PR #8, opened 2018-12-21
Share the app directory across multiple programs 

cburgdorf:
### What was wrong?

The `trinity-beacon` adds support for a new type of client that we need to support running side by side with the existing Ethereum 1 client. This client will have entirely different data, logs, ipc and pid files. 

Based on the discussion in https://github.com/ethereum/py-evm/pull/1603 we want to support running this client out of the same application directory as the Ethereum 1 client which means the application directory layout needs to be modified to support this.

### How was it fixed?

This change introduces multi-app support to trinity by introducing a new `app_identifier` config to the `TrinityConfig`.

Assuming there Ethereum 1 client uses an `app_identifier` of `""`(empty string)  and the Ethereum 2 beacon node uses an `app_identifier` of `beacon` the directory layout looks as follows.

```
/home/cburgdorf/.local/share/trinity
├── mainnet                       <-- one mainnet directory, but app specific sub directories
│   ├── chain
│   │   ├── full
|   |   |     └── db files
│   │   └── light
|   |         └── db files
│   ├── chain-beacon
│   │   ├── full
|   |   |     ├── beacon
│   │   │     |     └── db files
|   |   |     ├── shard-1
│   │   │     |     └── db files
|   |   |     └── shard-n
│   │   │           └── db files
│   │   └── light
|   |         ├── beacon
│   │         |     └── db files
|   |         ├── shard-1
│   │         |     └── db files
|   |         └── shard-n
│   │               └── db files
│   ├── logs
│   │   ├── trinity.log
│   │   ├── ...
│   │   ├── trinity.log.7
│   ├── logs-beacon
│   │   └── trinity.log
│   │   ├── ...
│   │   └── trinity.log.7
│   ├── ipc
│   │   ├── db.ipc
│   │   └── ...
│   ├── ipc-beacon
│   │   ├── db.ipc
│   │   └── ...
│   ├── pid
│   │   ├── database.pid
│   │   └── ...
│   ├── pid-beacon
│   │   ├── database.pid
│   │   └── ...
│   └── nodekey
|
└── ropsten
    ├── chain
    │   ├── full
    |   |     └── db files
    │   └── light
    |         └── db files
    ├── chain-beacon
    │   ├── full
    |   |     ├── beacon
    │   │     |     └── db files
    |   |     ├── shard-1
    │   │     |     └── db files
    |   |     └── shard-n
    │   │           └── db files
    │   └── light
    |         ├── beacon
    │         |     └── db files
    |         ├── shard-1
    │         |     └── db files
    |         └── shard-n
    │               └── db files
    ├── logs
    │   ├── trinity.log
    │   ├── ...
    │   ├── trinity.log.7
    ├── logs-beacon
    │   └── trinity.log
    │   ├── ...
    │   └── trinity.log.7
    ├── ipc
    │   ├── db.ipc
    │   └── ...
    ├── ipc-beacon
    │   ├── db.ipc
    │   └── ...
    ├── pid
    │   ├── database.pid
    │   └── ...
    ├── pid-beacon
    │   ├── database.pid
    │   └── ...
    └── nodekey
```

#### Cute Animal Picture

### DIFF
--- a/trinity/config.py
+++ b/trinity/config.py
@@ -37,7 +37,11 @@ from p2p.constants import (
 from trinity.constants import (
     ASSETS_DIR,
     DEFAULT_PREFERRED_NODES,
+    IPC_DIR,
+    LOG_DIR,
+    LOG_FILE,
     MAINNET_NETWORK_ID,
+    PID_DIR,
     ROPSTEN_NETWORK_ID,
     SYNC_FULL,
     SYNC_LIGHT,
@@ -47,7 +51,6 @@ from trinity._utils.chains import (
     get_data_dir_for_network_id,
     get_database_socket_path,
     get_jsonrpc_socket_path,
-    get_logfile_path,
     get_nodekey_path,
     load_nodekey,
 )
@@ -72,7 +75,6 @@ if TYPE_CHECKING:
     from trinity.chains.light import LightDispatchChain  # noqa: F401
 
 DATABASE_DIR_NAME = 'chain'
-
 
 MAINNET_EIP1085_PATH = ASSETS_DIR / 'eip1085' / 'mainnet.json'
 ROPSTEN_EIP1085_PATH = ASSETS_DIR / 'eip1085' / 'ropsten.json'
@@ -216,18 +218,19 @@ class TrinityConfig:
 
     def __init__(self,
                  network_id: int,
+                 app_identifier: str="",
                  genesis_config: Dict[str, Any]=None,
                  max_peers: int=25,
                  trinity_root_dir: str=None,
                  data_dir: str=None,
                  nodekey_path: str=None,
-                 logfile_path: str=None,
                  nodekey: PrivateKey=None,
                  sync_mode: str=SYNC_FULL,
                  port: int=30303,
                  use_discv5: bool = False,
                  preferred_nodes: Tuple[KademliaNode, ...]=None,
                  bootstrap_nodes: Tuple[KademliaNode, ...]=None) -> None:
+        self.app_identifier = app_identifier
         self.network_id = network_id
         self.max_peers = max_peers
         self.sync_mode = sync_mode
@@ -276,9 +279,6 @@ class TrinityConfig:
         elif nodekey is not None:
             self.nodekey = nodekey
 
-        if logfile_path is not None:
-            self.logfile_path = logfile_path
-
     def get_chain_config(self) -> ChainConfig:
         # the `ChainConfig` object cannot be pickled so we can't cache this
         # value since the TrinityConfig is sent across process boundaries.
@@ -286,6 +286,10 @@ class TrinityConfig:
             return ChainConfig.from_preconfigured_network(self.network_id)
         else:
             return ChainConfig.from_eip1085_genesis_config(self.genesis_config)
+
+    @property
+    def app_suffix(self) -> str:
+        return "" if len(self.app_identifier) == 0 else f"-{self.app_identifier}"
 
     @property
     def sync_mode(self) -> str:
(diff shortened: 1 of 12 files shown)


cburgdorf:
This has been approved before the repo migration (See https://github.com/ethereum/py-evm/pull/1654)

tests/trinity/conftest.py
```diff
@@ -50,12 +50,27 @@ def xdg_trinity_root(monkeypatch, tmpdir):
...
     Ensure proper test isolation as well as protecting the real directories.
     """
-    dir_path = tmpdir.mkdir('trinity')
-    monkeypatch.setenv('XDG_TRINITY_ROOT', str(dir_path))
+    trinity_root_dir = str(tmpdir.mkdir('trinity'))
+
+    # The default path that pytest generates are too long to be allowed as
+    # IPC Paths (hard UNIX rule). We are shorten them from something like:
+    # /tmp/pytest-of-<username>/pytest-88/<test-name>_command1_0/trinity
+    # to /tmp/pyt-<username>/88/<test-name>_command1_0/trinity
```
pipermerriam: What do you think about us just taking everything between the `/tmp/` prefix and the `trinity` suffix and running it though a quick *small* hash function.

`/tmp/fancy-pytest/test-something/trinity` becomes `/tmp/{hashlib.md5("fancy-pytest/test-something").hexdigest()}/trinity`

Benefits I see are:

- fixed length identifier
- safe against future changes to how pytest structures these directories
- reasonably strong resistance to collisions

trinity/config.py
```diff
@@ -287,6 +287,10 @@ def get_chain_config(self) -> ChainConfig:
         else:
             return ChainConfig.from_eip1085_genesis_config(self.genesis_config)
 
+    @property
+    def app_suffix(self) -> str:
+        return "" if len(self.app_identifier) == 0 else f"-{self.app_identifier}"
```
pipermerriam: Should this logic be updated now that we aren't using any empty identifiers (i.e. should we disallow an empty app identifier?)

tests/trinity/conftest.py
cburgdorf: I think the argument against this (which if I recall correctly @carver once brought up when we first experienced this issue) is that if you really need to dig into this data (because you are debugging) then you'll have a hard time finding it.

The (shortened) pytest scheme uses incremential numbers so if you are debugging a bunch of test runs you'll have an easy time finding what you are looking for.

Also, **if** we would simply use the `/tmp/<hash>/trinity` scheme, why would not just use a UUID but hash the proposed pytest path instead? Is it so that you'll at least theoretically would be able to drive the hash path from the originally proposed path? (Which still would have pretty bad ergonomics for debugging sessions)

trinity/config.py
cburgdorf: Yep, I was wondering that to. We could throw if `TrinityConfig`is instantiated with a zero length `app_identifier`. I guess that would lead to a bunch of tests that maybe need to be updated but if we want to disallow empty `app_identifier` that's fine with me. I think I'm neutral about this.

cburgdorf: I think I slightly prefer continuing support for empty `app_identifier` just for the better ergonomics. I could be convinced otherwise but I didn't want it to hold back the PR from getting merged.

tests/trinity/conftest.py
cburgdorf: I merged this now but if you feel strongly, feel free to discuss this further and then I'm happy to make the change in another PR.

pipermerriam: I don't feel strongly.  UUID instead of hash seems fine.  Also, I realized that *if* unix is ok with using an IPC path `/some/short-path/jsonrpc.ipc` where `/some/short-path` is actually a symlink to the original path then we might have an optimal solution.

veox:
This directory structure change warrants a warning in the release notes (change log?..).

Just stepped on this rake: my `chain` dir is on a separate (faster, non-RAID) drive than the rest of the stuff (keys, logs, PID files...) in the hierarchy. Pulling in `master` started syncing from scratch. I only really noticed because the debug log (which also had a path change) was no longer scrolling.

cburgdorf:
@veox yes, absolutely. Unfortunately we are lacking a good process for change log generation to keep track of these things.

What I did now is, I created a milestone for the next release and put a reminder in the description of the milestone that I hope we will pay attention to when we cut the next release.

https://github.com/ethereum/trinity/milestone/1

veox:
JIC, issue on process: #101.

(GitLab's approach linked there would've worked in this case, it seems.)

PR #11, opened 2018-12-21
Add async variants for some iter tools

jannikluhn:
### What was wrong?

Working with asynchronous iterators can be painful without the utilities in `itertools` and `toolz`.

### How was it fixed?

Implemented async versions of

- `chain`
- `cons`
- `take`
- `sliding_window`

The functions work for both synchronous and asynchronous iterables, in particular to allow for mixing them (e.g. `chain`).

(Motivation: https://github.com/ethereum/py-evm/pull/1650#discussion_r243006657)

#### Cute Animal Picture

### DIFF
--- a/trinity/_utils/async_iter.py
+++ b/trinity/_utils/async_iter.py
@@ -6,13 +6,10 @@ from typing import (
     Set,
     Tuple,
     TypeVar,
-    Union,
 )
 
 
 T = TypeVar('T')
-
-AnyIterable = Union[Iterable[T], AsyncIterable[T]]
 
 
 async def contains_all(async_gen: AsyncIterable[str], keywords: Set[str]) -> bool:
@@ -32,28 +29,27 @@ async def contains_all(async_gen: AsyncIterable[str], keywords: Set[str]) -> boo
     return False
 
 
-async def async_iter_wrapper(iterable: AnyIterable[T]) -> AsyncIterator[T]:
-    if isinstance(iterable, Iterable):
-        for item in iterable:
-            yield item
-    elif isinstance(iterable, AsyncIterable):
-        async for item in iterable:
-            yield item
+async def async_iterator(iterable: Iterable[T]) -> AsyncIterator[T]:
+    for item in iterable:
+        yield item
 
 
-async def async_chain(*iterables: AnyIterable[T]) -> AsyncIterator[T]:
-    for iterable in iterables:
-        async for item in async_iter_wrapper(iterable):
+async def async_chain(*async_iterables: AsyncIterable[T]) -> AsyncIterator[T]:
+    for async_iterable in async_iterables:
+        async for item in async_iterable:
             yield item
 
 
 def async_cons(first_item: T,
-               iterable: AnyIterable[T]) -> AsyncIterator[T]:
-    return async_chain([first_item], iterable)
+               async_iterable: AsyncIterable[T]) -> AsyncIterator[T]:
+    return async_chain(
+        async_iterator([first_item]),
+        async_iterable,
+    )
 
 
 def async_take(n: int,
-               iterable: AnyIterable[T]) -> AsyncIterator[T]:
+               async_iterable: AsyncIterable[T]) -> AsyncIterator[T]:
     if n < 0:
         raise ValueError("Number of elements to take must be non-negative")
 
@@ -62,7 +58,7 @@ def async_take(n: int,
             return
 
         yielded_items = 0
-        async for item in async_iter_wrapper(iterable):
+        async for item in async_iterable:
             yield item
 
             yielded_items += 1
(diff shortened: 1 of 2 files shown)


trinity/_utils/async_iter.py
```diff
@@ -19,3 +31,70 @@
...
+async def async_iter_wrapper(iterable: AnyIterable[T]) -> AsyncIterator[T]:
+    if isinstance(iterable, Iterable):
+        for item in iterable:
+            yield item
+    elif isinstance(iterable, AsyncIterable):
+        async for item in iterable:
+            yield item
+
+
+async def async_chain(*iterables: AnyIterable[T]) -> AsyncIterator[T]:
```
cburgdorf: I think there's a small trade off between consistency and ergonomics here. If all these `async_*` functions would work strictly on `AsyncIterable[T]` instead of `AnyIterable[T]` that would feel slightly more consistent to me. It makes things less ergonomic on the user side (or maybe mostly for the tests? :thinking: ).

I'm not against accepting `AnyIterable[T]` per se. Just pointing it out because in the past we have decided against abstracting away `sync` / `async` differences under one common API.

jannikluhn: Good point, I'm fine with making most of them only work with `AsyncIterable` for the sake of consistency. But `async_chain` is a bit of a special case as it can chain synchronous and asynchronous iterables together which can be useful.

trinity/_utils/async_iter.py
```diff
@@ -19,3 +27,67 @@
...
+        yield item
+
+
+async def async_chain(*async_iterables: AsyncIterable[T]) -> AsyncIterator[T]:
+    for async_iterable in async_iterables:
+        async for item in async_iterable:
+            yield item
+
+
+def async_cons(first_item: T,
```
pipermerriam: TIL `cons` is what I should have been using in a bunch of places where I've done `concatv([thing], other_things)`

trinity/_utils/async_iter.py
```diff
@@ -19,3 +27,67 @@
...
+def async_take(n: int,
+               async_iterable: AsyncIterable[T]) -> AsyncIterator[T]:
+    if n < 0:
+        raise ValueError("Number of elements to take must be non-negative")
+
+    async def async_generator() -> AsyncIterator[T]:
+        if n == 0:
+            return
+
+        yielded_items = 0
```
pipermerriam: Might be improved by using an `itertoolz.count()`

```python
counter = itertools.count(1)
async for item in async_iterable:
    yield item
    if counter.next() >= n:
        break
```

trinity/_utils/async_iter.py
```diff
@@ -19,3 +27,67 @@
...
+        async for item in async_iterable:
+            yield item
+
+            yielded_items += 1
+            if yielded_items >= n:
+                break
+
+    return async_generator()
```
pipermerriam: Why do we need this `async_generator` function?  Can't we just make `async_take` be a generator?

trinity/_utils/async_iter.py
```diff
@@ -19,3 +27,67 @@
...
+
+def async_cons(first_item: T,
+               async_iterable: AsyncIterable[T]) -> AsyncIterator[T]:
+    return async_chain(
+        async_iterator([first_item]),
+        async_iterable,
+    )
+
+
+def async_take(n: int,
```
pipermerriam: Can we rename to `num_items`.  :-1: on single letter variable name.

trinity/_utils/async_iter.py
```diff
@@ -19,3 +27,67 @@
...
+    if window_size < 0:
+        raise ValueError("size of sliding window must be non-negative")
+
+    async def async_generator() -> AsyncIterator[Tuple[T, ...]]:
+        if window_size == 0:
+            return
+
+        d = collections.deque(
+            [item async for item in async_take(window_size, async_iterable)],
+            maxlen=window_size,
```
pipermerriam: This is quite the clever way to do this.  I didn't know about this parameter for `deque`

trinity/_utils/async_iter.py
jannikluhn: We could do that as well, but then the parameter validation at the top would only run when the generator is first used, not already when it's initialized. In most cases that wouldn't matter very much. However, when the generator is only initialized but never used (for instance because it's zipped with something empty) then calling it with wrong arguments would not throw an exception which feels wrong.

pipermerriam: I didn't think about that part and I'm :+1: with keeping it as is, however, I think it's worth adding code comment(s) to explain why so that a future reader doesn't change it for the same reasons.  That is, unless there are tests which excercise this functionality in which case it's good-to-go as-is.

pipermerriam:
@jannikluhn closing as it looks like `aiostreams` is the way to go here.

Issue #13, opened 2018-12-20
Trinity prints noisy error message when being port-scanned


veox:
* py-evm Version: `trinity-v0.1.0-alpha.20-71-g8841a698` (commit 8841a6980254b93294c80a889ed3901fd6baf9c6, slightly past release)
* OS: Arch Linux
* Python Version (python --version): 3.7.1
* Environment (output of `pip freeze`): MISSING

### What is wrong?

Running `trinity` as

``` sh
% nice ionice -c3 trinity --network-id=1 --sync-mode=light --discv5 $ETHSTATS_OPTIONS
```

and then checking if the port is accessible with

``` sh
% nc -vz localhost 30303
```

results in this error message being printed:

``` python
   ERROR  12-20 17:58:10           LightServer  Unexpected error handling handshake
Traceback (most recent call last):
  File "/home/veox/src/py-evm/trinity/server.py", line 185, in receive_handshake
    await self._receive_handshake(reader, writer)
  File "/home/veox/src/py-evm/trinity/server.py", line 199, in _receive_handshake
    ip, socket, *_ = writer.get_extra_info("peername")
TypeError: cannot unpack non-iterable NoneType object
```

Not sure if fresh or regression.

### How can it be fixed

Not sure.

Checking how "regular" sync mode reacts to this and handles it would be a good start.

carver:
Theoretically fixed by #104

PR #104, opened 2018-12-21
Handle errors on incoming connections

pipermerriam:
migrated from: https://github.com/ethereum/py-evm/pull/1661

### What was wrong?

fixes: https://github.com/ethereum/py-evm/issues/1657

When an incoming connection's socket doesn't have address information an exception escapes the Trinity process.

Also, we end up leaving the TCP transports open.

### How was it fixed?

- Adds handling for the case where the incoming connection doesn't have the proper address info and log what data we can about the connection.
- Adds handling for the error cases while receiving the handshake to close the TCP transport.

#### Cute Animal Picture

441ab30cbf83ea412ec0ebb34bec6fc0

### DIFF
--- a/trinity/server.py
+++ b/trinity/server.py
@@ -232,7 +232,7 @@ class BaseServer(BaseService, Generic[TPeerPool]):
         try:
             ephem_pubkey, initiator_nonce, initiator_pubkey = decode_authentication(
                 msg, self.privkey)
-        except DecryptionError:
+        except DecryptionError as non_eip8_err:
             # Try to decode as EIP8
             got_eip8 = True
             msg_size = big_endian_to_int(msg[:2])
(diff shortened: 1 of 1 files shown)


trinity/server.py
```diff
@@ -214,9 +231,10 @@ def _make_request_server(self) -> BaseRequestServer:
             try:
                 ephem_pubkey, initiator_nonce, initiator_pubkey = decode_authentication(
                     msg, self.privkey)
-            except DecryptionError as e:
-                self.logger.debug("Failed to decrypt handshake: %s", e)
-                return
+            except DecryptionError as err:
+                raise HandshakeFailure(f"Failed to decrypt handshake: {err}")
```
carver: If you like the idea of getting more information about *both* decryption failures:

- Maybe add a name to the first failure in line 223, like `except DecryptionError as non_eip8_err`, then here `raise ... from non_eip8_err`, and explicitly mention EIP8 in this exception message. 

- Alternatively, something like `raise HandshakeFailure(f"Failed to decrypt handshake as EIP8: {err}, and without: {non_eip8_err}").

veox:
Fixes #13. ;)

Issue #76, opened 2018-11-19
Add command line flag for doing a full "regular" sync


pipermerriam:
### What is wrong?

It should be possible to perform a "regular" sync. Meaning sequential download of blocks and sequential application of those blocks to build the chain state.

### How can it be fixed

First, we need to fix an internal naming issue where we refer to our *fast* sync using `SYNC_FULL`.  This should be renamed to `SYNC_FAST`, and use the *fast* terminology in place of *full* everywhere that is currently used.

Then, add a new option for `SYNC_FULL` which just does a regular block-by-block sync.  The cli should support the option `--sync-mode=light/full/fast` with `fast` being the default.

The `trinity.sync` module will need to be re-organized as well, moving `trinity.sync.full` to `trinity.sync.fast` and adding the new syncer under `trinity.sync.full`.

PR #107, opened 2018-12-23
Fix sync mode naming: change all of the SYNC_FULL to SYNC_FAST

glaksmono:
### What was wrong?

We need to fix an internal naming issue where we refer to our fast sync using `SYNC_FULL`. This should be renamed to `SYNC_FAST`, and use the fast terminology in place of full everywhere that is currently used.

Issue Reference: https://github.com/ethereum/trinity/issues/76

### How was it fixed?

Change the name from `SYNC_FULL` to `SYNC_FAST`

#### Cute Animal Picture

### DIFF
--- a/tests/core/configuration/test_trinity_config_object.py
+++ b/tests/core/configuration/test_trinity_config_object.py
@@ -29,7 +29,7 @@ def test_trinity_config_computed_properties(xdg_trinity_root):
 
     assert trinity_config.network_id == 1
     assert trinity_config.data_dir == data_dir
-    assert trinity_config.database_dir == data_dir / DATABASE_DIR_NAME / "full"
+    assert trinity_config.database_dir == data_dir / DATABASE_DIR_NAME / "fast"
     assert trinity_config.nodekey_path == get_nodekey_path(data_dir)
 
 
@@ -42,7 +42,7 @@ def test_trinity_config_computed_properties_custom_xdg(tmpdir, xdg_trinity_root)
 
     assert trinity_config.network_id == 1
     assert trinity_config.data_dir == data_dir
-    assert trinity_config.database_dir == data_dir / DATABASE_DIR_NAME / "full"
+    assert trinity_config.database_dir == data_dir / DATABASE_DIR_NAME / "fast"
     assert trinity_config.nodekey_path == get_nodekey_path(data_dir)
 
 
--- a/trinity/config.py
+++ b/trinity/config.py
@@ -40,6 +40,7 @@ from trinity.constants import (
     MAINNET_NETWORK_ID,
     ROPSTEN_NETWORK_ID,
     SYNC_FULL,
+    SYNC_FAST,
     SYNC_LIGHT,
 )
 from trinity._utils.chains import (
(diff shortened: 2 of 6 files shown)


tests/core/configuration/test_trinity_config_object.py
```diff
@@ -42,7 +42,7 @@ def test_trinity_config_computed_properties_custom_xdg(tmpdir, xdg_trinity_root)
 
     assert trinity_config.network_id == 1
     assert trinity_config.data_dir == data_dir
-    assert trinity_config.database_dir == data_dir / DATABASE_DIR_NAME / "full"
+    assert trinity_config.database_dir == data_dir / DATABASE_DIR_NAME / "fast"
```
cburgdorf: This change doesn't make sense to me (but that may be *my* misunderstanding). In my world, from the perspective of persistence, there only exists `light` or `full`. The term `fast` refers to a specific syncing scheme but that doesn't change how the data is stored. E.g, you can sync trinity in `fast` mode and then change to a `full` sync later on, it would still be the same database.

trinity/config.py
```diff
@@ -381,6 +382,8 @@ def database_dir(self) -> Path:
         """
         if self.sync_mode == SYNC_FULL:
             return self.data_dir / DATABASE_DIR_NAME / "full"
+        elif self.sync_mode == SYNC_FAST:
+            return self.data_dir / DATABASE_DIR_NAME / "fast"
```
cburgdorf: See my comment above, why would we write into a different database when we switch from `fast` to `full` sync? We should still be writing to the same database, no?

trinity/config.py
```diff
@@ -524,6 +527,10 @@ def from_parser_args(cls,
     def is_light_mode(self) -> bool:
         return self.trinity_config.sync_mode == SYNC_LIGHT
 
+    @property
+    def is_fast_mode(self) -> bool:
+        return self.trinity_config.sync_mode == SYNC_FAST
```
cburgdorf: At this point, I'd like to restate what @carver said a few PRs back and propose to turn this into an enum.

trinity/config.py
```diff
@@ -536,7 +543,7 @@ def node_class(self) -> Type['Node']:
         from trinity.nodes.full import FullNode
         from trinity.nodes.light import LightNode
 
-        if self.is_full_mode:
+        if self.is_full_mode or self.is_fast_mode:
```
cburgdorf: I think we should split things up. In my world, there are only two main operation modes: `light` and `full`. This is why we only have a `FullNode` and a `LightNode`. These two things also dictate (afaik) where to put the database. 

And then there's another mode which is the **sync mode**. The specific *sync* mode whether it is `fast`, `full` or potentially `warp` doesn't change the general operational mode. No matter if we are syncing `fast`, `full` or `warp` we would always create a `FullNode` here and we would always persist into the same db. The only reason to return a `LightNode` here and to write into a different db would be if we would operate a light node which to me is like an entirely different operational mode and could even have its own different sub syncing strategies.

tests/core/configuration/test_trinity_config_object.py
pipermerriam: I concur/agree

trinity/config.py
pipermerriam: No opposition from me :+1:

cburgdorf:
I'm going to look into this today to potentially move it forward.

cburgdorf:
I looked at it and slipped down a rabbit whole of refactoring. I stopped and put out https://github.com/ethereum/trinity/pull/123 so that we have some code to discuss before things move forward.

cburgdorf:
Sorry, this turned out to be a rabbit whole deeper than initially expected. Thanks for kicking this off but we are going forward with https://github.com/ethereum/trinity/pull/123

PR #113, opened 2019-01-05
Allow `DEBUG2`-logging in sub-protocols + nitpick debug messages + various

veox:
### What was wrong?

* Sub-protocols couldn't use `DEBUG2` logging level;
* `DEBUG`-level logging was a bit too noisy, and displaying `int`/`bytes` where hex-strings make more sense;
* --help has not been updated for the TRACE -> DEBUG2 rename.

### How was it fixed?

Major:

* ETH and BCC sub-protocols now also derive from `HasExtendedDebugLogger`;
* `ETH/Status` and `BCC/Status` sub-protocol-level logging was moved to level `DEBUG2`.

Minor:

* A few debug messages now have `to_hex()`;
* `trinity --help` output for log levels now generated on-the-fly;

Extra:

* peer (IP) is printed in generic "Disconnecting" message;
* update date in `LICENSE` file.

Detailed reasoning is in commit messages.

### Cute Animal Picture

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -578,7 +578,7 @@ class BasePeer(BaseService):
             raise ValueError(
                 f"Reason must be an item of DisconnectReason, got {reason}"
             )
-        self.logger.debug("Disconnecting from remote peer; reason: %s", reason.name)
+        self.logger.debug("Disconnecting from remote peer %s; reason: %s", self.remote, reason.name)
         self.base_protocol.send_disconnect(reason.value)
         self.close()
 
--- a/trinity/protocol/eth/proto.py
+++ b/trinity/protocol/eth/proto.py
@@ -42,7 +42,7 @@ if TYPE_CHECKING:
     from .peer import ETHPeer  # noqa: F401
 
 
-# HasExtendedDebugLogger must come first so there's self.logger.debug2()
+# HasExtendedDebugLogger must come before Protocol so there's self.logger.debug2()
 class ETHProtocol(HasExtendedDebugLogger, Protocol):
     name = 'eth'
     version = 63
--- a/trinity/protocol/eth/proto.py
+++ b/trinity/protocol/eth/proto.py
@@ -1,4 +1,3 @@
-import logging
 from typing import (
     List,
     Tuple,
@@ -37,11 +36,14 @@ from .commands import (
     Transactions,
 )
 
+from trinity._utils.logging import HasExtendedDebugLogger
+
 if TYPE_CHECKING:
     from .peer import ETHPeer  # noqa: F401
 
 
-class ETHProtocol(Protocol):
+# HasExtendedDebugLogger must come first so there's self.logger.debug2()
+class ETHProtocol(HasExtendedDebugLogger, Protocol):
     name = 'eth'
     version = 63
     _commands = [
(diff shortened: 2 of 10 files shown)


p2p/peer.py
```diff
@@ -578,7 +578,7 @@ def _disconnect(self, reason: DisconnectReason) -> None:
             raise ValueError(
                 f"Reason must be an item of DisconnectReason, got {reason}"
             )
-        self.logger.debug("Disconnecting from remote peer; reason: %s", reason.name)
+        self.logger.debug(f"Disconnecting from remote peer {self.remote}; reason: {reason.name}")
```
pipermerriam: This should be converted back to `%` style string formatting.

https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#formatting-log-messages

trinity/protocol/eth/proto.py
```diff
@@ -62,7 +60,6 @@ def send_handshake(self, chain_info: ChainInfo) -> None:
             'genesis_hash': chain_info.genesis_hash,
         }
         cmd = Status(self.cmd_id_offset)
-        self.logger.debug("Sending ETH/Status msg: %s", resp)
```
pipermerriam: Rather than remove why not move to `TRACE` level?

p2p/peer.py
veox: Right! Will do tomorrow - will check other places where `f"{stuff}"` was introduced, too.

trinity/protocol/eth/proto.py
veox: I tried logging at `debug2` in https://github.com/ethereum/trinity/pull/113/commits/25990d86a7f3345ff4e18a46c02410b7c5686a74, but then https://github.com/ethereum/trinity/pull/113/commits/3ad83509b3bdab2392d120b662adc2efaaea8e12 happened.

In all honesty, I got agitated. Could review `Protocol` to have `debug2` included instead.

pipermerriam: > Could review Protocol to have `debug2` included in its `Logger` instead.

I'm :+1: on this route.

veox: This might actually be an issue in `mypy`:

https://github.com/python/mypy/issues/5973

veox: `trinity/proto` sub-protocols should now be able to use `logger.debug()`.

veox:
@jannikluhn @hwwhww @ChihChengLiang I've taken the liberty of touching BCC (not just ETH) sub-protocol in this PR, in commit f46efa8cb9671736643e4eb86fc98739a67b0b63. (For ETH, it was 363bb8c20975e0cb9a21b2320ec9f11695c5cdcf.)

The `"Sending BCC/Status msg"` log line will now appear on level `DEBUG2` (instead of previous `DEBUG`), and from logger `trinity.protocol.eth.proto.BCCProtocol` (i.o.p. `p2p.bcc.BCCProtocol`).

Do ping if this change is undesired, it's (still) easily-revertable.

veox:
@pipermerriam I think this change - allowing `logger.debug2()` - makes the PR unsquashable. It'd make unrelated changes inseparable. :(

Issue #79, opened 2018-11-20
Event bus all-the-things


pipermerriam:
### What is wrong?

Our use of the proxy object pattern has had continued fallout and un-intended results.  Specific cases that I'm thinking about are:

- #1299 has been difficult due to issues with pickle-ability
- Weird APIs where we are *kind-of* interacting with a chain object, but not really since it's actually over a proxy object.
- Weird APIs where we wrap the non-async Chain methods to make them async.
- Issues with the JSON-RPC server and it needing direct access to the Chain object which produces awkward coupling/dependencies

### How can it be fixed

I think we need to go ahead and get everything over onto an explicit event bus pattern.  Please discuss.

PR #121, opened 2019-01-08
Christoph/feat/rpc beacon modules

cburgdorf:
### What was wrong?

RPC Modules were locked to depend on `BaseAsyncChain` .

### How was it fixed?

1. `RPCModule` is genreic over `TChain`
2. RPCServer dropped dependency on the chain so that all it cares about is the `name` which means it is satisfied by the new non-generic `BaseRPCModule`.
3. The chain reset mechanism is now event bus based (making it possible to drop the `chain` dependency on the `RPCServer`

#### Cute Animal Picture

### DIFF
--- a/trinity/rpc/modules/__init__.py
+++ b/trinity/rpc/modules/__init__.py
@@ -1,6 +1,10 @@
 from .main import (  # noqa: F401
+    BaseRPCModule,
+    BeaconRPCModule,
+    ChainReplacementEvent,
+    Eth1RPCModule,
     initialize_modules,
-    RPCModule
+    RPCModule,
 )
 
 from .beacon import Beacon  # noqa: F401
--- a/trinity/rpc/modules/main.py
+++ b/trinity/rpc/modules/main.py
@@ -43,16 +43,16 @@ class BaseRPCModule(ABC):
 class RPCModule(BaseRPCModule, Generic[TChain]):
 
     def __init__(self, chain: TChain, event_bus: Endpoint) -> None:
-        self._chain = chain
-        self._event_bus = event_bus
+        self.chain = chain
+        self.event_bus = event_bus
 
-        self._event_bus.subscribe(
+        self.event_bus.subscribe(
             ChainReplacementEvent,
             lambda ev: self.set_chain(ev.chain)
         )
 
     def set_chain(self, chain: TChain) -> None:
-        self._chain = chain
+        self.chain = chain
 
 
 Eth1RPCModule = RPCModule['BaseAsyncChain']
--- a/trinity/rpc/modules/main.py
+++ b/trinity/rpc/modules/main.py
@@ -3,40 +3,66 @@ from abc import (
     ABC,
 )
 from typing import (
+    Any,
+    Generic,
     Iterable,
     Type,
+    TypeVar,
+    TYPE_CHECKING,
 )
 
 from eth_utils import (
     to_tuple,
 )
 from lahja import (
+    BaseEvent,
     Endpoint
 )
 
-from trinity.chains.base import BaseAsyncChain
+if TYPE_CHECKING:
+    from trinity.chains.base import BaseAsyncChain  # noqa: F401
 
 
-class RPCModule(ABC):
-    _chain = None
+TChain = TypeVar('TChain')
 
-    def __init__(self, chain: BaseAsyncChain, event_bus: Endpoint) -> None:
-        self._chain = chain
-        self._event_bus = event_bus
+
+class ChainReplacementEvent(BaseEvent, Generic[TChain]):
+
+    def __init__(self, chain: TChain):
+        self.chain = chain
+
+
+class BaseRPCModule(ABC):
 
     @property
     @abstractmethod
     def name(self) -> str:
         pass
 
-    def set_chain(self, chain: BaseAsyncChain) -> None:
+
+class RPCModule(BaseRPCModule, Generic[TChain]):
+
+    def __init__(self, chain: TChain, event_bus: Endpoint) -> None:
+        self._chain = chain
+        self._event_bus = event_bus
+
+        self._event_bus.subscribe(
+            ChainReplacementEvent,
+            lambda ev: self.set_chain(ev.chain)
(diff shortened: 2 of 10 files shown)


cburgdorf:
The JSON-RPC Server already runs in it's own isolated process and the `Chain` is a *local* chain with only the DB ending up being proxied through another process as that's a current requirement that comes from using `LevelDB`.

As [I said](https://github.com/ethereum/trinity/issues/79#issuecomment-440224770) in #79, I don't think we should funnel all db read/writes through the event bus and instead focus on moving to something like RocksDB to get rid of remaining process hoping (caused by the db proxy to leveldb).

I believe funneling all db access through the eventbus will almost certainly have a worse performance than going through the db proxy, because:

- The native Python proxy code is much more likely to be already heavily optimized for performance whereas the EventBus is still super young alpha software

- The EventBus is optimized for convenience and e.g. uses multiple process hops internally whereas the db proxy has a more direct connection.

cburgdorf:
Oh on

>This leaves us with option A which requires us to create a new service that exposes Chain data over the event bus.

I want to highlight that just exposing the chain data through another service doesn't buy us anything here because that other service would still need to talk to the `DbProxy` unless you want to create one dedicated service for all db access as a replacement for the `DbProxy` which I had assumed with my previous comment.

So, if we just add another service to expose Chain access you end up with: `JSON-RPC process` -> `Chain process` -> `DbProxy` whereas in this PR it is just `JSON-RPC process` -> `DbProxy` (since the chain is already a local one).

pipermerriam:
:+1: on the correction that this is *less* about the `Chain` and more about the core database access which basically means my whole comment can be disregarded.  There's one section I want to comment on but I'll do it separately.

trinity/rpc/modules/main.py
```diff
@@ -3,40 +3,66 @@
...
+
+Eth1RPCModule = RPCModule['BaseAsyncChain']
+BeaconRPCModule = RPCModule[Any]
 
 
 @to_tuple
-def initialize_modules(modules: Iterable[Type[RPCModule]],
-                       chain: BaseAsyncChain,
-                       event_bus: Endpoint) -> Iterable[RPCModule]:
+def initialize_modules(modules: Iterable[Type[RPCModule[TChain]]],
```
pipermerriam: I'm curious about us re-working this API to remove the `chain` component and instead have the signature be something like `initialize_modules(module_classes, trinity_config, event_bus)`.  This should allow any module that needs `Chain` access to connect to the chain itself, which I *think* also means that we could drop the `Generic`.

pipermerriam: > The `chain` is what most of the modules want to access anyway.

There are plenty of modules that don't need the `chain` which is why I want to get the `Chain` requirement fully removed from the `RPCModule` concept.

As for testing and needing to inject a `TestChain` there are plenty of ways we can solve that problem.

### RPC APIs that don't need the `Chain`

- `net`
- `admin` from https://github.com/ethereum/go-ethereum/wiki/Management-APIs
- `shh` (Whisper)
- `pss` (PSS that Swarm uses)
- `bzz` if we ever get swarm implemented.

cburgdorf: >There are plenty of modules that don't need the chain which is why I want to get the `Chain` requirement fully removed from the `RPCModule` concept.

Ah ok. Well, I kinda removed the `Chain` already by creating an `BaseRPCModule` that is basically just an interface that defines an abstract `name` property.

```
class BaseRPCModule(ABC):

    @property
    @abstractmethod
    def name(self) -> str:
        pass
```

That's the only thing RPC modules need to implement to work with the `RPCServer` because `BaseRPCModule` is the only thing the server cares about. 

But I realize that I missed the opportunity to make that more clear by providing one class `RPCModule` that uses a chain and using that class even for modules that don't need the chain (e.g. `net`).

I'm going to iterate on that and will further comment after I refactored some things to be clearer.

trinity/rpc/modules/__init__.py
```diff
@@ -14,5 +29,14 @@
...
 
 
-ETH1_RPC_MODULES = (Eth, EVM, Net, Web3)
-BEACON_RPC_MODULES = (Beacon,)
+@to_tuple
+def initialize_eth1_modules(chain: BaseAsyncChain, event_bus: Endpoint) -> Iterable[BaseRPCModule]:
+    yield Eth(chain, event_bus)
+    yield EVM(chain, event_bus)
+    yield Net(event_bus)
+    yield Web3()
```
cburgdorf: ↑
The `BaseRPCModule` has *zero* implementation and only defines an abstract `name` (so more of an interface than a base class). We initialize every module with the dependencies that the specific module needs.

trinity/rpc/modules/main.py
```diff
@@ -40,29 +35,20 @@ def name(self) -> str:
         pass
 
 
-class RPCModule(BaseRPCModule, Generic[TChain]):
+class ChainBasedRPCModule(BaseRPCModule, Generic[TChain]):
```
cburgdorf: This is just to have a DRY base class for all the modules that *do depend* on a chain. No part of the RPC system cares about it except the module itself. Every other code place just deals with `BaseRPCModule`.

trinity/rpc/modules/net.py
```diff
@@ -1,10 +1,15 @@
...
 from trinity.constants import TO_NETWORKING_BROADCAST_CONFIG
 from trinity.nodes.events import NetworkIdRequest
-from trinity.rpc.modules import Eth1RPCModule
+from trinity.rpc.modules import BaseRPCModule
+
 
+class Net(BaseRPCModule):
 
-class Net(Eth1RPCModule):
+    def __init__(self, event_bus: Endpoint):
```
cburgdorf: The `Net` module even defines its very own `__init__` as it needs an event bus but no chain.

trinity/rpc/modules/main.py
cburgdorf: @pipermerriam I pushed another commit for some cleanups and added inline comments to explain.

PR #123, opened 2019-01-09
Refactor handling of sync mode

cburgdorf:
### What was wrong?

There's some ambiguity around the `sync_mode` that this PR tries to address on the way towards #76 

### How was it fixed?

- move `database_dir` off the `TrinityConfig` (into the app configs)
- move `sync_mode` off the `TrinityConfig`
- provide the `sync_mode` via plugin so that we can can decide whether to use it for Eth1, Beacon, Validator etc.
- get rid of `--light` shortcut as proposed in #76
- move the db-shell into a plugin (to not replicate the args handling)

#### Cute Animal Picture


cburgdorf:
@glaksmono @pipermerriam this is an incomplete attempt to move #107 forward. It gets hairy quickly which is why I'm stopping here to get some feedback before I dive in deeper. I'll make some inline comments.

trinity/config.py
```diff
@@ -369,20 +370,6 @@ def data_dir(self) -> Path:
     def data_dir(self, value: str) -> None:
         self._data_dir = Path(value).resolve()
 
-    @property
-    def database_dir(self) -> Path:
-        """
-        Path where the chain database will be stored.
```
cburgdorf: The `database_dir` **for the Ethereum 1 node** should only be either `full` or `light` independently of the specific sync mode (e.g `fast`, `full`) . Considering that this may look totally different for the beacon node it seems to make more sense to move it into the `Eth1AppConfig` (which @pipermerriam had pointed out in a previous PR as well)

trinity/initialization.py
```diff
@@ -26,9 +26,6 @@ def is_data_dir_initialized(trinity_config: TrinityConfig) -> bool:
     if not os.path.exists(trinity_config.data_dir):
         return False
 
-    if not os.path.exists(trinity_config.database_dir):
-        return False
-
```
cburgdorf: That's the part that I dislike quite a bit. If we move the `database_dir` off the `TrinityConfig` we can't easily check and initialize it from here.

trinity/main.py
```diff
@@ -164,6 +168,32 @@ def launch_node(args: Namespace, trinity_config: TrinityConfig, endpoint: Endpoi
...
 
 
+@setup_cprofiler('run_database_process')
+@with_queued_logging
+def run_database_process(trinity_config: TrinityConfig, db_class: Type[BaseDB]) -> None:
+    with trinity_config.process_id_file('database'):
+        app_config = trinity_config.get_app_config(Eth1AppConfig)
+
+        if not app_config.database_dir.exists():
+            app_config.database_dir.mkdir(parents=True, exist_ok=True)
```
cburgdorf: This whole `run_database_process` was moved from a shared space into the eth1 space. Notice that we do the `database_dir` check and initialization here now.

trinity/main_beacon.py
```diff
@@ -62,33 +52,10 @@ def trinity_boot(args: Namespace,
 
     event_bus.start()
 
-    # First initialize the database process.
```
cburgdorf: The whole database process bootstrapping was heavily coupled to Eth1 and wasn't working anyway yet. I removed it for now without replacement to reduce the scope of this PR.

trinity/bootstrap.py
```diff
@@ -229,27 +221,6 @@ def display_launch_logs(trinity_config: TrinityConfig) -> None:
     logger.info("Trinity DEBUG log file is created at %s", str(trinity_config.logfile_path))
 
 
-@setup_cprofiler('run_database_process')
-@with_queued_logging
-def run_database_process(trinity_config: TrinityConfig, db_class: Type[BaseDB]) -> None:
-    with trinity_config.process_id_file('database'):
-        base_db = db_class(db_path=trinity_config.database_dir)
```
cburgdorf: This used to be something that was shared between Eth1 and beacon node but:

1. it depends on `trinity_config.database_dir` so something had to be done
2. It's heavily coupled to Eth1 anyway as it stands.

trinity/cli_parser.py
```diff
@@ -248,8 +249,8 @@ def __call__(self,
 mode_parser = syncing_parser.add_mutually_exclusive_group()
 mode_parser.add_argument(
     '--sync-mode',
-    choices={SYNC_LIGHT, SYNC_FULL},
-    default=SYNC_FULL,
+    choices={SYNC_LIGHT, SYNC_FAST, SYNC_FULL},
+    default=SYNC_FAST,
```
cburgdorf: Maybe the entire `--sync-mode` should move off the `TrinityConfig`. I guess it is best to assume sync modes between Eth1 and Beacon would be different. But that means we need to touch the parser as well

cburgdorf: Ah, ok. That touches an even bigger box than what I meant. Today, we already have the beacon node separated by an entirely different command (`trinity-beacon`). But yep, I also already thought about that longer term we may wanna switch to `trinity <eth1|beacon|validator> <whatever-follows>`.

What I actually meant is that today the `sync-mode` is a parameter that is hardcoded in the parser (with it's available options) and hence applies to both `trinity` and `trinity-beacon` in the same way.

I will take a look how to decouple that. It would actually be nice if the sync mode would be open for extensions. E.g. if someone could write a warp sync plugin it would be good if that plugin could cause `warp` to become a valid value for `sync-mode` only when that plugin is loaded.

trinity/initialization.py
```diff
@@ -112,3 +109,8 @@ def initialize_database(chain_config: ChainConfig,
         chaindb.get_canonical_head()
     except CanonicalHeadNotFound:
         chain_config.initialize_chain(base_db)
+
+
+def ensure_eth1_dirs(app_config: Eth1AppConfig) -> None:
+    if not app_config.database_dir.exists():
```
cburgdorf: I found splitting this into `is_initialized` / `initialize` pairs a bit too much for my taste here which is why I went for the `ensure` pattern where we just create things on the fly that are missing.

pipermerriam: I suspect you're correct that these operations can be made idempotent that we can convert to a single `ensure_` pattern.

trinity/plugins/builtin/json_rpc/plugin.py
```diff
@@ -65,15 +65,15 @@ def setup_eth1_modules(self, trinity_config: TrinityConfig) -> Tuple[Eth1ChainRP
...
             header_db = db_manager.get_headerdb()  # type: ignore
             event_bus_light_peer_chain = EventBusLightPeerChain(self.context.event_bus)
             chain = chain_config.light_chain_class(header_db, peer_chain=event_bus_light_peer_chain)
-        elif eth1_app_config.is_full_mode:
+        elif eth1_app_config.uses_full_db:
             db = db_manager.get_db()  # type: ignore
             chain = chain_config.full_chain_class(db)
         else:
-            raise NotImplementedError(f"Unsupported mode: {trinity_config.sync_mode}")
+            raise Exception(f"Unreachable code")
```
carver: Wasn't the old warning more meaningful here, if we add a new db type and forget to add a new `if` branch?

cburgdorf: The `uses_full_db` / `uses_light_db` property pair is mutually exclusive.

```
    @property
    def uses_full_db(self) -> bool:
        return not self.uses_light_db
```

Even if we'd add another sync mode such as `warp`, `uses_full_mode` would still be `True` and `uses_light_mode` would be `False` leaving the `else` clause as unreachable code.

However, I added another commit that gets rid of both boolean properties and turns them into an enum `database_mode` and then the `else` clause can reference that.

cburgdorf:
@pipermerriam @carver Do you think we can merge this? I'm in the process of digging out the previously started RocksDB work in #173 and since this PR touches the db handling in respect to the config it would be convenient to have this in master. (I could rebase #173 on top of this but I already have #155 depending on this PR so I'd like to push this forward)

cburgdorf:
>+1 the SyncModeParserPlugin strikes me as a little awkward but it's not a blocker.

It **is** awkward haha.

 >Basically I assume that it's a step towards having Eth1.0 syncing be fully plugin based and that it will get rolled up into something more comprehensive at some later point?

Exactly, in fact, it is already getting a little refactored and more useful with #155

Issue #127, opened 2019-01-09
Doesn't connect to preferred node when discovery is on


lithp:
* trinity Version: b78c3287

This is actually two issues bundled into one.

When trinity is run with a command like:

    trinity --preferred-node enode://long-key@whatever --disable-discovery

We get a lot of lines like this:

    WARNING  01-09 13:12:59.825           ETHPeerPool  [0x42ba] Discovery did not answer PeerCandidateRequest in time

This is understandable, Discovery is not answering at all :)

The second problem occurs when you drop the `--disable-discovery` flag and just use `--preferred-node`. As far as I can tell from looking at the logs, trinity never attempts to connect to the preferred node!

Some ideas:
- When `--disable-discovery` is passed `DiscoveryService` should not run discovery in the background but should still answer requests, using only the set of bootstrap and preferred peers.
- When `--disable-discovery` is passed `BasePeerPool` should not attempt to talk to it, and should instead use just the set of preferred peers? This feels like a much worse solution.
- I'm not sure why ETHPeerPool never ends up trying to connect to the preferred peer. Maybe when discovery is on it never gets a chance because it's too busy trying to connect to other peers. This feels like it's related to why trinity has a hard time finding peers so it's probably worth looking into.

lithp:
#129 fixes the first problem, I still need to look into the second problem.

veox:
> I'm not sure why ETHPeerPool never ends up trying to connect to the preferred peer. Maybe when discovery is on it never gets a chance because it's too busy trying to connect to other peers.

`trinity` now DEBUG-logs reasons why it doesn't connect to the preferred node, e.g.:

```
   DEBUG  04-14 13:02:34        SQLitePeerInfo  skipping <Node(0x053d@79.98.29.154)>, it failed because "TooManyPeersFailure" and is not usable until 2019-04-14 13:03:52
```

carver:
Solved by https://github.com/ethereum/trinity/pull/436 ?

carver:
AFAIK, this is fixed. I have been regularly running in this mode (discovery on, with a preferred peer).

PR #129, opened 2019-01-09
Stop --disable-discovery from emitting WARNINGs

lithp:
Part of #127

TODO:
- [ ] Add some kind of test that this works?

When --disable-discovery was set it turn the discovery
plugin completely turned itself off. However, PeerPool still expected
something to respond to `PeerCandidatesRequest` and `RandomBootnodeRequest`
messages, so the log was full of WARNINGS complaining about unanswered
messages.

This commit adds `NoopDiscoveryService` which is used when the
--disable-discovery flag is passed. It answers every message with the
empty set.

frozen-whiskers

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -955,6 +955,38 @@ class DiscoveryByTopicProtocol(DiscoveryProtocol):
         return tuple(seen_nodes)
 
 
+class NoopDiscoveryService(BaseService):
+    'A stub "discovery service" which does nothing'
+
+    def __init__(self, event_bus: Endpoint, token: CancelToken = None) -> None:
+        super().__init__(token)
+        self._event_bus = event_bus
+
+    async def handle_get_peer_candidates_requests(self) -> None:
+        async for event in self._event_bus.stream(PeerCandidatesRequest):
+            self.logger.debug("Servicing request for more peer candidates")
+
+            self._event_bus.broadcast(
+                event.expected_response_type()(tuple()),
+                event.broadcast_config()
+            )
+
+    async def handle_get_random_bootnode_requests(self) -> None:
+        async for event in self._event_bus.stream(RandomBootnodeRequest):
+            self.logger.debug("Servicing request for boot nodes")
+
+            self._event_bus.broadcast(
+                event.expected_response_type()(tuple()),
+                event.broadcast_config()
+            )
+
+    async def _run(self) -> None:
+        self.run_daemon_task(self.handle_get_peer_candidates_requests())
+        self.run_daemon_task(self.handle_get_random_bootnode_requests())
+
+        await self.cancel_token.wait()
+
+
 class DiscoveryService(BaseService):
     _last_lookup: float = 0
     _lookup_interval: int = 30
@@ -969,9 +1001,6 @@ class DiscoveryService(BaseService):
         self.port = port
         self._event_bus = event_bus
         self._lookup_running = asyncio.Lock()
-
-        self.run_daemon_task(self.handle_get_peer_candidates_requests())
-        self.run_daemon_task(self.handle_get_random_bootnode_requests())
 
     async def handle_get_peer_candidates_requests(self) -> None:
         async for event in self._event_bus.stream(PeerCandidatesRequest):
@@ -998,6 +1027,9 @@ class DiscoveryService(BaseService):
             )
 
     async def _run(self) -> None:
+        self.run_daemon_task(self.handle_get_peer_candidates_requests())
+        self.run_daemon_task(self.handle_get_random_bootnode_requests())
+
         await self._start_udp_listener()
         self.run_task(self.proto.bootstrap())
         await self.cancel_token.wait()
--- a/trinity/plugins/builtin/peer_discovery/plugin.py
+++ b/trinity/plugins/builtin/peer_discovery/plugin.py
@@ -23,6 +23,7 @@ from p2p.discovery import (
     DiscoveryByTopicProtocol,
     DiscoveryProtocol,
     DiscoveryService,
+    NoopDiscoveryService,
     PreferredNodeDiscoveryProtocol,
 )
 from p2p.kademlia import (
@@ -91,8 +92,12 @@ class DiscoveryBootstrapService(BaseService):
     Bootstrap discovery to provide a parent ``CancellationToken``
     """
 
-    def __init__(self, event_bus: Endpoint, trinity_config: TrinityConfig) -> None:
+    def __init__(self,
+                 disable_discovery: bool,
+                 event_bus: Endpoint,
+                 trinity_config: TrinityConfig) -> None:
         super().__init__()
+        self.is_discovery_disabled = disable_discovery
         self.event_bus = event_bus
         self.trinity_config = trinity_config
 
@@ -120,12 +125,18 @@ class DiscoveryBootstrapService(BaseService):
                 self.cancel_token,
             )
 
-        discovery_service = DiscoveryService(
-            discovery_protocol,
-            self.trinity_config.port,
-            self.event_bus,
-            self.cancel_token,
-        )
+        if self.is_discovery_disabled:
+            discovery_service: BaseService = NoopDiscoveryService(
+                self.event_bus,
+                self.cancel_token,
+            )
+        else:
+            discovery_service = DiscoveryService(
+                discovery_protocol,
+                self.trinity_config.port,
+                self.event_bus,
+                self.cancel_token,
+            )
 
         await discovery_service.run()
 
@@ -140,8 +151,7 @@ class PeerDiscoveryPlugin(BaseIsolatedPlugin):
         return "Peer Discovery"
 
     def on_ready(self) -> None:
-        if not self.context.args.disable_discovery:
-            self.start()
+        self.start()
 
     def configure_parser(self, arg_parser: ArgumentParser, subparser: _SubParsersAction) -> None:
         arg_parser.add_argument(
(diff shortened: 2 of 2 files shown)


p2p/discovery.py
```diff
@@ -955,6 +955,45 @@ def get_nodes_to_connect(self, count: int) -> Iterator[kademlia.Node]:
...
+            event_bus: Endpoint,
+            token: CancelToken = None) -> None:
+        super().__init__(token)
+        self._event_bus = event_bus
+
+        self.preferred_nodes = tuple(to_uris(preferred_nodes))
+        self.logger.info('Using static set of peers: %s', self.preferred_nodes)
+
+        self.run_daemon_task(self.handle_get_peer_candidates_requests())
+        self.run_daemon_task(self.handle_get_random_bootnode_requests())
```
lithp: This pattern is strange for two reasons:
- It's a lot of boilerplace! for what are just RPCs. Maybe either lahja or `BaseService` could make this easier, so that all we could have to do is declare a mapping from `PeerCandidatesRequest` to `handle_get_peer_candidates_request` and write a function which responds to a message.
- We're implementing the same interface that DiscoveryService implements, but there's no abc to ensure that's true!

p2p/discovery.py
```diff
@@ -955,6 +955,45 @@ def get_nodes_to_connect(self, count: int) -> Iterator[kademlia.Node]:
...
+        async for event in self._event_bus.stream(RandomBootnodeRequest):
+            self.logger.debug("Servicing request for boot nodes")
+
+            self._event_bus.broadcast(
+                event.expected_response_type()(self.preferred_nodes),
+                event.broadcast_config()
+            )
+
+    async def _run(self) -> None:
+        await self.cancel_token.wait()
```
lithp: Maybe `BaseService._run` could do this for us?

trinity/plugins/builtin/peer_discovery/plugin.py
```diff
@@ -120,12 +125,21 @@ def __init__(self, event_bus: Endpoint, trinity_config: TrinityConfig) -> None:
                 self.cancel_token,
             )
 
-        discovery_service = DiscoveryService(
-            discovery_protocol,
-            self.trinity_config.port,
-            self.event_bus,
-            self.cancel_token,
-        )
+        discovery_service: Union[DiscoveryService, StaticDiscoveryService] = None
```
cburgdorf: you can get rid of that line if you just give a little type hint further down.

```
discovery_service: BaseService = StaticDiscoveryService(
                self.trinity_config.preferred_nodes,
                self.event_bus,
                self.cancel_token,
            )
```

The reason for that is, that mypy does not automatically see both as `BaseService` (even if they are) but defaults to infer the actual type of each. Just giving a small type hint solves this.

trinity/plugins/builtin/peer_discovery/plugin.py
```diff
@@ -91,8 +93,11 @@ class DiscoveryBootstrapService(BaseService):
     Bootstrap discovery to provide a parent ``CancellationToken``
     """
 
-    def __init__(self, event_bus: Endpoint, trinity_config: TrinityConfig) -> None:
+    def __init__(
+            self, disable_discovery: bool, event_bus: Endpoint,
+            trinity_config: TrinityConfig) -> None:
```
cburgdorf: nitpick. Preferred style would be:

```python
def __init__(self, 
                   disable_discovery: bool, 
                   event_bus: Endpoint,
                   trinity_config: TrinityConfig) -> None:
```

cburgdorf:
>`--disable-discovery` was broken. When it is set the discovery
plugin completely turned itself off. However, `PeerPool` still expected
something to respond to `PeerCandidatesRequest` and `RandomBootnodeRequest`
messages

Mmh..I added the `--disable-discovery` when I turned the discovery into an isolated plugin basically just because it was easy to add and I thought it might be a useful convention to be able to disable every plugin that otherwise runs automatically. If only for easier debugging of certain cases. So, having the `PeerPool` run into warnings when `--disable-discovery` is passed was expected behavior to me.

But it seems the bigger problem is that connecting to preferred nodes doesn't work as expected even when the plugin is active. Maybe that is because of the way the `PreferredNodeDiscoveryProtocol` is implemented with its `preferred_node_recycle_time` :thinking: 

https://github.com/ethereum/trinity/blob/d14ad79bfb679d7a8430f5a73605c2c57e542fb0/p2p/discovery.py#L849-L851

I'm wondering if, instead of using the `StaticDiscoveryService` when `--disable-discovery` is passed, maybe instead we should use that with something like `--static-peer`?

So, the idea would be that `--disable-discovery` still continues to entirely disable the plugin (with all hard consequences) but if one wants to **only** connect to a static set of peers, it could be done with

```
trinity --static-peer enode://long-key@whatever
```

That would then bootstrap the `StaticDiscoveryService` instead of the regular one. 

Benefits:

- the plugin can still be deactivated entirely with `--disable-discovery`
- connecting to static nodes does not need the extra `--disable-discovery` flag but is handled implicitly when a node is passed with `--static-peer`.

Thoughts?

p2p/discovery.py
cburgdorf: >It's a lot of boilerplate

Agree! Feel free to propose improved ergonomics in another PR

>the type system has no idea

See my [comment](https://github.com/ethereum/trinity/pull/129#discussion_r246685253) on how to do that without introducing that extra `Union` type

p2p/discovery.py
```diff
@@ -955,6 +955,46 @@ def get_nodes_to_connect(self, count: int) -> Iterator[kademlia.Node]:
...
+            preferred_nodes: Sequence[kademlia.Node],
+            event_bus: Endpoint,
+            token: CancelToken = None) -> None:
+        super().__init__(token)
+        self._event_bus = event_bus
+
+        self.preferred_nodes = tuple(to_uris(preferred_nodes))
+        self.logger.info('Using static set of peers: %s', self.preferred_nodes)
+
+        self.run_daemon_task(self.handle_get_peer_candidates_requests())
```
pipermerriam: I don't think it's appropriate to do this in `__init__`. I believe we need any asyncio/task running should be scoped within the `_run` method.  Otherwise, there are no guarantees that these will be properly shutdown (since the service may never be run)

pipermerriam:
I think there is merit to the `--static-peer` approach so :+1: to exploring that.  I think that it *properly* reflects the two use cases

- preferred peers: part of discovery and are used first before using random discovery peers
- static peers: should probably be *injected* into the `PeerPool` and their connections *managed* in a separate mechanism.

This also hints at us needing an architecture that supports multiple peer pools since a *static peer* should not be subjected to the same rules as a random peer we connect to.  For example we probably want to retain a connection to a static peer even if they are a very poorly performing peer with whom we would normally disconnect.  Or alternatively, we want to maintain a connection to static peers even if our peer pool is full.

trinity/plugins/builtin/peer_discovery/plugin.py
lithp: Oops, great point!

lithp:
@cburgdorf ah, that makes sense. What's your vision for wanting to be able to disable the plugin? What would you like to be able to debug?

I like the idea of having a `--disable-discovery` flag. Geth has `--nodiscover` and parity has `--no-discovery`; it seems like a common use-case to want to disable kademlia. The current warnings make this flag far less useful though, this should be an expected configuration! Some potential solutions:

1) Move `--disable-discovery` into the top-level parser, and don't attempt to send messages to the plugin if we know it won't be running.
2) Add an API to lahja which lets you inspect subscriptions. If there are no subscriptions to `PeerCandidatesRequest`s then peer pool should detect that and not send any requests.
3a) `--disable-discovery` works almost exactly like it does in this PR, except instead of `StaticDiscoveryService` it enables a `NoopDiscoveryService` which always returns the empty set when asked for nodes.
3b) Add a flag or config file which lets you disable some/all plugins. The plugin discovery mechanism reads this flag/file and lets you do a hard shutdown of this plugin, including the warnings, if you really want that.

I'm kind of leaning for (3a) and implementing (3b) once it becomes necessary. What do you think?

p2p/discovery.py
lithp: The mypy was an example, but I think there's a deeper problem than that! Both `DiscoveryService` and `StaticDiscoveryService` expose the same interface: `PeerCandidatesRequest` and `RandomBootnodeRequest`, but _that_ doesn't show up in the type system anywhere. If I had implemented `PeerCandidatesRequest` but forgotten `RandomBootnodeRequest` then nothing would have complained until I started getting `WARNING`s.

cburgdorf:
>What's your vision for wanting to be able to disable the plugin? What would you like to be able to debug?

Nothing in particular. It was more like a feeling that it would be useful to disable it. But you bring up a valid point that these warning reduce the usefulness of that.

>I'm kind of leaning for (3a) and implementing (3b) once it becomes necessary.

I'm on board. :+1:

lithp:
I'm trying to incrementally add support for static peers, starting with https://github.com/ethereum/trinity/pull/151. While I do that I've changed this PR to have `--disable-discovery` just turn off the plugin. It's still around but it doesn't talk to the network and it answers all messages with the empty set.

trinity/plugins/builtin/peer_discovery/plugin.py
```diff
@@ -91,8 +92,12 @@ class DiscoveryBootstrapService(BaseService):
     Bootstrap discovery to provide a parent ``CancellationToken``
     """
 
-    def __init__(self, event_bus: Endpoint, trinity_config: TrinityConfig) -> None:
+    def __init__(self,
+                 disable_discovery: bool,
+                 event_bus: Endpoint,
+                 trinity_config: TrinityConfig) -> None:
         super().__init__()
+        self.disable_discovery = disable_discovery
```
pipermerriam: *minor nitpick*

Just noting that we have a *loose* convention for this to be `is_discovery_disabled` as that makes it a bit clearer that it is a boolean.  As it stands a naive read of the code could easily interpret this name as a method.

But it appears we've never written any of those conventions down....

lithp: It's a good convention though, pushed a change :+1:

lithp: Opened a PR for this convention against the style guide: https://github.com/ethereum/snake-charmers-tactical-manual/pull/35

PR #151, opened 2019-01-11
Add support for w3.admin.addPeer()

lithp:
You can now inject peers into the pool by calling the `w3.admin.addPeer()` endpoint.

Some future work?
- If you try to add a peer when the pool is full your request [will be silently dropped](https://github.com/ethereum/trinity/blob/master/p2p/peer.py#L985). That's not very user friendly.
- If you add a peer with the format `enode://the-key:[::]:30303` the RPC will be accepted, even though this is the wrong format, we don't know how to connect to `[::]`! It will then fail with a `DEBUG` rather than a more user friendly `INFO`.

I think fixing these requires turning this into an AddPeerRequest/Response pair. That might be the right thing to do eventually but I think this is good enough for now?

#### Cute Animal Picture

sheep

### DIFF
--- a/p2p/kademlia.py
+++ b/p2p/kademlia.py
@@ -29,6 +29,8 @@ from eth_keys import (
 
 from eth_hash.auto import keccak
 
+from trinity._utils.validation import validate_enode_uri
+
 k_b = 8  # 8 bits per hop
 
 k_bucket_size = 16
@@ -101,6 +103,7 @@ class Node:
 
     @classmethod
     def from_uri(cls, uri: str) -> 'Node':
+        validate_enode_uri(uri)  # Be no more permissive than the validation
         parsed = urlparse.urlparse(uri)
         pubkey = keys.PublicKey(decode_hex(parsed.username))
         return cls(pubkey, Address(parsed.hostname, parsed.port))
--- a/p2p/kademlia.py
+++ b/p2p/kademlia.py
@@ -20,6 +20,7 @@ from urllib import parse as urlparse
 from eth_utils import (
     big_endian_to_int,
     decode_hex,
+    remove_0x_prefix,
 )
 
 from eth_keys import (
@@ -110,8 +111,7 @@ class Node:
 
     def uri(self) -> str:
         hexstring = self.pubkey.to_hex()
-        if hexstring.startswith('0x'):
-            hexstring = hexstring[2:]
+        hexstring = remove_0x_prefix(hexstring)
         return f'enode://{hexstring}@{self.address.ip}:{self.address.tcp_port}'
 
     def __str__(self) -> str:
--- a/p2p/kademlia.py
+++ b/p2p/kademlia.py
@@ -109,7 +109,10 @@ class Node:
         return cls(pubkey, Address(parsed.hostname, parsed.port))
 
     def uri(self) -> str:
-        return f'enode://{self.pubkey.to_hex()}@{self.address.ip}:{self.address.tcp_port}'
+        hexstring = self.pubkey.to_hex()
+        if hexstring.startswith('0x'):
+            hexstring = hexstring[2:]
+        return f'enode://{hexstring}@{self.address.ip}:{self.address.tcp_port}'
 
     def __str__(self) -> str:
         return '<Node(%s@%s)>' % (self.pubkey.to_hex()[:6], self.address.ip)
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -91,6 +91,7 @@ from .constants import (
 )
 
 from .events import (
+    ConnectToNodeCommand,
     PeerCandidatesRequest,
     PeerCountRequest,
     PeerCountResponse,
@@ -798,6 +799,12 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
         self._subscribers: List[PeerSubscriber] = []
         self.event_bus = event_bus
 
+    async def accept_connect_commands(self) -> None:
+        async for command in self.event_bus.stream(ConnectToNodeCommand):
+            # TODO: This adds nodes in series, it won't handle a high rate of commands
+            self.logger.debug('Received request to connect to %s', command.node)
+            await self.connect_to_nodes(from_uris([command.node]))
+
     async def handle_peer_count_requests(self) -> None:
         async for req in self.event_bus.stream(PeerCountRequest):
                 # We are listening for all `PeerCountRequest` events but we ensure to only send a
@@ -925,6 +932,7 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
         if self.event_bus is not None:
             self.run_daemon_task(self.handle_peer_count_requests())
             self.run_daemon_task(self.maybe_connect_more_peers())
+            self.run_daemon_task(self.accept_connect_commands())
         self.run_daemon_task(self._periodically_report_stats())
         await self.cancel_token.wait()
 
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -801,9 +801,8 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
 
     async def accept_connect_commands(self) -> None:
         async for command in self.event_bus.stream(ConnectToNodeCommand):
-            # TODO: This adds nodes in series, it won't handle a high rate of commands
             self.logger.debug('Received request to connect to %s', command.node)
-            await self.connect_to_nodes(from_uris([command.node]))
+            self.run_task(self.connect_to_nodes(from_uris([command.node])))
 
     async def handle_peer_count_requests(self) -> None:
         async for req in self.event_bus.stream(PeerCountRequest):
--- a/tests/core/_utils/test_validation.py
+++ b/tests/core/_utils/test_validation.py
@@ -24,6 +24,9 @@ GOOD_KEY = (
         ('enode://public_key:[::]:30303', MALFORMED_URI),
         ('enode://nonhex@add', 'public key must be a 128-character hex string'),
         ('enode://00@[::]:30303', 'public key must be a 128-character hex string'),
+
+        # The following tests check for validations which are done by functions internal
+        # to validate_enode_uri
         (f'enode://{GOOD_KEY}@10:30303', "'10' does not appear to be an IPv4 or IPv6 address"),
         (f'enode://{GOOD_KEY}@[::]:3000000', "Port out of range 0-65535"),
         (f'enode://{GOOD_KEY}@[::/24]:30303', "Invalid IPv6 URL"),
--- a/tests/core/_utils/test_validation.py
+++ b/tests/core/_utils/test_validation.py
@@ -1,0 +1,67 @@
+import pytest
+
+from trinity._utils.validation import (
+    validate_enode_uri
+)
+
+
+MALFORMED_URI = 'enode string must be of the form "enode://public-key@ip:port"'
+GOOD_KEY = (
+    '717ad99b1e67a93cd77806c9273d9195807da1db38ecfc535aa207598a4bd599'
+    '645599a0d11f72c1107ba4fb44ab133b0188a2f0b47ff16ba12b0a5ec0350202'
+)
+
+
+@pytest.mark.parametrize(
+    'uri,message',
+    (
+        ('/:hi', MALFORMED_URI),
+        ('wrongscheme://public_key:[::]:30303', MALFORMED_URI),
+        ('enode://public_key:[::]:30303', MALFORMED_URI),
+        ('enode://nonhex@add', 'public key must be a 128-character hex string'),
+        ('enode://00@[::]:30303', 'public key must be a 128-character hex string'),
+        (f'enode://{GOOD_KEY}@10:30303', "'10' does not appear to be an IPv4 or IPv6 address"),
+        (f'enode://{GOOD_KEY}@[::]:3000000', "Port out of range 0-65535"),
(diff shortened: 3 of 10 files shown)


trinity/rpc/modules/admin.py
```diff
@@ -0,0 +1,26 @@
...
+        self.event_bus = event_bus
+
+    @property
+    def name(self) -> str:
+        return 'admin'
+
+    async def addPeer(self, node: str) -> None:
+
+        # Throw an exception to show the user if {node} has the wrong format
+        enode = Node.from_uri(node)  # noqa: F841
```
lithp: Looking at this again, I think I can remove the need for a `#noqa` by just not assigning to anything, I'll try that tomorrow.

pipermerriam: While functionally sound, this is relying on the side-effect of `Node.from_uri`.  Wrapping this in a small stand-alone function `validate_enode` and adding a small set of tests for that function would help future proof this from breakage if that side-effect changes at some point in the future.

lithp:
I've added a `validate_enode_uri` and written some tests for it.

@pipermerriam I looked through the existing tests to try to find what you meant by `it is something we should have all of the primitives necessary to test it.` but couldn't find anything which is already testing the json-rpcs. What did you have in mind?

I could write a test that calling the json-rpc endpoint sends the expected message to lahja? And maybe also a test that firing the message causes connect_to_node to be called? Or, I could do more of an integration tests and assert that a handshake is eventually send when the json-rpc endpoint is called?

lithp:
Well, just found `tests/core/json-rpc/test_ipc.py`

p2p/kademlia.py
```diff
@@ -101,12 +103,16 @@ def __init__(self, pubkey: datatypes.PublicKey, address: Address) -> None:
...
     def from_uri(cls, uri: str) -> 'Node':
+        validate_enode_uri(uri)  # Be no more permissive than the validation
         parsed = urlparse.urlparse(uri)
         pubkey = keys.PublicKey(decode_hex(parsed.username))
         return cls(pubkey, Address(parsed.hostname, parsed.port))
 
     def uri(self) -> str:
-        return f'enode://{self.pubkey.to_hex()}@{self.address.ip}:{self.address.tcp_port}'
+        hexstring = self.pubkey.to_hex()
+        if hexstring.startswith('0x'):
```
lithp: This is always true but I wanted something which was self-documenting. Maybe I could replace it with an `assert hexstring.startswith('0x')`?

tests/core/p2p-proto/test_server.py
```diff
@@ -161,3 +163,28 @@ def get_server(privkey, address):
...
+    event_bus.broadcast(
+        ConnectToNodeCommand(RECEIVER_REMOTE.uri()),
+        TO_NETWORKING_BROADCAST_CONFIG
+    )
+
+    await asyncio.sleep(0.5)
+
+    assert len(server.peer_pool.connected_nodes) == 1
+
+    await initiator_peer_pool.cancel()
```
lithp: This line blocks for 5 seconds before timing out on the `accept_connect_commands` task. Now I see why https://github.com/ethereum/trinity/pull/166/files#r247747992 is necessary

lithp:
Okay, I've added a test that `PeerPool`s respect `ConnectToNodeCommand`s. I think this is ready for review again.

tests/core/p2p-proto/test_server.py
```diff
@@ -161,3 +163,28 @@ def get_server(privkey, address):
...
+@pytest.mark.asyncio
+async def test_peer_pool_answers_connect_commands(event_loop, event_bus, server):
+    # This is the PeerPool which will accept our message and try to connect to {server}
+    initiator_peer_pool = ParagonPeerPool(
+        privkey=INITIATOR_PRIVKEY,
+        context=ParagonContext(),
+        event_bus=event_bus,
+    )
+    asyncio.ensure_future(initiator_peer_pool.run(), loop=event_loop)
+    await initiator_peer_pool.events.started.wait()
```
lithp: I think this code is wrong, because the started event is fired before `_run()` completes:

https://github.com/ethereum/trinity/blob/e6f5e4c0cff011266a22503a4ce6f455cb2666f7/p2p/service.py#L101-L103

The symptom should be flaky tests? Or, in the best case, tests which suddenly stop working when something like the version of asyncio changes.

I think that the solution is to change BaseService; it feels useful to be able to do an await like this line. 
- `_run()` should not be allowed to block execution (to enforce this it times out in 0.5 seconds)
- `self.events.started.set()` should happen after it returns
- `BaseService` can take responsibility for calling `await self.cancel_token.wait()` right after the event is set.

p2p/kademlia.py
pipermerriam: You could use [`eth_utils.remove_0x_prefix`](https://eth-utils.readthedocs.io/en/latest/utilities.html#remove-0x-prefix-value-string) which is idempotent, removing the need for a conditional.

p2p/peer.py
```diff
@@ -798,6 +799,12 @@ def __init__(self,
         self._subscribers: List[PeerSubscriber] = []
         self.event_bus = event_bus
 
+    async def accept_connect_commands(self) -> None:
+        async for command in self.event_bus.stream(ConnectToNodeCommand):
+            # TODO: This adds nodes in series, it won't handle a high rate of commands
```
pipermerriam: If you adjusted the line below to be `self.run_task(self.connect_to_nodes(...))` then the connection would be offloaded into a background task allowing higher throughput..

tests/core/_utils/test_validation.py
```diff
@@ -0,0 +1,67 @@
...
+        (f'enode://{GOOD_KEY}@10:30303', "'10' does not appear to be an IPv4 or IPv6 address"),
+        (f'enode://{GOOD_KEY}@[::]:3000000', "Port out of range 0-65535"),
+        (f'enode://{GOOD_KEY}@[::/24]:30303', "Invalid IPv6 URL"),
+    ),
+)
+def test_validate_enode_failures(uri, message):
+    try:
+        validate_enode_uri(uri)
+    except ValueError as e:
+        assert message == str(e)
```
pipermerriam: The idiomatic way to do this is:

```python
with pytest.raises(ValueError, match=message):
    validate_enode_uri
```

tests/core/json-rpc/test_ipc.py
```diff
@@ -526,3 +527,74 @@ def mock_syncing(is_syncing, progress=None):
...
+    '645599a0d11f72c1107ba4fb44ab133b0188a2f0b47ff16ba12b0a5ec0350202'
+)
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    'description, request_msg, expected',
+    (
+        (
+            'Validation occurs',
```
pipermerriam: Instead of a `description` field `pytest` has a mechanism for assigning test names yourself:

https://docs.pytest.org/en/latest/example/parametrize.html#different-options-for-test-ids

tests/core/p2p-proto/test_server.py
pipermerriam: I'm :+1: on something like `starting/started` assuming we can find a clean mechanism.

There was another conversation at some point about refactoring our `BaseService` to use composition such that the `Service` class got passed a sort of `ServiceLogic` class that was responsible for the actual implementation, keeping the `Service` class in charge of running the `ServiceLogic` class.  I think these two things are *semi* orthogonal but the thing your trying to accomplish *may* be easier in the later architecture were you to want to take a stab at that.  cc @carver who was part of the original conversation.

And second, take a look at the `Peer` class and the `BootManager` mechanism there as it is seems somewhat similar to what you are wanting which is an event to say, "I'm both started **and** I've done all the things I need to do to be fully functional".

trinity/_utils/validation.py
```diff
@@ -42,3 +50,23 @@ def validate_transaction_call_dict(transaction_dict: Dict[str, Any], vm: BaseVM)
...
+
+
+def validate_enode_uri(enode: str, require_ip: bool = False) -> None:
+    parsed = urlparse.urlparse(enode)
+
+    if parsed.scheme != 'enode' or not parsed.username:
+        raise ValueError('enode string must be of the form "enode://public-key@ip:port"')
+
+    if not re.match('^[0-9a-fA-F]{128}$', parsed.username):
+        raise ValueError('public key must be a 128-character hex string')
```
pipermerriam: Be believe these should be converted to raise `eth_utils.ValidationError`

trinity/_utils/validation.py
```diff
@@ -42,3 +50,23 @@ def validate_transaction_call_dict(transaction_dict: Dict[str, Any], vm: BaseVM)
...
+
+    decoded_username = decode_hex(parsed.username)
+    ip = ipaddress.ip_address(parsed.hostname)
+
+    if require_ip and ip in (ipaddress.ip_address('0.0.0.0'), ipaddress.ip_address('::')):
+        raise ValueError('A concrete IP address must be specified')
+
+    # The rest of these already throw ValueError when something bad happens
+    keys.PublicKey(decoded_username)
```
pipermerriam: I'm a little uncomfortable with us using implicit validation mechanisms like this since this functionality isn't necessarily part of the `eth-keys` API but more of a side effect.  However, I can get over it as long as we 1) have tests which cover this code path and 2) we comment those tests to make it clear that they are to ensure that the implicit reliance on the internal validation mechanisms within `eth-keys` are still being enforced.

p2p/kademlia.py
lithp: Oh, that's much better :+1:

trinity/_utils/validation.py
lithp: Just pushed [a commit](https://github.com/ethereum/trinity/pull/151/commits/20d4fb7d61f487437f9e7ab3079e195496d5f139) which addresses your last two comments.

I'm not super happy with re-raising the errors to change their type but my rationale is:
- The function is a lot easier to use if you only have to think about `ValidationError` when using it. Where the exception came from is an implementation detail which should be abstracted away.
- The "right" answer might be to inline each of the checks which functions like urlparse are expected to perform (as is currently being done with `keys.PublicKey`), but that seems like an excessive amount of work, I think all the common cases are already accounted for.

trinity/_utils/validation.py
lithp: There's no way of writing a test to enforce this but I've inlined all the checks which `keys.PublicKey` performs, after looking at eth_keys I'm pretty sure it's not possible for it to throw a `ValueError`.

lithp: I've added a comment marking the tests which ensure these functions perform the validations we expect them to, I think that's a good idea :+1:

lithp:
Ready for review again, I'll tackle starting/started in a different PR.

p2p/kademlia.py
```diff
@@ -101,12 +104,15 @@ def __init__(self, pubkey: datatypes.PublicKey, address: Address) -> None:
...
     def from_uri(cls, uri: str) -> 'Node':
+        validate_enode_uri(uri)  # Be no more permissive than the validation
         parsed = urlparse.urlparse(uri)
         pubkey = keys.PublicKey(decode_hex(parsed.username))
         return cls(pubkey, Address(parsed.hostname, parsed.port))
 
     def uri(self) -> str:
-        return f'enode://{self.pubkey.to_hex()}@{self.address.ip}:{self.address.tcp_port}'
+        hexstring = self.pubkey.to_hex()
+        hexstring = remove_0x_prefix(hexstring)
```
pipermerriam: *nitpick*

We try to avoid overwriting variables like this.  This would typicall be written as something like:

```python
hexstring_with_prefix = self.pubkey.to_hex()
hexstring = remove_0x_prefix(hexstring_with_prefix)
```

In this case the only reason is to conform with the convention of not overwriting things.  The general reason for not re-using variable names is that in situations where there is an iterative process of *mungers* on some value, if we re-use the variable name then we destroy historical state which makes debugging harder.

So when we don't care about historical state we'll often do something like:

```python
value = pipe(base_val, munger_1, munger_2)
```

and when we want intermediate state:

```python
raw_val = get_raw_val()
val_state_a = munge_state_a(raw_val)
val_state_b = munge_state_b(val_state_a)
val = final_munge(val_state_b)
```

tests/core/p2p-proto/test_server.py
carver: > Why does that block for 5 seconds and how does it relate to the need for `wait_iter`?

Right, the a subtask of the peer pool service is waiting on the event bus stream. When the service is `cancel()`ed, it just waits for its tasks to die. If they don't die within a 5s timeout, they are killed more harshly.

The event bus stream doesn't shut itself down when the service is cancelled, but the `wait_iter()` exits immediately when the service is cancelled, so the task cleans up right away.

tests/core/p2p-proto/test_server.py
carver: Yeah, these are significant changes to how services work at the moment. `run()` is not a method that kicks off the service. It *is* the service, right now. It doesn't return until the service has completed.

I'm definitely excited to keep improving `BaseService`, but there are a *lot* of classes using it right now, so we may be better off adding some kind of boot mechanism (similar to the peer's boot manager) rather than modifying `run()` to exit early.

lithp: I spent some time looking into just how many services there are, it turns out [there are a lot](https://gist.github.com/lithp/f2753a3aca7240230a4e3fca5ec5ac11)

pipermerriam: I'd advocate for developing a parallel service concept were we to want to re-work anything fundamental, especially if it goes for the composition based approach

PR #154, opened 2019-01-11
Add routines for updating validator status

hwwhww:
### What was wrong?
#139 part 2: Add [routines for updating validator status](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#routines-for-updating-validator-status).

### How was it fixed?
- Add  `eth2/beacon/validator_status_helpers.py`
    - `activate_validator`
    - `initiate_validator_exit`
    - `exit_validator`
    - `penalize_validator`
    - `prepare_validator_for_withdrawal`

#### Cute Animal Picture

cute-pictures-of-animals-wearing-glasses-27 1

### DIFF
--- a/eth2/beacon/validator_status_helpers.py
+++ b/eth2/beacon/validator_status_helpers.py
@@ -4,6 +4,10 @@ from typing import (
     Sequence,
 )
 
+from eth_utils import (
+    ValidationError,
+)
+
 from eth2.beacon.enums import (
     ValidatorRegistryDeltaFlag,
     ValidatorStatusFlags,
@@ -25,11 +29,24 @@ from eth2.beacon.typing import (
 # Helper for updating tuple item
 #
 def update_tuple_item(tuple_data: Sequence[Any],
-                      index: ValidatorIndex,
+                      index: int,
                       new_value: Any) -> Iterable[Any]:
+    """
+    Update the ``index``th item of ``tuple_data`` to ``new_value``
+    """
     list_data = list(tuple_data)
-    list_data[index] = new_value
-    return tuple(list_data)
+
+    try:
+        list_data[index] = new_value
+    except IndexError:
+        raise ValidationError(
+            "the length of the given tuple_data is {}, the given index {} is out of index".format(
+                len(tuple_data),
+                index,
+            )
+        )
+    else:
+        return tuple(list_data)
 
 
 #
@@ -135,6 +152,10 @@ def _settle_penality_to_validator_and_whistleblower(
     Apply penality/reward to validator and whistleblower and update the meta data
 
     More intuitive pseudo-code:
+    current_epoch_penalization_index = (state.slot // EPOCH_LENGTH) % LATEST_PENALIZED_EXIT_LENGTH
+    state.latest_penalized_exit_balances[current_epoch_penalization_index] += (
+        get_effective_balance(state, index)
+    )
     whistleblower_index = get_beacon_proposer_index(state, state.slot)
     whistleblower_reward = get_effective_balance(state, index) // WHISTLEBLOWER_REWARD_QUOTIENT
     state.validator_balances[whistleblower_index] += whistleblower_reward
@@ -142,19 +163,19 @@ def _settle_penality_to_validator_and_whistleblower(
     validator.penalized_slot = state.slot
     """
     # Update `state.latest_penalized_exit_balances`
-    last_penalized_epoch = (state.slot // epoch_length) % latest_penalized_exit_length
+    current_epoch_penalization_index = (state.slot // epoch_length) % latest_penalized_exit_length
     effective_balance = get_effective_balance(
         state.validator_balances,
         validator_index,
         max_deposit,
     )
     penalized_exit_balance = (
-        state.latest_penalized_exit_balances[last_penalized_epoch] +
+        state.latest_penalized_exit_balances[current_epoch_penalization_index] +
         effective_balance
     )
     latest_penalized_exit_balances = update_tuple_item(
         tuple_data=state.latest_penalized_exit_balances,
-        index=last_penalized_epoch,
+        index=current_epoch_penalization_index,
         new_value=penalized_exit_balance,
     )
     state = state.copy(
@@ -193,6 +214,11 @@ def penalize_validator(state: BeaconState,
                        whistleblower_reward_quotient: int,
                        entry_exit_delay: int,
                        max_deposit: Ether) -> BeaconState:
+    """
+    Penalize the validator with the given ``index``.
+
+    Exit the validator, penalize the validator, and reward the whistleblower.
+    """
     state = exit_validator(state, index, entry_exit_delay)
     state = _settle_penality_to_validator_and_whistleblower(
         state=state,
@@ -206,6 +232,9 @@ def penalize_validator(state: BeaconState,
 
 
 def prepare_validator_for_withdrawal(state: BeaconState, index: ValidatorIndex) -> BeaconState:
+    """
+    Set the validator with the given ``index`` with ``WITHDRAWABLE`` flag.
+    """
     validator = state.validator_registry[index]
     validator = validator.copy(
         status_flags=validator.status_flags | ValidatorStatusFlags.WITHDRAWABLE
--- a/eth2/beacon/validator_status_helpers.py
+++ b/eth2/beacon/validator_status_helpers.py
@@ -1,0 +1,215 @@
+from typing import (
+    Any,
+    Iterable,
+    Sequence,
+)
+
+from eth2.beacon.enums import (
+    ValidatorRegistryDeltaFlag,
+    ValidatorStatusFlags,
+)
+from eth2.beacon.helpers import (
+    get_beacon_proposer_index,
+    get_effective_balance,
+)
+from eth2.beacon.types.states import BeaconState
+from eth2.beacon.types.validator_registry_delta_block import ValidatorRegistryDeltaBlock
+from eth2.beacon.typing import (
+    Ether,
+    SlotNumber,
+    ValidatorIndex,
+)
+
+
+#
+# Helper for updating tuple item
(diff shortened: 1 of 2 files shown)


eth2/beacon/validator_status_helpers.py
```diff
@@ -0,0 +1,215 @@
...
+
+    More intuitive pseudo-code:
+    whistleblower_index = get_beacon_proposer_index(state, state.slot)
+    whistleblower_reward = get_effective_balance(state, index) // WHISTLEBLOWER_REWARD_QUOTIENT
+    state.validator_balances[whistleblower_index] += whistleblower_reward
+    state.validator_balances[index] -= whistleblower_reward
+    validator.penalized_slot = state.slot
+    """
+    # Update `state.latest_penalized_exit_balances`
+    last_penalized_epoch = (state.slot // epoch_length) % latest_penalized_exit_length
```
djrtwo: The name here implies that it is an "epoch" but really it is an "index". What we are effectively naming is the index defined by the "current epoch" into the `latest_penalized_exit_balances` .

Maybe change to `current_epoch_penalization_index`.
^ that isn't a great name either but might be better (?) :/

eth2/beacon/validator_status_helpers.py
```diff
@@ -0,0 +1,215 @@
...
+    return state
+
+
+def penalize_validator(state: BeaconState,
+                       index: ValidatorIndex,
+                       epoch_length: int,
+                       latest_penalized_exit_length: int,
+                       whistleblower_reward_quotient: int,
+                       entry_exit_delay: int,
+                       max_deposit: Ether) -> BeaconState:
```
djrtwo: + docstring

eth2/beacon/validator_status_helpers.py
```diff
@@ -0,0 +1,215 @@
...
+        latest_penalized_exit_length=latest_penalized_exit_length,
+        whistleblower_reward_quotient=whistleblower_reward_quotient,
+        epoch_length=epoch_length,
+        max_deposit=max_deposit,
+    )
+    return state
+
+
+def prepare_validator_for_withdrawal(state: BeaconState, index: ValidatorIndex) -> BeaconState:
+    validator = state.validator_registry[index]
```
djrtwo: + docstring

tests/eth2/beacon/test_validator_status_helpers.py
```diff
@@ -0,0 +1,317 @@
...
+    ),
+    [
+        (True),
+        (False),
+    ]
+)
+def test_activate_validator(ten_validators_state, genesis, genesis_slot, entry_exit_delay):
+    state = ten_validators_state
+    index = 1
+    result_state = activate_validator(
```
djrtwo: Is it worth doing a sanity check that the `validator` at `index` is not active at the start?

tests/eth2/beacon/test_validator_status_helpers.py
```diff
@@ -0,0 +1,317 @@
...
+    )
+    if genesis:
+        state.validator_registry[index].activation_slot == genesis_slot
+    else:
+        state.validator_registry[index].activation_slot == state.slot + entry_exit_delay
+
+
+def test_initiate_validator_exit(ten_validators_state):
+    state = ten_validators_state
+    index = 1
```
djrtwo: Is it worth doing a sanity check that the `validator` at `index` does not have have `INITIATED_EXIT` already set?

tests/eth2/beacon/test_validator_status_helpers.py
```diff
@@ -0,0 +1,317 @@
...
+def test_exit_validator(num_validators,
+                        entry_exit_delay,
+                        committee,
+                        state_slot,
+                        exit_slot,
+                        validator_registry_exit_count,
+                        not_previous_exited,
+                        ten_validators_state):
+    # Unchanged
+
```
djrtwo: unnecessary new line

tests/eth2/beacon/test_validator_status_helpers.py
```diff
@@ -0,0 +1,317 @@
...
+    validator_registry_exit_count = 2
```
djrtwo: Why overwriting this var? It was specified as a param/arg above

djrtwo: Could set to 2 (or any other value) in the parametrize above

tests/eth2/beacon/test_validator_status_helpers.py
```diff
@@ -0,0 +1,317 @@
...
+    state_slot = 100
```
djrtwo: same as above. this is set in `parametrize`

tests/eth2/beacon/test_validator_status_helpers.py
```diff
@@ -0,0 +1,317 @@
...
+    state = ten_validators_state.copy(
+        slot=state_slot,
+        validator_registry_exit_count=validator_registry_exit_count,
+    )
+    index = 1
+
+    # Set validator exit_slot
```
djrtwo:
```suggestion
    # Set validator exit_slot prior to running `exit_validator`
```

tests/eth2/beacon/test_validator_status_helpers.py
```diff
@@ -0,0 +1,317 @@
...
+
+    ],
+)
+def test_exit_validator(num_validators,
+                        entry_exit_delay,
+                        committee,
+                        state_slot,
+                        exit_slot,
+                        validator_registry_exit_count,
+                        not_previous_exited,
```
djrtwo: Suggest not using a negative in a boolean var name. Makes reading conditionals difficult

tests/eth2/beacon/test_validator_status_helpers.py
```diff
@@ -0,0 +1,317 @@
...
+    state = state.update_validator_registry(
+        validator_index=index,
+        validator=validator,
+    )
+    result_state = exit_validator(
+        state=state,
+        index=index,
+        entry_exit_delay=entry_exit_delay,
+    )
+    if not_previous_exited:
```
djrtwo: I think the name of this conditional is wrong. It looks like this is checking _if_ the validator was previously exited and then ensuring that the state was not modified. (because a double exit is a no-op)

tests/eth2/beacon/test_validator_status_helpers.py
```diff
@@ -0,0 +1,317 @@
...
+    effective_balance = max_deposit * GWEI_PER_ETH
+    whistleblower_reward = (
+        effective_balance //
+        whistleblower_reward_quotient
+    )
+    whistleblower_index = get_beacon_proposer_index(state, state.slot, epoch_length)
+
+    assert (
+        (
+            state.validator_balances[whistleblower_index] -
```
djrtwo: Recommend assigning this to a separate var before the assertion -- `balance_difference` or something similar

tests/eth2/beacon/test_validator_status_helpers.py
hwwhww: Ohhh good catch! I forgot I'm using `ten_validators_state`... Thanks!

eth2/beacon/validator_status_helpers.py
```diff
@@ -0,0 +1,215 @@
...
+
+#
+# Helper for updating tuple item
+#
+def update_tuple_item(tuple_data: Sequence[Any],
+                      index: ValidatorIndex,
+                      new_value: Any) -> Iterable[Any]:
+    list_data = list(tuple_data)
+    list_data[index] = new_value
+    return tuple(list_data)
```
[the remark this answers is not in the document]
hwwhww: Benchmarked, it surprised me that the old approach is faster:

-  `update_tuple_item`: the current function with `list()` and `tuple()`

-  `update_tuple_item_2`
    ```python
    def update_tuple_item_2(tuple_data: Sequence[Any],
                            index: int,
                            new_value: Any) -> Iterable[Any]:
        assert len(tuple_data) > index
        return tuple_data[:index] + (new_value,) + tuple_data[index+1:]
    ```

- Test case
    ```python
    tuple_data = tuple(i for i in range(10000))
    def do_it():
        for i in range(10000):
            result_1 = update_tuple_item(
                tuple_data=tuple_data,
                index=i,
                new_value=10
            )
    ```

- Benchmark 5 times:
    - `update_tuple_item`:
        ```
        0.5724449157714844 sec
        0.5826718807220459 sec
        0.5666313171386719 sec
        0.5777051448822021 sec
        0.57430100440979 sec
        avg 0.5747508525848388 sec
        ```
    - `update_tuple_item_2`:
        ```
        0.766359806060791 sec
        0.7576439380645752 sec
        0.7507579326629639 sec
        0.7591519355773926 sec
        0.7523770332336426 sec
        avg 0.757258129119873 sec
        ```

Anything I missed?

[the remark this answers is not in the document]
hwwhww: Also I think we need to ensure `assert len(tuple_data) > index` in both approaches.

tests/eth2/beacon/test_validator_status_helpers.py
hwwhww: You're right! Removed `not_previous_exited` flag.

eth2/beacon/validator_status_helpers.py
hwwhww: I can't figure out other better name too... use `current_epoch_penalization_index` now. Thanks!

hwwhww:
@pipermerriam I'm merging it now, please let me know if there's any post-merge feedback and I will add them in the incoming part 3.

eth2/beacon/validator_status_helpers.py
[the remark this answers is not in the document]
pipermerriam: :+1: to verifying via benchmark, leaving as-is fine with me.

PR #156, opened 2019-01-11
Implement `BeaconChain`

hwwhww:
### What was wrong?
Implement `BeaconChain`.

### How was it fixed?

1. tests: extract codes from `test_demo` to `create_mock_signed_attestation`
2. Modify `BeaconStateMachine.__init__`: now takes `(chaindb, block, parent_block_class)`
3. BeaconChainDB: add `block_root` -> `slot` for accessing per-version `block_class`
    * In `_persist_block_chain`, also persist the lookup of `block_root` -> `slot`
    * Add `get_canonical_head_root` to get the root of canonical head
    * Add `get_slot_by_root` to lookup the `slot` by `block_root`
3. `BeaconBlock`: add `from_parent`
4. Add `BeaconChain`

#### Cute Animal Picture

640px-vombatus_ursinus_-maria_island_national_park


setup.py
```diff
@@ -73,7 +73,8 @@
         "lru-dict>=1.1.6",
         "mypy_extensions>=0.4.1,<1.0.0",
         "py-ecc>=1.4.7,<2.0.0",
-        "rlp==1.0.3",
+        "rlp>=1.1.0,<2.0.0",
+        "py-evm>=0.2.0a38",
```
pipermerriam: Probably best to do exact version pinning though I'm curious what happens if this version isn't the same as the version in the `'trinity'` section as it may not be possible.

setup.py
```diff
@@ -73,7 +73,8 @@
         "lru-dict>=1.1.6",
         "mypy_extensions>=0.4.1,<1.0.0",
```
pipermerriam: This dependency can be elevated to the `'trinity'` section as it's actual a core dependency that we only have implicitely because of `py-evm` installing it.

tests/eth2/beacon/chains/test_chain.py
```diff
@@ -0,0 +1,88 @@
...
+
+        assert block == valid_chain.get_canonical_block_by_slot(
+            block.slot
+        )
+        assert block.root == valid_chain.get_canonical_block_root(
+            block.slot
+        )
+        blocks += (block,)
+
+    valid_chain_2 = copy.deepcopy(valid_chain)
```
pipermerriam: My intuition says that this will also copy the underlying base database which means that this test *might* not be testing what you expect since you may just be re-importing blocks that are already in the chain.  Worth double checking that `valid_chain_2.get_canonical_head()` returns the genesis block as I assume is expected.

tests/eth2/beacon/conftest.py
```diff
@@ -694,3 +807,44 @@ def create_mock_signed_attestation(state,
         )
 
     return create_mock_signed_attestation
+
+
+@pytest.fixture
+def create_mock_block(privkeys, pubkeys):
+    def create_mock_block(state, block_class, parent_block, config, slot=None):
```
pipermerriam: Worth naming this differently with a `_` prefix or something so that you aren't colliding with the parent function name.

tests/eth2/beacon/conftest.py
```diff
@@ -694,3 +807,44 @@ def create_mock_signed_attestation(state,
...
+        if slot is None:
+            slot = state.slot
+
+        block = block_class.from_parent(
+            parent_block=parent_block,
+            slot=slot,
+        )
+
+        # Sign block
+        beacon_proposer_index = get_beacon_proposer_index(
```
pipermerriam: This block of code to sign a block looks like something that may deserve a utility function **or** even a formal API on the block class itself.

eth2/beacon/chains/base.py
```diff
@@ -0,0 +1,380 @@
+from __future__ import absolute_import
```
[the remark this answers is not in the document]
cburgdorf: @hwwhww this seems to be still in

eth2/beacon/chains/base.py
```diff
@@ -0,0 +1,380 @@
...
+)
+
+from eth._utils.datatypes import (
+    Configurable,
+)
+from eth._utils.rlp import (
+    validate_imported_block_unchanged,
+)
+
+from eth2.beacon.db.chain import (  # noqa: F401
```
cburgdorf: The `# noqa: F401` doesn't seem to be needed here

eth2/beacon/chains/base.py
```diff
@@ -0,0 +1,380 @@
...
+        self.chaindb = self.get_chaindb_class()(base_db)
+
+    #
+    # Helpers
+    #
+    @classmethod
+    def get_chaindb_class(cls) -> Type['BaseBeaconChainDB']:
+        if cls.chaindb_class is None:
+            raise AttributeError("`chaindb_class` not set")
+        return cls.chaindb_class
```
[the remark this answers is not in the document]
cburgdorf: I kinda wonder the opposite. Why don't we treat the `BaseBeaconChain` like a true interface (think: C#, Java, TypeScript interfaces) and keep it entirely free of implementations, with every single method being abstract?

eth2/beacon/chains/base.py
```diff
@@ -0,0 +1,380 @@
...
+        Returns the ``StateMachine`` instance for the given block slot number.
+        """
+        return cls.get_sm_class_for_block_slot(block.slot)
+
+    @abstractmethod
+    def get_sm(self, block: BaseBeaconBlock) -> 'BaseBeaconStateMachine':
+        pass
+
+    @classmethod
+    def get_sm_class_for_block_slot(cls, slot: SlotNumber) -> Type['BaseBeaconStateMachine']:
```
cburgdorf: I know that this is just following what `Py-EVM` does with the Eth1 code but related to my [comment](https://github.com/ethereum/trinity/pull/156/files#r249111618) I'm wondering why we don't treat this whole class as a pure interface without implementation (think Java, C#, TypeScript interfaces).

@pipermerriam made the opposite recommendation of moving more implementation from `BeaconChain` into `BaseBeaconChain` and I don't feel strongly about it but I think it's worth discussing.

eth2/beacon/chains/base.py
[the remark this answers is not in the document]
pipermerriam: I'm fine with the base class being fully abstract :+1:

eth2/beacon/chains/base.py
```diff
@@ -0,0 +1,380 @@
...
+    #
+    @classmethod
+    def get_sm_class(cls, block: BaseBeaconBlock) -> Type['BaseBeaconStateMachine']:
+        """
+        Returns the ``StateMachine`` instance for the given block slot number.
+        """
+        return cls.get_sm_class_for_block_slot(block.slot)
+
+    @abstractmethod
+    def get_sm(self, block: BaseBeaconBlock) -> 'BaseBeaconStateMachine':
```
cburgdorf: I know that this is just mimicking the `get_vm` API but I'd like us to think about the name for a bit. If you ask a software developer what a VM is I'm certain the majority thinks of a "Virtual Machine" instantly...I don't think the same can be said about "SM" though. I for one, have different pictures in my head when I hear "SM" so maybe it would be worth to rename this to the longer form `get_state_machine()

eth2/beacon/chains/base.py
```diff
@@ -0,0 +1,380 @@
...
+        block_root = self.chaindb.get_canonical_head_root()
+
+        block_class = self.get_block_class(block_root)
+        return self.chaindb.get_block_by_root(block_root, block_class)
+
+    def get_score(self, block_root: Hash32) -> int:
+        """
+        Return the score of the block with the given hash.
+
+        Raises ``BlockNotFound`` if there is no matching black hash.
```
cburgdorf: Nitpick: Raise~s~

eth2/beacon/db/chain.py
```diff
@@ -54,6 +57,9 @@
 class BaseBeaconChainDB(ABC):
     db = None  # type: BaseAtomicDB
 
+    def __init__(self, db: BaseAtomicDB) -> None:
+        pass
```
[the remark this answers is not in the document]
cburgdorf: This certainly looks like a bug/leftover to. Guess it was meant to be `self.db = db` instead of `pass`

eth2/beacon/exceptions.py
```diff
@@ -0,0 +1,10 @@
+from eth.exceptions import (
+    PyEVMError,
+)
+
+
+class SMNotFound(PyEVMError):
```
cburgdorf: Same here. `StateMachineNotFound` looks nicer to me tbh. Not sure how decided we are on the `SM` naming already?

eth2/beacon/exceptions.py
```diff
@@ -0,0 +1,10 @@
...
+    """
+    Raise when no StateMachine is available for the provided block slot number.
```
cburgdorf: Raise -> Raised

eth2/beacon/state_machines/forks/serenity/__init__.py
```diff
@@ -21,3 +24,6 @@ class SerenityStateMachine(BeaconStateMachine):
     state_class = SerenityBeaconState  # type: Type[BeaconState]
     state_transition_class = SerenityStateTransition  # type: Type[BaseStateTransition]
     config = SERENITY_CONFIG
+
+    # methods
+    create_block_from_parent = staticmethod(create_serenity_block_from_parent)   # type: ignore
```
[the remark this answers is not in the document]
cburgdorf: I played with that for a minute and even after annotating it as `StaticMethod[Callable[[BaseBeaconBlock, KWArg()], BaseBeaconBlock]]`, the revealed type was still `Any`. So, I propose we do what you said and implement it as:

```python
    def create_serenity_block_from_parent(parent_block: BaseBeaconBlock,
                                      **block_params: Any) -> BaseBeaconBlock:
        return create_serenity_block_from_parent(parent_block, **block_params)
```

In fact, we could apply that to existing code to get rid of some `type: ignore` and increase type safety.

eth2/beacon/types/blocks.py
```diff
@@ -194,3 +204,37 @@ def from_root(cls, root: Hash32, chaindb: 'BaseBeaconChainDB') -> 'BeaconBlock':
...
+            state_root=parent_block.state_root,
+            randao_reveal=ZERO_HASH32,
+            candidate_pow_receipt_root=parent_block.candidate_pow_receipt_root,
+            signature=EMPTY_SIGNATURE,
+            body=cls.block_body_class.create_empty_block_body(),
+        )
+
+    @classmethod
+    def cast_block(cls,
+                   block: 'BaseBeaconBlock') -> 'BaseBeaconBlock':
```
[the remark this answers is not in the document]
cburgdorf: >Should the return type here be `"BeaconBlock"` instead of `BaseBeaconBlock`?

I think so. 

Also, should this whole API be renamed from `cast_block` to `convert_block`? Because from what it does, this is more like *converting* than *casting*. A cast in my book is: "Please dear compiler|system, treat this as `<Foo>` I know what I'm doing."

But this is an actual *conversion*, where we do in fact call the constructor and pass all parameters, so what is happening here is that we may in fact be converting from type `A` to type `B` (instead of just *telling* to *treat* it differently)

eth2/beacon/chains/base.py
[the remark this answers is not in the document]
cburgdorf: Oh, sorry. When I read this I thought this was an *old* comment from @pipermerriam that got lost. Now I realized that I just commented at the same time as Piper, rendering my comment stupid haha

eth2/beacon/chains/base.py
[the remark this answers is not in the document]
cburgdorf: Btw, loosely related: *structual typing*  a la TypeScript is coming to `mypy` (or already is behind the `typing_extensions` import) and I think we could benefit from that a lot now that we maintain the beacon node and the eth1 node in one code base.

https://mypy.readthedocs.io/en/latest/protocols.html#defining-subprotocols-and-subclassing-protocols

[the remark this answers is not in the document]
hwwhww: will change to fully abstract scheme. 👍

eth2/beacon/db/chain.py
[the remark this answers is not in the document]
hwwhww: Took a look the `eth` side, I think I should have added `@abstractmethod`? `self.db = db` will be set in `BeaconChainDB`.

eth2/beacon/state_machines/forks/serenity/__init__.py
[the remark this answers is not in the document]
hwwhww: @cburgdorf

Since the API name is `create_block_from_parent`, do you mean:
```python
class SerenityStateMachine(BeaconStateMachine):
    ...
    @staticmethod
    def create_block_from_parent(parent_block: BaseBeaconBlock,
                                 **block_params: Any) -> BaseBeaconBlock:
        return create_serenity_block_from_parent(parent_block, **block_params)
```

eth2/beacon/state_machines/forks/serenity/blocks.py
```diff
@@ -1,12 +1,19 @@
...
 
-class SerenityBeaconBlockBody(BeaconBlockBody):
+class SerenityBeaconBlock(BeaconBlock):
     pass
 
 
-class SerenityBeaconBlock(BeaconBlock):
-    pass
+def create_serenity_block_from_parent(parent_block: BaseBeaconBlock,
+                                      **block_params: Any) -> BaseBeaconBlock:
```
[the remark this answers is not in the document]
hwwhww: 1. I just found we actually just pass optional `slot: SlotNumber` to `from_parent` for `BeaconBlock`. Maybe we can just set it to `slot: SlotNumber=None`?
2. The current `**block_params` is convenient for passing optional params. Please correct me if I misunderstand it, I tried:
```python
class FromBlockParams(TypedDict):
    slot: SlotNumber
```
If we use `TypedDict`, it seems we have to explicitly pass `block_params=FromBlockParams(slot=None)` when we want to use default value?

hwwhww:
@pipermerriam @cburgdorf  Thank a lot for the review!

> `BaseBeaconChain` vs `BeaconChain`

Using the fully abstract scheme now.


Resolved the most issues except for:
1. Question on using `TypedDict` in this case: https://github.com/ethereum/trinity/pull/156#discussion_r249254297
2. Not 100% sure if this change is correct: https://github.com/ethereum/trinity/blob/5efb8702a748e622191e4f891b10f0ec62f11d5c/eth2/beacon/state_machines/forks/serenity/__init__.py#L33-L36
2. Creating some `eth.tools.builder` utils and the signer utils:
Agreed and that’s the plan! 👍 I’m inclined to create it in another PRs, and maybe we can sort out the functions according to: https://github.com/ethereum/eth2.0-specs/pull/444

eth2/beacon/db/chain.py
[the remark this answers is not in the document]
cburgdorf: @hwwhww sounds right

eth2/beacon/state_machines/forks/serenity/__init__.py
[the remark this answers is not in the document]
cburgdorf: @hwwhww yep, that's what I meant. :+1: Sloppy me, copy and pasting nonsense haha

eth2/beacon/chains/base.py
```diff
@@ -0,0 +1,406 @@
...
+                    sm_class.block_class
+                )
+            )
+
+        chaindb = cls.get_chaindb_class()(db=base_db)
+        chaindb.persist_state(genesis_state)
+        return cls.from_genesis_block(base_db, genesis_block)
+
+    @classmethod
+    def from_genesis_block(cls,
```
jannikluhn: Is it legal to create a beacon chain with only a genesis block but without a genesis state? If not, maybe rename to `_from_genesis_block`?

eth2/beacon/chains/base.py
```diff
@@ -0,0 +1,406 @@
...
+
+    #
+    # StateMachine API
+    #
+    @classmethod
+    def get_state_machine_class(cls, block: BaseBeaconBlock) -> Type['BaseBeaconStateMachine']:
+        """
+        Returns the ``StateMachine`` instance for the given block slot number.
+        """
+        return cls.get_state_machine_class_for_block_slot(block.slot)
```
jannikluhn: Should we have a `self.ensure_block` in here, similar to `get_state_machine`?

eth2/beacon/chains/base.py
```diff
@@ -0,0 +1,406 @@
...
+    def create_block_from_parent(self,
+                                 parent_block: BaseBeaconBlock,
+                                 block_params: FromBlockParams) -> BaseBeaconBlock:
+        """
+        Passthrough helper to the ``StateMachine`` class of the block descending from the
+        given block.
+        """
+
+        return self.get_state_machine_class_for_block_slot(
+            slot=parent_block.slot + 1,
```
jannikluhn: maybe
```suggestion
            slot=parent_block.slot + missed_slots + 1,
```
where `missed_slots` is an additional parameter with default `0`?

eth2/beacon/chains/base.py
```diff
@@ -0,0 +1,406 @@
...
+        """
+        Return the score of the block with the given hash.
+
+        Raise ``BlockNotFound`` if there is no matching black hash.
+        """
+        return self.chaindb.get_score(block_root)
+
+    def ensure_block(self, block: BaseBeaconBlock=None) -> BaseBeaconBlock:
+        """
+        Return ``block`` if it is not ``None``, otherwise return the block
```
jannikluhn:
```suggestion
        Return ``block`` if it is not ``None``, otherwise return a new child of
```

eth2/beacon/chains/base.py
```diff
@@ -0,0 +1,406 @@
...
+
+        Raise ``BlockNotFound`` if there's no block with the given number in the
+        canonical chain.
+        """
+        return self.chaindb.get_canonical_block_root(slot)
+
+    def import_block(
+            self,
+            block: BaseBeaconBlock,
+            perform_validation: bool=True
```
jannikluhn:
```suggestion
            perform_validation: bool=True,
```

eth2/beacon/chains/base.py
```diff
@@ -0,0 +1,406 @@
...
+                "Attempt to import block #{}.  Cannot import block {} before importing "
+                "its parent block at {}".format(
+                    block.slot,
+                    block.hash,
+                    block.parent_root,
+                )
+            )
+        base_block_for_import = self.create_block_from_parent(
+            parent_block,
+            FromBlockParams(slot=None)
```
jannikluhn:
```suggestion
            FromBlockParams(slot=None),
```

eth2/beacon/chains/base.py
```diff
@@ -0,0 +1,406 @@
...
+        )
+        state, imported_block = self.get_state_machine(
+            base_block_for_import
```
jannikluhn:
```suggestion
            base_block_for_import,
```

eth2/beacon/db/chain.py
```diff
@@ -326,6 +374,13 @@ def _persist_block_chain(
         except StopIteration:
             return tuple(), tuple()
 
+        no_canonical_head = False
```
jannikluhn: Maybe it's subjective, but I'd put this either to the end of the `try` block or in an `else` block after `try...except`.

cburgdorf: >or in an `else` block after `try...except`.

:+1: I think that would be the most idiomatic handling in respect to this code base

eth2/beacon/typing.py
```diff
@@ -11,9 +16,13 @@
...
 Ether = NewType('Ether', int)  # uint64
 Gwei = NewType('Gwei', int)  # uint64
 
 Timestamp = NewType('Timestamp', int)
 Second = NewType('Second', int)
+
+
+# TypedDict
+class FromBlockParams(TypedDict):
+    slot: SlotNumber
```
cburgdorf: @hwwhww 

If we'd use `NamedTuple` instead of `TypedDict`, we'd be free to use default values for some properties, which makes it more ergonomic to use.

```python
from typing import Optional, NamedTuple


class FooBar(NamedTuple):
    mandatory_property: int
    some_property: Optional[int] = None
    some_other_property: int = 100

x = FooBar(mandatory_property=500)
```

eth2/beacon/state_machines/forks/serenity/blocks.py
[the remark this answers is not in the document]
cburgdorf: If we derive from `NamedTuple` instead of `TypedDict`, we can use default values to make this more ergonomic. 

https://github.com/ethereum/trinity/pull/156/files#r249385491

eth2/beacon/typing.py
```diff
@@ -11,9 +16,13 @@
...
 
-
 Ether = NewType('Ether', int)  # uint64
 Gwei = NewType('Gwei', int)  # uint64
 
 Timestamp = NewType('Timestamp', int)
 Second = NewType('Second', int)
+
+
+# TypedDict
```
cburgdorf: Btw, I think we shouldn't do these type of comments. They quickly go out of sync (e.g. if we decide to derive from `NamedTuple`  instead) and they don't seem to help comprehension of the code base.

eth2/beacon/typing.py
carver: Yeah, Python3.6 is the default in the Ubuntu LTS that's been out for nearly a year. We've already been planning the drop of python3.5 for a while in py-evm. So :+1: for 3.6 syntax and an issue to track the change in py-evm (if it's not already there).

PR #163, opened 2019-01-14
Bcc syncing

jannikluhn:
[moved from ethereum/py-evm/pull/1650]

This implements syncing of the beacon chain according to the following basic strategy:

- Select the peer with the highest announced slot number.
- Request blocks starting with the last finalized block.
- Stop when the peer does not provide any more blocks (i.e. we've reached their head or something went wrong).

Non-beacon chain code I touched: I've added an empty `TypedDict` as a fourth option for p2p payload types in order to use my `StatusMessage`, `BeaconBlocksMessage`, and `GetBeaconBlocksMessage` properly. For some reason, this meant that I had to add explicit type hints for "normal" dictionaries in a few instances (e.g. [here](https://github.com/ethereum/py-evm/compare/ethereum:48be6ca...jannikluhn:724bd22#diff-2fe3625b5324b16f88b7a92761437638R107)). Otherwise mypy expects dictionaries without entries (even though `Dict[str, Any]` is still allowed). Referencing #845 

#### Cute Animal Picture


trinity/sync/beacon/chain.py
```diff
@@ -0,0 +1,144 @@
...
+    def __init__(self,
+                 chain_db: BaseBeaconChainDB,
+                 peer_pool: BCCPeerPool,
+                 token: CancelToken = None) -> None:
+        super().__init__(token)
+
+        self.chain_db = chain_db
+        self.peer_pool = peer_pool
+
+        self.sync_peer: BCCPeer = None
```
pipermerriam: Orthogonal to this PR **but** worth pointing out.  With a reasonably high confidence interval, we'll be migrating all of the existing syncing code's peer and peer pool interactions to be event bus based.  See #79 

I don't have a recommendation for whether you should stick with the current pattern or try to model this under the new paradigm since the new way will likely require some discovery/exploration/figuring-it-out work but conversely, this old way is likely to present some problems and overhead with-respect-to complexity that manifests when trying to ensure your networking process doesn't get overwhelmed.

Just wanted to highlight this eventuality to be sure you were aware.

trinity/sync/beacon/chain.py
```diff
@@ -0,0 +1,144 @@
...
+            self.logger.info("No suitable peers to sync with")
+
+        await self.sync()
+
+        new_head = self.chain_db.get_canonical_head(BeaconBlock)
+        self.logger.info(f"Sync with {self.sync_peer} finished, new head: {new_head}")
+
+    def select_sync_peer(self) -> BCCPeer:
+        if len(self.peer_pool) == 0:
+            return None
```
pipermerriam: Maybe exception raising instead of this which requires checking return values?

trinity/sync/beacon/chain.py
```diff
@@ -0,0 +1,144 @@
...
+        if best_peer.head_slot <= finalized_head_slot:
+            return None
+
+        return best_peer
+
+    async def sync(self) -> None:
+        self.logger.info(
+            "Syncing with %s (their head slot: %d, our finalized slot: %d)",
+            self.sync_peer,
+            self.sync_peer.head_slot,
```
pipermerriam: Just making sure this is all intentional (and if so, maybe add a comment saying as much).

When `select_sync_peer()` doesn't find a suitable peer, `self.sync_peer` will be `None`.  This will in-turn error out with an attribute error.  If you're looking for a simple short-term solution, I'd suggest something like

1. a coroutine that will try for up to `N` seconds to find a suitable peer to sync with.
2. an event or simple sleep loop at the beginning of the `sync` coroutine which waits for `self.sync_peer` to be populated.
3. aborting/cancelling the whole sync process if no suitable peer arrives within `N` seconds.

trinity/sync/beacon/chain.py
```diff
@@ -0,0 +1,144 @@
...
+        parent_root = batch[0].parent_root
+        parent_slot = batch[0].slot - 1
+
+        if parent_slot < 0:
+            raise Exception(
+                "Invariant: Syncing starts with the child of a finalized block, so never with the "
+                "genesis block"
+            )
+
+        canonical_cousin = self.chain_db.get_canonical_block_by_slot(parent_slot, BeaconBlock)
```
pipermerriam: Why the use of the term `cousin` here?  Seems like it would be `canonical_parent`?

trinity/sync/beacon/chain.py
cburgdorf: I think it is ok to stick with what we have right now. We need to come up with a `PeerPool` that natively exposes all interactions through the `EventBus` and then I *think* the actual syncing code won't look too much different except that it uses something like `PeerPoolAPI(event_bus)` to interact with the `PeerPool`.

Notice that #155 already moves the syncing into a plugin but that isn't an `BaseIsolatedPlugin` yet which means it still runs in the `networking` process and interacts with the current `PeerPool`. I believe once this PR lands, it should be relatively easy to modify #155 to bootstrap the sync for the beacon node.

Issue #119, opened 2019-01-08
FastChainBodySyncer crashes(?) and sync stalls


lithp:
* py-evm Version: c4e38746

### What is wrong?

```

DEBUG  01-07 13:57:40   FastChainBodySyncer  Duplicate headers during fast sync (<BlockHeader #214104 d837fccd>, <BlockHeader #214105 d1aaf1c3>, <BlockHeader #214106 cf8
[this line is insanely long, it includes 10929 instances of "BlockHeader"]

[40 minutes where the only lines look like this]
DEBUG  01-07 13:57:43     DiscoveryProtocol  Ignoring find_node request from unknown node <Node(0x77c7@172.104.162.211)>
DEBUG  01-07 13:57:44     DiscoveryProtocol  Ignoring find_node request from unknown node <Node(0x9043@142.93.125.215)>
DEBUG  01-07 13:57:45     DiscoveryProtocol  Ignoring find_node request from unknown node <Node(0x80a3@54.37.85.202)>
[the 40 minutes end]

    DEBUG  01-07 14:36:20   FastChainBodySyncer  <coroutine object FastChainBodySyncer._launch_prerequisite_tasks at 0x7f18b84c1a48> finished while <trinity.sync.full.chain.
FastChainBodySyncer object at 0x7f18b84c2cf8> is still running, terminating as well
ESC[1mESC[33m WARNING  01-07 14:36:20   FastChainBodySyncer  Task <coroutine object FastChainBodySyncer._launch_prerequisite_tasks at 0x7f18b82686c8> finished unexpectedly:
 Cannot prepare task <CompletionForBlockPersistPrereqs(<BlockHeader #136877 f3590012>, done=set(), remaining={<BlockPersistPrereqs.StoreBlockBodies: 1>, <BlockPersistPrereq
s.StoreReceipts: 2>})> with id b'\xf3Y\x00\x12I$\x15\'*=\rr@\xcbBe\x93\x1fj^\x10b\x07\x1e:\x85"\x89<\x04\x0fX' and dependency b'\x04I\xdf[>\xf1\xda\\iN\x82f\x94\x13\x80\xf1
C\x8a\\\xf0h\x0fn\xd9\xbc\xcd/\xd0\xf1nn\x0c' before preparing its dependencyESC[0m
   DEBUG  01-07 14:36:20   FastChainBodySyncer  Task failure traceback
Traceback (most recent call last):
[...]

PR #166, opened 2019-01-15
Resolve some service shutdown hangups

carver:
### What was wrong?

Lots of services were failing to shut down cleanly. They would permanently hang on one of their child tasks or services. One noted side effect is that when the syncer crashes, the node keeps running. See #119 

Out of scope for this PR: even after the `FullNode` shuts down, the processes continue running, so there's still an awkward ending where you have to `Ctrl-C` out.

@cburgdorf would you mind figuring out how to make all the processes shut down when the `FullNode` exits? The bootstrapping code changed a lot, and is heavily entangled with the event bus, so the "right" way to do this isn't obvious to me anymore. I'll open an issue, but I would appreciate your review here anyway (which is also good background for the issue).

### How was it fixed?

Hunt down the places that were preventing the node services from shutting down. Some general classes of things found:

- tasks that were hanging waiting for an event bus event
- tasks that were waiting for a cancellation event, but the `cancel_token` trigger was never generating a cancellation event
- perhaps the most tangential (I'm happy to pull this into another PR if anyone asks): when fast sync exits abnormally, regular sync was starting. now it just exits immediately.

Note that there are some more asyncio warnings on shutdown now, but I still think it's forward progress.

#### Cute Animal Picture

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -1003,7 +1003,7 @@ class DiscoveryService(BaseService):
         self._lookup_running = asyncio.Lock()
 
     async def handle_get_peer_candidates_requests(self) -> None:
-        async for event in self._event_bus.stream(PeerCandidatesRequest):
+        async for event in self.wait_iter(self._event_bus.stream(PeerCandidatesRequest)):
 
             self.run_task(self.maybe_lookup_random_node())
 
@@ -1016,7 +1016,7 @@ class DiscoveryService(BaseService):
             )
 
     async def handle_get_random_bootnode_requests(self) -> None:
-        async for event in self._event_bus.stream(RandomBootnodeRequest):
+        async for event in self.wait_iter(self._event_bus.stream(RandomBootnodeRequest)):
 
             nodes = tuple(to_uris(self.proto.get_random_bootnode()))
 
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -821,7 +821,7 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
             self.run_task(self.connect_to_nodes(from_uris([command.node])))
 
     async def handle_peer_count_requests(self) -> None:
-        async for req in self.event_bus.stream(PeerCountRequest):
+        async for req in self.wait_iter(self.event_bus.stream(PeerCountRequest)):
                 # We are listening for all `PeerCountRequest` events but we ensure to only send a
                 # `PeerCountResponse` to the callsite that made the request.  We do that by
                 # retrieving a `BroadcastConfig` from the request via the
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -1005,7 +1005,7 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
 
     async def connect_to_nodes(self, nodes: Iterator[Node]) -> None:
         for node in nodes:
-            if self.is_full:
+            if self.is_full or not self.is_operational:
                 return
 
             # TODO: Consider changing connect() to raise an exception instead of returning None,
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -305,6 +305,12 @@ class BaseService(ABC, CancellableMixin):
     def is_running(self) -> bool:
         return self._run_lock.locked()
 
+    async def cancellation(self) -> None:
+        """
+        Pause until this service is cancelled
+        """
+        await self.wait(self.events.cancelled.wait())
+
     async def threadsafe_cancel(self) -> None:
         """
         Cancel service in another thread. Block until service is cleaned up.
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -165,7 +165,7 @@ class BaseService(ABC, CancellableMixin):
                         awaitable,
                         self,
                     )
-                    self.cancel_token.trigger()
+                    self.cancel_nowait()
         self.run_task(_run_daemon_task_wrapper())
 
     def run_child_service(self, child_service: 'BaseService') -> None:
@@ -217,7 +217,7 @@ class BaseService(ABC, CancellableMixin):
                         service,
                         self,
                     )
-                    self.cancel_token.trigger()
+                    self.cancel_nowait()
 
         self.run_task(_run_daemon_wrapper())
 
@@ -327,7 +327,7 @@ class BaseService(ABC, CancellableMixin):
 
         Should return or raise OperationCancelled when the CancelToken is triggered.
         """
-        raise NotImplementedError()
+        pass
 
     async def _cleanup(self) -> None:
         """Clean up any resources held by this service.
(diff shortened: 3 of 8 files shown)


p2p/discovery.py
```diff
@@ -974,7 +974,7 @@ def __init__(self,
         self.run_daemon_task(self.handle_get_random_bootnode_requests())
 
     async def handle_get_peer_candidates_requests(self) -> None:
-        async for event in self._event_bus.stream(PeerCandidatesRequest):
+        async for event in self.wait_iter(self._event_bus.stream(PeerCandidatesRequest)):
```
lithp: I'm having some trouble understand the flow here. Is all this right?

Previously this daemon task would blocked, waiting on an event from the queue to come in. It now waits on that event through the cancel_token, so cancellation causes an exception to be raised here and the task is exited.

`BaseService.cancel()` calls cancel on the futures, which I think raises a `CancelledError` on this line, so the task should be cleaned up even though it's not using a cancellation token?

I guess there's a bug if this service was stopped with `BaseService.cleanup()`. `cleanup()` would wait forever trying to gather() the task, but that would mean that trinity would always wait forever, and it often exited cleanly.

cburgdorf: >so the task should be cleaned up even though it's not using a cancellation token?

Exactly. AFAIK, there shouldn't be a need to wrap `event_bus.stream(...)` with `wait_iter` as the `stream` API should handle a proper cleanup when the future is cancelled.

Need to take a closer look.

cburgdorf:
>@cburgdorf would you mind figuring out how to make all the processes shut down when the FullNode exits?

#168 demonstrates how to do that. Basically, we just need to send a `ShutdownRequest` which the `main` process listens to and then shuts down all processes. I'm not entirely sure if there would be a better place where to put that though. I put it at the very last step of the `Node.run(...)` function which means, it technically is also called when it is not needed but it doesn't seem to be harmful either.

Ultimately, as the architecture becomes more dominated by plugins (e.g. #155 moves the bootstrap of the `sync` into a plugin), essential plugins probably need to decide if they want to cause a full shutdown or not.

p2p/discovery.py
carver: > Previously this daemon task would blocked, waiting on an event from the queue to come in. It now waits on that event through the cancel_token, so cancellation causes an exception to be raised here and the task is exited.

yes

> `BaseService.cancel()` calls cancel on the futures, which I think raises a `CancelledError` on this line, so the task should be cleaned up even though it's not using a cancellation token?

Yes, but `.cancel()` is never called. The future cancellation is happening when `wait_iter` exits on the token, and cancels the other pending future (the `.stream()` call here). As you pointed out, when a service dies by `cancel_token` it must wait co-operatively for the tasks to end: https://github.com/ethereum/trinity/blob/7b3a26e6ccfc9225042ce4bead1ef49892b6446a/p2p/service.py#L251-L254

It's only when `service.cancel()` is *explicitly* called that the tasks are forcibly cancelled after a timeout: https://github.com/ethereum/trinity/blob/7b3a26e6ccfc9225042ce4bead1ef49892b6446a/p2p/service.py#L277-L285

Because these services were dying by `cancel_token`, their `cleanup()` was hanging indefinitely, and failing to cancel other services up the chain.

I'm not claiming any of this is *good*, just how I understand it to work.

> I guess there's a bug if this service was stopped with `BaseService.cleanup()`. `cleanup()` would wait forever trying to gather() the task, but that would mean that trinity would always wait forever, and it often exited cleanly.

The reason it was sometimes exiting cleanly (I assume you mean on `Ctrl-C`) is that `SIGTERM` also sends a termination signal to the event bus, so the stream eventually closes and the tasks can complete.

When the sync fails, the event bus doesn't know to exit yet (and probably won't until the full `Node` service exits).

carver: > Exactly. AFAIK, there shouldn't be a need to wrap `event_bus.stream(...)` with `wait_iter` as the `stream` API should handle a proper cleanup when the future is cancelled.

Ah, this goes against the expectations of a service, which is to have a clean shutdown, instead of a forced cancellation of pending tasks. The two main options I see for the task to shut itself down are to:
- use `wait_iter()`
- pass the `cancel_token` to `event_bus.stream()`

carver:
> I put it at the very last step of the `Node.run(...)` function which means, it technically is also called when it is not needed but it doesn't seem to be harmful either.

Interesting, maybe that actually belongs in the `Node._cleanup()` since any crash in the middle of `_run()` will skip that shutdown step. I'll try it out.

carver:
> Interesting, maybe that actually belongs in the `Node._cleanup()` since any crash in the middle of `_run()` will skip that shutdown step. I'll try it out.

Yup, just tried this and it worked nicely. I'll call the change out of scope for this PR and count on #168 for the change.

p2p/discovery.py
lithp: This all makes sense, thanks for the thorough explanation! :+1: 

Currently the contract is that tasks should respect cancellation_token, which is like a SIGINT. If they don't then the task will be cancelled, which is like a SIGTERM (there's no SIGKILL because the task could `shield()` or catch `CancelledError`)

This seems easy to forget and leads to slightly more verbose code. Are there tasks which need the warning? It would be simpler to have cancellation be the normal flow: tasks are given `cancel()` and expected to catch it if they really need the extra time to clean up.

carver: > Are there tasks which need the warning?

I don't know. It wouldn't surprise me that some things would get wacky if we could interrupt any task at any time. So I'm not convinced that's safe to do, yet.

I'm not opposed to adding some kind of timeout error like we have on `cancel()` to eventually force the task shutdown (instead of hanging forever).

cburgdorf: Wow, thorough thread! Thanks for the in depth explanation @carver!

>pass the cancel_token to `event_bus.stream()`

We've been there for a brief moment in time but [discussing adding `CancelToken` support](https://github.com/ethereum/lahja/pull/11) to all lahja APIs with @pipermerriam we eventually decided against it. It doesn't seem to be a good idea to make the `EventBus` depend on the `CancelToken` library (and the whole pattern).

>It would be simpler to have cancellation be the normal flow: tasks are given cancel() and expected to catch it if they really need the extra time to clean up.

:+1: I agree to that.

carver:
So... where did we land on this? Is anyone :+1: on the changes? I haven't come up with any alternatives to `wait_iter()` that will cause a clean shutdown. (at least without forcibly cancelling all running tasks immediately, which I'm not convinced is safe)

cburgdorf:
I'm :+1: on these changes and in the process of making the change to #166 (moving it into `_cleanup`) I also experienced first hand the hangup that you explained. So, I believe `wait_iter()` is our best short term solution so far. Long term, I'd love to see us exploring alternatives to our `CancelToken` approach though.

PR #168, opened 2019-01-15
Shutdown all plugins when node finishs

cburgdorf:
### What was wrong?

Triggered by #166.

Currently, if the `Node` (aka the current `networking` process)  finishes, all external processes continue running unaffected. This PR changes this so that when the `Node` finishes (for whatever reason), all other running processes are instructed to shut down.

### How was it fixed?

Send a `ShutdownRequest` at the last thing in `Node.run(...)`.
As we move more and more responsibilities into plugins (e.g. #155 moves the sync into a plugin) this responsibility would shift so that essential plugins would trigger this. 

#### Cute Animal Picture

### DIFF
--- a/trinity/nodes/base.py
+++ b/trinity/nodes/base.py
@@ -27,6 +27,9 @@ from trinity.config import (
     ChainConfig,
     TrinityConfig,
 )
+from trinity.events import (
+    request_shutdown,
+)
 from trinity.extensibility.events import (
     ResourceAvailableEvent
 )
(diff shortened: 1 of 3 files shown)


pipermerriam:
I restarted the integration test failure CI run to see if it's a fluke.

trinity/nodes/base.py
```diff
@@ -141,3 +144,4 @@ def notify_resource_available(self) -> None:
         self.notify_resource_available()
         self.run_daemon_task(self.handle_network_id_requests())
         await self.get_p2p_server().run()
+        request_shutdown(self.event_bus, "Node finished unexpectedly")
```
carver: This works quite nicely when moved into `def _cleanup`. I tested on my local branch for #166 :+1:

cburgdorf: Yep, moved it. It still depends on #166 though because with that, we hang and never get to `_cleanup` (related to your what you explained in that other thread).

PR #173, opened 2019-01-15
Migrate to RocksDB

cburgdorf:
### What was wrong?

Trinity is moving towards an architecture with many different processes. Unfortunately our current database `LevelDB` does only allow one process to read/write to it at a time. So far, we've been dealing with that by having one process entirely dedicated for db access.

So, if process `A` wants to *read* some key from the database, the information flow is as follows:

```
Process A

+        ^
|        |
|        |
|        |
|        |
v        +


DB Manager Process

+        ^
|        |
|        |
|        |
|        |
v        +

 Actual Database Process

```

Obviously, moving data between process boundaries is always a bottleneck and under this model we are crossing process boundaries 4 times for a simple read.

Enter `RocksDB` which is a `LevelDB` fork by facebook. With `RocksDB` we can have one **read/write** and lots of **readonly** processes connected to the database.

There are a lot of parts of Trinity that do not need to write to the database and can hence benefit from a local readonly connection the database which means we cut process boundary crossing down to 2 (request/response to the actual database process)

### How was it fixed?

This is mostly just digging out the work @pipermerriam had started a while back plus a bit of care to ensure the dev experience works well.

Some points:

- in contrast to the previous attempt, instead of supporting different db engines, this goes all-in on `RocksDB` as goal is to get rid of some of the proxy overhead and lead toward a different architecture so trying to continue `LevelDB` support feels like an unnecessary burden holding us back

- This demos how to benefit from a readonly connection for the JSON-RPC API

If the existing db is a LevelDB db, Trinity will great like this:

image

### TODOs

- [x] ensure the installation story works well and is documented for both Linux and OSX
- [x] Check if the database is correctly initialized (aka handle LevelDB breaking change) as soon as #123 landed (because then, this needs to happen in the Eth1 branch of checks)
- [x] refactor `readonly_database`  property after #123 landed

#### Cute Animal Picture


cburgdorf:
Would anyone with a Mac like to try out this PR especially in respect to the given docs and if the listed steps work well to get Trinity to work with RocksDB? @lithp are you working on a Mac by any chance?

lithp:
Sorry, I'm on Ubuntu, but I love the idea behind this PR! Looks like Parity also uses RocksDB. I guess that the thorough test fixtures give Ethereum the confidence to use different databases without fear of a Bitcoin-style leveldb-caused fork?

cburgdorf:
>I'm on Ubuntu

Same. Guess all of us are on Linux which is kinda nice ;)

>Parity also uses RocksDB

Ah, I didn't know about Parity but Lighthouse and Nimbus also use RocksDB. If I recall correctly, the installation story is more user friendly in Rust (which Lighthouse and Parity use) because I *think* they have a native crate (like a `pip` package) that takes care of the compilation step whereas for us you either have to compile it yourself or install a binary if one exists (ubuntu has it as a native package).

But at this point, I think we should value performance slightly more than developer UX.

What does everyone think about hosting the `RocksDB` abstraction inside Trinity as this PR currently has it? @lithp @pipermerriam @carver 

The `LevelDB` abstraction lives in `Py-EVM` but I personally don't see it as a must to have the `RocksDB` abstraction in `Py-EVM`. After all, the main motivation to use `RocksDB` in Trinity is because it allows concurrent readonly connections from multiple processes which clearly is a requirement out of the scope of the Py-EVM.

The nice thing about having the `RocksDB` abstraction in `Py-EVM` would be to have benchmarks, so we could compare how much of a performance gain `RocksDB` brings for block processing etc.

I would propose that I do a second PR to also add `RocksDB` to `Py-EVM` and then, after the next `Py-EVM` release, we can drop it from Trinity again and import it from `Py-EVM`.

>without fear of a Bitcoin-style leveldb-caused fork?

I didn't know that happened for Bitcoin. But yeah, this is one more reason to have `RocksDB` in Py-EVM as well, because we can run it through more existing tests there.

So, how does everyone feel about moving on with this?

pipermerriam:
> What does everyone think about hosting the `RocksDB` abstraction inside Trinity as this PR currently has it?

`LevelDB` abstraction could move into `Trinity` too.  The **only** justification I can see for having these available in `Py-EVM` is if someone wants to manually instantiate a `Chain` instance connected to an existing trinity database from a separate environment than whatever trinity runs in (which is reasonable to assume the python environments would be separate since we're likely to handle virtual environment setup if installed via `brew` or a PPA.)

cburgdorf:
>The **only** justification I can see for having these available in Py-EVM is if someone wants to manually instantiate a Chain instance connected to an existing trinity database

From a developer experience, whatever it is that you want to build on top of Py-EVM, if it needs persistence, then you'll need some sort of connector. Not having any connectors in Py-EVM makes these things less ergonomic. You'd either need to implement `BaseAtomicDB` yourself or import it from *somewhere* where *somewhere* should not be Trinity I think :)

E.g. the benchmark suite would be less useful if it only runs against `MemoryDB`. In fact, it would be good if it can run against `RocksDB` as well which would be one reason for moving `RocksDB` into Py-EVM.

.gitignore
```diff
@@ -74,3 +74,6 @@ monkeytype.sqlite3
 .idea/**
 
 !tests/integration/fixtures/*.ldb.zip
+
+# rocksdb
+rocksdb/
```
carver: *nitpick* newline at end

.circleci/install_rocksdb.sh
```diff
@@ -0,0 +1,20 @@
...
+
+echo "Building on $NPROC cores. Pass number of cores (e.g. install_rocksdb.sh 4) to run concurrently"
+
+# More about installation of RocksDB https://github.com/facebook/rocksdb/blob/master/INSTALL.md
+
+[ -f rocksdb/rocksdb-$ROCKSDBVER/Makefile ] || { rm -rf rocksdb ; mkdir -p rocksdb; cd rocksdb; wget https://github.com/facebook/rocksdb/archive/v$ROCKSDBVER.tar.gz && tar xvf v$ROCKSDBVER.tar.gz; cd ..; }
+cd rocksdb/rocksdb-$ROCKSDBVER
+[ -f util/build_version.cc ] || { make util/build_version.cc ; } # use cached version if possible
+export NO_UPDATE_BUILD_VERSION=1
+make shared_lib -j$NPROC && make install-shared INSTALL_PATH=/usr
```
carver: As you already implied, it's a big bummer to have all this custom install logic that lives outside the normal install process. I don't have a solution, just grumpy.

docs/contributing.rst
```diff
@@ -1,22 +1,32 @@
 Contributing
 ------------
 
-Thank you for your interest in contributing! We welcome all contributions no matter their size. Please read along to learn how to get started. If you get stuck, feel free to reach for help in our `Gitter channel <https://gitter.im/ethereum/py-evm>`_.
+Thank you for your interest in contributing! We welcome all contributions no matter their size.
+Please read along to learn how to get started. If you get stuck, feel free to reach for help in
+our `Gitter channel <https://gitter.im/ethereum/py-evm>`_.
```
carver:
```suggestion
our `Gitter channel <https://gitter.im/ethereum/trinity>`_.
```

docs/guides/quickstart.rst
```diff
@@ -4,61 +4,100 @@ Quickstart
...
 ~~~~~~~~~~~~
 
-This is the quickstart guide for Trinity. If you only care about running a Trinity node, this
-guide will help you to get things set up. If you plan to develop on top of Py-EVM or contribute
-to the project you may rather want to checkout the :doc:`Contributing Guide </contributing>` which
-explains how to set everything up for development.
+This is the quickstart guide for Trinity. If all we care about is *running* a Trinity node,
+this guide covers everything to set things up.
+
+For developers who want to *work* on Trinity and *contribute* to the project, the
```
carver: Could drop the italics on contribute. Starts to feel like a little too much emphasis. If you want to further clarify, you could maybe change `*work* on` -> `*modify*`. (To resolve the tiny bit of ambiguity, like maybe you want to work on _top_ of trinity, but not modify it. You don't need to clone the repo for that.)

trinity/db/rocksdb.py
```diff
@@ -0,0 +1,140 @@
...
+    def __delitem__(self, key: bytes) -> None:
+        exists, _ = self.db.key_may_exist(key)
+        if not exists:
+            raise KeyError(key)
+        self.db.delete(key)
+
+    @contextmanager
+    def atomic_batch(self) -> Iterator['RocksDBWriteBatch']:
+        import rocksdb  # noqa: F811
+        batch = rocksdb.WriteBatch()
```
carver: Feels awkward to reimport this at the beginning of every write batch. Maybe it's not a lot better, but we could assign `self._WriteBatch = rocksdb.WriteBatch` during init or something, to avoid it.

trinity/plugins/builtin/json_rpc/plugin.py
```diff
@@ -70,8 +70,7 @@ def setup_eth1_modules(self, trinity_config: TrinityConfig) -> Tuple[Eth1ChainRP
             event_bus_light_peer_chain = EventBusLightPeerChain(self.context.event_bus)
             chain = chain_config.light_chain_class(header_db, peer_chain=event_bus_light_peer_chain)
         elif eth1_app_config.database_mode is Eth1DbMode.FULL:
-            db = db_manager.get_db()  # type: ignore
-            chain = chain_config.full_chain_class(db)
+            chain = chain_config.full_chain_class(eth1_app_config.readonly_database)
```
carver: The way the rpc tests work, I don't think this line actually gets tested. Have you tried locally attaching and running web3 requests on a rocksdb node?

Maybe we ought to add an attach test in `test_trinity_cli` that uses a web3 call to hit the database.

All the other tests I know of build a custom chain (which means a custom `AtomicDB` in-memory database).

carver:
Oh, and this should definitely use the new release notes scheme, so folks like @veox get notified before losing access to their db. X)

.circleci/install_rocksdb.sh
cburgdorf: Kinda depends on the system though. For newer Ubuntu (including current 18.10), installing rocksdb can be as simple as `apt-get install librocksdb-dev`. I installed it from source for CI to have total control over the installed version and be more independent from the `apt` sources. Not sure if that's too paranoid and if we should install it via `apt` in CI?

trinity/db/rocksdb.py
cburgdorf: Ah actually, if we are going all in with `RocksDB` we could also just import it on top of the file.

trinity/plugins/builtin/json_rpc/plugin.py
cburgdorf: I'm pretty sure I tested it locally but I like the idea of expanding our interactive `attach` tests to cover that :+1:

.circleci/install_rocksdb.sh
carver: Ah, if it's available in apt, then I'd say use it. We are already using that approach on other libraries, and were using it in leveldb. I think it would be quite useful to find out if trinity stops working with a traditional `apt` install.

trinity/db/rocksdb.py
carver: I'm good with that in trinity :+1: 

:-1: for doing that in py-evm though.

cburgdorf: I'm on the same page: It's a hard requirement for Trinity (if we want to leverage concurrent readonly connections) but Py-EVM does not need a hard dependency against RocksDB.

.circleci/install_rocksdb.sh
cburgdorf: >I think it would be quite useful to find out if trinity stops working with a traditional apt install.

I've tested that and it works fine. The main reason I went for the source install was really to be able to pin it down to exactly the version we want.

I'll give the `apt` install a try for CI but would recommend we leave the install script somewhere in the repo. Could come in handy for when we want to install specific versions locally.

cburgdorf: Ah too bad. None of the available Circle CI docker images supports the version that we need. So, installing from source is currently our best option for CI I guess.

cburgdorf:
For some reason the `py3*-wheel-cli` job is failing more often than not. Not sure why yet. The same test runs as part of the integration job without any issues.

.circleci/install_rocksdb.sh
carver: > I've tested that and it works fine.

Right, I didn't mean that it would be nice to know it works now. I meant it would be nice to know if it stops working in the future.



> None of the available Circle CI docker images supports the version that we need.

But yeah, that ends the discussion.

veox:
The [`py37-wheel-cli` CI job](https://circleci.com/gh/ethereum/trinity/7945?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) has:

```
============================= test session starts ==============================
platform linux -- Python 3.7.2, pytest-4.1.1, py-1.7.0, pluggy-0.8.1
cachedir: .tox/py37-wheel-cli/.pytest_cache
rootdir: /home/circleci/repo, inifile: pytest.ini
plugins: asyncio-0.10.0
collected 17 items / 2 deselected                                              

tests/integration/test_trinity_cli.py ........F.....Too long with no output (exceeded 10m0s)
```

Looks to have hanged.

In a different, unrelated PR, I've seen an `OSError: [Errno 12] Cannot allocate memory`. Luckily, memory was already allocated to capture error messages...

Running tests locally on `master` without considering there's already a process that claimed port 30303 quite expectedly also results in a failure.

-----

I'll try this PR locally.

veox:
Guess: in the failing test, `w3.net.version` in the console of `trinity attach` is sent before the console is done attaching; the test then waits forever to read an answer for a question no one heard.

veox:
Note for Arch Linux users:

[`rocksdb` is in AUR](https://aur.archlinux.org/packages/rocksdb/) and therefore has to be built locally.

There is [an issue running with `python-rocksdb`](https://bugs.archlinux.org/index.php?do=details&action=details.addvote&task_id=54711), described also in https://github.com/jemalloc/jemalloc/issues/955 and linking to https://github.com/jemalloc/jemalloc/issues/937 as culprit; there are several work-arounds; I'm currently using the one where `libjemalloc.so` gets pre-loaded (as this doesn't require me to rebuild anything):

`LD_PRELOAD=/usr/lib/libjemalloc.so.2 pytest tests/integration/test_trinity_cli.py`

cburgdorf:
@veox Thanks for helping with the debugging! Here's what I figured out:

1. The `py36-wheel-cli` job regulary runs out of memory.

2. The same tests run in `py36-integration` without any issues though

3. I tried to reduce the number of CLI tests that `py36-wheel-cli` runs (because we run the tests as integration tests anyway, so for the wheel test it should be enough to run *some*) but even though I don't run out of memory then, it still fails for other reasons (as you although figured out)

4. I wasn't able to reproduce that locally yet

veox:
@cburgdorf PR #231 evaluates the other option; I'll try it on top of this PR once it can stand by itself on top of `master`.

carver:
> 3\. I tried to reduce the number of CLI tests that `py36-wheel-cli` runs (because we run the tests as integration tests anyway, so for the wheel test it should be enough to run _some_)

If we are going to drop any tests (not sure we should), then at least drop the non-wheel tests. The `wheel` tests are needed to catch packaging issues. For example, there might be some packaging issue that affects ropsten, but not mainnet. Or maybe we start including some asset that is only used by light sync, etc. The `wheel` tests are our final defense against production deployment bugs.

veox:
Got a feeling CI failure is not product of this PR. :/

PR #231 tries to contain them; it also fails.

cburgdorf:
>then at least drop the non-wheel tests.

That's the thing, the same tests run just fine as part of the `py36-integration` job.

Butyea, this is just some back and forth trying to figure out what exactly is it that makes the `py36-wheel-cli` tests fail. It's not even related to this PR.

veox:
Most likely fix for `py36-wheel-cli` job described in PR https://github.com/ethereum/trinity/pull/227#issuecomment-457908353.

cburgdorf:
Ok, I just noticed something weird. When using the `readonly_database` for the JSON-RPC server, it seems that it does not serve *live* changes. Only those that were available when the `readonly_database` was instantiated. Like a fixed snapshot. If that would be the case, that'd be a big bummer. Need to investigate that further.

tests/core/database/test_rocksdb_with_multiple_processes.py
```diff
@@ -92,3 +96,66 @@ def test_database_read_access_across_multiple_processes_with_ongoing_writes(db,
 
     for key, value in data_to_write.items():
         assert db[key] == value
+
+
+def test_native_readonly_connection_cant_access_data_written_after_connection(db, db_path):
```
cburgdorf: @pipermerriam I suppose that the `test_database_read_access_across_multiple_processes_with_ongoing_writes` that you had originally written was supposed to cover the same thing. But apparently, this test proves that *natively* a readonly connection to RocksDB **does not allow**  reading data that was written **after** the readonly connection has been established.

We can easily solve that by introducing our own `ReadonlyRocksDB` at the expense of creating a fresh readonly connection for every single `get` access. Alternatively we may want to add a cache to the connection for a reasonable amount of reads?

I wonder if we should even keep the `test_database_read_access_across_multiple_processes_with_ongoing_writes`  test?

cburgdorf:
>When using the readonly_database for the JSON-RPC server, it seems that it does not serve live changes

I added tests to back the claim that a readonly connection to does not have access to data that was written *after* the readonly connection was made.

I also added a `ReadonlyRocksDB` that overcomes this at the expense of creating a fresh connection for every `get` access. I haven't checked the overhead of that yet. We could also throw a small cache on that.

It's a bit weird that I haven't found this behavior documented in RocksDB anywhere so I still trying to find out if I could be wrong somehow.

pipermerriam:
@cburgdorf maybe it'd be nice to have some sort of read session API that allows for multiple reads before ditching the connection.  That should be acceptable for things like JSON-RPC server reads that need to do things like execute a `eth_call` which will need a **lot** of reads.


```python
with db.read_only_session() as base_db:
    ...  # establishes a new read-only connection for the duration of this context block.
```

cburgdorf:
@pipermerriam 

>maybe it'd be nice to have some sort of read session API that allows for multiple reads before ditching the connection

I like that idea. Just keep in mind though that this means they will all be served from the same snapshot that was taken when the connection is first made. So, if you have some *async* code doing some thing and coming back a second later and then using that same *session* to read some data you can not expect data to be present that was written half a second before. Which may be an ok thing in most circumstances. Just bringing it up as a thought.

pipermerriam:
@cburgdorf this makes me think about https://docs.djangoproject.com/en/2.1/topics/db/transactions/

I honestly have no idea how to model this but having some type of API that allowed us to reason about database access might be really nice.

- Know that all reads within a certain block will be *consistent*
- Transactional commits (we already have a version of this)
- Database level batch writing (been talked about but never implemented to the best of my knowledge)

cburgdorf:
They replied that they are working on it https://github.com/facebook/rocksdb/issues/908#issuecomment-459803173

pipermerriam:
> As far as I can tell, our atomic_batch API uses a native RocksDB WriteBatch underneath which should achieve that.

I'm pretty sure that we aren't actually leveraging batch writes at the rocksdb/leveldb level.  Looking at the code, our `DBProxy` uses `eth.db.diff.DBDiff` to do the writes, and `DBDiff.apply_to` does single key/value setting across the diff.  We would need to update that code/logic to correctly pass the full set of key/value pairs and to have the written as a batch.

carver:
> > As far as I can tell, our atomic_batch API uses a native RocksDB WriteBatch underneath which should achieve that.
> 
> I'm pretty sure that we aren't actually leveraging batch writes at the rocksdb/leveldb level. Looking at the code, our `DBProxy` uses `eth.db.diff.DBDiff` to do the writes, and `DBDiff.apply_to` does single key/value setting across the diff. We would need to update that code/logic to correctly pass the full set of key/value pairs and to have the written as a batch.

Although we write one-by-one into the `WriteBatch` object, they are committed all at once when exiting this context (at least that's how it works in `LevelDB`, and presumable `RocksDB` is the same):

https://github.com/ethereum/py-evm/blob/100b8c7d5267a6a6c9ae9c0472963db217b390db/eth/db/backends/level.py#L68-L73

docs: https://plyvel.readthedocs.io/en/latest/user.html#write-batches

veox:
This PR:

``` markdown
* Closes #36 as implemented.
```

carver:
Good to close for now and reopen whenever we want to take another shot @cburgdorf ?

cburgdorf:
@carver I'd prefer to keep it open both as a reminder and to signal that working is being done on that issue. It should be easy to rebase and I plan to get back to it as soon as I finished the other work that I'm currently busy with.

carver:
It's blocked on RocksDB making changes to how read processes work, right?

pipermerriam:
> It's blocked on RocksDB making changes to how read processes work, right?

I don't actually see this as a blocker assuming IUC.   When we make a *read* connection we get a static view of the db.  The places we need *read-only* connections are things like:

- serving peer requests
- serving JSON-RPC requests

Both of these seem totally viable with a temporary static read connection, or even some simple *pooling* of these connections.

cburgdorf:
Just noting (mostly to my future self coming back to this) that I *think* the feature has landed in rocksdb 6.0. Need to test it out.

carver:
I'll say that if the motivation for this is mostly performance, then let's back it up with some runtime analysis that this will have a major impact. I suspect that other structural things (like separating sync into its own process) is important for both performance and architecture. So I'd advocate for bumping that up your list.

On performance, which I've been digging into, it's been mindblowing which things are slow. Intuition is a really bad guide. For example, did you know that this one method is taking 7% of long `import_block()` runs? https://github.com/ethereum/py-evm/blob/6b1796c419fcd07e4f2bf8270c54f8e72ba4081d/eth/db/journal.py#L66-L71
(I've got a fix to make it 90% faster, which will be pushed up today/tomorrow)

That's just one reason that I think it's really critical to let instrumentation guide performance improvements.

cburgdorf:
>I suspect that other structural things (like separating sync into its own process) is important for both performance and architecture. So I'd advocate for bumping that up your list.

Absolutely! That's pretty much on top of my list but I'm currently blocked on that. It all boils down to the `PeerPool` exposing events on the event bus and we need to get the lahja changes finalized before that is viable (But it's looking to be ready soon!).

>(I've got a fix to make it 90% faster, which will be pushed up today/tomorrow)

Woah! That's pretty cool. Did you use `py-spy` to find it?

>I'll say that if the motivation for this is mostly performance, then let's back it up with some runtime analysis that this will have a major impact

Absolutely! Unfortunately, measuring the impact of that is more fuzzy. The theory behind the performance change goes as follows:

1. This PR (still untested wip changes) makes it so that EthStatsService, JSON-RPC Server and the Peer Request Server all **don't call** into the DB Proxy Process anymore. Instead, they do a local readonly connection to the db.

2. The theory goes on that taking pressure from the DB Proxy Process frees it up to more/faster *writes* for syncing.

3. Based on the assumption that rocksdb is same or faster speed than leveldb in general, these performance gains do sound plausible to me.

However, there are lots of unknown unknowns that can totally ruin the picture which is why I absolutely agree that we should measure the impact. Unfortunately, measuring this on a live network isn't 100 % accurate as things depend on the peers that you are connected to etc.

Maybe you have a better idea how to measure the impact, but my current plan goes as follows:

Perform `n` runs from block `0` to block `x` for each `master` and this branch. Measure the time it takes. It's not accurate but if let's say `n` is `5` and every single run of this branch is faster than every run on `master` than this should be somewhat representative I think.

In fact, I think I'm going to throw together a quick plugin that helps with that. Something like `trinity --track-perf --track-perf-label "First run custom build" --track-perf-target-block 500000`

Then, the plugin would automatically shutdown trinity after block `500000` was hit and it would print performance statistics to the console as well as append them to a file. So, you could run it on several branches back and forth and it would append the relevant info to the file in a structured way.

Do you think that would be valuable? Do you have other ideas how to measure this. This could also be very useful if we start Trinity without discovery and pre-defined peers to get us a bit closer to proper "lab conditions"

carver:
> > (I've got a fix to make it 90% faster, which will be pushed up today/tomorrow)
> 
> Woah! That's pretty cool. Did you use `py-spy` to find it?
> 
> >

I used `cProfile` and a tool called `pyprof2calltree` to convert the profiling data into something that can be visualized with `KCachegrind`. I'm pretty happy with it so far. It's rough around the edges, but gets the job done much better than just looking at `cProfile` summary tables.

> In fact, I think I'm going to throw together a quick plugin that helps with that. Something like `trinity --track-perf --track-perf-label "First run custom build" --track-perf-target-block 500000`
> 
> Then, the plugin would automatically shutdown trinity after block `500000` was hit and it would print performance statistics to the console as well as append them to a file. So, you could run it on several branches back and forth and it would append the relevant info to the file in a structured way.
> 
> Do you think that would be valuable? Do you have other ideas how to measure this. This could also be very useful if we start Trinity without discovery and pre-defined peers to get us a bit closer to proper "lab conditions"

- Yeah I think that would be super valuable!
- Another option is something like: being able to artificially insert delays into particular `lahja` events and see if it causes any slowdown. If you can a particular event type and it doesn't affect total time, then we know that there will be no impact from making it faster (heavily borrowed from an offhand comment by @lithp )

I doubt we'll end up with one performance monitor to rule them all, but a toolbox for different scenarios and aspects.

pipermerriam:
I'd not put the motivation for this primarily in the performance bucket, but rather in the architecture bucket.  A database that allows us to have multiple connections should make our lives a lot easier when we have multiple auxiliary processes that only need read access (JSON-RPC, PeerRequestServer, EthStats).  They would no longer need a handle to the database provided to them and could establish their own.

cburgdorf:
>They would no longer need a handle to the database provided to them and could establish their own

Not sure how big of an advantage that really is though. On a service level, all that matters is getting a `BaseAsyncChainDB` injected. And on the outer level, say, the plugin code bootstrapping the service, getting a handle to the database is pretty straight forward under the current proxy model already.

```
self._db_manager = create_db_consumer_manager(trinity_config.database_ipc_path)
self._chaindb = self._db_manager.get_chaindb()  # type: ignore
```

So, to me, the potential performance benefit is what matters most. Here are some numbers that I crunched with the new performance tracking plugin (#588)

```
Performance tracking finished run: LevelDB Run #1
Hit target block 30000
Total duration: 3m50s
Actual duration: 3m35s
#####################
Performance tracking finished run: LevelDB Run #2
Hit target block 30000
Total duration: 10m55s
Actual duration: 10m40s
#####################
Performance tracking finished run: LevelDB Run #3
Hit target block 30000
Total duration: 7m45s
Actual duration: 7m35s
#####################

ROCKS DB AFTER THIS:

Performance tracking finished run: RocksDB Run #1
Hit target block 30000
Total duration: 3m50s
Actual duration: 3m25s
#####################
Performance tracking finished run: RocksDB Run #2
Hit target block 30000
Total duration: 2m55s
Actual duration: 2m40s
#####################
Performance tracking finished run: RocksDB Run #3
Hit target block 30000
Total duration: 3m35s
Actual duration: 3m20s
#####################
```

So far RocksDB is ahead but I admit that the measurement results may not be representative enough. Let me bump up the target block number a bit more and run the benchmark again.

cburgdorf:
Here's another one syncing blocks to 50k.

```
LEVELDB
Performance tracking finished run: LevelDB Run #1
Hit target block 50000
Total duration: 7m
Actual duration: 6m40s
#####################
Performance tracking finished run: LevelDB Run #2
Hit target block 50000
Total duration: 7m50s
Actual duration: 7m45s
#####################

3rd run crashed -> https://github.com/ethereum/trinity/issues/590

ROCKSDB

#####################
Performance tracking finished run: RocksDB Run #1
Hit target block 50000
Total duration: 7m45s
Actual duration: 7m30s
#####################
Performance tracking finished run: RocksDB Run #2
Hit target block 50000
Total duration: 9m30s
Actual duration: 9m20s
#####################
Performance tracking finished run: RocksDB Run #3
Hit target block 50000
Total duration: 6m50s
Actual duration: 6m35s
#####################

```

This time the LevelDB performance is closer to RocksDB but still worse (matching the two successful leveldb runs against the two best rocksdb runs). It doesn't feel like proper science but I guess with enough runs continuously showing the rocksdb branch winning, it might be slowly building up evidence.

pipermerriam:
Can you have them alternate back and forth indefinitely?  If you give me a branch I can run that on I'll help gather more data.

carver:
There is so much variance between each run that we're not even close to meaningful results yet. You're just connecting to live mainnet? Maybe connect to a dedicated peer that you're running on another machine, and turn off discovery. Having discovery on adds way too much variation, I'm guessing we'd have to do something like 1000s of runs to get statistically significant results.

cburgdorf:
>Can you have them alternate back and forth indefinitely?

Not yet, but I'm planning to add that. So far, I have to run `n` subsequent runs on branch `x` followed by the same number of runs on branch `y`.

I'm currently running https://github.com/cburgdorf/trinity/tree/christoph/feat/rocksdb-pertest vs https://github.com/cburgdorf/trinity/tree/christoph/feat/perf-measure-plugin

I'm running it through supervisord and the config for that looks like this:

```
[program:trinity]
user=ubuntu
environment=HOME="/home/ubuntu"
command=/home/ubuntu/trinity/venv/bin/trinity track-perf --track-perf-trinity-exec '/home/ubuntu/trinity/venv/bin/trinity' --track-perf-report-path '/home/ubuntu/trinity/performance_report.txt' --track-perf-target-block 50000 --track-perf-run-count 3
autostart=false
autorestart=false
redirect_stderr=true
stdout_logfile=/var/log/supervisor/trinity.log
```  

>You're just connecting to live mainnet?

yep

>Maybe connect to a dedicated peer that you're running on another machine, and turn off discovery. Having discovery on adds way too much variation, I'm guessing we'd have to do something like 1000s of runs to get statistically significant results.

That's true but connecting to just one single machine may also artificially skew results. I think perfect lab conditions would be if we had something like 10 boxes to connect to. None of these would be connected to anyone else except our testing box (so that the serving side also doesn't have much variation). 

But I think if we modify the performance benchmark so that it can alternate runs between two different versions continuously and then do something like 100 runs and if then the total time, median and average would show a clear winner, I guess that would mean something.

cburgdorf:
Quick update after 30 runs (15 on each side) from block 0 to 30k

image


The next thing on my list is teach the benchmark runner to pass additional params to the Trinity instance and then run this without discovery disabled and connected to a pre-defined single box.

cburgdorf:
Here's [another update](https://gist.github.com/cburgdorf/1d4678f135dea129147c3627dd609c4e) a/b testing without discovery and only connected to a single peer that I control.

There's still a lot of variance in these numbers but in general I've seen more faster runs with rocksdb compared to leveldb. The numbers aren't as strong as I'd hope. @pipermerriam are you still up for putting this to work on another machine to collect more evidence. I can guide you through the setup.

veox:
> But creating a fresh readonly db is very costly and currently in this PR we only cache readonly instances for 100ms. Given how costly the creation is and taking into consideration that the places where we need read access can probably also live with less fresh data, I think I'm going to bump this up to something like 1000ms and see if it has a notable effect in the live benchmarks.

They can probably be cached once per block; that is, once per "chain segment" import - regular or reorg; that is, once per DB write... But that might be too much coupling between the "read" and "write" sides, and would introduce thrashing in case of many consecutive writes.

cburgdorf:
There is still a chance that we can get entirely rid of the costly reconnecting because apparently the feature [is now available in rocksdb 6.1.](https://github.com/facebook/rocksdb/issues/908#issuecomment-491353242). It's just a matter of figuring out how to use it and probably contributing upstream changes to `python-rocksdb`.

trinity/main.py
```diff
@@ -198,11 +206,11 @@ def launch_node(args: Namespace, trinity_config: TrinityConfig) -> None:
 
 @setup_cprofiler('run_database_process')
 @with_queued_logging
-def run_database_process(trinity_config: TrinityConfig, db_class: Type[BaseDB]) -> None:
+def run_database_process(trinity_config: TrinityConfig) -> None:
```
veox: `trinity/main_beacon.py` has a (copy-paste?) function with same name and signature, and differs only by one hard-coded in-line parameter (`Eth1AppConfig` vs. `BeaconAppConfig`).

Refactoring this "launch infrastructure" is probably out of scope in this PR; just wanted to point out there's `LevelDB`s lurking elsewhere.

cburgdorf:
I'm going to close this. Not saying it dead but at least not a priority for now and the code has become stale.

PR #176, opened 2019-01-15
Postpone Constantinople as hotfix release

carver:
### What was wrong?

Constantinople likely to be postponed due to re-entrancy concerns.

See:
- https://medium.com/chainsecurity/constantinople-enables-new-reentrancy-attack-ace4088297d9
- https://gitter.im/ethereum/AllCoreDevs?at=5c3e0fb51cb70a372ae2c1f7

### How was it fixed?

Set fork block number by 9876543210

#### Cute Animal Picture


carver:
Yikes, it's going to be a big PITA to deal with the fact that eth & trinity split since the last release. Considering just doing a release of current master instead...

pipermerriam:
Test failures look to just be two numbers which are off in our chain configuration file.

carver:
> Test failures look to just be two numbers which are off in our chain configuration file.

Yeah, I realize now that really the only annoyance is dealing with putting the changes on both branches. I thought it was going to be more problematic. Current thought is that the hotpatch is the best option.

PR #179, opened 2019-01-16
Append cached replacement node when a node is removed in KBucket.

zsaladin:
### What was wrong?
I found Replacement nodes are not used in KBucket.

### How was it fixed?
When the number of nodes in KBucket decreases, one of the replacement nodes is added.

#### Cute Animal Picture

### DIFF
--- a/tests/p2p/test_kademlia.py
+++ b/tests/p2p/test_kademlia.py
@@ -123,21 +123,25 @@ def test_kbucket_remove():
     nodes = [random_node() for _ in range(bucket.k)]
     for node in nodes:
         bucket.add(node)
-    assert len(bucket) == bucket.k
+    assert bucket.nodes == nodes
+    assert bucket.replacement_cache == []
 
     replacement_count = 10
     replacement_nodes = [random_node() for _ in range(replacement_count)]
     for replacement_node in replacement_nodes:
         bucket.add(replacement_node)
-    assert len(bucket) == bucket.k
+    assert bucket.nodes == nodes
+    assert bucket.replacement_cache == replacement_nodes
 
     for node in nodes:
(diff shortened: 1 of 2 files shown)


zsaladin:
Added a test for this.

zsaladin:
I have a question.
If a replacement node of a KBucket is passed as a parameter in `remove_node` then what will happen?

Removing process will be ignored in current codes.
If the replacement node has to be removed from KBucket then I will fix it.
Please give me your opinion. @pipermerriam 

Thank you.

tests/p2p/test_kademlia.py
```diff
@@ -116,6 +116,30 @@ def test_kbucket_add():
...
+    replacement_nodes = [random_node() for _ in range(replacement_count)]
+    for replacement_node in replacement_nodes:
+        bucket.add(replacement_node)
+    assert len(bucket) == bucket.k
+
+    for node in nodes:
+        bucket.remove_node(node)
+    assert len(bucket) == replacement_count
+
+    for replacement_node in replacement_nodes:
```
pipermerriam: Maybe add one more set of assertions before removing these to check that each node in `replacement_nodes` is in the bucket, as well as asserting they are *not* in the bucket prior to removing all of the original nodes.

pipermerriam:
I opened an issue to track the issue about `remove_node` being called with a node in the list of replacement nodes https://github.com/ethereum/trinity/issues/187

tests/p2p/test_kademlia.py
zsaladin: I modified test codes to clear how it works.

pipermerriam:
@zsaladin thanks for finding and fixing this.  much appreciated.

PR #180, opened 2019-01-16
Make usage of proxy classes more robust

cburgdorf:
### What was wrong?

1. Our current proxy classes are a known source of fatal bugs. The reason for that is because they only derive from `BaseProxy` (e.g. `ChainDBProxy(BaseProxy)`) and then implement whatever they want. So, there is *zero* guarantee that a `ChainDBProxy` will actually have the methods of a `BaseChainDB` which has lead to many run time crashes in Trinity over time.

2. The complexity of classes involved is higher than it needs to be. E.g. taking a look at `AsyncHeaderDB` it may seem that the purpose of this class is to be an *abstract* definition.

https://github.com/ethereum/trinity/blob/0d76f212ac71ba8567533c813d549f640403e953/trinity/db/header.py#L70-L94

But that is not the case, the abstract definition is the `BaseAsyncHeaderDB`  as shown below:

https://github.com/ethereum/trinity/blob/0d76f212ac71ba8567533c813d549f640403e953/trinity/db/header.py#L29-L67

So, what is the purpose of `AsyncHeaderDB` if all it does is to derive from `HeaderDB` and throw in the async methods? 

The answer is that the manager instantiates an `AsyncHeaderDB` (which derives from `HeaderDB`) to use that as the backend for the `AsyncHeaderDBProxy`.

https://github.com/ethereum/trinity/blob/0d76f212ac71ba8567533c813d549f640403e953/trinity/db/manager.py#L29-L47

But there's no reason for that added complexity because the `AsyncHeaderDBProxy` just calls methods of the `HeaderDB` and so a `HeaderDB` serves just fine as a backend for that proxy.

In other words, there is no reason why we should write an implementation without implementation. That's what the *abstract*  base class is for. And since we use three different DB types + the beacon chain does the same, the complexity adds up (6 superfluous classes basically) .

### How was it fixed?

1. This removes all the extra classes that **were not abstract classes** yet did not provide an implementation.

2. This introduces a pattern of implementing a `*PreProxy` that **does** implement an abstract base class and hence throws if an abstract method is not implemented. It also adds a very simple test that just instantiates each `PreProxy`. **This eliminates a common source of Trinity bugs**. In fact, this PR already adds method implementations that the new tests revealed that were missing.

3. This fixes naming inconsistency issues. All abstract base classes begin with `Base*` to be more in line with the rest of the code base.

#### Cute Animal Picture

### DIFF
--- a/trinity/db/chain.py
+++ b/trinity/db/chain.py
@@ -14,19 +14,19 @@ from typing import (
 from eth_typing import Hash32
 
 from eth.db.chain import ChainDB
+from eth.db.header import HeaderDB
 from eth.rlp.blocks import BaseBlock
 from eth.rlp.headers import BlockHeader
 from eth.rlp.receipts import Receipt
 from eth.rlp.transactions import BaseTransaction
 
-from trinity.db.header import AsyncHeaderDB
 from trinity._utils.mp import (
     async_method,
     sync_method,
 )
 
 
-class AsyncChainDB(ChainDB, AsyncHeaderDB):
+class AsyncChainDB(ChainDB, HeaderDB):
     async def coro_get(self, key: bytes) -> bytes:
         raise NotImplementedError()
 
--- a/trinity/db/chain.py
+++ b/trinity/db/chain.py
@@ -1,3 +1,4 @@
+from abc import abstractmethod
 # Typeshed definitions for multiprocessing.managers is incomplete, so ignore them for now:
 # https://github.com/python/typeshed/blob/85a788dbcaa5e9e9a62e55f15d44530cd28ba830/stdlib/3/multiprocessing/managers.pyi#L3
 from multiprocessing.managers import (  # type: ignore
@@ -13,8 +14,8 @@ from typing import (
 
 from eth_typing import Hash32
 
-from eth.db.chain import ChainDB
-from eth.db.header import HeaderDB
+from eth.db.backends.base import BaseAtomicDB
+from eth.db.chain import BaseChainDB
 from eth.rlp.blocks import BaseBlock
 from eth.rlp.headers import BlockHeader
 from eth.rlp.receipts import Receipt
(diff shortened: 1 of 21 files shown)


trinity/db/chain.py
```diff
@@ -69,12 +89,31 @@ class ChainDBProxy(BaseProxy):
...
+    get_receipts = sync_method('get_receipts')
     get_score = sync_method('get_score')
+    get_transaction_by_index = sync_method('get_transaction_by_index')
+    get_transaction_index = sync_method('get_transaction_index')
     header_exists = sync_method('header_exists')
+    get_canonical_block_header_by_number = sync_method('get_canonical_block_header_by_number')
     get_canonical_block_hash = sync_method('get_canonical_block_hash')
+    persist_block = sync_method('persist_block')
     persist_header = sync_method('persist_header')
+    persist_header_chain = sync_method('persist_header_chain')
```
cburgdorf: @carver @pipermerriam @lithp notice how this PR enforces us to add these methods. I'm even fairly certain that some of these methods **not** being on the proxy right now will cause some runtime errors in Trinity today that we just didn't discover yet (like calling certain JSON-RPC APIs).

We could argue that some of them should only be used in their `async` counterpart and that it would be better to implement some of these as `raise NotImplementedError('Use coro_foobar instead')`. I happily do that if we agree on which one we want to discourage to be used as sync methods.

But the point of this PR is that CI **will enforce some action from us** if e.g. we add a new method on the `ChainDB` and we forgot to implement it on the proxy. Because the alternative (that we currently live with) is that we get a runtime error that e.g `persist_header_chain` does not exist on the object which the rest of Trinity does assume it should since it **thinks** it is dealing with a `BaseAsyncChainDB` when in fact it is only dealing with a `BaseProxy` without any further guarantees.

PR #181, opened 2019-01-17
Update deposit structures

### LINKED ISSUE #149: Update `Deposit` and `DepositData` data structure
### What is wrong?

Per https://github.com/ethereum/eth2.0-specs/pull/388, some renamings and adjustment.

### How can it be fixed

- Update `eth2.beacon.types.deposits.Deposit` fields
- Update `eth2.beacon.types.deposit_data.DepositData` fields

ralexstokes:
### What was wrong?

Fixes #149.

### How was it fixed?

Update the order and naming of the fields to match the spec.

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/types/deposit_data.py
+++ b/eth2/beacon/types/deposit_data.py
@@ -15,20 +15,21 @@ class DepositData(rlp.Serializable):
     Note: using RLP until we have standardized serialization format.
     """
     fields = [
-        ('deposit_input', DepositInput),
         # Amount in Gwei
         ('amount', uint64),
         # Timestamp from deposit contract
         ('timestamp', uint64),
+        # Deposit input
+        ('deposit_input', DepositInput),
     ]
 
     def __init__(self,
-                 deposit_input: DepositInput,
                  amount: Gwei,
-                 timestamp: Timestamp) -> None:
+                 timestamp: Timestamp,
+                 deposit_input: DepositInput) -> None:
 
         super().__init__(
-            deposit_input,
             amount,
             timestamp,
+            deposit_input,
         )
--- a/eth2/beacon/types/deposit_data.py
+++ b/eth2/beacon/types/deposit_data.py
@@ -9,7 +9,10 @@ from eth2.beacon.typing import (
 
 class DepositData(rlp.Serializable):
     """
-    Not in spec, this is for fields in Deposit
+    ``DepositData`` corresponds to the data broadcast from the Ethereum 1.0 deposit
+    contract after a successful call to the `deposit` function.
+
+    Note: using RLP until we have standardized serialization format.
     """
     fields = [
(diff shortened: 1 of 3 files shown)


eth2/beacon/types/deposits.py
```diff
@@ -37,7 +41,7 @@ def __init__(self,
                  merkle_tree_index: int,
                  deposit_data: DepositData)-> None:
         super().__init__(
-            merkle_branch,
-            merkle_tree_index,
-            deposit_data,
+            branch=merkle_branch,
```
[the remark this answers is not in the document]
hwwhww: - I think the params should be changed too.
- And also, `sample_deposit_params` in `tests/eth2/beacon/conftest.py`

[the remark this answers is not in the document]
ralexstokes: the thinking was that if all someone had was the function signature, the param names as they are provide more information than just "branch" and "index".

i'll agree it is a small thing and we may value consistency more here... i'll update now

eth2/beacon/types/deposit_data.py
```diff
@@ -10,23 +9,27 @@
 
 class DepositData(rlp.Serializable):
     """
-    Not in spec, this is for fields in Deposit
+    ``DepositData`` corresponds to the data broadcast from the Ethereum 1.0 deposit
```
pipermerriam: For RST notation this would be:

```python
class Thing:
    """
    :class:`~eth2.beacon.types.deposit_data.DepositData`
    """
```

This will properly render as `DepositData` in the docs while properly referencing the class definition.

eth2/beacon/types/deposits.py
```diff
@@ -20,24 +20,28 @@
 
 class Deposit(rlp.Serializable):
     """
+    A ``Deposit`` contains the data represented by an instance of ``DepositData``,
```
pipermerriam: Same here for cross-referencing classes in docstrings.  Can you convert to restructured text `:class:` references.

eth2/beacon/types/deposit_data.py
```diff
@@ -10,23 +9,27 @@
...
+    contract after a successful call to the `deposit` function.
```
pipermerriam: Same here for the `deposit` function assuming you are referencing a local function, otherwise, this should be converted to double backticks to properly render in the docs.

eth2/beacon/types/deposit_data.py
```diff
@@ -10,23 +9,27 @@
...
+
+    Note: using RLP until we have standardized serialization format.
```
pipermerriam: Can you convert this to a sphinx `.. note::` directive which will improve rendering in the docs.

ralexstokes:
@pipermerriam addressed your comments!

PR #188, opened 2019-01-17
Prepare BCC protocol functionality for syncing

jannikluhn:
Changes to the protocol and boilerplate in preparation for BCC syncing (originally from #163 without any new code).

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/bcc/commands.py
+++ b/trinity/protocol/bcc/commands.py
@@ -1,4 +1,5 @@
 from typing import (
+    Tuple,
     Union,
 )
 
@@ -10,6 +11,10 @@ from mypy_extensions import (
 
 from eth_typing import (
     Hash32,
+)
+
+from eth2.beacon.typing import (
+    SlotNumber,
 )
 
 from p2p.protocol import (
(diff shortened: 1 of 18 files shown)


trinity/protocol/bcc/commands.py
```diff
@@ -24,17 +29,35 @@
 from eth2.beacon.types.attestations import Attestation
 
 
+RequestMessage = TypedDict("RequestMessage", {
```
cburgdorf: I quick scan through the Trinity code base shows that we've already been mixing the Python 3.5 and the "modern Python" form of `TypedDict` usage. 

https://github.com/ethereum/trinity/blob/cdb617df0bb11610b3bb7636465ae60d67938556/trinity/rpc/modules/eth.py#L285-L288

If there's no specific reason (which I can't think of from the top of my head), could we update this to use the class based syntax?

cburgdorf: Did my part to migrate the existing place that used the old syntax https://github.com/ethereum/trinity/pull/195

PR #189, opened 2019-01-17
Bcc syncing logic

jannikluhn:
The actual BCC syncer. This PR builds on top of #188. It's originally from #163 and contains no new code.

#### Cute Animal Picture


trinity/sync/beacon/chain.py
```diff
@@ -0,0 +1,177 @@
...
+        if parent_slot < 0:
+            raise Exception(
+                "Invariant: Syncing starts with the child of a finalized block, so never with the "
+                "genesis block"
+            )
+
+        canonical_parent = self.chain_db.get_canonical_block_by_slot(parent_slot, BeaconBlock)
+        if canonical_parent.hash != parent_root:
+            self.logger.info(f"Peer has different block finalized at slot #{parent_slot}")
+            raise ValidationError()
```
cburgdorf: Maybe put `f"Peer has different block finalized at slot #{parent_slot}"` in a variable and reuse that in `ValidationError(msg)`.

PR #191, opened 2019-01-17
Bootstrap beacon node db process

cburgdorf:
### What was wrong?

Beacon node needs a db process it can interact with

### How was it fixed?

- split `trinity.db` into `trinity.db.eth1` and `trinity.db.beacon` namespaces
- some copy, some paste, some renames and some cleanups :)

#### Known issues

This does not yet handle initialization of the beacon chain because:

a.) I think that might depend on #156 
b.) I'm not sure if we can even use the same static initialization scheme that Ethereum 1 uses.
c.) it shouldn't matter at this point

#### Cute Animal Picture


cburgdorf:
@pipermerriam I believe for something as low level as reading/updating a key from the database using the `EventBus` as-is will have a worse performance. Keep in mind that every single event going through the EventBus is a multi process hop, thing. So, as much as I like the `EventBus`, I do believe stdlib proxy objects are still our best tool for low high frequency / low level things such as database access.

Nothing though that #173 is pretty much ready to bring back your RocksDB work so that we can use a *local*  readonly connection to the database where ever possible.

PR #193, opened 2019-01-17
[conceptual] move compression to base peer

carver:
### What was wrong?

Something still feels wrong about passing around the snappy boolean everywhere. Inspired by #175

### How was it fixed?

Move compression (like encryption) to the base peer.

#### Cute Animal Picture

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -189,6 +189,7 @@ class BasePeer(BaseService):
     listen_port = 30303
     # Will be set upon the successful completion of a P2P handshake.
     sub_proto: protocol.Protocol = None
+    proto_version: int = None
 
     def __init__(self,
                  remote: Node,
@@ -406,6 +407,12 @@ class BasePeer(BaseService):
                 self.logger.debug("%r disconnected: %s", self, e)
                 return
 
+    def decompress(self, maybe_compressed: bytes) -> bytes:
+        if self.is_compressing:
+            return maybe_compressed[0] + snappy.decompress(maybe_compressed[1:])
+        else:
+            return maybe_compressed
+
     async def read_msg(self) -> Tuple[protocol.Command, protocol.PayloadType]:
         header_data = await self.read(HEADER_LEN + MAC_LEN)
         try:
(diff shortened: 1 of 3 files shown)


p2p/peer.py
```diff
@@ -406,6 +407,12 @@ def is_closing(self) -> bool:
                 self.logger.debug("%r disconnected: %s", self, e)
                 return
 
+    def decompress(self, maybe_compressed: bytes) -> bytes:
+        if self.is_compressing:
```
veox: FTR, it is this branching (on every payload, here and in `compress()`) that I was trying to avoid in PR #175. Maybe that's too frugal, agreed; haven't run any profiling.

carver: Ah, yeah. It doesn't seem likely to me that the `if` check is a meaningful performance cost.

To me, the passing around of the `snappy_compression` boolean is the bigger problem, and this gets to remove all of it.

---

I had intended this as an idea that you might be interested in picking up in your PR. If you're not into it, though, I'll go ahead and flesh it out into a real one.

veox: Yes, `snappy` being passed around is the real deal.

I tried the other approach in [your comment in PR #175](https://github.com/ethereum/trinity/pull/175/files#r248863617), moving it to `Command` (technically, leaving it as a property of `Protocol`, and viewing that as a collection of `Command`s); but encountered difficulties. That "solution" is now in a `git stash` somewhere.

If you have more pressing things to do, I'd like to attempt this a few more times.

-----

OT: I'm [a bit concerned with compression](https://ethereum-magicians.org/t/random-critique-of-eip-706-devp2p-snappy-compression-too-much-compression-will-kill-you/2469) being an all-or-nothing feature. Moving it into `BasePeer` makes it harder for only some messages to use it.

carver: > So is `cmd_offset`, probably...

Ooh, interesting. One cool thing about that is each `Command` could be a class instead of an instance, which would save on a good deal of object initialization.

> That "solution" is now in a git stash somewhere.

:laughing: 

> If you have more pressing things to do, I'd like to attempt this a few more times.

:+1: Please go ahead and keep hacking at it.

---

> I'm [a bit concerned with compression](https://ethereum-magicians.org/t/random-critique-of-eip-706-devp2p-snappy-compression-too-much-compression-will-kill-you/2469) being an all-or-nothing feature. Moving it into `BasePeer` makes it harder for only some messages to use it.

Yeah, fair enough. I can imagine leaning that way in another version of devp2p. (although is compression more computationally intensive than encryption?)

veox: > (although is compression more computationally intensive than encryption?)

Having slept on it - probably not.

veox: See PR #205 for first WIP demo that can do fast-sync.

PR #197, opened 2019-01-18
Fix KBucket.add for a duplicate node of replacement_cache.

zsaladin:
### What was wrong?
- `KBucket.add` does not handle a duplicate node of `KBucket.replacement_cache`.

### How was it fixed?
- Add codes similar to handling duplicate nodes in `KBucket.nodes`

#### Cute Animal Picture

### DIFF
--- a/tests/p2p/test_kademlia.py
+++ b/tests/p2p/test_kademlia.py
@@ -1,4 +1,5 @@
 import random
+import itertools
 
 import pytest
 
@@ -116,17 +117,6 @@ def test_kbucket_add():
     assert bucket.head == node2
     assert bucket.replacement_cache == [node3]
 
-    node4 = random_node()
-    assert bucket.add(node4) == node2
-    assert bucket.nodes == [node2, node]
-    assert bucket.head == node2
-    assert bucket.replacement_cache == [node3, node4]
-
-    assert bucket.add(node3) == node2
-    assert bucket.nodes == [node2, node]
-    assert bucket.head == node2
-    assert bucket.replacement_cache == [node4, node3]
-
 
 def test_kbucket_remove():
     bucket = kademlia.KBucket(0, 100)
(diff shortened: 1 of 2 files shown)


tests/p2p/test_kademlia.py
```diff
@@ -114,6 +114,18 @@ def test_kbucket_add():
...
+    node4 = random_node()
+    assert bucket.add(node4) == node2
+    assert bucket.nodes == [node2, node]
+    assert bucket.head == node2
+    assert bucket.replacement_cache == [node3, node4]
+
+    assert bucket.add(node3) == node2
+    assert bucket.nodes == [node2, node]
+    assert bucket.head == node2
+    assert bucket.replacement_cache == [node4, node3]
```
zsaladin: Original codes works like below.
bucket.replacement_cache == [node3, node4, node3]

pipermerriam (approved):
It'd be nice to have a specific test that calls `bucket.remove()` since it's the simplest example.  :+1: to the current test but it mostly tests this using side effects and having at least one simple direct test would be nice (ideally as a completely independent test)

tests/p2p/test_kademlia.py
```diff
@@ -144,6 +146,49 @@ def test_kbucket_remove():
...
+    assert bucket.nodes == list(reversed(nodes))
+    assert bucket.replacement_cache == replacement_nodes[1:] + replacement_nodes[:1]
+
+    for replacement_node in reversed(replacement_nodes):
+        assert bucket.add(replacement_node) == nodes[-1]
+    assert bucket.nodes == list(reversed(nodes))
+    assert bucket.replacement_cache == list(reversed(replacement_nodes))
+
+    all_nodes = list(itertools.chain(nodes, replacement_nodes))
+    random.shuffle(all_nodes)
```
carver: Having the test depend on `random` in this way would make it hard to reproduce if there was a particular ordering that was problematic. `hypothesis` is a good tool for this kind of thing. I think these are the main properties we are looking for in these particular tests:
- after the bucket is full, a newly added node is always at the end of the `replacement_cache`
- at all times, the `replacement_cache` is unique, ie~ `len(set(bucket.replacement_cache)) == len(bucket.replacement_cache)`

If you just throw a list of random nodes as input via hypothesis and meet those criteria ^ I think we are showing that the method is working correctly. That should make for a shorter, easier-to-read test.

(BTW, I see you aren't the first to introduce `random` into these tests, but let's not continue the old mistake :))

carver (commented):
@zsaladin this change is small enough that it would be fine to add the `bucket.remove()` updates to the same PR. That will make for some more interesting hypothesis tests as well, like: add and then remove never leaves that node remaining in `.nodes` or `.replacement_cache`.

Looks like it needs a rebase, too.

carver:
Oh and chances are good that the `cli` test failure was just our CI being wacky before and you shouldn't have problems on the next force-push.

PR #202, opened 2019-01-21
Fix some broken imports

cburgdorf:
### What was wrong?

Looks like #188 broke some imports.

### How was it fixed?

Replaced `eth.beacon` with `eth2.beacon`.

#### Cute Animal Picture


cburgdorf:
I'm going into full anarchy mode and merge this when it's green to unbreak CI.

PR #203, opened 2019-01-21
Fix #190: verify_slashable_vote_data

### LINKED ISSUE #190: Fix `domain` in `verify_slashable_vote_data`
### What is wrong?

https://github.com/ethereum/eth2.0-specs/pull/465

### How can it be fixed
Update `eth2/beacon/helpers.py` and `tests/eth2/beacon/test_helpers.py`.

NIC619:
Fix #190 

### How was it fixed?
- Change `state.slot` to `vote_data.data.slot`.
- Add a test to check that slashable data prior to a fork is still slashable.

### DIFF
--- a/tests/eth2/beacon/test_helpers.py
+++ b/tests/eth2/beacon/test_helpers.py
@@ -877,30 +877,6 @@ def test_verify_slashable_vote_data_signature(num_validators,
     invalid_votes = SlashableVoteData(**invalid_params)
     assert not verify_slashable_vote_data_signature(state, invalid_votes)
 
-    # Test that slashable data is still valid after fork
-    # Slashable data slot = 10, fork slot = 15, current slot = 20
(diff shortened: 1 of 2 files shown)


tests/eth2/beacon/test_helpers.py
```diff
@@ -877,6 +877,30 @@ def test_verify_slashable_vote_data_signature(num_validators,
...
+
+    valid_params = _correct_slashable_vote_data_params(
+        num_validators,
+        sample_slashable_vote_data_params,
+        messages,
+        privkeys,
+        future_state.fork_data,
+    )
+    valid_votes = SlashableVoteData(**valid_params)
+    assert verify_slashable_vote_data_signature(future_state, valid_votes)
```
hwwhww: I'd suggest you add another new test case instead of adding it right after `invalid_votes` tests.

PR #204, opened 2019-01-21
Remove shard committees from the state

### LINKED ISSUE #161: Remove shard committees from the state
### What is wrong?

https://github.com/ethereum/eth2.0-specs/pull/408
1. Remove shard committees from the state and instead in the spec specifies them as needing to be calculated in real time.
2. Implement the cache of shard committees.

### How can it be fixed

hwwhww:
### What was wrong?
Fix #161

### How was it fixed?
Rewrite:
- `get_crosslink_committees_at_slot`
- `get_shuffling`
- `get_attestation_participants`

Add: 
- `get_committee_count_per_slot`
- `get_previous_epoch_committee_count_per_slot`
- `get_current_epoch_committee_count_per_slot`

Fix all the functions that called them.

#### Cute Animal Picture

640px-baby_fur_seal _south_georgia_by_Ville_Miettinen


eth2/beacon/helpers.py
```diff
@@ -106,274 +102,299 @@ def get_block_root(
...
-            total_validator_count=total_validator_count,
+def get_committee_count_per_slot(active_validator_count: int,
+                                 shard_count: int,
+                                 epoch_length: int,
+                                 target_committee_size: int) -> int:
+    return max(
+        1,
+        min(
+            shard_count // epoch_length,
+            active_validator_count // epoch_length // target_committee_size,
```
pipermerriam: Isn't this `max/min` pattern effectively implemented in `eth._utils.clamp`

eth2/beacon/helpers.py
```diff
@@ -106,274 +102,299 @@ def get_block_root(
...
+            target_committee_size=target_committee_size,
+            shard_count=shard_count,
+        )
+
+    for i in range(committees_per_slot):
+        index = committees_per_slot * offset + i
+        committee = shuffling[index]
+        yield (
+            committee,
+            ShardNumber((slot_start_shard + i) % shard_count),
```
pipermerriam: This value is constant over the lifeycle of the loop.  You could instantiate it once and re-use the value rather than re-instantiating for each iteration.

eth2/beacon/helpers.py
```diff
@@ -106,274 +102,299 @@ def get_block_root(
...
+    return first_committee[slot % len(first_committee)]
 
 
 #
 # Bitfields
 #
+@to_tuple
+def _get_committees_for_shard(
+        crosslink_committees: Sequence[Tuple[Sequence[ValidatorIndex], ShardNumber]],
+        shard: ShardNumber) -> Iterable[Iterable[ValidatorIndex]]:
```
pipermerriam: Should this be `Iterable[Tuple[ValidatorIndex, ...]]`?  I assume the inner type here is actually known and not just a generic iterable.

eth2/beacon/helpers.py
```diff
@@ -106,274 +102,299 @@ def get_block_root(
...
+    # Find the committee in the list with the desired shard
+    crosslink_committees = get_crosslink_committees_at_slot(
+        state=state,
+        slot=attestation_data.slot,
+        epoch_length=epoch_length,
+        target_committee_size=target_committee_size,
+        shard_count=shard_count,
     )
+
+    if attestation_data.shard not in [shard for _, shard in crosslink_committees]:
```
pipermerriam: Use a `set` comprehension instead since they are better suited for inclusion checks.

eth2/beacon/helpers.py
```diff
@@ -106,274 +102,299 @@ def get_block_root(
...
+                attestation_data.shard,
+            )
+        )
 
-    if len(participation_bitfield) != get_bitfield_length(len(crosslink_committee.committee)):
+    committee_size = len(committee)
+    if len(aggregation_bitfield) != get_bitfield_length(committee_size):
         raise ValidationError(
             'Invalid bitfield length,'
             "\texpected: %s, found: %s" % (
```
pipermerriam: Use `.format()` style string formatting for consistency.

eth2/beacon/validation.py
```diff
@@ -8,8 +8,39 @@
...
+
+    if state_epoch_slot > slot + epoch_length:
+        raise ValidationError(
+            "state_epoch_slot ({}) should be less than or equal to slot + "
+            "epoch_length ({})".format(
+                state_epoch_slot,
+                slot + epoch_length,
+            )
+        )
+    if slot >= state_epoch_slot + epoch_length:
```
pipermerriam: *nitpick* 

Either make this an `elif` or add a line of whitespace to visually separate these conditionals.

tests/eth2/beacon/conftest.py
```diff
@@ -678,7 +674,7 @@ def create_mock_signed_attestation(state,
         signatures = [
             bls.sign(
                 message,
-                privkeys[crosslink_committee.committee[committee_index]],
+                privkeys[committee[committee_index]],
```
djrtwo: We should probably be using a lookup in the `keymap` using the validator's pubkey from the state rather than the direct array access here. This is assuming that the validator indices will correspond 1-to-1 with the privkeys. This is generally true everywhere in the testing currently but is brittle

tests/eth2/beacon/state_machines/test_attestation_validation.py
```diff
@@ -243,22 +245,33 @@ def test_validate_serenity_attestation_aggregate_signature(genesis_state,
...
+                                                           is_valid,
+                                                           target_committee_size,
+                                                           shard_count,):
     state = genesis_state
 
     # choose committee
     slot = 0
-    crosslink_committee = state.crosslink_committees_at_slots[slot][0]
-    committee_size = len(crosslink_committee.committee)
+    # crosslink_committee = state.crosslink_committees_at_slots[slot][0]
```
djrtwo: delete this line

```suggestion
```

eth2/beacon/helpers.py
hwwhww: I think the `Iterable[ValidatorIndex]` is directly come from `Sequence[ValidatorIndex]` inside the input `Sequence[Tuple[Sequence[ValidatorIndex], ShardNumber]]`?

eth2/beacon/_utils/random.py
```diff
@@ -27,10 +26,10 @@
 
 
 @to_tuple
-def shuffle(values: Sequence[Any],
-            seed: Hash32) -> Iterable[Any]:
+def shuffle(values: Sequence[TItem],
```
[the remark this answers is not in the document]
hwwhww: Ahhh right, we recently just discussed it in `py-ssz` (https://github.com/ethereum/py-ssz/pull/22#discussion_r244277437). After we change the output types of `get_shuffling()`, `mypy` told me to make output type more explict than `Any`. The usage here is to note the relationship between the input and output values.

eth2/beacon/validation.py
```diff
@@ -8,8 +8,39 @@
     validate_is_integer,
 )
 
+from eth2.beacon.typing import (
+    SlotNumber,
+)
+
 
-def validate_slot(slot: Any, title: str="Slot") -> None:
+def validate_slot(slot: int, title: str="Slot") -> None:
```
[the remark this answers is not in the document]
hwwhww: Ahhh right, I tried `SlotNumber` during this PR and found that if I do it, I have to update the whole `BeaconChainDB` types which is off this PR. Opening an issue for it. 👍

eth2/beacon/helpers.py
hwwhww: Yes, there a small `i`.
Refactored to the following to avoid confusion:
```python
    for index in range(committees_per_slot):
        committee = shuffling[committees_per_slot * offset + index]
        yield (
            committee,
            ShardNumber((slot_start_shard + index) % shard_count),
        )
```

eth2/beacon/_utils/random.py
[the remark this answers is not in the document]
pipermerriam: It's a *generic* which is a way for the type system to say, this function operates on a sequence of values, but the type of the value doesn't matter, only that it takes a sequence of them and returns a new (shuffled) sequence of them.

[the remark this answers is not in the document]
pipermerriam: The name comes from the convention setup in the standard library [`typing`](https://docs.python.org/3.5/library/typing.html) module's documentation.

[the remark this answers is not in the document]
djrtwo: Got it! thanks

Issue #90, opened 2018-12-04
Move functionality from `Peer` class down into `Protocol` class.


pipermerriam:
### What is wrong?

A while back I did a reasonably big effort to get the *EVM* stuff out of the `p2p` module.  One of the harder parts was the `Peer` classes.  This lead us to the idea that the functionality that is currently wrapped up in the `Peer` class should be split.

### How can it be fixed

The basic idea is:

- `Protocol` class gets the low level stuff
- `Peer` class gets the higher level stuff
- `Peer` class gets instantiated with a `Protocol` instance.

So, the `Protocol` class becomes a *Service* and gets instantiated with the reader/writer streams and other connection information.  It also gets the `read_msg` loop and needs to expose an API for consumer to stream those messages.  I believe that a single consumer model is appropriate here.  It seems prudent to have the *base* protocol messages partitioned from the *sub* protocol messages so that we can have a dedicated handler for `Ping/Pong/Disconnect`.

The `Peer` class should be instantiated with a running `Protocol` instance.  Running the peer would *stream* the incoming messages from the `Protocol` instance and handle them however it deems appropriate.

This design **should** simplify our concept of what a `Peer` is, removing the *need* for us to actually go through the handshake/connection by separating/decoupling the networking component.

veox:
After fiddling some with PR #205, it seems that `Protocol`, too, could avoid object instantiation, if the reference to `peer` that created it could be removed...

This functionality that perhaps doesn't belong in `Peer` - it seems to me that it doesn't belong in `Protocol`, also.

Should it be isolated then into a new `MessageStream`/`ProtocolMultiplexer` class?..

(Just thinking out loud.)

pipermerriam:
Now that this is underway I've got a bit of a map of what things can/should/might come next.  None of this is set in stone of course or even necessarily the right idea but rather musings that lead towards something hopefully better.

### Further shrink the `Peer` concept.

Basically this: https://github.com/pipermerriam/trinity/pull/2

We'll call this a `Connection` for lack of a better term right now.  A connection binds a `Protocol` to a `Transport`, implementing the business logic needed to do the sub-protocol handshake.

Basically this code and the parts directly related to it become a new class:

https://github.com/ethereum/trinity/blob/926e95cf96159e557a30a7c78f43dbe10c919be6/trinity/protocol/eth/peer.py#L69-L95

I do not believe the `Connection` class needs to be a service.

[...]

PR #205, opened 2019-01-21
[WIP] Make `Command` class "stateless"

veox:
WIP, demo.

Closes #175 (is part of this PR).
Closes #193 (different solution).

See linked PRs for earlier discussion, and some of my arguments.

### What was wrong?

* Every peer's got a `Protocol` (or sub-class) instance, and every instance of that holds (identical) instances of `Command`s in the protocol.
* `cmd_offset` and `snappy_support` - protocol implementation detail and per-connection protocol feature, respectively - are passed around needlessly, requiring every sub-protocol developer to know of their existence.

### How was it fixed?

Only dirty so far.

* A module-level "look-up table" for `Command`s and a function to operate it are introduced.
* `Command`s are referenced by class.
* Each `Protocol` instance maintains a list of command-plus-feature classes they support, plus `dict` mappings (similarly to how it was done previously).
* `Protocol`s are still per-peer objects. This can be addressed later.

TODO:

- [ ] `BCC` and `LES` need to be updated, too!
- [ ] Run tests locally and fix failures.
- [ ] Address TODOs and FIXMEs in code.

### Cute Animal Picture

Source: [Pumpkin the raccoon](https://cuteandkids.com/inspire/pumpkin-the-raccoon/), by Laura Young

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -283,12 +283,12 @@ class BasePeer(BaseService):
         """
         await self.send_sub_proto_handshake()
         cmd, msg = await self.read_msg()
-        if isinstance(cmd, Ping):
+        if cmd.cmd_type == Ping:
             # Parity sends a Ping before the sub-proto handshake, so respond to that and read the
             # next one, which hopefully will be the actual handshake.
             self.base_protocol.send_pong()
             cmd, msg = await self.read_msg()
-        if isinstance(cmd, Disconnect):
+        if cmd.cmd_type == Disconnect:
             msg = cast(Dict[str, Any], msg)
             # Peers sometimes send a disconnect msg before they send the sub-proto handshake.
             raise HandshakeFailure(
@@ -311,7 +311,7 @@ class BasePeer(BaseService):
         except MalformedMessage as e:
             raise HandshakeFailure("Got malformed message during handshake") from e
 
-        if isinstance(cmd, Disconnect):
+        if cmd.cmd_type == Disconnect:
             msg = cast(Dict[str, Any], msg)
             # Peers sometimes send a disconnect msg before they send the initial P2P handshake.
             raise HandshakeFailure(
@@ -431,12 +431,14 @@ class BasePeer(BaseService):
 
     def handle_p2p_msg(self, cmd: protocol.Command, msg: protocol.PayloadType) -> None:
         """Handle the base protocol (P2P) messages."""
-        if isinstance(cmd, Disconnect):
+        cmd_type = cmd.cmd_type
+
+        if cmd_type == Disconnect:
             msg = cast(Dict[str, Any], msg)
             raise RemoteDisconnected(msg['reason_name'])
-        elif isinstance(cmd, Ping):
+        elif cmd_type == Ping:
             self.base_protocol.send_pong()
-        elif isinstance(cmd, Pong):
+        elif cmd_type == Pong:
             # Currently we don't do anything when we get a pong, but eventually we should
             # update the last time we heard from a peer in our DB (which doesn't exist yet).
             pass
@@ -444,7 +446,7 @@ class BasePeer(BaseService):
             raise UnexpectedMessage(f"Unexpected msg: {cmd} ({msg})")
 
     def handle_sub_proto_msg(self, cmd: protocol.Command, msg: protocol.PayloadType) -> None:
-        cmd_type = type(cmd)
+        cmd_type = cmd.cmd_type
 
         if self._subscribers:
             was_added = tuple(
@@ -470,16 +472,16 @@ class BasePeer(BaseService):
     async def process_p2p_handshake(
             self, cmd: protocol.Command, msg: protocol.PayloadType) -> None:
         msg = cast(Dict[str, Any], msg)
-        if not isinstance(cmd, Hello):
+        if not cmd.cmd_type == Hello:
             await self.disconnect(DisconnectReason.bad_protocol)
             raise HandshakeFailure(f"Expected a Hello msg, got {cmd}, disconnecting")
 
-        # Check whether to support Snappy Compression or not
-        # based on other peer's p2p protocol version
+        # Snappy compression is negotiated indirectly, by comparing protocol versions;
+        # if both sides support v5, then compression is mandatory
         snappy_support = msg['version'] >= SNAPPY_PROTOCOL_VERSION
-
-        # Now update the base protocol to support snappy compression
-        self.base_protocol.snappy_support = snappy_support
+        if snappy_support:
+            # TODO: select protocol directly, instead of using side-effect of property
+            self.base_protocol.snappy_support = snappy_support
 
         remote_capabilities = msg['capabilities']
         try:
@@ -693,7 +695,7 @@ class PeerSubscriber(ABC):
     def add_msg(self, msg: PeerMessage) -> bool:
         peer, cmd, _ = msg
 
-        if not self.is_subscription_command(type(cmd)):
+        if not self.is_subscription_command(cmd.cmd_type):
             if hasattr(self, 'logger'):
                 self.logger.debug2(  # type: ignore
                     "Discarding %s msg from %s; not subscribed to msg type; "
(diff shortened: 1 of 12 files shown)


p2p/peer.py
```diff
@@ -301,12 +299,12 @@ def remove_subscriber(self, subscriber: 'PeerSubscriber') -> None:
         """
         await self.send_sub_proto_handshake()
         cmd, msg = await self.read_msg()
-        if isinstance(cmd, Ping):
+        if cmd.cmd_type == Ping:
```
veox: This (here and throughout) is the first very-WIP'y thing. `cmd` is now a class, not an instantiated object of that class.

p2p/protocol.py
```diff
@@ -159,6 +164,37 @@ def encode(self, data: PayloadType) -> Tuple[bytes, bytes]:
         body = _pad_to_16_byte_boundary(enc_cmd_id + compressed_payload)
         return header, body
 
+### FIXME: move to p2p._utils module?..
```
veox: Here and below, module-level stuff that's probably undesirable.

p2p/protocol.py
```diff
@@ -159,6 +164,37 @@ def encode(self, data: PayloadType) -> Tuple[bytes, bytes]:
...
+# FIXME: use kwargs for feature specification; they shouldn't be spelled out here!
+def get_command_class(cmd_class, cmd_id_offset, snappy_support):
+    # TODO: use NamedTuple?..
+    specifier = (cmd_class, cmd_id_offset, snappy_support)
+
+    # use existing if available
+    if specifier in command_classes.keys():
+        return command_classes[specifier]
+
+    class CommandClassInstance(cmd_class):
```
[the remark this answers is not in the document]
veox: Comment outdated because name of class changed.

I did try to use `metaclass` (it didn't work, not with the little effort I gave); but now think that doing `type(cmd)` elsewhere in the code would obfuscate that:

* `cmd` now refers to a class, not an object; in fact, it's a sub-class of what would've previously been returned by `type(cmd)`;
* trying to use return value of `type(cmd)` directly to create another variable referring to same class `cmd` would not yield the expected result.

In short, this would make things even more confusing than already. I'm not well-versed to pursue this.

p2p/peer.py
```diff
@@ -488,23 +488,20 @@ def process_msg(self, cmd: protocol.Command, msg: protocol.PayloadType) -> None:
...
+        # Snappy compression is negotiated indirectly, by comparing protocol versions;
+        # if both sides support v5, then compression is mandatory
         snappy_support = msg['version'] >= SNAPPY_PROTOCOL_VERSION
-
         if snappy_support:
-            # Now update the base protocol to support snappy compression
-            # This is needed so that Trinity is compatible with parity since
-            # parity sends Ping immediately after Handshake
-            self.base_protocol = P2PProtocol(self, snappy_support=snappy_support)
+            # TODO: select protocol directly, instead of using side-effect of property
```
pipermerriam: :+1:

p2p/protocol.py
```diff
@@ -55,94 +55,99 @@ class TypedDictPayload(TypedDict):
...
-        self.cmd_id = cmd_id_offset + self._cmd_id
-        self.snappy_support = snappy_support
-
+    # FIXME: does this even work?
     @property
-    def logger(self) -> logging.Logger:
-        if self._logger is None:
-            self._logger = logging.getLogger(f"p2p.protocol.{type(self).__name__}")
-        return self._logger
+    def logger(cls) -> logging.Logger:
```
pipermerriam: This won't work if we're interacting with these as classes.  It'll need to be converted to a static property **or** a metaclass if we want to get fancy and have the logger instance be a class variable but not have to manually set it each time.

p2p/protocol.py
```diff
@@ -55,94 +55,99 @@ class TypedDictPayload(TypedDict):
...
-    @property
-    def is_base_protocol(self) -> bool:
-        return self.cmd_id_offset == 0
+    # TODO: add abstract methods for is_base_protocol
 
-    def __str__(self) -> str:
-        return f"{type(self).__name__} (cmd_id={self.cmd_id})"
+    # FIXME: reference to `self`
+    # def __str__(self) -> str:
+    #     return f"{type(self).__name__} (cmd_id={cls.cmd_id})"
```
pipermerriam: This might require a metaclass which I don't see being a problem.

https://stackoverflow.com/questions/8144026/how-to-define-a-str-method-for-a-class

p2p/protocol.py
```diff
@@ -159,6 +164,38 @@ def encode(self, data: PayloadType) -> Tuple[bytes, bytes]:
...
+# FIXME: use kwargs for feature specification; they shouldn't be spelled out here!
+def get_command_class(cmd_class, cmd_id_offset, snappy_support):
+    # TODO: use NamedTuple?..
+    specifier = (cmd_class, cmd_id_offset, snappy_support)
+
+    # use existing if available
+    if specifier in command_classes.keys():
+        return command_classes[specifier]
+
+    class CommandWithSettings(cmd_class):
```
pipermerriam: Dynamically generated classes are going to make type checking incredibly hard and this strikes me as equivalent to using instances of the command class so it seems like it defeats the purpose of this pull request since now you're effectively replacing instances with dynamically constructed classes...

veox: This is somewhat tangential, but objects of this class don't have to be re-instantiated on every new connection (via `Protocol` (via `Peer`)) - so they no longer are. It is the same in-memory object (describing a class and its class methods), referenced through thousands of connections within a lifetime of a node run (mostly ending prematurely with "wrong network" and "too many peers"). So, at the very least, it's not equivalent.

"Tangential", because while striving for `snappy_support` to not be passed around in [sub-]protocol implementations, this PR passes around references to some `Command` class in general. The implementing `Protocol` then looks up a sub-class of the `Command` that it's supposed to be _always_ using (with appropriate `cmd_id_offset` and `snappy_support`).

The purpose of this PR is to stop referencing [sub-]protocol features when issuing commands belonging to that [sub-]protocol; and to stop creating thousands of identical `Command` objects; in fact, stop creating `Command` objects altogether.

p2p/peer.py
veox: I don't think this can be done in a meaningful manner until `Protocol`s have their own `get_protocol_class()` similar to this PR's `get_command_class()`.

To do it right now, the existing line can be kept. The difference is only that previously, a new `P2PProtocol` object must be created, whereas with this PR, only most of it (the property and the command look-up tables). That, and the fact that "P2PProtocol" features again (compare to "whatever is already in use"), and the `self` is passed.

p2p/peer.py
veox: This line is currently testing for the sub-class. The generic, "protocol-unbound" class, e.g.:

`<class 'trinity.protocol.eth.commands.Transactions'>`

... is [saved as a class attribute](https://github.com/ethereum/trinity/pull/205/files#diff-631c68b238129bf0fb825fe6abb229cbR204) when getting a "protocol-bound" variant of the command with `cmd_id_offset=16` (always the same number for a specific protocol) and `snappy_support=[True|False]` (depending on whether compression is enabled). This "protocol-bound" class is currently ungracefully `__repr__`'d as:

 `<class 'trinity.protocol.eth.commands.Transactions'><cmd_id_offset=16,snappy_support=`[True|False]>`)

Point is, it's not exact matching being tested here, intentionally.

p2p/protocol.py
pipermerriam: @veox evidently I never submitted this the other day.  Take a look at https://github.com/ethereum/py-wasm/blob/5b6b61790fc226ff069ecdf179857647113703df/wasm/_utils/singleton.pyfor a working implementation.

veox: Yes, that seems much more concise.

I've decided to lay this off myself for now, as I didn't feel competent enough; but I could try plugging that in (and undo the jolly "why, we'll be passing classes around instead"), if no one else does.

veox:
Probably best to close this, since it's pretty stale; and might be easier to pull off after the plans described in https://github.com/ethereum/trinity/issues/90#issuecomment-488193338 are settled/implemented.

Issue #55, opened 2018-09-04
Concurrent chain data and state sync


pipermerriam:
### What is wrong?

Currently we *first* sync the chain data (blocks, receipts, etc) and then sync the state trie.  This approach won't reliably work for doing a full sync of the chain because:

- State sync currently takes around 5-10 days
- The state root which it syncs against is static
- Geth (and probably parity) garbage collect old state tries (geth defaults to only keeping state data for 128 blocks)

This means that unless trinity happens to be connected to a node running in archive mode (extremely rare) that by the time it completes the state sync there will be missing trie nodes which will not be available from most any peer it connects to.

And **even if** trinity happens to be connected to an archive node, in order to hit the necessary performance numbers for Trinity to be viable, we will need to process these two syncs concurrently.

### How can it be fixed

An initial exploratory attempt can be found here: https://github.com/ethereum/py-evm/pull/1231

Running the two sync processes concurrently is reasonably easy to do, however, the problems that need to be solved are related to how we orchestrate these two sync processes.

### HEAD tracking

See: https://github.com/ethereum/trinity/issues/54

We need a reference to the HEAD chain header that we are syncing towards.  

- For chain data sync, this merely acts as an anchor for when we have fully synced.
- For state trie, we use the `state_root` as our sync target.

Currently the chain data sync updates the target but the state trie sync does not.   We need to address the security issues laid out in #1233 as well as implement a mechanism for the state sync to update the `state_root` that it is targeting.

### Decouple header and body syncing

> This isn't exactly required, however I believe it will make this easier to manage.

Currently chain data sync syncs headers and bodies together.  This should be separated such that we sync the full header chain **ahead** of the block bodies, effectively making it into two sync processes.

- One process which syncs only the headers (akin to light sync)
- Another process which syncs the bodies.

By moving the header sync into a dedicated process we can institute the head tracking in that process and then cue off of the current target HEAD in the block body and state trie sync.

### Updating the target `state_root`

A naive approach for updating the `state_root` is to simply abandon the previous syncer and start a new one.  This however is likely to incur potentially significant performance overhead since it will require re-walking the entire state trie.

Intelligent updating of the state root would ideally involve 

- preserving all of the `SyncRequest` objects that are still valid for the new state root
- discarding all of the `SyncRequest` objects which are no longer valid **and** removing their entries from the database.

[...]

PR #210, opened 2019-01-22
[proof of concept] State Sync Rebasing

lithp:
This is a not-quite-functional slightly-embarrassing spike of a solution to #55, getting state syncing and chain syncing to happen concurrently.

- `StateDownloader` no longer accepts a `root_hash` argument. Instead, it spawns a `ETHVerifiedTipMonitor` and changes it's sync target whenever the monitor says that a new highest block has appeared.
- `ETHVerifiedTipMonitor` is like a `ChainTipMonitor` except instead of firing whenever a possibly-relevant message is received it does a little work to extract the block header and return that to its listeners. (This is related to https://github.com/ethereum/trinity/issues/54) It fetches the tip from every peer that connects, it also listens for `NewBlock` messages, and it yields the new highest-known header every time a new one is discovered.

Because `ETHVerifiedTipMonitor` does a fair amount of work (and a proper implementation will do even more) there should only ever be one instance. Maybe it can be started early by the `FullServer` and broadcast `NewHead` events over lajha for any interested parties. This might be the wrong factoring, because the header syncer should probably have some control over this process, while header sync is running it'll collect information which could out the tip we're using as malicious.

Remaining work:
- `ETHVerifiedTipMonitor` doesn't verify the headers yet.
- `HexarySync` accepts a `new_root_hash` command but does nothing with it, so the state sync never actually rebases.
- Some tests. I've just been running this manually against a parity node I'm running.

#### Cute Animal Picture

turtle-lead

### DIFF
--- a/p2p/kademlia.py
+++ b/p2p/kademlia.py
@@ -30,7 +30,6 @@ from eth_keys import (
 
 from eth_hash.auto import keccak
 
-from trinity._utils.validation import validate_enode_uri
 
 k_b = 8  # 8 bits per hop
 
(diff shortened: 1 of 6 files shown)


lithp:
Closing since this has been subsumed by our plan to create a new state sync protocol

Issue #89, opened 2018-11-30
Write an EventBus-friendly PeerPool that supports the Beacon node


cburgdorf:
### What is wrong?

This issue is an attempt to revive #1198 with a concrete plan to spearhead this effort for the new beacon node that Trinity aims to support.

### Background

Python is known to perform poorly in in-process concurrency ([GIL](http://www.dabeaz.com/python/GIL.pdf)). To make it feasible to still write applications like an Ethereum node in Python, one has to do far more gymnastics compared to other languages that have a better concurrency story. In particular, that means, applications need to make use of multiple processes in order to achieve concurrency. Of course, this comes with it's own set of problems.

Over the recent months we've put a lot of effort into creating a multi processing foundation for Trinity that abstracts away the pain of working with lots of different processes. We've created a [generic EventBus](https://lahja.readthedocs.io) that enables seamless communication between different processes.

We've also created a plugin foundation that enables us to write plugins that run in their own isolated process but can talk to the rest of the application out of the box via the event bus ([Docs](https://py-evm.readthedocs.io/en/latest/guides/trinity/writing_plugins.html#plugins-that-spawn-their-own-new-isolated-process))

Using this mechanism, we've already managed to successfully move out a lot of things into their own processes (e.g. EthStats Reporting, JSON-RPC server are both written as isolated plugins)

### Current PeerPool shortcomings

Our current peer pool doesn't yet work well with this multi process vision. This is holding us back because it means that things such as syncing or the transaction pool aren't easy to move into isolated plugins (read: separate processes) because they rely on being in the same process as the event pool.

Ideally, if we redesign the PeerPool I guess it should have the following characteristics:

- run in its own isolated plugin
- expose events on the `EventBus` such as `PeerJoined`, `PeerLeft` etc.
- these events are shallow [DTOs](https://en.wikipedia.org/wiki/Data_transfer_object) and can be consumed by any interested party in any process
- should expose an API that can be consumed from any process (which is easy to build on top of the EventBus)

E.g every process could use an API such as the following to interact with the PeerPool.

```py
class PeerPoolAPI:

    def __init__(self, event_bus: Endpoint):
        self.event_bus = event_bus


    async def get_peer_count() -> int:
        response = await self.event_bus.request(PeerCountRequest())
        return response.peer_count

    async def other_peer_pool_thingy() -> Whatever:
        ...
```

[...]

PR #213, opened 2019-01-22
Goal: Make it possible to move PeerPool into an isolated plugin

cburgdorf:
### What was wrong?

I think we are more or less aligned in our vision that we need to push forward the multiprocessing architecture that we've started. Similar to how modern browsers work (or many other applications), Trinity needs to run many processes in parallel to be able to handle the workload it is forced to deal with. This is especially true in a language such as Python that does not handle threading efficiently.

Today, we already have the [JSON-RPC Server](https://github.com/ethereum/trinity/blob/fa370299106b6391e203e8945f44f09f2c9d8703/trinity/plugins/builtin/json_rpc/plugin.py#L43), the [Peer Discovery](https://github.com/ethereum/trinity/blob/fa370299106b6391e203e8945f44f09f2c9d8703/trinity/plugins/builtin/peer_discovery/plugin.py#L144) and the [ETH Stats Reporting](https://github.com/ethereum/trinity/blob/fa370299106b6391e203e8945f44f09f2c9d8703/trinity/plugins/builtin/ethstats/plugin.py#L31) running as isolated processes (by deriving from `BaseIsolatedPlugin` which handles the heavy lifting).

Often these processes need to talk to other parts of Trinity. E.g. the `PeerPool` needs to talk to the discovery, the ETH Stats Reporting needs to ask the `PeerPool` for the number of connected nodes etc.

We leverage a central `EventBus` to handle such inter-process communication.

However, most of the networking is still running rather tightly coupled in one hardcoded (read: not managed through a plugin) process known as the `networking` process.

To be able to move core stuff such as *syncing*, *tx pool* etc into its own isolated processes, we need to get rid of the fact that most interactions with the `PeerPool` still rely on the code being run in the same process as the `PeerPool`.

In other words, what I believe, the best way forward would look like, is to have the `PeerPool` itself run as an isolated process and have all interactions be made through the `EventBus` (related: [event-bus-all-the-things](https://github.com/ethereum/trinity/issues/79), [write-an-eventbus-friendly-peerpool](https://github.com/ethereum/trinity/issues/89))

This PR is an attempt to experiment with that idea.

### How was it fixed?

Not yet. Just getting started.

#### Cute Animal Picture

### DIFF
--- a/p2p/events.py
+++ b/p2p/events.py
@@ -37,22 +37,3 @@ class RandomBootnodeRequest(BaseRequestResponseEvent[PeerCandidatesResponse]):
     @staticmethod
     def expected_response_type() -> Type[PeerCandidatesResponse]:
         return PeerCandidatesResponse
-
-
-class PeerCountResponse(BaseEvent):
-
-    def __init__(self, peer_count: int) -> None:
-        self.peer_count = peer_count
-
-
-class PeerCountRequest(BaseRequestResponseEvent[PeerCountResponse]):
-
-    @staticmethod
-    def expected_response_type() -> Type[PeerCountResponse]:
-        return PeerCountResponse
-
-
-class ConnectToNodeCommand(BaseEvent):
-
-    def __init__(self, node: str) -> None:
-        self.node = node
--- a/trinity/protocol/bcc/peer.py
+++ b/trinity/protocol/bcc/peer.py
@@ -1,5 +1,6 @@
 from typing import (
     cast,
+    NamedTuple,
 )
 
 from eth_utils import (
@@ -18,6 +19,7 @@ from eth2.beacon.typing import (
 from p2p.peer import (
     BasePeer,
     BasePeerFactory,
+    IdentifiablePeer,
 )
 from p2p.peer_pool import (
     BasePeerPool,
@@ -39,6 +41,12 @@ from trinity.protocol.bcc.commands import (
 from trinity.protocol.bcc.context import (
     BeaconContext,
 )
+
+
+class BCCPeerDTO(NamedTuple):
+    head_slot: Slot
+    network_id: int
+    uri: str
 
 
 class BCCPeer(BasePeer):
@@ -94,6 +102,13 @@ class BCCPeer(BasePeer):
         if self._requests is None:
             self._requests = BCCExchangeHandler(self)
         return self._requests
+
+    def to_dto(self) -> IdentifiablePeer:
+        return BCCPeerDTO(
+            head_slot=self.head_slot,
+            network_id=self.network_id,
+            uri=self.remote.uri(),
+        )
 
 
 class BCCPeerFactory(BasePeerFactory):
--- a/docs/release_notes/trinity.rst
+++ b/docs/release_notes/trinity.rst
@@ -4,7 +4,8 @@ Trinity
 Unreleased (latest source)
 --------------------------
 
-None
+- `#291 <https://github.com/ethereum/trinity/pull/291>`_: Performance: Move the Request server into its own isolated plugin process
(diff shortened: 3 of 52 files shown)


cburgdorf:
@jannikluhn This is basically starting what @pipermerriam pointed out with his recent [comment on your PR](https://github.com/ethereum/trinity/pull/163/files#r247567007)

So, this will most likely be related to your current work with the beacon node. I wanted to put this PR out very early on for transparency. I also tried to give some background on the direction in the description.

If you feel like doing your own experimentation on the same direction (motivated by getting the beacon node syncing), please do not feel obligated to wait for this PR. In fact, I'd be happy if someone would explore similar territory to hedge against my own failure haha.

p2p/events.py
```diff
@@ -56,3 +57,17 @@ class ConnectToNodeCommand(BaseEvent):
 
     def __init__(self, node: str) -> None:
         self.node = node
+
+
+class ConnectedPeersResponse(BaseEvent):
+
+    # FIXME: it's going to become really hard to get rid of `Any` here :-/
```
pipermerriam: Can you expand on why?  Intuition says this would be done with `typing.TypeVar('TPeer', bound=p2p.peer.BasePeer)` but I assume there's a reason why that doesn't work.

trinity/protocol/bcc/peer.py
```diff
@@ -41,6 +43,12 @@
 )
 
 
+class BCCPeerDTO(NamedTuple):
+    head_slot: SlotNumber
+    network_id: int
+    uri: str
```
pipermerriam: maybe including a referencable unique identifier like the enode public key or the full enode address itself would make sense here?

p2p/events.py
cburgdorf: Yes, but only if we make the `BasePeerPool` generic over `TPeer` (`BasePeerPool[TPeer]`). Actually, it's probably not going to be **that** hard as the comment suggest. I'm being overly dramatic here haha.

trinity/protocol/bcc/peer.py
cburgdorf: Isn't `uri` (coming from `Node.uri`) that unique identifier already?

pipermerriam: quadruple-facepalm

cburgdorf:
I created a basic benchmark to see how communication via the EventBus compares to using native Python proxies: https://github.com/ethereum/lahja/pull/29

It looks like are currently looking at a 15x poorer performance compared to native Python proxies which makes me believe that we should do the following:

1. Use proxy based access for things that are frequently needed (e.g. performing actions on peers)
2. Only expose not-too-frequent events on the EventBus so far (e.g. maybe sth like `NewChainTip`)
3. Look into ways to make the EventBus more efficient (e.g. get rid of extra process hop, use something like gRPC underneath etc)

cburgdorf:
@pipermerriam The code is still ugly but this PR now contains a commit that moves the `ETHRequestServer` into its own isolated plugin. This already lays the ground work to do the same with the sync. I'm going to continue with that now and clean it all up once I have more confidence things will work out fine.

p2p/events.py
```diff
@@ -56,3 +63,24 @@ class ConnectToNodeCommand(BaseEvent):
...
+class ConnectedPeersResponse(BaseEvent):
+
+    def __init__(self, peers: Tuple[IdentifiablePeer, ...]) -> None:
+        self.peers = peers
+
+
+class ConnectedPeersRequest(BaseRequestResponseEvent[ConnectedPeersResponse]):
+
+    @staticmethod
+    def expected_response_type() -> Type[ConnectedPeersResponse]:
```
pipermerriam: not for this PR but updating this method to be `get_expected_response_type` might be more inline with our other naming conventions.

cburgdorf:
Replaced by #721

PR #214, opened 2019-01-22
Randao validation

jannikluhn:
This is a copy of https://github.com/ethereum/py-evm/pull/1686 by @prateekreddy that I only rebased to Trinity.

Note that merging it right now is not a good idea as there are still unaddressed comments and it soon needs to be updated anyways once https://github.com/ethereum/eth2.0-specs/pull/483 is merged.

### What was wrong?
Implement `validate_randao` as per Eth 2.0 spec https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#randao

#### Cute Animal Picture

### DIFF
--- a/eth2/_utils/numeric.py
+++ b/eth2/_utils/numeric.py
@@ -134,4 +134,4 @@ def bitwise_xor(a: Hash32, b: Hash32) -> Hash32:
     c = b''
     for i in range(32):
         c += bytes([a[i] ^ b[i]])
-    return c
+    return Hash32(c)
--- a/eth2/_utils/numeric.py
+++ b/eth2/_utils/numeric.py
@@ -1,0 +1,137 @@
+import functools
+import itertools
+from typing import (
+    Union,
+)
+
+from cytoolz import (
+    curry,
+)
+from eth_typing import (
+    Hash32,
+)
+
+from eth.constants import (
+    UINT_255_MAX,
+    UINT_256_MAX,
+    UINT_256_CEILING,
+)
+
+
+def int_to_bytes32(value: Union[int, bool]) -> Hash32:
+    if not isinstance(value, int) or isinstance(value, bool):
+        raise ValueError(
+            "Value must be an integer: Got: {0}".format(
+                type(value),
+            )
+        )
+    if value < 0:
+        raise ValueError(
+            "Value cannot be negative: Got: {0}".format(
+                value,
+            )
+        )
+    if value > UINT_256_MAX:
+        raise ValueError(
+            "Value exeeds maximum UINT256 size.  Got: {0}".format(
+                value,
+            )
+        )
+    value_bytes = value.to_bytes(32, 'big')
+    return Hash32(value_bytes)
+
+
+def ceilXX(value: int, ceiling: int) -> int:
+    remainder = value % ceiling
+    if remainder == 0:
+        return value
+    else:
+        return value + ceiling - remainder
+
+
+ceil32 = functools.partial(ceilXX, ceiling=32)
+ceil8 = functools.partial(ceilXX, ceiling=8)
+
+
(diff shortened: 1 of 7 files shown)


hwwhww:
https://github.com/ethereum/eth2.0-specs/pull/483 is merged. 🎉

eth2/_utils/numeric.py
```diff
@@ -0,0 +1,149 @@
...
+)
+
+from eth.constants import (
+    UINT_255_MAX,
+    UINT_256_MAX,
+    UINT_256_CEILING,
+)
+
+
+def int_to_bytes32(value: Union[int, bool]) -> Hash32:
```
pipermerriam: It appears that this should be `value: int` since the next line explicitely disallows bools.

eth2/_utils/numeric.py
```diff
@@ -0,0 +1,149 @@
...
+    return value % 2 == 0
+
+
+def is_odd(value: int) -> bool:
+    return value % 2 == 1
+
+
+def get_highest_bit_index(value: int) -> int:
+    value >>= 1
+    for bit_length in itertools.count():
```
pipermerriam: Can we set a reasonable upper bound on this loop?

eth2/_utils/numeric.py
```diff
@@ -0,0 +1,149 @@
...
+
+@curry
+def clamp(inclusive_lower_bound: int,
+          inclusive_upper_bound: int,
+          value: int) -> int:
+    """
+    Bound the given ``value`` between ``inclusive_lower_bound`` and
+    ``inclusive_upper_bound``.
+    """
+    if value <= inclusive_lower_bound:
```
pipermerriam: I don't know how performance critical this is, but maybe it's better to optimize the common case?  Assuming the common case is that the value **is** within range then maybe start this with:

```python
if inclusive_lower_bound <= value <= inclusive_upper_bound:
     return value
elif value <= ...:
    ...
elif value >= ...:
    ...
else:
    raise Exception("Unreachable")

File: eth2/_utils/numeric.py [hunk 4]
```diff
@@ -0,0 +1,149 @@
...
+    if value < 0:
+        raise ValueError(
+            "Value cannot be negative: Got: {0}".format(
+                value,
+            )
+        )
+
+    x = value
+    y = (x + 1) // 2
+    while y < x:
```
pipermerriam: I think we worked up a faster version of this using a combination of `decimal.Decimal` and logarithms.

eth2/_utils/numeric.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+    if len(b) != 32:
+        raise ValueError(
+            "Arguments must be of size 32: Got: {0}".format(
+                len(b)
+            )
+        )
+    c = b''
+    for i in range(32):
+        c += bytes([a[i] ^ b[i]])
```
pipermerriam: Same, we had a better version of this in some other PR.

eth2/beacon/_utils/hash.py
```diff
@@ -13,3 +13,13 @@ def hash_eth2(data: Union[bytes, bytearray]) -> Hash32:
...
+
+
+def repeat_hash_eth2(data: bytes, repeats: int) -> Hash32:
+    """
+    Return Hash onion of ``data`` with ``repeats`` layers
+    """
+    if repeats == 0:
+        return Hash32(data)
+    else:
+        return repeat_hash_eth2(bytes(hash_eth2(data)), repeats - 1)
```
pipermerriam: And a non-recursive version of this using `cytoolz.pipe` and `itertools.repeat(hash_eth, iterations)`

jannikluhn:
Replaced by #252

PR #215, opened 2019-01-22
Ensure PeerPool does not hang on ConnectToNodeCommand

cburgdorf:
### What was wrong?

PeerPool hangs on cancel, waiting for `ConnectToNodeCommand` to finish.

### How was it fixed?

Wrapped with `wait_iter(...)`.
There are probably other parts of the code base where the same fix applies but this one I came across while working on #213 so it is verified that this is indeed needed.

#### Cute Animal Picture

### DIFF
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -93,7 +93,7 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
         self.event_bus = event_bus
 
     async def accept_connect_commands(self) -> None:
-        async for command in self.event_bus.stream(ConnectToNodeCommand):
+        async for command in self.wait_iter(self.event_bus.stream(ConnectToNodeCommand)):
             self.logger.debug('Received request to connect to %s', command.node)
             self.run_task(self.connect_to_nodes(from_uris([command.node])))
 


p2p/peer_pool.py
```diff
@@ -93,7 +93,7 @@ def __init__(self,
         self.event_bus = event_bus
 
     async def accept_connect_commands(self) -> None:
-        async for command in self.event_bus.stream(ConnectToNodeCommand):
+        async for command in self.wait_iter(self.event_bus.stream(ConnectToNodeCommand)):
```
pipermerriam: If we see this continue to be an issue maybe we should include a timeout as well as enforcing a maximum number of allowed timeouts for a given duration of time to allow occasional failures if this is intermittent.

cburgdorf:
>I was under the impression from other comments that you thought wrapping these in `wait_iter` wasn't necessary.

Yes, I thought so initially but the interplay with our `BaseService` model makes this slightly more complex.

Reading this [thread](https://github.com/ethereum/trinity/pull/166#discussion_r247747992), is probably the best explanation for now (should really turn this into some doc section).

When a  `BaseService` dies because the `CancelToken` was triggered, it waits for all tasks to finish. It's only when `BaseService.cancel()` was triggered explicitly that it forces tasks to shutdown after some timeout.

I think it would be worth trying to change this model as @lithp suggested in that thread.

>It would be simpler to have cancellation be the normal flow: tasks are given `cancel()` and expected to catch it if they really need the extra time to clean up.

But this may open up a pandora's box so I'm willing to kick the can down the road and just wrap `.stream(...)` with `wait_iter(...)` for the time being.

PR #221, opened 2019-01-23
Tune EthStats interval and expose it to CLI

cburgdorf:
### What was wrong?

The EthStats reporting could use a bit of tuning to reduce pressure on the `networking` process and report peer count more reliable. 

### How was it fixed?

1. Report stats less often
2. Wait a little longer on peer count response
3. Expose interval via --ethstats-interval option

#### Cute Animal Picture

### DIFF
--- a/trinity/plugins/builtin/ethstats/plugin.py
+++ b/trinity/plugins/builtin/ethstats/plugin.py
@@ -31,6 +31,7 @@ DEFAULT_SERVERS_URLS = {
 class EthstatsPlugin(BaseIsolatedPlugin):
     server_url: str
     server_secret: str
+    stats_interval: int
     node_id: str
     node_contact: str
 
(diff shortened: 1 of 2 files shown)


trinity/plugins/builtin/ethstats/plugin.py
```diff
@@ -70,6 +71,11 @@ def configure_parser(self, arg_parser: ArgumentParser, subparser: _SubParsersAct
             help='Node contact information for stats server',
             default=os.environ.get('ETHSTATS_NODE_CONTACT', ''),
         )
+        ethstats_parser.add_argument(
+            '--ethstats-interval',
+            help='The interval at which data is reported back',
+            default=10,
```
pipermerriam: Would 15 seconds (expected block time) be more appropriate? or something like 13/14 to come in just under block time?  I recognize I'm bike shedding.  10 is fine too :smile_cat:

cburgdorf: The good thing about this PR is that you can now have it your own way with `--ethstats-interval 15` ;)

PR #224, opened 2019-01-23
Make docstrings consistent with most common style

davesque:
### What was wrong?

The style of some docstrings was inconsistent with the majority of docstrings in the codebase.

### How was it fixed?

Fixed them.

#### Cute Animal Picture

### DIFF
--- a/trinity/plugins/builtin/ethstats/ethstats_service.py
+++ b/trinity/plugins/builtin/ethstats/ethstats_service.py
@@ -109,7 +109,7 @@ class EthstatsService(BaseService):
             await self.sleep(self.stats_interval)
 
     def get_node_info(self) -> EthstatsData:
-        '''Getter for data that should be sent once, on start-up.'''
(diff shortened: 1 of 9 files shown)


trinity/plugins/builtin/ethstats/ethstats_service.py
```diff
@@ -123,7 +123,7 @@ def get_node_info(self) -> EthstatsData:
         }
 
     def get_node_block(self) -> EthstatsData:
-        '''Getter for data that should be sent on every new chain tip change.'''
+        """Getter for data that should be sent on every new chain tip change."""
```
pipermerriam: Not to give you busy work but I'd be :+1: to changing these to multi-line style assuming there isn't opposition to this being the convention.

davesque: I bet that docstring formatter tool can do that for us. How about we make that change in a separate PR?

PR #227, opened 2019-01-24
Ensure trinity attach tests also hit the database fix bug in process pool executor handling

cburgdorf:
### What was wrong?

The interactive `trinity attach` tests that we have do currently not cover any APIs that hit the database. A missed opportunity to provide some extra guarantees.

This is a spin off from #173 

However, adding these caused our CI to fail more often than not. During debugging it was discovered (shout out to @veox :tada:) that we unintentionally create tons of processes that sit idle. That is because in `__init__` of `BaseService` we called `get_asyncio_executor` which creates a `ProcessPoolExecutor` with `n` processes where `n` is the number of CPU cores - 1. But since we use the `BaseService` not only in the `networking` process but also in every plugin, this caused tons of extra processes to be created for every plugin that derives from `BaseIsolatedProcess` (which run their own process).

### How was it fixed?

For the first part:

Added a test that calls `w3.eth.getBlock('latest')` (against a freshly initialized db) and asserts everything is as expected.

For the second part:

1. Change the `BaseService` to not depend or create an `Executor` implicitly. Instead, change the `_run_in_executor` API to require an `Executor` to be passed in.

2. Handle lifecycle of `Executor` explicitly in boot / teardown of `Node`

3. Rename `get_asyncio_executor` to `ensure_global_asyncio_executor` to better reflect what this API does

#### Cute Animal Picture

### DIFF
--- a/docs/release_notes/trinity.rst
+++ b/docs/release_notes/trinity.rst
@@ -3,6 +3,7 @@ Trinity
 
 Unreleased (latest source)
 
+- `#227 <https://github.com/ethereum/trinity/pull/227>`_: Bugfix: Do not accidentially create many processes that sit idle
 - `#227 <https://github.com/ethereum/trinity/pull/227>`_: Tests: Cover APIs that also hit the database in `trinity attach` tests
 - `#155 <https://github.com/ethereum/trinity/pull/155>`_: Feature: Disable syncing entirely with `--sync-mode none`
 - `#155 <https://github.com/ethereum/trinity/pull/155>`_: Feature: Allow running `--sync-mode full` directly
--- a/p2p/_utils.py
+++ b/p2p/_utils.py
@@ -44,7 +44,7 @@ CPU_EMPTY_VALUES = {None, 0}
 _executor: Executor = None
 
 
-def get_asyncio_executor(cpu_count: int=None) -> Executor:
+def ensure_global_asyncio_executor(cpu_count: int=None) -> Executor:
     """
     Returns a global `ProcessPoolExecutor` instance.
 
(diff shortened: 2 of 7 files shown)


cburgdorf:
The `py36-wheel-cli` is consistently running out of memory. Debugging this locally does not show hints of a memory leak to me. All processes are cleaned up cleanly. I also tried a few things to free up more memory in general but had no luck.

Maybe we can pay for some extra RAM? https://circleci.com/docs/2.0/configuration-reference/#resource_class

cburgdorf:
But even if I limit the `py36-wheel-cli` test to a subset of tests, they fail consistently (not because of memory though) even though the same test works just fine as part of the integration test run. It also works fine locally. Meh.

veox:
Some debugging showed that in CI, the `py36-wheel-cli` reaches the limit of allowed processes / file descriptors - not memory, as shown previously. The OS-generated error messages are somewhat misleading.

Commit https://github.com/ethereum/trinity/pull/227/commits/13aec4ea8ff673763d9f153efe9d5fbcc9c634e5 works to mitigate by not creating executor pools for plugins (which don't use them anyway).

-----

Currently, CircleCI jobs run on 36-core machines without strict KVM-like VM isolation; a running `trinity` instance would create (36-1)*3 == 105 processes for executor pools alone.

The test that most often failed was the one with `trinity attach` against an existing `trinity` instance; in effect, the process number doubled that of the above.

This is not counting all the rest of the processes that are running, e.g. the test harness...

cburgdorf:
@veox Thanks again for your help discovering this! I changed the description of the PR to reflect this. I also put the PR to work on a server and it seems to be working fine so far :)

Issue #141, opened 2018-12-28
Implement `process_justification`


hwwhww:
### What is wrong?

- Implement per-epoch processing: justification logic as [spec](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#justification)
- See if we can set `justification_bitfield` as `bytes`.

### How can it be fixed
- Implement `total_balance`
    - In the spec: Let `total_balance(shard_committee) = sum([get_effective_balance(state, i) for i in shard_committee.committee])`...looks buggy.
    - It can be put in `eth/beacon/helpers.py` for now.
- Implement `process_justification(state: BeaconState, config: BeaconConfig) -> BeaconState` in `eth/beacon/state_machines/forks/serenity/epoch_transitions.py`
- Call `process_justification` in `eth.beacon.state_machines.forks.serenity.per_epoch_transition`.

hwwhww:
@ChihChengLiang 
Note that we also need to fix this: https://github.com/ethereum/eth2.0-specs/pull/523 in this PR.

PR #228, opened 2019-01-25
Implement process_crosslink

### LINKED ISSUE #147: Implement `process_crosslinks`
### What is wrong?

- Implement per-epoch processing: crosslinks logic as [spec](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#crosslinks)

### How can it be fixed
- Implement `process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState` in `eth/beacon/state_machines/forks/serenity/epoch_transitions.py`
- Call `process_crosslinks` in `eth.beacon.state_machines.forks.serenity.SerenityStateTransition.per_epoch_transition`.

NIC619:
### What was wrong?
Fix #147 

### How was it fixed?
**NOTE: Epoch transition rule is still subject to change in https://github.com/ethereum/eth2.0-specs/pull/492.**

- Implement helper functions
    - [x] `get_current_epoch_attestations`/`get_previous_epoch_attestations`
    - [x] `get_attesting_validator_indices`
    - [x] `get_winning_root`
    - [x] tests
- Implement `process_crosslinks`
    - [x] `process_crosslinks`
    - [x] test
- Fix
    - bitfield length check in `get_attestation_participants`

### DIFF
--- a/eth2/beacon/helpers.py
+++ b/eth2/beacon/helpers.py
@@ -18,10 +18,6 @@ from eth_typing import (
     Hash32,
 )
 
-from eth.constants import (
-    ZERO_HASH32,
-)
-
 from eth2._utils.bitfield import (
     get_bitfield_length,
     has_voted,
@@ -37,6 +33,7 @@ from eth2.beacon._utils.random import (
     shuffle,
     split,
 )
+from eth2.beacon.exceptions import NoWinningRootError
 from eth2.beacon.enums import (
     SignatureDomain,
 )
@@ -411,10 +408,7 @@ def get_attestation_participants(state: 'BeaconState',
     if len(aggregation_bitfield) != get_bitfield_length(committee_size):
         raise ValidationError(
             'Invalid bitfield length,'
-            "\texpected: {}, found: {}".format(
-                get_bitfield_length(committee_size),
-                len(aggregation_bitfield),
-            )
+            "\texpected: {get_bitfield_length(committee_size)}, found: {len(aggregation_bitfield)}"
         )
 
     # Find the participating attesters in the committee
@@ -481,6 +475,30 @@ def get_attesting_validator_indices(
             )
 
 
+def get_total_attesting_balance(
+        *,
+        state: 'BeaconState',
+        shard: ShardNumber,
+        shard_block_root: Hash32,
+        attestations: Sequence[PendingAttestationRecord],
+        epoch_length: int,
+        max_deposit: Ether,
+        target_committee_size: int,
+        shard_count: int) -> int:
+    return sum(
+        get_effective_balance(state.validator_balances, i, max_deposit)
+        for i in get_attesting_validator_indices(
+            state=state,
+            attestations=attestations,
+            shard=shard,
+            shard_block_root=shard_block_root,
+            epoch_length=epoch_length,
+            target_committee_size=target_committee_size,
+            shard_count=shard_count,
+        )
+    )
+
+
 def get_winning_root(
         *,
         state: 'BeaconState',
@@ -490,24 +508,23 @@ def get_winning_root(
         max_deposit: Ether,
         target_committee_size: int,
         shard_count: int) -> Hash32:
-    winning_root = ZERO_HASH32
+    winning_root = None
     winning_root_balance = 0
     visited_shard_block_root: List[Hash32] = []
     for a in attestations:
         if a.data.shard_block_root in visited_shard_block_root:
+            # Already get the balance of this block root
             continue
         else:
-            root_balance = sum(
-                get_effective_balance(state.validator_balances, i, max_deposit)
-                for i in get_attesting_validator_indices(
-                    state=state,
-                    attestations=attestations,
-                    shard=shard,
-                    shard_block_root=a.data.shard_block_root,
-                    epoch_length=epoch_length,
-                    target_committee_size=target_committee_size,
-                    shard_count=shard_count,
-                )
+            root_balance = get_total_attesting_balance(
+                state=state,
+                shard=shard,
+                shard_block_root=a.data.shard_block_root,
+                attestations=attestations,
+                epoch_length=epoch_length,
+                max_deposit=max_deposit,
+                target_committee_size=target_committee_size,
+                shard_count=shard_count,
             )
             if root_balance > winning_root_balance:
                 winning_root = a.data.shard_block_root
@@ -517,6 +534,8 @@ def get_winning_root(
                 winning_root = a.data.shard_block_root
 
             visited_shard_block_root.append(a.data.shard_block_root)
+    if winning_root is None:
+        raise NoWinningRootError
     return winning_root
 
 
--- a/eth2/beacon/helpers.py
+++ b/eth2/beacon/helpers.py
@@ -413,7 +413,7 @@ def get_attestation_participants(state: 'BeaconState',
             'Invalid bitfield length,'
             "\texpected: {}, found: {}".format(
                 get_bitfield_length(committee_size),
-                len(attestation_data),
+                len(aggregation_bitfield),
             )
         )
 
--- a/eth2/beacon/helpers.py
+++ b/eth2/beacon/helpers.py
@@ -416,6 +416,35 @@ def get_attestation_participants(state: 'BeaconState',
             yield validator_index
 
 
+@to_tuple
+def get_attesting_validator_indices(
+        state: 'BeaconState',
+        epoch_length: int,
+        attestations: Sequence[PendingAttestationRecord],
+        target_committee_size: int,
+        shard_count: int,
+        shard: ShardNumber,
+        shard_block_root: Hash32) -> Iterable[ValidatorIndex]:
+    """
+    Loop through ``attestations`` and check if ``shard``/``shard_block_root`` in the attestation
+    matches the given ``shard``/``shard_block_root``.
+    If the attestation matches, get the index of the participating validators.
+    Finally, return the union of the indexes.
+    """
+    indexes_tuple = ()
+    for a in attestations:
+        if a.data.shard == shard and a.data.shard_block_root == shard_block_root:
+            indexes_tuple += get_attestation_participants(
+                state,
+                a.data,
+                a.participation_bitfield,
+                epoch_length,
+                target_committee_size,
+                shard_count,
+            )
+    return set(indexes_tuple)
+
+
 #
 # Misc
 #
--- a/eth2/beacon/helpers.py
+++ b/eth2/beacon/helpers.py
@@ -11,6 +11,7 @@ import functools
 from eth_utils import (
     big_endian_to_int,
     to_tuple,
+    to_set,
     ValidationError,
 )
 from eth_typing import (
@@ -39,7 +40,6 @@ from eth2.beacon._utils.random import (
 from eth2.beacon.enums import (
     SignatureDomain,
 )
-from eth2.beacon.state_machines.configs import BeaconConfig
 from eth2.beacon.types.pending_attestation_records import (
     PendingAttestationRecord,
 )
@@ -423,25 +423,55 @@ def get_attestation_participants(state: 'BeaconState',
             yield validator_index
 
 
+#
+# Per-epoch processing helpers
+#
 @to_tuple
-def get_attesting_validator_indices(
+def get_current_epoch_attestations(
         state: 'BeaconState',
-        epoch_length: int,
+        epoch_length: int) -> Iterable[PendingAttestationRecord]:
+    """
+    Note that this function should only be called at epoch boundaries as
+    it's computed based on current slot number.
+    """
+    for attestation in state.latest_attestations:
+        if state.slot - epoch_length <= attestation.data.slot < state.slot:
+            yield attestation
+
+
+@to_tuple
+def get_previous_epoch_attestations(
+        state: 'BeaconState',
+        epoch_length: int) -> Iterable[PendingAttestationRecord]:
+    """
+    Note that this function should only be called at epoch boundaries as
+    it's computed based on current slot number.
+    """
+    for attestation in state.latest_attestations:
+        if state.slot - 2 * epoch_length <= attestation.data.slot < state.slot - epoch_length:
+            yield attestation
+
+
+@to_tuple
+@to_set
+def get_attesting_validator_indices(
+        *,
+        state: 'BeaconState',
         attestations: Sequence[PendingAttestationRecord],
-        target_committee_size: int,
-        shard_count: int,
         shard: ShardNumber,
-        shard_block_root: Hash32) -> Iterable[ValidatorIndex]:
+        shard_block_root: Hash32,
+        epoch_length: int,
+        target_committee_size: int,
+        shard_count: int) -> Iterable[ValidatorIndex]:
     """
     Loop through ``attestations`` and check if ``shard``/``shard_block_root`` in the attestation
     matches the given ``shard``/``shard_block_root``.
     If the attestation matches, get the index of the participating validators.
-    Finally, return the union of the indexes.
+    Finally, return the union of the indices.
     """
-    indexes_tuple = ()
     for a in attestations:
         if a.data.shard == shard and a.data.shard_block_root == shard_block_root:
-            indexes_tuple += get_attestation_participants(
+            yield from get_attestation_participants(
                 state,
                 a.data,
                 a.aggregation_bitfield,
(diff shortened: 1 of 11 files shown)


eth2/beacon/helpers.py
```diff
@@ -363,9 +373,94 @@ def get_attestation_participants(state: 'BeaconState',
...
+                shard_count,
+            )
+    return set(indexes_tuple)
+
+
 #
 # Misc
 #
+@to_tuple
+def get_current_epoch_attestations(
```
hwwhww: Instead of putting under `Misc`, could you reorg and put all "Per-epoch processing helpers" together?

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -0,0 +1,82 @@
...
+            config.SHARD_COUNT,
+        )
+        for crosslink_committee, shard in crosslink_committees_at_slot:
+            # Filter out attestations not attesting to this shard so we don't need
+            # to going over irrelevent attestations over and over again.
+            to_this_shard_attestations = [
+                a
+                for a in current_epoch_attestations
+                if a.data.shard == shard
+            ]
```
hwwhww: @djrtwo @NIC619 
does it make sense to `break` if `len(to_this_shard_attestations) == 0` (skipped slot)?

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -0,0 +1,82 @@
...
+                config.TARGET_COMMITTEE_SIZE,
+                config.SHARD_COUNT,
+                shard,
+                winning_root,
+            )
+            total_attesting_balance = sum(
+                [
+                    get_effective_balance(state.validator_balances, i, config.MAX_DEPOSIT)
+                    for i in attesting_validators_indices
+                ]
```
hwwhww: Per https://github.com/ethereum/trinity/issues/141#issuecomment-452057519, you can remove the inner `[...]`.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -0,0 +1,82 @@
...
+            )
+            total_balance = sum(
+                [
+                    get_effective_balance(state.validator_balances, i, config.MAX_DEPOSIT)
+                    for i in crosslink_committee
+                ]
+            )
```
hwwhww: ditto

eth2/beacon/helpers.py
```diff
@@ -363,9 +373,94 @@ def get_attestation_participants(state: 'BeaconState',
...
+                    for i in get_attesting_validator_indices(
+                        state,
+                        config.EPOCH_LENGTH,
+                        attestations,
+                        config.TARGET_COMMITTEE_SIZE,
+                        config.SHARD_COUNT,
+                        shard,
+                        a.data.shard_block_root)
+                ]
+            )
```
hwwhww: Per https://github.com/ethereum/trinity/issues/141#issuecomment-452057519, you can remove the inner `[...]`.

eth2/beacon/helpers.py
```diff
@@ -363,9 +373,94 @@ def get_attestation_participants(state: 'BeaconState',
...
+        epoch_length: int) -> Sequence[PendingAttestationRecord]:
+    return [
+        a for a in state.latest_attestations
+        if state.slot - 2 * epoch_length <= a.data.slot < state.slot - epoch_length
+    ]
+
+
+def get_winning_root(
+        state: 'BeaconState',
+        config: BeaconConfig,
```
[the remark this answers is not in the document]
NIC619: Gotcha. Sorry I wasn't aware it.

NIC619:
@hwwhww Rebased and updated with PR feedbacks.

eth2/beacon/helpers.py
```diff
@@ -375,6 +385,103 @@ def get_attestation_participants(state: 'BeaconState',
...
+        epoch_length: int,
+        max_deposit: Ether,
+        target_committee_size: int,
+        shard_count: int) -> Hash32:
+    winning_root = ZERO_HASH32
+    winning_root_balance = 0
+    visited_shard_block_root: List[Hash32] = []
+    for a in attestations:
+        if a.data.shard_block_root in visited_shard_block_root:
+            continue
```
hwwhww: Could you add some comment of why `continue`? I suppose it's because that all `attestations` will be iterated in `get_attesting_validator_indices`.

eth2/beacon/helpers.py
```diff
@@ -375,6 +385,103 @@ def get_attestation_participants(state: 'BeaconState',
...
+                for i in get_attesting_validator_indices(
+                    state=state,
+                    attestations=attestations,
+                    shard=shard,
+                    shard_block_root=a.data.shard_block_root,
+                    epoch_length=epoch_length,
+                    target_committee_size=target_committee_size,
+                    shard_count=shard_count,
+                )
+            )
```
hwwhww: Should this `sum(...)` code be a helper function `get_total_attesting_balance` as described in the spec?

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -1,21 +1,112 @@
...
+                shard=shard,
+                shard_block_root=winning_root,
+                epoch_length=config.EPOCH_LENGTH,
+                target_committee_size=config.TARGET_COMMITTEE_SIZE,
+                shard_count=config.SHARD_COUNT,
+            )
+            total_attesting_balance = sum(
+                get_effective_balance(state.validator_balances, i, config.MAX_DEPOSIT)
+                for i in attesting_validators_indices
+            )
```
hwwhww: `total_attesting_balance` was calculated in `get_winning_root`?

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -1,21 +1,112 @@
...
+            attesting_validators_indices = get_attesting_validator_indices(
+                state=state,
+                attestations=_get_attestations_by_shard(current_epoch_attestations, shard),
+                shard=shard,
+                shard_block_root=winning_root,
+                epoch_length=config.EPOCH_LENGTH,
+                target_committee_size=config.TARGET_COMMITTEE_SIZE,
+                shard_count=config.SHARD_COUNT,
+            )
```
hwwhww: `attesting_validators_indices` was calculated in `get_winning_root`?

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -1,21 +1,112 @@
...
+            config.SHARD_COUNT,
+        )
+        for crosslink_committee, shard in crosslink_committees_at_slot:
+            # Use `_get_attestations_by_shard` to filter out attestations not attesting to
+            # this shard so we don't need to going over irrelevent attestations over and over again.
+            winning_root = get_winning_root(
+                state=state,
+                shard=shard,
+                attestations=_get_attestations_by_shard(current_epoch_attestations, shard),
+                epoch_length=config.EPOCH_LENGTH,
```
pipermerriam: not for this PR but an idea I've had a few times.

I think it is *good* that each of these nice functional utilities takes only the set of config values that it needs.  This helps make it easy to see what config values each function depends on and keeps the implementations decoupled from the config object.

however, it's also really unfortunate how verbose some of this code ends up because you end up having to explicitly provide each parameter...

So I was thinking about an API that allowed you to take a `config` and auto-magically partial out all of these helper functions.  Here's some pseudo code to illustrate.

```python
class BeaconHelperAPI:
    def __init__(self, config):
        # option 1: do it dynamically in constructor (con: hard to maintain type safety)
        self.get_attesting_validator_indices = functools.partial(
            get_attesting_validator_indices,
            epoch_length=config.EPOCH_LENGTH ,
            ...
        )

    def get_attesting_validator_indices(state, attestations, shard, shard_block_root):
        # option 2: mirror the API (con: heavy on boilerplate, duplicates all function signatures)
        return get_attesting_validator_indices(..., epoch_length=self.config.EPOCH_LENGTH, ...)
```

I'm sure there are some other options for how it could be implemented as well.  Here's an overly clever one that could be really good if done well but also might be too clever and end up bad.

```python
class Wrapper:
    def __init__(self, fn):
        self.fn = fn
    def __call__(self, *args, **kwargs):
         return fn(*args, **kwargs)
    def partial(self, config):
         kwargs = {kw: getattr(config, kw.upper()) for kw in keywords_to_partial}
         return functools.partial(self.fn, **kwargs)

def configurable(*keywords_to_partial):
    def outer(fn):
        return Wrapper(fn)
    return outer

@configurable('epoch_length', 'target_committee_size', 'shard_count')
def get_attesting_validator_indices(stat, attestations, shard, shard_block_root, epoch_length, target_committee_size, shard_count):
    ...

# using it in code
get_attestating_validator_indices.partial(config)(state, attestations, shard, shard_block_root)
```

Anyways, some ideas on how you might make calling these helpers a little less cumbersome, but I'd advocate for ensuring that whatever solution you end up using, that you ensure that you maintain type safety. 

cc @jannikluhn @hwwhww @ralexstokes (and anyone else I'm forgetting).

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -1,21 +1,112 @@
...
+            total_attesting_balance = sum(
+                get_effective_balance(state.validator_balances, i, config.MAX_DEPOSIT)
+                for i in attesting_validators_indices
+            )
+            total_balance = sum(
+                get_effective_balance(state.validator_balances, i, config.MAX_DEPOSIT)
+                for i in crosslink_committee
+            )
+            if 3 * total_attesting_balance >= 2 * total_balance:
+                latest_crosslinks = update_tuple_item(
```
pipermerriam: Consider using a new variable name like `updated_latest_crosslinks` (and you can just do `updated_latest_crosslinks = latest_crosslinks` in an `else` clause) as to not overwrite the local variable.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -1,25 +1,111 @@
...
+                            slot=state.slot,
+                            shard_block_root=winning_root,
+                        )
+                    )
+                else:
+                    # Don't update the crosslink of this shard
+                    pass
+            except NoWinningRootError:
+                # No winning shard block root found for this shard.
+                pass
```
hwwhww: Could narrow down the `try..` scope by refactoring to `try... except... else...`:
```python
            try:
                winning_root, total_attesting_balance = get_winning_root(....)
            except NoWinningRootError:
                # No winning shard block root found for this shard.
                pass
            else:
                total_balance = sum(...)
                if 3 * total_attesting_balance >= 2 * total_balance:
                    ...
                else:
                    # Don't update the crosslink of this shard
                    pass
```

eth2/beacon/helpers.py
```diff
@@ -405,10 +412,7 @@ def get_attestation_participants(state: 'BeaconState',
     if len(aggregation_bitfield) != get_bitfield_length(committee_size):
         raise ValidationError(
             'Invalid bitfield length,'
-            "\texpected: {}, found: {}".format(
-                get_bitfield_length(committee_size),
-                len(attestation_data),
-            )
+            "\texpected: {get_bitfield_length(committee_size)}, found: {len(aggregation_bitfield)}"
```
hwwhww: Missing `f""`.

eth2/beacon/helpers.py
```diff
@@ -375,6 +385,103 @@ def get_attestation_participants(state: 'BeaconState',
...
+@to_tuple
+def get_current_epoch_attestations(
+        state: 'BeaconState',
+        epoch_length: int) -> Iterable[PendingAttestationRecord]:
+    """
+    Note that this function should only be called at epoch boundaries as
+    it's computed based on current slot number.
+    """
+    for attestation in state.latest_attestations:
+        if state.slot - epoch_length <= attestation.data.slot < state.slot:
```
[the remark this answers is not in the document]
hwwhww: After offline discuss, leave it after #235.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
jannikluhn: :+1: Something like this would be really cool. However, I'm not sure if type safety can be kept, because we have to dynamically extract arguments from the config tuple (and, according to a quick google search, mypy doesn't seem to understand decorators if they change the function signature.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -1,25 +1,112 @@
...
+        if attestation.data.shard == shard:
+            yield attestation
+
+
+def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
+    """
+    Implement 'per-epoch-processing.crosslinks' portion of Phase 0 spec:
+    https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#crosslinks
+
+    For the shard to be crosslinked during each slots of the past two epochs, find the shard block
```
djrtwo:
```suggestion
    For each shard from the past two epochs, find the shard block
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -1,25 +1,112 @@
...
+    root with the most stake attestted to.
```
djrtwo:
```suggestion
    root that has been attested to by the most stake.
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -1,25 +1,112 @@
...
+    If enough(>= 2/3 total stake) attesting stake, update the crosslink record of that shard.
+    Return resulting ``state``
+    """
+    latest_crosslinks = state.latest_crosslinks
+    current_epoch_attestations = get_current_epoch_attestations(state, config.EPOCH_LENGTH)
+    # TODO: STUB, in spec it was
+    # `for slot in range(state.slot - 2 * config.EPOCH_LENGTH, state.slot):``
```
djrtwo: `- 2 *config.EPOCH_LENGTH` and using all of the `state.attestations` is more correct. Note that in the new spec changes, we are doing a range from `(prev_epoch_start_slot, next_epoch_start_slot)`.

djrtwo: mainly, just ensure that you use `state.attestations` when you fix this TODO

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -1,25 +1,112 @@
...
+                    for i in crosslink_committee
+                )
+                if 3 * total_attesting_balance >= 2 * total_balance:
+                    latest_crosslinks = update_tuple_item(
+                        latest_crosslinks,
+                        shard,
+                        CrosslinkRecord(
+                            slot=state.slot,
+                            shard_block_root=winning_root,
+                        )
```
djrtwo:
```suggestion
                        ),
```

tests/eth2/beacon/conftest.py
```diff
@@ -358,6 +358,21 @@ def ten_validators_state(filled_beacon_state, max_deposit_amount):
     )
 
 
+@pytest.fixture()
+def hundred_validators_state(filled_beacon_state, max_deposit_amount):
```
djrtwo: Should we add `n_validators_state(filled_beacon_state, max_deposit_amount, n)` that both of these helpers use?

tests/eth2/beacon/state_machines/forks/test_serenity_epoch_processing.py
```diff
@@ -311,3 +325,104 @@ def test_process_final_updates(genesis_state,
...
+
+
+@settings(max_examples=1)
+@given(random=st.randoms())
+@pytest.mark.parametrize(
+    (
+        'epoch_length,'
+        'target_committee_size,'
+        'shard_count,'
+        'succes_crosslink_in_cur_epoch,'
```
djrtwo: typo -- should be `success_crosslink_in_cur_epoch`

tests/eth2/beacon/state_machines/forks/test_serenity_epoch_processing.py
```diff
@@ -311,3 +325,104 @@ def test_process_final_updates(genesis_state,
...
+        for _ in range(shard_count)
+    ])
+    state = hundred_validators_state.copy(
+        slot=current_slot,
+        latest_crosslinks=initial_crosslinks,
+    )
+
+    # Generate current epoch attestations
+    cur_epoch_attestations = []
+    for slot_in_cur_epoch in range(state.slot - config.EPOCH_LENGTH + 1, state.slot + 1):
```
djrtwo: why `+ 1`?
this will be on range -- (65, 129). this misses 64 and adds in 128. 128 is the start of the next epoch.

djrtwo: This loop only adds a single attestation for `shard`. Is this the intention?

NIC619: > This loop only adds a single attestation for shard. Is this the intention?

Yes, I pack the votes into single attestation but they can be divided and separated into different attestations if desired.

NIC619:
@djrtwo I'm going ahead to merge this. Please let me know if there's anything still needs to be corrected and I will fix them in another PR.

djrtwo:
looks good!

PR #231, opened 2019-01-25
tests/integration: address some intermittent failures of `py3*-wheel-cli` CI jobs

veox:
### What was wrong?

Many PRs failing `py3*-wheel-cli` jobs, e.g.: https://github.com/ethereum/trinity/pull/173#issuecomment-457366460, for two reasons:

* In the failing test, `w3.net.version` in the console of `trinity attach` is (sometimes) sent before the console is done attaching; the test then waits forever to read an answer for a question no one heard.
* For some reason, some process gets a `kill -9` from the OS (likely the initial one, with DB connection and IPC socket and whatnot); the test then waits forever for an answer from a process that's no longer there. Described in more detail in PR https://github.com/ethereum/trinity/pull/227#issuecomment-457908353.

### How was it fixed?

- [x] Wait for `In [1]:` of IPython console output. It seems this is enough to stop `py37-wheel-cli` from failing.
- [x] Test for `trinity attach` starts background `trinity` with no discovery and 0 max peers, since it's not strictly required in this test. That should limit resource consumption, and prevent unexpected failures due to networking exceptions bubbling up to the chain syncer, which now (since PR ???) has the power to shut down the node on unexpected exception.

The following was debugged, but left for PR https://github.com/ethereum/trinity/pull/227#issuecomment-457908353 to address more gracefully:

- [ ] Lower the number of CPUs used so CI machine's file handle / process number limits are not hit, and the OS doesn't intervene by `kill -9`.

### Cute Animal Picture

Waiting for tests to go green.

Via [Pinterest](https://www.pinterest.com/pin/325948091754940630/) (source unknown)

### DIFF
--- a/tests/integration/test_trinity_cli.py
+++ b/tests/integration/test_trinity_cli.py
@@ -129,6 +129,7 @@ async def test_web3(command, async_process_runner):
     attached_trinity = pexpect.spawn('trinity', ['attach'], logfile=sys.stdout)
     try:
         attached_trinity.expect("An instance of Web3 connected to the running chain")
+        attached_trinity.expect(r"In \[.*\]:")  # ipython console ready to accept commands
         attached_trinity.sendline("w3.net.version")
         attached_trinity.expect("'1'")
         attached_trinity.sendline("w3")
--- a/tests/integration/test_trinity_cli.py
+++ b/tests/integration/test_trinity_cli.py
@@ -114,11 +114,12 @@ async def test_light_boot(async_process_runner, command):
 @pytest.mark.parametrize(
     'command',
     (
-        ('trinity', ),
+        # no network sync -> less resources used -> no kill in CI run
+        ('trinity', '--disable-discovery', '--max-peers=0',),
     )
 )
 @pytest.mark.asyncio
-async def test_web3(command, async_process_runner):
+async def test_web3(async_process_runner, command):
     await async_process_runner.run(command, timeout_sec=30)
     assert await contains_all(async_process_runner.stderr, {
         "Started DB server process",
--- a/tox.ini
+++ b/tox.ini
@@ -68,14 +68,12 @@ deps = {[common-wheel-cli]deps}
 whitelist_externals = {[common-wheel-cli]whitelist_externals}
 commands = {[common-wheel-cli]commands}
 skip_install=true
-use_develop=false
 
 [testenv:py37-wheel-cli]
 deps = {[common-wheel-cli]deps}
 whitelist_externals = {[common-wheel-cli]whitelist_externals}
 commands = {[common-wheel-cli]commands}
 skip_install=true
-use_develop=false
 
 [common-integration]
 deps = .[p2p,trinity,test]


veox:
Looks like same failure, plus some more. :/

veox:
[`py37-wheel-cli`](https://circleci.com/gh/veox/trinity/831?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) now shows:

``` python
=============================== warnings summary ===============================
tests/integration/test_trinity_cli.py:132
  /home/circleci/repo/tests/integration/test_trinity_cli.py:132: DeprecationWarning: invalid escape sequence \[
    attached_trinity.expect("In \[1\]:")  # ipython console ready to accept commands
```

veox:
Only [`py36-wheel-cli` is failing](https://circleci.com/gh/veox/trinity/847?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) now, with out-of-memory kill, as @cburgdorf established in https://github.com/ethereum/trinity/pull/173#issuecomment-457580007. :/

veox:
@pipermerriam Does the approval mean that, after all, expecting IPython to be present is correct?..

If so, I'll clean up the PR (instead of closing) once it's able to build the `py36` wheel. It still seems a 50/50 proposition so far.

pipermerriam:
@veox with respect to the regex, what do you think about running these tests with whatever the equivalent of `--colors=NoColor`.  Here's the docs on doing it with code.

https://ipython.readthedocs.io/en/stable/config/details.html?highlight=environment#terminal-colors

And some more on general configuration patterns: https://ipython.readthedocs.io/en/stable/config/intro.html#the-config-magic

Current solutino is almost definitely fine, I just worry about it overmatching which is probably not a big worry.

pipermerriam:
ipython is part of the trinity dependencies.  We decided that the vanilla shell is bad UX and that bundling ipython is acceptable.  Does that answer your question?

veox:
> Does that answer your question?

Yes, in general; it also turns out that not using IPython doesn't cure OOM, so the point is moot.

As to overmatching, I'll try passing the regex in a raw string (as I should've done in the first commit).

veox:
Updated the preamble to more closely reflect the findings: there are two cases that `py36-wheel-cli` tends to fail. So far, this PR only deals with one.

tox.ini
```diff
@@ -60,22 +58,21 @@ whitelist_externals=
 commands=
     /bin/rm -rf build dist
     python setup.py sdist bdist_wheel
-    /bin/bash -c 'pip install --upgrade "$(ls dist/*.whl)""[p2p,trinity]"'
+    # install the rest of dependencies needed for testing, as they're specified in the wheel
+    /bin/bash -c 'pip install --upgrade "$(ls dist/*.whl)""[p2p,trinity,test]"'
```
cburgdorf: @veox the one downside of adding `test` to the `pip install` here is that if we would not catch the case where Trinity might accidentally depend on `test` (which it obviously shouldn't for the regular user). But if this is important to cure the CI fails I'm ok with it. Just wanted to point out the advantage of the previous setup in this regard.

veox: I agree; if this doesn't test for what it was meant to, then it is useless.

It is not important to cure CI fails. It is important to make no expected mistakes.

(Where is the downward rocket emojicon?  I blame github. I blame microsoft.)

tests/integration/test_trinity_cli.py
```diff
@@ -114,11 +114,12 @@ def async_process_runner(event_loop):
 @pytest.mark.parametrize(
     'command',
     (
-        ('trinity', ),
+        # no network sync -> less resources used -> no kill in CI run
```
veox: The "no kill in CI run" part is incorrect.

cburgdorf: @veox do we want to keep this (it should not be needed anymore for CI to work but it still does save resources in general). If we want to keep it, maybe we just change the comment and remote the `no kill in CI run` part?

tests/integration/test_trinity_cli.py
```diff
@@ -129,6 +130,7 @@ def async_process_runner(event_loop):
     attached_trinity = pexpect.spawn('trinity', ['attach'], logfile=sys.stdout)
     try:
         attached_trinity.expect("An instance of Web3 connected to the running chain")
+        attached_trinity.expect(r"In \[.*\]:")  # ipython console ready to accept commands
```
cburgdorf: The one bad thing about this is that it makes the test dependent on `ipython` being present. I wonder how related this failure was with the fact that the host was just overloaded with processes? In other words: do we still need this? I haven't seen any fail anymore since we got rid of the faulty processes.

veox: Currently, `ipython` is installed as a dependency, including in test runs.

The [particular job](https://circleci.com/gh/ethereum/trinity/7918?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) that made me add this actually failed due to an OS-initiated kill, so it's hard to say how often the send-before-console-ready happens, and whether the input isn't in fact buffered and fed to the console anyway. :/

tests/integration/test_trinity_cli.py
veox: It's quite likely no longer required; if we do something like this, it should probably be applied to all cases not requiring networking, not just this one.

veox:
Actually, it seems this PR no longer contains fixes for something that happens regularly (even semi-regularly...).

Closing.

PR #236, opened 2019-01-29
Faster task pruning

### LINKED ISSUE #200: Profiling: a third of CPU time is spent looking up sync task dependencies
... and another 9% constructing sync task objects?..

-----

* py-evm Version: `master` and a WIP branch, see below
* OS: Arch Linux
* Python Version (python --version): 3.7.2
* Environment (output of `pip freeze`): [gist](https://gist.github.com/veox/d46ed02d3973cb9b0489b697cbeb9515)

### What is wrong?

Currently syncing main-net, around block `#4_000_000`: often blocked due to single-CPU execution.

The following (selective) log illustrates it well:

```
    INFO  01-20 03:18:15   FastChainBodySyncer  blks=0     txs=0      bps=1    tps=100   elapsed=7.2  head=#3931833 15f4..f26f  age=1y6m3w                                                                                                    
    INFO  01-20 03:18:30   FastChainBodySyncer  blks=0     txs=0      bps=1    tps=95    elapsed=14.8  head=#3931833 15f4..f26f  age=1y6m3w                                                                                                   
    INFO  01-20 03:18:35   FastChainBodySyncer  blks=0     txs=0      bps=1    tps=91    elapsed=5.0  head=#3931833 15f4..f26f  age=1y6m3w                                                                                                    
    INFO  01-20 03:18:59   FastChainBodySyncer  blks=0     txs=0      bps=1    tps=86    elapsed=23.9  head=#3931833 15f4..f26f  age=1y6m3w                                                                                                   
    INFO  01-20 03:19:18   FastChainBodySyncer  blks=0     txs=0      bps=1    tps=82    elapsed=18.7  head=#3931833 15f4..f26f  age=1y6m3w                                                                                                   
    INFO  01-20 03:19:29   FastChainBodySyncer  blks=11    txs=696    bps=1    tps=81    elapsed=11.7  head=#3931844 526b..2b18  age=1y6m3w                                                                                                   
    INFO  01-20 03:19:34   FastChainBodySyncer  blks=299   txs=18832  bps=4    tps=265   elapsed=5.0  head=#3932143 7cc1..8aa5  age=1y6m3w                                                                                                    
    INFO  01-20 03:19:39   FastChainBodySyncer  blks=240   txs=15116  bps=6    tps=402   elapsed=5.0  head=#3932383 d0ac..ff77  age=1y6m3w                                                                                                    
    INFO  01-20 03:20:20   FastChainBodySyncer  blks=28    txs=1829   bps=6    tps=385   elapsed=40.6  head=#3932411 534c..09b4  age=1y6m3w                                                                                                   
    INFO  01-20 03:20:25   FastChainBodySyncer  blks=21    txs=1387   bps=6    tps=379   elapsed=5.0  head=#3932432 92ba..2488  age=1y6m3w                                                                                                    
    INFO  01-20 03:20:30   FastChainBodySyncer  blks=3     txs=111    bps=5    tps=361   elapsed=5.1  head=#3932435 35e5..1ea4  age=1y6m3w                                                                                                    
    INFO  01-20 03:20:35   FastChainBodySyncer  blks=376   txs=20108  bps=9    tps=544   elapsed=5.0  head=#3932811 34b0..e4c5  age=1y6m3w                                                                                                    
    INFO  01-20 03:20:56   FastChainBodySyncer  blks=12    txs=909    bps=8    tps=519   elapsed=21.0  head=#3932823 03e1..8183  age=1y6m3w                                                                                                   
    INFO  01-20 03:21:05   FastChainBodySyncer  blks=125   txs=7801   bps=9    tps=538   elapsed=8.7  head=#3932948 f6c5..33e5  age=1y6m3w                                                                                                    
    INFO  01-20 03:21:16   FastChainBodySyncer  blks=0     txs=0      bps=8    tps=511   elapsed=10.8  head=#3932948 f6c5..33e5  age=1y6m3w                                                                                                   
    INFO  01-20 03:22:33   FastChainBodySyncer  blks=3     txs=250    bps=8    tps=485   elapsed=77.2  head=#3932951 14b0..221e  age=1y6m3w                                                                                                   
    INFO  01-20 03:22:38   FastChainBodySyncer  blks=6     txs=462    bps=7    tps=466   elapsed=5.1  head=#3932957 25ef..e801  age=1y6m3w                                                                                                    
    INFO  01-20 03:22:43   FastChainBodySyncer  blks=343   txs=17344  bps=10   tps=616   elapsed=5.0  head=#3933300 bd78..c471  age=1y6m3w                                                                                                    
    INFO  01-20 03:22:48   FastChainBodySyncer  blks=114   txs=5943   bps=11   tps=644   elapsed=5.0  head=#3933414 f7e2..b44f  age=1y6m3w                                                                                                    
    INFO  01-20 03:22:55   FastChainBodySyncer  blks=0     txs=0      bps=10   tps=612   elapsed=7.6  head=#3933414 f7e2..b44f  age=1y6m3w                                                                                                    
    INFO  01-20 03:23:02   FastChainBodySyncer  blks=0     txs=0      bps=10   tps=581   elapsed=6.2  head=#3933414 f7e2..b44f  age=1y6m3w                                                                                                    
    INFO  01-20 03:23:11   FastChainBodySyncer  blks=0     txs=0      bps=9    tps=552   elapsed=9.0  head=#3933414 f7e2..b44f  age=1y6m3w                                                                                                    
    INFO  01-20 03:23:18   FastChainBodySyncer  blks=0     txs=0      bps=9    tps=525   elapsed=6.9  head=#3933414 f7e2..b44f  age=1y6m3w                                                                                                    
    INFO  01-20 03:23:23   FastChainBodySyncer  blks=0     txs=0      bps=8    tps=498   elapsed=5.1  head=#3933414 f7e2..b44f  age=1y6m3w                                                                                                    
    INFO  01-20 03:23:39   FastChainBodySyncer  blks=0     txs=0      bps=8    tps=473   elapsed=15.9  head=#3933414 f7e2..b44f  age=1y6m3w
```

Guess: data starts coming in, but is soon shut off due to block/transaction processing not being able to keep up.

Therefore evaluating inefficiencies and multi-CPU options.

Stats (moved from https://github.com/ethereum/trinity/issues/120#issuecomment-455797061):

On my [`sync-hodge-podge`](https://github.com/veox/trinity/commits/sync-hodge-podge) branch, sync-task pruning takes up 31.6%. :( Adding some debug statements shows most tasks are pruned at depth 1 (no big surprise, I guess).

On `master` (at commit 8cedd0529020270cbe7eba1f67b62a7655562eca), it's 33.9%, so about the same as above (but less stable).

**Profiling for the above:** [in gist](https://gist.github.com/veox/ff3b7114a2869b2d9067ac2aed623755) (<10 minutes run each).

`sync-hodge-podge` is a WIP thing with a lot of garbage and some sync-fixing PRs; of which only PR #165 is still unmerged, and seems merge-worthy to me.

`s-h-p` is able to keep limping on for hours, and demonstrates the above log pattern coming and going in waves. CPU core load remains at constant >=99.9% on syncing thread - therefore my conclusion about saturation above.

### How can it be fixed

Review sync-tasks, in particular:

https://github.com/ethereum/trinity/blob/8cedd0529020270cbe7eba1f67b62a7655562eca/trinity/_utils/datastructures.py#L650-L658

Previous profiling (many months ago) had `sedes`-related processing in first place (now also major, but second). Have not reviewed `sedes`-providing package for inefficiencies (e.g. object-based processing vs. class-based).

`py-evm` has PR https://github.com/ethereum/py-evm/pull/1497 to move some syncing into a plugin (and therefore separate thread). I've not looked at it yet.

carver:
### What was wrong?

Fixes #200 (According to local simplified profiling)

It was quite slow to prune the tasks from `OrderedTaskPreparation`. According to the included test, it took ~5ms to prune one node in a simple 10k node line.

### How was it fixed?

Track references to the source points in the list, adjusted as the tree grows and is pruned. In the best case, a single line of 10k nodes, that means a single lookup instead of 10k lookups. In the included test, that means dropping from 5ms to ~0.05ms to prune.

TODO
- [x] add release notes entry
- [x] rebase/squash commits
- [x] linting & type hints
- [x] docstring comments on new classes & methods
- [x] remove assertions, after confirming that they are all be covered by tests now
- [x] make the veox bug non-fatal (unverified)
- [x] resolve the bug that veox reported
- [x] only prune node if it is `ready`

Moving to other issues/PRs:
- [ ] prune dangling uncles (ready or not!)
- [ ] handle case where a bunch of pruned tasks already in the database are re-queued in the fast syncer
- [ ] Fairly common PoW failures, appears to be that the local validation is failing, not that the received header is wrong
- [ ] There is no back-pressure when importing blocks is the bottleneck: we will race ahead getting more and more data from peers, sitting in memory waiting to be imported (in my sync, head was inching along around block 700k, but the in-memory data was racing ahead to 1.27M, using ~10G of RAM)

#### Cute Animal Picture

### DIFF
--- a/tests/core/utils/test_root_tracker.py
+++ b/tests/core/utils/test_root_tracker.py
@@ -1,0 +1,177 @@
+import pytest
+
+from eth_utils import ValidationError
+from hypothesis import (
+    given,
+    settings,
+    strategies as st,
+)
+
+from trinity._utils.tree_root import Tree, RootTracker
+
+
+@given(st.permutations(range(10)))
+def test_tree_linking(add_order):
+    # node and parent id
+    nodes = (
+        ('A0', '_'),
+        ('A1', '_'),
+        ('B0', 'A0'),
+        ('C0', 'B0'),
+        ('C1', 'B0'),
+        ('C2', 'B0'),
+        ('D0', 'C0'),
+        ('D1', 'C0'),
+        ('D2', 'C0'),
+        ('E0', 'D0'),
+    )
+
+    tree = Tree()
+    for node_idx in add_order:
+        node, parent = nodes[node_idx]
+        tree.add(node, parent)
+
+    assert not tree.has_parent('A0')
+    assert not tree.has_parent('A1')
+    assert not tree.has_node('_')
+    assert tree.has_node('A0')
+    assert tree.has_node('A1')
+    assert tree.children_of('A0') == ('B0',)
+    assert tree.children_of('A1') == tuple()
+
+    for node, parent in nodes[2:]:
+        assert tree.has_node(node)
+
+        assert tree.has_parent(node)
+        assert tree.parent_of(node) == parent
+
+        children = tree.children_of(parent)
+        assert node in children
+
+        if parent in ('B0', 'C0'):
+            assert len(children) == 3
+        elif parent in ('A0', 'D0'):
+            assert len(children) == 1
+        else:
+            assert len(children) == 0
+
+
+@given(st.permutations(range(10)))
+def test_out_of_order_line(insertion_order):
+    tracker = RootTracker()
+    for node in insertion_order:
+        tracker.add(node, node - 1)
+
+    for expected_root in range(9):
+        for node in range(expected_root, 10):
+            root, depth = tracker.get_root(node)
+            assert depth == node - expected_root
+            assert root == expected_root
+
+        # always prune from the end
+        prune_root_id, _ = tracker.get_root(9)
+        tracker.prune(prune_root_id)
+
+        # root should not be retrievable anymore
+        with pytest.raises(ValidationError):
+            tracker.get_root(prune_root_id)
+
+
+@given(st.lists(st.integers(min_value=0, max_value=9)))
+def test_prune_reinsert_root_tracking(element_flipping):
+    tracker = RootTracker()
+
+    present = set()
+    for node in element_flipping:
+        if node in present:
+            prune_root_id, _ = tracker.get_root(node)
+            tracker.prune(prune_root_id)
+            present.remove(prune_root_id)
+        else:
+            tracker.add(node, node - 1)
+            present.add(node)
+
+        # validate all the present nodes have valid roots
+        for test_node in present:
+            root_id, depth = tracker.get_root(test_node)
+
+            # make sure parent is *not* present
+            assert root_id - 1 not in present
+
+            # make sure depth is correct
+            assert depth == test_node - root_id
+
+
+FULL_BINARY_TREE = [(layer, column) for layer in [0, 1, 2, 3] for column in range(2**layer)]
+
+
+@given(st.permutations(FULL_BINARY_TREE))
+def test_full_branching(insertion_order):
+    """Test full binary tree, in random order"""
+    def parent(node):
+        return (node[0] - 1, node[1] // 2)
+
+    tracker = RootTracker()
+    for node in insertion_order:
+        tracker.add(node, parent(node))
+
+    # prune all the way to the leaf of (3, 0)
+    for num_prunings in range(3):
+        root_id, depth = tracker.get_root((3, 0))
+        assert root_id[0] == num_prunings
+        assert depth == 3 - num_prunings
+        tracker.prune(root_id)
+        assert tracker.get_root((3, 7)) == ((1, 1), 2)
+
+
+@st.composite
+def subset_and_order(draw):
+    nodes_to_insert = draw(st.lists(st.sampled_from(FULL_BINARY_TREE), unique=True))
+    prune_order = draw(st.permutations(range(len(nodes_to_insert))))
+    return (nodes_to_insert, prune_order)
+
+
+@given(subset_and_order())
+@settings(max_examples=500)
+def test_sparse_branching(test_data):
+    nodes_to_insert, prune_order = test_data
+
+    def parent(node):
+        return (node[0] - 1, node[1] // 2)
+
+    def get_expected_root(node, present_nodes):
+        expected_depth = 0
+        expected_root = node
+        parent_node = parent(node)
+        while parent_node in present_nodes:
+            expected_depth += 1
+            expected_root = parent_node
+            parent_node = parent(parent_node)
+        return expected_root, expected_depth
+
+    tracker = RootTracker()
+    for node in nodes_to_insert:
+        tracker.add(node, parent(node))
+
+    # verify parent and depth of partially-built tree
+    for node in nodes_to_insert:
+        actual_root, actual_depth = tracker.get_root(node)
+        expected_root, expected_depth = get_expected_root(node, nodes_to_insert)
+        assert actual_root == expected_root
+        assert actual_depth == expected_depth
+
+    # prune
+    remaining_nodes = set(nodes_to_insert)
+    for prune_idx in [idx for idx in prune_order if idx < len(nodes_to_insert)]:
+        node_to_prune_from = nodes_to_insert[prune_idx]
+        if node_to_prune_from not in remaining_nodes:
+            continue
+        prune_root_id, _ = tracker.get_root(node_to_prune_from)
+        tracker.prune(prune_root_id)
+        remaining_nodes.remove(prune_root_id)
+
+        for node in remaining_nodes:
+            actual_root, actual_depth = tracker.get_root(node)
+            expected_root, expected_depth = get_expected_root(node, remaining_nodes)
+            assert actual_root == expected_root
+            assert actual_depth == expected_depth
--- a/trinity/_utils/datastructures.py
+++ b/trinity/_utils/datastructures.py
@@ -12,14 +12,11 @@ from functools import (
 )
 from itertools import (
     count,
-    repeat,
 )
-from operator import attrgetter
 from typing import (
     Any,
     Callable,
     Dict,
-    Generator,
     Generic,
     Iterable,
     Set,
@@ -33,15 +30,7 @@ from eth_utils import (
     to_tuple,
 )
 from eth_utils.toolz import (
-    compose,
-    concat,
-    curry,
-    do,
     identity,
-    iterate,
-    mapcat,
-    nth,
-    pipe,
 )
 
 from eth.typing import (
@@ -51,6 +40,9 @@ from eth.typing import (
 from trinity._utils.queues import (
     queue_get_batch,
     queue_get_nowait,
+)
+from trinity._utils.tree_root import (
+    RootTracker,
 )
 
 TPrerequisite = TypeVar('TPrerequisite', bound=Enum)
@@ -436,9 +428,12 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
     _dependency_of: StaticMethod[Callable[[TTask], TTaskID]]
 
     # by default, how long should the integrator wait before pruning?
-    _default_max_depth = 10000  # not sure how to pick a good default here
+    _default_max_depth = 10  # not sure how to pick a good default here
 
     _prereq_tracker: Type[BaseTaskPrerequisites[TTask, TPrerequisite]]
+
+    # track roots
+    _roots: RootTracker[TTaskID]
 
     NoPrerequisites = NoPrerequisites
     """
@@ -457,7 +452,6 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
         self._prereq_tracker = BaseTaskPrerequisites.from_enum(prerequisites)
         self._id_of = id_extractor
         self._dependency_of = dependency_extractor
-        self._oldest_depth = 0
         self._accept_dangling_tasks = accept_dangling_tasks
 
         # how long to wait before pruning
@@ -488,6 +482,8 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
         # Declared finished with set_finished_dependency()
         self._declared_finished: Set[TTaskID] = set()
 
+        self._roots = RootTracker()
+
     def set_finished_dependency(self, finished_task: TTask) -> None:
         """
         Mark this task as already finished. This is a bootstrapping method. In general,
@@ -511,6 +507,7 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
             )
         self._tasks[task_id] = completed
         self._declared_finished.add(task_id)
+        self._roots.add(task_id, self._dependency_of(finished_task))
         # note that this task is intentionally *not* added to self._unready
 
     def register_tasks(self, tasks: Tuple[TTask, ...], ignore_duplicates: bool = False) -> None:
@@ -551,6 +548,7 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
                 self._tasks[task_id] = prereq_tracker
                 self._unready.add(task_id)
                 self._dependencies[dependency_id].add(task_id)
+                self._roots.add(task_id, dependency_id)
 
                 if prereq_tracker.is_complete and self._is_ready(prereq_tracker.task):
                     # this is possible for tasks with 0 prerequisites (useful for pure ordering)
@@ -593,16 +591,29 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
         else:
             return False
 
-    def _mark_complete(self, task_id: TTaskID) -> None:
+    @to_tuple
+    def _mark_complete_task_tree(self, task_id: TTaskID) -> Iterable[TTaskID]:
         qualified_tasks = tuple([task_id])
         while qualified_tasks:
-            qualified_tasks = tuple(concat(
-                self._mark_one_task_complete(task_id)
-                for task_id in qualified_tasks
-            ))
+            next_layer_tasks: Tuple[TTaskID, ...] = tuple()
+            for completed_task in qualified_tasks:
+                child_tasks = self._mark_one_task_complete(completed_task)
+                if not child_tasks:
+                    yield completed_task
+                else:
+                    next_layer_tasks = next_layer_tasks + child_tasks
+            qualified_tasks = tuple(next_layer_tasks)
+
+    def _mark_complete(self, task_id: TTaskID) -> None:
+        new_completed_leaves = self._mark_complete_task_tree(task_id)
+        for prune_from_leaf in new_completed_leaves:
+            # Attempt pruning at least twice (to eventually catch up after forks)
+            # re-running is okay, because pruning limits the prune depth
+            self._prune_finished(prune_from_leaf)
+            self._prune_finished(prune_from_leaf)
 
     @to_tuple
-    def _mark_one_task_complete(self, task_id: TTaskID) -> Generator[TTaskID, None, None]:
+    def _mark_one_task_complete(self, task_id: TTaskID) -> Iterable[TTaskID]:
         """
         Called when this task is completed and its dependency is complete, for the first time
 
@@ -616,9 +627,6 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
         # note that this task has been made ready
         self._unready.remove(task_id)
 
-        # prune any completed tasks that are too old
-        self._prune_finished(task_id)
-
         # resolve tasks that depend on this task
         for depending_task_id in self._dependencies[task_id]:
             # we already know that this task is ready, so we only need to check completion
@@ -627,65 +635,20 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
 
     def _prune_finished(self, task_id: TTaskID) -> None:
         """
-        This prunes any data starting more than _max_depth in history.
+        This prunes the oldest data, if it starts more than _max_depth in history.
         It is called when the task becomes ready.
         """
-        try:
-            oldest_id = self._find_oldest_unpruned_task_id(task_id)
-        except ValidationError:
-            # No tasks are old enough to prune, can end immediately
+        root_task_id, depth = self._roots.get_root(task_id)
+        num_to_prune = depth - self._max_depth
+        if num_to_prune <= 0:
             return
+        else:
+            self._prune(root_task_id)
 
-        root_id, depth = self._find_root(oldest_id)
-        unpruned = self._prune_forward(root_id, depth)
-        if oldest_id not in unpruned:
-            raise ValidationError(
-                f"Expected {oldest_id} to be in {unpruned!r}, something went wrong during pruning."
-            )
-
-    def _validate_has_task(self, task_id: TTaskID) -> None:
-        if task_id not in self._tasks:
-            raise ValidationError(f"No task {task_id} is present")
-
-    def _find_oldest_unpruned_task_id(self, finished_task_id: TTaskID) -> TTaskID:
-        get_dependency_of_id = compose(
-            curry(do)(self._validate_has_task),
-            self._dependency_of,
-            attrgetter('task'),
-            self._tasks.get,
-        )
-        ancestors = iterate(get_dependency_of_id, finished_task_id)
-        return nth(self._max_depth, ancestors)
-
-    def _find_root(self, task_id: TTaskID) -> Tuple[TTaskID, int]:
-        """
-        return the oldest root, and the depth to it from the seed task
-        """
-        root_candidate = task_id
-        get_dependency_of_id = compose(self._dependency_of, attrgetter('task'), self._tasks.get)
-        # We'll use the maximum saved history (_max_depth) to cap how long the stale cache
-        # of history might get, when pruning. Increasing the cap should not be a problem, if needed.
-        for depth in range(0, self._max_depth):
-            dependency = get_dependency_of_id(root_candidate)
-            if dependency not in self._tasks:
-                return root_candidate, depth
-            else:
-                root_candidate = dependency
-        raise ValidationError(
-            f"Stale task history too long ({depth}) before pruning. {dependency} is still in cache."
-        )
-
-    def _prune_forward(self, root_id: TTaskID, depth: int) -> Tuple[TTaskID]:
-        """
-        Prune all forks forward from the root
-        """
-        def prune_parent(prune_task_id: TTaskID) -> Set[TTaskID]:
-            children = self._dependencies.pop(prune_task_id, set())
-            del self._tasks[prune_task_id]
-            if prune_task_id in self._declared_finished:
-                self._declared_finished.remove(prune_task_id)
-            return children
-
-        prune_parent_list = compose(tuple, curry(mapcat)(prune_parent))
-        prune_trunk = repeat(prune_parent_list, depth)
-        return pipe((root_id, ), *prune_trunk)
+    def _prune(self, prune_task_id: TTaskID) -> None:
+        # _roots.prune() has validation in it, so if there is a problem, we should skip the rest
+        self._roots.prune(prune_task_id)
+        del self._dependencies[prune_task_id]
+        del self._tasks[prune_task_id]
+        if prune_task_id in self._declared_finished:
+            self._declared_finished.remove(prune_task_id)
--- a/trinity/_utils/datastructures.py
+++ b/trinity/_utils/datastructures.py
@@ -31,6 +31,7 @@ from eth_utils import (
 )
 from eth_utils.toolz import (
     identity,
+    mapcat,
 )
 
 from eth.typing import (
@@ -415,6 +416,8 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
     The memory needs of this class would naively be unbounded. Any newly
     registered task might depend on any other task in history. To prevent
     unbounded memory usage, old tasks are pruned after a configurable depth.
+    Pruning is triggered when `ready_tasks()` is called, starting from the
+    tail of the *previous* ready_tasks() result.
 
     Vocab:
 
@@ -483,6 +486,7 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
         self._declared_finished: Set[TTaskID] = set()
 
         self._roots = RootTracker()
+        self._last_yielded_tasks: Tuple[TTask, ...] = tuple()
 
     def set_finished_dependency(self, finished_task: TTask) -> None:
         """
@@ -578,7 +582,15 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
         Return the next batch of tasks that are ready to process. If none are ready,
         hang until at least one task becomes ready.
         """
-        return await queue_get_batch(self._ready_tasks)
+        for completed_task in self._last_yielded_tasks:
+            task_id = self._id_of(completed_task)
+            # Attempt pruning at least twice (to eventually catch up after forks)
+            # re-running is okay, because pruning limits the prune depth
+            self._prune_finished(task_id)
+            self._prune_finished(task_id)
+
+        self._last_yielded_tasks = await queue_get_batch(self._ready_tasks)
+        return self._last_yielded_tasks
 
     def _is_ready(self, task: TTask) -> bool:
         dependency = self._dependency_of(task)
@@ -591,26 +603,13 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
         else:
             return False
 
-    @to_tuple
-    def _mark_complete_task_tree(self, task_id: TTaskID) -> Iterable[TTaskID]:
+    def _mark_complete(self, task_id: TTaskID) -> None:
         qualified_tasks = tuple([task_id])
         while qualified_tasks:
-            next_layer_tasks: Tuple[TTaskID, ...] = tuple()
-            for completed_task in qualified_tasks:
-                child_tasks = self._mark_one_task_complete(completed_task)
-                if not child_tasks:
-                    yield completed_task
-                else:
-                    next_layer_tasks = next_layer_tasks + child_tasks
-            qualified_tasks = tuple(next_layer_tasks)
-
-    def _mark_complete(self, task_id: TTaskID) -> None:
-        new_completed_leaves = self._mark_complete_task_tree(task_id)
-        for prune_from_leaf in new_completed_leaves:
-            # Attempt pruning at least twice (to eventually catch up after forks)
-            # re-running is okay, because pruning limits the prune depth
-            self._prune_finished(prune_from_leaf)
-            self._prune_finished(prune_from_leaf)
+            qualified_tasks = tuple(mapcat(
+                self._mark_one_task_complete,
+                qualified_tasks,
+            ))
 
     @to_tuple
     def _mark_one_task_complete(self, task_id: TTaskID) -> Iterable[TTaskID]:
@@ -636,7 +635,24 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
     def _prune_finished(self, task_id: TTaskID) -> None:
         """
         This prunes the oldest data, if it starts more than _max_depth in history.
-        It is called when the task becomes ready.
+        It is called when the task has been consumed by the caller via
+        `ready_tasks()` and completed. The workflow looks something like:
+
+        :return: True if a node was pruned
+
+        ::
+
+            otp = OrderedTaskPreparation(...)
+            otp.register_tasks(range(3))
+            otp.finish_prereq(OnlyPrereq, (0, 1))
+            assert await otp.ready_tasks() == (0, 1)
+
+            # Do some processing on ready tasks (0, 1) ...
+            # Complete processing on ready tasks (0, 1)
+
+            await otp.ready_tasks()
+            # ^ when this is called, pruning is triggered from
+            # the tip of task 1 (whether or not task 2 is ready)
         """
         root_task_id, depth = self._roots.get_root(task_id)
         num_to_prune = depth - self._max_depth
--- a/trinity/_utils/datastructures.py
+++ b/trinity/_utils/datastructures.py
@@ -592,6 +592,9 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
         self._last_yielded_tasks = await queue_get_batch(self._ready_tasks, max_results)
         return self._last_yielded_tasks
 
+    def has_ready_tasks(self) -> bool:
+        return not self._ready_tasks.empty()
+
     def _is_ready(self, task: TTask) -> bool:
         dependency = self._dependency_of(task)
         if dependency in self._declared_finished:
--- a/trinity/_utils/datastructures.py
+++ b/trinity/_utils/datastructures.py
@@ -577,7 +577,7 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
             if task_completion.is_complete and self._is_ready(task):
                 self._mark_complete(task_id)
 
-    async def ready_tasks(self) -> Tuple[TTask, ...]:
+    async def ready_tasks(self, max_results: int = None) -> Tuple[TTask, ...]:
         """
         Return the next batch of tasks that are ready to process. If none are ready,
         hang until at least one task becomes ready.
@@ -589,7 +589,7 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
             self._prune_finished(task_id)
             self._prune_finished(task_id)
 
-        self._last_yielded_tasks = await queue_get_batch(self._ready_tasks)
+        self._last_yielded_tasks = await queue_get_batch(self._ready_tasks, max_results)
         return self._last_yielded_tasks
 
     def _is_ready(self, task: TTask) -> bool:
--- a/trinity/_utils/tree_root.py
+++ b/trinity/_utils/tree_root.py
@@ -1,0 +1,360 @@
+from typing import (
+    Dict,
+    Generic,
+    Optional,
+    Tuple,
+    TypeVar,
+)
+from eth_utils import ValidationError
+
+TNodeID = TypeVar('TNodeID')
+
+
+class Tree(Generic[TNodeID]):
+    """
+    A graph built by adding nodes which only know their parents. Each node
+    must have exactly one parent, but it need not be present at insertion time.
+
+    Because the Tree is built out-of-order, it may actually be a forest of trees
+    at any time.
+
+    Nodes can only be removed if their parent is not present
+    """
+    def __init__(self) -> None:
+        self._parents: Dict[TNodeID, TNodeID] = {}
+        self._children: Dict[TNodeID, Tuple[TNodeID, ...]] = {}
+
+    def has_node(self, node_id: TNodeID) -> bool:
+        return node_id in self._parents
+
+    def has_parent(self, node_id: TNodeID) -> bool:
+        return self.has_node(node_id) and self.has_node(self._parents[node_id])
+
+    def parent_of(self, node_id: TNodeID) -> TNodeID:
+        parent = self._parents[node_id]
+        if self.has_node(parent):
+            parent_children = self.children_of(parent)
+            if node_id not in parent_children:
+                raise ValidationError(
+                    f"Node {node_id} has parent {parent}, but the parent only has children "
+                    f"{parent_children}, not the original node."
+                )
+
+        return parent
+
+    def children_of(self, node_id: TNodeID) -> Tuple[TNodeID, ...]:
+        return self._children.get(node_id, tuple())
+
+    def add(self, node_id: TNodeID, parent_id: TNodeID) -> None:
+        if self.has_node(node_id):
+            raise ValidationError(f"must not re-add same node twice: {node_id}")
+        else:
+            self._parents[node_id] = parent_id
+            children = self._children.get(parent_id, tuple())
+            self._children[parent_id] = children + (node_id, )
+
+    def prune(self, node_id: TNodeID) -> None:
+        if self.has_parent(node_id):
+            raise ValidationError(
+                f"must not prune a node {node_id} that has a parent {self.parent_of(node_id)}"
+            )
+        else:
+            parent_id = self._parents.pop(node_id)
+
+            # remove node ID from parent's children
+            parent_children = self.children_of(parent_id)
+            new_parent_children = tuple(child for child in parent_children if child != node_id)
+            if len(new_parent_children):
+                self._children[parent_id] = new_parent_children
+            else:
+                del self._children[parent_id]
+
+
+class TreeRoot(Generic[TNodeID]):
+    """
+    This class tracks the root of a node in a tree (the node with no parents).
+    The root may be extended to recursively point to the eventual "true" root.
+
+    A key point is that the root is mutable. In this way, when a node is pruned,
+    all the children nodes that point to the root are updated at once, by updating
+    this single tree root.
+
+    The tree root extends in two cases:
+        - the root node was added before its parent, OR
+        - the root node is known to be a fork from its parent at addition time
+
+    Why is it important for these to be extensions, instead of updating the root node
+    to the new root?
+        - When a root node is added before its parent, the depths to the true root are
+        unknown, so the extension provides the opportunity to modify the offset to account
+        for the new depth.
+        - When a root node is a known fork, it is important to be able to mutate the root
+        for one side of the fork, and then mutate the root for the other side of the fork,
+        with distinct root objects.
+    """
+    def __init__(self, root_node_id: TNodeID) -> None:
+        # _root_id is the best-known root node at the time of node insertion
+        self._root_id = root_node_id
+        self._depth_offset = 0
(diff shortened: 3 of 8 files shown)


trinity/_utils/tree_source.py
```diff
@@ -0,0 +1,236 @@
...
+            assert self._get_depth(child) == self._get_depth(node_id) + 1
+
+    def _get_depth(self, node_id):
+        _, depth = self.get_source(node_id)
+        return depth
+
+    def get_source(self, node_id):
+        if node_id not in self._sources:
+            raise ValidationError(f"Node {node_id} is not in the tree")
+        if node_id in self._cache:
```
pipermerriam: *nitpick*

`elif`?

carver:
Ok, now, I thing it should be a bit more reviewable, with some docstrings. Recommended order:
- `tests/core/utils/test_root_tracker.py`
- `trinity._utils.tree_root.Tree`
- `trinity._utils.tree_root.RootTracker`
- `trinity._utils.tree_root.TreeRoot`
- `trinity._utils.datastructures.OrderedTaskPreparation` & tests

If reviewers like, I can pull out the `OrderedTaskPreparation` changes easily and move to a new PR. I left it in for the context of how `RootTracker` will actually be used.

veox:
(I have not really reviewed code in the PR to any extent.)

Running with `--profiling` shows `datastructures.py` is almost gone (maybe 1% of `tottime`); `tree_root.py` is sum of all locations is perhaps 8%.

This seems indeed an improvement, including "visually" during sync: there are no intermittent "pauses" in block download (as described in issue #200), the node is able to maintain continuous sync.

-----

A new (non-fatal) error started popping up; here's a snippet:

``` python
   DEBUG  01-30 00:44:07      HeaderMeatSyncer  Requesting 192 headers from ETHPeer <Node(0x2780@52.213.50.123)>                   
   DEBUG  01-30 00:44:07      HeaderMeatSyncer  Requesting 192 headers from ETHPeer <Node(0x9a69@13.250.15.120)>                       
   DEBUG  01-30 00:44:07            FullServer  Receiving handshake from Address(18.223.185.86:udp:59230|tcp:59230)                
   DEBUG  01-30 00:44:07               ETHPeer  Disconnecting from remote peer <Node(0x399a@18.223.185.86)>; reason: too_many_peers                                                                                                            
    INFO  01-30 00:44:07      HeaderMeatSyncer  Unexpected err while downloading headers from ETHPeer <Node(0x2780@52.213.50.123)>: must not prune a node that has a parent                                                                    
   DEBUG  01-30 00:44:07      HeaderMeatSyncer  Problem downloading headers from peer, dropping...                                                                                                                                             
  File "/home/veox/src/trinity/trinity/sync/common/headers.py", line 599, in _run_fetch_segment
    completed_headers = await peer.wait(self._fetch_segment(peer, parent_header, length))
  File "/home/veox/src/trinity/p2p/cancellable.py", line 20, in wait
  File "/home/veox/src/trinity/p2p/cancellable.py", line 42, in wait_first
  File "/home/veox/src/trinity/.virtualenv/trinity/lib/python3.7/site-packages/cancel_token/token.py", line 152, in cancellable_wait
    return done.pop().result()
  File "/home/veox/src/trinity/trinity/sync/common/headers.py", line 677, in _fetch_segment
    self._stitcher.register_tasks(headers, ignore_duplicates=True)
  File "/home/veox/src/trinity/trinity/_utils/datastructures.py", line 568, in register_tasks
    self._mark_complete(task_id)
  File "/home/veox/src/trinity/trinity/_utils/datastructures.py", line 625, in _mark_complete
    self._prune_finished(prune_from_leaf)
  File "/home/veox/src/trinity/trinity/_utils/datastructures.py", line 660, in _prune_finished
    self._prune(root_task_id)
  File "/home/veox/src/trinity/trinity/_utils/datastructures.py", line 699, in _prune
    self._roots.prune(prune_task_id)
  File "/home/veox/src/trinity/trinity/_utils/tree_root.py", line 265, in prune
    self._tree.prune(prune_off_id)
  File "/home/veox/src/trinity/trinity/_utils/tree_root.py", line 40, in prune
    raise ValidationError("must not prune a node that has a parent")
eth_utils.exceptions.ValidationError: must not prune a node that has a parent
```

-----

And another one, fatal:

``` python
   DEBUG  01-30 00:46:34   FastChainBodySyncer  Got receipts for 64/256 headers from ETHPeer <Node(0xc0d4@119.130.229.225)>, with 0 trivial headers                                                                                            
   DEBUG  01-30 00:46:34   FastChainBodySyncer  Requesting receipts for 256 headers from ETHPeer <Node(0xc0d4@119.130.229.225)>                                                                                                                
   DEBUG  01-30 00:46:34   FastChainBodySyncer  Got receipts for 62/256 headers from ETHPeer <Node(0x2780@52.213.50.123)>, with 0 trivial headers                                                                                              
   DEBUG  01-30 00:46:35   FastChainBodySyncer  Requesting receipts for 256 headers from ETHPeer <Node(0x2780@52.213.50.123)>                                                                                                                  
   DEBUG  01-30 00:46:37   FastChainBodySyncer  <coroutine object FastChainBodySyncer._launch_prerequisite_tasks at 0x7ff918bf6148> finished while <trinity.sync.full.chain.FastChainBodySyncer object at 0x7ff918bdf470> is still running, ter
minating as well                                                                                                                                                                                                                               
   DEBUG  01-30 00:46:37   FastChainBodySyncer  Cancelling <trinity.sync.full.chain.FastChainBodySyncer object at 0x7ff918bdf470>                                                                                                              
 WARNING  01-30 00:46:37   FastChainBodySyncer  Task <coroutine object FastChainBodySyncer._launch_prerequisite_tasks at 0x7ff918befd48> finished unexpectedly: must not re-add same node twice                                                
   DEBUG  01-30 00:46:37   FastChainBodySyncer  Task failure traceback                                                                                                                                                                         
  File "/home/veox/src/trinity/p2p/service.py", line 141, in _run_task_wrapper                                                                                                                                                                 
    await awaitable                                                                                                                                                                                                                            
  File "/home/veox/src/trinity/p2p/service.py", line 160, in _run_daemon_task_wrapper                                                                                                                                                          
    await awaitable                                                                                                                                                                                                                            
  File "/home/veox/src/trinity/trinity/sync/full/chain.py", line 454, in _launch_prerequisite_tasks                                                                                                                                            
    self._block_persist_tracker.register_tasks(headers, ignore_duplicates=True)                                                                                                                                                                
  File "/home/veox/src/trinity/trinity/_utils/datastructures.py", line 564, in register_tasks                                                                                                                                                  
    self._roots.add(task_id, dependency_id)                                                                                                                                                                                                    
  File "/home/veox/src/trinity/trinity/_utils/tree_root.py", line 208, in add                                                                                                                                                                  
    self._tree.add(node_id, parent_id)                                                                                                                                                                                                         
  File "/home/veox/src/trinity/trinity/_utils/tree_root.py", line 32, in add                                                                                                                                                                   
    raise ValidationError("must not re-add same node twice")                                                                                                                                                                                   
eth_utils.exceptions.ValidationError: must not re-add same node twice                                                                                                                                                                          
   DEBUG  01-30 00:46:37   FastChainBodySyncer  <trinity.sync.full.chain.FastChainBodySyncer object at 0x7ff918bdf470> cancelled, cleaning up...                                                                                               
   DEBUG  01-30 00:46:37   FastChainBodySyncer  Waiting for tasks: [<Task pending coro=<FastChainBodySyncer._persist_ready_blocks() running at /home/veox/src/trinity/p2p/service.py:141> wait_for=<Future pending cb=[<TaskWakeupMethWrapper o
bject at 0x7ff69d84ed98>()]>>, <Task pending coro=<FastChainBodySyncer._assign_receipt_download_to_peers() running at /home/veox/src/trinity/p2p/service.py:141> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7ff68ff8c318>(
)]>>, <Task pending coro=<FastChainBodySyncer._display_stats() running at /home/veox/src/trinity/p2p/service.py:141> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7ff6f425ec18>()]>>, <Task pending coro=<BaseBodyChainSynce
r._assign_body_download_to_peers() running at /home/veox/src/trinity/p2p/service.py:141> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7ff8d042f2e8>()]>>, <Task pending coro=<WaitingPeers.put_nowait() running at /home/veo
x/src/trinity/p2p/service.py:141> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7ff69e11cbb8>()]>>]
... <node shutdown continues as usually>
```

-----

I can open new issues if these are not to be addressed in this PR.

veox:
It would seem that the two exceptions are related. The below is pure guesswork.

First, a task-node is to be pruned in `datastructures.py`, and [references are removed from the task list](https://github.com/ethereum/trinity/pull/236/files#diff-8a085c33e95b52e1aba4d210b0d288c3R694) of `OrderedTaskPreparation`, but at the last step of `_prune()`, `self._roots.prune()` fails (non-fatal); the task-node remains in `self._roots`.

Then, the task-node is to be [re-added](https://github.com/ethereum/trinity/pull/236/files#diff-8a085c33e95b52e1aba4d210b0d288c3R564), but the attempt fails (fatal), because the task-node is still in `self._roots`.

trinity/_utils/datastructures.py
```diff
@@ -551,6 +561,7 @@ def register_tasks(self, tasks: Tuple[TTask, ...], ignore_duplicates: bool = Fal
                 self._tasks[task_id] = prereq_tracker
                 self._unready.add(task_id)
                 self._dependencies[dependency_id].add(task_id)
+                self._roots.add(task_id, dependency_id)
```
veox: This is where the tasks-nodes are re-added, even though they are still present in `self._roots`.

It seems that them being `duplicates` is not detected at the beginning of function: after all, they have been removed form `self._tasks`, it is only in `self._roots` that some leftover remains.

trinity/_utils/datastructures.py
```diff
@@ -593,13 +604,27 @@ def _is_ready(self, task: TTask) -> bool:
...
+                else:
+                    next_layer_tasks.extend(child_tasks)
+            qualified_tasks = next_layer_tasks
+
+    def _mark_complete(self, task_id: TTaskID) -> None:
+        new_completed_leaves = self._mark_complete_task_tree(task_id)
+        for prune_from_leaf in new_completed_leaves:
+            # Attempt pruning at least twice (to eventually catch up after forks)
+            # re-running is okay, because pruning limits the prune depth
+            self._prune_finished(prune_from_leaf)
```
veox: When ["must not prune a node that has a parent"](https://gist.github.com/veox/b5d5a7215b3b247fae9174389951e508) is observed, all failures happen on first run (line 625), never second (line 626).

trinity/_utils/tree_root.py
```diff
@@ -0,0 +1,307 @@
...
+
+class Tree:
+    """
+    A graph built by adding nodes which only know their parents. Each node
+    must have exactly one parent, but it need not be present at insertion time.
+
+    Because the Tree is built out-of-order, it may actually be a forest of trees
+    at any time.
+
+    Nodes can only be removed if their parent is not present
```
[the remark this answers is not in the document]
carver: Yes, something like:

```py
rt = RootTracker()
# ... add nodes

root, depth = rt.get_root(leaf)
rt.add(parent_of_root, gp_of_root)

# the following would fail, as it is no longer the root
rt.prune(root)
```

At this point, it's the caller's job not to do something like that. Since there's no `await` in between the `get_root()` and the `prune()` call in `OrderedTaskPreparation`, I don't think this is the particular issue we're seeing.

trinity/_utils/datastructures.py
```diff
@@ -593,13 +604,27 @@ def _is_ready(self, task: TTask) -> bool:
         else:
             return False
 
-    def _mark_complete(self, task_id: TTaskID) -> None:
+    @to_tuple
+    def _mark_complete_task_tree(self, task_id: TTaskID) -> None:
```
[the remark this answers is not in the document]
carver: Yup, just hadn't gotten around to all the type fixups. Return type is correct now. :)

lithp:
Running this and checking with py-spy shows that pruning takes up very little time. The stack traces are enormous though: https://gist.github.com/lithp/8436b5842fa6a492242c30e450480a58#file-stack-trace

While running this I got an error I've never seen before. The node appeared to be making progress but the log was full of warnings: https://gist.github.com/lithp/8436b5842fa6a492242c30e450480a58#file-msg-queue-is-full

I don't think that error is related to these changes. Going to dive into the code now.

tests/core/utils/test_root_tracker.py
```diff
@@ -0,0 +1,146 @@
...
+        # always prune from the end
+        prune_root_id, _ = tracker.get_root(9)
+        tracker.prune(prune_root_id)
+
+        # root should not be retrievable anymore
+        with pytest.raises(ValidationError):
+            tracker.get_root(prune_root_id)
+
+
+FULL_BINARY_TREE = [(exp, num) for exp in [0, 1, 2, 3] for num in range(2**exp)]
```
lithp: ipython convinced me that this does create a binary tree and after some thought I think I know how it works but I think it could be made simpler:

```python
>>> FULL_BINARY_TREE = [(child, child // 2) for child in range(1, 16)]
>>> for child, parent in FULL_BINARY_TREE:
...   print(child, parent)
... 
1 0
2 1
3 1
4 2
5 2
6 3
7 3
8 4
9 4
10 5
11 5
12 6
13 6
14 7
15 7
```

Alternatively, after reading `test_tree_linking` I was thrown off by the fact that `(exp, num)` is not `(child, parent)`. Maybe it'd be easier to do the following? Now that I know how it works I'm no longer a good judge of what's confusing :sweat_smile: 

```python
FULL_BINARY_TREE = [
  # starting from the leftmost node for a given depth, (depth, i) is the i'th node
  ((depth, i), (depth-1, i // 2))  # (child, parent)
  for depth in [0, 1, 2, 3] for i in range(2**depth)
]
```

tests/core/utils/test_root_tracker.py
```diff
@@ -0,0 +1,146 @@
...
+    # verify parent and depth of partially-built tree
+    for node in nodes_to_insert:
+        actual_root, actual_depth = tracker.get_root(node)
+        expected_root, expected_depth = get_expected_root(node, nodes_to_insert)
+        assert actual_root == expected_root
+        assert actual_depth == expected_depth
+
+    # prune
+    remaining_nodes = set(nodes_to_insert)
+    for prune_idx in [idx for idx in prune_order if idx < len(nodes_to_insert)]:
```
lithp: This is probably not worth the extra complication, but you can simplify this line and have `prune_order` reference just the subset of nodes which were passed in by complicating the preceding lines:

```python
@st.composite
def subset_and_order(draw):
    nodes_to_insert = draw(st.lists(st.sampled_from(FULL_BINARY_TREE), unique=True))
    prune_order = draw(st.permutations(range(len(nodes_to_insert))))
    return (nodes_to_insert, prune_order)

@given(subset_and_order())
@settings(max_examples=500)
def test_sparse_branching(test_data):
    nodes_to_insert, prune_order = test_data
    ...
```

tests/core/utils/test_root_tracker.py
lithp: Though, now that I read `test_full_branching` my first proposal seems like a bad idea. `(3, 0)` is easier to understand than `8`.

lithp:
Still looking through the code, it's taking me a while to grok this. Am I blocking you at all?

trinity/_utils/tree_root.py
```diff
@@ -0,0 +1,315 @@
...
+
+    @property
+    def _is_extension(self) -> bool:
+        return self._extends_root is not None
+
+    def extend(self, tree_root: 'TreeRoot[TNodeID]', more_depth: int) -> None:
+        """
+        Indicate that what was once the original root now has a new root, because
+        a parent node was added. tree_root is the new root node, and more_depth
+        is how much deeper the new root is beyound the current root.
```
lithp: s/beyound/beyond

trinity/_utils/tree_root.py
```diff
@@ -0,0 +1,315 @@
...
+    This class tracks the root of a node in a tree (the node with no parents).
+    The root may be extended to recursively point to the eventual "true" root.
+
+    A key point is that the root is mutable. In this way, when a node is pruned,
+    all the children nodes that point to the root are updated at once, by updating
+    this single tree root.
+
+    The tree root extends in two cases:
+        - the root node was added before its parent, OR
+        - the root node is known to be a fork from its parent at addition time
```
lithp: These two lines are confusing, a tree root doesn't have parents!

trinity/_utils/tree_root.py
```diff
@@ -0,0 +1,315 @@
...
+        self._cache = {}
+        pruned_root = self._roots.pop(prune_off_id)
+        if pruned_root.node_id != prune_off_id:
+            raise ValidationError(
+                f"Can only prune of a root node, tried to prune {prune_off_id}, "
+                f"but the root is {pruned_root}"
+            )
+        child_nodes = self._tree.children_of(prune_off_id)
+        child_roots = tuple(self._roots[child] for child in child_nodes)
+        child_pairs = tuple(zip(child_nodes, child_roots))
```
lithp: nitpick: I usually understand comprehensions a little faster:

```python
child_pairs = (
  (child_node, self._roots[child_node])
  for child_node in child_nodes
)
```

trinity/_utils/tree_root.py
```diff
@@ -0,0 +1,315 @@
...
+                ideal_original_depth = parent_original_depth + 1
+                actual_original_depth = self._original_depth_to_root[child]
+                child_root.extend(parent_root, ideal_original_depth - actual_original_depth)
+
+        elif len(children) > 1:
+            for child in children:
+                extension_length = parent_original_depth - self._original_depth_to_root[child] + 1
+                self._roots[child].extend(parent_root, extension_length)
+
+    def _get_depth(self, node_id: TNodeID) -> int:
```
lithp: I'm surprised lint didn't catch this but this method is unused.

trinity/_utils/tree_root.py
```diff
@@ -0,0 +1,315 @@
...
+            node_root = TreeRoot(node_id)
+            original_depth = 0
+
+        return node_root, original_depth
+
+    def _link_children(
+            self,
+            parent_root: TreeRoot[TNodeID],
+            parent_original_depth: int,
+            children: Tuple[TNodeID, ...]) -> None:
```
lithp: I think that the logic is identical in the cases where you have one child and where you have multiple children. Maybe they could be combined?

```python
    def _link_children(
            self,
            parent_root: TreeRoot[TNodeID],
            parent_original_depth: int,
            children: Tuple[TNodeID, ...]) -> None:

        for child in children:
            child_root = self._roots[child]
            child_original_depth = self._original_depth_to_root[child]

            if child_original_depth != 0:
                # the only way for the original depth to be non-zero is from pruning
                # it's invalid to prune and re-add a node
                raise ValidationError("children who gain a parent must start at original depth 0")
            elif child_root.depth_offset != 0:
                raise ValidationError("children without parents must have net depth 0")

            extension_length = parent_original_depth - child_original_depth + 1
            child_root.extend(parent_root, extension_length)
```

trinity/_utils/datastructures.py
```diff
@@ -630,18 +645,12 @@ def _prune_finished(self, task_id: TTaskID) -> None:
         This prunes any data starting more than _max_depth in history.
```
lithp: This comment isn't true anymore, it only prunes the deepest task.

trinity/_utils/datastructures.py
```diff
@@ -675,17 +684,10 @@ def _find_root(self, task_id: TTaskID) -> Tuple[TTaskID, int]:
...
-        prune_parent_list = compose(tuple, curry(mapcat)(prune_parent))
-        prune_trunk = repeat(prune_parent_list, depth)
-        return pipe((root_id, ), *prune_trunk)
+    def _prune(self, prune_task_id: TTaskID) -> Set[TTaskID]:
+        children = self._dependencies.pop(prune_task_id, set())
+        del self._tasks[prune_task_id]
+        if prune_task_id in self._declared_finished:
+            self._declared_finished.remove(prune_task_id)
+        self._roots.prune(prune_task_id)
+        return children
```
lithp: This result is never consumed by any caller. Maybe it should return None?

trinity/_utils/datastructures.py
```diff
@@ -630,18 +645,12 @@ def _prune_finished(self, task_id: TTaskID) -> None:
         This prunes any data starting more than _max_depth in history.
         It is called when the task becomes ready.
         """
-        try:
-            oldest_id = self._find_oldest_unpruned_task_id(task_id)
-        except ValidationError:
-            # No tasks are old enough to prune, can end immediately
+        root_task_id, depth = self._roots.get_root(task_id)
```
lithp: Wow, I see what you mean. This was a lot of work for something you only need to call once!

trinity/_utils/tree_root.py
pipermerriam: I believe the linter only catches unused variables and not functions or methods.

trinity/_utils/tree_root.py
carver: We could replace the above comment with:

> All nodes have indicated parents. Those parents may not be in our tree yet, though. When the parent is not present, we consider the child a root. This is the only kind of root that exists, one where the parent is not present *yet*.
>
> We use `extend()` when a parent is added after its child. Then the child will report its new parent's root as the child's root.

---
I think this part could move to the `RootTracker` since it's less about `TreeRoot` and more about how we are using it:

> When using `TreeRoot` to track the root of a node, we reuse the same instance as often as possible. One of the key benefits of that is being able to mutate the root for all nodes with a single operation (eg~ when we prune a root away). But we use multiple `TreeRoot`s that get chained together by `extend()`. Why aren't we keeping exactly one `TreeRoot` for each entire subtree?
>
> First, subtrees can be joined. At join time, dropping from two roots to one would mean iterating through all the nodes in one subtree and changing its `TreeRoot` instance.
> 
> Second, a parent with n children can be pruned. That means creating at n-1 new `TreeRoot`s and worse, walking the n-1 subtrees of all the children to set a new `TreeRoot`.
> 
> So instead, we "chain" the TreeRoot's together when a parent is added after its child. We also chain the TreeRoot's together when a child joins and a sibling is present. That way, at prune time, there are already distinct instances of TreeRoot for each new subtree.

---

Would that be any clearer? Any suggestions for changes? It's *sooo* long that I worry no one would read it. Not sure how to make it shorter, though.

trinity/_utils/tree_root.py
carver: Yeah, since the private methods are only a convention, it could be a public API for all the linter knows. Thanks for the catch.

carver:
> The stack traces are enormous though

Yeah, I see the problem, and it certainly limits how deep the trees can be. I haven't come up with any good solutions unfortunately :(

Maybe when a parent comes in after the child, we could find a way to share the `TreeRoot`. Something like making the original depth of the parent -1 and increasing the depth offset. That's not too bad, but it doesn't work with the new parent is joining two subtrees. (this unfortunately happens a lot because we get tasks in batches, like task (4, 5, 6), then (1, 2, 3). So by the time we add 3, it is joining subtrees. The new tree root has already been created for 1, and used on 2. We could include a new `batch_add()` that tries to detect this case and add nodes in reverse. That's a lot of extra complexity, so I'd want to be confident that it's solving a significant problem.

carver:
> Still looking through the code, it's taking me a while to grok this. Am I blocking you at all?

Nope, I'm going to write up the `CREATE2` thing next. Besides, I still have to figure out why it broke for veox.

> While running this I got an error I've never seen before. The node appeared to be making progress but the log was full of warnings: https://gist.github.com/lithp/8436b5842fa6a492242c30e450480a58#file-msg-queue-is-full
> 
> I don't think that error is related to these changes. Going to dive into the code now.

Agree, it seems unrelated.

lithp:
> Maybe when a parent comes in after the child, we could find a way to share the `TreeRoot`. Something like making the original depth of the parent -1 and increasing the depth offset. That's not too bad, but it doesn't work with the new parent is joining two subtrees. (this unfortunately happens a lot because we get tasks in batches, like task (4, 5, 6), then (1, 2, 3). So by the time we add 3, it is joining subtrees. The new tree root has already been created for 1, and used on 2. We could include a new `batch_add()` that tries to detect this case and add nodes in reverse. That's a lot of extra complexity, so I'd want to be confident that it's solving a significant problem.

Heh, believe it or not I tried my hand at doing exactly this yesterday (Adding a parent with -1 and bumping the depth offset). I got a few errors like this:

https://github.com/ethereum/trinity/blob/d3598261026e37335be52e309ad91a02f9a5b3f8/trinity/_utils/tree_root.py#L298

and stopped because you're right, it feels like this optimization has limited benefit. If I understand it right the only place this optimization would help is `BaseHeaderChainSyncer`, because it's the only place with `accept_dangling_tasks=True`?

If I understand it right this code makes writes fast at the cost of slow worst-case reads. Merging subtrees would make both fast (in the case where there's no branching). This is already a big performance improvement though, I don't think that adding merging would require a large refactor so it makes sense to wait until we know that we need that extra performance.

trinity/_utils/tree_root.py
```diff
@@ -0,0 +1,308 @@
...
+                raise ValidationError(
+                    f"Error while pruning node: child root {child_root} is not extension "
+                    f"and is not the pruned node {self}"
+                )
+        prune_off_id = self.node_id
+        old_depth_offset = self.depth_offset
+        for child_id, child_root in child_pairs:
+            child_root._prune(prune_off_id, child_id, old_depth_offset)
+
+    @property
```
lithp: This is not the prettiest solution but it might be the least-worst. What do you think of fixing the huge stack traces by making these functions iterative?

```python
    @property
    def node_id(self) -> TNodeID:
        """
        Return the node all the way at the true root
        """
        node = self
        while node._is_extension:
            node = node._extends_root
        return node._root_id

    @property
    def depth_offset(self) -> int:
        """
        Return the depth all the way to the true root
        """
        node = self
        extension_depth = 0
        while node._is_extension:
            extension_depth += node._depth_extension
            node = node._extends_root
        return node._depth_offset + extension_depth
```

trinity/_utils/tree_root.py
```diff
@@ -0,0 +1,308 @@
...
+
+    def extend(self, tree_root: 'TreeRoot[TNodeID]', more_depth: int) -> None:
+        """
+        Indicate that what was once the original root now has a new root, because
+        a parent node was added. tree_root is the new root node, and more_depth
+        is how much deeper the new root is beyond the current root.
+        """
+        if self._is_extension:
+            raise ValidationError("Cannot extend a node that is already extended...")
+        self._depth_extension = more_depth
```
lithp: I don't know what convention is but `_depth_extension` is never declared in `__init__`, what do you think of putting

```python
self._depth_extension: Optional[int] = None
```

up there?

tests/core/utils/test_root_tracker.py
lithp: Oh yeah, good point!

tests/core/utils/test_root_tracker.py
lithp: That sounds great to me.

trinity/_utils/datastructures.py
```diff
@@ -627,21 +642,15 @@ def _mark_one_task_complete(self, task_id: TTaskID) -> Generator[TTaskID, None,
...
+        num_to_prune = depth - self._max_depth
+        if num_to_prune <= 0:
             return
-
-        root_id, depth = self._find_root(oldest_id)
```
lithp: Now that this line is gone `_find_root` is unused and can also be removed.

lithp:
While you're here it looks like `OrderedTaskPreparation._oldest_depth` is also unused and can be removed.

trinity/_utils/datastructures.py
```diff
@@ -627,21 +642,15 @@ def _mark_one_task_complete(self, task_id: TTaskID) -> Generator[TTaskID, None,
 
     def _prune_finished(self, task_id: TTaskID) -> None:
         """
-        This prunes any data starting more than _max_depth in history.
+        This prunes the oldest data, if it starts more than _max_depth in history.
         It is called when the task becomes ready.
         """
-        try:
-            oldest_id = self._find_oldest_unpruned_task_id(task_id)
```
lithp: Same here, as of this PR `_find_oldest_unpruned_task_id` is no longer used and can be removed.

lithp:
Sorry for the long comment, I've left it in the order I wrote it because this sequence might be the best way to figure out that's wrong with what I concluded. If you skip to the end you can see that conclusion:

---

This is wild speculation but it seems like a lot of the complication here is in keeping track of exactly how deep each node is. That's more information than is consumed though, `OrderedTaskPreparation` only cares about whether the root is deeper than some threshold. Maybe it's possible to simplify the code by doing the minimum necessary to allow pruning ancient nodes.

For both of the below I'm relying on the following: `_prune_finished` is only called on nodes which are ready, meaning their dependencies are all ready, meaning there are no missing gaps all the way back to whatever was first passed to `set_finished_dependency`. That means `_prune_finished` has no danger of being too aggressive and preemptively pruning a subtree, preventing said subtree from being linked to a parent.

1. In lieu of tracking depth and trying to cap it, what if `TreeRoot` capped the number of nodes it tracks. Roots are evicted if there are more than 10k nodes?

2. Why not always prune the root?

I think that currently the root isn't pruned because we might discover a deep fork, one which branches off of a point we've already pruned away. If no history was kept then such a fork would result in `MissingDependency`. (Worse, if `accept_dangling_tasks` is `True` then `OrderedTaskPrep` will wait forever, never realizing the dependency is so ready that it's already done) This sounds bad but I think that if that's the only reason then pruning ready nodes is always okay.

My reason is that `MissingDependency` is already a risk! If a fork appears from before trinity started then the fork will occur before whatever was passed to `set_finished_dependency`, this looks exactly like an `OrderedTaskPreparation` which pruned too aggressively. Because it's already a risk, it's already being handled:

https://github.com/ethereum/trinity/blob/2058916a589aa7854778d7dca331bfa292c2581a/trinity/sync/full/chain.py#L455-L479

There's no need for `OrderedTaskPreparation` to keep track of which nodes have already become ready, because the caller will always be able to tell by themselves.

Here are the other instances of `register_tasks`:
1. Independently of this PR `RegularChainBodySyncer` should be changed to do the same thing `FastChainBodySyncer` is doing, and once it does it'll be fine with aggressively pruned tasks:

https://github.com/ethereum/trinity/blob/d3598261026e37335be52e309ad91a02f9a5b3f8/trinity/sync/full/chain.py#L841-L852

2. `BaseHeaderChainSyncer` has two modes:
- `_full_skeleton_sync` fetches a skeleton from a single peer by calling `set_finished_dependency` on the first header it receives and only ever fetching descendants of that header. That it scans forward means pruning doesn't impact it, it never goes backwards and attempts to add nodes which might be branches of already-pruned nodes.
- `HeaderMeatSyncer` validates received segments and makes sure the first element matches what is expected before calling `register_tasks`. That validation ensures that a branch of an already-pruned node will never be added to `OrderedTaskPreparation`. The exception to this is duplicates. If, for some reason, a segment was fetched twice then the first might be pruned away by the time the second one is registered. I think that case can be prevented by changes to `HeaderMeatSyncer`.

---

I guess my idea is: it doesn't look like `OrderedTaskPreparation` needs to keep around so many nodes. It can prune away any ready root (and continue pruning until the root is not ready). This removes the need to keep track of depths, once that tracking is gone the remaining code would be simpler.

In fact, I think this means that nodes can be pruned as soon as they're passed to `_ready_tasks`! That sounds wrong but I'm not sure why it's wrong.

trinity/_utils/tree_root.py
carver: Seems better than blowing up the stack :+1: and basically the same speed.

carver:
> Heh, believe it or not I tried my hand at doing exactly this yesterday (Adding a parent with -1 and bumping the depth offset). I got a few errors like this:
> 
> [trinity/trinity/_utils/tree_root.py](https://github.com/ethereum/trinity/blob/d3598261026e37335be52e309ad91a02f9a5b3f8/trinity/_utils/tree_root.py#L298)
> 
> Line 298 in [d359826](/ethereum/trinity/commit/d3598261026e37335be52e309ad91a02f9a5b3f8)
> 
>  raise ValidationError("children who gain a parent must start at original depth 0") 

Yeah, I'm not certain that the validation is strictly required, but it made things much easier to reason about.

> and stopped because you're right, it feels like this optimization has limited benefit. If I understand it right the only place this optimization would help is `BaseHeaderChainSyncer`, because it's the only place with `accept_dangling_tasks=True`?

Right, this only affects dangling tasks.

> 
> If I understand it right this code makes writes fast at the cost of slow worst-case reads. Merging subtrees would make both fast (in the case where there's no branching). This is already a big performance improvement though, I don't think that adding merging would require a large refactor so it makes sense to wait until we know that we need that extra performance.

:+1: for focusing on whatever the next slow piece is.

Now, on to read that essay... X)

carver:
> I guess my idea is: it doesn't look like `OrderedTaskPreparation` needs to keep around so many nodes. It can prune away any ready root (and continue pruning until the root is not ready). This removes the need to keep track of depths, once that tracking is gone the remaining code would be simpler.

I don't see anything specifically problematic: as you mentioned, we have to deal with forks from before we started syncing, as well. I'm inclined to try it!

> 
> In fact, I think this means that nodes can be pruned as soon as they're passed to `_ready_tasks`! That sounds wrong but I'm not sure why it's wrong.

I'm not *sure* I want to get rid of all this machinery yet (though I do love deleting things!). In the meantime, we could simulate it today by setting `max_depth=0`...

Which brings up an interesting point: I think the current PR does not verify that the task is ready before pruning it! So that's definitely something worth fixing (and maybe checking to see if that somehow is related to veox's bug).

Anyway, since a `MissingDependency` causes a database lookup, and uncles are fairly common, I think it would be better to keep at least a few blocks in history. (say, 10)

Also, we have to deal with the issue that dangling uncles will never get pruned using the current (or former!) model. (especially dealing with dangling uncles, when OTP accepts dangling tasks, so they may never even become ready)

carver:
I have no idea how to reasonably prune uncle chains whose roots have been trimmed, other than explicitly tracking "created uncles" at prune time. That still doesn't help when you accept dangling tasks and get a fork whose parent isn't in memory. Unfortunately, that looks a whole lot like the out-of-order header segments that we get during skeleton sync.

lithp:
> I don't see anything specifically problematic: as you mentioned, we have to deal with forks from before we started syncing, as well. I'm inclined to try it!

Sorry again for the essay!

> Which brings up an interesting point: I think the current PR does not verify that the task is ready before pruning it! So that's definitely something worth fixing (and maybe checking to see if that somehow is related to veox's bug).

Your last commit looks like a nice defensive addition but I think that this PR already indirectly checks. Pruning is only called as a result of `_mark_complete`. A root will only be pruned if it's descendant was ready, which means the root must be ready.

> Anyway, since a `MissingDependency` causes a database lookup, and uncles are fairly common, I think it would be better to keep at least a few blocks in history. (say, 10)

Yeah, good point, and a `max_depth` of 10 won't hurt performance at all.

> I have no idea how to reasonably prune uncle chains whose roots have been trimmed, other than explicitly tracking "created uncles" at prune time.

OTP could check by only pruning nodes which are ready and for which all children are also ready. When a task is complete OTP looks up the root and tries to prune it (which is fast since `max_depth` is 10). Since pruning only occurs when all children are also ready this pruning will leave behind two trees, each of which are rooted with ready nodes.

> Also, we have to deal with the issue that dangling uncles will never get pruned using the current (or former!) model. (especially dealing with dangling uncles, when OTP accepts dangling tasks, so they may never even become ready)

> That still doesn't help when you accept dangling tasks and get a fork whose parent isn't in memory. Unfortunately, that looks a whole lot like the out-of-order header segments that we get during skeleton sync.

This is an annoying foot-gun but I can't think of an easy way to prevent it either. I think all current callers are safe, skeleton sync always fetches in-order and meat syncer checks that segments aren't dangling before committing them to OTP, the only case where forever-dangling-segments might be added is if it received the same segment twice.

Something which could be done is: skeleton sync knows the height of every header. I'm not sure whether it would be `BaseHeaderChainSyncer` or OTP's responsibility but something has gone wrong if there are ever unready tasks in OTP with a lower height than the last header which was emitted by `ready_tasks`. A check which makes sure there are no stale tasks in OTP would prevent forever-dangling tasks.

carver:
> Your last commit looks like a nice defensive addition but I think that this PR already indirectly checks. Pruning is only called as a result of `_mark_complete`. A root will only be pruned if it's descendant was ready, which means the root must be ready.

:man_facepalming: Right, maybe it's not worth the extra check, then, in performance-sensitive code.

> Sorry again for the essay!

Haha, no it was good, it helped!

> OTP could check by only pruning nodes which are ready and for which all children are also ready. When a task is complete OTP looks up the root and tries to prune it (which is fast since `max_depth` is 10). Since pruning only occurs when all children are also ready this pruning will leave behind two trees, each of which are rooted with ready nodes.

I don't follow how this helps. So you start with:
```
A --- B1
   \_ B2
```
Then you add C, say with no prerequisites, with `max_depth=1`, and get:
```
X --- B1 -- C
   \_ B2
```

Then the chain continues along adding D, E, etc. When does `B2` get pruned?

Maybe we a task's "height" and when pruning one task, we prune all tasks at that same height (ready or not!). That only helps if the uncles used to be in the same tree, so the height can be tracked, but that should only be a problem at startup time.

Which starts to sound like the other solution you mentioned. It doesn't seem right that other objects would be responsible for explicitly pruning. Maybe if dangling tasks are accepted, we require that a "height" be derivable from the task (like how we calculate the ID and prereq from a header). It means yet another data store, looking up from height to tasks, so that pruning can prune siblings, too.

I hate adding the complexity. Still looking for better options.

lithp:
> Then the chain continues along adding D, E, etc. When does `B2` get pruned?

Oh, now I see what you mean! My thought was that B1 gets pruned once C1 is ready, that's triggered by `finish_prereq`. B2 will likewise be pruned once a C2 is added and becomes ready. You're right though, what if C2 *never* comes in. I'm not sure what to do there.

lithp:
On further thought, I think this is okay:

```
X --- B1 -- C
   \_ B2
```

The proposed rule: `only prune nodes which are ready, have children, and all children are also ready` would not prune B2, and I think that's valuable, because C2 *might* come in! And when C2 is ready B2 will be pruned, so each branch doesn't take up many resources. This is a memory leak in a long-running trinity node which observes many uncles, but that's already a problem with the current code.

carver:
Unfortunately, I can't seem to sync from the Consensys office to try to reproduce veox 's bug. I am still stuck on what might be causing it. On the up-side, I *half* fixed it by making sure that pruning doesn't happen and *then* fail. So we should only get soft errors instead of hard errors now.

carver:
> This is a memory leak in a long-running trinity node which observes many uncles, but that's already a problem with the current code.

Yeah, okay, I guess you could say that we should separate the speedup in this PR from the memory leak bugfix.

Also, I just realized this: if we're only keeping 10 nodes in history, the speed of the old naive approach for finding the root might be probably fine :see_no_evil: -- Although I guess it depends on how long the unready queue gets.

lithp:
> Also, I just realized this: if we're only keeping 10 nodes in history, the speed of the old naive approach for finding the root might be probably fine  -- Although I guess it depends on how long the unready queue gets.

Ah, sorry, I could have been more explicit. That's what I was trying to say with the essay.

Re a long unready queue, that seems likely to happen with the skeleton syncer, if a lot of segments come in before the lowest segment then a bunch of tasks could become ready at once. This PR improves performance in that case, without it there could be occasional spikes of activity as pruning iterates the list again and again.

lithp:
Just profiled this branch as well, datastructures.py takes up around 2.3% of time, this is still a pretty big win over #266!

> That still doesn't help when you accept dangling tasks and get a fork whose parent isn't in memory.

This weekend I realized there's an easy partial solution to this. Currently `accept_dangling_tasks` is a property of the instance, but if it was a parameter of `register_tasks` then callers like [`HeaderMeatSyncer`](https://github.com/ethereum/trinity/blob/master/trinity/sync/common/headers.py#L677) could be safe from accidentally committing dangling tasks. Happy to open another PR for that if you'd like.

carver:
> This weekend I realized there's an easy partial solution to this. Currently `accept_dangling_tasks` is a property of the instance, but if it was a parameter of `register_tasks` then callers like [`HeaderMeatSyncer`](https://github.com/ethereum/trinity/blob/master/trinity/sync/common/headers.py#L677) could be safe from accidentally committing dangling tasks. Happy to open another PR for that if you'd like.

So this means that the bones are registered as dangling but the meat are registered as non-dangling. Seems reasonable, even as just a safety/validation measure. :+1:

I don't see the win on tracking uncle chains to trim, but doesn't seem like a problem.

lithp:
I agree that it doesn't solve the dangling uncle memory leak :cry:

trinity/_utils/tree_root.py
```diff
@@ -0,0 +1,340 @@
...
+            else:
+                return (root_node_id, root_depth)
+        else:
+            root = self._roots[node_id]
+            original_depth = self._original_depth_to_root[node_id]
+            (root_node_id, root_depth) = (root.node_id, original_depth + root.depth_offset)
+            if self._tree.has_parent(root_node_id):
+                parent = self._tree.parent_of(root_node_id)
+                raise ValidationError(
+                    f"{root_node_id} has parent {parent}, but was going to be returned as a root"
```
veox: This still seems to be happening. Dropped a log snippet to file in gist:

https://gist.github.com/veox/22f44209bc6e5d8f1d486869e4ad9918#file-3-and-another-one-py

carver: Yup, I'm still getting this too, even after the latest updates from 25 min ago (commit 58834c0). Digging deeper now...

carver:
Okay, so it looks like the  `was going to be returned as a root` bug was ultimately caused by pruning away and re-adding the pruned task nodes.

This was explicitly not yet supported, and was supposed to throw a noisier exception early on if it came up. Unfortunately, lack of support had a subtle effect on the tracking of a tree that prevented the warning from showing. Anyway, I added support for pruning and re-adding the same nodes, plus a relevant test that fails before the implementation.

Things have run smoothly on my local node test for over an hour now. The bug is no longer reproducible for me.

I'm adding and keeping peers, using this local variant to skip essentially all PoW checks, since they seem buggy:
```diff
diff --git a/p2p/constants.py b/p2p/constants.py
index b75cf0ac..fb233fc9 100644
--- a/p2p/constants.py
+++ b/p2p/constants.py
@@ -105,7 +105,7 @@ MAX_REORG_DEPTH = 24
 # 100 was the optimal value determined by geth devs
 # (https://github.com/ethereum/go-ethereum/pull/1889#issue-47241762), but in order to err on the
 # side of caution, we use a higher value.
-SEAL_CHECK_RANDOM_SAMPLE_RATE = 48
+SEAL_CHECK_RANDOM_SAMPLE_RATE = 9999999999999999999
 
 
 # The amount of time that the BasePeerPool will wait for a peer to boot before
diff --git a/trinity/sync/common/headers.py b/trinity/sync/common/headers.py
index 42d488ad..fb6fc9c6 100644
--- a/trinity/sync/common/headers.py
+++ b/trinity/sync/common/headers.py
@@ -158,7 +158,7 @@ class SkeletonSyncer(BaseService, Generic[TChainPeer]):
             pairs = tuple(zip(parents, children))
             try:
                 validate_pair_coros = [
-                    self.wait(self._chain.coro_validate_chain(parent, (child, )))
+                    self.wait(self._chain.coro_validate_chain(parent, (child, ), 9999999999999999999))
                     for parent, child in pairs
                 ]
                 await asyncio.gather(*validate_pair_coros, loop=self.get_event_loop())
```

carver:
@cburgdorf @lithp it would be awesome to get a trinity release out tomorrow that supports the new Constantinople (since it's hitting tomorrow), so I'm looking for a :+1: 

I think this PR gets us to good enough to release (as you can see by the checklist, still quite a few follow-up items, though).

carver:
So... now the biggest problem is that there is no back pressure when the peer syncing races ahead of the block import to database. The in-memory data for was up to block 1.5M while the database was back at block 700k. Memory usage went to at least 16GB, and trinity died sometime later (maybe because of OOM killer, but I couldn't confirm in the logs).

I think we'll have to add something to halt the peer download tasks when the queue of block import tasks gets too long. I'll sleep on whether I can do that without adding more API surface area to `OrderedTaskPreparation`.

tox.ini
```diff
@@ -115,7 +115,7 @@ commands=
     flake8 {toxinidir}/scripts
     flake8 {toxinidir}/eth2
     # TODO: Drop --ignore-missing-imports once we have type annotations for eth_utils, coincurve and cytoolz
-    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p p2p -p trinity -p eth2
+    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p p2p -p trinity
```
cburgdorf: Removing `eth2` looks like an unintentional change that slipped in here?

PR #241, opened 2019-01-30
[WIP] Fix `import_block` and add `state_root` update

hwwhww:
### What was wrong?

WIP

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/state_machines/base.py
+++ b/eth2/beacon/state_machines/base.py
@@ -72,7 +72,9 @@ class BaseBeaconStateMachine(Configurable, ABC):
     # Import block API
     #
     @abstractmethod
-    def import_block(self, block: BaseBeaconBlock) -> Tuple[BeaconState, BaseBeaconBlock]:
+    def import_block(self,
+                     block: BaseBeaconBlock,
+                     is_proposer_mode: bool=False) -> Tuple[BeaconState, BaseBeaconBlock]:
         pass
 
     @staticmethod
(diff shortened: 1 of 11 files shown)


eth2/beacon/state_machines/base.py
```diff
@@ -72,7 +72,9 @@ def state_transition(self) -> BaseStateTransition:
     # Import block API
     #
     @abstractmethod
-    def import_block(self, block: BaseBeaconBlock) -> Tuple[BeaconState, BaseBeaconBlock]:
+    def import_block(self,
+                     block: BaseBeaconBlock,
+                     is_proposer_mode: bool=False) -> Tuple[BeaconState, BaseBeaconBlock]:
```
jannikluhn:
```suggestion
                     check_proposer_signature: bool=False) -> Tuple[BeaconState, BaseBeaconBlock]:
```

Just a suggestion to make it more explicit what's going on.

eth2/beacon/state_machines/base.py
```diff
@@ -139,12 +137,17 @@ def state_transition(self) -> BaseStateTransition:
     #
     # Import block API
     #
-    def import_block(self, block: BaseBeaconBlock) -> Tuple[BeaconState, BaseBeaconBlock]:
+    def import_block(self,
```
jannikluhn: It's not part of this PR so it shouldn't be addressed here, but I'm not sure if `import_block` is a good name for this function. It sounds like some database stuff would go on in the background. I can't think of a good alternative though (maybe because I don't really know what the point of the function is).

hwwhww: The name `import_block` is copied from `eth` module. The similar API might be called `add_blcok` in pyethereum. I'm fine either way.
 /cc @pipermerriam

cburgdorf: >it sounds like some database stuff would go on in the background.

But that's in fact what is happening, no?

https://github.com/ethereum/trinity/blob/b2aa50982508614e1720550b22ea92665ef923d2/eth2/beacon/chains/base.py#L387

cburgdorf: Ah, wait. This isn't the `import_block` on the chain but on the `state_machine`. Scratch that.

Issue #235, opened 2019-01-28
Make epochs first class citizens and epoch trans at end of epoch


hwwhww:
### What is wrong?

Spec-PRs:
- Make epochs first-class citizens: https://github.com/ethereum/eth2.0-specs/pull/499
- Epoch trans at end of epoch: https://github.com/ethereum/eth2.0-specs/pull/492

### How can it be fixed
- Move epoch transition to the last slot in the epoch
- Change a lot `slot` to `epoch` (see https://github.com/ethereum/eth2.0-specs/pull/492)

These two goals are twisted together. Ideally, we can start at https://github.com/ethereum/eth2.0-specs/pull/499 and then do https://github.com/ethereum/eth2.0-specs/pull/492 separately, but I suspect it will be harder to review...

NIC619:
What's left from #228 : 
- `slot_to_epoch`
- move `get_current_epoch_attestations`/`get_previous_epoch_attestations` to `state`

PR #252, opened 2019-02-01
RANDAO processing

jannikluhn:
### What was wrong?

RANDAO processing was missing and the original PR for this was outdated.

### How was it fixed?

Implemented the function `process_randao` plus the helpers `validate_randao`, `slot_to_epoch`, and `get_current_epoch` according to the spec.

However, some [signature checking tests](https://github.com/ethereum/trinity/compare/master...jannikluhn:randao-processing?expand=1#diff-814adfb940e4a4f6bbba14b955593107R135) are failing and I don't understand why: Even though a different message is signed, `bls.verify` still returns `True`. Any idea what the issue is? (this is the reason why I labeled it as WIP.)

There are two `FIXME` comments that I think should be addressed in #235 (or here, depending what's merged earlier, cc @NIC619 ).

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/state_machines/forks/serenity/block_processing.py
+++ b/eth2/beacon/state_machines/forks/serenity/block_processing.py
@@ -50,6 +50,10 @@ def process_eth1_data(state: BeaconState,
     return state
 
 
+from eth2.beacon.typing import (
+    SlotNumber,
+)
+
 
 def process_randao(state: BeaconState,
                    block: BaseBeaconBlock,
@@ -73,8 +77,10 @@ def process_randao(state: BeaconState,
     )
 
     randao_mix_index = epoch % config.LATEST_RANDAO_MIXES_LENGTH
+    # FIXME: remove this once get_randao_mix is updated to accept epochs instead of slots
+    slot = SlotNumber(epoch)
     new_randao_mix = bitwise_xor(
-        get_randao_mix(state, epoch, config.LATEST_RANDAO_MIXES_LENGTH),
+        get_randao_mix(state, slot, config.LATEST_RANDAO_MIXES_LENGTH),
         hash_eth2(block.randao_reveal),
     )
 
--- a/eth2/beacon/state_machines/forks/serenity/block_processing.py
+++ b/eth2/beacon/state_machines/forks/serenity/block_processing.py
@@ -19,7 +19,6 @@ from eth2.beacon.state_machines.forks.serenity.block_validation import (
 
 from eth2.beacon.helpers import (
     get_beacon_proposer_index,
-    get_current_epoch,
     get_randao_mix,
 )
 
@@ -69,7 +68,7 @@ def process_randao(state: BeaconState,
     )
     proposer = state.validator_registry[proposer_index]
 
-    epoch = get_current_epoch(state, config.EPOCH_LENGTH)
+    epoch = state.current_epoch(config.EPOCH_LENGTH)
 
     validate_randao_reveal(
         randao_reveal=block.randao_reveal,
--- a/eth2/beacon/state_machines/forks/serenity/block_processing.py
+++ b/eth2/beacon/state_machines/forks/serenity/block_processing.py
@@ -8,7 +8,10 @@ from eth2._utils.numeric import (
 )
 from eth2.beacon._utils.hash import hash_eth2
 
-from eth2.beacon.configs import BeaconConfig
+from eth2.beacon.configs import (
+    BeaconConfig,
+    CommitteeConfig,
+)
 from eth2.beacon.types.states import BeaconState
 from eth2.beacon.types.blocks import BaseBeaconBlock
 from eth2.beacon.types.eth1_data_vote import Eth1DataVote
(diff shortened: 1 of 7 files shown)


tests/eth2/beacon/state_machines/forks/test_serenity_block_processing.py
```diff
@@ -0,0 +1,109 @@
...
+    mock_validator_record,
+)
+
+
+def test_randao_processing(sample_beacon_block_params,
+                           sample_beacon_state_params,
+                           sample_fork_params,
+                           config):
+    proposer_privkey = 1
+    proposer_pubkey = bls.privtopub(proposer_privkey)
```
hwwhww: FYI, because `privtopub` needs one ECMUL calculation, there's already `keymap`, `privkeys`, and `pubkeys` fixtures you can reuse.
```python
proposer_pubkey = pubkeys[0]
proposer_privkey = keymap[proposer_pubkey]  # or privkey[0].
```

tests/eth2/beacon/state_machines/forks/test_serenity_block_processing.py
```diff
@@ -0,0 +1,109 @@
...
+        for index, (updated, original) in enumerate(zip(updated_mixes, original_mixes))
+    )
+
+
+def test_randao_processing_validates_randao_reveal(sample_beacon_block_params,
+                                                   sample_beacon_state_params,
+                                                   sample_fork_params,
+                                                   config):
+    proposer_privkey = 1
+    proposer_pubkey = bls.privtopub(proposer_privkey)
```
hwwhww: ditto

tests/eth2/beacon/state_machines/forks/test_serenity_block_validation.py
```diff
@@ -119,3 +125,43 @@ def test_validate_proposer_signature(
...
+                                  sample_fork_params,
+                                  config):
+    message = epoch.to_bytes(32, byteorder="big")
+    slot = epoch * config.EPOCH_LENGTH
+    fork = Fork(**sample_fork_params)
+    domain = get_domain(fork, slot, SignatureDomain.DOMAIN_RANDAO)
+
+    randao_reveal = bls.sign(message, proposer_privkey, domain)
+
+    expected_proposer_pubkey = bls.privtopub(expected_proposer_privkey)
```
hwwhww: ditto

NIC619:
👍 on making them instance function of `BeaconState` though IIUC they are already implemented?
https://github.com/ethereum/trinity/blob/master/eth2/beacon/types/states.py#L329-L337

jannikluhn:
> +1 on making them instance function of `BeaconState` though IIUC they are already implemented?
> https://github.com/ethereum/trinity/blob/master/eth2/beacon/types/states.py#L329-L337

Didn't know they existed, thanks :+1:, will use those.

It would be nice if we could pass the whole config to `BeaconState` at initialization and use properties instead of getters, but I guess that's not possible without some changes in py-ssz...

hwwhww:
> It would be nice if we could pass the whole config to BeaconState at initialization and use properties instead of getters, but I guess that's not possible without some changes in py-ssz...

Yep...
FYI we have an outline of a "forkable" `BeaconBlock` (i.e., `SerenityBeaconBlock`), but not a forkable `BeaconState` (i.e., `SerenityBeaconState`) yet and I think we can delay the process of making them forkable after we have a stable phase 0 first release. (Now, 20 SSZ objects!!!!)

eth2/beacon/state_machines/forks/serenity/block_processing.py
```diff
@@ -0,0 +1,59 @@
...
+    validate_randao_reveal(
+        randao_reveal=block.randao_reveal,
+        proposer_pubkey=proposer.pubkey,
+        epoch=epoch,
+        fork=state.fork,
+    )
+
+    randao_mix_index = epoch % config.LATEST_RANDAO_MIXES_LENGTH
+    # FIXME: remove this once get_randao_mix is updated to accept epochs instead of slots
+    slot = SlotNumber(epoch)
```
djrtwo: `SlotNumber` just adds slot typing to an int, right?
The actual slot here would be `epoch * EPOCH_LENGTH`.

djrtwo: Or is this just a typing issue for `get_randao_mix`?

jannikluhn: Yeah, so this was just a hack until `get_randao_mix` is updated to accept epochs instead of slots. It is now, so I'll clean this up at the next rebase.

jannikluhn: It's fixed now

hwwhww:
I just realized that, one more thing, in this PR, `process_randao` is not called in `eth2.beacon.state_machines.forks.serenity.state_transitions.SerenityStateTransition.per_block_transition`: https://github.com/ethereum/trinity/blob/53a00906a21114472b5cdd1a9a38dcb4c93a8449/eth2/beacon/state_machines/forks/serenity/state_transitions.py#L108

I imagine you will have to modify `eth2/beacon/tools/builder/validator.py` to set randao reveal for `test_demo.py`. Do you think it makes sense to add in this PR? Or the next one?

PR #253, opened 2019-02-02
Implement per-block processing - eth1 data

jannikluhn:
### What was wrong?

eth1 data processing was missing

### How was it fixed?

Implemented it

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/state_machines/forks/serenity/block_processing.py
+++ b/eth2/beacon/state_machines/forks/serenity/block_processing.py
@@ -1,0 +1,39 @@
+from eth_utils.toolz import (
+    first,
+)
+
+from eth2._utils.tuple import update_tuple_item
+
+from eth2.beacon.types.states import BeaconState
+from eth2.beacon.types.blocks import BaseBeaconBlock
+from eth2.beacon.types.eth1_data_vote import Eth1DataVote
+
+from eth2.beacon.state_machines.configs import BeaconConfig
+
+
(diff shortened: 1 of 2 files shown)


eth2/beacon/state_machines/forks/serenity/block_processing.py
```diff
@@ -0,0 +1,39 @@
...
+
+def process_eth1_data(state: BeaconState,
+                      block: BaseBeaconBlock,
+                      config: BeaconConfig) -> BeaconState:
+    try:
+        vote_index, original_vote = first(
+            (index, eth1_data_vote)
+            for index, eth1_data_vote in enumerate(state.eth1_data_votes)
+            if block.eth1_data == eth1_data_vote.eth1_data
+        )
```
hwwhww: Oh, this `try, except StopIteration, else` pattern looks interesting to me! But since https://github.com/ethereum/eth2.0-specs/pull/561 clarified https://github.com/ethereum/eth2.0-specs/issues/555 and using `first` makes us lose the advantage of that we can break the for-loop earlier once we find `block.eth1_data == eth1_data_vote.eth1_data`. IMHO I think the formal boring for-if-break is better in this case now.

jannikluhn: First will also break early as it's given a generator instead of a list.

hwwhww: Ahh yes!

PR #261, opened 2019-02-06
[POC] peer stats + more aggressive peer discovery

lithp:
Do not merge, PR opened for discussion but the code is atrocious. I'll clean it up after getting some confirmation that these are useful ideas.

### What was wrong?

It takes a very long time for my trinity node to discover peers.

### How was it fixed?

1. `p2p/constants` now includes all the bootnodes which Parity uses
2. `BasePeerPool.connect_to_nodes()` now attempts to connect to up to 10 remotes simultaneously, rather than connecting to nodes in series.
3. `BasePeerPool` now has some memory of nodes. It remembers which nodes have failed with errors such as `WrongGenesisFailure` and doesn't attempt to connect to them again until some time has passed. There's a toml file called `nodedb-eth1` in the data dir which stores entries like this
 
    > ["enode://1356200[snip]cc441a77@147.135.138.85:30303"]
    > status = "do-not-use"
    > until = 1549502737
    > reason = "WrongNetwork"
    > consecutive-error-count = 1
    > network = 246

4. `BasePeerPool` also remembers which nodes we've successfully connected to. On startup it first attempts to re-connect to those:

    > ["enode://70c500e[snip]d0f27777@103.75.212.222:30303"]
    > status = "preferred"
    > when = 1549416256
    > reason = "successful-connection"

### Questions

- I think (1) is a change which I could open a PR for that could be accepted right now?
- I think (2) is also a good idea which could be merged now? Step one to connecting to nodes faster is to attempt to connect to nodes faster.
- How should information on nodes be stored? Right now I'm using a toml file called nodedb-eth1 in the data dir. It's easy for users to inspect but makes storing more complicated stats difficult. leveldb would be easiest to work with but harder for users to inspect, I'm not sure what the right tradeoff is there.
- (4) is a great improvement for me but I'm not sure if it's a good idea, maybe it makes something like an eclipse attack easier?
- I wrote a function called `bounded_wait` which runs a bunch of coroutines but only runs n of them concurrently. Is this already provided by something in asyncio?
- It was difficult to thread nodedb_path, a new attribute of `TrinityConfig`, all the way down to `PeerPool` where it's being used. I'm generally a fan of immutable globals, if something doesn't change during execution then making it a global is easier than passing it exactly where it's needed. How do you feel about moving TrinityConfig in that direction? Then `PeerPool` could do something like: `trinity.config.get_config().nodedb_path`

### Future Work

- This all needs to be tested.
- `PeerInfo` reads and writes the entire file each time it's called. This should be cached and periodically flushed?
- Look at what Parity and Geth do, do they have any equivalent of (3) or (4)?
- When the thread stalls for whatever reason we timeout our peers and drop them. I think there are a couple ways to fix this:
  - Not disconnecting peers after just a single timeout. Giving them a few retries before dropping them.
  - Merging https://github.com/ethereum/trinity/pull/236, so less work is done in this thread and stalls happen less often.
  - Moving networking into a separate process, as @cburgdorf has suggested.
  - Generating a WARNING whenever a coroutine blocks the main thread, to let this problem be noticed sooner.

- In (3) nodes `DiscoveryService` passes to `PeerPool` are filtered out, however it's wasteful for `DiscoveryService` to pass them at all, maybe `DiscoveryService` should also have access to the peer information.
- CancelToken/BaseService has no equivalent of `asyncio.wait` which returns the remaining tasks and doesn't cancel them
- Broadcasting peer candidates happens far more often than nodes are attempted. Not every candidate is attempted before `DiscoveryService` is asked for more peers.

- If we have no peers we'll accept an incoming connection; I think this is a bug.

- I've gotten into a state where I had 0 outbound connections and 3 inbound ones. I think this is bad? When we drop an outbound peer maybe we should also drop inbound peers to make sure they never account for more than half of our connections?

- Failed inbound connections are not added to `PeerInfo` and maybe they should be. (`FullServer.receive_handshake`)

#### Cute Animal Picture


p2p/peer_pool.py
```diff
@@ -274,32 +429,83 @@ def _add_peer(self,
...
+                )
+            except asyncio.futures.CancelledError:
+                for task in tasks:
+                    task.cancel()
+                raise
+            token_wait.cancel()
+            return done, remaining.difference({token_wait})
+
+        # TODO: how might you write tests for this?
+        async def bounded_wait(coros, concurrency: int):
```
lithp: A quick review of the asyncio documentation make it sound like `concurrent.futures.Executor` was very close to this concept but didn't quite support it. Does this already exist / is there a better way to implement it than I'm doing here?

pipermerriam:
>  How should information on nodes be stored?

Maybe overkill but what do people think about a squite database?  That would allow us to do easy querying and should be pretty easy to setup and manage using something like SQLAlchemy or if we don't want the dependency, just the built-in low level [`squite3`](https://docs.python.org/2/library/sqlite3.html) from the standard lib.

pipermerriam:
>  I think (1) is a change which I could open a PR for that could be accepted right now?

:+1: 

> I think (2) is also a good idea which could be merged now? Step one to connecting to nodes faster is to attempt to connect to nodes faster.

:+1:

pipermerriam:
>   (4) is a great improvement for me but I'm not sure if it's a good idea, maybe it makes something like an eclipse attack easier?

I think it's time we started toying with this concept of remembering good nodes so I'm :+1: to get it into the codebase.  As for attack vectors, preventing eclipse seems like it could be accomplished by using some randomness and ensuring that we're choosing peers based on good metrics.

1. Prefer peers that we have successfully connected to
2. Maybe record throughput metrics or longevity of connection.
3. Use a weighted random approach for picking peers to increase the difficulty of things like eclipse attacks.

pipermerriam:
> If we have no peers we'll accept an incoming connection; I think this is a bug.

I don't understand what this means.  Are you saying that we should not allow dial-in connections if we have no dial-out connections?

pipermerriam:
> If we have no peers we'll accept an incoming connection; I think this is a bug.

https://github.com/ethereum/trinity/blob/320162d096056465e2a40c1cb16c43d488488236/trinity/server.py#L263-L265

We have this code to ensure that our connection pool cannot be fully populated by dial-in connections.

lithp:
> I don't understand what this means. Are you saying that we should not allow dial-in connections if we have no dial-out connections?

Yeah, exactly. Maybe it's just a problem with this branch but the line you pointed out doesn't seem to be enough, I have parity node which is always trying to connect to this one. Very soon after starting up I get this:

```
   DEBUG  02-05 17:27:31            FullServer  Receiving handshake from Address(35.233.143.70:udp:50444|tcp:50444)
   DEBUG  02-05 17:27:31               ETHPeer  Finished P2P handshake with <Node(0xe440@35.233.143.70)>, using sub-protocol (eth, 63)
   DEBUG  02-05 17:27:31               ETHPeer  Finished (eth, 63) handshake with <Node(0xe440@35.233.143.70)>
```

And soon after we succeed in connecting to it.

p2p/peer_pool.py
```diff
@@ -68,6 +75,140 @@
...
+        # TODO: implement some kind of exponential backoff?
+        self.path = path
+        self.logger = logger.getChild('PeerInfo')
+        self.logger.debug('Reading from %s', path)
+
+    def _load(self):
+        # TODO: Use aiofiles?
+        try:
+            with open(self.path, 'r') as config_file:
+                return toml.load(config_file)
```
cburgdorf: Just noting that I'm a big fan of using a `.toml` configuration file in general and I foresee things like enabling/disabling plugins through some kind of `toml` file just for the purpose of persisting peer info, I think I agree that this may be better kept in a database as @pipermerriam suggested.

p2p/peer_pool.py
```diff
@@ -88,6 +230,8 @@ def __init__(self,
         self.max_peers = max_peers
         self.context = context
 
+        self.peer_info = PeerInfoPersistence(nodedb_path, self.logger)
```
cburgdorf: I think there should be an abstract `BasePeerInfoPersistence` and then the `BasePeerPool` should get an instance of that injected. That makes the system less coupled and improves testability.

p2p/peer_pool.py
cburgdorf: Not entirely sure but have you looked into `asyncio.Semaphore` and `asyncio.BoundedSemaphore`?

```python
sema = asyncio.BoundedSemaphore(5)
async with sema:
    await thingThatCanOnlyBeCalledWithAConcurrencyOfFive()
```

trinity/config.py
```diff
@@ -432,6 +432,10 @@ def nodekey(self, value: Union[bytes, PrivateKey]) -> None:
                 f"`PrivateKey` instance: got {type(self._nodekey)}"
             )
 
+    @property
+    def nodedb_path(self) -> Path:
```
cburgdorf: I *think* this should rather go onto the `Eth1AppConfig`. Even if we may need the same for the beacon node, there's a good chance that things will be still slightly different (e.g. maybe having multiple node dbs because we need one per shard?)

cburgdorf:
>Then `PeerPool` could do something like: `trinity.config.get_config().nodedb_path`

I'm personally not a fan of that pattern as it introduces stronger coupling and weakens testability imho. A class `BasePeerPool` that depends on a `BasePeerInfoService` is easier testable imho`.

>*it was difficult to thread nodedb_path, a new attribute of `TrinityConfig`, all the way down to `PeerPool` where it's being used.

I agree that this is a problem. However, I think the main problem here is that the current system has the `PeerPool` deeply buried underneath abstractions. I believe once the `PeerPool` is running as a standalone plugin it will be pretty close to the `TrinityConfig`.

In fact, that's why I recently already moved the bootstraping out of the sync out of the server and into a plugin. Because passing the new `--sync mode full` all the way down was equally painful and now that it is a plugin it has easy access to the `TrinityConfig`.

p2p/peer_pool.py
```diff
@@ -134,51 +140,45 @@ def handshake_failure(self, remote: Node, failure: HandshakeFailure):
...
-        if key in contents:
-            error_count = contents[key].get('consecutive-error-count', 0) + 1
-            contents[key].update({
-                'consecutive-error-count': error_count,
-                'until': int(time.time()) + timeout*error_count,
-                'reason': reason,
-                **kwargs
-            })
+        enode = remote.uri()
+        cursor = self.db.execute('SELECT * from bad_nodes WHERE enode = ?', (enode,))
```
pipermerriam: Just noting that it appears that as long as we correctly do parametrization of queries like this we are supposed to be safe from sql injection: https://stackoverflow.com/questions/3220005/escaping-chars-in-python-and-sqlite

carver:
Thoughts before reading the PR:

> * (4) is a great improvement for me but I'm not sure if it's a good idea, maybe it makes something like an eclipse attack easier?

I don't see an eclipse problem, as long as we only store nodes that we discovered out-bound.

> * If we have no peers we'll accept an incoming connection; I think this is a bug.
> * I've gotten into a state where I had 0 outbound connections and 3 inbound ones. I think this is bad? When we drop an outbound peer maybe we should also drop inbound peers to make sure they never account for more than half of our connections?

Agree to both.

> * Look at what Parity and Geth do, do they have any equivalent of (3) or (4)?

Almost certainly, but I don't know the details.

> * In (3) nodes `DiscoveryService` passes to `PeerPool` are filtered out, however it's wasteful for `DiscoveryService` to pass them at all, maybe `DiscoveryService` should also have access to the peer information.

I *think* this is mixing up responsibilities. `PeerPool`s and `Peer`s are subprotocol-aware, but I expect the discovery service to be subprotocol-agnostic. Maybe I'll see something different when I dive into code.

p2p/peer_pool.py
```diff
@@ -68,6 +76,135 @@
 )
 
 
+class PeerInfoPersistence:
+    def __init__(self, path: Path, logger: ExtendedDebugLogger):
+        # TODO: implement some kind of exponential backoff?
```
carver: Yeah, seems reasonable, but tricky when mixing different failure types :+1:

p2p/peer_pool.py
carver: Concretely: it could be helpful to implement an in-memory version of `PeerInfoPersistence` for testing.

p2p/peer_pool.py
```diff
@@ -139,7 +284,7 @@ def __init__(self,
                         )
                         continue
 
-                self.logger.debug2("Received candidates to connect to (%s)", response.candidates)
+                self.logger.debug("Received candidates to connect to (%s)", response.candidates)
```
carver: I assume this is just a temporary change while you're testing, and will go back to debug2 before merge. I think the preferred way to do this would be something like `trinity --log-level p2p.peer_pool.BasePeerPool=debug2` so that you can test it with those messages in the debug log. But I can never remember the syntax, so I do the same thing you did, but try to keep the change local. X)

p2p/peer_pool.py
```diff
@@ -209,8 +354,12 @@ def _add_peer(self,
         Appart from adding it to our list of connected nodes and adding each of our subscriber's
         to the peer, we also add the given messages to our subscriber's queues.
         """
-        self.logger.info('Adding %s to pool', peer)
         self.connected_nodes[peer.remote] = peer
+        self.logger.warning(
+            'Adding %s to pool. Have %s peers', peer, len(self.connected_nodes)
+        )
+        # TODO: I think peer_info should only be called if this was an outgoing connection
+        self.peer_info.connect_success(peer.remote)
```
carver: agree with TODO :+1:

p2p/peer_pool.py
```diff
@@ -209,8 +354,12 @@ def _add_peer(self,
         Appart from adding it to our list of connected nodes and adding each of our subscriber's
         to the peer, we also add the given messages to our subscriber's queues.
         """
-        self.logger.info('Adding %s to pool', peer)
         self.connected_nodes[peer.remote] = peer
+        self.logger.warning(
```
carver: Again, I assume this will be `info` before merge.

p2p/peer_pool.py
```diff
@@ -274,32 +425,82 @@ def _add_peer(self,
...
             # dump the full stacktrace in the debug logs
             self.logger.debug('Got malformed response from %r', remote, exc_info=True)
+        except WrongNetworkFailure as e:
+            self.peer_info.wrong_network_failure(remote, e)
+        except WrongGenesisFailure as e:
+            self.peer_info.wrong_genesis_failure(remote, e)
+        except HandshakeFailure as e:
+            if 'too_many_peers' in repr(e):
+                # TODO: this should become a separate exception?
+                self.peer_info.too_many_peers(remote)
```
carver: Separate exception seems reasonable. Maybe we could even head to something like:
```py
except HandshakeFailure as e:
  self.peer_info.handshake_failure(remote, e)
```
Where `handshake_failure()` would internally decide how to record the failure based on the subclass of `HandshakeFailure`.

p2p/peer_pool.py
```diff
@@ -274,32 +425,82 @@ def _add_peer(self,
...
+            else:
+                self.peer_info.handshake_failure(remote, e)
+                self.logger.debug("Could not complete handshake with %r: %s", remote, repr(e))
+        except UnreachablePeer as e:
+            self.peer_info.unreachable_peer(remote)
         except expected_exceptions as e:
+            # TODO: should TimeoutError also be added to peer_info?
```
carver: This is tricky, same with unreachable per and a few others. It's a non-attributable fault.

So if my wifi router goes down, I start marking all these peers as having timeouts. Maybe that's fine, but it seems like the penalty should be small (or maybe the penalty should be proportionally high if the timeout is not happening with other peers. But that's probably not worth the added complexity)

p2p/peer_pool.py
pipermerriam: Decoupling the storage from the logic might be appropriate.  

- `PeerInfoPersistence(peer_metadata_store=PeerSQLiteStore('/path/to/sqlite.db'))`
- `PeerInfoPersistence(peer_metadata_store=InMemoryPeerStore())`

p2p/peer_pool.py
pipermerriam: @lithp if/when you figure out the CLI way to do this can you write it down somewhere in the developer docs.

> If you want to get output for `<level>` from the debugger `trinity.some_path.Thing` then launch trinity with `trinity --log......????`

p2p/peer_pool.py
pipermerriam: Maybe go even further and have a single entry point for recording failures and do `peer_info.handle_failure(remote, exception)`

p2p/peer_pool.py
pipermerriam: I think @cburgdorf suggestion is probably a good mechanism for this.

trinity/server.py
```diff
@@ -110,14 +112,14 @@ def __init__(self,
 
         # child services
         self.upnp_service = UPnPService(port, token=self.cancel_token)
-        self.peer_pool = self._make_peer_pool()
+        self.peer_pool = self._make_peer_pool(nodedb_path)
```
pipermerriam: It seems appropriate to store this on `self.nodedb_path` so that it doesn't have to get passed around like this.

p2p/peer_pool.py
lithp: Thanks for all the ideas! https://github.com/ethereum/trinity/pull/270 has a `BasePeerInfoPersistance` with noop, disk-backed, and memory-backed instantiations.

I think breaking out the logic into a separate class makes sense. I started down that path by pulling out methods like `_update_node`, hopefully that makes the future refactor easy enough that YAGNI can win out for now.

trinity/config.py
lithp: Sounds good to me :+1: did so in #270!

p2p/peer_pool.py
lithp: toml seems really nice :+1: I was kind of frustrated here by the lack of lists but think it could be a great match for configuring trinity.

p2p/peer_pool.py
lithp: Yeah, definitely. An ideal algorithm would probably take into account the history of which errors which node has given us. I think I'll leave this for a future PR once we decide it's necessary.

pipermerriam:
This was done in other PRs

PR #250, opened 2019-02-01
Remove placeholder

hwwhww:
### What was wrong?

Fix #244

### How was it fixed?
- Update the type of `ValidatorIndex` from `uint24` to `uint64`
- Update `SlashableAttestation` and `verify_slashable_attestation` logic.
    - Modify `verify_slashable_attestation` (spec: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#verify_slashable_attestation) to `validate_slashable_attestation` - it raises `ValidatationError` when it's invalid
    - Implement `verify_bitfield`(spec: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#verify_bitfield) and change it to `validate_bitfield` - it raises `ValidatationError` when it's invalid
- Remove the placeholders
    - Remove `latest_vdf_outputs`, `persistent_committees`, `persistent_committee_reassignments`, and `custody_challenges` from `BeaconState`
    - Remove `custody_reseeds`, `custody_challenges`,  and `custody_responses` from `BaseBeaconBlock`
    - Remove `custody_commitment` from `ValidatorRecord` and `DepositInput`

[//]: # (For important changes, please add a new entry to the release notes file)
[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)

#### Cute Animal Picture

pexels-photo-1245356

PR #262, opened 2019-02-06
Move functions from `helpers.py` to `committee_helpers.py` and `epoch_processing_helpers.py`

hwwhww:
### What was wrong?

Too many functions in `helpers.py`, and the `test_helpers.py` is getting longer and longer.

### How was it fixed?
No logic changes in this PR.

- For the functions for shuffling, find the proposer ... this kind of functions are related to committee shuffling, move them to `committee_helpers.py`.
- For the functions that defined in https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#helpers, move them to `epoch_processing_helpers.py`.

Note the dependencies are:
- `committee_helpers.py` depends on `helper.py`.
- `epoch_processing_helpers` depends on `helper.py` and `committee_helpers.py`.

/cc @pipermerriam @djrtwo @ChihChengLiang @NIC619 @jannikluhn 

Any objection to this reorg? If no, I will continue to reorg `test_helpers.py` too.

Other thoughts on the file path reorg:
1. There may be another new category `attestation_helpers.py` for the helper functions that related to `Attestation` and `SlashableAttestation`, but I'm inclined to leave it after #250 and other in-progressing PR merged.
2. We can move all the `_helpers.py` files into `helpers` folder, and remove the suffix.

#### Cute Animal Picture

pexels-photo-1260803


djrtwo:
reorg lgtm

pipermerriam:
>     2. We can move all the `_helpers.py` files into `helpers` folder, and remove the suffix.

:+1: 

- `helpers.attestations`
- `helpers.committee`

That makes a lot more sense to me than suffixing.

PR #263, opened 2019-02-08
Migrate to new lahja version

cburgdorf:
### What was wrong?

The current lahja version routes every single event through the `main` process which has been proven to be a huge bottleneck and a show stopper for #213 

### How was it fixed?

There's a new lahja version coming up in https://github.com/ethereum/lahja/pull/30 that gets rid of the central relayer topology and instead connects every endpoint with every other endpoint directly (as far as they want to be connected).

To put things visually, currently we have it like this:

old_eventbus

With this change and the new Lahja version, we have it like this:

new_eventbus

Notice that we don't have to have a connection from every Endpoint to every other but it's a convenient default to begin with.

Under the new model, the **main** process is no different from any other `Endpoint` as there is no such thing as a central coordinator role anymore. Strictly speaking, that's not entirely true as there's still some **logical** centralization going on where every endpoint uses the `main` Endpoint as their "bootstrap endpoint" so to speak which they use to learn about all the other endpoints. But that is a purely logical role, from the technical side, all Endpoints are equal.

Notice that this design does also allow adding Endpoints at runtime (e.g. hot plugin replacement) which wasn't supported under the previous model.

### TODO

- [ ] change pinned version once Lahja is released

#### Cute Animal Picture


cburgdorf:
@pipermerriam This is based on the new lahja version (which I still need to get a review on). I've cut a Lahja dev release to be used in this PR and I have also put this PR to work on our server and it seems to be going great so far.

cburgdorf:
@lithp now that you are already deep into the latest Lahja changes you may want to review this PR as well. It's still based on the alpha that doesn't yet incorporate your PR feedback but I guess the changes should be rather small.

cburgdorf:
@lithp This is now rebased on the final Lahja release version. I do have some follow up changes for efficiency gains lined up in my head but I plan to introduce this at a later stage.

trinity/_utils/lahja_helper.py
```diff
@@ -0,0 +1,43 @@
...
+from trinity.constants import (
+    MAIN_EVENTBUS_ENDPOINT,
+)
+from trinity.events import (
+    AvailableEndpointsUpdated,
+    ShutdownRequest,
+)
+
+
+def request_shutdown(event_bus: Endpoint, reason: str) -> None:
```
lithp: That this requires a helper is kind of a code smell for me. It looks like trinity never uses `filter_event_id`, maybe the interface could be changed such that this becomes: `event_bus.broadcast(ShutdownRequest(reason), MAIN_EVENTBUS_ENDPOINT)`

That's not quite right because internally lahja uses `filter_event_id`, but maybe the calls could be broken out into `event_bus.send(endpoint, message)` and `event_bus.reply(event, message)`. I usually prefer two methods which do different things to one method which does different things depending on which parameters are passed into it.

That's not a criticism of this PR though, I might open a PR against lahja later.

trinity/bootstrap.py
```diff
@@ -194,13 +215,23 @@ def main_entry(trinity_boot: BootFn,
     if hasattr(args, 'func'):
         args.func(args, trinity_config)
     else:
+        # We postpone EventBus connection until here because we don't want one in cases where
+        # a plugin just redefines the `trinity` command such as `trintiy fix-unclean-shutdown`
```
lithp: s/trintiy/trinity/

trinity/main.py
```diff
@@ -159,17 +161,37 @@ def kill_trinity_with_reason(reason: str) -> None:
...
+        )
+        endpoint.connect_to_endpoints_blocking(
+            ConnectionConfig.from_name(MAIN_EVENTBUS_ENDPOINT, trinity_config.ipc_dir),
+            # Plugins that run within the networking process broadcast and receive on the
+            # the same endpoint
+            networking_connection_config,
+        )
+        endpoint.broadcast(
+            EventBusConnected(networking_connection_config),
+            BroadcastConfig(filter_endpoint=MAIN_EVENTBUS_ENDPOINT)
```
lithp: nitpick: this line is unnecessary, endpoint is only connected to `MAIN_EVENTBUS_ENDPOINT` so that's the only place broadcasts will go anyway.

further nitpick: I also think this might be easier to read if `endpoint.broadcast` was broken into different functions which each did different things. This call isn't broadcasting, it's talking to just one endpoint! Something like: `endpoint.send_to(MAIN_EVENTBUS_ENDPOINT, EventBusConnected(...))`

trinity/main.py
```diff
@@ -159,17 +161,37 @@ def kill_trinity_with_reason(reason: str) -> None:
...
+        )
+        logger = logging.getLogger("trinity.main.launch_node")
+        endpoint.subscribe(AvailableEndpointsUpdated, connect_to_other_endpoints(logger, endpoint))
```
lithp: I thiink there's a race condition here? I think the idea is that `AvailableEndpointsUpdated` plays two roles, it tells endpoints that new endpoints exist and it also tells new endpoints what all the existing endpoints are. Is there a chance the `AvailableEndpointsUpdated` comes in before this `subscribe()` call?

trinity/main.py
```diff
@@ -159,17 +161,37 @@ def kill_trinity_with_reason(reason: str) -> None:
...
+        endpoint = Endpoint()
         NodeClass = trinity_config.get_app_config(Eth1AppConfig).node_class
         node = NodeClass(endpoint, trinity_config)
         # The `networking` process creates a process pool executor to offload cpu intensive
         # tasks. We should revisit that when we move the sync in its own process
         ensure_global_asyncio_executor()
         loop = node.get_event_loop()
 
-        endpoint.connect_no_wait(loop)
+        networking_connection_config = ConnectionConfig.from_name(
```
lithp: In another thread I mentioned preferring that configuration be moved somewhere a little more global, accessing it could look similar to how `logging.getLogger()` works. This is an example of something which would be made easier, this PR has a few calls to `from_name` which generate what are effectively constants. This is a lot of boilerplate which could be replaced with something like `trinity_config('NETWORKING_BUS')`

trinity/events.py
```diff
@@ -1,22 +1,35 @@
...
 
     def __init__(self, reason: str="") -> None:
         self.reason = reason
+
+
+class EventBusConnected(BaseEvent):
+    """
+    Broadcasted when a new :class:`~lahja.endpoint.Endpoint` connects to the ``main``
+    :class:`~lahja.endpoint.Endpoint`.
+    """
```
lithp: This comment is correct but I think it might be misleading. It makes it sound like `main` sends this message when something new connects to it, but really the other endpoints send it to `main`.

Someone creating a new Endpoint might not realize they should send it. Maybe it could include a note like: "new Endpoints which connect to main should send EventBusConnected(self) to main. main will then send AvailableEndpointsUpdated to every endpoint. This will cause every endpoint to connect to every other endpoint"

trinity/bootstrap.py
```diff
@@ -144,6 +144,27 @@ def main_entry(trinity_boot: BootFn,
     if args.log_levels:
         setup_log_levels(args.log_levels)
 
+    connected_endpoints: Tuple[ConnectionConfig, ...] = tuple()
+
+    def handle_new_endpoints(ev: EventBusConnected) -> None:
+        # The only reason this could happen would be a malicious plugin trying
```
lithp: I'm not sure I understand, won't this happen every time? The only time the main event bus connects to another bus is through the call here. 

Other busses will connect to the main event bus before sending this message, but `is_connected_to` only looks at outbound connections.

trinity/extensibility/plugin_manager.py
```diff
@@ -202,7 +203,7 @@ def prepare(self,
                 plugin,
```
lithp: The comment starting on line 124 references `EventBus`, even though as of this PR that class no longer exists.

trinity/extensibility/plugin_manager.py
```diff
@@ -97,9 +95,12 @@ def create_plugin_context(self,
...
-            # Isolated plugins get an entirely new endpoint to be passed into that new process
+            # Isolated plugins use their own Endpoint that lives in the new process. It is only
+            # created here for API symmetry. Endpoints are pickable *before* they are connected,
+            # which means, this Endpoint will be pickled and transferred into the new process
+            # together with the rest of the `PluginContext`.
             plugin.set_context(PluginContext(
-                self.event_bus.create_endpoint(plugin.name),
+                Endpoint(),
                 boot_info,
             ))
```
lithp: This can now all fit on one line.

trinity/extensibility/plugin_manager.py
```diff
@@ -97,9 +95,12 @@ def create_plugin_context(self,
         uses to enable application wide event-driven communication even across process boundaries.
```
lithp: The line just above this one references the now non-existent `EventBus`

trinity/extensibility/plugin_manager.py
```diff
@@ -97,9 +95,12 @@ def create_plugin_context(self,
...
         """
         if isinstance(plugin, BaseIsolatedPlugin):
-            # Isolated plugins get an entirely new endpoint to be passed into that new process
+            # Isolated plugins use their own Endpoint that lives in the new process. It is only
+            # created here for API symmetry. Endpoints are pickable *before* they are connected,
```
lithp: s/pickable/pickleable/

trinity/extensibility/plugin_manager.py
```diff
@@ -74,8 +73,7 @@ def create_plugin_context(self,
 
 class MainAndIsolatedProcessScope(BaseManagerProcessScope):
 
-    def __init__(self, event_bus: EventBus, main_proc_endpoint: Endpoint) -> None:
-        self.event_bus = event_bus
+    def __init__(self, main_proc_endpoint: Endpoint) -> None:
         self.endpoint = main_proc_endpoint
```
lithp: `self.endpoint` is never used, is that intentional?

trinity/_utils/lahja_helper.py
cburgdorf: Yes, maybe `event_bus.send_to(event: BaseEvent, *endpoint_name:str)` would make sense. That would even allow something like `event_bus(some_event, first_receiver, second_receiver)`.

Happy to see an PR proposing changes!

trinity/extensibility/plugin_manager.py
```diff
@@ -97,9 +95,12 @@ def create_plugin_context(self,
         uses to enable application wide event-driven communication even across process boundaries.
         """
         if isinstance(plugin, BaseIsolatedPlugin):
```
lithp: nitpick: While you're here, I think this check is unnecessary and can be removed. This duplicates the check in ` is_responsible_for_plugin`, which is always called before `create_plugin_context`.

trinity/main.py
cburgdorf: >endpoint is only connected to `MAIN_EVENTBUS_ENDPOINT`

Nope! In fact, with the proposed changes, the `networking` endpoint connects to all other endpoints. See line 194:

`endpoint.subscribe(AvailableEndpointsUpdated, connect_to_other_endpoints(logger, endpoint))`

This is also very important because these connections are *push*  connections and the `networking` endpoint needs to push into all other endpoints so that these can receive events from the peer pool. E.g. the `Ethstats` plugin periodically wants to now the number of connected peers. Also #291 moves the `RequestServer` into its own isolated plugin which then subscribes to all `PeerPoolMessageEvent`s to answer peer requests.

In fact, I plan to further optimize that so that we have more fine grain control over which events go where and can tame down unnecessary event bus traffic.

trinity/main.py
cburgdorf: That's a good point. I swapped it so that we now `subscribe` first. Notice that the following line which now comes *after* the `subscribe` will trigger an `AvailableEndpointsUpdated` event itself, eliminating any risk for a race condition.

```python
endpoint.broadcast(
    EventBusConnected(networking_connection_config),
    BroadcastConfig(filter_endpoint=MAIN_EVENTBUS_ENDPOINT)
)
```

trinity/main.py
cburgdorf: It's certainly more ergonomic in a lot of places. Still I wonder if it would hurt testability or other aspects of the architecture. If you think that would be a good direction, feel free to take a :hocho: at it and we can look at some concrete code :)

trinity/events.py
cburgdorf: Updated :+1:

trinity/plugins/builtin/ethstats/plugin.py
```diff
@@ -77,7 +84,7 @@ def configure_parser(self, arg_parser: ArgumentParser, subparser: _SubParsersAct
             default=10,
         )
 
-    def on_ready(self) -> None:
+    def on_ready(self, manager_eventbus: Endpoint) -> None:
```
lithp: I don't really understand the tradeoffs here, why does this method now have an additional parameter instead of using `self.event_bus`?

trinity/plugins/builtin/tx_pool/plugin.py
```diff
@@ -54,7 +60,7 @@ def configure_parser(self, arg_parser: ArgumentParser, subparser: _SubParsersAct
             help="Enables the Transaction Pool (experimental)",
         )
 
-    def on_ready(self) -> None:
+    def on_ready(self, manager_eventbus: Endpoint) -> None:
```
lithp: Same here, couldn't this use self.event_bus?

trinity/bootstrap.py
cburgdorf: I think my comment is misleading. What I mean is that in a perfect world, there would be no reason to check for `if not main_endpoint.is_connected_to(ev.connection_config.name)` and just simply connect in the handler because we can assume that this is only raised once when an endpoint connects.

However, theoretically, you could write a plugin that emits `EventBusConnected` as often as it wants to which this safe guard then catches to not try to connect again. Let me try to refine my comment to make it more clear.

trinity/plugins/builtin/syncer/plugin.py
```diff
@@ -218,7 +221,7 @@ def extract_strategy_types(self) -> Iterable[Type[BaseSyncStrategy]]:
         for strategy in self.strategies:
             yield type(strategy)
 
-    def on_ready(self) -> None:
+    def on_ready(self, manager_eventbus: Endpoint) -> None:
```
lithp: I'm concerned about adding a new parameter to `on_ready` in part because of methods like this. It uses `self.event_bus` and also has a `manager_eventbus` and it's not at all clear to me why a future developer would want to choose one over another.

trinity/plugins/builtin/fix_unclean_shutdown/plugin.py
```diff
@@ -56,26 +56,14 @@ def fix_unclean_shutdown(self, args: Namespace, trinity_config: TrinityConfig) -
                     'pidfile %s was gone after killing process id %d' % (pidfile, process_id)
```
lithp: Oh, this plugin is nice! I've been deleting the files by hand. I wonder if this could by inlined and run on every boot!

trinity/extensibility/plugin_manager.py
cburgdorf: :+1: went through a view more places that referenced it and reworded it.

trinity/_utils/lahja_helper.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+
+@curry
+def connect_to_other_endpoints(logger: logging.Logger,
+                               event_bus: Endpoint,
+                               ev: AvailableEndpointsUpdated) -> None:
+
+    for connection_config in ev.available_endpoints:
+        if connection_config.name == event_bus.name:
```
lithp: I'm not sure removing this check would be enough to make sure that endpoints connect to themselves, but shouldn't endpoints connect to themselves? I'm thinking of the `SharedProcessScope` plugins, they have no way to talk to each other if they don't listen to their own endpoint.

trinity/plugins/builtin/ethstats/plugin.py
cburgdorf: Previously, endpoints where always created in the `main` process and then passed through to other processes. At the point where `on_ready` is called, a plugin has not yet started. It is in fact, the hook that gives the plugin the chance to decide if it wants to start or not. In the case of a `BaseIsolatedProcess` plugin it means that the process for that plugin does not even exist at this point. So, in order to give the plugin a chance to already raise events such as `ShutdownRequested` we can only raise them on the event bus from the `PluginManager`.

This worked previously because the `Endpoint` for the plugin was already created in the parent process and could already be used to broadcast events even if the actual process for which it was intended was not ready yet.

trinity/extensibility/plugin_manager.py
```diff
@@ -74,8 +73,7 @@ def create_plugin_context(self,
 
 class MainAndIsolatedProcessScope(BaseManagerProcessScope):
 
-    def __init__(self, event_bus: EventBus, main_proc_endpoint: Endpoint) -> None:
-        self.event_bus = event_bus
+    def __init__(self, main_proc_endpoint: Endpoint) -> None:
```
lithp: As far as I can tell this endpoint is never read, it might not need to be passed in.

trinity/bootstrap.py
```diff
@@ -103,12 +104,11 @@ def main_entry(trinity_boot: BootFn,
...
                sub_configs: Iterable[Type[BaseAppConfig]]) -> None:
-    event_bus = EventBus(ctx)
-    main_endpoint = event_bus.create_endpoint(MAIN_EVENTBUS_ENDPOINT)
-    main_endpoint.connect_no_wait()
+
+    main_endpoint = Endpoint()
 
     plugin_manager = setup_plugins(
-        MainAndIsolatedProcessScope(event_bus, main_endpoint),
+        MainAndIsolatedProcessScope(main_endpoint),
```
lithp: Not passing `main_endpoint` here (I think it's never used) would allow `main_endpoint = Endpoint()` to be pushed much further down in this function. It's a bit of a code smell for me that it's created and then used a few times across this function which doesn't fit in a single screen.

trinity/extensibility/plugin.py
```diff
@@ -176,19 +177,25 @@ def set_context(self, context: PluginContext) -> None:
...
         """
         Notify the plugin that it is ready to bootstrap itself. Plugins can rely
         on the :class:`~trinity.extensibility.plugin.PluginContext` to be set
         after this method has been called.
+        The ``manager_eventbus`` refers to the instance of the
+        :class:`~lahja.endpoint.Endpoint` that the
+        :class:`~trinity.extensibility.plugin_manager.PluginManager` uses which may or may not
+        be the same :class:`~lahja.endpoint.Endpoint` as the plugin uses depending on the type
+        of the plugin. The plugin should use this :class:`~lahja.endpoint.Endpoint` instance to
+        listen for events *before* the plugin has started.
```
lithp: This comment says that the passed `manager_eventbus` should be used to listen for events but I'm not sure what that means and I don't think any of the plugins do it!

trinity/plugins/builtin/fix_unclean_shutdown/plugin.py
cburgdorf: >run on every boot

I think that might be dangerous. If you would accidentally run a second instance it's better to just tell the user that there is another instance running instead of killing the IPC files. I think for the time being, its safer to just tell the user to use `trinity fix-unclean-shutdown` (and I think we should inform the user better about that option) and have them take the action.

trinity/plugins/builtin/syncer/plugin.py
cburgdorf: I agree but it is the only way (I think) to give the plugin a chance to broadcast events when the plugins own event bus is not ready yet. See [my comment here](https://github.com/ethereum/trinity/pull/263/files#r259286459)

trinity/plugins/builtin/tx_pool/plugin.py
cburgdorf: This plugin *could* because it is a `BaseAsyncStopPlugin` which runs inside the `networking` process instead or spawning its own isolated process. Notice that this category of plugin will soon go away together will all the added complexity of `BaseManagerProcessScope`.

Currently, we have two `PluginManager` instances. One that runs inside the `networking` process and one that runs inside the `main` process. The `networking` process will soon die and will essentially be transformed into three isolated plugins that all run their own processes (`RequestServer`, `PeerPool`, `Sync`). When that happens, the plugin API is ready for a makeover because things like `BaseAsyncStopPlugin` will just go away. In fact, once #291 lands, the `TxPlugin` could already be rewritten as a `BaseIsolatedProcessPlugin` because it then has access to all the peer pool events.

trinity/_utils/lahja_helper.py
cburgdorf: These plugins all use the `networking` endpoint which already connects to itself. They don't have their own endpoints. But notice that this will all go away when the `networking` process dies (see my [other comment](https://github.com/ethereum/trinity/pull/263/files#r259290481))

trinity/extensibility/plugin_manager.py
cburgdorf: It is ready in https://github.com/ethereum/trinity/blob/fe83a4982d728f3285bcb33924ae54a76a10a559/trinity/extensibility/plugin_manager.py#L159-L166

But notice my other comment about `BaseManagerProcessScope` soon going away anyway. It's a crutch because we have two instances of `PluginManager` and two different categories (three actually) of plugins which will also go away when the `networking` process dies.

trinity/plugins/builtin/ethstats/plugin.py
```diff
@@ -87,14 +94,14 @@ def on_ready(self) -> None:
             self.logger.error(
                 'You must provide ethstats server url using the `--ethstats-server-url`'
             )
-            self.context.shutdown_host("Missing EthStats Server URL")
+            request_shutdown(manager_eventbus, "Missing EthStats Server URL")
```
cburgdorf: @lithp this plugin uses the `manager_eventbus` to shutdown trinity on misconfiguration. It can not use its own event bus because at this point in time, the plugin is just *deciding* if it should start or not. This plugin runs its own process and the eventbus of this plugin is only usable once the plugin process has launched. The `manager_eventbus` gives it a chance to communicate *before* it has started.

trinity/extensibility/plugin.py
cburgdorf: Here's an example where it is used https://github.com/ethereum/trinity/pull/263/files#r259294476

cburgdorf:
Hey @lithp thanks for the review!

>* There are still a ton of references to EventBus and event_bus in the code, which as of this PR are outdated concepts

I scanned through the code base and updated all places that directly mentioned the old `EventBus` class. I think the general *concept* still holds that you have a communication channel that can be used all across Trinity. It is now certainly more depending on the specific topology you decide to build. But since that is all abstracted away behind `BaseIsolatedPlugin` and the hardcoded `networking` process is about to die, I think we can live with the quick updates for now and remake the docs once the bigger migration is completed.

>* Every endpoint which wants to be part of the mesh needs to implement those two messages. This makes sense, but I couldn't find any comments which lay it all out, the code which causes this to happen is scattered around a couple different files and it seems easy to forget to implement these

Well, the good thing is that we don't have to setup endpoints ourselves when we write plugins. It is all abstracted away behind `BaseIsolatedPlugin`. 

>* I think one way to resolve all three complaints at once would be to create a new file and subclass Endpoint.

I'm not sure if `* handle_new_endpoints` fits in their because it currently keeps a reference to `connected_endpoints` which is a variable that directly belongs to the `main` process keeping track of the available endpoints.

But it should work well for `request_shutdown` and `connect_to_other_endpoints`. I'll give it a shot!

>* Now that the architecture has changed I wonder if a larger refactor might be in order

Yes, absolutely. I've mentioned that in some of my inline comments. The `networking` process is going away and transitioning into three isolated plugins (which btw, also helps with maintenance and code reuse of the beacon node client) and once that is done we won't have two `PluginManager` instances anymore and a bunch of complexity around plugin support can be removed.

>I'm excited to merge this and get a big speedup.

So, can you *approve* the PR? I will still try out the `EventBus` subclass idea before merging.

trinity/endpoint.py
```diff
@@ -0,0 +1,86 @@
...
+    MAIN_EVENTBUS_ENDPOINT,
+)
+from trinity.events import (
+    AvailableEndpointsUpdated,
+    EventBusConnected,
+    ShutdownRequest,
+)
+
+
+class TrinityEventBusEndpoint(Endpoint):
```
cburgdorf: @lithp This commit implements your idea of collecting the different APIs on a subclass of `Endpoint`.

lithp:
I get some errors when running trinity. Here's a gist: https://gist.github.com/lithp/a533b06f6a3be4b2b947631a94d2aea9

1. `$ipc_dir/bjson-rpc-api.ipc` does not seem to be deleted when trinity exits, and it also does not seem to be deleted when trinity starts, so every time I start trinity I see this error: `OSError: [Errno 98] Address already in use`. This is despite `lsof` telling me that no processes are using this file. I think this means that JSON-RPC does not work unless you manually remove the ipc file.

    ((a) and(b) are not something I'm saying this PR needs to fix, just noting them here for future reference)

    (a) This was more difficult for me to debug than it should have been because `Address already in use` doesn't give me any indication of *which* address is already in use. Maybe this error should be caught and wrapped with something nicer (b) Trinity continues booting here even though the JSON-RPC endpoint crashed. Maybe an error like this should trigger a `ShutdownRequest`?

2. On shutdown I get a lot of errors which I think already existed, and also some new output: `Attempted to push into main while it isn't running.` this output isn't formatted like the other log lines, there's no "INFO" or timestamp or indication of where it was emitted from. I think that the lahja loggers aren't being setup correctly? And if it's an easy fix I'd like to see this PR shutdown cleanly.

cburgdorf:
>I get some errors when running trinity. Here's a gist:

I think the IPC file sometimes gets left behind when Trinity is stopped early in the booting phase. I think we have the same on `master`. It's annoying but I think it is reproducible on master.

If you run `trinity fix-unclean-shutdown` it should boot again.

The `StopAsyncIteration`  errors are also reproducible on master and I think they appeared after we started using `wait_iter` but I haven't followed up on it. The `wait_iter` is needed but there is something about it not yet playing well with `event_bus.stream(...)`

>1. Maybe this error should be caught and wrapped with something nicer (b) Trinity continues booting here even though the JSON-RPC endpoint crashed. Maybe an error like this should trigger a

Yes, that has bugged me for a while. The problem is that these exceptions happen in another process and I'm not sure we can cleanly communicate them back or even use `request_shutdown` at this point because the IPC connection may be borked already.

>2 I think that the lahja loggers aren't being setup correctly?

I noticed that as well but I'm not sure what would need to be done differently in Lahja here...mmh.

>And if it's an easy fix I'd like to see this PR shutdown cleanly.

Unfortunately, we don't have a clean shutdown without errors in current `master` either so I'd rather like to investigate that separately.

trinity/endpoint.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+
+    # FIXME: For some reason, using the logger of the Endpoint doesn't seem to work here
+    def track_and_propagate_available_endpoints(self, logger: logging.Logger) -> None:
+        """
+        Track new announced endpoints and propagate them across all other existing endpoints.
+        This API should only be called from the ``main`` process.
+        """
+
+        available_endpoints: Tuple[ConnectionConfig, ...] = tuple()
```
lithp: Now that this is part of a class `available_endpoints` and even `handle_new_endpoints` could become (private) members of `TrinityEventBusEndpoint`, which might clean up this code a little bit.

trinity/endpoint.py
```diff
@@ -0,0 +1,86 @@
...
+    def request_shutdown(self, reason: str) -> None:
+        """
+        Perfom a graceful shutdown of Trinity. Can be called from any process.
+        """
+        self.broadcast(
+            ShutdownRequest(reason),
+            BroadcastConfig(filter_endpoint=MAIN_EVENTBUS_ENDPOINT)
+        )
+
+    # FIXME: For some reason, using the logger of the Endpoint doesn't seem to work here
```
lithp: What do you mean doesn't work? This might be related to the log output I saw during shutdown, I think the lahja loggers aren't being correctly setup for some reason.

trinity/endpoint.py
lithp: Nice, thanks for doing this! This was a big refactor refactor.

trinity/endpoint.py
cburgdorf: I think I observed the same thing you observed, the log entries where oddly structured (just like regular `print` calls). Maybe Trinity has some special setup. I'm not too familiar with the logging tbh. If there's anything that needs to be changed in Lahja, it needs a follow up release. But I wouldn't want this to hold back this PR. I'm pretty sure we'll soon have another Lahja release where this could go into (have some changes piled up in my head and you also wanted to make some).

I want to push this forward so that we soon have request server, sync and peer pool as plugins which will also help the beacon client to reuse these things and become functional sooner.

I know it's a bit edgy but once we have these things running I plan to smooth out the edges (including a bigger plugin api makeover).

That said, if you have capacity to follow up on some of these things I'm super happy to have help :)

trinity/main.py
lithp: I mean to say, *when this line is called* it's only connected to the main endpoint. I know that later it connects to every other, but this broadcast is the mechanism by which it connects to them! It's alright to leave this in though, it probably adds some useful clarity.

I think the refactor which created `TrinityEventBusEndpoint` is a big improvement! Maybe this could also be pushed into `TrinityEventBusEndpoint`? All that callers care about is that they're a part of the `EventBus`, `EventBusConnected` seems like an implementation detail from their perspective, and it might be easy to accidentally forget to broadcast it, when `TrinityEventBusEndpoint` "knows" that `EventBusConnected` should always be broadcast.

nitpick: That might get tricky to implement but maybe it'll be easier if `TrinityEventBusEndpoint` is broken up into two classes. It seems like there are two different kinds of endpoint: the main endpoint which connects everything together, and the client endpoints which expect to be connected to all the others.

trinity/plugins/builtin/tx_pool/plugin.py
lithp: Okay, if you think this will be fixed soon then I guess I'm okay with merging it in, but this UX makes me really uncomfortable, this method has both a `manager_eventbus` and a `self.event_bus`, it uses both(!), and there's no indication of when future users should decide to use one or the other.

What do you think about clearing up confusion by doing something like removing `self.event_bus` and passing it into `self.do_start()`?

lithp:
> Unfortunately, we don't have a clean shutdown without errors in current `master` either so I'd rather like to investigate that separately.

I didn't mean to suggest this PR should fix everything! The `Attempted to push into main while it isn't running.` output is new though, and it'd be nice not to merge it if it's an easy fix.

trinity/plugins/builtin/tx_pool/plugin.py
cburgdorf: >What do you think about clearing up confusion by doing something like removing `self.event_bus` and passing it into `self.do_start()`?

This might work well. It's certainly something to take a closer look at when the plugin API gets cleaned up after the `networking` process is gone. I'm just trying to keep the scope of this PR in bounds ;)

cburgdorf:
>The `Attempted to push into main while it isn't running.` output is new though

Yes, we produce this warning in `Lahja` when a message is tried to get pushed into an endpoint that has already stopped (or not yet started). So, the warning in itself might be fine. It's just the oddly placed logger output that is annoying imho. Maybe we should allow passing a `logging.Logger` instance to the constructor of `Endpoint` so that we can pass in a trinity logger?

trinity/extensibility/__init__.py
```diff
@@ -8,6 +8,7 @@
     BasePlugin,
     DebugPlugin,
     PluginContext,
+    PluginStatus,
     TrinityBootInfo,
 )
 from trinity.extensibility.plugin_manager import (  # noqa: F401
```
lithp: Maybe this file should also include the exceptions from `trinity.extensibility.exceptions`? Users who try to use `PluginStatus` might want to catch `InvalidPluginStatus`

trinity/plugins/builtin/fix_unclean_shutdown/plugin.py
cburgdorf: Here's another idea that I think might be more safe! What if run something like `remove_dangling_ipcs` at the very last step of `kill_trinity_gracefully`.

https://github.com/ethereum/trinity/blob/fe83a4982d728f3285bcb33924ae54a76a10a559/trinity/bootstrap.py#L254

Reasoning is: At this step, we are shutting down a running Trinity instance so removing IPC files at this point is the natural thing for Trinity to do (instead of at launch where it might just be that we *forgot* we have another instance running).

Also, at this point we have given each process enough time to properly clean up, so if there are dangling IPC files at this point, it should be fair to remove them.

Thoughts?

lithp: Yeah, I wonder why the files are being left behind, it might be that the `InterruptException` is causing some `shutdown()` call to not happen.

I'm down for trying again to delete them during shutdown!

An alternative might be to run slightly different logic on startup: Trinity could check if another instance of trinity is running and if it isn't then deleting the socket is safe.

cburgdorf: Yep, I do wonder the same. The `fix-unclean-shutdown` plugin is a crutch that we introduced a long time ago when we only had the db process and the networking process. The more we go into a multiprocess architecture and especially now that Lahja starts managing ipc files as well, the more we are exposed to this annoying bug.

>I'm down for trying again to delete them during shutdown!

I just [pushed a commit](https://github.com/ethereum/trinity/pull/263/commits/5f478ed40dfab4928bf2fe80dbe5d7caa1efd567) that does that and it works really well.

>An alternative might be to run slightly different logic on startup: Trinity could check if another instance of trinity is running and if it isn't then deleting the socket is safe.

Like introducing a `.lock` file for Trinity? One problematic piece here is also that we *do support* multiple instances when the other instance is just doing a `trinity attach` :sweat_smile: 
But yes, if done correctly that could be even better than the cleanup on shutdown because there may still be cases where we leave IPC paths behind when the cleanup on shutdown wasn't executed for some reason.

PR #266, opened 2019-02-08
[WIP] Try a short pruning history

carver:
### What was wrong?

During #236 I realized that if we keep a short history, maybe the naive approach is fast enough. That would be a lot less code.

### How was it fixed?

Cut history from 10k nodes to 10. Still need to test live syncing performance (my current office seems to block connections).  @veox if you feel like testing the performance on this branch, that would be sweet. I'll probably pick it back up on Monday morning.

#### Cute Animal Picture

### DIFF
--- a/trinity/_utils/datastructures.py
+++ b/trinity/_utils/datastructures.py
@@ -436,7 +436,7 @@ class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
     _dependency_of: StaticMethod[Callable[[TTask], TTaskID]]
 
     # by default, how long should the integrator wait before pruning?
-    _default_max_depth = 10000  # not sure how to pick a good default here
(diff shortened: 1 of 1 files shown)


lithp:
I did some profiling of this branch, the raw data is in this gist: https://gist.github.com/lithp/f3d922b0a01ef481f76a835f5019360b

This makes `_mark_complete` go from taking 50% cpu time to 11% cpu time. FastChainBodySyncer used import 20 blocks per second and now imports 36.7 blocks per second.

carver:
> This makes `_mark_complete` go from taking 50% cpu time to 11% cpu time. FastChainBodySyncer used import 20 blocks per second and now imports 36.7 blocks per second.

Awesome, thx. 11% still sounds like too much, so I'm happy to just close and continue with #236

PR #268, opened 2019-02-12
Add more bootnodes from Geth & Parity

lithp:
Does this meet the bar for a release notes entry?

### What was wrong?

Trinity has a hard time finding peers, in part because the boot nodes it uses don't seem to be as useful as the ones Parity uses.

### How was it fixed?

All the bootnodes with Geth and Parity use were added to the list of mainnet bootnodes.

#### Cute Animal Picture

happy-turtle


lithp:
Pushed an update which adds `noqa` to all the new lines. I wish there was a way to disable an error for a given block.

Issue #240, opened 2019-01-30
Minimal viable trinity-beacon node


cburgdorf:
## Abstract

This issue tries to document the current state of the `trintiy-beacon` command and lays out steps to turn it into an actual minimal viable beacon node. This was sparked by @ChihChengLiang starting this [document](https://notes.ethereum.org/znnddsMNTLikDHVrkVosSQ?both) which means it is also trying to address questions outlined there.

### Background: How does the beacon node implementation fit into the code base?

In general, the relation between the beacon node and the existing Ethereum 1 client is very loose at this point.

1. The `trinity-beacon` command is an entirely different command from `trinity` and invokes its own bootstrapping routine at `trinity/main_beacon.py`. Bootstrapping code does share some common utility functions.

2. The long term vision for Trinity is to be mostly a composition of plugins with many essential parts just being written as builtin, pre-loaded core plugins (while other use defined plugins can be loaded at runtime)

We can see this already working out with the collection of current plugins.

https://github.com/ethereum/trinity/blob/7436d9c3fdef5fbeef5d7a1123c8ed2ecd16e033/trinity/plugins/registry.py#L49-L68

The plugin infrastructure also takes care of spawning isolated processes for each plugin that derives from `BaseIsolatedPlugin` and enables inter-process communication through a central `EventBus`.

3. Trinity provides a `TrinityConfig` that should be agnostic of anything that is specific to the Ethereum1 or beacon node. These specific configs need to go into the `BeaconAppConfig` which will be available in the context of the beacon node (similar to the `Eth1AppConfig` being available in the context of the Ethereum 1 node).

4. Currently, the beacon node uses all base plugins (**not** the Ethereum 1 specific plugins) which includes things like the *Peer Discovery* as well as the *JSON-RPC API*.

### Does the beacon node need a `Node` class?

Probably not. The `Node` / `FullNode` / `LightNode` classes are currently used in the Ethereum 1 node to run what is currently known as the `networking` process. This process does currently host the `PeerPool`, the server that answers request from other peers as well as the syncing code.

**We are actively migrating away from that architecture** towards an architecture where the responsibilities are further broken down into different processes that are assembled through plugins.

Ideally, we'll soon have something that looks like this:

1. Peer Pool Plugin (own process) -> manages peers, relays incoming messages on the `EventBus`

[...]

PR #270, opened 2019-02-12
Persist Peer Information between runs

lithp:
A cleaned up version of #261

### What was wrong?

Trinity wastes time trying to connect to nodes which it already knows won't work (they're on the wrong network or using the wrong genesis).

### How was it fixed?

Trinity now keeps a sqlite database with information on peers. It records failures and doesn't attempt to reconnect to useless peers until a timeout has passed.

This PR doesn't do things like remember good nodes or exponential backoff, but it hopefully lays a foundation which will make those changes easy in the future.

Todo:
- [x] aiosqlite? -- decide not to do this because the library seems rather immature and I don't expect sqlite will block the thread for very long, even though it's ugly that it blocks the thread
- [x] When logging use localtime? That or explicitly mention that the logs use UTC
- [x] Add an entry to release notes

#### Cute Animal Picture

cute-baby-elephants-50-5902071911e92__700


p2p/peer_pool.py
```diff
@@ -79,11 +83,16 @@ def __init__(self,
                  privkey: datatypes.PrivateKey,
                  context: BasePeerContext,
                  max_peers: int = DEFAULT_MAX_PEERS,
+                 peer_info: BasePeerInfoPersistance = None,
```
cburgdorf: Wouldn't it make sense to put `NoopPeerInfoPersistance()` as a default instead of `None`? And then removing line 92-95.

lithp: Yeah, good point. My instinct is to avoid passing objects as parameter defaults but `NoopPeerInfoPersistance` has no state so it's safe here.

p2p/persistance.py
```diff
@@ -0,0 +1,206 @@
...
+    def __init__(self, logger: ExtendedDebugLogger) -> None:
+        if logger:
+            logger = logger.getChild('PeerInfo')
+        self.logger = logger
+
+    @abc.abstractmethod
+    def record_failure(self, remote: Node, timeout: int, reason: str) -> None:
+        raise NotImplemented()
+
+    @abc.abstractmethod
```
cburgdorf: Even though we in general prefer using the module prefix, for `@abstractmethod` our style guide says we prefer it *without* the `abc` prefix for brevity / aesthetics / readability

p2p/persistance.py
```diff
@@ -0,0 +1,206 @@
...
+        super().__init__(None)
+
+    def record_failure(self, remote: Node, timeout: int, reason: str) -> None:
+        pass
+
+    def can_connect_to(self, remote: Node) -> bool:
+        return True
+
+
+class ClosedException(Exception):
```
cburgdorf: I think this should derive from `BaseP2PError` and probably even move into `p2p/exceptions.py`

ralexstokes:
looks nice :) this is minor but shall i suggest s/persistance/persistence/?

p2p/persistance.py
```diff
@@ -0,0 +1,206 @@
...
+    # methods of SQLPeerInfoPersistance cannot be called after it's been closed
+    pass
+
+
+T = TypeVar('T', bound=Callable[..., Any])
+
+
+def must_be_open(func: T) -> T:
```
cburgdorf: This looks like we are throwing away valuable type information here ( `T` has a bound of `Callable[..., Any]`).

Did you check if `mypy` still thinks `can_connect_to` will return `bool` after this decorator is applied on? You can test it if you put in `reveal_type(can_connect_to)` and run `mypy`.

p2p/persistance.py
```diff
@@ -0,0 +1,206 @@
...
+                'skipping %s, it failed because "%s" and is not usable until %s',
+                remote, row['reason'], row['until']
+            )
+            return False
+
+        return True
+
+    def _fetch_node(self, remote: Node) -> sqlite3.Row:
+        enode = remote.uri()
+        cursor = self.db.execute('SELECT * from bad_nodes WHERE enode = ?', (enode,))
```
cburgdorf: I haven't worked with SQL for ages and never did so in Python. I think @pipermerriam already commented on SQL injections in the previous PR so I assume it is double checked that this is the right way to do it and we are on the safe side?

p2p/persistance.py
```diff
@@ -0,0 +1,206 @@
...
+                '''
+                UPDATE bad_nodes
+                SET until = ?, reason = ?, error_count = ?
+                WHERE enode = ?
+                ''',
+                (time_to_str(until), reason, error_count, enode),
+            )
+
+    def close(self) -> None:
+        self.closed = True
```
cburgdorf: Shouldn't the `self.closed = True` go after the `self.db.close()`?

p2p/persistance.py
```diff
@@ -0,0 +1,206 @@
...
+            self.logger.error("malformed. try deleting %s", self.path)
+            raise Exception("cannot read nodedb: version %s is unsupported", version)
+
+        # schema_version exists and is 1, this database has already been initialized!
+        return True
+
+
+class MemoryPeerInfoPersistance(SQLPeerInfoPersistance):
+    def __init__(self, logger: ExtendedDebugLogger) -> None:
+        super().__init__(Path(":memory:"), logger)
```
cburgdorf: Sweet :heart:

p2p/persistance.py
```diff
@@ -0,0 +1,206 @@
...
+    return time.isoformat(timespec='seconds')
+
+
+def str_to_time(as_str: str) -> datetime.datetime:
+    # use datetime.datetime.fromisoformat once support for 3.6 is dropped
+    return datetime.datetime.strptime(as_str, "%Y-%m-%dT%H:%M:%S")
+
+
+class BasePeerInfoPersistance(abc.ABC):
+    def __init__(self, logger: ExtendedDebugLogger) -> None:
```
cburgdorf: rather than monkeypatching `current_time` in the tests I personally would prefer the `__init__` to take a parameter with `time_provider: Callable[[], datetime.datetime] = datetime.datetime.utcnow`. So with a sensible default value yet over-writable for testing without reaching for `monkeypatch`.

trinity/nodes/full.py
```diff
@@ -23,6 +24,9 @@ def __init__(self, event_bus: Endpoint, trinity_config: TrinityConfig) -> None:
         self._node_port = trinity_config.port
         self._max_peers = trinity_config.max_peers
 
+        app_config = trinity_config.get_app_config(Eth1AppConfig)
```
cburgdorf: Just noting that without further checking this would crash if the `Node` class gets reused for the beacon chain which however I don't *think* will happen (as described in #240) Just noting.

p2p/persistance.py
```diff
@@ -0,0 +1,206 @@
...
+def must_be_open(func: T) -> T:
+    @functools.wraps(func)
+    def run(self: 'SQLPeerInfoPersistance', *args: Any, **kwargs: Any) -> Any:
+        if self.closed:
+            raise ClosedException()
+        return func(self, *args, **kwargs)
+    return cast(T, run)
+
+
+class SQLPeerInfoPersistance(BasePeerInfoPersistance):
```
pipermerriam: *nitpick*

Lets call this `SQLitePeerInfo...` so that if we end up with some other SQL based backend we don't collide on naming.

p2p/persistance.py
```diff
@@ -0,0 +1,206 @@
...
+def time_to_str(time: datetime.datetime) -> str:
+    return time.isoformat(timespec='seconds')
+
+
+def str_to_time(as_str: str) -> datetime.datetime:
+    # use datetime.datetime.fromisoformat once support for 3.6 is dropped
+    return datetime.datetime.strptime(as_str, "%Y-%m-%dT%H:%M:%S")
+
+
+class BasePeerInfoPersistance(abc.ABC):
```
pipermerriam: Maybe `PeerInfoBackend` would carry more semantic meaning since the memory one isn't persistent? Or `PeerInfoDB` but I'm hesitant about using the term `DB` since we already have multiple other *database* things in these codebases.

p2p/exceptions.py
```diff
@@ -28,7 +28,20 @@ class HandshakeFailure(BaseP2PError):
     """
     Raised when the protocol handshake was unsuccessful.
     """
-    pass
+    # when this error occurs don't try to talk to the remote for x seconds
+    timeout = 10
+
+
+class WrongNetworkFailure(HandshakeFailure):
+    timeout = 60 * 60 * 24  # one day
```
pipermerriam: Maybe instead of storing these timeout on the exceptions we should keep this information separate in a mapping from `Exception -> timeout`

p2p/persistance.py
pipermerriam: Yes, using the `?` in the query string and placing the values as positional arguments does keep us safe from sql injection.

p2p/persistance.py
pipermerriam: Yeah, in memory SQLite is pretty sexy.  I recall loving this for testing Django apps.

p2p/persistance.py
pipermerriam: I'm minorly :-1: on this.  Time is one of the few things that I'm *ok* with mocking as I haven't come across a better mechanism.  I would advocate removing the `current_time` function in favor of just using `datetime.datetime.utcnow()` and then in our tests, using the `monkeypatch` fixture to swap that function out for one that returns the desired value.

tests/p2p/test_persistance.py
```diff
@@ -0,0 +1,150 @@
...
+    db.close()
+
+    # version 2 is not supported!
+    with pytest.raises(Exception):
+        SQLPeerInfoPersistance(dbpath, logger)
+
+
+def random_node():
+    address = kademlia.Address('127.0.0.1', 30303)
+    return kademlia.Node(generate_privkey(), address)
```
pipermerriam: Just noting this instpired me to create https://github.com/ethereum/trinity/issues/279

p2p/peer_pool.py
lithp: > How about just not having a default at this level and taking care of it at the level above this where it is instantiated.

My idea was that `BasePeerPool` would always have a peer_info without needing to put null checks everywhere. This also makes the interface a little simpler, if the caller doesn't care about persistence then it can implicitly not-care, vs explicitly passing in a Noop object.

p2p/persistance.py
lithp: Huh, I'm willing to do it but why do you think that error is more appropriate? This exception doesn't have anything to do with peers, it signals an invalid state which should never be reached. I thought about making it an `assert False` but think this stacktrace is a little easier to read.

p2p/persistance.py
lithp: > Did you check if `mypy` still thinks `can_connect_to` will return `bool` after this decorator is applied on?

I didn't but this decorator is called on more methods than `can_connect_to`, and I wanted it to work no matter what the signature of the wrapped method is. You're right that it's not very pretty though :cry: I found this ticket which tracks mypy support for this: https://github.com/python/mypy/issues/3157

p2p/persistance.py
lithp: It makes me sad that this is the correct way but it's indeed the correct way. [named params](https://www.python.org/dev/peps/pep-0249/#paramstyle) are nicer but this sqlite3 library doesn't support them.

p2p/persistance.py
pipermerriam: If seeing this is truly something a user should **never** encounter then downgrading to `Exception` is acceptable.  If it's helpful to have a custom exception but we don't want to expose it to users, then maybe add that explanation to the docstring.  If a user might encounter this in some situation (which it sounds like they shouldn't) then it should be in `p2p.exceptions` and inherit from the base exception.

Do as you see fit :smile_cat:

p2p/persistance.py
lithp: Actually, after looking at [the sqlite3 library](https://github.com/python/cpython/blob/62be74290aca26d16f3f55ece7ff6dad14e60e8d/Modules/_sqlite/statement.c#L206), params [are passed directly to sqlite](https://github.com/python/cpython/blob/62be74290aca26d16f3f55ece7ff6dad14e60e8d/Modules/_sqlite/statement.c#L276), which means named parameters [are possible](https://www.sqlite.org/lang_expr.html#varparam).

That means the code could look like this:

```python
self.db.execute(
    '''
    INSERT INTO bad_nodes (enode, until, reason, error_count)
    VALUES (:enode, :until, :reason, :error_count)
    ''',
    {
       ':enode': enode,
       ':until': time_to_str(until),
       ':reason': reason,
       ':error_count': error_count,
    }
)
```

But I think that's a level of boilerplate which the queries aren't yet big enough to justify

p2p/persistance.py
lithp: Sure, I can move it down :+1:

I guess the concern is that `db.close()` might have a reference to `self` and expect it to be open?

cburgdorf: It almost certainly won't practically matter but it *feels* more coherent to only flip the `bool` once the actual `action` has happened.

p2p/persistance.py
cburgdorf: Fine either way :+1:

p2p/peer_pool.py
```diff
@@ -246,8 +255,9 @@ def _add_peer(self,
         if remote in self.connected_nodes:
             self.logger.debug("Skipping %s; already connected to it", remote)
             return None
+        if not self.peer_info.can_connect_to(remote):
```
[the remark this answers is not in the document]
cburgdorf: I had pretty much the same comment drafted and than deleted it after tricking me into believing that's just my non-native language sensor being wrong. So, :+1: on a rename.

p2p/persistance.py
lithp: My idea on this is that there are two big categories of references, there's the "immutable" world of static references (immutable globals such as + or, in this file, sqlite3), and also the mutable world of objects which programs typically concern themselves with.

Dependency injection works well for the latter, there are legitimately different implementations of `BasePeerInfoPersistance` floating around and almost all classes won't care exactly which one they use, so by taking responsibility for constructing `BasePeerInfoPersistance` away from those classes we can make the code more flexible and much easier to test.

I think DI makes less sense for the former. Imports are fixed at the beginning of runtime, there aren't going to be multiple implementations of `datetime.datetime.utcnow()` floating around so you lose no flexibility by explicitly referencing it. And if you squint, python is already doing DI! This module is asking for an implementation of `datetime.datetime` and python is providing one. `test_persistence` isn't monkeypatching so much as changing which implementation of `utcnow()` this module receives.

p2p/persistance.py
lithp: makes sense

p2p/persistance.py
lithp: How about just `PeerInfo`?

p2p/exceptions.py
lithp: I was unsure on this. On the one hand it doesn't seem to be `BasePeerInfoPersistence`'s responsibility, but on the other hand you're right, this isn't *really* a property of the failure itself.

I could make a map in `p2p/persistence.py`, I didn't know if that was too much boilerplate for one PR

p2p/persistance.py
```diff
@@ -0,0 +1,206 @@
...
+        self.db.close()
+        self.db = None
+
+    @must_be_open
+    def setup_schema(self) -> None:
+        try:
+            if self._schema_already_created():
+                return
+        except Exception:
+            self.db.close()
```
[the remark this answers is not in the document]
lithp: oops, this was not intentional :sweat_smile:

p2p/peer_pool.py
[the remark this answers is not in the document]
lithp: Yeah, `should_connect_to` sounds like an improvement :+1:

p2p/persistance.py
```diff
@@ -0,0 +1,206 @@
...
+def str_to_time(as_str: str) -> datetime.datetime:
+    # use datetime.datetime.fromisoformat once support for 3.6 is dropped
+    return datetime.datetime.strptime(as_str, "%Y-%m-%dT%H:%M:%S")
+
+
+class BasePeerInfoPersistance(abc.ABC):
+    def __init__(self, logger: ExtendedDebugLogger) -> None:
+        if logger:
+            logger = logger.getChild('PeerInfo')
+        self.logger = logger
```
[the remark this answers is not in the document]
lithp: Yeah, good point, I broke my own rule by injecting the logger :sweat_smile:

lithp:
> looks nice :) this is minor but shall i suggest s/persistance/persistence/?

heh, oops, good point @ralexstokes

p2p/persistance.py
lithp: Makes sense, applied!

p2p/exceptions.py
lithp: Okay, I've moved them into a map, that's probably a big enough change to be worth looking at again: https://github.com/ethereum/trinity/pull/270/commits/41b8b5adf93f817535e173a0b3bd66d862ffb24b

p2p/persistance.py
```diff
@@ -0,0 +1,206 @@
...
+
+        # a schema_version table already exists, get the version
+        cur = self.db.execute("SELECT version FROM schema_version")
+        rows = cur.fetchall()
+        if len(rows) != 1:
+            self.logger.error(
+                "malformed nodedb. try deleting %s. (got rows: %s)",
+                self.path, rows,
+            )
+            raise Exception(
```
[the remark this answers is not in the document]
lithp: Added a `BadDatabaseError`

lithp:
> It doesn't have to be addressed in this PR but I want to give consideration to how this interacts with _preferred_ peers and the concept of _reserved_ connection slots. We'll _eventually_ need to ensure this mechanism doesn't blacklist a peer in a [vipnode](https://vipnode.org/) type setup where peers have paid for reserved slots.

I *believe* that they could be orthogonal notions. Even if you don't go all the way and have separate `PeerPool`s this seems like something which the `PeerPool` will have knowledge about. In that case, it's not a stretch to imagine the `PeerPool` skipping [the `PeerInfo` check](https://github.com/ethereum/trinity/pull/270/files#diff-72f857530e6a00a2c27fe44d64fdad90R256) and connecting even if we've previously seen an error from the node.

The alternative might be telling `PeerInfo` about which nodes are preferred/reserved? It can then choose to return `should_connect_to() -> True` for those nodes, but that feels like mixing the concern of remembering node quality with the concern of picking which nodes to connect to, so I think my final answer is that reserved peers are higher up in the stack than what `PeerInfo` cares about, historical data on peers, making it independent of this PR.

> Note that this is probably best solved via a multi-peer pool setup since it's probably not sane to try to support a single peer pool having multiple sets of rules for peer connections since those rules would very likely end of conflicting with each other.

Yeah, multiple peer pools with different rules seems like the simplest change. I'm trying to imagine alternatives: Maybe peers each have a `PeerManager`? Currently other parts of the code-base liberally call `peer.disconnect(...)`. A `PeerPool` for reserved instances might deal with that by noticing disconnects and immediately reconnecting. If a `PeerManager` was injected into each Peer and called inside of `peer.disconnect` it might contain logic like cancelling the disconnect (`ReservedPeerManager`) or waiting until 5 timeouts have passed before finally giving up on a peer (`ETHPeerManager`).

lithp:
Okay, I *believe* that I've addressed all the feedback. I finished my TODOs and squashed, if one of you could please give it a final lookover I'll address anything you notice and merge!

p2p/persistance.py
cburgdorf: >`test_persistence` isn't monkeypatching so much as changing which implementation of `utcnow()` this module receives.

Yes, which is why I lean on the side: Hey, if we do encounter a case where we want to swap the implementation of `utcnow()` against something else e.g. a controllable `fakeutcnow()` then, DI is my tool of choice.

And for a class where time plays an important role I wouldn't see it as uncommon to just depend on a `time_provider: Callable[[], int]`. But anyway, this can become quite a bit philosophical and I just happen to have build DI-heavy systems for many years so I'm just biased into that direction.

It's actually refreshing to look at other solutions so I'm not opposed to the existing code :+1:

Issue #146, opened 2019-01-10
Reorg helper functions


hwwhww:
### What is wrong?

There are many helper functions in `eth/beacon/helpers.py`. IMO Several reasons make it worth to be sorted out:
1. Long `tests/beacon/test_helpers.py` for the tests and *the helpers of the tests*. It’s not easy to divide the code sections.
2. More recent cleanup ideas on the spec side: https://github.com/ethereum/eth2.0-specs/issues/380 
3. Per https://github.com/ethereum/py-evm/pull/1684#discussion_r244824902, @pipermerriam: I'm always a little nervous with indexing into iterables naively like this.  I understand/recognize that the previous `get_beacon_proposer_index` provides some level of guarantee that the `beacon_proposer_index` is indeed present and accurate in the `validator_registry` iterable. What if this was abstracted away on the `BeaconState` class as something like `state.get_beacon_proposer(epoch_length)`:
    ```python
    proposer = state.get_beacon_proposer(epoch_length)
    proposer_pubkey = proposer.pubkey
    ```
	* I hesitate due to the pros and cons if we want to make all of them to instance methods:
		* Pros:
			* Intuitively,  `state.get_beacon_proposer(epoch_length)` makes a lot sense for the user.
		* Cons:
			* As @djrtwo mentioned (https://github.com/ethereum/eth2.0-specs/issues/380#issuecomment-450745575), pure functions are easier to tests. Also IMO it has clear boundaries of the essential inputs.
			* Not sure if we will apply it on the spec side; if not, our codebase will be quite different from the spec and makes it harder to sync with the spec.

### How can it be fixed
Some ideas, **TBD**:
1. `BeaconState`
	1. `get_block_root(latest_block_roots: Sequence[Hash32], current_slot: SlotNumber, slot: SlotNumber) -> Hash32` -> `BeaconState.get_block_root(slot: SlotNumber) -> Hash32`.
	2. `get_shard_committees_at_slot(state: 'BeaconState', slot: SlotNumber, epoch_length: int) -> Tuple[ShardCommittee]` -> `BeaconState.get_shard_committees_at_slot(slot: SlotNumber, epoch_length: int) -> Tuple[ShardCommittee]`
[...]

PR #271, opened 2019-02-12
Constants arguments clean-up and shuffling lookahead

### LINKED ISSUE #258: Update helpers for shuffling lookahead and boundary fixes
### What is wrong?

Spec PR: https://github.com/ethereum/eth2.0-specs/pull/520

* Update `get_crosslink_committees_at_slot` to be able to get potential committees for slots from the next epoch. add `registry_change` param to get next epoch committees with and without a registry change at the end of the current epoch
* Add `get_next_epoch_committee_count`
* Fill all positions of `latest_index_roots = genesis_active_index_root`
* Change bounds of `get_active_index_root` to always store up to `ENTRY_EXIT_DELAY` epochs in the future

### How can it be fixed

Pending on #256

### LINKED ISSUE #520: Tracking issue for peer blacklisting
### What is wrong?

In working on #24 I have found that we need to first implement better blacklisting.  Without this, we end up with a pool filled with lots of mediocre or bad peers.  Loose evaluation looks like the following end up being problems.

- Malicious peers:  Sends provably invalid or wrong responses.
- Useless peers: 
    - Predominantly incomplete responses.
    - Announced HEAD is too far behind our local HEAD
- Poorly connected peers: always times out
- Lazy peers: does not respond to queries they *should* be able to respond to.
    - Announced HEAD block number `N`, doesn't respond to query for block/header at height <= `N`.
    - Announced HEAD block number `N`, doesn't respond to `GetNodeData` for state root from block within `range(N - X, N)` blocks of 100, where `X` is somewhere in the range 64 <-> 128

### How can it be fixed

I'm looking at doing the following.

**Firs**t: New internal plugin that runs the blacklist tracker as a service over the event bus.  This is needed because we need to communicate with this service from many different parts of the codebase and it's not really viable to pass around a reference to the tracker to all of these different code locations.

Then, I plan to implement some version of the following rules.

#### Malicious peers

Immediate disconnect and put on blacklist for some appropriate amount of time.

#### Useless peers

Mark them as useless and blacklist for appropriate amount of time.  Disconnect oldest/random useless peer if pool is full.

#### Poorly connected, Lazy

Use token bucket based approach.  If token bucket is empty, disconnect and blacklist for appropriate amount of time.

hwwhww:
### What was wrong?
1. Refactor: clean up constants arguments by adding `CommitteeConfig`
2. Fix #258

#### 1. Refactor: clean up constants arguments by adding `CommitteeConfig`
Sorry for that this PR became more about refactoring than just add a new feature.
When I tried to implement #258, I have to revamp `get_crosslink_committees_at_slot` helper; and with this new feature, I have to pass **four more constants into `get_crosslink_committees_at_slot` and all the function that calls it**.
That's a stronger signal of that some 💩 code here that we need to fix... 
Per https://github.com/ethereum/trinity/issues/146#issuecomment-460908609, I mentioned passing the shuffling-required constants into `ShufflingConfig`, but after some recent changes, it looks better to have a subset of `BeaconConfig` - `CommitteeConfig` to mitigate it.

```python
class CommitteeConfig:
    def __init__(self, config: BeaconConfig):
        # Basic
        self.genesis_epoch = config.GENESIS_EPOCH
        self.shard_count = config.SHARD_COUNT
        self.epoch_length = config.EPOCH_LENGTH
        self.target_committee_size = config.TARGET_COMMITTEE_SIZE

        # For seed
        self.seed_lookahead = config.SEED_LOOKAHEAD
        self.entry_exit_delay = config.ENTRY_EXIT_DELAY
        self.latest_index_roots_length = config.LATEST_INDEX_ROOTS_LENGTH
        self.latest_randao_mixes_length = config.LATEST_INDEX_ROOTS_LENGTH
```

1. It's initialized with `BeaconConfig` in `StateMachine` layer.
2. Also, `BeaconConfig` has been moved from `eth2/beacon/state_machines/configs.py` → `eth2/beacon/configs.py`.
3. Pros: less ugly arguments passing.
4. Cons: lose some awareness of what are the configuration parameters we are testing. 

While this PR is WIP, @pipermerriam @jannikluhn @ralexstokes @djrtwo , any thoughts on this refactoring?

#### 2. Shuffling lookahead

- Update `get_crosslink_committees_at_slot` for shuffling lookahead function `get_next_epoch_committee_assignment ` for the validator to check if they are assigned in the incoming epoch and the context they need.
- Use case: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/validator/0_beacon-chain-validator.md#responsibility-lookahead
- `get_next_epoch_committee_assignment` returns `NoCommitteeAssignment` when no assignment for the given `validator_index`.

#### Cute Animal Picture

aquarium-ocean-fish-jumping-dolphin-animal-water-906176

### DIFF
--- a/eth2/beacon/committee_helpers.py
+++ b/eth2/beacon/committee_helpers.py
@@ -49,7 +49,7 @@ if TYPE_CHECKING:
 
 class CommitteeConfig:
     def __init__(self, config):
-        self.genesis_epoch = config.genesis_epoch
+        self.genesis_epoch = config.GENESIS_EPOCH
         self.shard_count = config.SHARD_COUNT
         self.epoch_length = config.EPOCH_LENGTH
         self.target_committee_size = config.TARGET_COMMITTEE_SIZE
@@ -173,7 +173,6 @@ def get_crosslink_committees_at_slot(
     """
     Return the list of ``(committee, shard)`` tuples for the ``slot``.
     """
-
     genesis_epoch = committee_config.genesis_epoch
     epoch_length = committee_config.epoch_length
     target_committee_size = committee_config.target_committee_size
@@ -239,8 +238,13 @@ def get_crosslink_committees_at_slot(
                 latest_index_roots_length=latest_index_roots_length,
                 latest_randao_mixes_length=latest_randao_mixes_length,
             )
-            shuffling_start_shard = (state.current_epoch_start_shard + current_committees_per_epoch) % SHARD_COUNT
-        elif epochs_since_last_registry_update > 1 and is_power_of_two(epochs_since_last_registry_update):
+            shuffling_start_shard = (
+                state.current_epoch_start_shard + current_committees_per_epoch
+            ) % shard_count
+        elif (
+            epochs_since_last_registry_update > 1 and
+            is_power_of_two(epochs_since_last_registry_update)
+        ):
             seed = generate_seed(
                 state=state,
                 epoch=next_epoch,
@@ -280,20 +284,14 @@ def get_crosslink_committees_at_slot(
 
 def get_beacon_proposer_index(state: 'BeaconState',
                               slot: SlotNumber,
-                              genesis_epoch: EpochNumber,
-                              epoch_length: int,
-                              target_committee_size: int,
-                              shard_count: int) -> ValidatorIndex:
+                              committee_config: CommitteeConfig) -> ValidatorIndex:
     """
     Return the beacon proposer index for the ``slot``.
     """
     crosslink_committees_at_slot = get_crosslink_committees_at_slot(
         state=state,
         slot=slot,
-        genesis_epoch=genesis_epoch,
-        epoch_length=epoch_length,
-        target_committee_size=target_committee_size,
-        shard_count=shard_count,
+        committee_config=committee_config,
     )
     try:
         first_crosslink_committee = crosslink_committees_at_slot[0]
@@ -322,10 +320,7 @@ def _get_committee_for_shard(
 def get_attestation_participants(state: 'BeaconState',
                                  attestation_data: 'AttestationData',
                                  bitfield: Bitfield,
-                                 genesis_epoch: EpochNumber,
-                                 epoch_length: int,
-                                 target_committee_size: int,
-                                 shard_count: int) -> Iterable[ValidatorIndex]:
+                                 committee_config: CommitteeConfig) -> Iterable[ValidatorIndex]:
     """
     Return the participant indices at for the ``attestation_data`` and ``bitfield``.
     """
@@ -333,10 +328,7 @@ def get_attestation_participants(state: 'BeaconState',
     crosslink_committees = get_crosslink_committees_at_slot(
         state=state,
         slot=attestation_data.slot,
-        genesis_epoch=genesis_epoch,
-        epoch_length=epoch_length,
-        target_committee_size=target_committee_size,
-        shard_count=shard_count,
+        committee_config=committee_config,
     )
 
     if attestation_data.shard not in set([shard for _, shard in crosslink_committees]):
--- a/eth2/beacon/committee_helpers.py
+++ b/eth2/beacon/committee_helpers.py
@@ -163,14 +163,15 @@ def get_crosslink_committees_at_slot(
     """
     Return the list of ``(committee, shard)`` tuples for the ``slot``.
     """
-    genesis_epoch = committee_config.genesis_epoch
-    epoch_length = committee_config.epoch_length
-    target_committee_size = committee_config.target_committee_size
-    shard_count = committee_config.shard_count
-    seed_lookahead = committee_config.seed_lookahead
-    entry_exit_delay = committee_config.entry_exit_delay
-    latest_index_roots_length = committee_config.latest_index_roots_length
-    latest_randao_mixes_length = committee_config.latest_randao_mixes_length
+    genesis_epoch = committee_config.GENESIS_EPOCH
+    shard_count = committee_config.SHARD_COUNT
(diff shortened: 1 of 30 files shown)


eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -269,11 +233,45 @@ def process_validator_registry(state: BeaconState,
 #
 # Final updates
 #
+def _update_latest_index_roots(state: BeaconState,
+                               config: BeaconConfig) -> BeaconState:
```
NIC619: What do you think about passing in `CommitteeConfig` instead of `BeaconConfig` here?

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -269,11 +233,45 @@ def process_validator_registry(state: BeaconState,
...
+    """
+    Return the BeaconState with updated `latest_index_roots`.
+    """
+    next_epoch = state.next_epoch(config.EPOCH_LENGTH)
+
+    # TODO: chanege to hash_tree_root
+    active_validator_indices = get_active_validator_indices(
+        state.validator_registry,
+        next_epoch,
```
NIC619: Should this be `next_epoch + config.ENTRY_EXIT_DELAY`?

eth2/beacon/committee_helpers.py
```diff
@@ -134,30 +139,53 @@ def get_current_epoch_committee_count(
...
     Return the list of ``(committee, shard)`` tuples for the ``slot``.
     """
+    genesis_epoch = committee_config.genesis_epoch
+    epoch_length = committee_config.epoch_length
+    target_committee_size = committee_config.target_committee_size
+    shard_count = committee_config.shard_count
+    seed_lookahead = committee_config.seed_lookahead
+    entry_exit_delay = committee_config.entry_exit_delay
+    latest_index_roots_length = committee_config.latest_index_roots_length
+    latest_randao_mixes_length = committee_config.latest_randao_mixes_length
```
NIC619: The config variables were loaded into capital letter variables in other place if I'm not mistaken. Should we conform to either one of them?

hwwhww: Yes, that's the plan if people think this refactoring is the right way to go! 😄

hwwhww: I'm thinking about, usually, use `genesis_epoch = committee_config.GENESIS_EPOCH` becasue there are still some other constants which are not in part of `CommitteeConfig` in some functions.

eth2/beacon/on_startup.py
```diff
@@ -161,18 +160,17 @@ def get_initial_beacon_state(*,
...
                 index.to_bytes(32, 'big')
                 for index in active_validator_indices
             ]
         )
     )
-    latest_index_roots = update_tuple_item(
-        state.latest_index_roots,
-        genesis_epoch % latest_index_roots_length,
-        index_root,
+    latest_index_roots = tuple(
```
pipermerriam: equivalent to either of the following if you like them more

```
# using itertools.repeat
tuple(itertools.repeat(genesis_active_index_root, latest_index_roots_length))

# using tuple multiplication
(genesis_active_index_root,) * latest_index_roots_length
```

eth2/beacon/validator_status_helpers.py
```diff
@@ -111,9 +111,11 @@ def _settle_penality_to_validator_and_whistleblower(
     state.validator_balances[index] -= whistleblower_reward
     validator.penalized_epoch = slot_to_epoch(state.slot)
     """
+    EPOCH_LENGTH = committee_config.EPOCH_LENGTH
```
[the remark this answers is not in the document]
hwwhww: Yeah, I prefer to use the lower letter for consistency too.

I remember that @ChihChengLiang introduced capital letters in some other functions, and now there are two different style in the codebase. Can we unify it?

eth2/beacon/committee_helpers.py
```diff
@@ -134,30 +139,54 @@ def get_current_epoch_committee_count(
...
+    target_committee_size = committee_config.TARGET_COMMITTEE_SIZE
+
+    seed_lookahead = committee_config.SEED_LOOKAHEAD
+    entry_exit_delay = committee_config.ENTRY_EXIT_DELAY
+    latest_index_roots_length = committee_config.LATEST_INDEX_ROOTS_LENGTH
+    latest_randao_mixes_length = committee_config.LATEST_RANDAO_MIXES_LENGTH
 
     epoch = slot_to_epoch(slot, epoch_length)
     current_epoch = state.current_epoch(epoch_length)
+    previous_epoch = state.previous_epoch(epoch_length, genesis_epoch)
```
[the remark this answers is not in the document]
hwwhww: Isn't `genesis_time` the unix timestamp when chain starts, while `genesis_slot` and (`genesis_epoch`) is the offset pivot to `slot`? Or do you mean, we can find a way to set  `genesis_slot` as a configuration inside `BeaconState`, not as SSZ field?
On the other hand, it seems it can be solved by https://github.com/ethereum/eth2.0-specs/issues/504.

eth2/beacon/validator_status_helpers.py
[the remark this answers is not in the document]
ChihChengLiang: Will fix them to lower cases in my existing PRs

eth2/beacon/committee_helpers.py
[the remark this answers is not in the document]
djrtwo: ahh. I see. Leave as is for now. This might change when we finish the great integer debate https://github.com/ethereum/eth2.0-specs/issues/626

PR #273, opened 2019-02-12
Implement process rewards and penalties

### LINKED ISSUE #239: Implement per-epoch processing - rewards and penalties
### What is wrong?
[Per-epoch processing - Rewards and penalties](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#rewards-and-penalties)

### How can it be fixed

NIC619:
### What was wrong?
Fix #239 .

### How was it fixed?
- [x] Rename `get_attesting_validator_indices` to `get_shard_block_root_attester_indices` and `get_total_attesting_balance` to `get_shard_block_root_total_attesting_balance`
- [x] Add helper and test: `get_previous_epoch_justified_attestations`
- [x] Add helper: `get_base_reward`
- [x] Add helper and test: `get_previous_epoch_head_attestations`
- [x] Add helper and test: `get_inclusion_info_map`
- [x] Add `process_rewards_and_penalties`:
    - [x] justification and finalization
    - [x] attestation inclusion
    - [x] crosslinks
    - [x] test
        - [x] test if validator/proposer is successfully rewarded/penalized
- [x] Update according to https://github.com/ethereum/eth2.0-specs/pull/597

#### Cute Animal Picture


eth2/beacon/epoch_processing_helpers.py
```diff
@@ -161,3 +202,51 @@ def get_winning_root(
...
+        index: ValidatorIndex,
+        previous_total_balance: Gwei,
+        base_reward_quotient: int,
+        max_deposit_amount: Gwei) -> Gwei:
+    _base_reward_quotient = int(math.sqrt(previous_total_balance)) // base_reward_quotient
+    return Gwei(
+        get_effective_balance(
+            state.validator_balances,
+            index,
+            max_deposit_amount
```
hwwhww: nitpick
```suggestion
            max_deposit_amount,
```

eth2/beacon/epoch_processing_helpers.py
```diff
@@ -161,3 +202,51 @@ def get_winning_root(
...
+
+
+def get_inclusion_info_map(
+        *,
+        state: 'BeaconState',
+        attestations: Sequence[PendingAttestationRecord],
+        genesis_epoch: EpochNumber,
+        epoch_length: int,
+        target_committee_size: int,
+        shard_count: int) -> Tuple[Dict[ValidatorIndex, SlotNumber], Dict[ValidatorIndex, int]]:
```
hwwhww: Oh maps. Request for docstring.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -132,6 +146,388 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
     return state
 
 
+def process_rewards_and_penalties(state: BeaconState, config: BeaconConfig) -> BeaconState:
```
hwwhww: It's a long function and we should make it easier to read and test. At least 3 sections:
1. `_process_rewards_and_penalties_for_finality` (justification and finalization)
2. `_process_rewards_and_penalties_for_attestation_inclusion`
3. `_process_rewards_and_penalties_for_crosslinks`

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -132,6 +146,388 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+                    for a in previous_epoch_justified_attestations
+                ]
+            )
+        )
+    else:
+        previous_epoch_justified_attester_indices = set()
+
+    # Compute previous epoch boundary attester indices and the total balance they account for
+    # for later use.
+    previous_epoch_boundary_attestations = [
```
hwwhww: `set()` it?

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -132,6 +146,388 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+                        config.EPOCH_LENGTH,
+                        config.TARGET_COMMITTEE_SIZE,
+                        config.SHARD_COUNT,
+                    )
+                    for a in previous_epoch_justified_attestations
+                ]
+            )
+        )
+    else:
+        previous_epoch_justified_attester_indices = set()
```
hwwhww: It seems we can extract the logic into

`get_attester_indices_from_attesttion(state, config, get_attestations_fn)`

and call

```python
previous_epoch_attester_indices = get_attester_indices_from_attesttion(state, config, get_previous_epoch_attestations)
previous_epoch_justified_attester_indices = get_attester_indices_from_attesttion(state, config, get_previous_epoch_justified_attestations)
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -132,6 +146,388 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+                        config.EPOCH_LENGTH,
+                        config.TARGET_COMMITTEE_SIZE,
+                        config.SHARD_COUNT,
+                    )
+                    for a in previous_epoch_head_attestations
+                ]
+            )
+        )
+    else:
+        previous_epoch_head_attester_indices = set()
```
hwwhww: It seems we can extract the logic into

`get_attester_indices_from_attesttions(state, config, get_attestations_fn)`

and call

```python
previous_epoch_attester_indices = get_attester_indices_from_attesttions(state, config, get_previous_epoch_attestations)
previous_epoch_justified_attester_indices = get_attester_indices_from_attesttions(state, config, get_previous_epoch_justified_attestations)
previous_epoch_head_attester_indices = get_attester_indices_from_attesttion(state, config, get_previous_epoch_head_attestations)
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -132,6 +146,388 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+            config.GENESIS_EPOCH,
+            config.EPOCH_LENGTH,
+            config.TARGET_COMMITTEE_SIZE,
+            config.SHARD_COUNT,
+        )
+        reward = base_reward_map[index] // config.INCLUDER_REWARD_QUOTIENT
+        reward_received_map[proposer_index] += reward
+
+    # 3. Process rewards and penalties for crosslinks
+    prev_epoch_start_slot = get_epoch_start_slot(
```
hwwhww: `prev_epoch_start_slot` -> `previous_epoch_start_slot` for consistency.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -132,6 +146,388 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+        state.previous_epoch(config.EPOCH_LENGTH, config.GENESIS_EPOCH),
+        config.EPOCH_LENGTH,
+    )
+    cur_epoch_start_slot = get_epoch_start_slot(
```
hwwhww: `cur_epoch_start_slot` -> `current_epoch_start_slot` for consistency.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -132,6 +146,388 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+            config.TARGET_COMMITTEE_SIZE,
+            config.SHARD_COUNT,
+        )
+        for crosslink_committee, shard in crosslink_committees_at_slot:
+            filtered_attestations = _filter_attestations_by_shard(
+                previous_epoch_attestations + current_epoch_attestations,
+                shard,
+            )
+            try:
+                winning_root, total_attesting_balance = get_winning_root(
```
hwwhww: It seems crosslink section and rewarding section are overlapping from the spec and it's weird that we'd need to call `get_winning_root` + for-loop again in per-epoch processing. (Right after `process_crosslinks`). :(

Does it make sense if we make `process_crosslinks` return `attesting_validator_indices: Dict[Tuple[crosslink_committee, shard], Tuple[ValidatorIndex, ...]]`, and pass it to `process_rewards_and_penalties`? It may make the state transition boundary dirty though.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
NIC619: I suppose the chance two identical attestations are included is rare? Or do you mean `tuple()`? (Indeed it should be tuple for consistency with `previous_epoch_justified_attestations` and `previous_epoch_head_attestations`.)

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
NIC619: I'm not strongly against it but personally leaning toward keeping the structure clean and separate.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -132,6 +146,363 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+        config: BeaconConfig,
+        prev_epoch_active_validator_indices: Set[ValidatorIndex],
+        previous_total_balance: Gwei,
+        previous_epoch_attester_indices: Iterable[ValidatorIndex],
+        previous_epoch_boundary_attester_indices: Iterable[ValidatorIndex],
+        previous_epoch_head_attester_indices: Iterable[ValidatorIndex],
+        inclusion_distance_map: Dict[ValidatorIndex, int],
+        effective_balance_map: Dict[ValidatorIndex, Gwei],
+        base_reward_map: Dict[ValidatorIndex, Gwei],
+        reward_received_map: Dict[ValidatorIndex, Gwei]) -> None:
```
hwwhww: I believe it should return something. `reward_received_map`?

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -132,6 +146,363 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+            reward_received_map[index] = Gwei(reward_received_map[index] - penalty)
+
+
+def _process_rewards_and_penalties_for_attestation_inclusion(
+        state: BeaconState,
+        config: BeaconConfig,
+        previous_epoch_attester_indices: Iterable[ValidatorIndex],
+        inclusion_slot_map: Dict[ValidatorIndex, SlotNumber],
+        base_reward_map: Dict[ValidatorIndex, Gwei],
+        reward_received_map: Dict[ValidatorIndex, Gwei]) -> None:
```
hwwhww: I believe it should return something. `reward_received_map`?

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -132,6 +146,363 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+        reward_received_map[proposer_index] = Gwei(reward_received_map[proposer_index] + reward)
+
+
+def _process_rewards_and_penalties_for_crosslinks(
+        state: BeaconState,
+        config: BeaconConfig,
+        previous_epoch_attestations: Iterable[PendingAttestationRecord],
+        effective_balance_map: Dict[ValidatorIndex, Gwei],
+        base_reward_map: Dict[ValidatorIndex, Gwei],
+        reward_received_map: Dict[ValidatorIndex, Gwei]) -> None:
```
hwwhww: I believe it should return something. `reward_received_map`?

tests/eth2/beacon/state_machines/forks/test_serenity_epoch_processing.py
```diff
@@ -447,3 +456,418 @@ def test_process_crosslinks(
...
+            5,
+            10,
+            3, 79,  # epochs_since_finality > 4
+            5,
+            100,
+            4,
+        ),
+    ]
+)
+def test_process_rewards_and_penalties(
```
hwwhww: Now we can test these functions seperately. 🖖 
1. `test_process_rewards_and_penalties_for_finality`
2. `test_process_rewards_and_penalties_for_attestation_inclusion`
3. `test_process_rewards_and_penalties_for_crosslinks`

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
hwwhww: I just realized that you are using mutable `reward_received_map`. I think we should remove the mutability between function passing.

hwwhww:
Try to enable it! 😆 
https://github.com/ethereum/trinity/blob/ba908037461b0c2d4495be46d4a5c0af07d483c1/eth2/beacon/state_machines/forks/serenity/state_transitions.py#L118

NIC619:
@hwwhww You are right. I forgot to enable it!

> After you enable it, you can add assertions to check that if the rewards are given to the block proposers in `test_demo.py`.

IMO the three tests for `process_rewards_and_penalties` already handles checking rewards/penalties been properly applied. And it seems to me that even if we do check, the best we can do is to check if the proposer's balance increases after epoch processing(under the premise that no penalties applied on the proposer).

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -132,6 +146,369 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+def process_rewards_and_penalties(state: BeaconState, config: BeaconConfig) -> BeaconState:
+    # Compute previous epoch active validator indices and the total balance they account for
+    # for later use.
+    prev_epoch_active_validator_indices = set(
+        get_active_validator_indices(
+            state.validator_registry,
+            state.previous_epoch(config.EPOCH_LENGTH, config.GENESIS_EPOCH)
+        )
+    )
+    previous_total_balance: Gwei = Gwei(
```
ChihChengLiang: Can use `get_total_balance` in eth2/beacon/helpers for this line now.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -132,6 +146,369 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+        if a.data.epoch_boundary_root == get_block_root(
+            state,
+            get_epoch_start_slot(
+                state.previous_epoch(config.EPOCH_LENGTH, config.GENESIS_EPOCH),
+                config.EPOCH_LENGTH,
+            ),
+            config.LATEST_BLOCK_ROOTS_LENGTH,
+        )
+    ]
+    previous_epoch_boundary_attester_indices = get_attester_indices_from_attesttion(
```
ChihChengLiang: Can use `get_epoch_boundary_attester_indices` now

NIC619:
Rebased and only needs an approval now :).

eth2/beacon/epoch_processing_helpers.py
```diff
@@ -223,3 +207,52 @@ def get_epoch_boundary_attesting_balances(
...
+
+
+def get_base_reward(
+        *,
+        state: 'BeaconState',
+        index: ValidatorIndex,
+        previous_total_balance: Gwei,
+        base_reward_quotient: int,
+        max_deposit_amount: Gwei) -> Gwei:
+    _base_reward_quotient = integer_squareroot(previous_total_balance) // base_reward_quotient
```
ChihChengLiang: `_base_reward_quotient` is calculated every time for each validator. Should calculate it outside once and then pass it to the function.

eth2/beacon/epoch_processing_helpers.py
```diff
@@ -223,3 +207,52 @@ def get_epoch_boundary_attesting_balances(
...
+            attestation.data,
+            attestation.aggregation_bitfield,
+            committee_config,
+        )
+        for index in participant_indices:
+            inclusion_slot = inclusion_slot_map.get(index)
+            if inclusion_slot is None:
+                inclusion_slot_map[index] = attestation.slot_included
+                inclusion_distance_map[index] = attestation.slot_included - attestation.data.slot
+            elif attestation.slot_included < inclusion_slot:
```
ChihChengLiang: This `elif` branch does exactly the same thing. How about: `if inclusion_slot is None or attestation.slot_included < inclusion_slot:`

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +276,368 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+                previous_epoch_attesting_balance //
+                previous_total_balance
+            )
+            reward_received_map[index] = Gwei(reward_received_map[index] + reward)
+        # Punish active validators not in `previous_epoch_attester_indices`
+        excluded_active_validators_indices = prev_epoch_active_validator_indices.difference(
+            set(previous_epoch_attester_indices))
+        for index in excluded_active_validators_indices:
+            penalty = base_reward_map[index]
+            reward_received_map[index] = Gwei(reward_received_map[index] - penalty)
```
pipermerriam: This `reward_received_map` is mutated across multiple loops, a pattern which I'm inclined to push back against.  It makes reasoning about what is going on here very difficult.  A brief read suggests this high level description of what happens:

- *first* populate it with the the base data:
  - `reward_received_map = old_reward_received_map.copy()`
- *second* apply for current reward
  - `reward_received_map[index] = Gwei(reward_received_map[index] + reward)`
- *third* apply penalties
  - `reward_received_map[index] = Gwei(reward_received_map[index] - penalty)`

And so on...

My recommendation would be to keep this data separate and then aggregate it at the latest possible point when you need the aggregate values.  This becomes something like:

- *first* populate it with the the base data:
  - `historical_rewards_map = old_reward_received_map.copy()`
- *second* apply for current reward
  - `reward_map = {index: _compute_reward(....) for index in ...}
- *third* apply penalties
  - `penalty_map = {index: _compute_penalty(...) for index in ...}

And so on.  Then at the point where you need aggregate data you can do the proper combining of these values.  Also, this may be a good way to reduce the overall size of this function body by splitting each of these individual components out into their own smaller functions, each of which could produce the individual maps which would *also* allow you to reduce the mutation by using things like the `to_dict` decorator and writing them as generator functions.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +276,368 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+                )
+            except NoWinningRootError:
+                # No winning shard block root found for this shard.
+                # Hence no one is counted as attesting validator.
+                attesting_validator_indices: Iterable[ValidatorIndex] = set()
+            else:
+                attesting_validator_indices = get_attester_indices_from_attesttion(
+                    state=state,
+                    committee_config=CommitteeConfig(config),
+                    attestations=[
```
pipermerriam: should probably be tuple instead of list

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +276,368 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+                        a
+                        for a in filtered_attestations
+                        if a.data.shard == shard and a.data.shard_block_root == winning_root
+                    ],
+                )
+            total_balance = sum(
+                effective_balance_map[i]
```
pipermerriam: I recognize this convention has spread across the codebase a bit so probably not a change that should happen as part of this PR but maybe a meta cleanup if there's agreement.  Historical naming convention would have us naming this `effective_balances`.  We have tended to not include metadata about the data type in our names.

Worth pointing out there are some counter examples such as the `state_dict` in py-evm.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +276,368 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+        state,
+        config,
+        previous_epoch_attester_indices,
+        inclusion_slot_map,
+        base_reward_map,
+        reward_received_map,
+    )
+
+    # 3. Process rewards and penalties for crosslinks
+    reward_received_map = _process_rewards_and_penalties_for_crosslinks(
```
pipermerriam: meta concern that applies more broadly than this specific PR.

I :heart: the architecture you all have chosen, focusing on composing smaller utility functions like this to create the larger functionality.  However, now I'm seeing a second order issue show up, which is these long function bodies with lots of function calls, many of which take 5+ positional arguments, many of which are repeated across each call like `state`, `config`, `base_reward_map` etc.

I touched on some ideas for how we might accomplish this here: https://github.com/ethereum/trinity/pull/228#discussion_r251901840 though they don't exactly map cleanly to this case.

I don't have a solid solution but I think this is probably worth someone taking a day to try some things out and see what patterns we can figure out to reduce some of the visual overhead in these function bodies.

pipermerriam: Wrote this to capture this in an easier to find place and ideally make it more actionable.

https://github.com/ethereum/trinity/issues/298

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +276,368 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+        previous_epoch_boundary_attester_indices,
+        previous_epoch_head_attester_indices,
+        inclusion_distance_map,
+        effective_balance_map,
+        base_reward_map,
+        reward_received_map,
+    )
+
+    # 2. Process rewards and penalties for attestation inclusion
+    reward_received_map = _process_rewards_and_penalties_for_attestation_inclusion(
```
pipermerriam: The repeated overwriting of this value suggest that this could be rewritten using something like `pipe` or, if we want to preserve the visual separation to choose different but semantically descriptive names for each of the three times we write to this variable.

eth2/beacon/epoch_processing_helpers.py
```diff
@@ -223,3 +207,52 @@ def get_epoch_boundary_attesting_balances(
...
+        committee_config: CommitteeConfig) -> Tuple[Dict[ValidatorIndex, SlotNumber], Dict[ValidatorIndex, int]]:  # noqa: E501
+    """
+    Return two maps. One with ``ValidatorIndex`` -> ``inclusion_slot`` and the other with
+    ``ValidatorIndex`` -> ``inclusion_distance``.
+
+    ``attestation.inclusion_slot`` is the slot during which the pending attestation is included.
+    ``inclusion_distance = attestation.inclusion_slot - attestation.data.slot``
+    """
+    inclusion_slot_map: Dict[ValidatorIndex, SlotNumber] = {}
+    inclusion_distance_map: Dict[ValidatorIndex, int] = {}
```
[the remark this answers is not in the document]
NIC619: This does look cleaner but I'm not sure where to put them. They are not the core data structure and hence to me they don't seem significant enough to have a `inclusion_info.py` created just for them.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +276,368 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
     return state
 
 
+def _process_rewards_and_penalties_for_finality(
```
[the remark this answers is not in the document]
NIC619: I will put it in my TODO list and come back around when I have a chance.

eth2/beacon/tools/builder/validator.py
```diff
@@ -311,6 +313,19 @@ def create_mock_signed_attestations_at_slot(
...
     )
+    beacon_block_root = get_block_root(
+        state,
+        attestation_slot,
+        config.LATEST_BLOCK_ROOTS_LENGTH,
+    )
+    epoch_boundary_root = get_block_root(
+        state,
+        get_epoch_start_slot(
+            max(slot_to_epoch(attestation_slot, config.SLOTS_PER_EPOCH), 0),
```
hwwhww: Why we need `max(..., 0)` here?  How would `slot_to_epoch(attestation_slot, config.SLOTS_PER_EPOCH)` be less than `0`?

hwwhww: It seems there is still bug of #294, where we need to:
1. Pass the right head `state`
2. Get `epoch_boundary_root` by head `state.slot`, not by `attestation_slot`.

NIC619: Revert the changes to `test_demo.py` and `create_mock_signed_attestations_at_slot`.
Wait for #294 to fix it.

eth2/beacon/epoch_processing_helpers.py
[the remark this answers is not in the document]
pipermerriam: It can be inline above this function or you could relegate it to something like `eth2.beacon.datastructures` module.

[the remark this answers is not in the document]
NIC619: Add it to `eth2.beacon.datastructures` module but make a little naming adjustment, `InclusionData` -> `InclusionInfo`, `included_at_slot` -> `inclusion_slot`.
- `InclusionInfo` because IMO `InclusionData` would make people think it is part of `PendingAttestationData` or something equivalent that should go into the state.
- `inclusion_slot` to conform to the spec naming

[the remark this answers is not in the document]
hwwhww: 1. Can we also make a new data structure for `old_rewards_received: Dict[ValidatorIndex, Gwei]`/`base_rewards: Dict[ValidatorIndex, Gwei]`? It's everywhere in this PR.
2. @pipermerriam  Any style guide recommended that using `datastructures` rather than `data_structures`? And what do you think about renaming `datastructures` to `data_models`?

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
hwwhww: +1

NIC619: Piped!

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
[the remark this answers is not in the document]
NIC619: @pipermerriam @hwwhww Add `_apply_rewards_and_penalties` to replace repeating procedure of applying rewards/penalties. Let me if it doesn't make sense or if there's still room for improvement.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +285,500 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+                base_rewards[index] *
+                config.MIN_ATTESTATION_INCLUSION_DELAY //
+                inclusion_infos[index].inclusion_distance
+            )
+            for index in previous_epoch_attester_indices
+        }
+        rewards_received, penalties_received = _apply_rewards_and_penalties(
+            rewards,
+            previous_epoch_attester_indices,
+            dict(),
```
hwwhww: Although I know you won't mutate the first 4 arguments. It still makes me nervous to pass `dict()` in arguments. I'd suggest you make another `NamedTuple` for it. e.g.,
```python
class SettlementContext(NamedTuple):
    rewards: Dict[ValidatorIndex, Gwei] = dict()
    indices_to_reward: Iterable[ValidatorIndex] = set()
    penalties: Dict[ValidatorIndex, Gwei] = dict()
    indices_to_penalize: Iterable[ValidatorIndex] = set()
    rewards_received: Dict[ValidatorIndex, Gwei]
    penalties_received: Dict[ValidatorIndex, Gwei]
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +285,500 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+        )
+    return rewards_received, penalties_received
+
+
+@curry
+def _process_rewards_and_penalties_for_finality(
+        old_rewards_received: Dict[ValidatorIndex, Gwei],
+        state: BeaconState,
+        config: BeaconConfig,
+        prev_epoch_active_validator_indices: Set[ValidatorIndex],
```
hwwhww: Remame all `prev_epoch_active_validator_indices` to `previous_epoch_active_validator_indices` for consistency.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +285,500 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+    )
+
+    # Compute inclusion slot/distance of previous attestations for later use.
+    inclusion_infos = get_inclusion_infos(
+        state=state,
+        attestations=previous_epoch_attestations,
+        committee_config=CommitteeConfig(config),
+    )
+
+    # Compute effective balance of each previous epoch active validator for later use
```
hwwhww: I see the point here for efficiency but I also like that we can make three lines into one function call if we use `get_total_balance` later.

Since this pattern happens 4 times, maybe we can add another helper function:
```python
def get_total_balance_from_effective_balances(
    effective_balances: Dict[ValidatorIndex, Gwei],
    validator_indices: Sequence[ValidatorIndex]) -> Gwei
    return sum(
        effective_balances[i]
        for i in validator_indices
    )
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +285,500 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+            set(),
+            inactivity_penalties,
+            excluded_active_validators_indices,
+            rewards_received,
+            penalties_received,
+        )
+
+        # Punish active validators not in `previous_epoch_head_attester_indices`
+        excluded_active_validators_indices = prev_epoch_active_validator_indices.difference(
+            previous_epoch_head_attester_indices)
```
hwwhww: nitpick
```suggestion
            previous_epoch_head_attester_indices,
        )
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +285,500 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+    if epochs_since_finality <= 4:
+        # 1.1 Expected FFG source:
+        previous_epoch_attesting_balance = sum(
+            effective_balances[i]
+            for i in previous_epoch_attester_indices
+        )
+        # Reward validators in `previous_epoch_attester_indices`
+        # # Punish active validators not in `previous_epoch_attester_indices`
+        excluded_active_validators_indices = prev_epoch_active_validator_indices.difference(
+            previous_epoch_attester_indices)
```
hwwhww: nitpick
```suggestion
            previous_epoch_attester_indices,
        )
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +285,500 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+
+        # 1.2 Expected FFG target:
+        previous_epoch_boundary_attesting_balance = sum(
+            effective_balances[i]
+            for i in previous_epoch_boundary_attester_indices
+        )
+        # Reward validators in `previous_epoch_boundary_attester_indices`
+        # Punish active validators not in `previous_epoch_boundary_attester_indices`
+        excluded_active_validators_indices = prev_epoch_active_validator_indices.difference(
+            previous_epoch_boundary_attester_indices)
```
hwwhww: nitpick
```suggestion
            previous_epoch_boundary_attester_indices,
        )
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +285,500 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+
+        # 1.3 Expected beacon chain head:
+        previous_epoch_head_attesting_balance = sum(
+            effective_balances[i]
+            for i in previous_epoch_head_attester_indices
+        )
+        # Reward validators in `previous_epoch_head_attester_indices`
+        # Punish active validators not in `previous_epoch_head_attester_indices`
+        excluded_active_validators_indices = prev_epoch_active_validator_indices.difference(
+            previous_epoch_head_attester_indices)
```
hwwhww: nitpick
```suggestion
            previous_epoch_head_attester_indices,
        )
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +285,500 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+            set(),
+            rewards_received,
+            penalties_received,
+        )
+
+    # epochs_since_finality > 4
+    else:
+        # Punish active validators not in `previous_epoch_attester_indices`
+        excluded_active_validators_indices = prev_epoch_active_validator_indices.difference(
+            previous_epoch_attester_indices)
```
hwwhww: nitpick
```suggestion
            previous_epoch_attester_indices,
        )
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +285,500 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+            set(),
+            inactivity_penalties,
+            excluded_active_validators_indices,
+            rewards_received,
+            penalties_received,
+        )
+
+        # Punish active validators not in `previous_epoch_boundary_attester_indices`
+        excluded_active_validators_indices = prev_epoch_active_validator_indices.difference(
+            previous_epoch_boundary_attester_indices)
```
hwwhww: nitpick
```suggestion
            previous_epoch_boundary_attester_indices,
        )
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +285,500 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+            penalties,
+            excluded_active_validators_indices,
+            rewards_received,
+            penalties_received,
+        )
+
+        # 1.2 Expected FFG target:
+        previous_epoch_boundary_attesting_balance = sum(
+            effective_balances[i]
+            for i in previous_epoch_boundary_attester_indices
```
hwwhww: Hmm, it seems we do assume that `previous_epoch_boundary_attester_indices` is a `Set`, but since you use `Iterable` in type hinting, it would be nice to add `set()` here.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +285,500 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+
+@curry
+def _process_rewards_and_penalties_for_finality(
+        old_rewards_received: Dict[ValidatorIndex, Gwei],
+        state: BeaconState,
+        config: BeaconConfig,
+        prev_epoch_active_validator_indices: Set[ValidatorIndex],
+        previous_total_balance: Gwei,
+        previous_epoch_attester_indices: Iterable[ValidatorIndex],
+        previous_epoch_boundary_attester_indices: Iterable[ValidatorIndex],
```
hwwhww: Hmm, it seems we do assume that `previous_epoch_boundary_attester_indices` is a `Set`, but since you use `Iterable` in type hinting, it would be nice to add `previous_epoch_boundary_attester_indices = set(previous_epoch_boundary_attester_indices)` before we use it.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +287,503 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+            penalties_received,
+        )
+    return rewards_received, penalties_received
+
+
+@curry
+def _process_rewards_and_penalties_for_finality(
+        old_rewards_received: Dict[ValidatorIndex, Gwei],
+        state: BeaconState,
+        config: BeaconConfig,
```
hwwhww: nitpick: order the parameters by
```
1. state: BeaconState,
2. config: BeaconConfig,
3. others ...
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +287,503 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+            historical_rewards_received,
+        )
+    return historical_rewards_received
+
+
+@curry
+def _process_rewards_and_penalties_for_attestation_inclusion(
+        old_rewards_received: Dict[ValidatorIndex, Gwei],
+        state: BeaconState,
+        config: BeaconConfig,
```
hwwhww: nitpick: order the parameters by
```
1. state: BeaconState,
2. config: BeaconConfig,
3. others ...
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +287,503 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+        reward = base_rewards[index] // config.ATTESTATION_INCLUSION_REWARD_QUOTIENT
+        rewards_received[proposer_index] = Gwei(rewards_received[proposer_index] + reward)
+    return rewards_received
+
+
+@curry
+def _process_rewards_and_penalties_for_crosslinks(
+        old_rewards_received: Dict[ValidatorIndex, Gwei],
+        state: BeaconState,
+        config: BeaconConfig,
```
hwwhww: nitpick: order the parameters by
```
1. state: BeaconState,
2. config: BeaconConfig,
3. others ...
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +287,503 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+                rewards_received=rewards_received,
+                penalties_received=penalties_received,
+            ),
+        )
+
+    historical_rewards_received = old_rewards_received.copy()
+    for index in rewards_received:
+        historical_rewards_received = _update_rewards_or_penalies(
+            index,
+            rewards_received[index] - penalties_received[index],
```
hwwhww: Note that `historical_rewards_received : Dict[ValidatorIndex, Gwei]`, here the `Gwei` might be a negative integer - which doesn't fit the definition in spec (https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#custom-types).
I think we'd need another way to explicitly note it, e.g., adding another custom type for this case like `SignedGwei/GweiSigned` for handling rewards.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -263,6 +287,503 @@ def process_crosslinks(state: BeaconState, config: BeaconConfig) -> BeaconState:
...
+            total_balance = get_total_balance_from_effective_balances(
+                effective_balances,
+                crosslink_committee,
+            )
+            for index in attesting_validator_indices:
+                reward = base_rewards[index] * total_attesting_balance // total_balance
+                rewards_received[index] = Gwei(rewards_received[index] + reward)
+            for index in set(crosslink_committee).difference(attesting_validator_indices):
+                penalty = base_rewards[index]
+                rewards_received[index] = Gwei(rewards_received[index] - penalty)
```
hwwhww: As `historical_rewards_received`, it's possible that `Gwei` becomes negative here.

tests/eth2/beacon/state_machines/forks/test_serenity_epoch_processing.py
```diff
@@ -637,3 +649,478 @@ def test_process_crosslinks(
...
+    ),
+    [
+        (
+            10,
+            2,
+            5,
+            2,
+            4,
+            10,
+            4, 15,  # epochs_since_finality <= 4
```
hwwhww: It seems the first 6 parameters ('n,' ... 'inactivity_penalty_quotient,') are fixed. You can extract that:
```python
@pytest.mark.parametrize(
    (
        'n,'
        'slots_per_epoch,'
        'target_committee_size,'
        'shard_count,'
        'min_attestation_inclusion_delay,'
        'inactivity_penalty_quotient'
    )
    [
        (
            10,
            2,
            5,
            2,
            4,
            10
        )
    ]
)
@pytest.mark.parametrize(
    (
        'finalized_epoch,current_slot,'
        'penalized_validator_indices,'
        'previous_epoch_active_validator_indices,'
        'previous_epoch_attester_indices,'
        'previous_epoch_boundary_head_attester_indices,'
        'inclusion_distances,'
        'effective_balance,base_reward,'
        'expected_rewards_received'
    ),
    [
        (
            4, 15,  # epochs_since_finality <= 4
            ....
        )
        (
            3, 15,  # epochs_since_finality > 4
            ....
        )
    ] 
```

eth2/beacon/datastructures/rewards_settlement_context.py
```diff
@@ -0,0 +1,19 @@
...
+    Set,
+)
+
+from eth2.beacon.typing import (
+    Gwei,
+    ValidatorIndex,
+)
+
+
+class RewardsSettlementContext(NamedTuple):
```
hwwhww: Suggestion: `RewardsSettlementContext` -> `RewardSettlementContext`

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
NIC619: Add `SignedGwei` and use it for `rewards_received` map where value might potentially be negative and leave `rewards`/`penalties` map as is.

PR #277, opened 2019-02-13
Remove dead code from tests

cburgdorf:
### What was wrong?

There's a bunch of code in `test_sync` that, to me, seems to be just extra noise. These tests all test the various syncer that we have but it is unclear to me why why the *serving* side should also setup a `RegularChainSyncer`. The tests pass without this code so if it serves any particular purpose it's at least unclear to me and deserves a comment.

### How was it fixed?

Deleting it to find out if anybody will call me a stupid :monkey_face: and points out why we should keep it.

#### Cute Animal Picture

### DIFF
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -43,22 +43,10 @@ async def test_fast_syncer(request, event_loop, chaindb_fresh, chaindb_20):
     client_peer_pool = MockPeerPoolWithConnectedPeers([client_peer])
     client = FastChainSyncer(ByzantiumTestChain(chaindb_fresh.db), chaindb_fresh, client_peer_pool)
     server_peer_pool = MockPeerPoolWithConnectedPeers([server_peer])
-    server = RegularChainSyncer(
-        ByzantiumTestChain(chaindb_20.db),
-        chaindb_20,
-        server_peer_pool,
-    )
-    asyncio.ensure_future(server.run())
     server_request_handler = ETHRequestServer(FakeAsyncChainDB(chaindb_20.db), server_peer_pool)
     asyncio.ensure_future(server_request_handler.run())
     server_peer.logger.info("%s is serving 20 blocks", server_peer)
     client_peer.logger.info("%s is syncing up 20", client_peer)
-
-    def finalizer():
-        event_loop.run_until_complete(server.cancel())
-        # Yield control so that server.run() returns, otherwise asyncio will complain.
-        event_loop.run_until_complete(asyncio.sleep(0.1))
-    request.addfinalizer(finalizer)
 
     # FastChainSyncer.run() will return as soon as it's caught up with the peer.
     await asyncio.wait_for(client.run(), timeout=2)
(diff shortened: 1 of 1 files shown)


tests/core/p2p-proto/test_sync.py
```diff
@@ -165,23 +126,14 @@ def finalizer():
...
-    )
-    asyncio.ensure_future(server.run())
+
     server_request_handler = LightRequestServer(FakeAsyncHeaderDB(chaindb_20.db), server_peer_pool)
     asyncio.ensure_future(server_request_handler.run())
     server_peer.logger.info("%s is serving 20 blocks", server_peer)
     client_peer.logger.info("%s is syncing up 20", client_peer)
 
     def finalizer():
-        event_loop.run_until_complete(asyncio.gather(
```
cburgdorf: Noting that this is the only test case that still seems to need the `finalizer` to shut down cleanly.

carver:
Hm, not sure how I feel about dropping the cleanup stuff just because the test doesn't fail. I think I'm :+1: on removing the `server` if it's not needed, but think we should probably still prefer to wait for the client to end.

cburgdorf:
>Hm, not sure how I feel about dropping the cleanup stuff just because the test doesn't fail.

They aren't needed as far I can tell. Two of these finalizer were just running `server.cancel()`. But as it turned out there is no need for `server` in this test. At least, not in the sense of what the test meant to be the `server`. The terminology is confusing here. The real server is the `ETHRequestServer` which is still there.

One of the `finalizer` had both `server.cancel()` and `client.cancel()` so we *could* keep `client.cancel()` but we don't need to cancel it as it cleanly shuts down when the sync finishes.

As for the last `finalizer` that still remains with `client.cancel()` this one still needs to be there as the `LightChainSyncer` does not shut down on its own without it.

As far as I can tell, all tests run cleanly without leaving any cruft behind.

tests/core/p2p-proto/test_sync.py
```diff
@@ -125,27 +101,12 @@ def finalizer():
...
-    asyncio.ensure_future(server.run())
+
     server_request_handler = ETHRequestServer(FakeAsyncChainDB(chaindb_20.db), server_peer_pool)
     asyncio.ensure_future(server_request_handler.run())
     server_peer.logger.info("%s is serving 20 blocks", server_peer)
     client_peer.logger.info("%s is syncing up 20", client_peer)
 
-    def finalizer():
-        event_loop.run_until_complete(asyncio.gather(
-            client.cancel(),
```
carver: Right, this `client.cancel()` is the one that makes sense to keep, I think. The only specific reason I can think of is that when the test fails, we avoid irrelevant warnings that might happen related to the client not shutting down. Other than that it just _feels_ like good hygiene. But if you really want to kill it, I wouldn't initiate a duel at dawn or anything :).

cburgdorf: >I wouldn't initiate a duel at dawn or anything



But I want my peers to be happy so I'll change it :)

PR #290, opened 2019-02-18
SSZ

jannikluhn:
### What was wrong?

We encoded objects in RLP, not SSZ.

### How was it fixed?

- add py-ssz as a requirement for the eth2 package
- change all types from `rlp.Serializable` to `ssz.Serializable` and the sedes objects accordingly
- change the encodings used in the database (note that this means that we encode integers for slot and score as fixed sized uints. Should the type change we must not to update them here as well!)
- use SSZ instead of RLP for hashing
- add a version of `validate_rlp_equal` that accepts SSZ objects (the implementation didn't change)

Not done:

- tree hashing instead of flat hashing (it might make sense to add this to py-ssz directly)
- networking stuff (but at this point it's unclear if we're going to use SSZ or RLP for that)

#### Cute Animal Picture


eth2/beacon/types/blocks.py
```diff
@@ -105,17 +105,17 @@ def cast_block_body(cls,
...
         ('slot', uint64),
-        ('parent_root', hash32),
-        ('state_root', hash32),
-        ('randao_reveal', binary),  # TODO: should be bytes96 once we transition to SSZ
+        ('parent_root', bytes32),
+        ('state_root', bytes32),
+        ('randao_reveal', bytes96),
         ('eth1_data', Eth1Data),
-        ('signature', binary),
+        ('signature', bytes_sedes),
```
hwwhww:
```suggestion
        ('signature', bytes96),
```

eth2/beacon/types/deposit_input.py
```diff
@@ -17,17 +15,15 @@
...
     fields = [
         # BLS pubkey
-        ('pubkey', binary),
+        ('pubkey', bytes_sedes),
         # Withdrawal credentials
-        ('withdrawal_credentials', hash32),
+        ('withdrawal_credentials', bytes32),
         # BLS proof of possession (a BLS signature)
-        ('proof_of_possession', binary),
+        ('proof_of_possession', bytes_sedes),
```
hwwhww:
```suggestion
        ('proof_of_possession', bytes96),
```

eth2/beacon/types/attestations.py
```diff
@@ -14,18 +14,16 @@
...
         ('data', AttestationData),
         # Attester aggregation bitfield
-        ('aggregation_bitfield', binary),
+        ('aggregation_bitfield', bytes_sedes),
         # Proof of custody bitfield
-        ('custody_bitfield', binary),
+        ('custody_bitfield', bytes_sedes),
         # BLS aggregate signature
-        ('aggregate_signature', binary),
+        ('aggregate_signature', bytes_sedes),
```
hwwhww:
```suggestion
        ('aggregate_signature', bytes96),
```

eth2/beacon/types/slashable_attestations.py
```diff
@@ -28,19 +27,17 @@
...
-        ('validator_indices', CountableList(uint64)),
+        ('validator_indices', List(uint64)),
         # Attestation data
         ('data', AttestationData),
         # Custody bitfield
-        ('custody_bitfield', binary),
+        ('custody_bitfield', bytes_sedes),
         # Aggregate signature
-        ('aggregate_signature', binary),
+        ('aggregate_signature', bytes_sedes),
```
hwwhww:
```suggestion
        ('aggregate_signature', bytes96),
```

hwwhww:
> 1. All signatures are `bytes96`.
> 2. Can we replace `bytes_sedes` with `List[bytes1]` now? (https://github.com/ethereum/py-ssz/issues/40)

BTW I think it's okay to leave (2) to other PRs. Just note here.

jannikluhn:
> 1. All signatures are `bytes96`.
> 2. Can we replace `bytes_sedes` with `List[bytes1]` now? ([ethereum/py-ssz#40](https://github.com/ethereum/py-ssz/issues/40))

Fixed (1). I think (2) is something we should fix in py-ssz (i.e. `bytes_sedes = List(bytes1)`).

There are still some tests failing. I'm trying to fix the core tests, but I'm quite confused by the failing `pyXX-wheel-cli` tests.

hwwhww:
I don't understand either, why it uses `pip` to install modules and test with local `tests/integration/`?

https://github.com/ethereum/trinity/blob/72d6db48ee73a71b99f1411eb441302abc1093d1/tox.ini#L61-L65

Page snake charmers @pipermerriam @carver @cburgdorf  @ralexstokes  🐍

ralexstokes:
@jannikluhn @hwwhww not that familiar w/ the tox setup/internals but it seems fine in terms of "install the correct dependencies and run the integration tests"

if you trace the dependency graph, we can see that `py-ssz` is only being pulled in as a dep for the `eth2` package although it is now transitively pulled in when running code in the `trinity` package and we get the `ImportError` when `py-ssz` cannot be resolved.

it seems like we would have to (at least for now) modify the `setup.py` line w/ `install_requires` to add `deps['eth2']` or modify the `common-wheel-cli` entry in the `tox.ini` so that it `pip install [p2p,trinity,eth2]` -- perhaps someone more familiar w/ the `tox` setup can chime in?

i say 'at least for now' as i've been wondering if there are any plans to separate out the `eth2` package into its own repo (like `py-evm`) so that it can be separately managed... in that case we can skip the above and just add `eth2` as a dep like `py-evm` and this should go away

jannikluhn:
Fixed the failing networking tests: Now we SSZ-encode the blocks before enveloping them in the RLP message.

I cheated a little with the dependencies and did not make eth2.0 a dependency of trinity to keep them as separate as possible, even though it technically is (because the BCC networking code imports `BeaconBlock`). But it should be installed in the relevant tox environments. What's confusing to me is that the pyXX-wheel-cli environments seem to not depend on any of our code: https://github.com/ethereum/trinity/blob/0f5e3ecdc375ed4562fc63917843b5c593593d34/tox.ini#L51-L56

> i say 'at least for now' as i've been wondering if there are any plans to separate out the `eth2` package into its own repo (like `py-evm`) so that it can be separately managed...

For now we decided to keep them in the same repository, because a change in eth2.0 quite often requires a change in trinity (see for instance this PR) and we'd like to avoid the added friction of making new releases constantly.

cburgdorf:
The wheel-cli tests (as far as I understand them) ensure users will by able to install and run Trinity via `pip`. They are basically creating a package, install it and then run the CLI tests which run a real Trinity instance and perform some basic tests against it.

In its current state, this PR seems to break this. In other words, a user trying to install and run Trinity would run into this missing dependency error.

I believe we should just add `eth2` here similar to how we handle `p2p` which also happens to be an external library that is still sitting in the same repo as Trinity.

https://github.com/ethereum/trinity/blob/0f5e3ecdc375ed4562fc63917843b5c593593d34/setup.py#L94

eth2/beacon/types/validator_records.py
```diff
@@ -19,15 +17,13 @@
...
-class ValidatorRecord(rlp.Serializable):
-    """
-    Note: using RLP until we have standardized serialization format.
-    """
+class ValidatorRecord(ssz.Serializable):
+
     fields = [
         # BLS public key
-        ('pubkey', binary),
+        ('pubkey', bytes_sedes),
```
hwwhww:
```suggestion
        ('pubkey', bytes48),
```

eth2/beacon/types/deposit_input.py
```diff
@@ -17,17 +16,15 @@
...
-class DepositInput(rlp.Serializable):
-    """
-    Note: using RLP until we have standardized serialization format.
-    """
+class DepositInput(ssz.Serializable):
+
     fields = [
         # BLS pubkey
-        ('pubkey', binary),
+        ('pubkey', bytes_sedes),
```
hwwhww:
```suggestion
        ('pubkey', bytes48),
```

jannikluhn:
> I believe we should just add `eth2` here similar to how we handle `p2p` which also happens to be an external library that is still sitting in the same repo as Trinity.

That did the trick, thanks a lot for your help!

@hwwhww Fixed the pubkeys and a leftover signature. There are two signatures left in `ProposerSlashing`, but I didn't touch those as the spec has diverged from our implementation here. Please let me know if I should update those as well. If not, I think this should be ready.

carver:
> The wheel-cli tests (as far as I understand them) ensure users will by able to install and run Trinity via `pip`. They are basically creating a package, install it and then run the CLI tests which run a real Trinity instance and perform some basic tests against it.
> 
> In its current state, this PR seems to break this. In other words, a user trying to install and run Trinity would run into this missing dependency error.

Yup, exactly, this is a great example of a kind of thing that the test is meant to catch.

I think the ideal solution is to update things so that a basic run of trinity on eth1 mainnet doesn't try to load eth2 libraries (then `eth2` dependencies would be optional again). But this solution seems okay in the short term. We definitely don't want to get in the way of eth2 progress.

PR #291, opened 2019-02-18
Move request server into an isolated plugin

### LINKED ISSUE #84: Move handling of peer requests to a plugin (and it's own process)
### What is wrong?

The code which responds to peer requests for data runs alongside everything else.  This is problematic because the CPU time is shared between it **and** the networking code **and** the synchronization code.

This results in a high peer count and high volume of incoming requests having a direct impact on both our baseline networking performance as well as sync performance.

### How can it be fixed

Fully move the handling of peer requests into an isolated process (probably a built-in plugin).

cburgdorf:
### What was wrong?

Fixes #84 

- Our `networking` process is doing too much.
- Request server code needs to be a plugin to be reusable across beacon chain client

### How was it fixed?

As explained in #213, this makes the first step towards nuking the `networking` process by moving the request server into its own isolated process.

#### Cute Animal Picture

### DIFF
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -35,10 +35,7 @@ from p2p.constants import (
     REQUEST_PEER_CANDIDATE_TIMEOUT,
 )
 from p2p.events import (
-    ConnectToNodeCommand,
     PeerCandidatesRequest,
-    PeerCountRequest,
-    PeerCountResponse,
     RandomBootnodeRequest,
 )
 from p2p.exceptions import (
@@ -85,7 +82,7 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
                  max_peers: int = DEFAULT_MAX_PEERS,
                  peer_info: BasePeerInfo = NoopPeerInfo(),
                  token: CancelToken = None,
-                 event_bus: Endpoint = None
+                 event_bus: Endpoint = None,
                  ) -> None:
         super().__init__(token)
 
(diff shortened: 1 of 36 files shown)


p2p/peer_pool.py
```diff
@@ -351,6 +339,14 @@ def __aiter__(self) -> AsyncIterator[BasePeer]:
             self.logger.debug("== End peer details == ")
             await self.sleep(self._report_interval)
 
+    # TODO: The lookup should not need looping through the peers
+    def super_shitty_peer_lookup(self, dto_peer: IdentifiablePeer) -> BasePeer:
```
cburgdorf: Oops..guess I still need to take care of this.

cburgdorf:
@jannikluhn @hwwhww This doesn't touch the beacon chain types yet but once it has gone through review and we polished out some things, we can add the beacon chain handling and then `RequestServerPlugin` can be used from the beacon node as well. It runs in its own process and communicates with the peer pool via event bus.

p2p/peer_pool_event_bus_request_handler.py
```diff
@@ -0,0 +1,84 @@
...
+
+if TYPE_CHECKING:
+    from p2p.peer_pool import (  # noqa: F401
+        BasePeerPool,
+    )
+
+TPeer = TypeVar('TPeer', bound=BasePeer)
+
+
+class BasePeerPoolEventBusRequestHandler(BaseService, Generic[TPeer]):
```
cburgdorf: Maybe this would make more sense in `trinity.protocol.common`?

cburgdorf:
Closing this. Aim to replace it with a new up to date implementation that uses the upcoming lahja release.

PR #300, opened 2019-02-22
Fix attestation, previous epoch logic and builder tools

### LINKED ISSUE #294: Test state.finalized_epoch in test_demo.py
### What is wrong?

The test in tests/eth2/beacon/state_machines/test_demo.py does not justify or finalize any epochs.

### How can it be fixed

- Need to add correct `epoch_boundary_root` and `justified_block_root` to attestations in `create_mock_signed_attestations_at_slot`.
  - https://github.com/ethereum/eth2.0-specs/issues/652 would be a block for this issue. A workaround is to skip attestations when `slot % epoch_length == 0`
- `assert` epoch processing variables such as `state.finalized_epoch` and `state.justified_epoch`, etc.

hwwhww:
### What was wrong?
To fix #294, also need to fix other issues to make `test_demo.py` work.

### How was it fixed?
1. Address ethereum/eth2.0-specs#672: fix up previous epoch logic around genesis
2. Address ethereum/eth2.0-specs#627: fix off by one attestaton issue
3. Fix `create_mock_signed_attestation`: Address ethereum/eth2.0-specs#671
4. Fix `test_demo.py`
	1. Use map `attestations_map: Dict[Slot, Sequence[Attestation]]` to keep the generated attestations, and include them when reache `MIN_ATTESTATION_INCLUSION_DELAY`.
	2. Add justification assertions

#### Cute Animal Picture
albino_kangaroo
(Foto: Heike Katthagen)

### DIFF
--- a/tests/eth2/beacon/conftest.py
+++ b/tests/eth2/beacon/conftest.py
@@ -57,7 +57,7 @@ def privkeys():
     Rationales:
     1. Making the privkeys be small integers to make multiplying easier for tests.
     2. Using ``2**i`` instead of ``i``:
-        The combinations of validators would not lead to unique pubkeys.
+        If using ``i``, the combinations of privkeys would not lead to unique pubkeys.
     """
     return [2 ** i for i in range(100)]
 
--- a/tests/eth2/beacon/conftest.py
+++ b/tests/eth2/beacon/conftest.py
@@ -8,7 +8,6 @@ from eth_utils import (
 )
 
 import eth2._utils.bls as bls
-from eth2.beacon._utils.hash import hash_eth2
 from eth2.beacon.configs import (
     BeaconConfig,
     CommitteeConfig,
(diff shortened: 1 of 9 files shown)


tests/eth2/beacon/state_machines/forks/test_serenity_block_attestation_validation.py
```diff
@@ -114,6 +113,7 @@ def test_validate_attestation_justified_epoch(
     )
 
     if is_valid:
+        # slot_to_epoch(attestation_data.slot + 1, slots_per_epoch) >= current_epoch
```
ChihChengLiang: Not sure if this comment makes sense here. The `is_valid` is determined only by justified_epoch here?

hwwhww: Should have removed this line. Thanks!

tests/eth2/beacon/conftest.py
```diff
@@ -54,7 +53,13 @@
 
 @pytest.fixture(scope="session")
 def privkeys():
-    return [int.from_bytes(hash_eth2(str(i).encode('utf-8'))[:4], 'big') for i in range(100)]
+    """
+    Rationales:
+    1. Making the privkeys be small integers to make multiplying easier for tests.
+    2. Using ``2**i`` instead of ``i``:
+        The combinations of validators would not lead to unique pubkeys.
```
NIC619: Should it be `...would lead to unique pubkeys`?

PR #307, opened 2019-02-25
Fix #278, minimal viable trinity-beacon node

### LINKED ISSUE #278: Implement minimal viable trinity-beacon node
### What is wrong?

Putting together all the components, we are building a node now. Per #240, this will be an IsolatedPlugin.

This is a temporary solution before the new PeerPool with EventBus-based API is built.

### How can it be fixed

- Need a new plugin, located at `trinity/plugins/eth2/beacon/`
- Create a `class BeaconNodePlugin(BaseIsolatedPlugin)` in `trinity/plugins/eth2/beacon/plugin.py`
    - It should bootstrap these
        - `PeerPool`
        - Chainserver
        - Syncer
- Modify `trinity/main_beacon.py` to add BeaconNodePlugin. So that when running `trinity-beacon` on the terminal, it would boot the plugin.

ChihChengLiang:
### What was wrong?

#278

### How was it fixed?

- [x] Create a plugin at trinity/plugins/eth2/beacon/
- ~~Add SerenityStateTransition somewhere~~

#### Cute Animal Picture

<a title="Appaloosa [CC BY-SA 3.0 (https://creativecommons.org/licenses/by-sa/3.0)], via Wikimedia Commons" href="https://commons.wikimedia.org/wiki/File:P.t.altaica_Tomak_Male.jpg">P.t.altaica Tomak Male</a>

### DIFF
--- a/tests/core/p2p-proto/bcc/test_requests.py
+++ b/tests/core/p2p-proto/bcc/test_requests.py
@@ -205,7 +205,7 @@ async def test_get_incomplete_canonical_block_range(request, event_loop):
     assert response.payload["request_id"] == 5
     blocks = tuple(ssz.decode(block, BeaconBlock) for block in response.payload["encoded_blocks"])
     assert len(blocks) == 5
-    assert [block.slot for block in blocks] == [genesis.slot + s for s in (3, 4, 5, 6, 7)]
+    assert [block.slot for block in blocks] == [genesis.slot + s for s in [3, 4, 5, 6, 7]]
     assert blocks == base_branch[-1:] + canonical_branch
 
 
--- a/tests/core/p2p-proto/bcc/test_requests.py
+++ b/tests/core/p2p-proto/bcc/test_requests.py
@@ -23,10 +23,14 @@ from .helpers import (
     create_branch,
     get_directly_linked_peers_in_peer_pools,
 )
+from eth2.beacon.state_machines.forks.serenity import SERENITY_CONFIG
 
 
 async def get_request_server_setup(request, event_loop, chain_db):
-    genesis = await chain_db.coro_get_canonical_block_by_slot(0, BeaconBlock)
+    genesis = await chain_db.coro_get_canonical_block_by_slot(
+        SERENITY_CONFIG.GENESIS_SLOT,
+        BeaconBlock,
+    )
     alice_chain_db = await get_chain_db((genesis,))
     alice, alice_peer_pool, bob, bob_peer_pool = await get_directly_linked_peers_in_peer_pools(
         request,
@@ -51,7 +55,7 @@ async def get_request_server_setup(request, event_loop, chain_db):
 
 @pytest.mark.asyncio
 async def test_get_single_block_by_slot(request, event_loop):
-    block = create_test_block(slot=0)
+    block = create_test_block()
     chain_db = await get_chain_db((block,))
     alice, response_buffer = await get_request_server_setup(request, event_loop, chain_db)
 
@@ -67,11 +71,11 @@ async def test_get_single_block_by_slot(request, event_loop):
 
 @pytest.mark.asyncio
 async def test_get_single_block_by_root(request, event_loop):
-    block = create_test_block(slot=0)
+    block = create_test_block()
     chain_db = await get_chain_db((block,))
     alice, response_buffer = await get_request_server_setup(request, event_loop, chain_db)
 
-    alice.sub_proto.send_get_blocks(0, 1, request_id=5)
+    alice.sub_proto.send_get_blocks(block.slot, 1, request_id=5)
     response = await response_buffer.msg_queue.get()
 
     assert isinstance(response.command, BeaconBlocks)
@@ -83,11 +87,11 @@ async def test_get_single_block_by_root(request, event_loop):
 
 @pytest.mark.asyncio
 async def test_get_no_blocks(request, event_loop):
-    block = create_test_block(slot=0)
+    block = create_test_block()
     chain_db = await get_chain_db((block,))
     alice, response_buffer = await get_request_server_setup(request, event_loop, chain_db)
 
-    alice.sub_proto.send_get_blocks(0, 0, request_id=5)
+    alice.sub_proto.send_get_blocks(block.slot, 0, request_id=5)
     response = await response_buffer.msg_queue.get()
 
     assert isinstance(response.command, BeaconBlocks)
@@ -99,11 +103,11 @@ async def test_get_no_blocks(request, event_loop):
 
 @pytest.mark.asyncio
 async def test_get_unknown_block_by_slot(request, event_loop):
-    block = create_test_block(slot=0)
+    block = create_test_block()
     chain_db = await get_chain_db((block,))
     alice, response_buffer = await get_request_server_setup(request, event_loop, chain_db)
 
-    alice.sub_proto.send_get_blocks(100, 1, request_id=5)
+    alice.sub_proto.send_get_blocks(block.slot + 100, 1, request_id=5)
     response = await response_buffer.msg_queue.get()
 
     assert isinstance(response.command, BeaconBlocks)
@@ -115,7 +119,7 @@ async def test_get_unknown_block_by_slot(request, event_loop):
 
 @pytest.mark.asyncio
 async def test_get_unknown_block_by_root(request, event_loop):
-    block = create_test_block(slot=0)
+    block = create_test_block()
     chain_db = await get_chain_db((block,))
     alice, response_buffer = await get_request_server_setup(request, event_loop, chain_db)
 
@@ -133,7 +137,7 @@ async def test_get_unknown_block_by_root(request, event_loop):
 async def test_get_canonical_block_range_by_slot(request, event_loop):
     chain_db = await get_chain_db()
 
-    genesis = create_test_block(slot=0)
+    genesis = create_test_block()
     base_branch = create_branch(3, root=genesis)
     non_canonical_branch = create_branch(3, root=base_branch[-1], state_root=b"\x00" * 32)
 
@@ -144,14 +148,14 @@ async def test_get_canonical_block_range_by_slot(request, event_loop):
 
     alice, response_buffer = await get_request_server_setup(request, event_loop, chain_db)
 
-    alice.sub_proto.send_get_blocks(2, 4, request_id=5)
+    alice.sub_proto.send_get_blocks(genesis.slot + 2, 4, request_id=5)
     response = await response_buffer.msg_queue.get()
 
     assert isinstance(response.command, BeaconBlocks)
     assert response.payload["request_id"] == 5
     blocks = tuple(ssz.decode(block, BeaconBlock) for block in response.payload["encoded_blocks"])
     assert len(blocks) == 4
-    assert [block.slot for block in blocks] == [2, 3, 4, 5]
+    assert [block.slot for block in blocks] == [genesis.slot + s for s in [2, 3, 4, 5]]
     assert blocks == base_branch[1:] + canonical_branch[:2]
 
 
@@ -159,7 +163,7 @@ async def test_get_canonical_block_range_by_slot(request, event_loop):
 async def test_get_canonical_block_range_by_root(request, event_loop):
     chain_db = await get_chain_db()
 
-    genesis = create_test_block(slot=0)
+    genesis = create_test_block()
     base_branch = create_branch(3, root=genesis)
     non_canonical_branch = create_branch(3, root=base_branch[-1], state_root=b"\x00" * 32)
     canonical_branch = create_branch(4, root=base_branch[-1], state_root=b"\x11" * 32)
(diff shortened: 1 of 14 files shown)


trinity/_utils/chains.py
```diff
@@ -137,8 +137,10 @@ def construct_trinity_config_params(
 
     if args.max_peers is not None:
         yield 'max_peers', args.max_peers
-    else:
+    elif "sync_mode" in args:
         yield 'max_peers', _default_max_peers(args.sync_mode)
```
cburgdorf: This is kinda sad since we've eliminated the `sync_mode` from the core code and moved it into the plugin space. But I don't have a better answer to it right now and think it is easier to address once the `PeerPool` has migrated into a plugin. Maybe put a comment on it:

>FIXME: This part of the code base should not know about `sync_mode`

trinity/plugins/eth2/beacon/plugin.py
```diff
@@ -0,0 +1,66 @@
...
+class BeaconNodePlugin(BaseIsolatedPlugin):
+
+    @property
+    def name(self) -> str:
+        return "Beacon Node"
+
+    def on_ready(self) -> None:
+        self.start()
+
+    def do_start(self) -> None:
```
cburgdorf: :+1: Nice and simple to get things going

trinity/plugins/eth2/beacon/plugin.py
```diff
@@ -0,0 +1,66 @@
...
+            keymap=dict(),
+            genesis_block_class=SerenityBeaconBlock,
+        )
+        chain_db.persist_block_chain((genesis_block,), BeaconBlock)
+        context = BeaconContext(
+            chain_db=chain_db,
+            network_id=1,
+        )
+
+        peer_pool = BCCPeerPool(
```
cburgdorf: To make things like the JSON-RPC API work, you need to pass an `event_bus` into the pool.

trinity/plugins/eth2/beacon/plugin.py
```diff
@@ -0,0 +1,66 @@
...
+            privkey=ecies.generate_privkey(),
+            context=context,
+        )
+
+        syncer = BeaconChainSyncer(
+            chain_db,
+            peer_pool,
+            peer_pool.cancel_token,
+        )
+        asyncio.ensure_future(syncer.run())
```
cburgdorf: Noting that in order to handle SIGINTS this need to be more like this:

```python
        asyncio.ensure_future(exit_with_service_and_endpoint(peer_pool, self.context.event_bus))
        asyncio.ensure_future(peer_pool.run())
        asyncio.ensure_future(exit_with_service(syncer))
        asyncio.ensure_future(syncer.run())
        loop.run_forever()
        loop.close()
```

pipermerriam: Can this get wrapped up in a utility?  It looks like a version of this is repeated in at least 3 places.

- external plugin example
- json_rpc/plugin.py
- peer_discodery/pluging.py

cburgdorf: >Can this get wrapped up in a utility?

Almost certainly :)

trinity/protocol/bcc/context.py
```diff
@@ -1,10 +1,14 @@
...
+from eth2.beacon.typing import Slot
 
 
 class BeaconContext(BasePeerContext):
 
-    def __init__(self, chain_db: BaseBeaconChainDB, network_id: int) -> None:
+    def __init__(self,
+                 chain_db: BaseBeaconChainDB,
+                 network_id: int,
+                 genesis_slot: Slot=Slot(0)) -> None:  # noqa: B008
```
hwwhww: Should make the default value to `GENESIS_SLOT` from config.

trinity/plugins/eth2/beacon/plugin.py
```diff
@@ -0,0 +1,119 @@
...
+    def on_ready(self, manager_eventbus: TrinityEventBusEndpoint) -> None:
+        if self.context.trinity_config.has_app_config(BeaconAppConfig):
+            self.start()
+
+    def do_start(self) -> None:
+
+        db = AtomicDB()
+        chain_db = BeaconChainDB(db)
+
+        genesis_state, genesis_block = create_mock_genesis(
```
hwwhww: Hmm, I think the mocking function shouldn't be called in codebase: (i) separate the tests and pro  (ii) too much overhead generate the genesis every time `do_start` is called. (iii) After the basic handshake tests, there's no validator in your genesis to perform state transition!

@cburgdorf  @ChihChengLiang 
Should we assign a certain `chain_db` to the `BeaconNodePlugin ` instance instead of implementing `db = AtomicDB(); chain_db = BeaconChainDB(db)` in `do_start`? If so, should genesis come from `chain_db`?

trinity/plugins/eth2/beacon/plugin.py
```diff
@@ -0,0 +1,119 @@
...
+class BeaconNodePlugin(BaseIsolatedPlugin):
+
+    @property
+    def name(self) -> str:
+        return "Beacon Node"
+
+    def configure_parser(self, arg_parser: ArgumentParser, subparser: _SubParsersAction) -> None:
+        arg_parser.add_argument(
+            "--bootstrap_nodes",
+            help="enode://node1@0.0.0.0:1234,enode://node2@0.0.0.0:5678",
```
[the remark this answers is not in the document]
ChihChengLiang: These are examples to specify `bootstrap_nodes`. So you can run something like 
```
trinity-beacon --bootstrap_nodes=enode://c289557985d885a3f13830a475d649df434099066fbdc840aafac23144f6ecb70d7cc16c186467f273ad7b29707aa15e6a50ec3fde35ae2e69b07b3ddc7a36c7@0.0.0.0:30303
```

trinity/plugins/eth2/beacon/plugin.py
cburgdorf: The database is running in it's own separate process (because only one process can have an open connection to it), so what this plugin needs to do to get a handle is to call 

`create_db_consumer_manager`

https://github.com/ethereum/trinity/blob/2577ecbdab0a8ef0e8c1f6e33d8c22a6c60e4d1c/trinity/db/beacon/manager.py#L39-L53

This will return a `DbManager` instance that you can call `get_chaindb()` on to get a proxy to the chain_db that you can work with across process boundaries.

Notice that the *serving side* of this manager is already created when the beacon node boots.

https://github.com/ethereum/trinity/blob/2577ecbdab0a8ef0e8c1f6e33d8c22a6c60e4d1c/trinity/main_beacon.py#L141

Regarding genesis handling, for Eth1, we create the db with a genesis state when the server part of the DbManager is created.

https://github.com/ethereum/trinity/blob/2577ecbdab0a8ef0e8c1f6e33d8c22a6c60e4d1c/trinity/db/eth1/manager.py#L23-L30

We are still missing that part for the beacon node as the comment here hints.

https://github.com/ethereum/trinity/blob/2577ecbdab0a8ef0e8c1f6e33d8c22a6c60e4d1c/trinity/db/beacon/manager.py#L16-L21

Could we handle it the same way and initialize it with a genesis state there?

ChihChengLiang: @cburgdorf Thanks for the clear walkthrough here, it's a lifesaver! One thing I'm still uncertain is when are we using the async methods from `AsyncBeaconChainDBProxy`?

trinity/db/beacon/chain.py
```diff
@@ -121,6 +125,23 @@ class AsyncBeaconChainDBPreProxy(BaseAsyncBeaconChainDB):
...
+    get_block_by_root = sync_method('get_block_by_root')
+    get_slot_by_root = sync_method('get_slot_by_root')
+    get_score = sync_method('get_score')
+    block_exists = sync_method('block_exists')
+    persist_block_chain = sync_method('persist_block_chain')
+    get_state_by_root = sync_method('get_state_by_root')
+    persist_state = sync_method('persist_state')
+    exists = sync_method('exists')
+    get = sync_method('get')
+
```
ChihChengLiang: The bcc/peer.py consumes `chaindb`, a `AsyncBeaconChainDBProxy` produced by the manager `create_db_consumer_manager` creates, it needs those synchronous methods to work.
https://github.com/ethereum/trinity/blob/master/trinity/protocol/bcc/peer.py#L55

trinity/config.py
```diff
@@ -570,12 +586,80 @@ def nodedb_path(self) -> Path:
         return config.with_app_suffix(config.data_dir / "nodedb")
 
 
+class BeaconChainConfig:
```
ChihChengLiang: Now almost all of the beacon chain related configurations and customizations coming into this class. In a future refactor can write a class method `from_preconfigured_network` to organize them better.

ChihChengLiang:
And a little script to play with it.
```
mkdir /tmp/alice
mkdir /tmp/bob
```

Terminal A
```
export PYTHONWARNINGS="ignore::DeprecationWarning"
trinity-beacon --trinity-root-dir=/tmp/alice --beacon-nodekey=6b94ffa2d9b8ee85afb9d7153c463ea22789d3bbc5d961cc4f63a41676883c19 -l debug
```

Terminal B
```
export PYTHONWARNINGS="ignore::DeprecationWarning"
trinity-beacon --trinity-root-dir=/tmp/bob --port=556 --beacon-nodekey=f5ad1c57b5a489fc8f21ad0e5a19c1f1a60b8ab357a2100ff7e75f3fa8a4fd2e --bootstrap_nodes=enode://c289557985d885a3f13830a475d649df434099066fbdc840aafac23144f6ecb70d7cc16c186467f273ad7b29707aa15e6a50ec3fde35ae2e69b07b3ddc7a36c7@0.0.0.0:30303  -l debug
```

trinity/plugins/eth2/beacon/plugin.py
cburgdorf: Happy to hear it was helpful!

>One thing I'm still uncertain is when are we using the async methods from `AsyncBeaconChainDBProxy`?

Most things that consume the db proxy should be written with `asyncio` using a single threaded, event-looped model. Under that model, it's the cardinal sin to block the event loop. When you read/write things from the DB, that's a blocking operation which means sync methods to interact with the db are a no-go when using an event loop. So, as a rule of thumb, the *async* methods should always be used (unless there are methods that don't really block but everything talking to the db qualifies as long-blocking).

This leads us to the question, why do we have the `sync` methods on the `BaseAsyncBeaconChainDB` at all? For Eth1 `BaseAsyncChainDB` inherits from `BaseChainDB` which means we have to implement the synchronous abstract methods to satisfy the rules. But I can't tell if there's a good reason to have `BaseAsyncChainDB` inherit from `BaseChainDB` in the first place. From the top of my head, it sees that we could refactor the `BaseAsyncChainDB` to **not inherit** from the `BaseChainDB` and then **only provide async** methods.

cburgdorf: @ChihChengLiang I've send a PR to fix it for the Eth1 side. Thanks for bringing it to my attention! https://github.com/ethereum/trinity/pull/439

jannikluhn:
Doesn't work for me :(

Some of the exceptions don't seem related to this PR. It could also be that my network here is a bit restrictive and blocks some of the necessary ports. For completeness though, here's the red part of the log output: https://gist.github.com/jannikluhn/2f7d65b12fad1dbb8708455ece507f9c

tests/core/p2p-proto/bcc/test_requests.py
```diff
@@ -194,22 +198,22 @@ def finalizer():
...
-    alice.sub_proto.send_get_blocks(3, 10, request_id=5)
+    alice.sub_proto.send_get_blocks(genesis.slot + 3, 10, request_id=5)
     response = await response_buffer.msg_queue.get()
 
     assert isinstance(response.command, BeaconBlocks)
     assert response.payload["request_id"] == 5
     blocks = tuple(ssz.decode(block, BeaconBlock) for block in response.payload["encoded_blocks"])
     assert len(blocks) == 5
-    assert [block.slot for block in blocks] == [3, 4, 5, 6, 7]
+    assert [block.slot for block in blocks] == [genesis.slot + s for s in (3, 4, 5, 6, 7)]
```
jannikluhn:
```suggestion
    assert [block.slot for block in blocks] == [genesis.slot + s for s in [3, 4, 5, 6, 7]]
```

for consistency

trinity/initialization.py
```diff
@@ -113,6 +131,16 @@ def initialize_database(chain_config: ChainConfig,
         chain_config.initialize_chain(base_db)
 
 
+def initialize_beacon_database(chain_config: BeaconChainConfig,
+                               chaindb: BaseBeaconChainDB,
+                               base_db: BaseAtomicDB,
+                               block_class: Type[BaseBeaconBlock]) -> None:
+    try:
```
jannikluhn: Shouldn't this throw an error if the DB is already initialized?

ChihChengLiang: Not necessarily, the `manager.py` checks `is_beacon_database_initialized` before calling the `initialize_beacon_database`. Even if `initialize_beacon_database` is called after the db is initialized, `chaindb.get_canonical_head(block_class)` should not raise `CanonicalHeadNotFound` exception and thus not initializing again.

trinity/plugins/eth2/beacon/plugin.py
ChihChengLiang: Looks like syncer can exit on its own. Removed `asyncio.ensure_future(exit_with_service(syncer))` for now.

ChihChengLiang:
Merging for now to unblock other tasks. Post PR comments are welcomed.

Issue #293, opened 2019-02-19
Libp2p daemon integration


mhchia:
### What is wrong?

Per the libp2p requirement in https://github.com/ethereum/trinity/issues/281, we will require the code to communicate with [libp2p daemon](https://github.com/libp2p/go-libp2p-daemon). I have some questions for future development:
- Should I make the PRs to another branch, e.g. libp2p-daemon?
- To interact with libp2p-daemon, I use [mhchia/py-libp2p-daemon-bindings](https://github.com/mhchia/py-libp2p-daemon-bindings) to do it. Should I move this package under the package `p2p`, or somewhere else, instead of import it as a third-party package?
- After roughly tracing the code, it seems I need to modify the package `p2p`(for the `Peer`, `PeerPool`, etc...), and also to add data structures for the shard chains. Does it make sense to add these data structures under `eth2`, e.g. a new directory `shards`?
- Does it make sense to add integration tests, with the go-libp2p-daemon spun up?

Thanks a lot!

cburgdorf:
Cool! Here are my thoughts:

>* Should I make the PRs to another branch, e.g. libp2p-daemon?

PRs against `master` seem fine to me.

>* Should I move this package under the package `p2p`, or somewhere else, instead of import it as a third-party package?

Mmh. Depends, would it make our life much harder if we used it as a `pip` dependency? Does it have to be updated very frequently?

>* I need to modify the package `p2p`(for the `Peer`, `PeerPool`, etc...),

You'll probably need some modifications directly in `p2p` but I *think* I'd recommend we try to keep things mostly out of the `p2p` package what doesn't absolutely need to be there. Depending on the scope (datastructure needed in `eth2` vs just trinity machinery) things should either go directly into `eth2` or somewhere in trinity. Where the *somewhere* in Trinity is a bit up in the air if you ask me. If it's stuff that can sit somewhere symmetrical with some existing types (e.g. `trinity.protocol.*`) then it should go there, if it is entirely new stuff then maybe something like `trinity.libp2p` could serve as a temporary bucket for the lack of a better place?

> * Does it make sense to add integration tests, with the go-libp2p-daemon spun up?

Sounds useful!

I'm sure the rest of us have thoughts on this as well.

mhchia:
> Mmh. Depends, would it make our life much harder if we used it as a `pip` dependency? Does it have to be updated very frequently?

The bindings code is subject to changes. In this case, it seems more suitable to move the code to trinity's codebase?

[...]

PR #309, opened 2019-02-25
Add libp2p daemon bindings

mhchia:
### What was wrong?
https://github.com/ethereum/trinity/issues/293

### How was it fixed?
This is the first step of the integration. Next step is to add the logics for Phase 0 beacon chain, based on the bindings. Sorry for this huge PR and possibly a mess structure. I would love to change it according to the feedback. Thanks a lot!

#### Done
- Added the [bindings codebase](https://github.com/mhchia/py-libp2p-daemon-bindings/) to trinity repo.
- Changed the tests to work with both py3.6 and py3.7: https://github.com/mhchia/py-libp2p-daemon-bindings/pull/27
- Added the tests to tox
- Added the config in circle CI
  - Added cache for `libp2p-daemon`, avoiding repeated downloading and building for the same version.
- Wrote a [script](https://gist.github.com/mhchia/977c8353ff4aea02d295dfa0b8ed7a56)(just a helper, not included in this PR) for moving p2pclient code to trinity

#### Todo
- [x] Fix mypy errors

#### Cute Animal Picture


tests/libp2p/p2pclient/test_serialization.py
```diff
@@ -0,0 +1,55 @@
...
+import pytest
+
+
+from libp2p.p2pclient.serialization import (
+    read_varint,
+    write_varint,
+)
+
+
+def test_serialize():
```
mhchia: Just found out the empty tests. Will fix them.

tests/libp2p/p2pclient/test_serialization.py
```diff
@@ -0,0 +1,55 @@
...
+    pass
+
+
+def test_pb_readwriter():
+    pass
```
mhchia: ditto

tests/libp2p/p2pclient/test_serialization.py
```diff
@@ -0,0 +1,55 @@
...
+    async def read_byte(s):
+        data = s.read(1)
+        return data[0]
+
+    result = await read_varint(s, read_byte)
+    assert value == result
+
+
+def test_read_varint_overflow():
+    pass
```
mhchia: ditto

mhchia:
py{36,37}-libp2p and py{36,37}-lint fail. Will fix them accordingly.

libp2p/p2pclient/p2pclient.py
```diff
@@ -0,0 +1,617 @@
...
+        self.logger.info("New incoming stream: %s", stream_info)
+        try:
+            handler = self.handlers[stream_info.proto]
+        except KeyError as e:
+            # should never enter here... daemon should reject the stream for us.
+            writer.close()
+            raise DispatchFailure(e)
+        await handler(stream_info, reader, writer)
+
+    async def _write_pb(self, writer: asyncio.StreamWriter, data_pb) -> None:
```
pipermerriam: This could be a `staticmethod`

libp2p/p2pclient/p2pclient.py
```diff
@@ -0,0 +1,617 @@
...
+        await read_pbmsg_safe(reader, resp)
+        writer.close()
+        raise_if_failed(resp)
+        peer_id_bytes = resp.identify.id
+        maddrs_bytes = resp.identify.addrs
+
+        maddrs = []
+        for maddr_bytes in maddrs_bytes:
+            maddr = Multiaddr(binascii.hexlify(maddr_bytes))
+            maddrs.append(maddr)
```
pipermerriam: 1. could we change the peer to store these as a `tuple` since they don't appear to be mutated.
2. can you change this to a comprehension to avoid the mutability.

libp2p/p2pclient/p2pclient.py
```diff
@@ -0,0 +1,617 @@
...
+    protocols.P_UNIX,
+)
+
+
+def parse_conn_protocol(maddr: Multiaddr) -> int:
+    proto_codes = set([
+        proto.code
+        for proto in maddr.protocols()
+    ])
+    supported_proto_codes = set(_supported_conn_protocols)
```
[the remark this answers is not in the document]
mhchia: You're right. Thanks for the pointer. It seems Iterable is accepted.

libp2p/p2pclient/p2pclient.py
mhchia: Yeah, no problem. Have changed them.

libp2p/p2pclient/p2pclient.py
mhchia: Thanks for pointing it out. I just changed it.

.circleci/config.yml
```diff
@@ -87,6 +87,66 @@ geth_steps: &geth_steps
...
+            make bin
+            mkdir -p $P2PD_DIR
+            cp `which p2pd` $P2PD_BINARY
+            cd -
+          fi
+          sudo ln -s $P2PD_BINARY /usr/local/bin/p2pd
+    - save_cache:
+        paths:
+          - ~/.p2pd
+        key: cache-v1-{{ arch }}-p2pd-{{ checksum "p2pd-version.txt" }}
```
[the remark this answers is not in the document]
mhchia: Agree! It seems a lot cleaner with your approach. I've moved them to ./.circleci/install_p2pd.sh and change the code in config.yml accordingly.

mhchia:
@pipermerriam Thanks a lot for the review! I just fixed the CI and improved the PR according to the feedback. I'm still reorganizing the leftover empty tests, and trying to add types for the protobuf messages.

mhchia:
Except for the typing for the specified protobuf messages e.g.  `DHTRequest`, this PR should be ready for review. Even with mypy-protobuf, there are syntax errors in the .pyi files. I currently use `google.protobuf.message.Message` as the protobuf message interface, and the typing on this interface works.

libp2p/p2pclient/pb/Makefile
```diff
@@ -0,0 +1,17 @@
+p2pd_path = ${GOPATH}/src/github.com/libp2p/go-libp2p-daemon
+p2pd_pb_path = $(p2pd_path)/pb
+p2pd_pbs = $(shell find $(p2pd_pb_path) -name *.proto)
+p2p_crypto_pb_path = ${GOPATH}/src/github.com/libp2p/go-libp2p-crypto/pb
+p2p_crypto_pbs = $(shell find $(p2p_crypto_pb_path) -name *.proto)
+
+all: p2pd_pb_pys p2p_crypto_pys
```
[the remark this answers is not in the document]
mhchia: > it seems that you have taken the (imo better) approach to just check in the compiled files.... in which case i would almost suggest we leave this Makefile out of the PR -- depends on if we expect the PB defs to change or not, i would say.

Agree. It also makes sense to me to move the Makefile out of the PR.

mhchia:
About the typing for pbs, I turned out to generate the mypy extensions(*.pyi) through [mypy-protobuf](https://github.com/dropbox/mypy-protobuf), and modified the `*.pyi`s a bit to get rid of the syntax error(due to `PSMessage.from` conflict with the keyword in Python), along with several issues. Will update them soon in this PR.

libp2p/p2pclient/p2pclient.py
```diff
@@ -0,0 +1,627 @@
...
+        resp = p2pd_pb.Response()
+        await read_pbmsg_safe(reader, resp)
+        writer.close()
+        raise_if_failed(resp)
+
+        topics = tuple(resp.pubsub.topics)
+        return topics
+
+    # FIXME: name conflicts: `list_topic_peers` is originally `list_peers` in pubsub
+    async def list_topic_peers(self, topic: str) -> Tuple[PeerID, ...]:
```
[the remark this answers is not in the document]
ralexstokes: i think it is less readable w/ the thing prefixed but it isn't the worst thing in the world.

i saw you had an issue to break up each category into its own batch of functionality (file or class etc) which i think would make this more manageable, regardless of what the names are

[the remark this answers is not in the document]
mhchia: Do you mean this one https://github.com/mhchia/py-libp2p-daemon-bindings/issues/9? Thanks a lot for the reminder. It seems more reasonable to restructure `Client`, to `P2PClient`(or other names, managing the connections between the daemon), `Control`(for control system including `identify`, `connect`, ...), `DHT`, `ConnMgr`. Will do it!

mhchia:
Update:
- Added stubs for protobuf messages, generated by mypy-protobuf and modified to avoid errors. One possible issue in the future is, I removed the `from` field in `PSMessage` in [this commit](https://github.com/ethereum/trinity/pull/309/commits/478458022bac84f970eec60c2310204b2c55d3ce#diff-edcef2e8f7e554d21efab877bc4f63bdR421). This field makes us unable to access the field directly, since `from` is the keyword in Python. Though we can bypass this with `setattr` and `getattr`.
- Restructured the `Client`, separate them into components `Client`, `ControlClient`, `ConnectionManagerClient`, `DHTClient`, and `PubSubClient` in [this commit](https://github.com/ethereum/trinity/pull/309/commits/4d4bd6356b96d7cb576ae3d1bfbe0126e68bacbc)

@pipermerriam @ralexstokes  IMO they are ready for a second review. Please review the updates if you have time. Thanks a lot!

ralexstokes:
@mhchia i would leave in `from` -- i doubt we will get upstream changes just because of language-specific restrictions.

can you call the field something like `from_field` (or `sender` or ...) and patch the `.pyi` file as needed to do what we want?

ralexstokes:
@mhchia and i reviewed the breakup of the client -- what are you thinking in terms of using the client? it looks like we had one monolithic client and now have 3 or so clients that all (effectively) subclass `Client`. Do we make one `Client` and then instantiate each specialized `*Client` as needed?

I suspect we will have all clients at a given time -- in which case it seems cumbersome to have a user of these bindings to juggle three different clients and make them determine which one to use for which activity.

i think what would be nicer is something like:

```python
from libp2p.p2pclient import Client
daemon_multiaddr = ...
client = Client(daemon_multiaddr)
client.do_stuff()
....
```

where `do_stuff` is one of the union of all the methods on each of the `*Client` classes you have. this way I can just tell the client to "do what i mean" and then not worry about which particular client to use. the `Client` class is responsible for delegating to the proper `*Client` based on the method I call. thoughts?

mhchia:
@ralexstokes 
> can you call the field something like from_field (or sender or ...) and patch the .pyi file as needed to do what we want?

Sure!

> where do_stuff is one of the union of all the methods on each of the *Client classes you have. this way I can just tell the client to "do what i mean" and then not worry about which particular client to use. the Client class is responsible for delegating to the proper *Client based on the method I call. thoughts?

You're right. It will be nicer to use if they are aggregated into a single `Client`. Do you mean something like the following snippet?
```python
class Client:
    control_maddr: Multiaddr
    listen_maddr: Multiaddr
    control: ControlClient
    dht: DHTClient

    def __init__(self, control_maddr, listen_maddr):
        self.control_maddr = control_maddr
        self.listen_maddr = listen_maddr
        self.control = ControlClient(self)
        self.dht = DHTClient(self)
        ...

    async def connect(self, ...):
        return await self.control.connect(...)

    async def find_peer(self, ...):
        return await self.dht.connect(...)
```

If yes, what do you think if I revert the "Restructuring" back to the original monolithic `Client`, compared to the structure above? I'm not sure which one is better. I would like to split the logics, but IMO the structure above seems to become more complicated.(Otherwise, move the dependency of `*Client`(ControlClient, DHTClient) on the `open_connection` and `control_maddr` to another class, e.g. `Connection`(actually the `Client` currently in my code), which becomes

```python
class Client:
    conn: Connection
    control_maddr: Multiaddr
    listen_maddr: Multiaddr
    control: ControlClient
    dht: DHTClient

    def __init__(self, control_maddr, listen_maddr):
        self.control_maddr = control_maddr
        self.listen_maddr = listen_maddr
        self.conn = Connection(control_maddr)

        self.control = ControlClient(conn, listen_maddr)
        self.dht = ControlClient(conn)
        ...

    async def connect(self, ...):
        return await self.control.connect(...)

    async def find_peer(self, ...):
        return await self.dht.connect(...)
```
IMO reverting back to the original monolithic `Client` might be easier, since afterwards we might still have a huge class, e.g. `libp2p.Host` wrapping these logics up, and the reverting is easy to do as well. Really appreciate your feedback, would love to hear your thoughts further. And I'm sorry for the time wasted on this restructuring, since it might be unnecessary.

ralexstokes:
@mhchia no worries on feedback :) i always like to say the most permanent solution is a temporary one so its worth some cycles to make this nice as it is unclear exactly how long we will be using it ;)

this being said to your point, if you feel like its useable enough, then let's go ahead and merge it. if you are curious to what i would prefer in a perfect world, its the following:

`Client` is an interface to all of the `libp2p` functionality we could hope for. I can make one `Client` by passing it the multiaddr of the daemon and then as a user of that `Client` I can call any method on `Client` to do any of the functionality afforded by the libp2p daemon interface. If it just so happens that the various logics are divided up *internally*, that's fine as *just* a user of this object I'm not poking around the internals and hopefully will not have to ("i'm too busy building some cool p2p app").

so as a user it looks like:
```python
client = Client(daemon_multiaddr)
some_peer_id = client.connect(some_other_peer)
topic = client.get_topic_id_for_topic("some-topic", some_peer_id)
subscription_id = client.subscribe_to(topic)
# do some stuff on this pubsub net i've just joined...
```
(note: i'm making some of those calls up in lieu of referring the actual API but you should get the idea)

then the question becomes: as an author/maintainer of such an interface, what is a sane way to accomplish our job? i think we both agree we want to avoid the moral hazard of having large files so we turn to our methods of composition in Python. one common answer is subclassing but i think we agree that isn't a great fit here. so what i would suggest is that `Client` essentially just holds the connection state (the bare multiaddrs for informational purposes, the runtime connection object to actually effect our wishes on the world) and delegates to a bunch of functions (imported from sibling and/or child modules to `client.py`). Each function can take a `Connection` object (supplied by the calling `Client`) along w/ any additional parameters for the specific call and then do the thing we want it to do, returning any response/error to the caller.

Make sense? I can open a PR to sketch out what I'm thinking if that is helpful :)

mhchia:
@ralexstokes Big thanks for the detailed explanation! I totally agree with your idea. May I try to interpret what your ideal structure looks like?

1. `Connection` holds the `control_maddr`, and expose a function for others to call, for others to open a TCP stream with. Components including `ControlClient`, `DHTClient`, ... have `Connection` object as the attribute and use it to create connection.
```python
class Connection:
    control_maddr: Multiaddr

    def open_connection(...)
        ...

class ControlClient:
    conn: Connection
    listen_maddr: Multiaddr

    def __init__(self, conn: Connection, listen_maddr: Multiaddr):
        self.conn = conn
        self.listen_maddr = listen_maddr

    def connect(...):
        reader, writer = await self.conn.open_connection(...)
        ...

    def listen(...):
        await asyncio.start_server(listen_maddr, ...)

class DHTClient:
    conn: Connection

    def __init__(self, conn: Connection):
        self.conn = conn

    def find_peer(...):
        ...

class Client:
    conn: Connection
    listen_maddr: Multiaddr
    control: ControlClient
    dht: DHTClient
    ...

    def __init__(self, control_maddr, listen_maddr):
        self.conn = Connection(control_maddr)

        self.control = ControlClient(conn, listen_maddr)
        self.dht = DHTClient(conn)
        ...

    async def connect(self, ...):
        return await self.control.connect(...)

    async def find_peer(self, ...):
        return await self.dht.connect(...)
```

2. Still have the same `Connection` as (1), but `Connection` object is passed to the functions in the components when the functions are called. IMO compared to (1), this is more flexible to swap between different `Connection`(though I think it is not common), and the object doesn't need to hold the connection state.
```python

class Connection:
    control_maddr: Multiaddr

    def open_connection(...)
        ...

class ControlClient:
    listen_maddr: Multiaddr

    def __init__(self, listen_maddr: Multiaddr):
        self.listen_maddr = listen_maddr

    def connect(self, conn: Connection, ...):
        reader, writer = await conn.open_connection(...)
        ...

    def listen(self, ...):
        await asyncio.start_server(listen_maddr, ...)

class DHTClient:

    def find_peer(self, conn: Connection, ...):
        ...
        reader, writer = await conn.open_connection(...)
        ...

class Client:
    conn: Connection
    listen_maddr: Multiaddr
    control: ControlClient
    dht: DHTClient
    ...

    def __init__(self, control_maddr, listen_maddr):
        self.conn = Connection(control_maddr)

        self.control = ControlClient(listen_maddr)
        self.dht = DHTClient()
        ...

    async def connect(self, ...):
        return await self.control.connect(self.conn, ...)

    async def find_peer(self, ...):
        return await self.dht.connect(self.conn, ...)
```

3, 4. The similar strucutre as 1, 2, respectively. But move the `listen_maddr`, and `listen` method from `ControlClient` to `Connection`. This moves all logic related to connections to `Connection`. However, since the `await asyncio.start_server(callback, host=host, port=port)` needs the callback function, which should finally call the function registered in `ControlClient`. Seems a little bit more complicated.

Which structure is more similar to your idea? If the none of them above are not what you're thinking about, please feel free to make a PR or directly push to this PR(my branch), which is feasible IIRC. Thanks a lot for the feedback:)

mhchia:
@ralexstokes May I merge this PR and open another issue to track the interface of `Client`? I would love to go with your idea. Just want to proceed to implement `BeaconNode` based on this PR and https://github.com/ethereum/trinity/pull/403. Really appreciate your suggestion:)

ralexstokes:
@mhchia yeah sorry for the delay! i think i was imagining something like (1) -- but i would say let's go ahead and merge this in.... we can always aim to improve the implementation and to be honest I think work on #403 will help inform us here :)

mhchia:
@ralexstokes 
Nevermind! 

> i think i was imagining something like (1)

Yeah, I think (1) makes sense, will change to it.

Btw., may I ask if this PR is causing errors or anything wrong? It is reverted now. I would like to revise it if anything is inappropriate. Thank you.

ralexstokes:
@mhchia yeah -- im seeing what was going on here -- it could have been some unrelated (?) CI stuff....

Issue #306, opened 2019-02-25
Rework withdrawability


hwwhww:
### What is wrong?

Spec PRs: 
1. Bring forward changes to withdrawability from phase 1: https://github.com/ethereum/eth2.0-specs/pull/615
2. Turned slashed and initiated_exit into booleans: https://github.com/ethereum/eth2.0-specs/pull/658
### How can it be fixed
1. Rework `ValidatorRecord` fields to latest version: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#validator
2. Remove status flags
3. Note: ignore `Transfer` functionality, for now, that could be implemented in another PR. IMO it's a lower priority.

Issue #223, opened 2019-01-23
Implement "per-epoch processing - validator registry" section


hwwhww:
### What is wrong?

Spec: per-epoch processing - validator registry https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#validator-registry

We don't have this section yet. 

It could be 4 parts:
1. Common logic: "update previous_epoch_calculation_slot and previous_epoch_start_shard.... " part
2. Conditions: "If the following are satisfied... " part
3. "If a validator registry update does **not** happen"
4. "If a validator registry update does happen"

*For MVP*, at least we need the part 1-3.

### How can it be fixed

- In `eth2/beacon/state_machines/forks/serenity/epoch_processing.py`, add `process_validator_registry(state: BeaconState, block: BaseBeaconBlock, config: BeaconConfig) -> BeaconState`.
- Call `process_validator_registry` in `eth2.beacon.state_machines.forks.serenity.state_transitions. SerenityStateTransition.per_epoch_transition`:
```python
    def per_epoch_transition(self, state: BeaconState, block: BaseBeaconBlock) -> BeaconState:
        # TODO: state = process_receipt_roots(state, block, self.config)
        # TODO: state = process_justification(state, block, self.config)
        # TODO: state = process_rosslinks(state, block, self.config)
        # TODO: state = process_rewards_and_penalties(state, block, self.config)
        # TODO: state = process_ejections(state, block, self.config)
        state = process_validator_registry(state, block, self.config)
        # TODO: state = process_final_updates(state, block, self.config)

        return state
```

### Note: recent incoming update: https://github.com/ethereum/eth2.0-specs/pull/476 will affect it.

hwwhww:
including part 1~3 into #225

hwwhww:
working on `update_validator_registry`

PR #314, opened 2019-02-26
Implement `update_validator_registry` and seek for a quick review

hwwhww:
### What was wrong?
Part of https://github.com/ethereum/trinity/issues/223 `update_validator_registry`

### How was it fixed?
I found that I need to finish #306 to upgrade it to the latest version.

But meanwhile, I implemented two versions of `update_validator_registry` and hope to get some suggestions/advices:

1. **`update_validator_registry_2`**: `update_validator_registry_2` is similar to the spec: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#validator-registry-and-shuffling-seed-data:
    ```python
    # Activate validators within the allowable balance churn
    # Iterate the validator registry, see if it's ready to be activated
    # Check the balance churn would be within the allowance
    # If so, activate the validator

    # Exit validators within the allowable balance churn
    # Iterate the validator registry, see if it's ready to be exited
    # Check the balance churn would be within the allowance
    # If so, exit the validator
    ```

2. `update_validator_registry`: there are some logic that could be refactored in `update_validator_registry_2`, so I made `update_validator_registry`… but after I finished it, it’s still not ideal since the arguments are not exactly the same.

What do you think about `update_validator_registry` v.s. `update_validator_registry_2`?
Any feedback would be appreciated. 🙏

#### Cute Animal Picture


eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -812,8 +822,206 @@ def _check_if_update_validator_registry(state: BeaconState,
     return True, num_shards_in_committees
 
 
-def update_validator_registry(state: BeaconState) -> BeaconState:
-    # TODO
+def is_ready_to_active(state: BeaconState,
```
ralexstokes: nitpick (spelling): s/active/activate/

may want to hold off if you think this function will not end up in the final PR

ralexstokes:
@hwwhww i prefer `update_validator_registry`. part of this is personal preference but i would rather orchestrate a bunch of small functions, than have to manage one big (monolithic) function.

the primary benefits come via ease of understanding/comprehension and ease of testing (the more names you have for a thing, the more control you have over it; here we go from no names for the subroutines to having at least one name.) the main argument against this implementation is some perceived difficulty in understanding. i would contend that there is a bit more upfront grokking required ("ok, what are these functions, how are they being used in the callee") but once that cost is incurred, all downstream maintenance is _much_ easier.

since you seem concerned about the `churn_fn_kwargs` not matching (correct me if I'm misreading your comment about "not exactly the same), i'll suggest another possible route to just pass a closure containing the necessary actions to perform. we can achieve uniformity by abstracting out the differences w/ a trusty `lambda`.

something like:

```python
# in `update_validator_registry`
churn_fn = lambda state, index: activate_validator(state, index, False, config.GENESIS_EPOCH, config.SLOTS_PER_EPOCH, config.ACTIVATION_EXIT_DELAY)
state = churn_validators(
       state=state,
       config=config,
       check_should_churn_fn=is_ready_to_active,
       churn_fn=churn_fn,
       current_epoch=current_epoch,
       max_balance_churn=max_balance_churn,
   )
```
you could do something similar w/ `functools.partial` but referring to that implementation it seems like we would need to change the parameters for constants from positional args to keyword args...

we could also make this into a "factory function" but i don't think it's necessary here given how simple this closure is....

( but for completeness, by factory function i just mean something like:

```python
def mk_f(f, args):
    @functools.wraps(f)
    def g(state, index):
        f(state, index, *args)
    return g
```
)

btw i think what is already here is fine, altho i was asking myself at first why duplicate some of the config constants

tests/eth2/beacon/state_machines/forks/test_serenity_epoch_processing.py
```diff
@@ -913,6 +920,68 @@ def test_check_if_update_validator_registry(genesis_state,
...
+                                   n_validators_state,
+                                   config,
+                                   slots_per_epoch,
+                                   update_validator_registry_fn):
+    validator_registry = list(n_validators_state.validator_registry)
+    activating_index = n
+    exiting_index = 0
+
+    activating_validator = ValidatorRecord.create_pending_validator(
+        b'\x10' * 48,
```
djrtwo: Use positional args here

hwwhww:
@ralexstokes @djrtwo  thanks for the feedback! I like `update_validator_registry` more after applying @ralexstokes's suggestion! Choosing `update_validator_registry` now and it's also easy to add `update_validator_registry_2` back if there's some radical change from the spec... 😬

Issue #308, opened 2019-02-25
Crash during fast sync


cburgdorf:
### What is wrong?

I'm pretty sure we already had this reported but I couldn't find it. This error is coming up *very* often during fast sync for me. 

```
126170: WARNING  02-24 23:14:07   FastChainBodySyncer  Task <coroutine object FastChainBodySyncer._launch_prerequisite_tasks at 0x7f4868d49830> finished unexpectedly: No header with hash 0xd50fcbd93c68d777d10c9e1033f001f2c9e6e1aa0137dbbf08021965297905a3 found
126171-   DEBUG  02-24 23:14:07   FastChainBodySyncer  Task failure traceback
126172-trinity._utils.mp.RemoteTraceback: 
126173-"""
126174-Traceback (most recent call last):
126175-  File "/home/ubuntu/trinity/venv/lib/python3.6/site-packages/eth/db/header.py", line 167, in _get_block_header_by_hash
126176-    header_rlp = db[block_hash]
126177-  File "/home/ubuntu/trinity/venv/lib/python3.6/site-packages/eth/db/backends/level.py", line 54, in __getitem__
126178-    raise KeyError(key)
126179-KeyError: b'\xd5\x0f\xcb\xd9<h\xd7w\xd1\x0c\x9e\x103\xf0\x01\xf2\xc9\xe6\xe1\xaa\x017\xdb\xbf\x08\x02\x19e)y\x05\xa3'
126180-
126181-During handling of the above exception, another exception occurred:
126182-
126183-Traceback (most recent call last):
126184-  File "/home/ubuntu/trinity/trinity/_utils/mp.py", line 68, in wrapper
126185-    return attr(*args, **kwargs)
126186-  File "/home/ubuntu/trinity/venv/lib/python3.6/site-packages/eth/db/header.py", line 156, in get_block_header_by_hash
126187-    return self._get_block_header_by_hash(self.db, block_hash)
126188-  File "/home/ubuntu/trinity/venv/lib/python3.6/site-packages/eth/db/header.py", line 170, in _get_block_header_by_hash
126189-    encode_hex(block_hash)))
126190-eth.exceptions.HeaderNotFound: No header with hash 0xd50fcbd93c68d777d10c9e1033f001f2c9e6e1aa0137dbbf08021965297905a3 found
126191-"""
126192-
126193-The above exception was the direct cause of the following exception:
126194-
126195-Traceback (most recent call last):
[...]

PR #321, opened 2019-02-26
Log a lot more info when header parent is missing

carver:
### What was wrong?

There is a pretty mysterious bug related to syncing headers, where a header comes in from a peer when its parent is not in the DB: #308 

### How was it fixed?

Added some logging to collect relevant info about the incoming invalid header.

#### Cute Animal Picture


carver:
This got cherry-picked into #236

PR #330, opened 2019-02-27
Ensure wait_iter handles StopAsyncIteration properly

cburgdorf:
### What was wrong?

Currently, when we run `trinity` and then stop it using `CTRL+C` we'll see three of these errors.

```
   ERROR  02-27 17:01:55               asyncio  Task exception was never retrieved
future: <Task finished coro=<async_generator_asend()> exception=StopAsyncIteration()>
StopAsyncIteration 
    INFO  02-27 17:01:55            FullServer  Closing server...
 WARNING  02-27 17:01:55  ETHHeaderChainSyncer  Tried to cancel <trinity.protocol.eth.sync.ETHHeaderChainSyncer object at 0x7f8155e07358>, but it was already cancelled
    INFO  02-27 17:01:55         PluginManager  Shutting down PluginManager with scope <class 'trinity.extensibility.plugin_manager.SharedProcessScope'>
    INFO  02-27 17:01:55         PluginManager  Stopping plugin: Syncer Plugin
    INFO  02-27 17:01:55         PluginManager  Successfully stopped plugin: Syncer Plugin
   ERROR  02-27 17:01:55               asyncio  Task exception was never retrieved
future: <Task finished coro=<async_generator_asend()> exception=StopAsyncIteration()>
StopAsyncIteration
   ERROR  02-27 17:01:55               asyncio  Task exception was never retrieved
future: <Task finished coro=<async_generator_asend()> exception=StopAsyncIteration()>
StopAsyncIteration
```

These errors appeared first when we started to use `wait_iter` to consume the `event_bus.stream(...)` API. The `wait_iter` is needed to ensure we stop streaming when a cancel token is triggered. However, it appears there's a problem where we don't handle the `StopAsyncIteration` exception properly.

We can get a bit more info if we run `PYTHONASYNCIODEBUG=1 trinity` which will then expand the error message (of a single error of these) into this.

```
   ERROR  02-27 17:06:29               asyncio  Task exception was never retrieved
future: <Task finished coro=<async_generator_asend()> exception=StopAsyncIteration() created at /home/cburgdorf/Documents/hacking/ef/trinity/venv/lib/python3.6/site-packages/cancel_token/token.py:126>
source_traceback: Object created at (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.6/multiprocessing/spawn.py", line 105, in spawn_main
    exitcode = _main(fd)
  File "/usr/lib/python3.6/multiprocessing/spawn.py", line 118, in _main
    return self._bootstrap()
  File "/usr/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/home/cburgdorf/Documents/hacking/ef/trinity/trinity/_utils/profiling.py", line 31, in inner
    return fn(*args, **kwargs)
  File "/home/cburgdorf/Documents/hacking/ef/trinity/trinity/_utils/logging.py", line 161, in inner
    return fn(*args, **inner_kwargs)
  File "/home/cburgdorf/Documents/hacking/ef/trinity/trinity/main.py", line 195, in launch_node
    loop.run_forever()
  File "/usr/lib/python3.6/asyncio/coroutines.py", line 126, in send
    return self.gen.send(value)
  File "/home/cburgdorf/Documents/hacking/ef/trinity/p2p/service.py", line 139, in _run_task_wrapper
    await awaitable
  File "/usr/lib/python3.6/asyncio/coroutines.py", line 110, in __next__
    return self.gen.send(None)
  File "/home/cburgdorf/Documents/hacking/ef/trinity/p2p/service.py", line 158, in _run_daemon_task_wrapper
    await awaitable
  File "/usr/lib/python3.6/asyncio/coroutines.py", line 110, in __next__
    return self.gen.send(None)
  File "/home/cburgdorf/Documents/hacking/ef/trinity/trinity/nodes/base.py", line 61, in handle_network_id_requests
    async for req in self.wait_iter(self.event_bus.stream(NetworkIdRequest)):
  File "/home/cburgdorf/Documents/hacking/ef/trinity/p2p/cancellable.py", line 66, in wait_iter
    timeout=timeout,
  File "/usr/lib/python3.6/asyncio/coroutines.py", line 110, in __next__
    return self.gen.send(None)
  File "/home/cburgdorf/Documents/hacking/ef/trinity/p2p/cancellable.py", line 20, in wait
  File "/usr/lib/python3.6/asyncio/coroutines.py", line 110, in __next__
    return self.gen.send(None)
  File "/home/cburgdorf/Documents/hacking/ef/trinity/p2p/cancellable.py", line 42, in wait_first
  File "/usr/lib/python3.6/asyncio/coroutines.py", line 110, in __next__
    return self.gen.send(None)
  File "/home/cburgdorf/Documents/hacking/ef/trinity/venv/lib/python3.6/site-packages/cancel_token/token.py", line 126, in cancellable_wait
    futures = [asyncio.ensure_future(a, loop=self.loop) for a in awaitables + (self.wait(),)]
  File "/home/cburgdorf/Documents/hacking/ef/trinity/venv/lib/python3.6/site-packages/cancel_token/token.py", line 126, in <listcomp>
    futures = [asyncio.ensure_future(a, loop=self.loop) for a in awaitables + (self.wait(),)]
StopAsyncIteration
```

### How was it fixed?

**Let's first make it clear that I believe the following is a hack and that my main motivation to open up this PR as is, is to fire up imagination so that some :unicorn: can show me how to do it properly** :sweat_smile: 

1. Previously in `wait_iter` we did this:

```
        while True:
            try:
                yield await self.wait(
                    aiter.__anext__(),
                    token=token,
                    timeout=timeout,
                )
            except StopAsyncIteration:
                break
```

However, it appears that we never catch the `StopAsyncIteration` this way.

2. This change moves the `__anext()__` into a helper function where we then catch the exception and trigger a special `CancelToken` .

3. The special token is then used in the `while` loop to decide whether it should move on or not

#### Cute Animal Picture

### DIFF
--- a/p2p/cancellable.py
+++ b/p2p/cancellable.py
@@ -35,10 +35,7 @@ class CancellableMixin:
 
         All pending futures are cancelled before returning.
         """
-        if token is None:
-            token_chain = self.cancel_token
-        else:
-            token_chain = token.chain(self.cancel_token)
+        token_chain = self._get_token_chain(token)
         return await token_chain.cancellable_wait(*awaitables, timeout=timeout)
 
     async def wait_iter(
@@ -57,13 +54,35 @@ class CancellableMixin:
 
         See :meth:`CancellableMixin.wait_first` for using arguments ``token`` and ``timeout``
         """
+        token_chain = self._get_token_chain(token)
         aiter = aiterable.__aiter__()
-        while True:
-            try:
-                yield await self.wait(
-                    aiter.__anext__(),
-                    token=token,
-                    timeout=timeout,
-                )
-            except StopAsyncIteration:
+
+        # A CancelToken, only to propagate back the StopAsyncIteration
+        aiter_token = CancelToken('__anext__')
+        while not token_chain.triggered and not aiter_token.triggered:
+            val = await self.wait(
+                self._handle_anext(aiter, aiter_token),
+                token=token,
+                timeout=timeout,
+            )
+
+            if aiter_token.triggered:
                 break
+            else:
+                yield val
+
+    async def _handle_anext(self,
+                            aiter: AsyncIterator[TReturn],
+                            aiter_token: CancelToken) -> TReturn:
+        """
+        Call __anext__ on an AsyncIterator[TReturn] and handle StopAsyncIteration exceptions
+        by triggering the passed aiter_token in order to propagate it back.
+        """
+        try:
+            return await aiter.__anext__()
+        except StopAsyncIteration:
+            aiter_token.trigger()
+            return None  # mypy wants this
+
+    def _get_token_chain(self, token: CancelToken = None) -> CancelToken:
+        return self.cancel_token if token is None else token.chain(self.cancel_token)
--- a/tests/p2p/test_service.py
+++ b/tests/p2p/test_service.py
@@ -70,6 +70,24 @@ async def test_cancel_exits_async_generator():
 
 
 @pytest.mark.asyncio
+async def test_wait_iter_stops_when_generator_is_done():
+    service = WaitService()
+    asyncio.ensure_future(service.run())
+
+    async def async_iterator():
+        yield 1
+        await asyncio.sleep(0.5)
+        yield 1
+
+    count = 0
+    async for val in service.wait_iter(async_iterator()):
+        count += val
+
+    assert count == 2
+    await service.cancel()
+
+
+@pytest.mark.asyncio
 async def test_service_tasks_do_not_leak_memory():
     service = WaitService()
     asyncio.ensure_future(service.run())


cburgdorf:
@carver I pinged *you* for a review here because you've written the `wait_iter` function. I know you are busy debugging #308 so don't feel pushed to review this any time soon. After all, it only fixes issues with unwinding Trinity which is a lower priority. Also, I'm almost certain there's a better approach how to solve it so take your time with this ;)

carver:
I expect `wait_iter()` to follow the same semantics as `wait()`, which is that it raises `OperationCancelled` so that nothing else in the method runs. I'm not certain that we've really nailed down the problem, either, because this test works before the PR:

```py
@pytest.mark.asyncio
async def test_wait_iter():
    service = WaitService()
    asyncio.ensure_future(service.run())

    async def generator():
        yield 1
        yield 2
        yield 3

    try:
        async for val in service.wait_iter(generator()):
            assert val == 1
            await service.cancel()
    except OperationCancelled:
        pass
    else:
        assert False, "Service must be cancelled during iteration"

    service = WaitService()
    asyncio.ensure_future(service.run())

    vals = [val async for val in service.wait_iter(generator())]
    assert vals == [1, 2, 3]
    await service.cancel()
    assert False, "check out logs"
```

(I added a failure at the end, so that pytest doesn't swallow the logs, and I could look for any warnings. None show up on master).

---

So that covers both simple cases: service cancelled mid-iteration, and normal iteration completion. Which other case are we missing? When the token is cancelled and the exception is raised, although that also seems to be handled here:

https://github.com/ethereum/asyncio-cancel-token/blob/135395a1a396c50731c03cf570e267c47c612694/cancel_token/token.py#L145-L149

^ tangent: why isn't this doing the cool inlining code thing?

---

I notice that if we're *not* cancelled yet, and get two results, then we're not cleaning out the exceptions the way we do when we are cancelled:
https://github.com/ethereum/asyncio-cancel-token/blob/135395a1a396c50731c03cf570e267c47c612694/cancel_token/token.py#L153

We're just ignoring the other items in `done`. Seems like we should probably iterate those other ones and access `exception()`. But I'd love to see the test case that reproduces the issue on master, first. Then we can fiddle around with trying to fix it.

cburgdorf:
>But I'd love to see the test case that reproduces the issue on master, first. Then we can fiddle around with trying to fix it.

Yes, I agree. It's a bit of a mystery to me why I can't come up with a test that nails it down. We clearly seem to never catch `StopAsyncIteration` with the current `wait_iter` implementation which causes all these nasty errors when Trinity shuts down but I haven't been able yet to isolate that in a clear test case.

carver:
Kept futzing around with it. I couldn't get a reproducible (isolated) test case yet and have to move on to other things. The two possibilities that come to mind are:

1. the service is not cancelled, and there might be two `done` futures, despite it being a `WAIT_FIRST` (recap of earlier theory
2. A `CancelledError` is raised, but somehow some of the futures got run first, and an `AsyncStopIteration` is in one of those futures.

lithp:
Here's a test-case which reliably reproduces it:

```python
@pytest.mark.asyncio
async def test_wait_iter():
    service = WaitService()
    fut = asyncio.ensure_future(service.run())

    await asyncio.sleep(0)  # give the service a chance to start

    # a badly-behaved async generator
    async def error_on_cancel():
        try:
            await asyncio.sleep(10)
            for i in itertools.count:
                yield i
        except asyncio.CancelledError:
            pass

    async def wait_on_iter():
        async for i in service.wait_iter(error_on_cancel()):
            print(i)

    fut = asyncio.ensure_future(wait_on_iter())
    await asyncio.sleep(0)  # give the generator a chance to start
    fut.cancel()  # raise a cancellation

    await service.cancel()
    assert False
```

I've used a generator here but I believe the same problem would occur for any [coroutine which raised an exception after it had been cancelled] that was passed into `CancelToken.cancellable_wait`.

The unclean shutdown problem was caused by a method (`Endpoint.stream`) in lahja which did this. I've opened a PR which gets rid of almost all the messy lines during trinity's shutdown! https://github.com/ethereum/lahja/pull/156

cburgdorf:
That's awesome! Maybe contribute that test case to the PR that eventually upgrades to a fixed lahja version. Or, alternatively, maybe a similar test case could be added to the lahja PR that fixes the issue so that it gets caught early in the lahja library.

lithp:
I thought about this for a while but I couldn't think of a good thing to test for. I can certainly add a test that `Endpoint.stream()` doesn't throw an exception when cancelled, but it seems unlikely that problem'll be introduced again.

To make this problem easier to catch in the future:

- Maybe `CancelToken.cancellable_wait` could do something (I'm not sure what) to notice that cancelled coros are misbehaving, and print an additional warning which makes debugging this problem easier if it happens again in the future.

- We could add a lint rule. I think it's almost always wrong to catch `CancelledError` and not re-raise it, maybe we could add a check to our repos that required adding an `# ignore: swallows-cancellation` to the times where swallowing cancellation is really the behavior we want.

cburgdorf:
Yeah, I see what you mean. I like both suggestion. I'm not sure how hard it would be to write a flake8 extension that does this but I think it would be very useful :+1:

cburgdorf:
Closing this. The exceptions that this PR fixed have been long fixed meanwhile.

PR #332, opened 2019-02-27
Print client version strings in (debug-level) summary of remotes

veox:
### What was wrong?

Nothing.

### How was it fixed?

This gives an additional insight into remotes, and is a first step in debugging possible remote-specific failures.

The string-to-print part that will be retained in memory is limited in size; and is minimally "sanitised". First, tabs/newlines and other "whitespace" is stripped (always); if the string is still unprintable at this point, then illegal characters are "replaced" with question marks:

https://docs.python.org/3/howto/unicode.html#the-string-type

(search `replace`).

Result looks like:

```
    INFO  03-01 17:10:07           ETHPeerPool  Connected peers: 2 inbound, 0 outbound                                                                                                                                                         
    INFO  03-01 17:10:07           ETHPeerPool  Peer subscribers: 4, longest queue: ETHChainTipMonitor(0)                                                                                                                                      
   DEBUG  03-01 17:10:07           ETHPeerPool  == Peer details ==                                                                                                                                                                             
   DEBUG  03-01 17:10:07           ETHPeerPool  ETHPeer <Node(0x19ac@122.224.9.37)>: uptime=0:00:12:56, received_msgs=3001, most_received=Transactions (cmd_id=18)(2657)                                                                       
   DEBUG  03-01 17:10:07           ETHPeerPool  client_version_string='Geth/v1.8.11-stable/linux-amd64/go1.9.2'
   DEBUG  03-01 17:10:07           ETHPeerPool      BlockBodies: msgs=110  items=103249  rtt=1.50/4.77/0.97  ips=811.37318  timeouts=3  quality=95                                                                                             
   DEBUG  03-01 17:10:07           ETHPeerPool      BlockHeaders: msgs=32  items=5954  rtt=4.31/1.34/0.33  ips=365.00735  timeouts=0  quality=80                                                                                               
   DEBUG  03-01 17:10:07           ETHPeerPool      NodeData: None
   DEBUG  03-01 17:10:07           ETHPeerPool      Receipts: msgs=62  items=119995  rtt=3.11/6.11/1.25  ips=939.43918  timeouts=1  quality=94
   DEBUG  03-01 17:10:07           ETHPeerPool  ETHPeer <Node(0x1835@47.95.231.69)>: uptime=0:00:12:55, received_msgs=1010, most_received=Transactions (cmd_id=18)(609)                 
   DEBUG  03-01 17:10:07           ETHPeerPool  client_version_string='Geth/v1.8.8-stable-2688dab4/linux-amd64/go1.10'
   DEBUG  03-01 17:10:07           ETHPeerPool      BlockBodies: msgs=140  items=141741  rtt=0.96/2.86/0.56  ips=1220.73794  timeouts=0  quality=99                                                                                            
   DEBUG  03-01 17:10:07           ETHPeerPool      BlockHeaders: msgs=28  items=4994  rtt=5.39/1.08/0.26  ips=346.78782  timeouts=1  quality=71                                                                                               
   DEBUG  03-01 17:10:07           ETHPeerPool      NodeData: None                                                                                                                                                                             
   DEBUG  03-01 17:10:07           ETHPeerPool      Receipts: msgs=75  items=149558  rtt=1.81/3.41/0.61  ips=1481.79904  timeouts=1  quality=93                                                                                                
   DEBUG  03-01 17:10:07           ETHPeerPool  == End peer details ==
```

### Cute Animal Picture

Source: [petmd.com](https://www.petmd.com/dog/care/emotional-support-pets-explained)

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -195,7 +195,7 @@ class BasePeer(BaseService):
         self.privkey = privkey
 
         # The self-identifying string that the remote names itself.
-        self.client_version_string = None
+        self.client_version_string = ''
 
         # Networking reader and writer objects for communication
         self.reader = connection.reader
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -489,7 +489,7 @@ class BasePeer(BaseService):
         original_version = msg['client_version_string']
         client_version_string = original_version[:256] + ('...' if original_version[256:] else '')
         if client_version_string.isprintable():
-            self.client_version_string = client_version_string.strip('\t\n\r\x0b\x0c')
+            self.client_version_string = client_version_string.strip()
         else:
             self.client_version_string = repr(client_version_string)
 
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -486,10 +486,12 @@ class BasePeer(BaseService):
 
         # limit number of chars to be displayed, and try to keep printable ones only
         # MAGIC 256: arbitrary, "should be enough for everybody"
-        client_version_string = msg['client_version_string'][:256].strip('\t\n\r\x0b\x0c')
-        if not client_version_string.isprintable():
-            client_version_string = client_version_string.encode("utf-8", "replace").decode()
-        self.client_version_string = client_version_string
+        original_version = msg['client_version_string']
+        client_version_string = original_version[:256] + ('...' if original_version[256:] else '')
+        if client_version_string.isprintable():
+            self.client_version_string = client_version_string.strip('\t\n\r\x0b\x0c')
+        else:
+            self.client_version_string = repr(client_version_string)
 
         # Check whether to support Snappy Compression or not
         # based on other peer's p2p protocol version
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -194,6 +194,9 @@ class BasePeer(BaseService):
         # The private key this peer uses for identification and encryption.
         self.privkey = privkey
 
+        # The self-identifying string that the remote names itself.
+        self.client_version_string = None
+
         # Networking reader and writer objects for communication
         self.reader = connection.reader
         self.writer = connection.writer
@@ -480,6 +483,13 @@ class BasePeer(BaseService):
         if not isinstance(cmd, Hello):
             await self.disconnect(DisconnectReason.bad_protocol)
             raise HandshakeFailure(f"Expected a Hello msg, got {cmd}, disconnecting")
+
+        # limit number of chars to be displayed, and try to keep printable ones only
+        # MAGIC 256: arbitrary, "should be enough for everybody"
+        client_version_string = msg['client_version_string'][:256].strip('\t\n\r\x0b\x0c')
+        if not client_version_string.isprintable():
+            client_version_string = client_version_string.encode("utf-8", "replace").decode()
+        self.client_version_string = client_version_string
 
         # Check whether to support Snappy Compression or not
         # based on other peer's p2p protocol version
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -346,6 +346,7 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
                     "%s: uptime=%s, received_msgs=%d, most_received=%s(%d)",
                     peer, peer.uptime, peer.received_msgs_count,
                     most_received_type, count)
+                self.logger.debug("client_version_string='%s'", peer.client_version_string)
                 for line in peer.get_extra_stats():
                     self.logger.debug("    %s", line)
             self.logger.debug("== End peer details == ")


veox:
> I prefer `replace` over `ignore` just so we can tell that something wacky happened.

OK, one sec.

> Not sure why we would limit the length at all here. Maybe just in the display?

It is my intuition that sanitising should happen as soon as possible.

veox:
Tests fail, because `paragon` hands over the data as a string already (not bytes, as they would be received over network).

p2p/peer.py
```diff
@@ -481,6 +484,11 @@ def process_msg(self, cmd: protocol.Command, msg: protocol.PayloadType) -> None:
             await self.disconnect(DisconnectReason.bad_protocol)
             raise HandshakeFailure(f"Expected a Hello msg, got {cmd}, disconnecting")
 
+        # limit number of chars that will be displayed in version strings,
+        # and replace invalid characters with question marks
+        # MAGIC 96: even overly-verbose parity debug build strings are shorter than this
+        self.client_version_string = msg['client_version_string'][:96].decode("utf8", "replace")
```
veox: To think about it, should probably also strip non-printable characters, too. Including line feeds and carriage returns.

(FTR, I'm so stringent on sanitising this field, since it's a low-effort way for a malicious remote to push bogus data: they don't even need to have an actual node, just something that knows how to handshake over `devp2p`.)

carver: I'm totally on-board with sanitizing non-printable and control characters.

I think we can be fairly loose with the length of the client version string, given how few peers we connect to. I suppose some limit makes sense, but maybe even 256 characters is okay.

veox:
Fixed CI failure (was `str`/`bytes` misunderstanding on my part); PR seems ready.

p2p/peer.py
```diff
@@ -481,6 +484,13 @@ def process_msg(self, cmd: protocol.Command, msg: protocol.PayloadType) -> None:
             await self.disconnect(DisconnectReason.bad_protocol)
             raise HandshakeFailure(f"Expected a Hello msg, got {cmd}, disconnecting")
 
+        # limit number of chars to be displayed, and try to keep printable ones only
+        # MAGIC 256: arbitrary, "should be enough for everybody"
+        client_version_string = msg['client_version_string'][:256].strip('\t\n\r\x0b\x0c')
+        if not client_version_string.isprintable():
+            client_version_string = client_version_string.encode("utf-8", "replace").decode()
```
carver: Now that we know that we're encoding from a (unicode) string to utf-8 bytes, I don't think there's any way we *can* have errors to replace. I'm also not sure that this round-trip helps us with removing all non-printables.

What about something like this:
```py
        original_version = msg['client_version_string']
        client_version_string = original_version[:256] + ('...' if original_version[256:] else '')
        if client_version_string.isprintable():
            self.client_version_string = client_version_string
        else:
            self.client_version_string = repr(client_version_string)
```
According to the `isprintable()` docs (nice use of stdlib, btw):
> Note that printable characters in this context are those which should not be escaped when repr() is invoked on a string.

`repr()` lets us see the full original string, and it might be fun to see what shenanigans people try. Also, `repr()` apparently doesn't miss converting any characters that `isprintable()` has caught.

veox: Ah, yes, that passage did catch my eye! But it slipped.

I'll just copy-paste what you did there.

veox:
Fixed linting (hopefully).

JIC, this PR is so small - it's definitely OK to squash-merge.

p2p/peer.py
```diff
@@ -481,6 +484,15 @@ def process_msg(self, cmd: protocol.Command, msg: protocol.PayloadType) -> None:
             await self.disconnect(DisconnectReason.bad_protocol)
             raise HandshakeFailure(f"Expected a Hello msg, got {cmd}, disconnecting")
 
+        # limit number of chars to be displayed, and try to keep printable ones only
+        # MAGIC 256: arbitrary, "should be enough for everybody"
+        original_version = msg['client_version_string']
+        client_version_string = original_version[:256] + ('...' if original_version[256:] else '')
+        if client_version_string.isprintable():
+            self.client_version_string = client_version_string.strip('\t\n\r\x0b\x0c')
```
carver: Strip is a noop here, though, because:

```py
In [3]: for char in '\t\n\r\x0b\x0c':
   ...:     print(repr(char), char.isprintable())    
   ...:     
'\t' False
'\n' False
'\r' False
'\x0b' False
'\x0c' False
```

If you like, we could do `.strip()` to get rid of useless whitespace on the edges. Think I'd rather not leave the noop strip in, though.

```suggestion
            self.client_version_string = client_version_string.strip()
```

veox: Argh! My idea with this was to prevent

```
   DEBUG  03-06 14:07:26           ETHPeerPool  ETHPeer <Node(0x04f8@88.198.47.121)>: uptime=0:00:41:29, received_msgs=1076, most_received=Transactions (cmd_id=18)(576)                                                                      
   DEBUG  03-06 14:07:26           ETHPeerPool  client_version_string='Geth/
╔════════════════════╗
║    BUY SCAMCOIN    ║
╚════════════════════╝
v1.8.11-stable-dea1ce05/linux-amd64/go1.10'      
   DEBUG  03-06 14:07:26           ETHPeerPool      BlockBodies: None                                                                                                                                                                         
   DEBUG  03-06 14:07:26           ETHPeerPool      BlockHeaders: msgs=1  items=2  rtt=19.01/0.19/0.00  ips=0.51390  timeouts=0  quality=5                                                                                                    
   DEBUG  03-06 14:07:26           ETHPeerPool      NodeData: None                                                                                                                                                                            
   DEBUG  03-06 14:07:26           ETHPeerPool      Receipts: msgs=283  items=1523595  rtt=1.14/1.93/0.21  ips=5241.20800  timeouts=32  quality=10
```

but I failed to see that the characters are non-printable (and would therefore force processing the string with `repr()`).

One sec...

PR #335, opened 2019-02-28
Remove broken code in test

cburgdorf:
### What was wrong?

The monkeypatch doesn't work here.  We can clearly see that when we run:

```
pytest tests/core/p2p-proto/test_sync.py -k 'test_fast_syncer' --log-cli-level 0
```

relevant logs:

```
Successfully decoded GetBlockHeaders (cmd_id=19) msg: {'block_number_or_hash': 21, 'max_headers': 192, 'skip': 0, 'reverse': False}
```

### How was it fixed?

Removed it for now but I actually wish it would work because what it tries to do is actually useful. One can set something like `client_peer.max_headers_fetch` to get it working for the most part but the `SkeletonSync` does also import and use `MAX_HEADERS_FETCH` so this solution doesn't work everywhere.

I tried to figure out why the monkeypatch doesn't work here and I guess it has to do with the module being loaded earlier from somewhere but I couldn't figure something out that works.

@lithp This is the kind of stuff that makes me cautious regarding globals. Do you have an idea how to to properly change these constants under test so that the changes take effect everywhere where they are importet?

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/eth/peer.py
+++ b/trinity/protocol/eth/peer.py
@@ -28,13 +28,12 @@ from .commands import (
     NewBlock,
     Status,
 )
-from .constants import MAX_HEADERS_FETCH
+from . import constants
 from .proto import ETHProtocol
 from .handlers import ETHExchangeHandler
 
 
 class ETHPeer(BaseChainPeer):
-    max_headers_fetch = MAX_HEADERS_FETCH
 
     supported_sub_protocols = [ETHProtocol]
     sub_proto: ETHProtocol = None
@@ -44,6 +43,10 @@ class ETHPeer(BaseChainPeer):
     def get_extra_stats(self) -> List[str]:
         stats_pairs = self.requests.get_stats().items()
         return ['%s: %s' % (cmd_name, stats) for cmd_name, stats in stats_pairs]
+
+    @property
+    def max_headers_fetch(self) -> int:
+        return constants.MAX_HEADERS_FETCH
 
     @property
     def requests(self) -> ETHExchangeHandler:
--- a/trinity/protocol/les/peer.py
+++ b/trinity/protocol/les/peer.py
@@ -30,13 +30,13 @@ from trinity.protocol.common.peer import (
     BaseChainPeerPool,
 )
 
+from . import (
+    constants,
+)
 from .commands import (
     Announce,
     Status,
     StatusV2,
-)
-from .constants import (
-    MAX_HEADERS_FETCH,
 )
 from .proto import (
     LESProtocol,
@@ -46,7 +46,6 @@ from .handlers import LESExchangeHandler
 
 
 class LESPeer(BaseChainPeer):
-    max_headers_fetch = MAX_HEADERS_FETCH
 
     supported_sub_protocols = [LESProtocol, LESProtocolV2]
     sub_proto: LESProtocol = None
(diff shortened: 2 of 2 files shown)


carver:
Yeah, you could mock the constant where it is imported inside the syncer class, but that's super implementation-specific and brittle.

At some point, we've talked about tracking stats on specific peers if they return less items than we requested. Then we can request fewer items next time. So if we do that, then maybe we could actually configure the peer side to say that we should only request 10 headers at a time from that particular peer. (without monkeypatching)

pipermerriam:
Even if we get peer classes adjusted so we can configure these, the limits will still have to exist somewhere.  I'd suggest we just use a simple class to store these.

```python
class EthProtocolConfig:
    MAX_HEADER_BATCH = 192
    ...
```

That gives us the option of:

1. just monkeypatching the object itself.
2. using instances of the class in places we want localized versions that aren't globals

cburgdorf:
I looked into this again today and noticed creating a class `EthProtocolConfig` as suggested by @pipermerriam doesn't solve the actual issue with the monkeypatch that our current sync test uses. The problem remains that `max_headers_fetch` is set *before* the monkeypatch had a chance to do the mutation. Plus, even if the timing were different so that the monkeypatching would go in effect, I believe with the current implementation there is no way for the monkeypatcher to revert back to the old value.

We can make this actual work if we turn `max_headers_fetch` into an actual property that directly reads from the `constants` module.

```
@property
    def max_headers_fetch(self) -> int:
        return constants.MAX_HEADERS_FETCH
```

I pushed new code into this PR that does that and now it uses the the correct value during the test.

>DEBUG  04-29 12:00:13            servers.py  ETHPeer <Node(0xb590@0.0.0.0)> requested headers: {'block_number_or_hash': 399, 'max_headers': 10, 'skip': 0, 'reverse': False}

In fact, now that it uses the value that the test wanted it to use, it turns out that it broke the skeleton syncer tests...looking into that.

cburgdorf:
It appears that the skeleton sync, if run with `MAX_HEADERS_FETCH` that is not `192`, will run into the following error...

>   DEBUG  04-29 12:29:46            headers.py  Exiting sync and booting ETHPeer <Node(0x7d21@0.0.0.0)> due to validation error: Invalid header chain; <BlockHeader #772 2a09d3ae> has parent b'\x9e\x7f\xa4\xe3\xff\xe2u\xc5n\xa0H\xc8\xaa}N08i\xf9\xe8%HV\x9c\\\xc76\x89T)\x99X', but expected b'\xd7M\xc5\x7f\x91\xf2\xc3\x95\xfe\x89*\x9a\x0f\xcc+\xd7[\xec\x0e\x90\x98&v\xf6\x03\x89/$\xb8\xf8\xe6$'

cburgdorf:
@carver That's the thing I was talking about. Without the change in this PR, the following code in our tests does not have the intended effect.

https://github.com/ethereum/trinity/blob/7412b0270c7bfc0f9adb9e881aa937bf1d384aed/tests/core/p2p-proto/test_sync.py#L33-L40

One can observe in the logs that requests are still made with `192` max headers. With this PR, we **do** request the intentional smaller batches. However, now something in the `SkeletonSyncer` breaks and I *think* it is due to the fact that the `SkeletonSyncer` does probably still internally calculates with `192` which lead me down a road of adjusting imports to use `constants.MAX_HEADERS_FETCH` everywhere but that seems like a big footgun to forget. And it didn't even fix the error...

carver:
Hm, yeah, you're right, it's probably stitching the headers together at the wrong splice point when the expected max header size is different. If you link me to this:

> adjusting imports to use `constants.MAX_HEADERS_FETCH` everywhere

I can take a look.

cburgdorf:
>If you link me to this:

I threw this away because it felt really fragile to do that in the first place (easy to forget you **have to** import it that way).

carver:
> > If you link me to this:
> 
> I threw this away because it felt really fragile to do that in the first place (easy to forget you **have to** import it that way).

Oh I misunderstood before. It's the monkeypatch support that's the problem, not that sync had the number hard-coded somehow. :+1:

PR #337, opened 2019-02-28
Upgrade to py-evm v0.2.0-alpha.42

carver:
### What was wrong?

Need the py-evm which includes the latest Constantinople version info.

### How was it fixed?

- Upgrade to alpha 42
- update genesis files for mainnet & ropsten
- update chain proxy class for new method `get_receipt_by_index`

#### Cute Animal Picture


carver:
Cool, I want to rebase on top of #236 and make sure all tests pass, then it's release time!

PR #341, opened 2019-03-01
Add sparse merkle utils

### LINKED ISSUE #289: Implement sparse merkle tree utils
### What is wrong?

current merkle tree utils in `eth2/_utils/merkle` requires all leaves to compute tree root. It will not be practical to use it to compute deposit tree in Deposit contract(which is a sparse merkle tree).

### How can it be fixed
1. Rename the old `eth2/_utils/merkle.py` to `eth2/_utils/merkle_normal.py`
2. Add `eth2/_utils/merkle_sparse.py`

NIC619:
### What was wrong?
Fix #289 

### How was it fixed?
- [x] Rename `/_utils/merkle.py` to `/_utils/merkle_normal.py`
- [x] Add `/_utils/merkle_sparse.py`
    - [x] Add fixed tree height `32`
    - [x] Add empty node hashes
    - [x] Append empty node hash of that layer if number of nodes in the layer is an odd number
    - The rest is mostly the same as `merkle_normal`
    - currently parameters, e.g., tree roots, of the tests are from the tree built by deposit contract


eth2/_utils/merkle_sparse.py
```diff
@@ -0,0 +1,93 @@
...
+
+from .merkle_normal import (  # noqa: F401
+    _calc_parent_hash,
+    _hash_layer,
+    get_branch_indices,
+    get_root,
+    MerkleTree,
+    MerkleProof,
+    verify_merkle_proof,
+)
```
hwwhww: Due to the symmetry of  `merkle_normal` and `merkle_sparse`, it's natural to me to say that we could make an abstract interface... but also I don't want to increase the complexity and I do like the current pure functions...

Ping @pipermerriam and co to comment.

eth2/_utils/merkle_sparse.py
```diff
@@ -0,0 +1,93 @@
...
+    Calculate the Merkle root corresponding to a list of items.
+    """
+    return get_root(calc_merkle_tree(items))
+
+
+def calc_merkle_tree_from_leaves(leaves: Sequence[Hash32]) -> MerkleTree:
+    if len(leaves) == 0:
+        raise ValueError("No leaves given")
+    tree = tuple()  # type: ignore
+    tree = (leaves,) + tree
```
ralexstokes: seems like L80-81 can be simplified?

```python
tree = (leaves,)
```
right?

eth2/_utils/merkle_sparse.py
ralexstokes: @NIC619 @hwwhww 
i would suggest we remove all of the duplicate functions in this new file and only keep the ones that are different.

it seems like the only differences are the index validation on `get_merkle_proof` and the actual construction `calc_merkle_tree_from_leaves`. i would ask if there is a way to unify the index validation check b/t the two implementations bc i would guess we can reduce the variance.

and following up on what @hwwhww is saying we should then think about how to provide a simple interface for users of this logic. i would suggest we keep it simple and just provide high-level functions for what we need.

example usage:

```python
# validator_client.py or something
## i want to join the validator set!

from eth2._utils.merkle_sparse import calc_merkle_tree, get_merkle_proof
deposits = fetch_deposits_from_eth1_as_bytes()
my_deposit = deposits[-1]
assert is_my_deposit(my_deposit)
my_deposit_index = len(deposits)-1
tree = calc_merkle_tree(deposits)
proof = get_merkle_proof(tree, my_deposit_index)
deposit_operation = make_deposit_operation(my_deposit, proof) # and whatever else...
broadcast(deposit_operation)
```

it will be on users to decide if they want a 'normal' tree or a 'sparse' tree. we could make an argument for some kind of generic interface -- "common" logic in a `_utils/merkle_tree.py` with the specialized logic in a `_utils/merkle_normal.py` and `_utils/merkle_sparse.py`. as a caller, you are still responsible for picking the right thing and then the actual callsite gets *slightly* clumsier as you have to pass in the appropriate 'builder' function (aka which impl of `calc_merkle_tree_from_leaves` you want).

all things considered, while i do recognize the symmetry here i don't really see enough value in creating an abstraction against these two implementations at the moment. this could change as our actual usages of these things evolve. i would see more reason to do so if the things we retain in the abstraction are the things we are really after (i.e. make it so a caller never has to think about which impl they want) or if we end up w/ a bunch of implementations of this 'proof tree' interface. and i don't see either of those being the case anytime soon :)

NIC619:
@ralexstokes Thanks for the review! Sorry I did not push the updates on time. As per @hwwhww 's suggestion, I move the common utils into `_utils/common.py` and `merkle_normal`/`merkle_sparse` would import from there.(IMO `_utils/merkle_tree.py` strikes me more as a real implementation of a merkle tree instead of just a place for common utils.)

For the  index validation check, I remove the additional check in `get_merkle_proof` from `merkle_sparse` so now both `merkle_normal` and `merkle_sparse` will import `get_merkle_proof` from `common.py`.
(The reason for the additional check is that in `merkle_sparse` tree, an `EmptyNodeHash` is appended to each layer of nodes if the number of nodes of that layer is an odd number, the additional check was there to prevent user from getting a merkle proof of an (appended) empty nodes. But now I feels like it's more of a user side error hence remove the additional check.)

eth2/_utils/merkle/sparse.py
```diff
@@ -80,7 +76,7 @@ def calc_merkle_tree_from_leaves(leaves: Sequence[Hash32]) -> MerkleTree:
     if len(leaves) == 0:
         raise ValueError("No leaves given")
     tree = tuple()  # type: ignore
-    tree = (leaves,) + tree
+    tree = (leaves,)
```
ralexstokes: and then we can get rid of the L78 right?

 no need to declare `tree = tuple()` because we instantiate it on the next line L79.

ralexstokes:
the current status of this branch still duplicates some stuff we could pull out but it would turn some of these functions into "factories" which could be viewed as raising the cost to understand them -- and to this end we can prolly leave them out and go ahead and merge this in w/o worrying about the abstractions for now...

but just to illustrate, something like this in `normal.py`:

```python
from common import calc_merkle_tree_maker

calc_merkle_tree = calc_merkel_tree_maker(calc_merkle_tree_from_leaves)
```

and then we can condense the logic to build a tree into the `common` module. but like i said we may be at the point of diminishing returns :)

NIC619:
@ralexstokes The compromise between code duplication and readability here seems fair to me! But after re-read your comment again, I probably have misunderstood the factory you mentioned. The `calc_merkle_tree_maker` I made is more like a decorator that does the preprocessing of converting `items` to `leaves`. Please let me know if I'm doing it right!

eth2/_utils/merkle/sparse.py
pipermerriam: @NIC619 as a general rule, overwriting an already defined variable should be a *code smell*.  We avoid doing it other than in cases where the alternative significantly reduces readability, there is a justifiable performance issue, or some other explicit justification.

ralexstokes:
@NIC619 yeah i wouldn't get too caught up in it -- i may have not been very clear but i'm not sure my suggestion adds enough value at this point to be warranted. i would address the stuff to clean up in the other comments above and then we can go ahead and merge this in :)

NIC619:
I do prefer this factory approach tho 😃  but like you said I'm not sure how much impact this has on readability either. Reverted and is ready to merge.

Issue #343, opened 2019-03-01
Sync with the spec - data structures


hwwhww:
### What is wrong?
1. [#349] `Attestation`: fix fields ordering https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#attestation
2. [#355] Spec PR https://github.com/ethereum/eth2.0-specs/pull/633
    1. Rename `ProposalSignedData` to `Proposal`
    2. Move `ProposerSlashing.proposal_signature_1` and `ProposerSlashing.proposal_signature_2` into per `Proposal`'s `signature`
3. [#353] Add `Transfer` data structure (cherry-pick from #342)
4. [#351] Update `AttesationData`: `latest_crosslink_root: Hash32` to `latest_crosslink: CrosslinkRecord`

### How can it be fixed

Fill this in if you know how to fix it.

PR #349, opened 2019-03-02
Fix `Attestation` ordering: `aggregation_bitfield` before `data`

hwwhww:
### What was wrong?
Sync with the spec
#343 task 1: `Attestation`: fix fields ordering https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#attestation

### How was it fixed?
Simply reorder it.

#### Cute Animal Picture

hedgehog-child-3636026_640

### DIFF
--- a/eth2/beacon/types/attestations.py
+++ b/eth2/beacon/types/attestations.py
@@ -18,9 +18,10 @@ from eth2.beacon.constants import EMPTY_SIGNATURE
 class Attestation(ssz.Serializable):
 
     fields = [
-        ('data', AttestationData),
         # Attester aggregation bitfield
         ('aggregation_bitfield', bytes_sedes),
+        # Attester aggregation bitfield
+        ('data', AttestationData),
         # Proof of custody bitfield
         ('custody_bitfield', bytes_sedes),
         # BLS aggregate signature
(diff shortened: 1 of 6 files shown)


eth2/beacon/types/attestations.py
```diff
@@ -18,23 +18,24 @@
 class Attestation(ssz.Serializable):
 
     fields = [
-        ('data', AttestationData),
         # Attester aggregation bitfield
         ('aggregation_bitfield', bytes_sedes),
+        # Attester aggregation bitfield
```
ChihChengLiang: Wrong comment

tests/core/p2p-proto/bcc/test_commands.py
```diff
@@ -171,7 +172,6 @@
             justified_epoch=SERENITY_CONFIG.GENESIS_EPOCH,
             justified_block_root=ZERO_HASH32,
         ),
-        aggregation_bitfield=b"\x00\x00\x00",
         custody_bitfield=b"\x00\x00\x00",
     )
```
ChihChengLiang: ⏬ The `Attestation` in the `test_send_multiple_attestations` should be reordered too.

PR #352, opened 2019-03-02
Fix `validate_attestation_slot`

### LINKED ISSUE #345: Fix `validate_attestation_slot`
### What is wrong?
https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#attestations-1
1. Add assertion for `Verify that attestation.data.slot >= GENESIS_SLOT`
```python
    # Can't submit attestations that are too far in history (or in prehistory) 
    assert attestation.data.slot >= GENESIS_SLOT
    assert state.slot < attestation.data.slot + SLOTS_PER_EPOCH
    # Can't submit attestations too quickly
    assert attestation.data.slot + MIN_ATTESTATION_INCLUSION_DELAY <= state.slot
```

2. Fix other assertions around `MIN_ATTESTATION_INCLUSION_DELAY` (https://github.com/ethereum/eth2.0-specs/pull/669)

### How can it be fixed

hwwhww:
### What was wrong?

Fix #345 

### How was it fixed?

Sync with the [latest spec logic](https://github.com/ethereum/eth2.0-specs/blob/fa327a18ea7e2d1551e0df5b982a1e2b66ee8462/specs/core/0_beacon-chain.md#attestations-1):
```python
    # Can't submit attestations that are too far in history (or in prehistory) 
    assert attestation.data.slot >= GENESIS_SLOT
    assert state.slot < attestation.data.slot + SLOTS_PER_EPOCH
    # Can't submit attestations too quickly
    assert attestation.data.slot + MIN_ATTESTATION_INCLUSION_DELAY <= state.slot
```

#### Cute Animal Picture
hedgehog-248827_640

### DIFF
--- a/eth2/beacon/state_machines/forks/serenity/block_validation.py
+++ b/eth2/beacon/state_machines/forks/serenity/block_validation.py
@@ -310,7 +310,7 @@ def validate_attestation(state: BeaconState,
         state.slot,
         slots_per_epoch,
         min_attestation_inclusion_delay,
-        get_epoch_start_slot(committee_config.GENESIS_EPOCH, committee_config.SLOTS_PER_EPOCH),
+        committee_config.GENESIS_SLOT,
     )
 
     validate_attestation_justified_epoch(
--- a/eth2/beacon/state_machines/forks/serenity/block_validation.py
+++ b/eth2/beacon/state_machines/forks/serenity/block_validation.py
@@ -310,6 +310,7 @@ def validate_attestation(state: BeaconState,
         state.slot,
         slots_per_epoch,
         min_attestation_inclusion_delay,
+        get_epoch_start_slot(committee_config.GENESIS_EPOCH, committee_config.SLOTS_PER_EPOCH),
     )
 
     validate_attestation_justified_epoch(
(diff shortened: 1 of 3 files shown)


eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -307,6 +307,7 @@ def validate_attestation(state: BeaconState,
         state.slot,
         slots_per_epoch,
         min_attestation_inclusion_delay,
+        get_epoch_start_slot(committee_config.GENESIS_EPOCH, committee_config.SLOTS_PER_EPOCH),
```
djrtwo: This should be `genesis_slot` rather than the current epoch start slot.

ralexstokes: it looks like it is? in lieu of passing in the `GENESIS_SLOT` directly, we just calculate it here

hwwhww: Ah, I think I was trying to avoid adding one more constant in `CommitteeConfig`. But I will change to `GENESIS_SLOT` to align with the spec.

Issue #238, opened 2019-01-30
Implement the rest of block operations


hwwhww:
### What is wrong?
`eth2/beacon/state_machines/forks/serenity/operation_processing.py`

- [x] [`process_proposer_slashings(state, block, self.config)`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#proposer-slashings-1)
- [x] [`process_attester_slashings(state, block, self.config)`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#attester-slashings-1)
    - Note: rename `CasperSlashing` to `AttesterSlashing`
- [ ] [`process_deposits(state, block, self.config)`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#deposits-1)
- [x] [`process_voluntary_exits(state, block, self.config)`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#exits-1)

### How can it be fixed

Fill this in if you know how to fix it.

hwwhww:
`process_deposits(state, block, self.config)` is pending on #289 and spec for generating the test vectors of deposit tree.

PR #358, opened 2019-03-03
Implement `process_voluntary_exits`

hwwhww:
### What was wrong?
#238 - `process_voluntary_exits`

### How was it fixed?
Ref spec: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#voluntary-exits-1

#### Cute Animal Picture

african-lion-951778_640

### DIFF
--- a/eth2/beacon/state_machines/forks/serenity/block_validation.py
+++ b/eth2/beacon/state_machines/forks/serenity/block_validation.py
@@ -54,7 +54,7 @@ from eth2.beacon.types.proposal_signed_data import ProposalSignedData
 from eth2.beacon.types.slashable_attestations import SlashableAttestation  # noqa: F401
 from eth2.beacon.types.proposer_slashings import ProposerSlashing
 from eth2.beacon.types.states import BeaconState  # noqa: F401
-from eth2.beacon.types.voluntary_exit import VoluntaryExit  # noqa: F401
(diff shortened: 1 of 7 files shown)


hwwhww:
19 lines in the spec -> 419 lines here(!)

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -703,3 +705,74 @@ def validate_slashable_attestation(state: 'BeaconState',
...
         )
+
+
+#
+# Voluntary Exit
+#
+def validate_voluntary_exit(state: 'BeaconState',
+                            voluntary_exit: 'VoluntaryExit',
+                            slots_per_epoch: int,
+                            activation_exit_delay: int) -> None:
```
ChihChengLiang: Not in this PR: Realized that `slots_per_epoch` should be `Slot` and `activation_exit_delay` should be `Epoch`. Created an issue for these: #360 .

hwwhww: Ah, not really. `SlotNumber` was renamed to `SlotNumber`, `EpochNumber` was renamed to `Epoch`.
We discussed something similar to this case. The `5`th slot != `5` slots. `slots_per_epoch` is set to `64` slots, and `activation_exit_delay` is set to `2**2` epochs.

PR #364, opened 2019-03-04
A more helpful db-shell

pipermerriam:
### What was wrong?

The `db-shell` is quite nice, but it only exposes the `ChainDB`.  Given this is a low level utility it would be nice to have access to more of the *stuff*

### How was it fixed?

Expanded the variables available in the local context to include

- `db` the base db instance
- `trinity_config`: the `TrinityConfig` instances
- `chain_config`: the `Eth1AppConfig` instance
- `chain`: the `Chain` instance.

#### Cute Animal Picture

foxwdinner

### DIFF
--- a/trinity/plugins/builtin/attach/console.py
+++ b/trinity/plugins/builtin/attach/console.py
@@ -15,6 +15,9 @@ from eth.db.backends.level import LevelDB
 from trinity._utils.log_messages import (
     create_missing_ipc_error_message,
 )
+from trinity.config import (
+    TrinityConfig,
+)
 
 
 DEFAULT_BANNER: str = (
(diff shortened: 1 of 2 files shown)


trinity/plugins/builtin/attach/console.py
```diff
@@ -91,18 +94,28 @@ def rpc(method: str, params: Dict[str, Any] = None) -> str:
...
 
-    shell(use_ipython, {'chaindb': chaindb}, DB_SHELL_BANNER + greeter)
+    namespace = {
+        'db': db,
+        'chaindb': chaindb,
+        'trinity_config': trinity_config,
+        'chain_config': chain_config,
+        'chain': chain,
+    }
+    shell(use_ipython, namespace, DB_SHELL_BANNER + greeter)
```
cburgdorf: Maybe also update `DB_SHELL_BANNER` to list all the new exposed goodies.

trinity/plugins/builtin/attach/console.py
```diff
@@ -91,18 +94,35 @@ def rpc(method: str, params: Dict[str, Any] = None) -> str:
...
     greeter = f"""
     Head: #{head.block_number}
     Hash: {head.hex_hash}
     State Root: {encode_hex(head.state_root)}
+
+    Available Context Variables:
+      - `db`: base database object
+      - `chaindb`: `ChainDB` instance
+      - `trinity_config`: `TrinityConfig` instance
+      - `chain_config`: `Eth1AppConfig` instance
```
cburgdorf: `Eth1AppConfig` should be `ChainConfig` (in fact, having the `ChainConfig` being tightly coupled to the `TrinityConfig` and not to the `Eth1AppConfig` is still one of the things that we need to over come.

PR #367, opened 2019-03-05
Enable process randao

### LINKED ISSUE #284: Enable `process_randao`
### What is wrong?
`process_randao` is implemented in #252, but it has not been enabled in `SerenityStateTransition`.

### How can it be fixed
- Uncomment https://github.com/ethereum/trinity/blob/b6667b6a6ed5748c1b0037f6b4135606cc0987d8/eth2/beacon/state_machines/forks/serenity/state_transitions.py#L101

- Add helpers in `eth2/beacon/tools/builder/validator.py` to set randao reveals for `test_demo.py`.

ralexstokes:
### What was wrong?

Fixes #284. 

### How was it fixed?

* Added the beginnings of a RANDAO gadget (will want to use in a validator client, someday)
* Enabled the RANDO processing
* Updated the tests

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/state_machines/forks/serenity/block_processing.py
+++ b/eth2/beacon/state_machines/forks/serenity/block_processing.py
@@ -69,6 +69,7 @@ def process_randao(state: BeaconState,
 
     validate_randao_reveal(
         randao_reveal=block.randao_reveal,
+        proposer_index=proposer_index,
         proposer_pubkey=proposer.pubkey,
         epoch=epoch,
         fork=state.fork,
--- a/eth2/beacon/state_machines/forks/serenity/block_validation.py
+++ b/eth2/beacon/state_machines/forks/serenity/block_validation.py
@@ -602,6 +602,7 @@ def validate_attestation_aggregate_signature(state: BeaconState,
 
 
 def validate_randao_reveal(randao_reveal: BLSSignature,
+                           proposer_index: ValidatorIndex,
                            proposer_pubkey: BLSPubkey,
                            epoch: Epoch,
                            fork: Fork) -> None:
@@ -618,8 +619,9 @@ def validate_randao_reveal(randao_reveal: BLSSignature,
     if not is_randao_reveal_valid:
         raise ValidationError(
             f"RANDAO reveal is invalid. "
-            f"reveal={randao_reveal}, proposer_pubkey={proposer_pubkey}, "
-            f"message_hash={message_hash}, domain={domain}"
+            f"proposer_index={proposer_index}, proposer_pubkey={proposer_pubkey}, "
+            f"reveal={randao_reveal}, "
+            f"message_hash={message_hash}, domain={domain}, epoch={epoch}"
         )
 
 
--- a/eth2/beacon/state_machines/forks/serenity/epoch_processing.py
+++ b/eth2/beacon/state_machines/forks/serenity/epoch_processing.py
@@ -1308,7 +1308,7 @@ def process_final_updates(state: BeaconState,
         ),
         latest_randao_mixes=update_tuple_item(
             state.latest_randao_mixes,
-            next_epoch % config.LATEST_SLASHED_EXIT_LENGTH,
+            next_epoch % config.LATEST_RANDAO_MIXES_LENGTH,
             get_randao_mix(
                 state=state,
                 epoch=current_epoch,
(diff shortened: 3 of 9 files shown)


ralexstokes:
i'm trying out the "draft" feature on this one...

it is WIP as there is some strange corner case i've been trying to track down around RANDAO generation some epochs into the future (failure is on the last block of `test_demo.py`. will un-WIP once I sort out what is going on.

ralexstokes:
i also wanted to open the floor for some discussion -- we generally want to be able to generate a correct RANDAO reveal some time into the future (based on what we currently know about the network). there is problem w/ the current API in that it requires a block to execute the state transition. what we also want is a way to say "fast forward the state to this slot", which i will then use to query things about the future state (like if I'm a block proposer or not). there are a number of ways to imagin implementing this and quickly touches on issues of visibility in Python (should this be public or private, etc.).

i have implemented one method by just tacking on the block-less state transition but would love input on this approach. the other obvious thing is to allow `apply_state_transition` to take a `None` block -- this keeps the code DRYer and so may be preferrable. my thinking w/ the block-less approach is that while there is some duplication here, i would be concerned about seeding the transformation of `apply_state_transition` into a kitchen sink method (i.e. it is doing too much at once!)

thoughts? @hwwhww @pipermerriam @djrtwo i would think have an opinion on this -- happy to hear anyone else's take :)

hwwhww:
@ralexstokes 

1. For the lookahead problem: I opened a PR in your repo: https://github.com/ralexstokes/trinity/pull/1 with `get_committee_assignment` ([in v-guide spec](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/validator/0_beacon-chain-validator.md#validator-assignments)).
2. The block-less state transition: agreed that even we don't need `apply_state_transition_without_block` for lookahead, we still need it for the skipped slot cases. I prefer to resue `apply_state_transition` too. 👍

djrtwo:
I think you may be running into a bug because we are still processing epoch transition at the end of hte last slot in an epoch rather than the beginning of the next slot in the epoch. After merging https://github.com/ethereum/eth2.0-specs/pull/711, I'll add a new PR to swap the epoch transition location.

The one challenge of having `apply_state_transition` accept `None` for block is that you'd also have to pass in a separate `target_slot` which makes it a little bit more cumbersome to overload.

ralexstokes:
ok i'll be on the lookout for the change to *when* to process an epoch change...

i would also suggest leaving it as is on this PR at the moment and **not** overloading `apply_state_transition` -- but if i'm reading you correctly @hwwhww you feel differently?

the signature would have to change from this:
```python
def apply_state_transition(self, state: BeaconState, block: BaseBeaconBlock, **options) -> BeaconState: pass
```
to something like:

```python
def apply_state_transition(self, state: BeaconState, block: Option[BaseBeaconBlock], target_slot: Slot, **options) -> BeaconState: pass
```

which imo breaks the purity of "(state, block) -> new-state". this being said it does move the burden to the caller to figure out which function they want to call. so i could be talked into upgrading `apply_state_transition` to essentially take `Option[Block], Option[Slot]`... i would think we would throw an exception if slot was present and block.slot was also present

ralexstokes:
@hwwhww i like the use of `get_committee_assignment` where we can use it however a number of the tests are failing where we are looking ahead more than one epoch (e.g. `git checkout 
e9a97f3 && pytest tests/eth2/beacon/state_machines/test_state_transition.py`) so i would elect to revert back to the `apply_state_transition_without_block` as it is the more general method here.

otherwise we have to be to disable the validation that the requested epoch is within the bounds of the state's current epoch. and that seems riskier than having this method `apply_state_transition_without_block` sitting on the side, possibly w/ warnings to only use in testing situations... thoughts?

hwwhww:
@ralexstokes 
Ah, right I see the test failed in `test_per_slot_transition`... where I am not able to do lookahead for a large number of skipped slots with `get_committee_assignment`. Agreed that `apply_state_transition_without_block` would be better for mocking test case here. 👍 


Yeah it seems `def apply_state_transition(self, state: BeaconState, block: Option[BaseBeaconBlock], target_slot: Slot, **options) -> BeaconState: pass` is less generic than I thought...

So except for testing, we may have several cases:
1. Apply state transition with a block:
    - if there's any skipped slot, it will handle that too. 
    -  Timing: chain syncing, upon receiving block and the parent block is valid.
2. Apply state transition to next slot, without a block:
    - Timing: time's up (i.e., maybe half of `SECONDS_PER_SLOT`, depending on the validator's setting) for the attester to vote on the slot.

I think they are different cases and easy to distinguish them. So I'm okay with adding `apply_state_transition_without_block` too.

ralexstokes:
@hwwhww yeah those cases makes sense

i'll go ahead and revert the use of `get_committee_assignment` as that should allow the tests to pass and then this will be ready for merge. we will keep our eyes open for uses of `apply_state_transition_without_block` that are liable to be mis-used.

ralexstokes:
meta: tried using the github ui to resolve the merge conflict. didn't like it.

ralexstokes:
this should be ready for merge!

PR #371, opened 2019-03-06
Update genesis slot

### LINKED ISSUE #259: Set `GENESIS_SLOT` to `2**32`
### What is wrong?

Spec PRs: ethereum/eth2.0-specs#505 (master branch, January Release) and ethereum/eth2.0-specs#537 (dev branch)
- Set `GENESIS_SLOT` to `2**32`
### How can it be fixed

ralexstokes:
### What was wrong?

Fixes #259.

### How was it fixed?

Updated the default genesis slot to the specified value, `2**32`. Moreover, we now compute the genesis epoch from the genesis slot value.

Many tests were hard-coded to a genesis slot of `0` -- they should all pass but I put all of those changes into its own commit in case you just want to see the configuration changes.

I also marked some tests as `slow` or `long`. You can {run, skip} these tests w/ `pytest -{m,k}-{slow,long}` which may be helpful to you.

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/state_machines/forks/serenity/configs.py
+++ b/eth2/beacon/state_machines/forks/serenity/configs.py
@@ -14,6 +14,9 @@ from eth2.beacon.typing import (
     Slot,
 )
 
+GENESIS_SLOT = Slot(2**32)
+SLOTS_PER_EPOCH = 2**6
+
 
 SERENITY_CONFIG = BeaconConfig(
     # Misc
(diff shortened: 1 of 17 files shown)


eth2/beacon/state_machines/forks/serenity/configs.py
```diff
@@ -14,6 +14,9 @@
     Slot,
 )
 
+GENESIS_SLOT = Slot(2**32)
+SLOTS_PER_EPOCH = 2**6
```
hwwhww: That seems a good way to resolve the dependency. 👍

tests/eth2/beacon/state_machines/forks/test_serenity_operation_processing.py
```diff
@@ -35,8 +35,9 @@ def test_process_max_attestations(genesis_state,
                                   sample_beacon_block_params,
                                   sample_beacon_block_body_params,
                                   config,
-                                  keymap):
-    attestation_slot = 0
+                                  keymap,
+                                  genesis_slot):
+    attestation_slot = genesis_slot
```
hwwhww: Since here the `config` is imported, alternatively:
```python
    attestation_slot = config.GENESIS_SLOT
```

tests/eth2/beacon/state_machines/forks/test_serenity_operation_processing.py
```diff
@@ -92,8 +93,9 @@ def test_process_proposer_slashings(genesis_state,
                                     keymap,
                                     block_root_1,
                                     block_root_2,
-                                    success):
-    current_slot = 1
+                                    success,
+                                    genesis_slot):
+    current_slot = genesis_slot + 1
```
hwwhww: ditto, can use `config.GENESIS_SLOT`.

ralexstokes:
seems like this hit an unreleated CI issue/bug w/ an integration test. rerunning CI (if i can)

PR #374, opened 2019-03-07
HandshakeBase no longer accepts a cancel_token

lithp:
This made the interface more complicated for little benefit. The only
caller of connect could just call the cancel_token itself.

(Opened this because I'm working on a PR which uses a HandshakeResponder but doesn't have a cancel_token handy)

#### Cute Animal Picture

bear-token


p2p/auth.py
```diff
@@ -55,8 +55,8 @@
     the remote.
     """
     use_eip8 = False
-    initiator = HandshakeInitiator(remote, privkey, use_eip8, token)
-    reader, writer = await initiator.connect()
+    initiator = HandshakeInitiator(remote, privkey, use_eip8)
+    reader, writer = await token.cancellable_wait(initiator.connect())
```
cburgdorf: Shouldn't this use the `timeout=REPLY_TIMEOUT` to mimic the current behavior? https://github.com/ethereum/trinity/pull/374/files#diff-3625c6e13b05919e68d2af3a86cd75afL133

lithp: Heh, yep, nice catch! I could have sworn my original commit included this, but also can't imagine any of the rebases I did would have wiped this away :thinking:

tests/core/p2p-proto/test_server.py
```diff
@@ -97,7 +97,7 @@ def get_server(privkey, address):
 async def test_server_incoming_connection(monkeypatch, server, event_loop):
     use_eip8 = False
     token = CancelToken("initiator")
-    initiator = HandshakeInitiator(RECEIVER_REMOTE, INITIATOR_PRIVKEY, use_eip8, token)
+    initiator = HandshakeInitiator(RECEIVER_REMOTE, INITIATOR_PRIVKEY, use_eip8)
     reader, writer = await initiator.connect()
```
carver: Should this be `token.cancellable_wait(initiator.connect())` now? Maybe there are other `connect()` calls around that also need the new cancellable wait...

pipermerriam:
Closing as this isn't viable.

pipermerriam:
see https://github.com/ethereum/trinity/pull/545

PR #375, opened 2019-03-07
Enable eth1 data voting per epoch

### LINKED ISSUE #326: Implement `eth1_data_vote` processing per block
### What is wrong?

We need to implement the per-epoch processing for eth1 data votes.

Reference:  https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#eth1-data-1 

### How can it be fixed

* Implement the spec
* uncomment the line that is marked as `# TODO: state = process_eth1_data_votes(state, self.config)` in the codebase.

edit: add correct link -- this is for epoch processing, not block processing

ralexstokes:
### What was wrong?

Fixes #326.

### How was it fixed?

Adds the per-epoch processing for eth1 data votes.

- [x] TODO: add tests

#### Cute Animal Picture

### DIFF
--- a/tests/eth2/beacon/state_machines/forks/test_serenity_epoch_processing.py
+++ b/tests/eth2/beacon/state_machines/forks/test_serenity_epoch_processing.py
@@ -1,4 +1,5 @@
 import pytest
+import random
 
 from hypothesis import (
     given,
@@ -12,6 +13,11 @@ from eth._utils.numeric import (
 
 from eth.constants import (
     ZERO_HASH32,
+)
+
+from eth_utils.toolz import (
+    assoc,
+    curry,
 )
 
 from eth2._utils.tuple import (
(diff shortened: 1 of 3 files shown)


tests/eth2/beacon/state_machines/forks/test_serenity_epoch_processing.py
```diff
@@ -75,6 +87,118 @@
...
+        (12,),
+        # NOTE: we are accepting more than one block per slot if
+        # there are multiple majorities so no need to test this
+    )
+)
+def test_ensure_update_eth1_vote_if_exists(sample_beacon_state_params,
+                                           config,
+                                           vote_offsets):
+    # one less than a majority is the majority divided by 2
+    threshold = _majority_threshold(config) / 2
```
hwwhww: To avoid `float`, maybe use `//` here?
```suggestion
    threshold = _majority_threshold(config) // 2
```

tests/eth2/beacon/state_machines/forks/test_serenity_epoch_processing.py
```diff
@@ -75,6 +87,118 @@
...
+    data_votes = tuple(
+        Eth1DataVote(
+            eth1_data=Eth1Data(
+                deposit_root=_random_bytes(32),
+                block_hash=_random_bytes(32),
```
hwwhww: IMO you can just use `offset` as the entropy to generate `hash`, e.g., `hash_eth2(b'block_hash' + offset.to_bytes(32, 'little'))`. It may be easier to reproduce the testcase.

ralexstokes: good catch!

PR #347, opened 2019-03-01
Fix pruning with consecutive finished dependencies

carver:
### What was wrong?

Fixes #344 

I added a test to reproduce: if you set two chained finished dependencies, then try to prune the oldest one, it crashes because we didn't link one finished dependencies as a dependency of the other.

### How was it fixed?

Check if the parent of a finished dependency is already present, and link it as a dependency, if so. Pruning works as expected, now.

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #389, opened 2019-03-09
Several Fast sync fixes: header parent log, consecutive header batching & no unnecessary backtracking

### LINKED ISSUE #387: HeaderNotFound during sync
I can only see two options for how this happened, either one is a bug, and I can't see how either might be possible, yet:

1. Header sync returned a header with a gap: the parent was not yet in the DB
2. Header sync returned an old header that was already sent (which is valid), but the `OrderedTaskPreparation` object had already pruned it's parent before its parent had been written to the DB. (that part shouldn't happen)

Note that #347 now has a failsafe restart in case the bug reappears so trinity shouldn't crash, but it's still a bug that needs to be hunted down.

---

This is from `7b9e956fedf936c63ed1413ebec7face6962fe5c` a recent version of #347.

`HeaderNotFound`: https://gist.github.com/cburgdorf/3d74f0b9a1c56ee8c9bda108ac7fcf51

_Originally posted by @cburgdorf in https://github.com/ethereum/trinity/pull/347#issuecomment-468957767_

carver:
### What was wrong?

#347 had a bug during logging, due to a stray comma.

Also, fix #387 

### How was it fixed?

Fix the bug by deleting the comma and add more logging.

Force header syncer to send consecutive headers in a single batch.

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/common/headers.py
+++ b/trinity/sync/common/headers.py
@@ -6,8 +6,9 @@ from random import randrange
 from typing import (
     AsyncIterator,
     Callable,
-    Generic,
     FrozenSet,
+    Generic,
+    Iterable,
     Tuple,
     Type,
     cast,
@@ -722,6 +723,19 @@ class HeaderMeatSyncer(BaseService, PeerSubscriber, Generic[TChainPeer]):
             raise
 
 
+def first_nonconsecutive_header(headers: Iterable[BlockHeader]) -> int:
+    """
+    :return: index of first child that does not match parent header, or a number
+        past the end if all are consecutive
+    """
+    for index, (parent, child) in enumerate(sliding_window(2, headers)):
+        if child.parent_hash != parent.hash:
+            return index + 1
+
+    # return an index off the end to indicate that all headers are consecutive
+    return index + 2
+
+
 class BaseHeaderChainSyncer(BaseService, HeaderSyncerAPI, Generic[TChainPeer]):
     """
     Generate a skeleton header, then use all peers to fill in the headers
@@ -780,7 +794,7 @@ class BaseHeaderChainSyncer(BaseService, HeaderSyncerAPI, Generic[TChainPeer]):
             max_batch_size: int = None) -> AsyncIterator[Tuple[BlockHeader, ...]]:
 
         while self.is_operational:
-            next_header_batch = await self.wait(self._stitcher.ready_tasks(max_batch_size))
+            headers = await self.wait(self._stitcher.ready_tasks(max_batch_size))
             if self._stitcher.has_ready_tasks():
                 # Even after clearing out a big batch, there is no available capacity, so
                 # pause any coroutines that might wait for capacity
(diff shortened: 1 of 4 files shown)


trinity/sync/full/chain.py
```diff
@@ -538,15 +545,16 @@ def register_peer(self, peer: BasePeer) -> None:
             canonical_tip = None
 
         self.logger.debug(
-            "Header syncer returned header %s, which is not in our DB. "
-            "Instead at #%d, our header is %s, whose parent is %s, with canonical tip %s. ",
-            "The highest received header is %d.",
+            "Header syncer returned header %s, which has no parent in our DB. "
+            "Instead at #%d, our header is %s, whose parent is %s, with canonical tip %s. "
+            "The highest received header is %d. Triggered by missing dependency: %s",
```
pipermerriam: I *think* wrapping this in one extra layer of parens would have made this error easier to spot.

carver:
Ok, so far it looks like requiring consecutive headers to come out of the header syncer is doing the trick (which makes sense).

I did notice that once, the really old headers sent sync back in time, which should be an easy fix. I'm going to add that and then this should be ready for review.

carver:
Fixed the problem that was causing backtracking, plus a bug in the consecutive headers batching. Ready for review!

carver:
@cburgdorf are you familiar enough with sync now to take a look at this PR?

trinity/sync/common/headers.py
```diff
@@ -780,15 +774,22 @@ def _reset_buffer(self) -> None:
...
                 self._buffer_capacity.set()
-            yield cast(Tuple[BlockHeader, ...], next_header_batch)
+
+            while headers:
+                split_idx = first_nonconsecutive_header(headers)
+                consecutive_batch, headers = headers[:split_idx], headers[split_idx:]
+                if headers:
+                    # Note lack of capacity if the headers are non-consecutive
+                    self._buffer_capacity.clear()
+                yield cast(Tuple[BlockHeader, ...], consecutive_batch)
```
cburgdorf: Do you know why this `cast` is needed?

trinity/sync/full/chain.py
```diff
@@ -466,40 +469,59 @@ def register_peer(self, peer: BasePeer) -> None:
...
                 except HeaderNotFound:
-                    await self._log_missing_parent(headers[0], highest_block_num)
+                    await self._log_missing_parent(new_headers[0], highest_block_num, missing_exc)
 
                     # Nowhere to go from here, reset and try again
                     await self._header_syncer.clear_buffer()
 
-                    # Don't try to process `headers`, wait for new ones to come in
+                    # Give time for queue to flush out
+                    await self.sleep(60)
```
cburgdorf: Maybe this should be a constant in `trinity.sync.common.constants`?

carver: Actually, there was some kind of bug with the reset, so I will just remove it instead.

trinity/sync/common/headers.py
carver: It was moved from a previous incantation. Appears to not be necessary anymore... :+1:

PR #395, opened 2019-03-11
Use new Lahja `filter_predicate` API

cburgdorf:
### What was wrong?

Currently, there are a number of high frequency events that we can not broadcast efficiently because it would waste a lot of resources in Trinity, slowing other important tasks down.

A good example for that are new transactions broadcasted by peers. It's our goal to migrate to an architecture where the `PeerPool` runs in its own process and is only concerned with pure peer IO leaving the handling of messages (and hence db lookups etc) to other processes.

Under that model, the `PeerPool` would simply broadcast new incoming transactions leaving it to other processes to deal with them. The `PeerPool` should not know and care *who* is interested in the new transactions (a transaction pool plugin, a transaction analyzer plugin etc.).

However, simply pushing these transactions into each and every process isn't the right answer either. Processes should be able to opt-in to these events.

### How was it fixed?

The upcoming [lahja release](https://github.com/ethereum/lahja/pull/34) introduces a new feature that allows lahja endpoints to shield themselve from events that they don't need. This does **not** affect any events that are deliberately directed towards some endpoint - those do always reach the recipient! 

This PR uses that feature and sets things up so that `BasePlugin` has a method `should_receive_broadcast(event: BaseEvent) -> bool` that shields plugins from all events that are not deliberately sent to it (With the exception of `AvailableEndpointsUpdated` that all plugins should receive).

To stick with our example and assuming we have a `PeerPool` that relays events on the event bus, a plugin that wants to receive new transactions can overwrite `should_receive_broadcast` to allow such event to reach it. 

The PR also renames some APIs to better reflect what they do and align with the renames in the lahja library.

#### Cute Animal Picture

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -25,6 +25,7 @@ from eth.vm.forks.spurious_dragon import SpuriousDragonVM
 
 from lahja import (
     ConnectionConfig,
+    ListenerConfig,
 )
 
 from trinity.config import (
@@ -136,7 +137,9 @@ async def event_bus(event_loop):
         path=ipc_path
     )
     await endpoint.start_serving(networking_connection_config, event_loop)
-    await endpoint.connect_to_endpoints(networking_connection_config)
+    await endpoint.add_listener_endpoints(
+        ListenerConfig.from_connection_config(networking_connection_config)
+    )
     try:
         yield endpoint
     finally:
--- a/trinity/extensibility/plugin.py
+++ b/trinity/extensibility/plugin.py
@@ -24,6 +24,7 @@ from typing import (
 
 from lahja import (
     ConnectionConfig,
+    ListenerConfig,
 )
 
 from trinity.config import (
@@ -34,6 +35,9 @@ from trinity.constants import (
 )
 from trinity.endpoint import (
     TrinityEventBusEndpoint,
+)
+from trinity.events import (
+    AvailableEndpointsUpdated,
 )
 from trinity.extensibility.events import (
     BaseEvent,
@@ -280,6 +284,19 @@ class BaseIsolatedPlugin(BasePlugin):
         """
         return self._process
 
+    @staticmethod
+    def should_receive_broadcast(event: BaseEvent) -> bool:
+        """
+        Default ``filter_predicate`` that prevents events that aren't deliberately send
+        to the endpoint of this plugin to get pushed into it. Plugins can overwrite this
+        to control fine-grained access about the events it wants to receive.
+        """
+
+        if isinstance(event, AvailableEndpointsUpdated):
+            return True
+
+        return False
+
     def start(self) -> None:
         """
         Prepare the plugin to get started and eventually call ``do_start`` in a separate process.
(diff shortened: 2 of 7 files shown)


lithp:
I'll take a look at this soon, probably tomorrow!

tests/conftest.py
```diff
@@ -136,7 +137,9 @@ def async_process_runner(event_loop):
         path=ipc_path
     )
     await endpoint.start_serving(networking_connection_config, event_loop)
-    await endpoint.connect_to_endpoints(networking_connection_config)
+    await endpoint.add_listener_endpoints(
+        ListenerConfig.from_connection_config(networking_connection_config)
```
lithp: This is more of a nitpick but I'm always going to push back on code like this. :)

I'm really sensitive to making new objects, in this case `ListenerConfig` is performing zero work, all it's doing is wrapping the `ConnectionConfig`! It signifies that there is no `filter_predicate`, but passing the `ConnectionConfig` would do the same trick. Could `filter_predicate` be an attribute of the `ConnectionConfig`?

trinity/extensibility/plugin.py
```diff
@@ -280,6 +284,19 @@ def process(self) -> Process:
...
 
+    @staticmethod
+    def should_receive_broadcast(event: BaseEvent) -> bool:
+        """
+        Default ``filter_predicate`` that prevents events that aren't deliberately send
+        to the endpoint of this plugin to get pushed into it. Plugins can overwrite this
+        to control fine-grained access about the events it wants to receive.
+        """
+
+        if isinstance(event, AvailableEndpointsUpdated):
```
lithp: I think this signifies a leaky abstraction. Anyone who wants to use their own `filter_predicate` has to remember to let `AvailableEndpointsUpdated` through, but that's an implementation detail of `MainEventBusEndpoint`, not something consumers of this class care about!

tests/conftest.py
cburgdorf: >Could `filter_predicate` be an attribute of the `ConnectionConfig`

The thing is that `ConnectionConfig` is used for the `start_serving_*` APIs and it doesn't make sense for those to accept a `filter_predicate` that's why I separated these two into different configuration types.

trinity/extensibility/plugin.py
cburgdorf: I know what you mean. Here's a real-world example from my sync work branch that I'm currently using.

```
@staticmethod
    def should_receive_broadcast(event: BaseEvent) -> bool:

        return any((
            isinstance(event, PeerJoinedEvent),
            isinstance(event, PeerLeftEvent),
            isinstance(event, PeerPoolMessageEvent) and isinstance(event.cmd, NewBlock),
            BaseIsolatedPlugin.should_receive_broadcast(event),
        ))
```

I guess if that method wasn't `staticmethod` but `classmethod` that `super().should_receive_broadcast(event)` would also work making it more ergonomic. It's still a thing that plugin authors need to remember though.

cburgdorf:
Thank you for the review Brian!

>I think it's possible to get the performance improvement without changing any code in trinity!

Interesting! Haven't read the lahja review yet. Will check it out in a moment.

>his PR doesn't wire anything up, do you have an example of what the code will look like once the plugins start using it?

it does wire the `should_receive_broadcast` up so as a plugin author all you need to do is to overwrite that method with your own implementation that has custom whitelisting logic.

Here's a real world example that I'm currently using to work on the isolated process sync:

https://github.com/cburgdorf/trinity/blob/c84f49fa6ca1c04e13af0f211e2194de9594fcd9/trinity/plugins/builtin/tmp_sync/plugin.py#L62-L70

cburgdorf:
This won't be needed when @lithp lands his latest lahja PR https://github.com/ethereum/lahja/pull/36

PR #396, opened 2019-03-12
Fix for sync getting stuck (deadlocked)

carver:
### What was wrong?

In certain scenarios, switching header syncers could cause sync to get stuck waiting on capacity.

### How was it fixed?

Header sync: reset capacity if queue is empty before loop.

When non-consecutive headers are split into batches in header sync, the syncer considers that a loss of capacity to handle more headers. This never gets reset, so if the queue is totally empty, then you could get deadlocked. Now, we move the queue clearing check to the end of the loop, which should always prevent that deadlock.

#### Cute Animal Picture


carver:
Fast sync seems to be rolling along quite smoothly for me after this change. (I haven't touched state sync yet, though)

PR #398, opened 2019-03-12
Avoid unnecessary clamp on `previous_epoch` by `genesis_epoch`

### LINKED ISSUE #382: Make `get_previous_epoch` return `current_epoch - 1` in all cases
### What is wrong?
Part of Trinity #379
Spec PRs:

### How can it be fixed
Update `BeaconState.previous_epoch()` function

ralexstokes:
### What was wrong?

Fixes #382.

### How was it fixed?

Updated the calculation hanging off the `BeaconState`.

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/epoch_processing_helpers.py
+++ b/eth2/beacon/epoch_processing_helpers.py
@@ -56,12 +56,11 @@ if TYPE_CHECKING:
 def get_previous_epoch_boundary_attestations(
         state: 'BeaconState',
         slots_per_epoch: int,
-        genesis_epoch: Epoch,
         latest_block_roots_length: int) -> Iterable[PendingAttestationRecord]:
     beacon_block_root = get_block_root(
         state,
         get_epoch_start_slot(
-            state.previous_epoch(slots_per_epoch, genesis_epoch),
+            state.previous_epoch(slots_per_epoch),
             slots_per_epoch,
         ),
         latest_block_roots_length,
@@ -75,7 +74,6 @@ def get_previous_epoch_boundary_attestations(
 def get_previous_epoch_matching_head_attestations(
         state: 'BeaconState',
         slots_per_epoch: int,
-        genesis_epoch: Epoch,
         latest_block_roots_length: int) -> Iterable[PendingAttestationRecord]:
     for attestation in state.previous_epoch_attestations:
         beacon_block_root = get_block_root(
@@ -167,7 +165,6 @@ def get_epoch_boundary_attesting_balances(
         previous_epoch: Epoch,
         state: 'BeaconState',
         config: 'BeaconConfig') -> Tuple[Gwei, Gwei]:
-
     previous_epoch_boundary_root = get_block_root(
         state,
         get_epoch_start_slot(previous_epoch, config.SLOTS_PER_EPOCH),
--- a/eth2/beacon/epoch_processing_helpers.py
+++ b/eth2/beacon/epoch_processing_helpers.py
@@ -164,7 +164,7 @@ def get_epoch_boundary_attester_indices(
 def get_attesting_indices(state: 'BeaconState',
                           attestations: Sequence[PendingAttestationRecord],
                           config: CommitteeConfig) -> Sequence[ValidatorIndex]:
-    output = set()
+    output: Set[ValidatorIndex] = set()
     for a in attestations:
         participants = get_attestation_participants(state, a.data, a.aggregation_bitfield, config)
         output = output.union(participants)
--- a/eth2/beacon/epoch_processing_helpers.py
+++ b/eth2/beacon/epoch_processing_helpers.py
@@ -147,28 +147,83 @@ def get_winning_root_and_participants(
 def get_epoch_boundary_attester_indices(
         state: 'BeaconState',
         attestations: Sequence[PendingAttestationRecord],
-        expected_justified_epoch: Epoch,
-        epoch_for_root: Epoch,
-        config: 'BeaconConfig') -> Iterable[ValidatorIndex]:
+        epoch: Epoch,
+        root: Hash32,
+        committee_config: CommitteeConfig) -> Iterable[ValidatorIndex]:
     for a in attestations:
-        root = get_block_root(
-            state,
-            get_epoch_start_slot(
-                epoch_for_root,
-                config.SLOTS_PER_EPOCH
-            ),
-            config.LATEST_BLOCK_ROOTS_LENGTH
-        )
-        matches_justified_epoch = a.data.justified_epoch == expected_justified_epoch
-        matches_block_root = a.data.epoch_boundary_root == root
-        should_include = matches_justified_epoch and matches_block_root
-        if should_include:
+        if a.data.justified_epoch == epoch and a.data.epoch_boundary_root == root:
             yield from get_attestation_participants(
                 state,
                 a.data,
                 a.aggregation_bitfield,
-                CommitteeConfig(config),
+                committee_config,
             )
+
+
+@to_tuple
+def get_attesting_indices(state: 'BeaconState',
+                          attestations: Sequence[PendingAttestationRecord],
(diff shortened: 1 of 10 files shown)


eth2/beacon/epoch_processing_helpers.py
```diff
@@ -142,16 +140,25 @@ def get_winning_root(
 def get_epoch_boundary_attester_indices(
```
ralexstokes: looks like @hwwhww and @ChihChengLiang were involved here -- would love your input on this...

don't exactly like getting the block root inside this function (mainly bc we then have to `monkeypatch` it later to test it) but the way the spec is written we don't want to call this function unless there are some attestations in `attestations`.

open to changing this bit if we can think of a nicer/cleaner way.... i'll update if i think of something i like better

ralexstokes:
@hwwhww @ChihChengLiang  this became more involved than the spec change would suggest (surprise!) but is ready for review -- i called attention to the part i'm not super excited about and would love to hear any thoughts!

eth2/beacon/epoch_processing_helpers.py
ChihChengLiang: We don't want to call this function again and again for every attestations, either.
Maybe something like:
```
if len(attestations)>0:
    root = helpers.get_block_root(...)
```

ralexstokes: @ChihChengLiang i don't follow -- given that we are iterating over `attestations` then we only call `helpers.get_block_root` if `len(attestations)>0` so that condition would always be true....

i think what will be nicer is if I trace up the call stack here and just avoid calling this entirely -- it seems like there is only one place it is being called in the function that determines if the various epochs are even justifiable.

ralexstokes:
Ok, I took advantage of some fresh helpers from spec version `v0.5.0` which handle things like "previous epoch to genesis" gracefully... 

I deleted the prior test for the function that no longer explicitly exists, with the understanding that it is only used by the function that gets the balances for the given boundary attestations which is already tested (if there is a bug in this covered function then it will fail the wrapping function's tests)

eth2/beacon/epoch_processing_helpers.py
```diff
@@ -138,75 +134,92 @@ def get_winning_root(
...
+        config: 'BeaconConfig') -> Sequence[PendingAttestationRecord]:
+    slot = get_epoch_start_slot(epoch, config.SLOTS_PER_EPOCH)
+    return [
+        a for a in attestations
+        if a.data.epoch_boundary_root == get_block_root(
+            state,
+            slot,
+            config.LATEST_BLOCK_ROOTS_LENGTH
+        )
+    ]
```
hwwhww: I think we would try to avoid returning mutable `List`.

eth2/beacon/epoch_processing_helpers.py
```diff
@@ -138,75 +134,92 @@ def get_winning_root(
...
 
 
-def get_epoch_boundary_attesting_balances(
-        current_epoch: Epoch,
-        previous_epoch: Epoch,
+def _get_current_epoch_boundary_attestations(
         state: 'BeaconState',
-        config: 'BeaconConfig') -> Tuple[Gwei, Gwei]:
-
+        config: 'BeaconConfig') -> Sequence[PendingAttestationRecord]:
```
hwwhww: As https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#function-variance mentioned, we should narrow down the scope of return type:

- Normally, use `Sequence` for the input parameters.
- If it returns known tuple, use `Tuple` for the return type.
- If there's `@to_tuple` + `yield` for output, use `Iterable` for the return type.

eth2/beacon/epoch_processing_helpers.py
```diff
@@ -138,75 +134,92 @@ def get_winning_root(
...
-        CommitteeConfig(config),
-    )
 
-    previous_epoch_boundary_attesting_balance = get_total_balance(
-        state.validator_balances,
-        previous_epoch_boundary_attester_indices,
-        config.MAX_DEPOSIT_AMOUNT,
+def _get_previous_epoch_boundary_attestations(
+        state: 'BeaconState',
+        config: 'BeaconConfig') -> Sequence[PendingAttestationRecord]:
```
hwwhww: ditto

eth2/beacon/epoch_processing_helpers.py
```diff
@@ -138,75 +134,92 @@ def get_winning_root(
...
-                state,
-                a.data,
-                a.aggregation_bitfield,
-                committee_config,
-            )
+        config: 'BeaconConfig') -> Sequence[PendingAttestationRecord]:
+    slot = get_epoch_start_slot(epoch, config.SLOTS_PER_EPOCH)
+    return [
+        a for a in attestations
+        if a.data.epoch_boundary_root == get_block_root(
```
NIC619: We can compute the epoch boundary root beforehand.

eth2/beacon/epoch_processing_helpers.py
```diff
@@ -138,75 +134,92 @@ def get_winning_root(
...
-@to_set
-def get_epoch_boundary_attester_indices(
+def get_attesting_indices(state: 'BeaconState',
+                          attestations: Sequence[PendingAttestationRecord],
+                          config: CommitteeConfig) -> Sequence[ValidatorIndex]:
+    output: Set[ValidatorIndex] = set()
+    for a in attestations:
+        participants = get_attestation_participants(state, a.data, a.aggregation_bitfield, config)
+        output = output.union(participants)
+    return sorted(list(output))
```
ChihChengLiang: Looks like it can sort the set directly
```suggestion
    return sorted(list(output))
    return sorted(output)
```

eth2/beacon/epoch_processing_helpers.py
ralexstokes: good call -- as long as `Tuple` is a subtype of `Sequence` here then I'm happy w/ it :)

i don't know `mypy` well enough to answer this question but I suspect the subtyping relation holds...

eth2/beacon/epoch_processing_helpers.py
ralexstokes: if you are referring to the call to `get_block_root` i started having the same thought but then i remembered why i was doing this in the first place.... and it was because we end up w/ a call to this function for essentially slots before the genesis epoch (where `get_block_root` will explode)

we purposefully put the call inside the conditional so that it is lazily evaluated.... if this becomes a hotspot perf-wise then we can do something like

```python
slot = # ....
if len(attestations) != 0:
    block_root = get_block_root(...)
return tuple(
    a for a in ... # rest of expression
)
```

or even just add some memoization to this function.... tbh i don't expect we will need to do this in this case :)

eth2/beacon/epoch_processing_helpers.py
ralexstokes: nice! let's get that unnecessary list constructor out of there...

ralexstokes:
@hwwhww would love your thoughts on what I did here https://github.com/ethereum/trinity/pull/398/commits/03d3ddf4474ac6b3825f065994e87b43863c6caf to satisfy mypy... i feel like the `@to_tuple + generator expression` is less efficient than what I had before but mypy was saying I would have to otherwise `cast` to a `Tuple[T]` (as the inferred type w/ the prior implementation was `Tuple[T,...]`

hwwhww:
@ralexstokes (sorry for the late reply!)
Yes, it seems right about how to make mypy happy - either `@to_tuple` + `yield` + `Iterable` or tuple comprehension + `Tuple`. IMO they are both fine as we can avoid `List` in both approaches.

/cc @pipermerriam any other trade-off here?

ralexstokes:
this got out of date -- i'll rebase and then merge this in

ralexstokes:
Closing in favor of #469.

PR #400, opened 2019-03-13
Fix broken configuration of individual loggers

cburgdorf:
### What was wrong?

Trinity can be configured to set log levels individually per module. However that wasn't working properly because it was only setup for the main process leaving out the `db`, `networking` as well as all isolated plugin processes.

E.g. if one tries to run `trinity -l DEBUG2 -l p2p.discovery=ERROR` to enable all `DEBUG2` logs except for those from `p2p.discovery`, it just doesn't have any effect.

### How was it fixed?

Used `setup_log_levels` in all processes.

#### Cute Animal Picture

### DIFF
--- a/docs/release_notes/trinity.rst
+++ b/docs/release_notes/trinity.rst
@@ -4,6 +4,7 @@ Trinity
 Unreleased (latest source)
 --------------------------
 
+- `#400 <https://github.com/ethereum/trinity/pull/400>`_: Bugfix: Respect configuration of individual logger (e.g -l p2p.discovery=ERROR)
 - `#386 <https://github.com/ethereum/trinity/pull/386>`_: Slightly reduce eventbus traffic that the peer pool causes
 - `#336 <https://github.com/ethereum/trinity/pull/336>`_: Bugfix: Ensure Trinity shuts down if the process pool dies (fatal error)
 
--- a/tests/integration/test_trinity_cli.py
+++ b/tests/integration/test_trinity_cli.py
@@ -159,3 +159,20 @@ async def test_does_not_throw(async_process_runner, command):
     # This is our last line of defence. This test basically observes the first
     # 20 seconds of the Trinity boot process and fails if Trinity logs any exceptions
     await run_command_and_detect_errors(async_process_runner, command, 20)
+
+
+@pytest.mark.parametrize(
+    'command, expected_to_contain_log',
+    (
+        (('trinity', '-l=DEBUG2'), True),
+        # We expect not to contain it because we set the p2p.discovery logger to only log errors
+        (('trinity', '-l=DEBUG2', '-l', 'p2p.discovery=ERROR'), False,)
+    )
+)
+@pytest.mark.asyncio
+async def test_logger(async_process_runner, command, expected_to_contain_log):
+    await async_process_runner.run(command, timeout_sec=20)
+    actually_contains_log = await contains_all(async_process_runner.stderr, {
+        "DiscoveryProtocol  >>> ping",
+    })
+    assert actually_contains_log == expected_to_contain_log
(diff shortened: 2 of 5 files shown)


docs/release_notes/trinity.rst
```diff
@@ -4,6 +4,7 @@ Trinity
 Unreleased (latest source)
 --------------------------
 
+- `#400 <https://github.com/ethereum/trinity/pull/400>`_: Fix bug where configuration of individual logger (e.g -l p2p.discovery=ERROR) had no effect
```
carver: *nitpick* We've been using `Bugfix: ` as a prefix for all the others in the release nodes.

tests/integration/test_trinity_cli.py
```diff
@@ -45,7 +45,7 @@
 @pytest.mark.asyncio
 async def test_full_boot(async_process_runner, command):
     # UPNP discovery can delay things, we use a timeout longer than the discovery timeout
-    await async_process_runner.run(command, timeout_sec=40)
+    await async_process_runner.run(command, timeout_sec=20)
```
carver: Is 20 definitely long enough if upnp discovery times out? Or was this just a local change that slipped in?

tests/integration/test_trinity_cli.py
```diff
@@ -159,3 +159,22 @@
     # This is our last line of defence. This test basically observes the first
     # 20 seconds of the Trinity boot process and fails if Trinity logs any exceptions
     await run_command_and_detect_errors(async_process_runner, command, 20)
+
+
+@pytest.mark.parametrize(
+    'command',
+    (
+        ('trinity', '-l=DEBUG2',),
+        # This test is expected to fail because we mute the p2p.discovery logger
```
carver: Ah, this was a surprise. Maybe we can still set it up to go green instead, with a new helper method `does_not_contain()`

tests/integration/test_trinity_cli.py
cburgdorf: I *think* it is enough but I honestly just wanted to try and see if it works at the advantage of shorter CI builds. It seems to work well with 20 and if it turns out it doesn't we can change it back. Went all-in on *move fast and break things* here :sweat_smile:

tests/integration/test_trinity_cli.py
cburgdorf: I thought about this as well but that would mean that we do one of these things:

- Add a second test that uses the `does_not_contain` (dislike that because it creates two distinct tests that feel better be grouped under one)
- Add another parameter e.g. `assume_existence` on which we then either use `contains_all` or `does_not_contain` (or maybe `contains_none` ?) inside the test.

Based on these two options I found the `xfail` more elegant but I can be convinced the other way :)

tests/integration/test_trinity_cli.py
carver: I just ran a quick blame to see if there were any commits that increased the value like "this was broken at a lower number". Didn't see any :+1:

tests/integration/test_trinity_cli.py
```diff
@@ -159,3 +159,22 @@
...
+        pytest.param(
+            ('trinity', '-l=DEBUG2', '-l', 'p2p.discovery=ERROR'),
+            marks=pytest.mark.xfail
+        ),
+    )
+)
+@pytest.mark.asyncio
+async def test_logger(async_process_runner, command):
+    await async_process_runner.run(command, timeout_sec=20)
+    assert await contains_all(async_process_runner.stderr, {
```
carver: Maybe another boolean param `expected_to_contain_log`:
```py
actually_contains_log = await contains_all(async_process_runner.stderr, { ...

assert actually_contains_log == expected_to_contain_log

cburgdorf: Uh. The simple things can hide themselves so well at times :laughing:

PR #401, opened 2019-03-14
Fix broken docker image build

gsmadi:
### What was wrong?

During docker image build, snappy was not being installed which is a requirement now.

### How was it fixed?

Install `libsnappy-dev` during docker image build.

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -234,6 +234,13 @@ jobs:
         environment:
           TOXENV: py37-wheel-cli
 
+  docker-image-build-test:
+    machine: true
+    steps:
+      - checkout
+      - run: docker build -t ethereum/trinity:test-build .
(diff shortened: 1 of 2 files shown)


pipermerriam:
Is there a way for us to test the Docker image within our CI setup?

gsmadi:
Yes most definitely feasible to build a test image with CircleCI. Will include a job in this PR.

Maybe, we could do:
- Build test image to ensure docker builds work
- Start a container with the built test image and run `trinity --help` to ensure all was installed correctly

cburgdorf:
@gsmadi that sounds like a good approach :+1:

gsmadi:
The above suggestion is implemented and build passing should be a good indicator of good docker builds so we can start distributing more frequently.

PR #402, opened 2019-03-14
Fix RTD doc build

gsmadi:
### What was wrong?

Trinity docs build were failing due to dependency on snappy for required packages during installation. RTD builds docs in docker containers and installing `libsnappy` in the environment is not feasible.

### How was it fixed?

Simplified doc build process by removing necessity to install trinity packages as suggested by @carver. Separated Sphinx requirements into single txt file.

This might require more work. Let me know how to improve it so we can start auto shipping latest docs.

#### Cute Animal Picture

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,10 +1,12 @@
 # https://docs.readthedocs.io/en/latest/yaml-config.html
+version: 2
 build:
   image: latest
 python:
   version: 3.6
-  pip_install: True
-  extra_requirements:
-    - p2p
-    - trinity
-    - doc
+  install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - trinity
+        - doc
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -50,6 +50,7 @@ autodoc_default_options = {
     'undoc-members': None,
 }
 
+autodoc_mock_imports = ["snappy"]
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+import os
 from setuptools import setup, find_packages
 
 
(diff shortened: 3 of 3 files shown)


.readthedocs.yml
```diff
@@ -1,10 +1,8 @@
...
+version: 2
 build:
   image: latest
 python:
   version: 3.6
-  pip_install: True
-  extra_requirements:
-    - p2p
-    - trinity
-    - doc
```
cburgdorf: Isn't it that if we remove this, autodoc won't be able to extract API docs (e.g. https://trinity-client.readthedocs.io/en/latest/api/extensibility/api.extensibility.plugin.html#plugincontext)

I will take this branch and deploy it to RTD and see what happens.

cburgdorf: I just tried it out and it does in fact break autodoc :-(

cburgdorf: I've played with this for a bit but couldn't come up with anything that works. I think RTD needs to install Trinity in order to extract API docs from it. So, what I'm thinking is, maybe there would be some way to only install `python-snappy` when it is not a RTD install. Not sure how we would differentiate but maybe someone has bright ideas.

This would probably cause another problem because, then, some imports wouldn't work but it turns out sphinx has some sort of auto-mocking for exactly that use case.

http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports

gsmadi: Hehe good catch on the autodoc miss in the PR.

We could check for the `READTHEDOCS` environment variable and if false append `python-snappy` as a requirement. That way we avoid having to install snappy in the RTD build environment.

Thanks for pointing out this import mock. Let me try it out.

gsmadi: Thanks for your suggestion @cburgdorf. This is working now using the autodoc mock import for snappy.

cburgdorf: Awesome! I test this live real quick!

cburgdorf:
This looks good for now and I tested it out and it works! Thank you very much @gsmadi for your contribution :pray:

PR #405, opened 2019-03-15
Use less aggressive timeouts for CLI tests

cburgdorf:
### What was wrong?

Yesterday, I reduced timeouts for the CLI test in an attempt to reduce CI times. Turns out reducing from 40 to 20 seconds was a bit too aggressive and it introduced some flakiness. 

### How was it fixed?

Trying if we can live with 30 seconds

#### Cute Animal Picture


cburgdorf:
I'm going wildwest crazy and merge this without review to not slow other PRs down. Judge me later.

pipermerriam:
Im-judging-you

> Evidently this is the top result for "I'm judging you" on the internet....

PR #422, opened 2019-03-17
Fix missing merkle module

NIC619:
### What was wrong?
`ModuleNotFoundError: No module named 'eth2._utils.merkle'`

### How was it fixed?
- [x] Create `eth2._utils.merkle` module
- [x]  fix mypy type checking.
    - [x] Change helpers in `common.py` to `@to_tuple + yield + Iterable type` pattern
    - [x] ignore type check for assigning concatenated tuples to tuple since mypy does not support it yet as suggested in [this issue](https://github.com/python/mypy/issues/224)

### DIFF
--- a/eth2/_utils/merkle/common.py
+++ b/eth2/_utils/merkle/common.py
@@ -9,6 +9,9 @@ from cytoolz import (
     partition,
     take,
 )
+
+from eth_utils import to_tuple
+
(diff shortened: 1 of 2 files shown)


ChihChengLiang:
Why isn't the ci triggered 🤔

PR #428, opened 2019-03-17
Make `test_demo.py` to a single CI job

### LINKED ISSUE #354: Make `test_demo.py` to a single CI job
### What is wrong?
`tests/eth2/beacon/state_machines/test_demo.py` takes ~100 seconds on CircleCI. The main reason why it's so slow is that it takes a lot of time to generate the genesis with proof of possession. Although we should optimize it, it also makes sense to make it to an integration/script tests.

### How can it be fixed
Since it only runs with one config setting once, we can remove the dependency of fixtures and make it more isolated.

hwwhww:
### What was wrong?

Fix #354 

### How was it fixed?
Extract `test_demo` to `py36-eth2-demo`.

#### Cute Animal Picture

animal-2123467_640


tox.ini
```diff
@@ -19,7 +19,8 @@ passenv =
     TRAVIS_EVENT_TYPE
 commands=
     core: pytest -n 4 {posargs:tests/core/}
-    eth2: pytest -n 4 {posargs:tests/eth2}
+    eth2: pytest -n 4 --ignore={posargs:tests/eth2/beacon/demo/} {posargs:tests/eth2}
```
pipermerriam: Rather than using `--ignore` mechanism, how about just adding a layer of depth to the namespacing:

```
tests/eth2/core
tests/eth2/demo
```

ralexstokes:
@hwwhww what do you think about the following:

`/tests/eth2/core/beacon/*`
`/tests/eth2/integration/beacon/*`

so that we can `s/beacon/shard/` when the time comes
and that the "demo" stuff goes into `integration` for now

ralexstokes:
@hwwhww i started making the changes and i'm happy to do so -- just want to confirm as it will involve some import patching, etc. i would prefer to only do once

hwwhww:
Sorry for the delay of this PR! I was thinking about that we can mock BLS verification in this test, as:
https://github.com/ethereum/trinity/blob/ee5951a788fd697bf7c16a4269f472b4bcc8495e/tests/eth2/beacon/state-fixtures/test_minimal_state.py#L29-L52

@ralexstokes the directory looks good to me! Feel free to close this PR and make another one if you've started to work on this. 👍

PR #430, opened 2019-03-19
Initialize index to fix unhandled exception

cburgdorf:
### What was wrong?

When `first_nonconsecutive_header` is called with an empty `Iterable`it crashes with `index used before assignment`

### How was it fixed?

Assign `index` to `0` at the begin of the function. I'm not entirely sure if `2` is the semantically valid response for an empty Iterable but the function should certainly not crash when called with an empty `Iterable`. 

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/common/headers.py
+++ b/trinity/sync/common/headers.py
@@ -8,7 +8,7 @@ from typing import (
     Callable,
     FrozenSet,
     Generic,
-    Iterable,
+    Sequence,
     Tuple,
     Type,
 )
@@ -691,7 +691,7 @@ class HeaderMeatSyncer(BaseService, PeerSubscriber, Generic[TChainPeer]):
             raise
 
 
-def first_nonconsecutive_header(headers: Iterable[BlockHeader]) -> int:
+def first_nonconsecutive_header(headers: Sequence[BlockHeader]) -> int:
     """
     :return: index of first child that does not match parent header, or a number
         past the end if all are consecutive
@@ -701,7 +701,7 @@ def first_nonconsecutive_header(headers: Iterable[BlockHeader]) -> int:
             return index + 1
 
     # return an index off the end to indicate that all headers are consecutive
-    return index + 2
+    return len(headers)
 
 
 class BaseHeaderChainSyncer(BaseService, HeaderSyncerAPI, Generic[TChainPeer]):


trinity/sync/common/headers.py
```diff
@@ -696,6 +696,7 @@ def first_nonconsecutive_header(headers: Iterable[BlockHeader]) -> int:
     :return: index of first child that does not match parent header, or a number
         past the end if all are consecutive
     """
+    index = 0
```
pipermerriam: I *think* instead of this we can add an `else` clause that returns `0`.

```python
for index, ... in enumerate(...):
    if condition:
        return index + 1
else:
    return 0
```

carver:
Yeah, sorry I haven't had a chance to push this one up yet, it was sitting in my working directory with the state sync changes. My patch went something like:

```diff
index 60872ba2..852a7d6a 100644
--- a/trinity/sync/common/headers.py
+++ b/trinity/sync/common/headers.py
@@ -8,7 +8,7 @@ from typing import (
     Callable,
     FrozenSet,
     Generic,
-    Iterable,
+    Sequence,
     Tuple,
     Type,
 )
@@ -691,7 +691,7 @@ class HeaderMeatSyncer(BaseService, PeerSubscriber, Generic[TChainPeer]):
             raise
 
 
-def first_nonconsecutive_header(headers: Iterable[BlockHeader]) -> int:
+def first_nonconsecutive_header(headers: Sequence[BlockHeader]) -> int:
     """
     :return: index of first child that does not match parent header, or a number
         past the end if all are consecutive
@@ -701,7 +701,7 @@ def first_nonconsecutive_header(headers: Iterable[BlockHeader]) -> int:
             return index + 1
 
     # return an index off the end to indicate that all headers are consecutive
-    return index + 2
+    return len(headers)
```

trinity/sync/common/headers.py
cburgdorf: As far as I understand we can not generalize to return `0` if we did encounter a header mismatch because what we need to return then is an index *higher* than the *highest* index. 

If we don't have any items `0` works (as well as any other positive number) but in other cases returning `0` would lead to undesired effect here afaict.

Anyway, I think I'm going with Jasons idea which also works and feels more pure.

PR #432, opened 2019-03-19
Ignore lock time in request timeout

carver:
### What was wrong?

When we get a lot of consistently quick responses from a peer, our timeout reduces. When we locally queue up two peer requests of the same time, one of them locks waiting for the other. Right now, that lock time is counted toward the peer timeout (but not toward the time in the tracker!). So we start timing out quick peers and dropping them!

cc @cburgdorf -- thanks for the indirect reminder to post some of the patches that were in my local branches.

### How was it fixed?

- Stop using the lock time as part of the request timeout
- Leave a little wiggle room in the lock timeout for the previous few requests to complete

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/common/constants.py
+++ b/trinity/protocol/common/constants.py
@@ -4,3 +4,10 @@ ROUND_TRIP_TIMEOUT = 20.0
 # Timeout used when performing the check to ensure peers are on the same side of chain splits as
 # us.
 CHAIN_SPLIT_CHECK_TIMEOUT = 15
+
+
+# We send requests to peers one at a time, but might initiate a few locally before
+# they are sent. This is an estimate of how many get queued locally. The reason we
+# estimate the queue length is to determine how long a timeout to use when
+# waiting for the lock to send the next queued peer request.
+NUM_QUEUED_REQUESTS = 3
--- a/trinity/protocol/common/managers.py
+++ b/trinity/protocol/common/managers.py
@@ -82,8 +82,6 @@ class ResponseCandidateStream(
         """
         outer_timeout = self.response_timeout if timeout is None else timeout
 
-        start_at = time.perf_counter()
-
         # The _lock ensures that we never have two concurrent requests to a
         # single peer for a single command pair in flight.
         try:
@@ -93,6 +91,8 @@ class ResponseCandidateStream(
                 f"Timed out waiting for {self.response_msg_name} request lock "
                 f"or peer: {self._peer}"
             )
+
+        start_at = time.perf_counter()
 
         if timeout is not None or tracker.total_msgs < 20:
             inner_timeout = outer_timeout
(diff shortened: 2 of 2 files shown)


carver:
Argh, master lint is busted

veox:
You'll have to poke the linting jobs (PR #433 should've fixed this).

trinity/protocol/common/constants.py
```diff
@@ -4,3 +4,10 @@
 # Timeout used when performing the check to ensure peers are on the same side of chain splits as
 # us.
 CHAIN_SPLIT_CHECK_TIMEOUT = 15
+
+
+# We send requests to peers one at a time, but might initiate a few locally before
+# they are sent. This is an estimate of how many get queued locally. The reason we
+# estimate the queue length is to determine how long a timeout to use when
+# waiting for the lock to send the next queued peer request.
+NUM_QUEUED_REQUESTS = 3
```
pipermerriam: orthogonal to this PR but related.

Most of these are *fine* as constants, however, I've regularly wanted to override this one:

https://github.com/ethereum/trinity/blob/373ca910f1c5689522001af7ddff58d504720f29/trinity/sync/full/constants.py#L3

And have in the past accidentally  committed that change.  I've wanted a simple environment variable based configuration API for overriding these.  Something like this from the py-evm codebase:

https://github.com/ethereum/py-evm/blob/8d1601a036545b521b439470922f450df29a54e0/eth/_utils/env.py

Not sure it's really worth it at this stage, but something I have continually come back to.

trinity/protocol/common/managers.py
```diff
@@ -82,18 +82,18 @@ def __init__(
         """
         outer_timeout = self.response_timeout if timeout is None else timeout
 
-        start_at = time.perf_counter()
-
         # The _lock ensures that we never have two concurrent requests to a
         # single peer for a single command pair in flight.
         try:
-            await self.wait(self._lock.acquire(), timeout=outer_timeout)
+            await self.wait(self._lock.acquire(), timeout=outer_timeout * NUM_QUEUED_REQUESTS)
```
pipermerriam: I wrote up a big thing about trying to optimize this timeout value for the case where requests are not stacking up and they are happening one at at time.  My motivation was that this timeout value is *correct* for the case where we have congestion, but it is roughly 3x the desired value in the case where there is no congestion.

After writing it up it feels really complicated and so I deleted it, but I wanted to mention this here in case you 1) hadn't thought of it and 2) have any bright ideas on how to account for this.

I don't think it's very important.  :+1:

carver: Hm, if I'm thinking about it right... it won't matter because when it's not congested, then the request won't be stuck waiting for the lock.

pipermerriam: Sooooo obvious once you point it out :woman_facepalming:

PR #434, opened 2019-03-19
Attempt to fix flaky CI with a much longer timeout

ralexstokes:
### What was wrong?

Trying to remove flaky CI testing w/ a longer timeout.

While not ideal, it seems that we are otherwise managing daemons/clients properly so it seems that sometimes CI doesn't launch the `p2pd` daemon subprocess in a timely fashion and it leads to tests blowing up.

### How was it fixed?

Extending the timeout.

#### Cute Animal Picture

### DIFF
--- a/tests/libp2p/p2pclient/test_p2pclient_integration.py
+++ b/tests/libp2p/p2pclient/test_p2pclient_integration.py
@@ -34,7 +34,7 @@ import libp2p.p2pclient.pb.p2pd_pb2 as p2pd_pb
 
 
 NUM_P2PDS = 4
-TIMEOUT_DURATION = 30  # seconds
+TIMEOUT_DURATION = 120  # seconds
 
 
 @pytest.fixture(scope="module")


pipermerriam:
On some of these where we're smoke testing for random errors, I believe they behave as follows:

- run command ....
- monitor stream for errors, early exit if failure detected.
- exit once time limit is hit

Maybe we could add a secondary early exit which watches for known lines that mean we're far enough into the run to justify an early exit.  For example, we could probably exit once we've seen 3-4 lines from the fast-sync status messageds showing the various blocks/per/second data since that means we've successfully booted the client and spun up the main sync code (though admittedly it likely won't have actually started syncing since it is unlikely to have a peer connection).

We could improve this by threading a `geth` node in the background against a pre-seeded database and directly connecting the two clients.

pipermerriam:
Oh, another thing that could speed these commands up is letting them bypass the creation of the genesis state.  IIRC trinity takes a bit of time to create the genesis state on first startup.  Maybe we could freeze this data as a fixture and load it into place for most of the tests?

ralexstokes:
both of these are great suggestions and stuff we will want to keep in mind, esp. as we get closer and closer to a 'mainnet' beacon node!

the timeout mentioned here is just to bound how long it may take for the daemon to boot up... i can't get this to trigger locally and would guess the same for the reader :) aiui usually CI environments can give you very 'bursty' compute and i would guess this is contributing to the issue here.

while im thinking about it, probably the better thing to do here was to re-use the daemon across the entire `libp2p` test suite and definitely something we should do moving forward. i'll keep this in mind as we integrate more and more of the `libp2p` stack in `trinity`

PR #435, opened 2019-03-20
Add additional sleep and remove useless `read`

mhchia:
### What was wrong?
CI fails several times in `tests/libp2p/p2pclient/test_p2pclient_integration.py::test_pubsub_client_subscribe` at , e.g. [this failure](https://circleci.com/gh/ethereum/trinity/19539). I guess it is because `StreamWriter` is not closed successfully before the query `await p2pds[0].pubsub.get_topics()`. The code is [here](https://github.com/ethereum/trinity/blob/master/tests/libp2p/p2pclient/test_p2pclient_integration.py#L1033). 

### How was it fixed?
Sleep for a short period, to yield the execution to the closing transport. Still need to confirm if this is the exact reason. I ran the test in my local machine with this patch 100 times, and no failure happens.

#### Cute Animal Picture

### DIFF
--- a/tests/libp2p/p2pclient/test_p2pclient_integration.py
+++ b/tests/libp2p/p2pclient/test_p2pclient_integration.py
@@ -1029,6 +1029,6 @@ async def test_pubsub_client_subscribe(p2pds):
     assert PeerID(pubsub_msg_2_0.from_field) == peer_id_0
     # test case: unsubscribe by closing the stream
     writer_0.close()
-    await asyncio.sleep(0.1)
+    await asyncio.sleep(0)
     assert topic not in await p2pds[0].pubsub.get_topics()
     assert peer_id_0 not in await p2pds[1].pubsub.list_peers(topic)
--- a/tests/libp2p/p2pclient/test_p2pclient_integration.py
+++ b/tests/libp2p/p2pclient/test_p2pclient_integration.py
@@ -1029,6 +1029,6 @@ async def test_pubsub_client_subscribe(p2pds):
     assert PeerID(pubsub_msg_2_0.from_field) == peer_id_0
     # test case: unsubscribe by closing the stream
     writer_0.close()
-    await reader_0.read() == b""
+    await asyncio.sleep(0.1)
     assert topic not in await p2pds[0].pubsub.get_topics()
     assert peer_id_0 not in await p2pds[1].pubsub.list_peers(topic)


tests/libp2p/p2pclient/test_p2pclient_integration.py
```diff
@@ -1029,6 +1029,6 @@ def _get_current_running_dispatched_handler():
     assert PeerID(pubsub_msg_2_0.from_field) == peer_id_0
     # test case: unsubscribe by closing the stream
     writer_0.close()
-    await reader_0.read() == b""
+    await asyncio.sleep(0.1)
```
pipermerriam: In tests, `0.1` is a long time.  Can you drop this to just an `await asyncio.sleep(0)` and `0.01` if that doesn't work.

mhchia: Sure. Changed it to `0`.

carver:
Hm, I'm still seeing libp2p issues: https://circleci.com/gh/ethereum/trinity/19947?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I'll take a quick look.

mhchia:
@carver  It seems the failing tests were trying to start `p2pd` before the ones started in the previous test were cleaned up. I will look more into this issue. Thanks for pointing it out.

carver:
FWIW, I think that one got resolved by #443 -- but there is a different one that seems to be popping up sporadically: https://circleci.com/gh/ethereum/trinity/20331?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

mhchia:
sorry I failed to notice the PR #443 . Thanks a lot for fixing the issue.

>  but there is a different one that seems to be popping up sporadically: https://circleci.com/gh/ethereum/trinity/20331?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

This one involves the interaction between two `p2pd`. I can come up with two options:
1. Use the [sleep time](https://github.com/libp2p/go-libp2p-pubsub/blob/master/floodsub_test.go#L793) in `go-libp2p-pubsub` for the subscription information to be propagated to the peers. The main concern from me is, in my impression sometimes the tests in `go-libp2p-pubsub` also failed due to this kind of issue.
2. Add polling here like [this](https://github.com/ethereum/trinity/blob/master/tests/libp2p/p2pclient/test_p2pclient_integration.py#L371-L380), which keeps polling `list_peers` several times. If after too many trials fail then we fail the test. The disadvantage is the uglier polling loop.

Personally, I prefer the second option, and will make a PR for this. Really sorry for incurring the problems.

PR #436, opened 2019-03-20
Connect to preferred nodes even if discovery off

carver:
### What was wrong?

I wanted discovery off but to connect to a particular preferred node, ie~
```sh
trinity --disable-discovery --preferred-node "enode..."
```

Trinity never attempted to connect to the preferred node.

### How was it fixed?

Replace the `NoopDiscoveryService` with a `StaticDiscoveryService` which only connects to the preferred nodes. (This should cleanly downgrade to the noop case if no preferred nodes are set)

#### Cute Animal Picture

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -61,9 +61,14 @@ from eth.tools.logging import ExtendedDebugLogger, DEBUG2_LEVEL_NUM
 
 from cancel_token import CancelToken, OperationCancelled
 
-from p2p.events import PeerCandidatesRequest, RandomBootnodeRequest
+from p2p.events import (
+    PeerCandidatesRequest,
+    RandomBootnodeRequest,
+    BaseRequestResponseEvent,
+    PeerCandidatesResponse,
+)
 from p2p.exceptions import AlreadyWaitingDiscoveryResponse, NoEligibleNodes, UnableToGetDiscV5Ticket
-from p2p.kademlia import to_uris
+from p2p.kademlia import to_uris, Node as KademliaNode
 from p2p import kademlia
 from p2p import protocol
 from p2p.service import BaseService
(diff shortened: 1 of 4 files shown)


carver:
First time I've tinkered in this area. Thoughts on the approach, or ideas for a better one @cburgdorf ?

pipermerriam:
I only did a shallow pass to refresh my memory of the code.  Here's what I think.

1. The concept of preferred nodes should be completely removed from discovery.
2. A preferred node plugin should be written up that activates if the `--preferred-node` flag is provided.  That plugin monitors the peer pool, and sends `ConnectToNodeCommand` events when it finds that the peer pool is not empty and isn't connected to one of the preferred nodes.

This would completely decouple discovery and the concept of preferred nodes.

carver:
Seems reasonable

> 2\. when it finds that the peer pool is not empty and isn't connected to one of the preferred nodes.

Why wait until the peer pool isn't empty? Could/should a preferred node be the first (maybe only) connection?

(This is what inspired the PR, because I wanted to connect to a local geth instance without joining the network)

p2p/discovery.py
```diff
@@ -962,6 +967,54 @@ def get_nodes_to_connect(self, count: int) -> Iterator[kademlia.Node]:
...
+        )
+
+    async def _run(self) -> None:
+        self.run_daemon_task(self.handle_get_peer_candidates_requests())
+        self.run_daemon_task(self.handle_get_random_bootnode_requests())
+
+        await self.cancel_token.wait()
+
+
 class NoopDiscoveryService(BaseService):
```
cburgdorf: It looks like this PR makes the `NoopDiscoveryService` obsolete (as the `StaticDiscoveryService` steps into this role if no `preferred_nodes` are provided. So I think `NoopDiscoveryService` should be removed because otherwise it just ends up as dead code not being used at all.

carver:
Any objections to me merging this and making a follow-up issue for decoupling the two?

I'd prefer to focus on syncing for the moment.

Otherwise I am happy to just leave it open until I can get back to it.

pipermerriam:
no objections as long as you create a tracking issue.

p2p/discovery.py
carver: Since we'll probably use it again as soon as the decoupling (#440) happens, maybe I'll just leave it in for now.

carver:

PR #439, opened 2019-03-21
Get rid of sync methods in BaseAsyncChainDB

cburgdorf:
### What was wrong?

`BaseAsyncChainDB` inherited from `BaseChainDB` and therefore inherited all *abstract synchronous* which then further leads to us having to implement these *synchronous*  methods on the `AsyncChainDBProxy` which is problematic for two reasons:

1. [It is confusing](https://github.com/ethereum/trinity/pull/307#discussion_r267678406) (should I choose the sync or async version)?

2. It can lead to bugs or performance issues when one uses a sync methods where an async method should be used (in fact, this PR discovered such thing)

### How was it fixed?

1. Make `BaseAsyncChainDB` inherit from `BaseAsyncHeaderDB` instead of `BaseChainDB`. Not only fixes that our problems, it is also *symmetric* with `BaseChainDB` inheriting from `BaseHeaderDB`.

2. Fix all issues that were discovered by this change

Shout out to @ChihChengLiang who reminded me of this inconsistency in the code base.

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/light/service.py
+++ b/trinity/sync/light/service.py
@@ -301,7 +301,7 @@ class LightPeerChain(PeerSubscriber, BaseService, BaseLightPeerChain):
         head_number = peer.head_number
         if head_number - header.block_number > MAX_REORG_DEPTH:
             # The peer claims to be far ahead of the header we requested
-            if self.headerdb.get_canonical_block_hash(header.block_number) == block_hash:
+            if await self.headerdb.coro_get_canonical_block_hash(header.block_number) == block_hash:
                 # Our node believes that the header at the reference hash is canonical,
                 # so treat the peer as malicious
                 raise BadLESResponse(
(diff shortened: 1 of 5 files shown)


trinity/sync/light/service.py
```diff
@@ -301,7 +301,7 @@ def callback(r: protocol._DecodedMsgType) -> None:
         head_number = peer.head_number
         if head_number - header.block_number > MAX_REORG_DEPTH:
             # The peer claims to be far ahead of the header we requested
-            if self.headerdb.get_canonical_block_hash(header.block_number) == block_hash:
```
carver: Is this a bugfix worth mentioning in the release notes?

cburgdorf: Afaict, the old code worked just fine because the `sync` version *could* be used, it's just blocking the event loop so this new version *might* be more efficient (kinda depends on the overhead of offloading things to the `ThreadPoolExecutor` but that's a different story).

Might be too subtle to justify a release note entry.

PR #441, opened 2019-03-21
Set a custom network ID if genesis file supplied

carver:
### What was wrong?

I wanted to run with a custom genesis file, and my instinct was to set a custom network ID (didn't seem right to use the mainnet ID). This was explicitly forbidden.

### How was it fixed?

- I relaxed the constraint, so that if you supply a genesis file, then you can also use a custom network ID.
- Also updated the schema to accept <32 characters for genesis extraData, since that seems permitted by the yellow paper (note: the chain builder tool creates a genesis with `extra_data=b''` by default)
- Set bootstrap nodes to empty if they aren't supplied with a custom network ID

Note that if you use a custom network id without bootnodes or preferred nodes then obviously it won't connect to anyone. (somewhat related to #436 )

#### Cute Animal Picture

### DIFF
--- a/trinity/assets/eip1085.schema.json
+++ b/trinity/assets/eip1085.schema.json
@@ -115,7 +115,7 @@
         },
         "extraData": {
           "type": "string",
-          "pattern": "^0x([0-9a-fA-F]{2}){32}$"
+          "pattern": "^0x([0-9a-fA-F]{2}){0,32}$"
         },
         "gasLimit": {
           "type": "string",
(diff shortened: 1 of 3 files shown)


trinity/assets/eip1085.schema.json
```diff
@@ -115,7 +115,7 @@
         },
         "extraData": {
           "type": "string",
-          "pattern": "^0x([0-9a-fA-F]{2}){32}$"
+          "pattern": "^0x([0-9a-fA-F]{2}){0,32}$"
```
carver: I know this isn't related much to the rest of the PR (just came up while I was doing genesis testing). Since the yellow paper explicitly permits <32 byte extradata, this seemed reasonable to me. In fact, I created a chain that would be impossible to generate a genesis file for if we require 32 bytes here.

pipermerriam: :+1: EIP1085 isn't by any means well defined or finalized and this schema is of my own making.

carver:

PR #442, opened 2019-03-21
Switch to long-running release notes PR

carver:
### What was wrong?

Some issues with including release notes in each PR:
- ripe for conflicts: people likely to add their own line to the beginning or end of the list
- requires pushing ones to get the PR number, then again. Potentially wastes a whole suite of test-runs.

### How was it fixed?

- Created a new label that should go on the running PR
- Change the notes to reference it

This should probably not merge until we have created the running PR (will do after I track down this libp2p issue that broke master)

#### Cute Animal Picture

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,7 +5,8 @@
 ### How was it fixed?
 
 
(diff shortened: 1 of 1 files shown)


carver:

carver:
For example: #445

PR #444, opened 2019-03-21
Make the block importing functionality swappable

carver:
### What was wrong?

In the new "beam sync", I want to reuse *almost* all the functionality of the regular syncer, but with a modified block import.

### How was it fixed?

Modified the regular syncer to take a new class is that is solely responsible for block import. In the new "beam sync", we swap in a block importer that requests any missing trie nodes that are found while executing the block.

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/common/chain.py
+++ b/trinity/sync/common/chain.py
@@ -1,3 +1,4 @@
+from abc import ABC, abstractmethod
 from typing import (
     AsyncIterator,
     Tuple,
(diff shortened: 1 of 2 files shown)


cburgdorf:
>I have a vague memory of him working on something that moved block importing into an isolated process.

That's correct. It never moved past the PoC stage and will be obsolete once the entire isolated plugin process sync kicks in (which I promise will happen this century)

PR #446, opened 2019-03-21
Unify some fast & regular syncer logic

carver:
### What was wrong?

The fast syncer got a lot of love, and the regular syncer missed out on a lot of those updates, because it was duplicated. When testing out regular sync, it was obvious it needed all the same upgrades.

### How was it fixed?

Combined the relevant logic, isolating a few pieces that are different. Had to ask mypy very nicely to allow me to make the changes.

#### Cute Animal Picture

### DIFF
--- a/trinity/_utils/datastructures.py
+++ b/trinity/_utils/datastructures.py
@@ -368,7 +368,6 @@ class BaseOrderedTaskPreparation(ABC, Generic[TTask, TTaskID]):
         pass
 
 
-
 class OrderedTaskPreparation(
         BaseOrderedTaskPreparation[TTask, TTaskID],
         Generic[TTask, TTaskID, TPrerequisite]):
--- a/trinity/_utils/datastructures.py
+++ b/trinity/_utils/datastructures.py
@@ -1,3 +1,4 @@
+from abc import ABC, abstractmethod
 from asyncio import (
     AbstractEventLoop,
     Lock,
@@ -349,7 +350,28 @@ class NoPrerequisites(Enum):
     pass
 
 
-class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
+class BaseOrderedTaskPreparation(ABC, Generic[TTask, TTaskID]):
+    @abstractmethod
+    def set_finished_dependency(self, finished_task: TTask) -> None:
+        pass
+
+    @abstractmethod
+    def register_tasks(self, tasks: Tuple[TTask, ...], ignore_duplicates: bool = False) -> None:
+        pass
+
+    @abstractmethod
+    async def ready_tasks(self, max_results: int = None) -> Tuple[TTask, ...]:
+        pass
+
+    @abstractmethod
+    def has_ready_tasks(self) -> bool:
+        pass
+
+
+
+class OrderedTaskPreparation(
+        BaseOrderedTaskPreparation[TTask, TTaskID],
+        Generic[TTask, TTaskID, TPrerequisite]):
     """
     This class is useful when a series of tasks with prerequisites must be run
     sequentially. The prerequisites may be finished in any order, but the
--- a/trinity/_utils/headers.py
+++ b/trinity/_utils/headers.py
@@ -1,16 +1,17 @@
 from typing import (
+    AsyncIterator,
+    Awaitable,
+    Callable,
     cast,
     Iterable,
-    AsyncIterator,
     Tuple,
     TypeVar,
 )
 
 from eth.constants import UINT_256_MAX
 from eth.rlp.headers import BlockHeader
-from p2p.service import BaseService
+from eth.tools.logging import ExtendedDebugLogger
 
-from trinity.db.eth1.header import BaseAsyncHeaderDB
 from trinity.exceptions import OversizeObject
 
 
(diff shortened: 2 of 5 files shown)


trinity/_utils/datastructures.py
```diff
@@ -349,7 +350,27 @@ class NoPrerequisites(Enum):
     pass
 
 
-class OrderedTaskPreparation(Generic[TTask, TTaskID, TPrerequisite]):
+class BaseOrderedTaskPreparation(ABC, Generic[TTask, TTaskID]):
```
carver: This allowed the usage of two OrderedTaskPrep objects that have different prerequisites but the same task & ID definitions.

trinity/sync/full/chain.py
```diff
@@ -114,13 +115,16 @@ def __init__(self,
...
                  peer_pool: ETHPeerPool,
+                 header_syncer: HeaderSyncerAPI,
                  token: CancelToken = None) -> None:
         super().__init__(token=token)
         self.chain = chain
         self.db = db
         self._peer_pool = peer_pool
         self._pending_bodies = {}
 
+        self._header_syncer = header_syncer
```
carver: Simple deduplication from both syncers

trinity/sync/full/chain.py
```diff
@@ -130,10 +134,131 @@ def __init__(self,
         buffer_size = MAX_BODIES_FETCH * REQUEST_BUFFER_MULTIPLIER
         self._block_body_tasks = TaskQueue(buffer_size, attrgetter('block_number'))
 
+        # Track if there is capacity for more block importing
+        self._db_buffer_capacity = asyncio.Event()
+        self._db_buffer_capacity.set()  # start with capacity
```
carver: This was in the fast syncer, but not regular.

trinity/sync/full/chain.py
```diff
@@ -130,10 +134,131 @@ def __init__(self,
...
+
+        # Track if any headers have been received yet
+        self._got_first_header = asyncio.Event()
```
carver: This was in regular syncer, but not fast. (helpful for things like deciding when to show logs and/or initiate timers)

trinity/sync/full/chain.py
```diff
@@ -130,10 +134,131 @@ def __init__(self,
...
+
     async def _run(self) -> None:
         with self.subscribe(self._peer_pool):
             await self.cancellation()
 
+    async def _sync_from_headers(
+            self,
```
carver: This is pretty close to straight copy of the original `_launch_prerequisites` in fast sync, with a couple differences noted below.

trinity/sync/full/chain.py
```diff
@@ -130,10 +134,131 @@ def __init__(self,
...
+                task_integrator.register_tasks(headers, ignore_duplicates=True)
+            except MissingDependency as missing_exc:
+                # The parent of this header is not registered as a dependency yet.
+                # Some reasons this might happen, in rough descending order of likelihood:
+                #   - a normal fork: the canonical head isn't the parent of the first header synced
+                #   - a bug: headers were queued out of order in new_sync_headers
+                #   - a bug: old headers were pruned out of the tracker, but not in DB yet
+
+                # Skip over all headers found in db, (could happen with a long backtrack)
+                new_headers = await self._skip_existing_headers(headers, completion_check)
```
carver: The different sync strategies have different concepts of what a "done" block looks like. So this method was changed to have a generic completion check that's passed in.

trinity/sync/full/chain.py
```diff
@@ -130,10 +134,131 @@ def __init__(self,
...
+                )
+                # Set first header's parent as finished
+                task_integrator.set_finished_dependency(parent_header)
+                # Re-register the header tasks, which will now succeed
+                task_integrator.register_tasks(new_headers, ignore_duplicates=True)
+                # Clobber the headers variable so that the follow-up work below is consistent with
+                # or without exceptions (ie~ only add headers not in DB to body/receipt queue)
+                headers = new_headers
+
+            yield headers
```
carver: Instead of actually launching the tasks, this method is now only responsible for deciding which headers to register, and dealing with what happens when the registration goes bad. Whatever method iterates over `_sync_from_headers` is now responsible for actually launching the tasks that will cause completion of the OrderedTaskPrep.

trinity/sync/full/chain.py
```diff
@@ -130,10 +134,131 @@ def __init__(self,
...
+        # Track the highest registered block header by number, purely for stats/logging
+        highest_block_num = -1
+
+        async for headers in self.wait_iter(get_headers_coro):
+            self._got_first_header.set()
+            try:
+                # We might end up with duplicates that can be safely ignored.
+                # Likely scenario: switched which peer downloads headers, and the new peer isn't
+                # aware of some of the in-progress headers
+                task_integrator.register_tasks(headers, ignore_duplicates=True)
```
carver: which task integration object you use is different per chain, so that is passed in as an argument. (I was deliberately avoiding subclassing methods here, preferring composition over inheritance)

trinity/_utils/headers.py
```diff
@@ -45,30 +47,32 @@ def sequence_builder(start_number: T,
...
+        completion_check: Callable[[BlockHeader], Awaitable[bool]]) -> AsyncIterator[BlockHeader]:
     """
     We only want headers that are missing, so we iterate over the list
     until we find the first missing header, after which we return all of
     the remaining headers.
     """
     iter_headers = iter(headers)
     for header in iter_headers:
-        is_present = await service.wait(db.coro_header_exists(header.hash))
+        is_present = await completion_check(header)
```
carver: different callers have different concepts of what it means for a header to be done, generalized a bit.

trinity/sync/full/chain.py
```diff
@@ -299,6 +393,46 @@ def _mark_body_download_complete(
 
         return block_body_bundles
 
+    async def _log_missing_parent(
```
carver: This was only in fast, but useful for regular

trinity/sync/full/chain.py
```diff
@@ -457,69 +585,22 @@ def register_peer(self, peer: BasePeer) -> None:
...
-                    parent_header,
-                )
-                # Set first header's parent as finished
-                self._block_persist_tracker.set_finished_dependency(parent_header)
-                # Re-register the header tasks, which will now succeed
-                self._block_persist_tracker.register_tasks(new_headers, ignore_duplicates=True)
-                # Clobber the headers variable so that the follow-up work below is consistent with
-                # or without exceptions (ie~ only add headers not in DB to body/receipt queue)
-                headers = new_headers
+        async for headers in self._sync_from_headers(
```
carver: Now `_launch_prerequisite_tasks` is muuuuch shorter in each of fast and regular sync.

carver:
Pretty sure the failure is unrelated

trinity/_utils/headers.py
carver: I will add a note that calling services should do `self.wait(skip_complete_headers)`. I actually wanted to pull service out as a dependency altogether, just passing in the logger.

PR #448, opened 2019-03-22
Fix async beacon chain db

ChihChengLiang:
### What was wrong?

Sync methods block the loop https://github.com/ethereum/trinity/pull/307#discussion_r267699499, and BCC peer is still using them to access chain database.

### How was it fixed?

- Use BaseAsyncBeaconChainDB and async methods in BeaconChainSyncer and BeaconContext
- Change abstract methods in BaseAsyncBeaconChainDB from sync to async
- Add FakeAsyncBeaconChainDB to test 
- Fix all related tests

Also, remove sync methods in AsyncHeaderDBPreProxy

(Genesis slots are not addressed in this PR)

#### Cute Animal Picture

image
This cat caused my eyes allergy today 🙀

### DIFF
--- a/tests/core/p2p-proto/bcc/test_full_sync.py
+++ b/tests/core/p2p-proto/bcc/test_full_sync.py
@@ -42,19 +42,20 @@ async def get_sync_setup(request, event_loop, alice_chain_db, bob_chain_db):
 async def test_sync_from_genesis(request, event_loop):
     genesis = create_test_block(slot=0)
     bob_blocks = (genesis,) + create_branch(length=99, root=genesis)
-    alice_chain_db = get_chain_db((genesis,))
-    bob_chain_db = get_chain_db(bob_blocks)
+    alice_chain_db = await get_chain_db((genesis,))
+    bob_chain_db = await get_chain_db(bob_blocks)
 
     alice_syncer = await get_sync_setup(request, event_loop, alice_chain_db, bob_chain_db)
 
     await alice_syncer.events.finished.wait()
 
-    assert alice_chain_db.get_canonical_head(BeaconBlock).slot == 99
-    assert (alice_chain_db.get_canonical_head(BeaconBlock) ==
-            bob_chain_db.get_canonical_head(BeaconBlock))
+    alice_head = await alice_chain_db.coro_get_canonical_head(BeaconBlock)
+    bob_head = await bob_chain_db.coro_get_canonical_head(BeaconBlock)
+    assert alice_head.slot == 99
+    assert alice_head == bob_head
     for slot in range(100):
-        alice_block = alice_chain_db.get_canonical_block_by_slot(slot, BeaconBlock)
-        bob_block = bob_chain_db.get_canonical_block_by_slot(slot, BeaconBlock)
+        alice_block = await alice_chain_db.coro_get_canonical_block_by_slot(slot, BeaconBlock)
+        bob_block = await bob_chain_db.coro_get_canonical_block_by_slot(slot, BeaconBlock)
         assert alice_block == bob_block
 
 
@@ -63,19 +64,21 @@ async def test_sync_from_old_head(request, event_loop):
     genesis = create_test_block(slot=0)
     alice_blocks = (genesis,) + create_branch(length=49, root=genesis)
     bob_blocks = alice_blocks + create_branch(length=50, root=alice_blocks[-1])
-    alice_chain_db = get_chain_db(alice_blocks)
-    bob_chain_db = get_chain_db(bob_blocks)
+    alice_chain_db = await get_chain_db(alice_blocks)
+    bob_chain_db = await get_chain_db(bob_blocks)
 
     alice_syncer = await get_sync_setup(request, event_loop, alice_chain_db, bob_chain_db)
 
     await alice_syncer.events.finished.wait()
 
-    assert alice_chain_db.get_canonical_head(BeaconBlock).slot == 99
-    assert (alice_chain_db.get_canonical_head(BeaconBlock) ==
-            bob_chain_db.get_canonical_head(BeaconBlock))
+    alice_head = await alice_chain_db.coro_get_canonical_head(BeaconBlock)
+    bob_head = await bob_chain_db.coro_get_canonical_head(BeaconBlock)
+
+    assert alice_head.slot == 99
+    assert alice_head == bob_head
     for slot in range(100):
-        alice_block = alice_chain_db.get_canonical_block_by_slot(slot, BeaconBlock)
-        bob_block = bob_chain_db.get_canonical_block_by_slot(slot, BeaconBlock)
+        alice_block = await alice_chain_db.coro_get_canonical_block_by_slot(slot, BeaconBlock)
+        bob_block = await bob_chain_db.coro_get_canonical_block_by_slot(slot, BeaconBlock)
         assert alice_block == bob_block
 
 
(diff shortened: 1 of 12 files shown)


trinity/sync/beacon/chain.py
```diff
@@ -133,7 +134,7 @@ def select_sync_peer(self) -> BCCPeer:
             last_block = batch[-1]
 
             try:
-                self.chain_db.persist_block_chain(batch, BeaconBlock)
+                await self.chain_db.coro_persist_block_chain(batch, BeaconBlock)  #type:ignore
```
ChihChengLiang: Not sure why mypy can resolve other async types but not this

tests/core/p2p-proto/bcc/test_full_sync.py
```diff
@@ -42,19 +42,20 @@ def finalizer():
...
-            bob_chain_db.get_canonical_head(BeaconBlock))
+    alice_head = await alice_chain_db.coro_get_canonical_head(BeaconBlock)
+    bob_head = await bob_chain_db.coro_get_canonical_head(BeaconBlock)
+    assert alice_head.slot == 99
+    assert alice_head == bob_head
     for slot in range(100):
-        alice_block = alice_chain_db.get_canonical_block_by_slot(slot, BeaconBlock)
-        bob_block = bob_chain_db.get_canonical_block_by_slot(slot, BeaconBlock)
+        alice_block = await alice_chain_db.coro_get_canonical_block_by_slot(slot, BeaconBlock)
+        bob_block = await bob_chain_db.coro_get_canonical_block_by_slot(slot, BeaconBlock)
```
ChihChengLiang: Not sure if I'm doing right with these loops

trinity/db/eth1/header.py
```diff
@@ -85,16 +84,6 @@ def __init__(self, db: BaseAtomicDB) -> None:
     coro_persist_header = async_method('persist_header')
     coro_persist_header_chain = async_method('persist_header_chain')
 
-    get_block_header_by_hash = sync_method('get_block_header_by_hash')
```
cburgdorf: Oh! Looks like I forgot to remove those in my previous PR. Thanks for the cleanup :+1:

trinity/sync/beacon/chain.py
```diff
@@ -133,7 +134,7 @@ def select_sync_peer(self) -> BCCPeer:
             last_block = batch[-1]
 
             try:
-                self.chain_db.persist_block_chain(batch, BeaconBlock)
+                await self.chain_db.coro_persist_block_chain(batch, BeaconBlock)  # type:ignore
```
cburgdorf: Do you know why the ignore is needed here? What's the error you are seeing?

trinity/sync/beacon/chain.py
```diff
@@ -68,7 +68,7 @@ def is_sync_peer_selected(self) -> bool:
                 raise Exception("Invariant: Cannot exceed max retries")
 
             try:
-                self.sync_peer = self.select_sync_peer()
+                self.sync_peer = await self.select_sync_peer()
```
cburgdorf: I *think* that this should be wrapped as `self.wait(self.select_sync_peer())`. An `await` expression that is wrapped with `self.wait(...)` will stop *awaiting*  when the `CancelToken` is triggered. I want to add though that I'm not super confident on judging which parts should be wrapped with `self.wait(...)` and which shouldn't. My gut feeling says that `await` expression that will turn back to the caller by itself anyway (such as: get me the current head from the db) don't need to be wrapped with `self.wait(...)` but those that would otherwise be *awaiting*  for a long / infinite time should be wrapped. I think we should `await get_expert_opinion(`@carver`)`

trinity/sync/beacon/chain.py
ChihChengLiang: mypy was complaining about
```
trinity/sync/beacon/chain.py:137: error: Incompatible types in "await" (actual type "Tuple[Tuple[BaseBeaconBlock, ...], Tuple[BaseBeaconBlock, ...]]", expected type "Awaitable[Any]")
```
but it was fixed after I change the abstract methods in BaseAsyncBeaconChainDB from sync methods to async methods.

trinity/db/beacon/chain.py
```diff
@@ -107,19 +116,20 @@ class AsyncBeaconChainDBPreProxy(BaseAsyncBeaconChainDB):
...
+    coro_get_canonical_head = async_method('get_canonical_head')
+    coro_get_finalized_head = async_method('get_finalized_head')
+    coro_get_block_by_root = async_method('get_block_by_root')
+    coro_get_score = async_method('get_score')
+    coro_block_exists = async_method('block_exists')
+    coro_persist_block_chain = async_method('persist_block_chain')
+    coro_get_state_by_root = async_method('get_state_by_root')
+    coro_persist_state = async_method('persist_state')
+    coro_exists = async_method('exists')
+    coro_get = async_method('get')
```
ChihChengLiang: Added these diff to make #307 work. Otherwise these error messages are shown:

```
  File "/Users/liangcc/projects/EAPL/eth2.0s/trinity/p2p/peer_pool.py", line 277, in connect
    peer = await self.wait(handshake(remote, self.get_peer_factory()))
  File "/Users/liangcc/projects/EAPL/eth2.0s/trinity/p2p/cancellable.py", line 20, in wait
  File "/Users/liangcc/projects/EAPL/eth2.0s/trinity/p2p/cancellable.py", line 42, in wait_first
  File "/usr/local/var/pyenv/versions/3.7.0/envs/py-evm/lib/python3.7/site-packages/cancel_token/token.py", line 152, in cancellable_wait
    return done.pop().result()
  File "/Users/liangcc/projects/EAPL/eth2.0s/trinity/p2p/peer.py", line 127, in handshake
    await peer.do_sub_proto_handshake()
  File "/Users/liangcc/projects/EAPL/eth2.0s/trinity/p2p/peer.py", line 290, in do_sub_proto_handshake
    await self.send_sub_proto_handshake()
  File "/Users/liangcc/projects/EAPL/eth2.0s/trinity/trinity/protocol/bcc/peer.py", line 59, in send_sub_proto_handshake
    BeaconBlock,
  File "/Users/liangcc/projects/EAPL/eth2.0s/trinity/trinity/_utils/mp.py", line 33, in method
    args,
  File "/usr/local/var/pyenv/versions/3.7.0/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/local/var/pyenv/versions/3.7.0/lib/python3.7/multiprocessing/managers.py", line 811, in _callmethod
    raise convert_to_error(kind, result)
multiprocessing.managers.RemoteError:
---------------------------------------------------------------------------
  File "/usr/local/var/pyenv/versions/3.7.0/lib/python3.7/multiprocessing/managers.py", line 248, in serve_client
    (methodname, type(obj), exposed)
AttributeError: method 'coro_get_canonical_block_by_slot' of <class 'trinity._utils.mp.TracebackRecorder'> object is not in exposed={'get_state_by_root', 'exists', 'get_finalized_head', 'persist_block_chain', 'persist_state', 'get_canonical_head_root', 'get_block_by_root', 'block_exists', 'get_score', 'get', 'get_canonical_block_by_slot', 'get_canonical_block_root_by_slot', 'get_slot_by_root', 'get_canonical_head', 'get_canonical_block_root', 'persist_block'}

  File "/usr/local/var/pyenv/versions/3.7.0/lib/python3.7/multiprocessing/managers.py", line 271, in serve_client
    fallback_func = self.fallback_mapping[methodname]
KeyError: 'coro_get_canonical_block_by_slot'
---------------------------------------------------------------------------
```

trinity/sync/beacon/chain.py
carver: Yeah, wrapping in a `self.wait()` sounds right. Roughly always do it, but you can sometimes leave it out for an performance benefit.  If a class calls its own methods, then only the outermost one needs the `self.wait()` because it will cancel all the pending tasks underneath it.

The worst time to forget to add `wait()` is when it causes a service to hang on shutdown.

trinity/db/beacon/chain.py
```diff
@@ -30,47 +30,56 @@ class BaseAsyncBeaconChainDB:
     """
```
mhchia: Should `BaseAsyncBeaconChainDB` subclass `ABC`?

cburgdorf:
Good catch! It should.

trinity/protocol/bcc/servers.py
```diff
@@ -123,21 +116,20 @@ def _get_blocks(self,
 
         yield start_block
 
-        blocks_generator = cons(start_block, (
-            # TODO: pass accurate `block_class: Type[BaseBeaconBlock]` under
-            # per BeaconStateMachine fork
-            self.db.get_canonical_block_by_slot(slot, BeaconBlock)
-            for slot in itertools.count(start_block.slot + 1)
-        ))
-        max_blocks_generator = take(max_blocks, blocks_generator)
```
ChihChengLiang: I ruined the beautiful generators here, it seems like they don't support async methods well.

trinity/sync/beacon/chain.py
```diff
@@ -90,10 +90,10 @@ def is_sync_peer_selected(self) -> bool:
 
         await self.sync()
```
cburgdorf: I *think*  this should also be wrapped as `self.wait(self.sync())` and then you should be on the safe side because the rest is then happening further down the stack from `sync()` so it should be implicitly covered by that outer `self.wait(...)`.

PR #449, opened 2019-03-22
Bugfix Regular sync when incoming header is not in memory

carver:
### What was wrong?

#446 had a bug (although it didn't actually make things worse, just didn't fix something that it promised to X)

In the scenario where a header comes in that's not in memory (say because you pruned old headers and switched peers), the check for whether a header was in the db was wrong.

### How was it fixed?

The fix was an easy one-liner. The longer part was making a test that actually triggers it.

The rough idea was to start up a two regular syncers at the same time. They both consider the db to be at genesis.
- start syncer 1 while syncer 2 is paused
- only let syncer 1 grab a single header, and then after it finishes syncing the block, make it shut down
- after that block is synced, start up syncer 2
- syncer 2 now receives a header at block # 2, which is a surprise, but one the syncer needs to handle

This code path is what wasn't being tested, and why the bug from #446 got missed

#### Cute Animal Picture

### DIFF
--- a/tests/core/integration_test_helpers.py
+++ b/tests/core/integration_test_helpers.py
@@ -104,6 +104,7 @@ class FakeAsyncRopstenChain(RopstenChain):
 
 class FakeAsyncMainnetChain(MainnetChain):
     chaindb_class = FakeAsyncChainDB
+    coro_get_canonical_head = async_passthrough('get_canonical_head')
     coro_import_block = coro_import_block
     coro_validate_chain = async_passthrough('validate_chain')
     coro_validate_receipt = async_passthrough('validate_receipt')
@@ -111,6 +112,7 @@ class FakeAsyncMainnetChain(MainnetChain):
 
 class FakeAsyncChain(MiningChain):
     coro_import_block = coro_import_block
+    coro_get_canonical_head = async_passthrough('get_canonical_head')
     coro_validate_chain = async_passthrough('validate_chain')
     coro_validate_receipt = async_passthrough('validate_receipt')
     chaindb_class = FakeAsyncChainDB
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -2,10 +2,16 @@ import asyncio
 
 import pytest
 
+from p2p.service import BaseService
+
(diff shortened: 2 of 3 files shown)


pipermerriam:
Wrote this up to try and capture an idea for how we could make testing this stuff easier: https://github.com/ethereum/trinity/issues/454

carver:
> LGTM. I don't _fully_ follow the scenario, I need to re-read it tomorrow when I can think straight. From a cursory review, this looks good.

I'll see if I can do a better job giving the background @cburgdorf :

Before worrying about how we end up in the state, the main point is that the `OrderedTaskPrep` object is throwing a `MissingDependency` which means we are trying to track a new header task when the OTP doesn't have its parent registered. We handle this by checking if the parent is already in the database, and if so then we just tell the OTP object: we know the dependency is good, carry on. There was a dumb bug in that path, and we weren't exercising it.

So how do we end up in that state where the OTP doesn't know about the parent? The primary scenario is that the header syncer gets a long chain, then has those headers pulled out of the queue, then resets to a new peer and gets a long chain again.

Step 1:
The header syncer has headers queued up: A, B, C, D, E
The body syncer queue is empty

Step 2:
The header syncer queue is empty
The body syncer has download tasks in progress for headers: A, B, C, D, E

Step 3:
The header syncer switches peers, and queues up headers: B, C, D, E, F again
The body syncer still has download tasks in progress for headers: A, B, C, D, E

Step 4:
The header syncer has headers in queue: B, C, D, E, F
The body syncer finishes tasks, writes them to database, and prunes the old tasks from memory. Maybe memory only has references to headers: D, E

Step 5:
The body syncer pulls header B, and gets a `MissingDependency` from its `OrderedTaskPrep`, so it checks the database to make sure A is already persisted. It is, and realizes the children are too, so it skips over all processed headers, and starts processing header F

---

But that scenario would be *really* tricky to simulate in a test, so I just manually persist one block, then restart another body syncer that doesn't know about the persisted block from the beginning, and will throw a `MissingDependency` right away.

carver:
I'll go ahead and merge, and fix forward if something else comes up :+1:

cburgdorf:
@carver Great explanation! Thanks for the extra effort Jason! It's much clearer now!

PR #397, opened 2019-03-12
Headerification and root juggling

ralexstokes:
### What was wrong?

Closes #380.

### How was it fixed?

Implementing the changes described in that issue.

edit: sorry this has become a monster PR, happy to split it up if that is helpful for reviewing

changes inside:

- [x]  add `BeaconBlockHeader` type
- [x]  add `get_temporary_block_header`
- [x]  add `get_state_root`
- [x]  add `latest_state_roots`
- [x]  update the `SignatureDomain` enum
- [x]  proposer slashing has two conflicting block headers now instead of proposal
- [x]  move `randao_reveal` and `eth1_data` into block body
- [x]  combine batched roots into `historical_roots` in the `BeaconState`
- [x]  update name for `genesis_eth1_data`
- [x]  update `process_slot_transition` to mirror the spec function `advance_slot`
- [x]  add function `process_block_header`
- [x]  update `process_proposer_slashing`:
- [x]   - moved ahead of spec#711 to ensure that header's epochs are the same (rather than slot)
- [x]   - assert roots of headers are different
- [x]   - updates to verify headers (rather than `Proposals`)
- [x]  various book-keeping for the above changes per-{slot, block, epoch}
- [x] add `latest_block_header` to `BeaconState`

and 
- [ ] updates to tests or new tests as required

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

Issue #380, opened 2019-03-08
Parent root reconstruction + granular state roots + body/header segregation


hwwhww:
### What is wrong?
Part of Trinity #379
Spec PRs:
   - Merged PR: ethereum/eth2.0-specs#711
   - Origin PR: ethereum/eth2.0-specs#649

Issue #391, opened 2019-03-10
epoch transition at start of epoch


hwwhww:
### What is wrong?

Spec PR: https://github.com/ethereum/eth2.0-specs/pull/732

It seems twisted with #380 

### How can it be fixed

Fill this in if you know how to fix it.

ralexstokes:
i started doing this in #397 and will get it all together on `master` :)

ralexstokes:
i have a patch for this: https://gist.github.com/ralexstokes/37f6c3779038705c7795962fd0147c1d

it introduced a bug in the mock attestation utility when we hit an epoch boundary and change the shuffling of the validators... i have not fully debugged but can do so soon.

PR #390, opened 2019-03-09
Merklization friendly pending attestations and update per-epoch processing: crosslink, rewards and penalties

NIC619:
### What was wrong?
Fix #381  #383  #384 

### How was it fixed?
- [x] Split `state.latest_attestations` into `state.previous_epoch_attestations` and `state.current_epoch_attestations`
- [x] Change tie breaks condition for `winning_root`
- [x] Update condition for punishing penalized active validators
- [x] Handle `get_base_reward` when no previous balance
- [x] Remove `SignedGwei` and `RewardSettlementContext`

[//]: # (For important changes, please add a new entry to the release notes file)
[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)

PR #458, opened 2019-03-25
Root juggling

ralexstokes:
### What was wrong?

Number of updates from https://github.com/ethereum/eth2.0-specs/pull/711 concerning the addition of state roots and historical batching.

Addresses part of #380.

Part of breaking up #397.

### How was it fixed?

- [x]  add `get_state_root` helper
- [x]  add `latest_state_roots` to the state
- [x]  combine batched roots into `historical_roots` in the `BeaconState`
- [x]  some bookkeeping w/ the roots

#### Cute Animal Picture


hwwhww:
Somehow this PR needs a rebasing.

ralexstokes:
@hwwhww ok -- rebased!

ralexstokes:
@NIC619 touched some of the stuff in #390 during the rebase so if you could skim that stuff, would be great!

eth2/beacon/state_machines/forks/serenity/slot_processing.py
```diff
@@ -1,37 +1,60 @@
...
+        slots_per_historical_root,
+        previous_block_root,
+    )
 
     state = state.copy(
-        latest_block_roots=tuple(updated_latest_block_roots),
-        batched_block_roots=updated_batched_block_roots,
+        latest_block_roots=updated_latest_block_roots,
+        latest_state_roots=updated_latest_state_roots,
     )
```
hwwhww: Side note:  this function now equals to `advance_slot` in spec 711, but will be torn apart to `cache_state` in #391.

eth2/beacon/helpers.py
```diff
@@ -48,23 +48,26 @@ def get_epoch_start_slot(epoch: Epoch, slots_per_epoch: int) -> Slot:
     return Slot(epoch * slots_per_epoch)
 
 
-def _get_block_root(
-        latest_block_roots: Sequence[Hash32],
+def _get_historical_root(
```
NIC619: I wonder if this will be confused as the function to get root in `state.historical_roots`, but maybe it's just me.

eth2/beacon/state_machines/forks/serenity/slot_processing.py
ralexstokes: yeah, see: https://github.com/ethereum/trinity/issues/391#issuecomment-478187337

like you say, it is in fact tied w/ #380 :)

eth2/beacon/helpers.py
ralexstokes: ah good point -- i agree it could be a point of confusion.

i'm inclined to leave it as is though for the following reasons:

1) there is not a getter for accessing the `historical_roots` in the current spec (this could change)
2) i intentionally mean to communicate this function should be private to this module w/ the `_` prefix. it would be great if we have greater assurance that our visibility will be respected, but such is python :)

PR #459, opened 2019-03-26
[WIP] Add `ValidatorPlugin`

mhchia:
### What was wrong?
Based on https://github.com/ethereum/trinity/pull/307, we can start implementing validator's functionalities. 

### How was it fixed?
- Add a `ValidatorPlugin` as an `BaseIsolatedPlubgin`, without the actual `ValidatorServer` running, which should be finished in the next step.
- Add the flag `--validator` to `trinity-beacon` for `ValidatorPlugin`.

#### TODO
- [ ] Get access to data/event from `BeaconNode`, through `EventBus` or other ways.
- [ ] Run `ValidatorServer` in `ValidatorPlugin` that broadcast blocks when needed.
  - Steps
    - [ ] The service broadcasts a block through `BeaconNode` every N seconds.
    - [ ] The service broadcasts a block only if it finds itself selected as a proposer.

#### Cute Animal Picture

### DIFF
--- a/trinity/main_beacon.py
+++ b/trinity/main_beacon.py
@@ -46,6 +46,7 @@ from trinity.initialization import (
 )
 from trinity.plugins.registry import (
     BASE_PLUGINS,
+    ETH2_NODE_PLUGINS,
 )
 from trinity._utils.ipc import (
     wait_for_ipc,
@@ -66,7 +67,12 @@ from trinity._utils.proxy import (
 
 
 def main_beacon() -> None:
-    main_entry(trinity_boot, APP_IDENTIFIER_BEACON, BASE_PLUGINS, (BeaconAppConfig,))
+    main_entry(
+        trinity_boot,
+        APP_IDENTIFIER_BEACON,
+        BASE_PLUGINS + ETH2_NODE_PLUGINS,
+        (BeaconAppConfig,),
+    )
 
 
 def trinity_boot(args: Namespace,
--- a/trinity/plugins/eth2/validator/plugin.py
+++ b/trinity/plugins/eth2/validator/plugin.py
@@ -1,0 +1,38 @@
+import asyncio
+from argparse import (
+    ArgumentParser,
+    _SubParsersAction,
+)
+from trinity.endpoint import TrinityEventBusEndpoint
+from trinity.extensibility import BaseIsolatedPlugin
+
+
+class ValidatorPlugin(BaseIsolatedPlugin):
+
+    @property
+    def name(self) -> str:
+        return "Validator"
+
+    def configure_parser(self, arg_parser: ArgumentParser, subparser: _SubParsersAction) -> None:
+        arg_parser.add_argument(
+            "--validator",
+            action="store_true",
+            help="Run as the validator node",
+        )
+
+    def on_ready(self, manager_eventbus: TrinityEventBusEndpoint) -> None:
+        args = self.context.args
+        if args.validator:
+            self.start()
+
+    def do_start(self) -> None:
+        loop = asyncio.get_event_loop()
+        # trinity_config = self.context.trinity_config
+
+        # TODO:
+        #   - Add a `ValidatorService` that print something every N seconds.
+        #   - Get access to data/event from `BeaconNode`, through `EventBus` or other ways.
+        #   - The service broadcasts a block through `BeaconNode` every N seconds.
+        #   - The service broadcasts a block if it finds itself selected as a proposer.
+        loop.run_forever()
+        loop.close()
--- a/trinity/plugins/registry.py
+++ b/trinity/plugins/registry.py
@@ -30,6 +30,7 @@ from trinity.plugins.builtin.syncer.plugin import (
     SyncerPlugin,
 )
 from trinity.plugins.eth2.beacon.plugin import BeaconNodePlugin
+from trinity.plugins.eth2.validator.plugin import ValidatorPlugin
 from trinity.plugins.builtin.tx_pool.plugin import (
     TxPlugin,
 )
@@ -52,7 +53,6 @@ BASE_PLUGINS: Tuple[BasePlugin, ...] = (
     FixUncleanShutdownPlugin(),
     JsonRpcServerPlugin(),
     PeerDiscoveryPlugin(),
-    BeaconNodePlugin(),
 )
 
 
@@ -67,6 +67,12 @@ ETH1_NODE_PLUGINS: Tuple[BasePlugin, ...] = (
         NoopSyncStrategy(),
     ), FastThenFullSyncStrategy),
     TxPlugin(),
+)
+
+
+ETH2_NODE_PLUGINS: Tuple[BasePlugin, ...] = (
+    BeaconNodePlugin(),
+    ValidatorPlugin(),
 )
 
 


mhchia:
@cburgdorf Sorry for a dumb question: is it possible to get the `peer_pool` and `db_manager` through event bus in a `BaseIsolatedPlugin`? I saw the [pattern waiting for resources](https://github.com/ethereum/trinity/blob/master/trinity/plugins/builtin/syncer/plugin.py#L234-L249) before the plugin starts. I suspect it works because it is a `BaseAsyncStopPlugin`, but I still wonder if it is possible in our case to get `peer_pool` in a similar way. If not, I guess I can still make `ValidatorPlugin` broadcast things through the events between `BeaconNodePlugin`? Thank you:)

cburgdorf:
>Sorry for some dumb questions:

These aren't dumb at all!

>Is it possible to get the (...) `db_manager` through event bus in a `BaseIsolatedPlugin`

Let's get the easy one out of the way first. You can run this from any process to get a hold of a `DbManager` instance that you can work with.

https://github.com/ethereum/trinity/blob/d56150babd53998e8d315954b5d6fb6774321233/trinity/plugins/builtin/json_rpc/plugin.py#L64


>Is it possible to get the `peer_pool` (...) in a BaseIsolatedPlugin?

Not yet. Interaction with the `PeerPool` is limited from an isolated plugin. This will change *soon* when the `PeerPool` runs as a plugin itself and we'll have a `ProxyPeerPool` that can be used from any process. Here's a more [detailed write-up](https://github.com/ethereum/trinity/issues/460) and this is the [branch](https://github.com/cburgdorf/trinity/tree/christoph/feat/sync-plugin-fiddling) that can give you an outlook into that future with the Eth1 sync running in an isolated process as well.

>I suspect it works because it is a `BaseAsyncStopPlugin` instead of an isolated one

Correct! This kind of plugin runs in the `networking` process which is a hardcoded (read: not bootstrapped through a plugin) that today hosts the Eth1 Node which is basically just the request server and the syncing logic at this point. This will soon be broken up into isolated processes but it all depends on the `ProxyPeerPool` work that I'm currently busy with. The entire concept of `networking` process and `BaseAsyncStopPlugin` will soon retire.

>If it is not possible, would you suggest any pattern to communicate with the peer pool and database across plugins?

Yes, you can communicate via events between the processes (as long as they are picklable). We do the same today for things like allowing the JSON-RPC API (which runs as an `BaseIsolatedPlugin`) to add a peer to the pool

https://github.com/ethereum/trinity/blob/d56150babd53998e8d315954b5d6fb6774321233/trinity/rpc/modules/admin.py#L19-L25

Or to request the current peer count.

https://github.com/ethereum/trinity/blob/d56150babd53998e8d315954b5d6fb6774321233/trinity/rpc/modules/net.py#L27-L35

You just have to pick up the events in the other process and send the response back. Current `master` answers these two from directly inside the `PeerPool` but the branch that I'm working on moves these out of the `PeerPool` and makes things more customizable for derived peer pools.

https://github.com/cburgdorf/trinity/blob/737e8b3f63903c53efd4552c9eb247da44952087/trinity/protocol/common/peer_pool_event_bus.py#L61-L102 

Similarly, you can relay all the events from the `PeerPool` on the event bus which is exactly what my branch is doing. https://github.com/cburgdorf/trinity/blob/737e8b3f63903c53efd4552c9eb247da44952087/trinity/protocol/common/peer_pool_event_bus.py#L119-L156

However, relaying all messages of the `PeerPool` isn't super efficient yet and my branch depends on some unreleased lahja changes to improve efficiency (https://github.com/ethereum/lahja/pull/36 and https://github.com/ethereum/lahja/pull/35)

I'm sorry that the situation is still such a mess. I'm doing my best to push things forward as fast as I can.

mhchia:
@cburgdorf 
Really appreciate your awesome explanations. It does help a lot! I tried `request`/`stream`/`broadcast` pattern between the beacon node plugin and the validator plugin, and it works like a charm. Lahja is really powerful and I still need to dig more into how it works. We probably only need to broadcast specific data from beacon node to the validator, instead of the whole information in peer pool. So IMO this pattern should be enough for our usage for now. 

I have another question about the event-bus when I tried to send request from Validator plugin to Beacon Node plugin through the event-bus. I haven’t investigated deeper to confirm, so please correct me if I’m wrong.

https://github.com/mhchia/trinity/blob/feature/eth2-validator-plugin-handler/trinity/plugins/eth2/beacon/beacon_node_bridge.py#L96-L102

It seems that the request should be made after the endpoints "validator" and "beacon node" are connected, otherwise the request is not received by beacon node. (if the code snippet is commented out, it seems the request is gone.) Is this the desired behavior or I misunderstand something?

FYI for the faster experiment for the testnet, we made the validator functionality along with the slot-ticker live in Beacon Node plugin instead of another Validator plugin in https://github.com/ethereum/trinity/pull/465. We will investigate more on where the Validator should reside after the experiment. (probably a daemon using RPC to communicate with the beacon node or a plugin?)

Thank you:)

cburgdorf:
Thank you for the kind words! I'm happy to hear it's working well for you. We have lots of work on the table but I believe the plugin system with the event-driven architecture will eventually deliver great benefits that attract developers to build on top of Trinity because it's easy to extend with custom features :)

>It seems that the request should be made after the endpoints "validator" and "beacon node" are connected, otherwise the request is not received by beacon node.

Correct! If an event/request is broadcasted it is totally possible for another endpoint to miss out on this event if the connection wasn't setup at the time when the event was fired.

There's nothing built-in to Lahja to help coordinating this initial readyness. When you boot `trinity` as an Eth1 node you'll also notice this warning once or twice in the beginning.

>Discovery did not answer PeerCandidateRequest in time

This is basically the same thing where the `PeerPool` is sending a request to the `discovery` but the first one or two requests end up in oblivion because the discovery plugin (also an isolated process) is just booting and missed out on these events.

So, if it doesn't exactly matter which event is going to be the first one received by the other party you can just wrap the request in a `self.wait(..., timeout=REASONABLE_SECONDS)` and skip waiting for the other party to respond after some time.

https://github.com/ethereum/trinity/blob/20c23d3364fe772111fa504400565855bd5e91d0/p2p/peer_pool.py#L129-L139

You could also use this paradigm to send some form of *ping*  event to the other party and wait until it has first responded with some *pong*.

Another way that you already discovered is to block until `is_connected_to` returns True. I think that Trinity should provide some form of assistance to deal with this type of problem but I don't have a good answer how exactly that should look like from the top of my head.

Let me add two random things that you might find useful:

1. Whenever you do `event_bus.request(...)` or `event_bus.broadcast(...)` the event can be received by any listening party. You can provide a `BroadcastConfig(filter_endpoint='NAME_OF_RECEIVING_ENDPOINT`) to ensure that the event is *only* routed to a specific party. I guess that most requests should really be routed to a specific party but there are many valid cases where `broadcast(...)` should not be restricted so that any plugin can react upon events such as new blocks, transactions etc.

2. More importantly, when you *receive* a request and want to respond to it you should *derive* the broadcast config from the incoming request via `request.broadcast_config()`

https://github.com/ethereum/trinity/blob/20c23d3364fe772111fa504400565855bd5e91d0/p2p/discovery.py#L1025-L1032

This ensures that the response will be routed exclusively to the code path that has made the `request(...)`. In other words, even if you would have multiple code paths making requests of same types or would be listening for the same response type, it would still be guaranteed that the response is exclusively ending up at the right place.

mhchia:
@cburgdorf  Sorry for the late reply, and thanks a lot for your suggestion!! They are indeed helpful:)

mhchia:
Closed in favor of https://github.com/ethereum/trinity/pull/525.

PR #461, opened 2019-03-27
Refactor how PeerPool listens for external EventBus requests

cburgdorf:
### What was wrong?

Currently, the `PeerPool` directly handles requests from the event bus to get the peer count or to connect to a specific node. This hinders extensibility and pollutes the peer pool with things it should not be concerned with.

### How was it fixed?

This is a spin off from my branch that breaks the peer pool, request server and sync into isolated plugins. This is an attempt to slice out some chunks of that work.

The does:

- create a `BasePeerPoolEventBusRequestHandler` that is run by whoever runs the peer pool
- moves all the events that aren't intrinsically needed by the `PeerPool` to `trinity.protocol.common.events.py`

#### Cute Animal Picture


cburgdorf:
@pipermerriam you had reviewed this before as #251 and I think your main concern was

PR #462, opened 2019-03-27
Refactor how PeerPool listens for external EventBus requests

cburgdorf:
### What was wrong?

Currently, the `PeerPool` directly listens for certain events on the event bus that it reacts upon. This is problematic because:

- it pollutes the `PeerPool` with concerns it should not care about
- it's bad for extensibility of such actions for various different peer pool implementations

### How was it fixed?

- Move handling of such events to a `BasePeerPoolEventBusRequestHandler`
- Move all events that aren't intrinsically needed by the `PeerPool` to `trinity.protocol.common.events`

#### Cute Animal Picture

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -23,6 +23,7 @@ from typing import (
     Type,
     TYPE_CHECKING,
 )
+import typing_extensions
 
 import sha3
 
@@ -163,6 +164,32 @@ class BasePeerBootManager(BaseService):
 
 class BasePeerContext:
     pass
+
+
+class IdentifiablePeer(typing_extensions.Protocol):
+    """
+    A protocol used to identify a peer based on the presence of an ``uri`` property. The
+    peer pool uses this to lookup and match DTO peers against the actual real peer instance.
+    The ``BaseDTOPeer`` qualifies for this protocol but usage of the ``BaseDTOPeer`` isn't
+    strictly needed. E.g. one might implement a different DTO class that derives from
+    ``NamedTuple`` which is fine as long as it defines an ``uri`` property.
+    """
+
+    @property
+    @abstractmethod
+    def uri(self) -> str:
+        pass
+
+
+class BaseDTOPeer:
+    """
+    A peer solely meant as a Data Transfer Object (DTO) to travel across process boundaries.
+    It's a shallow, pickleable representation of a peer that carries enough information to
+    make the peer identifiable within the actual pool of real peers.
+    """
+
+    def __init__(self, uri: str):
+        self.uri = uri
 
 
 class BasePeer(BaseService):
--- a/trinity/protocol/common/events.py
+++ b/trinity/protocol/common/events.py
@@ -1,0 +1,53 @@
+from typing import (
+    Type,
+)
+
+from lahja import (
+    BaseEvent,
+    BaseRequestResponseEvent,
+)
+
+from p2p.peer import (
+    IdentifiablePeer,
+)
+from p2p.p2p_proto import (
+    DisconnectReason,
+)
+
+
+class ConnectToNodeCommand(BaseEvent):
+    """
+    Event that wraps a node URI that the pool should connect to.
+    """
+
+    def __init__(self, node: str) -> None:
+        self.node = node
+
+
+class PeerCountResponse(BaseEvent):
+    """
+    Response event that wraps the count of peers connected to the pool.
+    """
+
+    def __init__(self, peer_count: int) -> None:
+        self.peer_count = peer_count
+
+
+class PeerCountRequest(BaseRequestResponseEvent[PeerCountResponse]):
+    """
+    Request event to get the count of peers connected to the pool.
+    """
+
+    @staticmethod
+    def expected_response_type() -> Type[PeerCountResponse]:
+        return PeerCountResponse
+
+
+class DisconnectPeerEvent(BaseEvent):
+    """
+    Event broadcasted when we want to disconnect from a peer
+    """
+
+    def __init__(self, peer: IdentifiablePeer, reason: DisconnectReason) -> None:
+        self.peer = peer
+        self.reason = reason
--- a/trinity/protocol/common/peer_pool_event_bus.py
+++ b/trinity/protocol/common/peer_pool_event_bus.py
@@ -1,0 +1,101 @@
+from typing import (
+    cast,
+    Generic,
+    TypeVar,
+)
+from cancel_token import (
+    CancelToken,
+)
+from lahja import (
+    Endpoint,
+)
+
+from p2p.exceptions import (
+    PeerConnectionLost,
+)
+from p2p.kademlia import (
+    from_uris,
+)
+from p2p.peer import (
+    BasePeer,
+    IdentifiablePeer,
+)
+from p2p.peer_pool import (
+    BasePeerPool,
+)
+from p2p.service import (
+    BaseService,
+)
+
+from .events import (
+    ConnectToNodeCommand,
+    DisconnectPeerEvent,
+    PeerCountRequest,
+    PeerCountResponse,
+)
+
+
+TPeer = TypeVar('TPeer', bound=BasePeer)
+
+
+class PeerPoolEventServer(BaseService, Generic[TPeer]):
+    """
+    Base request handler that listens for requests on the event bus that should be delegated to
+    the peer pool to either perform an action or return some response. Subclasses should extend
+    this class with custom event handlers.
+    """
+
+    def __init__(self,
+                 event_bus: Endpoint,
+                 peer_pool: BasePeerPool,
+                 token: CancelToken = None) -> None:
+        super().__init__(token)
+        self.peer_pool = peer_pool
+        self.event_bus = event_bus
+
+    async def accept_connect_commands(self) -> None:
+        async for command in self.wait_iter(self.event_bus.stream(ConnectToNodeCommand)):
+            self.logger.debug('Received request to connect to %s', command.node)
+            self.run_task(self.peer_pool.connect_to_nodes(from_uris([command.node])))
+
+    async def handle_peer_count_requests(self) -> None:
+        async for req in self.wait_iter(self.event_bus.stream(PeerCountRequest)):
+            self.event_bus.broadcast(
+                PeerCountResponse(len(self.peer_pool)),
+                req.broadcast_config()
+            )
+
+    async def handle_disconnect_peer_events(self) -> None:
+        async for ev in self.wait_iter(self.event_bus.stream(DisconnectPeerEvent)):
+            try:
+                peer = self.get_peer(ev.peer)
+            except PeerConnectionLost:
+                pass
+            else:
+                peer.disconnect(ev.reason)
+
+    async def _run(self) -> None:
+        self.logger.debug("Running PeerPoolEventServer")
+
+        self.run_daemon_task(self.handle_peer_count_requests())
+        self.run_daemon_task(self.accept_connect_commands())
+        self.run_daemon_task(self.handle_disconnect_peer_events())
+
+        await self.cancel_token.wait()
+
+    def get_peer(self, dto_peer: IdentifiablePeer) -> TPeer:
+
+        try:
+            peer = self.peer_pool.connected_nodes[dto_peer.uri]
+        except KeyError:
+            self.logger.debug("Peer %s does not exist in the pool anymore", dto_peer.uri)
+            raise PeerConnectionLost()
+        else:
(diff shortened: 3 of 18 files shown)


trinity/server.py
```diff
@@ -121,6 +127,15 @@ def __init__(self,
     def _make_peer_pool(self) -> TPeerPool:
         pass
 
+    def _make_peer_pool_request_handler(
+            self,
+            peer_pool: TPeerPool) -> BasePeerPoolEventBusRequestHandler[Any]:
```
cburgdorf: I used `Any` here because:

- we don't really care about the specific type of `BasePeerPoolEventBusRequestHandler`, we just want to run it
- I wanted to avoid adding yet another generic to `BaseServer` especially considering that this whole code will soon be refactored away anway.

cburgdorf:
@pipermerriam this is almost like #251 that you had reviewed a while but it already has the suggestion applied that you made back then. So maybe you'd like to review this? I'm trying to slice out chunks of my isolated process sync stuff...

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -0,0 +1,97 @@
...
+    async def _run(self) -> None:
+        self.logger.info("Running BasePeerPoolEventBusRequestHandler")
+
+        self.run_daemon_task(self.handle_peer_count_requests())
+        self.run_daemon_task(self.accept_connect_commands())
+        self.run_daemon_task(self.handle_disconnect_peer_events())
+
+        await self.cancel_token.wait()
+
+    def maybe_return_peer(self, dto_peer: IdentifiablePeer) -> Optional[TPeer]:
```
cburgdorf: We need to deal with the case that a peer we want to run some action on isn't available anymore. Alternatively, we could maybe also raise `PeerConnectionLost` here?

p2p/peer.py
```diff
@@ -165,6 +166,32 @@ class BasePeerContext:
...
+
+class BaseDTOPeer:
+    """
+    A peer solely meant as a Data Transfer Object (DTO) to travel across process boundaries.
+    It's a shallow, pickleable representation of a peer that carries enough information to
+    make the peer identifiable within the actual pool of real peers.
+    """
+
+    def __init__(self, uri: str):
+        self.uri = uri
```
pipermerriam: What do you think about going ahead and tackling whatever the underlying issue is with our `Node` class not being pickleable.  Seems like something that should be addressable and would simplify, allowing us to dump this DTO object.

trinity/protocol/common/peer_pool_event_bus.py
pipermerriam: Returning `None` here means return value checking.  While the method is named in a way to try to make that clear, I think it'd be better to just call this `get_peer()` and raise an appropriate exception if the peer is no longer connected (like `PeerConnectionLost`) and force call sites to handle this.

cburgdorf: Sounds good. In fact, down the road (not in this PR), raising a `PeerConnectionLost` exception actually makes things even a bit easier. :+1:

p2p/peer.py
cburgdorf: This actually opens up a discussion I wanted to open up in a future PR. For the sync, we send over a special DTO representations of the Peer which can be anything as long as it contains a `url` property. So, these objects contain things like the `head_td` etc. So, it would be more than just the `Node`.

But this is one of the weak points of the current wip design because this leads towards a design where the proxy peer carries properties such `head_td` but which aren't *live* but rather a snapshot of whatever the `head_td` was at the time that the peer traveled across the process boundaries.

I *think* having these sync properties on the proxy peer is bad because from a developer perspective it can be a big surprise that these values do not necessarily reflect the actual live value of that property of that peer in the actual `PeerPool`. Currently, I update these values in the proxy peer pool whenever a peer travels across the process boundaries so it *does update* but not on a guaranteed schedule.

On the other hand, it feels wasteful *not* to include these data points (that the sync or potentially other consumers need) whenever the peer travels across process domains. So, my current thinking is that maybe on the proxy peer there should be an API such as `async def get_peer_data(self) -> TPeerData` where one can request the data explicitly but I'm still unsure if and how it should be communicated initially.

Also, looking a bit more into where the updates for this data would come from it seems updates come from an `Announce` command so the `ProxyPeerPool` could track these itself :thinking: 

Does what I write make any sense? We could also ignore this part for now and pick this up when I send the actual PRs that tackle the syncing bits.

cburgdorf:
>Someone else should probably look this over. I did a quick pass and things looks right but it'd be good to loop others in on this architecture since it's going to expand pretty quickly and become very core to lots of stuff.

@carver mind giving this a second look?

p2p/peer.py
carver: > Also, looking a bit more into where the updates for this data would come from it seems updates come from an `Announce` command so the `ProxyPeerPool` could track these itself

At first glance, this makes the most sense to me.

pipermerriam: I'm not sure I agree with having the peer pool track those.  Doesn't feel natural to require the peer pool to sit in between code that wants to know about peer metadata and the peer itself.

It feels more appropriate to continue in this direction and have the peer be responsible for sending this information over the event bus.  Code that wanted to interact with a peer would do something like the following.

1. Ask peer pool over the event bus for information about what peers exist.
2. Pick a peer and then as it over the event bus for it's metadata (head_td, etc)
3. .... profit?

This does have the inherent inefficiency of requiring more event bus traffic to pick an appropriate peer but this seems solvable at a second layer, and, once we the process has an event bus based connection to a peer it should be able to *listen* for updates to the peer information without needing to re-request them, assuming we build an API for that which seems somewhat akin to our existing `PeerSubscriber` API but event bus flavored.

Did I mis-understand something or does this make sense.

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -0,0 +1,101 @@
...
+        self.run_daemon_task(self.handle_disconnect_peer_events())
+
+        await self.cancel_token.wait()
+
+    def get_peer(self, dto_peer: IdentifiablePeer) -> TPeer:
+
+        try:
+            peer = self.peer_pool.connected_nodes[dto_peer.uri]
+        except KeyError:
+            self.logger.warning("Peer %s does not exist in the pool anymore", dto_peer.uri)
```
carver: Same here, I think this should be debug.

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -0,0 +1,101 @@
...
+        async for ev in self.wait_iter(self.event_bus.stream(DisconnectPeerEvent)):
+            try:
+                peer = self.get_peer(ev.peer)
+            except PeerConnectionLost:
+                pass
+            else:
+                peer.disconnect(ev.reason)
+
+    async def _run(self) -> None:
+        self.logger.info("Running BasePeerPoolEventBusRequestHandler")
```
carver: Debug is probably fine for this

p2p/peer.py
cburgdorf: I'm thinking back and forth about this. In the actual `PeerPool` process, these things are currently tracked directly by the `ETHPeer` / `LESPeer` classes inside the `handle_sub_proto_msg` and the `process_sub_proto_handshake` method calls and while tracking these updates from the `ProxyPeerPool` is possible it would either mean to duplicate some of the handling or have `ETHPeer` / `LESPeer` send out events via the event bus whenever these properties change (Currently, these peer classes do not have access to the event bus).

What if:

Instead of having things like `head_td` sit directly on the peer classes, peer classes would have *extra data*  sitting one level deeper at a special property `meta_data` (name open for suggestions`.

Maybe, even have this property be of `async` nature (or use an `async` method getter / setter approach if async properties feel weird ([according to SO they do](https://stackoverflow.com/questions/54984337/how-should-you-create-properties-when-using-asyncio)).

Then, an `ETHProxyPeer` would have the same `meta_data` property but what happens here behind the scene is that the `ETHProxyPeer` (or really the underlying `BaseProxyPeer` would send out an event bus request (similar to how @pipermerriam suggests) to fetch this information from the actual peer in the peer pool process.

This has the nice property that we could even test drive this handling with the *current* sync and hence move the implementation forward in small chunks to reduce the code impact of my follow up changes.

The one scary thing about doing something like `if head_td > await peer.extra_data.head_td`is that it might lead to code that looks simple but causes multiple requests to the actual peer pool process to fetch meta data. Not sure if this could be solved with caching or if just method access alone (e.g  `meta_data = await peer.get_meta_data()` would encourage batch access of the data enough).

Thoughts?

trinity/protocol/common/events.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Request event to get the count of peers connected to the pool.
+    """
+
+    @staticmethod
+    def expected_response_type() -> Type[PeerCountResponse]:
+        return PeerCountResponse
+
+
+class DisconnectPeerEvent(BaseEvent):
```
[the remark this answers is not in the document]
cburgdorf: The idea was to keep `p2p` free from things that do not *need* to be there. And these events are used by trinity to hook the current peer pool up with trinity's event bus to give other parts of trinity access to peer pool information. From the current peer pools perspective these events are just not *intrinsically* needed by the peer pool and are an implementation detail of how Trinity wants to expose some of the peer pool APIs.

On the other hand though. The current peer pool already depends on discovery that it can only access via event bus APIs. I believe in the future we might redesign some of the current peer pool internals which might change the picture but from today's perspective, these events felt like something that the peer pool just doesn't need and hence should not know about.

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -0,0 +1,101 @@
...
+    DisconnectPeerEvent,
+    PeerCountRequest,
+    PeerCountResponse,
+)
+
+
+TPeer = TypeVar('TPeer', bound=BasePeer)
+
+
+class BasePeerPoolEventBusRequestHandler(BaseService, Generic[TPeer]):
```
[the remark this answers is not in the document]
cburgdorf: Hahaha, you should have seen the interim names I came up with. Gosh, I had candidates that were way longer. I hate myself for that and there's something about it that makes me feel I must have worked at Microsoft in a previous life. I just can't come up with concise names.

`PeerPoolEventServer` sounds way better to me :+1:

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -0,0 +1,101 @@
...
+    def get_peer(self, dto_peer: IdentifiablePeer) -> TPeer:
+
+        try:
+            peer = self.peer_pool.connected_nodes[dto_peer.uri]
+        except KeyError:
+            self.logger.warning("Peer %s does not exist in the pool anymore", dto_peer.uri)
+            raise PeerConnectionLost()
+        else:
+            if not peer.is_operational:
+                self.logger.warning("Peer %s is not operational", peer)
```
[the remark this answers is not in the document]
cburgdorf: Totally. Strictly speaking, this is a leftover from debugging because I tend to use `warning` because they come up as yellow and bold :)

trinity/protocol/common/peer_pool_event_bus.py
cburgdorf: :+1:

trinity/protocol/common/peer_pool_event_bus.py
cburgdorf: :+1:

cburgdorf:
I'm going to merge this as the impact of these changes is rather small but already moves my sync work one step closer to `master`. We can kick the *peer metadata* issue down the road but I have some ideas to try out based on your input.

p2p/peer.py
pipermerriam: I've been resistant to `async` properties because they feel like foot-guns where as method calls carry some sort of implicit concept of *this has a cost so cache the result if you are going to need it again soon*.

My *gut* says that in general we're going to want to do the following, cutting some corners along the way where it makes sense to validate the concepts but ultimately headed in this main direction.

1. Straight forward way to define events with minimal boilerplate (potentially with a metaclass)
2. Easy way to *serve* the events for a given object that exposes APIs over the bus (`PeerPool`, `Peer`, etc)
3. Easy way to *proxy* object types that serve events so that we can interact with them in a semi-transparent way **but** also have the ability to easily dig down to the low level components when needed and explicitly make requests.

I know this isn't exactly an answer to your question. I think my answer is 1) do what you think is best and 2) lets get something written up that tries to capture the python API for the end goal for this API because I think we're close enough to understanding it to start defining it that way.

PR #469, opened 2019-03-29
Update previous epoch calculation

### LINKED ISSUE #382: Make `get_previous_epoch` return `current_epoch - 1` in all cases

ralexstokes:
### What was wrong?

Fixes #382.

Replaces #398.

### How was it fixed?

Started rebasing #398 and realized we want to handle some of the tests differently.

This has the same content but keeps some of the tests... after further inspection I think what I had in #398 was actually weakening the test coverage.

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/epoch_processing_helpers.py
+++ b/eth2/beacon/epoch_processing_helpers.py
@@ -56,12 +56,11 @@ if TYPE_CHECKING:
 def get_previous_epoch_boundary_attestations(
         state: 'BeaconState',
         slots_per_epoch: int,
-        genesis_epoch: Epoch,
         latest_block_roots_length: int) -> Iterable[PendingAttestationRecord]:
     beacon_block_root = get_block_root(
         state,
         get_epoch_start_slot(
-            state.previous_epoch(slots_per_epoch, genesis_epoch),
+            state.previous_epoch(slots_per_epoch),
             slots_per_epoch,
         ),
         latest_block_roots_length,
@@ -75,7 +74,6 @@ def get_previous_epoch_boundary_attestations(
 def get_previous_epoch_matching_head_attestations(
         state: 'BeaconState',
         slots_per_epoch: int,
-        genesis_epoch: Epoch,
         slots_per_historical_root: int) -> Iterable[PendingAttestationRecord]:
     for attestation in state.previous_epoch_attestations:
         beacon_block_root = get_block_root(
--- a/eth2/beacon/epoch_processing_helpers.py
+++ b/eth2/beacon/epoch_processing_helpers.py
@@ -13,7 +13,6 @@ from eth_typing import (
 
 from eth_utils import (
     to_tuple,
-    to_set,
 )
 
 from eth.constants import ZERO_HASH32
@@ -57,6 +56,9 @@ def get_previous_epoch_boundary_attestations(
         state: 'BeaconState',
         slots_per_epoch: int,
         latest_block_roots_length: int) -> Iterable[PendingAttestationRecord]:
+    if not state.previous_epoch_attestations:
+        return tuple()
+
     beacon_block_root = get_block_root(
         state,
         get_epoch_start_slot(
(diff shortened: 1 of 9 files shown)


ralexstokes:
@ChihChengLiang @hwwhww @NIC619 given that you reviewed the other PR, i'll tag you here.

like stated above, i started rebasing and realized we want to handle some of the testing in a better way.

the one thing i could call special attention to is `test_justification_without_mock`. it seems like it should yield a bitfield `0b0` _unless_ we provide attestations. and it doesn't seem like we will get `0b11` until we run through two epoch processings. so i think that the prior implementation was incorrect and has now been fixed (although is not testing the scenario where we do in fact hit `0b11`) and that there was a bug in the previous implementation of `_current_previous_epochs_justifiable`. if i'm wrong, then there is a bug in the new `_is_epoch_justifiable` and will happily track it down so let me know :)

thanks!

ChihChengLiang:
@ralexstokes Feel free to fix them, those functions are implemented before v0.2.0. Prior to this PR, previous epoch and current epoch is 0 on the genesis, that's the very reason why `test_justification_without_mock` is getting 0b11. `_current_previous_epochs_justifiable` is also an awkward move to address the behavior of previous epoch and current epoch near genesis. As the consistent behavior introduced, it is safe to nuke `_current_previous_epochs_justifiable` and replace it with `_is_epoch_justifiable`.

ralexstokes:
@ChihChengLiang great -- went ahead and did that! just to confirm, we do want this w/ the latest implementation correct? https://github.com/ethereum/trinity/pull/469/files#diff-9ffda674671b43bc2ac2193ad772c6a5R319

this is my understanding but we want to be extra careful about justifying the correct stuff :)

eth2/beacon/epoch_processing_helpers.py
```diff
@@ -145,69 +145,55 @@ def get_attestations_for(root: Hash32) -> Sequence[PendingAttestationRecord]:
...
-        previous_epoch_boundary_attester_indices,
-        config.MAX_DEPOSIT_AMOUNT,
+@to_tuple
+def _get_epoch_boundary_attesting_indices(state: 'BeaconState',
+                                          attestations: Sequence[PendingAttestationRecord],
+                                          epoch: Epoch,
+                                          config: Eth2Config) -> Iterable[ValidatorIndex]:
+    relevant_attestations = (
+        a for a in attestations
+        if a.data.epoch_boundary_root == get_block_root(
```
ChihChengLiang: This `get_block_root` is called with same inputs many times in the loop

eth2/beacon/epoch_processing_helpers.py
```diff
@@ -145,69 +145,55 @@ def get_attestations_for(root: Hash32) -> Sequence[PendingAttestationRecord]:
...
-        state.previous_justified_epoch,
-        previous_epoch_boundary_root,
-        CommitteeConfig(config),
-    )
 
-    previous_epoch_boundary_attesting_balance = get_total_balance(
-        state.validator_balances,
-        previous_epoch_boundary_attester_indices,
-        config.MAX_DEPOSIT_AMOUNT,
+@to_tuple
```
ChihChengLiang: No need for to_tuple if we are returning tuple already

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -128,47 +129,31 @@ def process_eth1_data_votes(state: BeaconState, config: Eth2Config) -> BeaconSta
...
-    previous_epoch_active_validator_indices = get_active_validator_indices(
+    active_validator_indices = get_active_validator_indices(
         state.validator_registry,
-        previous_epoch,
-    )
-    current_total_balance = get_total_balance(
-        state.validator_balances,
-        current_epoch_active_validator_indices,
-        config.MAX_DEPOSIT_AMOUNT,
+        epoch
```
ChihChengLiang:
```
epoch,
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -209,15 +194,20 @@ def _get_finalized_epoch(
...
     current_epoch = state.current_epoch(config.SLOTS_PER_EPOCH)
-    previous_epoch = state.previous_epoch(config.SLOTS_PER_EPOCH, config.GENESIS_EPOCH)
+    previous_epoch = state.previous_epoch(config.SLOTS_PER_EPOCH)
 
-    current_epoch_justifiable, previous_epoch_justifiable = _current_previous_epochs_justifiable(
+    current_epoch_justifiable = _is_epoch_justifiable(
         state,
+        state.current_epoch_attestations,
         current_epoch,
+        config
```
ChihChengLiang:
```
config,
```

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -209,15 +194,20 @@ def _get_finalized_epoch(
...
+    )
+    previous_epoch_justifiable = _is_epoch_justifiable(
+        state,
+        state.previous_epoch_attestations,
         previous_epoch,
-        config,
+        config
```
ChihChengLiang:
```
config,
```
FYI https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#dangling-commas

tests/eth2/beacon/state_machines/forks/test_serenity_epoch_processing.py
```diff
@@ -217,7 +217,7 @@ def test_only_process_eth1_data_votes_per_period(sample_beacon_state_params, con
         ),
     )
 )
-def test_current_previous_epochs_justifiable(
+def test_is_epoch_justifiable(
```
ChihChengLiang: We can get rid of `current_epoch` and `previous_epoch` in this test actually. This test was intended to test `_current_previous_epochs_justifiable` but now `_is_epoch_justifiable`, and only the former needs them both.

ChihChengLiang:
Great work! Only leave some nitpick here. Justification and finalization look good to me, their core logic did not change much in the spec and the test cases are still valid.

eth2/beacon/epoch_processing_helpers.py
ralexstokes: good catch

eth2/beacon/epoch_processing_helpers.py
ralexstokes: another good catch

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
ralexstokes: yep, totally

tests/eth2/beacon/state_machines/forks/test_serenity_epoch_processing.py
ralexstokes: 👍

Issue #466, opened 2019-03-28
Fix circular dependency when embedding block roots in state by using `signed_root`


hwwhww:
### What is wrong?
Spec PR: https://github.com/ethereum/eth2.0-specs/issues/797
Spec issue: https://github.com/ethereum/eth2.0-specs/pull/816

### How can it be fixed

Fill this in if you know how to fix it.

ralexstokes:
i'm tackling this as part of addressing the work in #397

ralexstokes:
fixed in #470.

Issue #410, opened 2019-03-16
Fix `cache_state` and `advance_slot`


hwwhww:
### What is wrong?

Spec PR: https://github.com/ethereum/eth2.0-specs/pull/741/
Related to #380 and #391

### How can it be fixed

Fill this in if you know how to fix it.

ralexstokes:
i also did this but it is tied up w/ #470.

ralexstokes:
i also did this but it is tied up w/ #470.

coreyandersonbuilders:
Fraud hi- jacked account

ralexstokes:
fixed in #470. closing now.

PR #470, opened 2019-03-30
Headerification

### LINKED ISSUE #380: Parent root reconstruction + granular state roots + body/header segregation

ralexstokes:
### What was wrong?

Fixes #380, #466, #410.

Part of breaking up #397.

### How was it fixed?

Adds the `BeaconBlockHeader` the change required to get it working.

Updates some bookkeeping during the state transition to get the right header data in the right place, at the right time.

#### Cute Animal Picture


ralexstokes:
ok -- this should *finally* be ready to go! 

it turns out the one commit that changes the order of the epoch processing introduced a bug in the mock attestation utility that i spent some time debugging thinking it was related to the header change. it in fact is not so that has been removed from this PR and i updated the first comment here to reflect that.

this PR removes the `Proposal` type and adds the `BlockHeader` type, along with all the changes that entails.

i'm happy to go thru the existing issues once we merge this and close the ones we took care of as i know this was a bit of a moving target across a couple different issues.

eth2/_utils/ssz.py
```diff
@@ -34,7 +34,7 @@ def diff_ssz_object(left: BaseBeaconBlock,
                         sub_left_value,
                         sub_right_value,
                     )
-            elif isinstance(field_type, (ssz.sedes.List, ssz.sedes.CountableList)):
+            elif isinstance(field_type, ssz.sedes.List):
```
hwwhww: @jannikluhn could you take a look if it's correct?

eth2/beacon/helpers.py
```diff
@@ -37,6 +50,33 @@
...
+
+def get_empty_block(genesis_slot: Slot,
+                    block_class: Type[BaseBeaconBlock]) -> BaseBeaconBlock:
+    return block_class(
+        slot=genesis_slot,
+        previous_block_root=ZERO_HASH32,
+        state_root=ZERO_HASH32,
+        body=BeaconBlockBody.create_empty_body(),
+        signature=EMPTY_SIGNATURE,
+    )
```
hwwhww: What do you think about making it to a classmethod of `BeaconBlock`?

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -82,22 +81,27 @@ def validate_block_slot(state: BeaconState,
         )
 
 
+def validate_block_previous_root(state: BeaconState,
+                                 block: BaseBeaconBlock) -> None:
+    expected_root = state.latest_block_header.root
```
hwwhww: It should be `expected_root = state.latest_block_header.signed_root`
 with https://github.com/ethereum/trinity/issues/410

tests/eth2/beacon/chains/test_chain.py
```diff
@@ -47,7 +47,7 @@ def test_canonical_chain(valid_chain, genesis_slot):
 
     block = genesis_block.copy(
         slot=genesis_block.slot + 1,
-        parent_root=genesis_block.root,
+        previous_block_root=genesis_block.root,
```
hwwhww: Logically I think it's better to set to:
```suggestion
        previous_block_root=genesis_block.signed_root,
```
Although they are the same since no genesis block's signature is `EMPTY_SIGNATURE`.

tests/eth2/beacon/db/test_beacon_chaindb.py
```diff
@@ -95,7 +95,7 @@ def test_chaindb_get_score(chaindb, sample_beacon_block_params):
     assert chaindb.get_score(genesis.root) == 0
 
     block1 = BeaconBlock(**sample_beacon_block_params).copy(
-        parent_root=genesis.root,
+        previous_block_root=genesis.root,
```
hwwhww:
```suggestion
        previous_block_root=genesis.signed_root,
```

tests/eth2/beacon/db/test_beacon_chaindb.py
```diff
@@ -143,15 +143,15 @@ def test_chaindb_get_canonical_head(chaindb, block):
 
     block_2 = block.copy(
         slot=block.slot + 1,
-        parent_root=block.root,
+        previous_block_root=block.root,
```
hwwhww:
```suggestion
        previous_block_root=block.signed_root,
```

tests/eth2/beacon/db/test_beacon_chaindb.py
```diff
@@ -143,15 +143,15 @@ def test_chaindb_get_canonical_head(chaindb, block):
...
     )
     chaindb.persist_block(block_2, block_2.__class__)
     result_block = chaindb.get_canonical_head(block.__class__)
     assert result_block == block_2
 
     block_3 = block.copy(
         slot=block_2.slot + 1,
-        parent_root=block_2.root,
+        previous_block_root=block_2.root,
```
hwwhww:
```suggestion
        previous_block_root=block_2.signed_root,
```

tests/eth2/beacon/helpers.py
```diff
@@ -31,13 +31,13 @@ def mock_validator_record(pubkey,
...
-    Get a pseudo chain, only slot and parent_root are valid.
+    Get a pseudo chain, only slot and previous_block_root are valid.
     """
     block = genesis_block.copy()
     yield block
     for slot in range(1, length * 3):
         block = genesis_block.copy(
             slot=slot,
-            parent_root=block.root
+            previous_block_root=block.root
```
hwwhww:
```suggestion
            previous_block_root=block.signed_root,
```

trinity/protocol/bcc/servers.py
```diff
@@ -126,7 +126,7 @@ def __init__(self,
                 # TODO: pass accurate `block_class: Type[BaseBeaconBlock]` under
                 # per BeaconStateMachine fork
                 block = await self.db.coro_get_canonical_block_by_slot(slot, BeaconBlock)
-                if block.parent_root == parent.hash:
+                if block.previous_block_root == parent.root:
```
hwwhww:
```suggestion
                if block.previous_block_root == parent.signed_root:
```

tests/core/p2p-proto/bcc/helpers.py
```diff
@@ -69,16 +68,14 @@ def __init__(self, db: BaseAtomicDB) -> None:
...
         "state_root": ZERO_HASH32,  # note: not the actual genesis state root
-        "randao_reveal": EMPTY_SIGNATURE,
-        "eth1_data": Eth1Data.create_empty_data(),
         "signature": EMPTY_SIGNATURE,
         "body": BeaconBlockBody.create_empty_body()
     }
 
     if parent is not None:
-        kwargs["parent_root"] = parent.root
+        kwargs["previous_block_root"] = parent.root
```
hwwhww:
```suggestion
        kwargs["previous_block_root"] = parent.signed_root
```

trinity/protocol/bcc/validators.py
```diff
@@ -78,7 +78,7 @@ def _validate_sequence(self, blocks: Tuple[BaseBeaconBlock, ...]) -> None:
         for parent, child in sliding_window(2, blocks):
             # check that the received blocks form a sequence of descendents connected by parent
             # hashes, starting with the oldest ancestor
-            if child.parent_root != parent.hash:
+            if child.previous_block_root != parent.root:
```
hwwhww:
```suggestion
            if child.previous_block_root != parent.signed_root:
```

eth2/beacon/types/blocks.py
```diff
@@ -229,23 +315,19 @@ def from_parent(cls,
 
         return cls(
             slot=slot,
-            parent_root=parent_block.root,
+            previous_block_root=parent_block.root,
```
hwwhww: As for #466

```suggestion
            previous_block_root=parent_block.signed_root,
```

eth2/beacon/types/blocks.py
```diff
@@ -220,7 +306,7 @@ def from_parent(cls,
                     block_params: FromBlockParams) -> 'BaseBeaconBlock':
         """
         Initialize a new block with the `parent` block as the block's
-        parent hash.
+        parent root.
```
hwwhww: Maybe use `signed_root` would be more clear here

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -82,22 +81,27 @@ def validate_block_slot(state: BeaconState,
...
+    previous_root = block.previous_block_root
+    if previous_root != expected_root:
+        raise ValidationError(
+            f"block.previous_block_root ({previous_root}) is not equal to "
+            f"state.latest_block_header.root ({expected_root})"
```
hwwhww:
```suggestion
            f"state.latest_block_header.signed_root ({expected_root})"
```

tests/eth2/beacon/test_helpers.py
```diff
@@ -54,7 +68,7 @@ def generate_mock_latest_historical_roots(
     chain_length = (current_slot // slots_per_epoch + 1) * slots_per_epoch
     blocks = get_pseudo_chain(chain_length, genesis_block)
     latest_block_roots = [
-        block.hash
+        block.root
```
hwwhww: For https://github.com/ethereum/trinity/issues/466
```suggestion
        block.signed_root
```

eth2/_utils/ssz.py
jannikluhn: yes, it's correct.

ralexstokes:
@hwwhww thanks for the feedback! i'm w/ you about the `signed_root` thing, i just took another approach to the problem.

refer this module doc here: https://github.com/ethereum/trinity/pull/470/files#diff-5ee31ee2aba7f5d75fae31984b6492f1R1

the idea is that in all of these cases (in addition to how we key blocks in the DB like you mentioned) we want to use the `signed_root`. before this PR, there are various uses of the `hash` or `root` with the `root` referring to the `hash`.

my goal is to put the mechanism of how we get a block's (or in general an SSZ object's) identifier behind an abstraction barrier (so that we can localize any future changes to just one spot in the codebase, at most) which we could imagine is some property on the Python object called `identifier`. i chatted a bit w/ some of the other snake charmers about this and we were discussing either `id` or `identifier` and wanted to go w/ the more explicit `identifier` (and avoid collision on the `id` builtin).

so if we go down this road, we need to replace all uses of `root`/`hash` with `identifier` and as i started to do this i realized we could save some time by just using `root` generically. if you need a specific SSZ calculation you can call that on the Python object, e.g. `obj.hash_tree_root` or `obj.signed_root` and then to get the canonical identifier for this object, you can just call `root`.

i tried to communicate this in the module comment but agree it is not very precise.  if you find it is confusing or want to avoid this communication overhead in the future (as any future maintainers will have to work through the same understanding!) then i think we should avoid overloading the word `root`. in that case i would propose we move to the `identifier` property and am happy to update this PR

would love your thoughts on this :) (cc @ChihChengLiang @NIC619 @jannikluhn)

eth2/beacon/helpers.py
ralexstokes: i like it!

jannikluhn:
We should just add the `root` property to `ssz.Serializable` in `py-ssz`. We might not need it for all SSZ objects, but conceptually it makes sense for all of them.

`signed_root` is a little more tricky as only some objects are signed. I still think we should implement it in `py-ssz` though, either via an additional base class `ssz.SignedSerializable` or a flag  `is_signed = True`.

What would be the reason for renaming `root` to `identifier`? I'd prefer `root` because it's closer to the spec and presumably other implementations. Unless we're trying to have the same property in Eth1 objects, but then I'm not sure if falling back to `hash` wouldn't be better.

coreyandersonbuilders:
Fraud

coreyandersonbuilders:
Hi-jacked account

coreyandersonbuilders:
Hi-jacked account

coreyandersonbuilders:
Thank you so very much it’s appreciated and I still need help closing this hi-jacked account out and verify the person who stated it

Corey Anderson @ Anderson Builders LLC

ralexstokes:
@hwwhww went w/ the explicit approach we talked about earlier :) going to go ahead and merge this in

Issue #471, opened 2019-03-30
The chaotic configs


ChihChengLiang:
### What is wrong?

In Trinity and Beacon chain, we have a lot of configs. Some renaming and clarification of responsibility could make life easier. This is just a quick brain dump.

- class TrinityConfig
- class BeaconAppConfig(BaseAppConfig)
- BeaconChainConfig
- SERENITY_CONFIG = BeaconConfig(...)

### How can it be fixed

- Rename BeaconConfig to Eth2Config
- network_id appeared in TrinityConfig and BeaconChainConfig, something must be wrong here

PR #473, opened 2019-03-30
Rename `BeaconConfig` to `Eth2Config` and move it to `eth2.configs`

hwwhww:
### What was wrong?
part of #471
The scope of these constants are not only for beacon chain but also shard chains. 

### How was it fixed?
1. Rename `BeaconConfig` to `Eth2Config`
2. Move `eth2/beacon/configs.py` → `eth2/configs.py`

#### Cute Animal Picture

🐨


tests/eth2/bls_stub.py
```diff
@@ -0,0 +1,10 @@
+def bls_verify(pubkey, message_hash, signature, domain):
+    return True
+
```
ChihChengLiang: is this file committed accidentally ?

hwwhww: whooops yes

PR #476, opened 2019-03-30
Create Testnet state machine

ChihChengLiang:
### What was wrong?

In the cow people programming branch `genesis_time`, we are accessing `eth2_config` everywhere in the plugin. This should be avoided because `eth2_config` is defined fork-by-fork.

The correct way to access `eth2_config` is to get it from the state machine that we intend to use on a particular slot. If a customize `eth2_config` is required for a testnet, then we should create a Testnet state machine class who has the customized config as property.

During this PR I also realize that some genesis information, such as genesis_time and genesis_slot, could be organized into a `BeaconGenesisData`, while some private validator information, such as their private and public key that's specific to the testnet use, could be organized into `testing_config.py`.

### How was it fixed?

- mkdir  `eth2/beacon/state_machines/forks/testnet/`
  - create `TestnetStateMachine` in `__init__.py`
  - move its `eth2_config` to `configs.py`
- remove the use of eth2_config everywhere
- aggregate genesis data into BeaconGenesisData
- aggregate all private validator information to `trinity/plugins/eth2/beacon/testing_config.py`

cc @hwwhww @mhchia @NIC619 

#### Cute Animal Picture

<a title="Keith Weller/USDA [Public domain], via Wikimedia Commons" href="https://commons.wikimedia.org/wiki/File:Cow_female_black_white.jpg">Cow female black white</a>


NIC619:
`BeaconGenesisData` sounds reasonable. However in a local testing or testnet, `genesis_time` is likely to change in each run while other genesis data, e.g., `genesis_slot` aren't. Would it be trivial to automate the process of updating `genesis_time` in each run of testing/testnet?

eth2/beacon/chains/testnet/__init__.py
```diff
@@ -0,0 +1,27 @@
+from typing import Tuple, Type   # noqa: F401
+from .constants import (
```
hwwhww: nitpick: move it to the last

eth2/beacon/chains/testnet/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    GENESIS_SLOT,
+    TESTNET_CHAIN_ID,
+)
+from eth2.beacon.state_machines.forks.xiao_long_bao import (
+    XiaoLongBaoStateMachine,
+)
+from eth2.beacon.chains.base import BeaconChain
```
hwwhww: nitpick: move it before `eth2.beacon.state_machines.forks.xiao_long_bao`

eth2/beacon/state_machines/forks/xiao_long_bao/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+    FromBlockParams,
+)
+
+from eth2.beacon.types.blocks import BaseBeaconBlock  # noqa: F401
+from eth2.beacon.types.states import BeaconState  # noqa: F401
+
+from eth2.beacon.state_machines.base import BeaconStateMachine
+from eth2.beacon.state_machines.state_transitions import BaseStateTransition  # noqa: F401
+
+from .configs import XIAO_LONG_BAO_CONFIG
```
hwwhww: nitpick: move it to the last

trinity/config.py
```diff
@@ -70,19 +71,18 @@
...
+from eth2.beacon.typing import (
+    Timestamp,
+    Slot,
+)
+from trinity.plugins.eth2.beacon.testing_config import (
+    Config as testing_config,
+    keymap,
 )
-from trinity.plugins.eth2.beacon.testing_blocks_generators import config as testing_config, keymap
+from eth2.beacon.chains.testnet import TestnetChain
```
hwwhww: nitpick: reorder the `import`

trinity/plugins/eth2/beacon/plugin.py
```diff
@@ -178,7 +174,6 @@ def new_slot(self, slot: int) -> None:
         """
         The callback for `SlotTicker`, to be called whenever new slot is ticked.
         """
-        assert slot > self.eth2_config.GENESIS_SLOT
         head = self.chain.get_canonical_head()
         state_machine = self.chain.get_state_machine(at_block=head)
```
hwwhww: If we have more than one fork, you're not building the new block on the new fork if you set `at_block=head`.

I *think* the correct logic is:
1. use `get_state_machine_class_for_block_slot` to get the `state_machine_class` of the *next* block.
2. generate a block from the parent with of the given `state_machine_class`.
3. use it to initiate a `StateMachine`.

So, could you try if it works?
```suggestion
        state_machine = self.chain.get_state_machine()
```
If `at_block` is `None`, it should do the above logic.

eth2/beacon/state_machines/forks/xiao_long_bao/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+from eth2.beacon.typing import (
+    FromBlockParams,
+)
+
+from .configs import (
+    XIAO_LONG_BAO_CONFIG,
+)
+
+
+class XiaoLongBaoStateMachine(BeaconStateMachine):
```
pipermerriam: TIL: https://en.wikipedia.org/wiki/Xiaolongbao

eth2/beacon/state_machines/forks/xiao_long_bao/configs.py
```diff
@@ -0,0 +1,15 @@
+from eth2.beacon.state_machines.forks.serenity.configs import (
+    SERENITY_CONFIG,
+)
+
+SLOTS_PER_EPOCH = 4
+
+
+XIAO_LONG_BAO_CONFIG = SERENITY_CONFIG._replace(
```
pipermerriam: Just noting that a private API shouldn't be called from outside of the class indicating that this API needs to be updated to not have an underscore prefix.

PR #483, opened 2019-04-05
Perf: Don't create intermediate tuples in normalizer

cburgdorf:
### What was wrong?

**THIS IS A WEIRD ONE:**

I *thought* I've found a potential performance issue within our normalizer code but after crunching some numbers, I'm actually not so sure anymore.

Our current code creates intermediate tuples and thereby loop over the data more often than needed. At least, that's how I understand it and we can in fact, prove that getting rid of the intermediate tuples runs faster.

Here's a simplified snipped demoing the performance difference.

With intermediate tuples:

```python
def multiply(x):
    return x * 10

def test():
    msg = tuple(x for x in range(5))

    uncles_hashes = tuple(map(
        multiply,
        tuple(x for x in msg)
    ))
    transaction_roots_and_trie_data = tuple(map(
        multiply,
        tuple(x for x in msg)
    ))

    body_bundles = tuple(zip(msg, transaction_roots_and_trie_data, uncles_hashes))
    return body_bundles

import timeit
print(timeit.timeit('test()', globals=globals()))
```

Prints: `5.469376810011454`

Without intermediate tuples:

```python
def multiply(x):
    return x * 10

def test():
    msg = tuple(x for x in range(5))

    uncles_hashes = map(
        multiply,
        (x for x in msg)
    )
    transaction_roots_and_trie_data = map(
        multiply,
        (x for x in msg)
    )

    body_bundles = tuple(zip(msg, transaction_roots_and_trie_data, uncles_hashes))
    return body_bundles

import timeit
print(timeit.timeit('test()', globals=globals()))
```

Prints: `4.732703522982774`

~~**However!** It turns out, that the more elements we have in `msg`, the number flip and the version with the intermediate tuples does actually run faster. E.g. if we change the code to be `msg = tuple(x for x in range(50))`, then, the version with the intermediate tuples runs consistently faster. This is surprising to me as I would have thought the opposite should be the case.~~ 

:middle_finger: Must have been something weird.

### How was it fixed?

- removed intermediate tuples
- turned one place into a loop to avoid traversing the same set twice

### TODO:

- [] add changelog entry for perf improvement

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/eth/normalizers.py
+++ b/trinity/protocol/eth/normalizers.py
@@ -1,9 +1,10 @@
 from typing import (
+    Iterable,
     Tuple,
 )
 
-from eth_utils.toolz import (
-    compose,
+from eth_utils import (
+    to_tuple,
 )
 from eth.db.trie import make_trie_root_and_nodes
 from eth_hash.auto import keccak
@@ -13,6 +14,7 @@ from trinity.protocol.common.normalizers import (
     BaseNormalizer,
 )
 from trinity.protocol.common.types import (
+    BlockBodyBundle,
     BlockBodyBundles,
     NodeDataBundles,
     ReceiptsBundles,
@@ -26,7 +28,7 @@ class GetNodeDataNormalizer(BaseNormalizer[Tuple[bytes, ...], NodeDataBundles]):
 
     @staticmethod
     def normalize_result(msg: Tuple[bytes, ...]) -> NodeDataBundles:
-        node_keys = tuple(map(keccak, msg))
+        node_keys = map(keccak, msg)
         result = tuple(zip(node_keys, msg))
         return result
 
@@ -36,7 +38,7 @@ class ReceiptsNormalizer(BaseNormalizer[ReceiptsByBlock, ReceiptsBundles]):
 
     @staticmethod
     def normalize_result(message: ReceiptsByBlock) -> ReceiptsBundles:
-        trie_roots_and_data = tuple(map(make_trie_root_and_nodes, message))
+        trie_roots_and_data = map(make_trie_root_and_nodes, message)
         return tuple(zip(message, trie_roots_and_data))
 
 
@@ -44,15 +46,9 @@ class GetBlockBodiesNormalizer(BaseNormalizer[Tuple[BlockBody, ...], BlockBodyBu
     is_normalization_slow = True
 
     @staticmethod
-    def normalize_result(msg: Tuple[BlockBody, ...]) -> BlockBodyBundles:
-        uncles_hashes = tuple(map(
-            compose(keccak, rlp.encode),
-            tuple(body.uncles for body in msg)
-        ))
-        transaction_roots_and_trie_data = tuple(map(
-            make_trie_root_and_nodes,
-            tuple(body.transactions for body in msg)
-        ))
-
-        body_bundles = tuple(zip(msg, transaction_roots_and_trie_data, uncles_hashes))
-        return body_bundles
+    @to_tuple
+    def normalize_result(msg: Tuple[BlockBody, ...]) -> Iterable[BlockBodyBundle]:
+        for body in msg:
+            uncle_hashes = keccak(rlp.encode(body.uncles))
+            transaction_root_and_nodes = make_trie_root_and_nodes(body.transactions)
+            yield body, transaction_root_and_nodes, uncle_hashes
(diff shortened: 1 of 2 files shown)


pipermerriam:
I wold suggest running those two with profiling enabled to try and see where the performance hit is happening because that is entirely un-intuitive to me.

veox:
Below is all conjecture, and might indeed benefit from profiling.

-----

From "performance flipping", I'd guess this has something to do with the fact that

``` python
(x for x in msg)
```

is a generator object (calls to its `next()` are made from `map()` and values (pointers?) returned one-by-one), whereas

``` python
tuple(x for x in msg)
```

is an instantiated tuple object (constructed in `tuple()` and pointer returned (once?) to `map()`).

-----

There's also the `tuple(map(...))` and plain `map(...)` difference, which is probably a different beast. In other words, there are four cases of tuplification to consider here, not two.

veox:
... Also, I don't see a "reversal". This might be a Python version peculiarity. Mine is CPython v3.7.3.

With [this glorious script](https://gist.github.com/veox/9350f2aaacc3634d62595ee1c9bdb714#file-pr-483-tuplification-py), I get:

``` sh
# nmsgelems = 5
test00: 3.281730979000713
test01: 3.5057063780004682
test10: 3.575567134001176
test11: 4.169835692000561
# nmsgelems = 50
test00: 14.890971224000168
test01: 15.964820151999447
test10: 15.694290583000111
test11: 17.228873800000656
# nmsgelems = 500
test00: 135.5852560960011
test01: 139.08642388600128
test10: 137.17986890300017
test11: 146.79928800699963
```

* `test00()` has no `tuple()` as described above;
* `test01()` has the inner `tuple(x for x in msg)`;
* `test10()` has the outer `tuple(map(...))`; and
* `test11()` has both occasions.

veox:
Using CPython v3.6.8:

``` sh
# nmsgelems = 5
test00: 5.746762596998451
test01: 6.3236174329995265
test10: 6.438446665999436
test11: 7.349682849000601
# nmsgelems = 50
test00: 24.74597435999931
test01: 25.555854600999737
test10: 25.6410359649999
test11: 27.19204468099997
# nmsgelems = 500
test00: 214.67359304300044
test01: 219.21980871099913
test10: 215.62727851199998
test11: 222.10079747300006
```

Still no "reversal".

But note how CPython 3.6 is less performant than CPython 3.7 in this particular task. ;)

trinity/protocol/eth/normalizers.py
```diff
@@ -45,14 +45,15 @@ class GetBlockBodiesNormalizer(BaseNormalizer[Tuple[BlockBody, ...], BlockBodyBu
...
-        ))
-        transaction_roots_and_trie_data = tuple(map(
+            (body.uncles for body in msg)
+        )
+        transaction_roots_and_trie_data = map(
             make_trie_root_and_nodes,
-            tuple(body.transactions for body in msg)
-        ))
+            (body.transactions for body in msg)
+        )
```
carver: There's also the fact that we're iterating through msg twice. Maybe this whole thing could be replaced by:

```py
@staticmethod
@to_tuple
def normalize_result(msg: Tuple[BlockBody, ...]) -> BlockBodyBundles:
  for body in msg:
    uncle_hashes = keccak(rlp.encode(body.uncles))
    transaction_root_and_nodes = make_trie_root_and_nodes(body.transactions)
    yield body, uncle_hashes, transaction_root_and_nodes
```

Just about the only downside I see is that we look up the `encode` attribute on `rlp` on every body iteration. If this is a real hotspot we could look it up once ahead of the loop, but it seems unlikely to be a notable win.

cburgdorf:
@veox Thanks for helping to measure this! It's weird. Meanwhile I have upgraded my OS (unrelated to this haha!) and now I can't reproduce the turning of these numbers either. It's consistently faster!

Also, back then I had opened up [a SO post about it and everyone kept saying they don't see the reversal](https://stackoverflow.com/questions/55535677/tuple-vs-generator-expression-why-does-performance-flip-the-longer-the-sequenc). So, it must have been me/my machine (can we blame the stupid machines, please?).

@carver Ha! Even better. I took your suggestion (with some minor corrections). Maybe you want to review it again because I had to introduce a new type `BlockBodyBundle` (singular) which is obviously very close to `BlockBodyBundles` so that might be a discussion point. I can't think of a better name though and I would still prefer it over the unaliased types.

Issue #488, opened 2019-04-07
Split and merge #465 to master


ChihChengLiang:
### What is wrong?

As #465 getting fat and the exploration is nearly maximized, it's time to split it to smaller PRs and merge them back to master.

### How can it be fixed

- [x] Update a state machine class and a chain class for the testnet. #489
- [x] Add Slot ticker #490
- [x] Add validator service #525
- [x] Add validator service and slot ticker to the beacon plugin. #594
- [x] Update sync and the receiving server. #580
- [x] Add integration tests for two `trinity-beacon` clients initiating and syncing #594
- [x] The network setup experience #516

PR #465, opened 2019-03-28
Not a plugin Validator

ChihChengLiang:
### How was it fixed?

[//]: # (For important changes, please add a new entry to the release notes file)
[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #489, opened 2019-04-07
Add testnet chain and Xiao Long Bao state machine

ChihChengLiang:
### What was wrong?

Part of #488

We need a state machine class and a chain class for the testnet. Unfortunately, for the state machine it requires a specific name. A choice with better taste than foo might be food, and the lowest hanging food outside of the coworking space is the Xiao Long Bao.

### How was it fixed?

- [x] Create a state machine fork XiaoLongBaoStateMachine
- [x] Create a chain class TestnetChain
- [x] Side quest: Fix coro_get_canonical_head_root
- [x] Protect them with some tests

#### Cute Animal Picture

<a title="Eason Lai from 上海 [CC BY-SA 2.0 (https://creativecommons.org/licenses/by-sa/2.0)], via Wikimedia Commons" href="https://commons.wikimedia.org/wiki/File:Xiao_Long_Bao_at_Nanxiang_Mantou_Dian_1.jpg">Xiao Long Bao at Nanxiang Mantou Dian 1</a>


eth2/beacon/state_machines/forks/xiao_long_bao/configs.py
```diff
@@ -0,0 +1,22 @@
...
+)
+from eth2.beacon.typing import (
+    Slot,
+)
+
+SLOTS_PER_EPOCH = 4
+GENESIS_SLOT = Slot(32)
+
+
+XIAO_LONG_BAO_CONFIG = SERENITY_CONFIG._replace(
```
ChihChengLiang: Continuing the discussion of https://github.com/ethereum/trinity/pull/476#pullrequestreview-223372029 here.
SERENITY_CONFIG is a Namedtuple from `typing` and is really hard to mutate it.
One possible way is to explicitly name all the config we want to copy from the `SERENITY_CONFIG`
```python
XIAO_LONG_BAO_CONFIG = Eth2Config(
    # the part we want to customize
    TARGET_COMMITTEE_SIZE=2,
    SHARD_COUNT=2,
    ...
    # the part we want to copy
    BASE_REWARD_QUOTIENT=SERENITY_CONFIG.BASE_REWARD_QUOTIENT
   ... (really lengthy here) ...
)
```


cc @pipermerriam

eth2/beacon/chains/base.py
```diff
@@ -60,7 +60,7 @@ class BaseBeaconChain(Configurable, ABC):
     """
     chaindb = None  # type: BaseBeaconChainDB
     chaindb_class = None  # type: Type[BaseBeaconChainDB]
-    sm_configuration = None  # type: Tuple[Tuple[int, Type[BaseBeaconStateMachine]], ...]
+    sm_configuration = None  # type: Tuple[Tuple[Slot, Type[BeaconStateMachine]], ...]
```
ChihChengLiang: Change the typing here to make mypy happy, because the inheritance is `BaseBeaconStateMachine` -> `BeaconStateMachine` -> `SerenityStateMachine`.
Not sure if it makes sense to you @hwwhww

eth2/beacon/state_machines/forks/xiao_long_bao/configs.py
pipermerriam: Named tuple's are really only great at being data storage objects.  I've found that as soon as you start adding methods you're probably better of with a custom object.

Including an API on this for creating a copy with some fields overridden seems like an easy to implement and useful API.

ChihChengLiang: Created #503 to address this

eth2/beacon/state_machines/forks/xiao_long_bao/configs.py
```diff
@@ -0,0 +1,22 @@
...
+)
+from eth2.beacon.state_machines.forks.serenity.configs import (
+    SERENITY_CONFIG,
+)
+from eth2.beacon.typing import (
+    Slot,
+)
+
+SLOTS_PER_EPOCH = 4
+GENESIS_SLOT = Slot(32)
```
NIC619: Should it be `2**32`?

eth2/beacon/state_machines/forks/xiao_long_bao/configs.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+XIAO_LONG_BAO_CONFIG = SERENITY_CONFIG._replace(
+    SLOTS_PER_EPOCH=SLOTS_PER_EPOCH,
+    GENESIS_SLOT=GENESIS_SLOT,
+    GENESIS_EPOCH=slot_to_epoch(GENESIS_SLOT, SLOTS_PER_EPOCH),
+    TARGET_COMMITTEE_SIZE=2,
+    SHARD_COUNT=2,
```
NIC619: With `SLOTS_PER_EPOCH` == 4, `TARGET_COMMITTEE_SIZE` == 2 and `SHARD_COUNT` == 2, does this mean we need at least 16 validators?

eth2/beacon/chains/base.py
ChihChengLiang: My bad, undo the change. Mypy was unhappy about `sm_configuration` in `BaseTestnetChain`.

eth2/beacon/state_machines/forks/xiao_long_bao/configs.py
ChihChengLiang: I'm trying to use a smaller genesis slot number here to read the change in slots easier.

eth2/beacon/chains/testnet/constants.py
```diff
@@ -0,0 +1,6 @@
+from eth2.beacon.typing import (
+    Slot,
+)
+
+GENESIS_SLOT = Slot(2**32)
```
ChihChengLiang: 🙈Just realized that there are GENESIS_SLOT in the chain constants and in the state machine. It seems wrong to  treat them as a constant in the state machine and should be parameterizable from the chain. But maybe access this constant from the Serenity config for the time being and address the issue after #896 to resolved.

eth2/beacon/state_machines/forks/xiao_long_bao/configs.py
ChihChengLiang: Will remove this value and use the 2**32 from the SERENITY_CONFIG

eth2/beacon/state_machines/forks/xiao_long_bao/configs.py
ChihChengLiang: Not sure if I understand, do we need to specify 16 validators somewhere?

NIC619: Ah just that in #465 we use 8 validators. Not a problem. Ignore me. 😅

eth2/beacon/chains/testnet/constants.py
NIC619: #896?

ChihChengLiang: Sorry https://github.com/ethereum/eth2.0-specs/pull/896

tests/eth2/beacon/chains/test_chains.py
```diff
@@ -0,0 +1,13 @@
...
+
+@pytest.mark.parametrize(
+    "chain_klass",
+    (
+        TestnetChain,
+    )
+)
+def test_chain_class_well_defined(chain_klass):
+    chain = chain_klass(None)
+    assert chain.sm_configuration is not ()
```
hwwhww: Should we check:
```suggestion
    assert chain.sm_configuration is not () and chain.sm_configuration is not None
```
?

tests/eth2/beacon/state_machines/forks/test_fork_classes.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+from eth2.beacon.state_machines.forks.xiao_long_bao import XiaoLongBaoStateMachine
+
+
+@pytest.mark.parametrize(
+    "sm_klass",
+    (
+        XiaoLongBaoStateMachine,
```
hwwhww: How about add `SerenityStateMachine`? Would it pass the test?

eth2/beacon/chains/testnet/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+)
+from eth2.beacon.chains.base import (
+    BeaconChain,
+)
+from eth2.beacon.state_machines.forks.xiao_long_bao import (
+    XiaoLongBaoStateMachine,
+)
+from .constants import (
+    TESTNET_CHAIN_ID,
+)
```
hwwhww: Move it after `from eth2.beacon.state_machines.forks.serenity.configs import ...`.

tests/eth2/beacon/state_machines/forks/test_fork_classes.py
ChihChengLiang: It passes the test.

PR #480, opened 2019-04-02
Run plugin tests in CI

gsmadi:
### What was wrong?

Our CI was not running plugin tests during builds. As we start ramping up the plugin refactor its important to constantly monitor the health of these tests.

### How was it fixed?

Simply added some jobs to execute plugin tests for Py 3.6 and 3.7.

Also included the CircleCI build badge in the README.

[//]: # (For important changes, please add a new entry to the release notes file)
[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #490, opened 2019-04-09
Fix #437, add slot ticker

### LINKED ISSUE #437: Slot ticker in the beacon-node
### What is wrong?

IIUC from the discussion in https://github.com/ethereum/eth2.0-specs/issues/497, a client should maintain a current slot view of the state, rather than updating slot only when a block is received.

So we need a slot ticker somewhere in the codebase that increases the slot number, performs `per_slot_transition`, and updates state every 6 seconds.

Not sure if it makes sense to let the slot ticker have its own service or plugin. It looks like Syncer, RequestServer, and PeerPool all have their own job and thus not the best place for the slot ticker.

#NOTE: probably related to #260

cc @cburgdorf @ralexstokes @hwwhww

ChihChengLiang:
### What was wrong?

#437 Add a slot ticker for the validator

### How was it fixed?

- [x] Put the slot ticker under beacon plugin
- [x] Add tests
- [x] Require #480 for tests to work

#### Cute Animal Picture

<a title="Christian Mehlführer, User:Chmehl [CC BY 2.5 (https://creativecommons.org/licenses/by/2.5)], via Wikimedia Commons" href="https://commons.wikimedia.org/wiki/File:MC_Drei-Finger-Faultier.jpg">MC Drei-Finger-Faultier</a>

### DIFF
--- a/tests/plugins/eth2/beacon/test_slot_ticker.py
+++ b/tests/plugins/eth2/beacon/test_slot_ticker.py
@@ -1,9 +1,11 @@
-import pytest
-from trinity.plugins.eth2.beacon.slot_ticker import (
-    SlotTicker,
-    NewSlotEvent,
-)
 import asyncio
+
+import pytest
+
+from trinity.plugins.eth2.beacon.slot_ticker import (
+    NewSlotEvent,
+    SlotTicker,
+)
 
 
 @pytest.mark.asyncio
--- a/tests/plugins/eth2/beacon/test_slot_ticker.py
+++ b/tests/plugins/eth2/beacon/test_slot_ticker.py
@@ -11,16 +11,9 @@ async def test_slot_ticker_ticking(event_bus, event_loop):
     slot_ticker = SlotTicker(
         genesis_slot=0,
         genesis_time=0,
-        chain=None,
+        seconds_per_slot=1,
         event_bus=event_bus,
     )
-
-    def get_seconds_per_slot():
-        return 1
-
-    # mock get_seconds_per_slot
-    slot_ticker.get_seconds_per_slot = get_seconds_per_slot
-
     asyncio.ensure_future(slot_ticker.run(), loop=event_loop)
     await slot_ticker.events.started.wait()
     try:
--- a/tests/plugins/eth2/beacon/test_slot_ticker.py
+++ b/tests/plugins/eth2/beacon/test_slot_ticker.py
@@ -1,0 +1,35 @@
+import pytest
+from trinity.plugins.eth2.beacon.slot_ticker import (
+    SlotTicker,
+    NewSlotEvent,
+)
+import asyncio
+
+
+@pytest.mark.asyncio
+async def test_slot_ticker_ticking(event_bus, event_loop):
+    slot_ticker = SlotTicker(
+        genesis_slot=0,
+        genesis_time=0,
+        chain=None,
+        event_bus=event_bus,
+    )
+
+    def get_seconds_per_slot():
+        return 1
+
+    # mock get_seconds_per_slot
+    slot_ticker.get_seconds_per_slot = get_seconds_per_slot
+
+    asyncio.ensure_future(slot_ticker.run(), loop=event_loop)
+    await slot_ticker.events.started.wait()
+    try:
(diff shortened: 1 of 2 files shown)


trinity/plugins/eth2/beacon/slot_ticker.py
```diff
@@ -0,0 +1,77 @@
...
+        self.latest_slot = Slot(0)
+        self.event_bus = event_bus
+
+    async def _run(self) -> None:
+        self.run_daemon_task(self._keep_ticking())
+        await self.cancellation()
+
+    def get_seconds_per_slot(self) -> Second:
+        state_machine = self.chain.get_state_machine()
+        return state_machine.config.SECONDS_PER_SLOT
```
pipermerriam: Thoughts on caching this return value?

ChihChengLiang: Think about this more, handling the possibility of the changing `SECONDS_PER_SLOT` at this stage might not be a good idea. Created #491 to delay this to the future when it really matters.

trinity/plugins/eth2/beacon/slot_ticker.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class NewSlotEvent(BaseEvent):
+    def __init__(self, slot: Slot, elapsed_time: Second):
+        self.slot = slot
+        self.elapsed_time = elapsed_time
+
+
+class SlotTicker(BaseService):
+
```
mhchia: nits: should the fields be declared with type ahead of the methods?

trinity/plugins/eth2/beacon/slot_ticker.py
```diff
@@ -0,0 +1,74 @@
...
+    async def _run(self) -> None:
+        self.run_daemon_task(self._keep_ticking())
+        await self.cancellation()
+
+    async def _keep_ticking(self) -> None:
+        while self.is_operational:
+            elapsed_time = Second(int(time.time()) - self.genesis_time)
+            if elapsed_time >= self.seconds_per_slot:
+                slot = Slot(elapsed_time // self.seconds_per_slot + self.genesis_slot)
+                if slot > self.latest_slot:
```
mhchia: nit: should `self.latest_slot` be `cast`ed to enforce them the same type?

tests/plugins/eth2/beacon/test_slot_ticker.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+from trinity.plugins.eth2.beacon.slot_ticker import (
+    SlotTicker,
+    NewSlotEvent,
+)
+import asyncio
```
mhchia: nit: probably need to sort the imports here?

tests/plugins/eth2/beacon/test_slot_ticker.py
```diff
@@ -0,0 +1,28 @@
...
+async def test_slot_ticker_ticking(event_bus, event_loop):
+    slot_ticker = SlotTicker(
+        genesis_slot=0,
+        genesis_time=0,
+        seconds_per_slot=1,
+        event_bus=event_bus,
+    )
+    asyncio.ensure_future(slot_ticker.run(), loop=event_loop)
+    await slot_ticker.events.started.wait()
+    try:
```
mhchia: Can also be changed to `with pytest.raises(asyncio.TimeoutError):`

mhchia: ahh please ignore this. I'm being dumb.

trinity/plugins/eth2/beacon/slot_ticker.py
ChihChengLiang: They are already the same `Slot` type in line 62? Since `slot` is cast to `Slot` at line 61.

tests/plugins/eth2/beacon/test_slot_ticker.py
ChihChengLiang: I disagree, you are being fabulous 🌈.

ChihChengLiang:
Merge to unblock other tasks, post PR feedbacks are welcomed. 🙏

PR #492, opened 2019-04-10
Loosen the test for `pubsub.publish`

mhchia:
### What was wrong?
In https://github.com/ethereum/trinity/pull/480, CI failed in https://circleci.com/gh/gsmadi/trinity/885?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link. The failure is quite strange. Since there are only two connected pubsub, the receiver should receive the data in the order they are published.

### How was it fixed?
I couldn't reproduce the error(run on my laptop with 100+ times). In this situation, since we might already need to assume the order varies when publishing(broadcasting), IMO loosen the check might make sense: only check that the receiver receives the same number of the data as the published ones, but don't check their arriving order.

#### Cute Animal Picture

### DIFF
--- a/tests/libp2p/p2pclient/test_p2pclient_integration.py
+++ b/tests/libp2p/p2pclient/test_p2pclient_integration.py
@@ -75,7 +75,7 @@ async def try_until_success(coro_func, timeout=TIMEOUT_DURATION):
             break
         if (time.monotonic() - t_start) >= timeout:
             # timeout
-            assert False, f"{coro_func} still failed after `timeout` seconds"
+            assert False, f"{coro_func} still failed after `{timeout}` seconds"
         await asyncio.sleep(0.01)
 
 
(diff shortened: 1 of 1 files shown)


tests/libp2p/p2pclient/test_p2pclient_integration.py
```diff
@@ -1015,10 +1015,9 @@ def _get_current_running_dispatched_handler():
     await p2pds[0].pubsub.publish(topic, another_data_1)
     pubsub_msg_1_0 = p2pd_pb.PSMessage()
     await read_pbmsg_safe(reader_1, pubsub_msg_1_0)
-    assert pubsub_msg_1_0.data == another_data_0
     pubsub_msg_1_1 = p2pd_pb.PSMessage()
     await read_pbmsg_safe(reader_1, pubsub_msg_1_1)
-    assert pubsub_msg_1_1.data == another_data_1
+    assert set([pubsub_msg_1_0.data, pubsub_msg_1_1.data]) == set([another_data_0, another_data_1])
```
pipermerriam: What if instead of allowing them to come in any order, we instead do the following.

1. Assert message comes in correct order.
2. If unexpected message comes:
    - monitor for additional messages that come and then fail after X seconds reporting any follow up messages that arrive.

mhchia: Sorry for the late summary of the investigation. I dug more into `go-libp2p-pubsub` and verified it is possible that later published data arrive earlier. Message signing is enabled [by default](https://github.com/libp2p/go-libp2p-daemon/blob/master/p2pd/main.go#L74), and messages are verified in their own goroutines([code](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pubsub.go#L642)). They are sent after their validation goroutines finish([code](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pubsub.go#L671)).

> What if instead of allowing them to come in any order, we instead do the following.
>
> 1. Assert message comes in correct order.
> 2. If unexpected message comes:
>     - monitor for additional messages that come and then fail after X seconds reporting any follow up messages that arrive.

@pipermerriam 
Personally, I expect messages are received in their sent order. However, just want to confirm if we still want to assert them in our expected order. From this fact I am afriad we will still get similar CI errors sometimes, though we will get follow up messages reported).

mhchia:
Close and reopen to make CI run

mhchia:
Redo the commit with signature.

tests/libp2p/p2pclient/test_p2pclient_integration.py
pipermerriam: Based on what you said, it sounds like we should instead loosen this to allow out-of-order messages.

PR #493, opened 2019-04-11
Attempt to connect to peers concurrently.

pipermerriam:
### What was wrong?

Two things:

1. There was a bug in how the `PeerCandidateResponse` was handled.  In the case where the peer pool was empty, the peer pool would first make a request for peer candidates sourced from the discovery protocol, but then it would overwrite that response with a request for bootnodes, resulting in trinity only ever sourcing it's first outbound connection when the peer pool is empty from a bootnode.
2. Attempts to connect to new peers were done synchronously, not attempting a new connection until the previous attempt had finished.  This is very slow.

### How was it fixed?

- Fixed the bug to ensure that the bootnodes do not overwrite the nodes found from discovery.
- Updated the connection code so that we attempt connections concurrently.
- Also changed how `PeerPool.connect` works to raise exceptions instead of returning `None` on a failure.

#### Cute Animal Picture

miniature cattle echuca

### DIFF
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -30,6 +30,7 @@ from lahja import (
     BroadcastConfig,
 )
 
+from p2p._utils import clamp
 from p2p.constants import (
     DEFAULT_MAX_PEERS,
     DEFAULT_PEER_BOOT_TIMEOUT,
@@ -318,8 +319,9 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
 
             # only attempt to connect to up to the maximum number of available
             # peer slots that are open.
-            available_peer_slots = min(10, max(1, self.max_peers - len(self)))
-            batch = tuple(take(available_peer_slots, nodes_iter))
+            available_peer_slots = self.max_peers - len(self)
+            batch_size = clamp(1, 10, available_peer_slots)
+            batch = tuple(take(batch_size, nodes_iter))
 
             # There are no more *known* nodes to connect to.
             if not batch:
@@ -331,11 +333,10 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
                 available_peer_slots,
             )
             # Try to connect to the peers concurrently.
-            # TODO: does this need to be wrapped in `self.wait`...
-            await asyncio.gather(
+            await self.wait(asyncio.gather(
                 *(self.connect_to_node(node) for node in batch),
                 loop=self.get_event_loop(),
-            )
+            ))
 
     async def connect_to_node(self, node: Node) -> None:
         """
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -23,6 +23,7 @@ from eth_keys import (
 )
 from eth_utils.toolz import (
     groupby,
+    take,
 )
 from lahja import (
     Endpoint,
@@ -41,6 +42,8 @@ from p2p.events import (
     RandomBootnodeRequest,
 )
 from p2p.exceptions import (
+    BaseP2PError,
+    IneligiblePeer,
     BadAckMessage,
     HandshakeFailure,
     MalformedMessage,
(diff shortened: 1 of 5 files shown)


p2p/peer_pool.py
```diff
@@ -139,9 +157,12 @@ def __init__(self,
                             "Discovery did not answer RandomBootnodeRequest in time"
                         )
                         continue
+                    candidates = response.candidates + bootnodes_response.candidates
+                else:
+                    candidates = response.candidates
```
pipermerriam: @carver this is actually a bugfix you might want to pull into your branch.  Without this, I believe that you will only ever initiate your first peer connection from a bootnode.

carver: Ah, interesting, so you start up with 0 peers, fall-back to the bootnodes and ignore the `PeerCandidatesRequest`. So that could definitely explain some difficulty in getting the first peer connection...

lithp: Wow, great catch, can't believe it took so long to notice this.

p2p/peer_pool.py
```diff
@@ -264,40 +282,97 @@ def _add_peer(self,
...
+            if not batch:
+                return
+
+            self.logger.debug(
+                'Initiating %d peer connection attempts with %d open peer slots',
+                len(batch),
+                available_peer_slots,
+            )
+            # Try to connect to the peers concurrently.
+            # TODO: does this need to be wrapped in `self.wait`...
```
lithp: I think it doesn't, because `self.connect` wraps the call to `handshake` in a `self.wait`. It might still be nice to add a `self.wait` here, this function would then make fewer assumptions about the rest of the code.

p2p/peer_pool.py
```diff
@@ -264,40 +282,97 @@ def _add_peer(self,
...
+        while True:
             if self.is_full or not self.is_operational:
                 return
 
-            # TODO: Consider changing connect() to raise an exception instead of returning None,
-            # as discussed in
-            # https://github.com/ethereum/py-evm/pull/139#discussion_r152067425
+            # only attempt to connect to up to the maximum number of available
+            # peer slots that are open.
+            available_peer_slots = min(10, max(1, self.max_peers - len(self)))
```
lithp: nitpick: this took me a second to understand. `min` is called min but is actually setting the upper bound, `max` is actually setting the lower bound. Stackoverflow has a suggestion which maps to: `sorted((1, self.max_peers - len(self), 10))[1]`, which *might* be easier to understand? Now that I've thought about both they both make sense.

p2p/peer_pool.py
```diff
@@ -264,40 +282,97 @@ def _add_peer(self,
...
+            batch = tuple(take(available_peer_slots, nodes_iter))
+
+            # There are no more *known* nodes to connect to.
+            if not batch:
+                return
+
+            self.logger.debug(
+                'Initiating %d peer connection attempts with %d open peer slots',
```
lithp: nitpick: this variable name and message are a little misleading. `available_peer_slots` is not equal to the number of open peer slots if there are more than 10 open slots or if there are few candidates.

p2p/peer_pool.py
```diff
@@ -264,40 +282,97 @@ def _add_peer(self,
...
+            return
+
+        # Check again to see if we have *become* full since the previous
+        # check.
+        if self.is_full:
+            self.logger.debug(
+                "Successfully connected to %s but peer pool is full.  Disconnecting.",
+                peer,
+            )
+            await peer.disconnect(DisconnectReason.too_many_peers)
```
lithp: An optimization here might be to try to disconnect an inbound peer but that's not worth complicating this PR for.

p2p/peer_pool.py
pipermerriam: I'm gonna write a utility. `clamp(lower_bound, upper_bound, value)`

pipermerriam:
Test passed, waiting for ssh session to timeout.  merging.

PR #496, opened 2019-04-12
Move sourcing of peer candidates into a backend system

pipermerriam:
### What was wrong?

We want to try and re-use known good peers from historical connection data.  This is a preliminary step towards this since the current construct doesn't lend itself well to extension outside of the `p2p` code which will be needed in the trinity side.

### How was it fixed?

Peer candidates are now sourced across a set of *backends*, each of which can source peers however it chooses.

#### Cute Animal Picture

85152cb53f52a7e54e317f5a4f7b64ec

### DIFF
--- a/p2p/_utils.py
+++ b/p2p/_utils.py
@@ -1,6 +1,4 @@
 import asyncio
-import bisect
-import collections
 from concurrent.futures import Executor, ProcessPoolExecutor
 import datetime
 import logging
@@ -9,7 +7,6 @@ import signal
 import time
 from typing import (
     AsyncGenerator,
-    Deque,
     Tuple,
     Union,
 )
@@ -117,31 +114,29 @@ def ensure_global_asyncio_executor(cpu_count: int=None) -> Executor:
     return _executor
 
 
-async def burstable_rate_limiter(num_per_time_unit: Union[int, float],
-                                 time_unit_in_seconds: Union[int, float],
-                                 ) -> AsyncGenerator[None, None]:
+async def token_bucket(rate: Union[int, float],
+                       capacity: Union[int, float],
+                       ) -> AsyncGenerator[None, None]:
     """
-    Given a rate, which represents number of attempts per second
+    rate: Number of token that can be consumed per second.
+    capacity: Maximum number of tokens
+
+    Each `await` call consumes a single token.
     """
-    history: Deque[float] = collections.deque()
-    sleep_time = time_unit_in_seconds / num_per_time_unit
+    num_tokens = capacity
+    last_refill = time.perf_counter()
+    seconds_per_token = 1 / rate
 
     while True:
         now = time.perf_counter()
+        num_tokens = min(
+            capacity,
+            num_tokens + (rate * (now - last_refill)),
+        )
+        last_refill = now
 
-        # discard any stale history
-        cutoff = now = time_unit_in_seconds
-        discard_index = bisect.bisect_left(history, cutoff)
-
-        for _ in range(discard_index):
-            history.popleft()
-
-        if len(history) < num_per_time_unit:
-            # not at capacity yet, can yield without waiting.
-            history.append(now)
+        if num_tokens >= 1:
+            num_tokens -= 1
             yield
         else:
-            # at capacity so must wait the average sleep time
-            await asyncio.sleep(sleep_time)
-            history.append(time.perf_counter())
-            yield
+            await asyncio.sleep((1 - num_tokens) * seconds_per_token)
--- a/p2p/_utils.py
+++ b/p2p/_utils.py
@@ -1,9 +1,18 @@
-import datetime
+import asyncio
+import bisect
+import collections
 from concurrent.futures import Executor, ProcessPoolExecutor
+import datetime
 import logging
 import os
 import signal
-from typing import Tuple
+import time
+from typing import (
+    AsyncGenerator,
+    Deque,
+    Tuple,
+    Union,
+)
 
 import rlp
 
@@ -106,3 +115,33 @@ def ensure_global_asyncio_executor(cpu_count: int=None) -> Executor:
         _executor._start_queue_management_thread()  # type: ignore
         signal.signal(signal.SIGINT, original_handler)
     return _executor
+
+
+async def burstable_rate_limiter(num_per_time_unit: Union[int, float],
+                                 time_unit_in_seconds: Union[int, float],
+                                 ) -> AsyncGenerator[None, None]:
+    """
+    Given a rate, which represents number of attempts per second
+    """
+    history: Deque[float] = collections.deque()
+    sleep_time = time_unit_in_seconds / num_per_time_unit
+
+    while True:
+        now = time.perf_counter()
+
+        # discard any stale history
+        cutoff = now = time_unit_in_seconds
+        discard_index = bisect.bisect_left(history, cutoff)
+
+        for _ in range(discard_index):
+            history.popleft()
+
+        if len(history) < num_per_time_unit:
+            # not at capacity yet, can yield without waiting.
+            history.append(now)
+            yield
+        else:
+            # at capacity so must wait the average sleep time
+            await asyncio.sleep(sleep_time)
+            history.append(time.perf_counter())
+            yield
--- a/p2p/peer_backend.py
+++ b/p2p/peer_backend.py
@@ -1,0 +1,64 @@
+from abc import ABC, abstractmethod
+from typing import (
+    Iterable,
+    NamedTuple,
+)
+
+from lahja import (
+    Endpoint,
+    BroadcastConfig,
+)
+
+from p2p.constants import (
+    DISCOVERY_EVENTBUS_ENDPOINT,
+)
+from p2p.kademlia import (
+    from_uris,
+    Node,
+)
+from p2p.events import (
+    PeerCandidatesRequest,
+    RandomBootnodeRequest,
+)
+
+
+class PoolMeta(NamedTuple):
+    num_connected_peers: int
+    num_requested: int
+
+
+class BasePeerBackend(ABC):
+    @abstractmethod
+    async def get_peer_candidates(self, pool_info: PoolMeta) -> Iterable[Node]:
+        pass
+
+
+TO_DISCOVERY_BROADCAST_CONFIG = BroadcastConfig(filter_endpoint=DISCOVERY_EVENTBUS_ENDPOINT)
+
+
+class DiscoveryPeerBackend(BasePeerBackend):
+    def __init__(self, event_bus: Endpoint) -> None:
+        self.event_bus = event_bus
+
+    async def get_peer_candidates(self, pool_info: PoolMeta) -> Iterable[Node]:
+        response = await self.event_bus.request(
+            PeerCandidatesRequest(pool_info.num_requested),
+            TO_DISCOVERY_BROADCAST_CONFIG,
+        )
+        return from_uris(response.candidates)
+
+
+class BootnodesPeerBackend(BasePeerBackend):
+    def __init__(self, event_bus: Endpoint) -> None:
+        self.event_bus = event_bus
+
+    async def get_peer_candidates(self, pool_info: PoolMeta) -> Iterable[Node]:
+        if pool_info.num_connected_peers == 0:
+            response = await self.event_bus.request(
+                RandomBootnodeRequest(),
+                TO_DISCOVERY_BROADCAST_CONFIG
+            )
+
+            return from_uris(response.candidates)
+        else:
+            return ()
--- a/p2p/peer_backend.py
+++ b/p2p/peer_backend.py
@@ -1,7 +1,6 @@
 from abc import ABC, abstractmethod
 from typing import (
     Iterable,
-    NamedTuple,
 )
 
 from lahja import (
@@ -22,14 +21,11 @@ from p2p.events import (
 )
 
 
-class PoolMeta(NamedTuple):
-    num_connected_peers: int
-    num_requested: int
-
-
 class BasePeerBackend(ABC):
     @abstractmethod
-    async def get_peer_candidates(self, pool_info: PoolMeta) -> Iterable[Node]:
+    async def get_peer_candidates(self,
+                                  num_requested: int,
+                                  num_connected_peers: int) -> Iterable[Node]:
         pass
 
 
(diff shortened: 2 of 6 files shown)


p2p/peer_backend.py
```diff
@@ -0,0 +1,65 @@
...
+)
+from p2p.events import (
+    PeerCandidatesRequest,
+    RandomBootnodeRequest,
+)
+
+
+class PoolMeta(NamedTuple):
+    max_peers: int
+    num_peers: int
```
carver: Looks like `max_peers` isn't used, could be dropped. Maybe `num_peers` -> `num_connected_peers` to disambiguate?

Actually, at that point, maybe `get_peer_candidates(self, num_requested: int, num_connected: int)` simplifies even more.

p2p/peer_pool.py
```diff
@@ -104,44 +105,42 @@ def __init__(self,
...
-
-                self.logger.debug2("Received candidates to connect to (%s)", response.candidates)
-                await self.connect_to_nodes(from_uris(response.candidates))
+                    else:
+                        self.logger.debug2(
+                            "Got candidates from backend %s (%s)",
+                            backend,
+                            candidates,
+                        )
+                        await self.connect_to_nodes(iter(candidates))
```
carver: We could potentially connect to more peers than our max, since `num_requested` is calculated once and broadcast to all backends (probably not with exactly the current backends, but easily from adding others, or even just reversing the order).

One option is to change `connect_to_nodes` to return the successful connections, then update `num_requested` before the next loop.

p2p/peer_pool.py
```diff
@@ -104,44 +105,42 @@ def __init__(self,
...
+                    max_peers=self.max_peers,
+                    num_peers=len(self),
+                    num_requested=available_slots,
+                )
 
-                # In some cases (e.g ROPSTEN or private testnets), the discovery table might be
-                # full of bad peers so if we can't connect to any peers we try a random bootstrap
-                # node as well.
-                if not len(self):
+                for backend in self.peer_backends:
```
carver: I vaguely want to do:

```py
await asyncio.gather(self._add_peers_from_backend(backend) for backend in self.peer_backends)
```
So that all the peer requests can all be running in parallel. But... that makes it even harder to maintaining the peer limit at the `backend.get_peer_candidates()` API. Maybe we could enforce the limit inside `connect_to_nodes()` instead. Should be straightforward there.

pipermerriam: I'll rework it to be parallel like this.

p2p/peer_backend.py
pipermerriam: Did the rename and removed `max_peers` but I still like having the data object unless you feel strongly.

p2p/peer_pool.py
pipermerriam: Take another look at this.  After having merged some other PRs I think this all now fits what you were wanting.

p2p/peer_pool.py
```diff
@@ -124,47 +125,46 @@ def __init__(self,
...
+                pool_info = PoolMeta(
+                    num_connected_peers=len(self),
+                    num_requested=available_slots,
+                )
 
-                # In some cases (e.g ROPSTEN or private testnets), the discovery table might be
-                # full of bad peers so if we can't connect to any peers we try a random bootstrap
-                # node as well.
-                if not len(self):
+                async def _add_peers_from_backend(backend: BasePeerBackend) -> None:
```
carver: Why redefine it on every loop instead of as a private method on `PeerPool` that you pass pool_info into?

p2p/peer_backend.py
carver: Eh, the indirection seems unnecessary (why this one and not other times we pass in more than one related argument?). Not a dealbreaker, but takes a smidge longer to read the implementations and I am not seeing the upside.

pipermerriam:
> Eh, the indirection seems unnecessary (why this one and not other times we pass in more than one related argument?). Not a dealbreaker, but takes a smidge longer to read the implementations and I am not seeing the upside.

Maybe just me being emotionally attached to my own code...  but I struggled with what arguments to pass into this and didn't like that `DiscoveryBackend` only uses `num_requested` and `BootnodesPeerBackend` only uses `num_connected_peers` and so for each of them there is an un-used argument.  It felt *less* messy if they were packaged as a single object.

Tell me what to do....  I cant sort out whether this is good or just replacing one unfortunate implementation detail with another (unfortunate that both arguments aren't needed in both places vs unfortunate that they are wrapped up in an object and thus less easily understood).

p2p/peer_pool.py
pipermerriam: I didn't want to re-compute `available_peer_slots` within that method and it felt wrong to pass it in...  I'll take another pass at it.  Private method seeming more fine now that I'm forced to look at it.

carver:
> Tell me what to do.... I cant sort out whether this is good or just replacing one unfortunate implementation detail with another (unfortunate that both arguments aren't needed in both places vs unfortunate that they are wrapped up in an object and thus less easily understood).

I say let's do:
```py
async def get_peer_candidates(self, num_requested: int, num_connected: int) -> Iterable[Node]:
```

In the trivial case, it doesn't bother me that the arguments aren't being used:

```py
class NoPeerBackend(BasePeerBackend):
    async def get_peer_candidates(self, num_requested: int, num_connected: int) -> Iterable[Node]:
        return []
```

So I guess it doesn't bother me if any particular implementation only uses some of the information it's supplied.

I feel the (same?) worry that maybe we might get to a place where we want to send another piece of data, and another piece of data, and have a huge signature. To me, that starts to imply that we have the wrong abstraction boundary anyway, and should refactor. Maybe to localize the relevant data, or provide a reference to the object that provides it, or whatever. We can decide what to do when the requirements start to become more apparent.

p2p/peer_pool.py
```diff
@@ -124,47 +124,46 @@ def __init__(self,
...
+    async def maybe_connect_more_peers(self) -> None:
+        while self.is_operational:
+            if self.is_full:
+                await self.sleep(DISOVERY_INTERVAL)
+                continue
+
+            await self.wait(asyncio.gather(*(
+                self._add_peers_from_backend(backend)
+                for backend in self.peer_backends
+            )))
```
pipermerriam: @carver one last substantive change here.  We no longer sleep between requests *if* we aren't full.  I don't see a reason why we'd want to not try to connect to as many peers as fast as possible...

carver: Ah interesting, I mostly agree. But flipped on its head: is it too much to spin this loop constantly? My peer set is essentially never full (though maybe we'll get this to change soon!). So how hard does this rev up?

I guess the latency to the peers during the connection attempts will keep us from really slamming the CPU in this loop. If discovery returns an empty set, this could potentially spin fast, though. (Like when we disable discovery?) Maybe it should just be a shorter sleep if we're not full, like `0.1s` (ugh, another constant). Another reason to be cautious about dropping the sleep entirely is: if discovery is not returning anything, then we might be really slamming the bootnodes with connection requests over and over.

So maybe something like:
```py
if self.is_full:
  await self.sleep(FULL_PEERPOOL_PAUSE)  # this could be pretty long, like 2s. I guess it could me merged with the next constant
  # ideally ^ would actually be something like below, which would trigger on an event when a peer drops:
  await self.wait(self.has_capacity())
  # not for this PR though, I think
  continue
elif not any(attempted_peers):  # I suppose this would require returning the attempted peers from `_add_peers_from_backend()` We don't have to merge them, since any([[], []]) is still False
  await self.sleep(EMPTY_DISCOVERY_PAUSE)  # we want to give discovery a chance to refill, maybe 2s
else:
  await self.sleep(DISCOVERY_INTERVAL)  # Could be very short like 0.05s. Maybe we don't need this sleep at all, I'm just worried about slamming bootnodes or when we have a small discovered peer group.
```

Not for this PR, but I'm wondering how often we keep trying to connect to the same peers if we try and fail, and how long in between retries. How good is discovery about skipping recently attempted peers? How does discovery know not to send us peers that we're already connected to? These are just idle musings. Obviously I should just go read the discovery code...

Also looks like some old cleanup: `DISOVERY_INTERVAL` -> `DISCOVERY_INTERVAL`

pipermerriam: As we briefly discussed, this is now handled by a rate limiter in this loop.  The loop is allowed to burst with back-to-back checks until it hits a rate of 10 checks per 20 seconds after which the rate of checks is capped at this rate.  Take a look and tell me what you think.

carver: Looks like the new commits didn't get pushed

pipermerriam: pushed

tests/p2p/test_burstable_rate_limiter.py
```diff
@@ -0,0 +1,31 @@
...
+    limiter = burstable_rate_limiter(10, 0.01)
+
+    start_at = time.perf_counter()
+    for _ in range(10):
+        await limiter.__anext__()
+
+    end_at = time.perf_counter()
+    delta = end_at - start_at
+    print('DELTA:', delta)
+    assert delta < 0.001
```
pipermerriam: Tests on my machine showed this has a buffer of about 2 more decimal places before the *fuzzy* part of this tests produces a failure.

tests/p2p/test_burstable_rate_limiter.py
```diff
@@ -0,0 +1,31 @@
...
+async def test_burstable_rate_limiter_hits_limit():
+    limiter = burstable_rate_limiter(5, 0.01)
+
+    start_at = time.perf_counter()
+    for _ in range(25):
+        await limiter.__anext__()
+
+    end_at = time.perf_counter()
+    delta = end_at - start_at
+    assert abs(delta - 0.05) < 0.01
```
pipermerriam: Tests on my machine showed that this has a buffer of about `0.015` (2.5x) before this fuzzyness fails.

p2p/peer_pool.py
```diff
@@ -124,47 +124,49 @@ def __init__(self,
...
+            await self.connect_to_nodes(iter(candidates))
+
+    async def maybe_connect_more_peers(self) -> None:
+        # allowed to operate at a maximum rate of 10 checks every 20 seconds.
+        rate_limiter = burstable_rate_limiter(10, 20)
+        while True:
+            if self.is_full:
+                await self.sleep(DISOVERY_INTERVAL)
+                continue
+            await rate_limiter.__anext__()
```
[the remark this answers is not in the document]
pipermerriam: I tried the `async for` approach but the linter didn't like it.  Updated to use `self.wait`.

p2p/_utils.py
```diff
@@ -106,3 +115,33 @@ def ensure_global_asyncio_executor(cpu_count: int=None) -> Executor:
...
+    Given a rate, which represents number of attempts per second
+    """
+    history: Deque[float] = collections.deque()
+    sleep_time = time_unit_in_seconds / num_per_time_unit
+
+    while True:
+        now = time.perf_counter()
+
+        # discard any stale history
+        cutoff = now = time_unit_in_seconds
```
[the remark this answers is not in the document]
pipermerriam: I was being sloppy earlier today, trying to get this done before an apt I had to make it to.  I just sat back down after looking over the TokenBucket implementation you linked and and then re-worked the tests so that they were indeed testing what I thought they needed to test.

I considered the linked existing implementation and I'm not opposed to using it, however, it isn't exactly clear to me *how* we can wrap that in asyncio without polling and/or doing some kind of deep inspection of the bucket to determine how long we should sleep for.

Take a look at the updated code and tell me what you think.  It's way cleaner.  I was strugling earlier to model the concept I wanted in code but the token bucket does it perfectly.

p2p/_utils.py
```diff
@@ -106,3 +112,31 @@ def ensure_global_asyncio_executor(cpu_count: int=None) -> Executor:
...
+            capacity,
+            num_tokens + (rate * (now - last_refill)),
+        )
+        last_refill = now
+
+        if num_tokens >= 1:
+            num_tokens -= 1
+            yield
+        else:
+            await asyncio.sleep((1 - num_tokens) * seconds_per_token)
```
carver: *nit* Can `num_tokens` be negative? I think this always reduces to:

```py
await asyncio.sleep(seconds_per_token)
```

I guess if `rate` was negative all kinds of things would blow up. Other than that, I can't see `num_tokens` going negative.

p2p/peer_pool.py
```diff
@@ -124,47 +124,49 @@ def __init__(self,
...
+                candidates,
+            )
+            await self.connect_to_nodes(iter(candidates))
+
+    async def maybe_connect_more_peers(self) -> None:
+        # allowed to operate at a maximum rate of 1 check every 2 seconds
+        rate_limiter = token_bucket(0.5, 5)
+        while True:
+            if self.is_full:
+                await self.sleep(DISOVERY_INTERVAL)
```
carver: *nit* from old code

```suggestion
                await self.sleep(DISCOVERY_INTERVAL)
```

(plus the import, etc)

p2p/peer_pool.py
```diff
@@ -124,47 +124,49 @@ def __init__(self,
...
+            self.logger.debug2(
+                "Got candidates from backend %s (%s)",
+                backend,
+                candidates,
+            )
+            await self.connect_to_nodes(iter(candidates))
+
+    async def maybe_connect_more_peers(self) -> None:
+        # allowed to operate at a maximum rate of 1 check every 2 seconds
+        rate_limiter = token_bucket(0.5, 5)
```
carver: :+1: looks like it's settled down enough to move the magic numbers to `p2p.constants`.

Something like `MAX_SEQUENTIAL_DISCOVERY = 5`. I suppose we can reuse the rate like so:

```suggestion
        rate_limiter = token_bucket(1 / DISCOVERY_INTERVAL, MAX_SEQUENTIAL_DISCOVERY)
```

(or inverting the constant to `DISCOVERY_RATE` and inverting the sleep below... which looks a tiny bit cleaner to my eyes)

tests/p2p/test_burstable_rate_limiter.py
```diff
@@ -0,0 +1,64 @@
...
+    # first 10 tokens should be roughly instant
+    # next 10 tokens should each take 1/1000th second each to generate.
+    for _ in range(20):
+        await limiter.__anext__()
+
+    end_at = time.perf_counter()
+
+    expected_delta = 10 / 1000
+    delta = end_at - start_at
+    print('DELTA:', delta, expected_delta)
```
carver:
```suggestion
```

p2p/_utils.py
pipermerriam: This takes into account that tokens are fractional and thus a refill could place the token count at `1.5`.  Then taking a token out brings the balance down to `0.5`.  At that point we only should wait the duration of `1/2` of the `seconds_per_token` since we only need `1/2` of a token.

carver: Ah right, they always decrement by 1, but they could start fractional...

lithp:
`test_token_bucket_hits_limit` occasionally fails: https://circleci.com/gh/ethereum/trinity/29739

pipermerriam:
yeah, I need to re-work those tests to just manually advance the `time.perf_counter` that the token bucket uses rather than trying to do precise timing in the tests.

PR #497, opened 2019-04-12
Fix ecies deprecation warning

pipermerriam:
### What was wrong?

Deprecation warnings were being issued from APIs we were using in the `p2p.ecies` modules.

### How was it fixed?

Followed instructions in the warning message to update to the non-depracated APIs.

#### Cute Animal Picture

060103_cattle_hmed_6p grid-6x2

### DIFF
--- a/p2p/ecies.py
+++ b/p2p/ecies.py
@@ -55,8 +55,8 @@ def ecdh_agree(privkey: datatypes.PrivateKey, pubkey: datatypes.PublicKey) -> by
     pubkey_bytes = b'\x04' + pubkey.to_bytes()
     try:
         # either of these can raise a ValueError:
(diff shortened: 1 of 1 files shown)


ChihChengLiang:
It looks like this requires an upgrade in `cryptography` package.
I'm getting an error of 
```
AttributeError: type object 'EllipticCurvePublicKey' has no attribute 'from_encoded_point'
```
in `pytest tests/p2p/test_ecies.py`, with `cryptography==2.4.2`

pipermerriam:
looks like you need `>=2.6`  tracking down the right place to do the dependency declaration now.

pipermerriam:
https://github.com/ethereum/trinity/pull/514

PR #498, opened 2019-04-12
Update p2p tests to use factory boy

pipermerriam:
### What was wrong?

We had a few scattered implementations of the same functions for generating things like random nodes and addresses.  Some of these were used across modules resulting in tests importing from within the` `tests/...` namespace.

### How was it fixed?

- Created factories using the [factory-boy](https://factoryboy.readthedocs.io/en/latest/) library for these various constructs and a new pytest fixture to make accessing them easy.
- Updated all tests to use these constructs.

#### Cute Animal Picture

huggie18may2012


setup.py
```diff
@@ -34,6 +34,7 @@
     'test': [
         "hypothesis==3.69.5",
         "pexpect>=4.6, <5",
+        "factory-boy==2.11.1",
```
lithp: Nitpick: `factory-boy` and `factory_boy` appear to be the same package, I think it'd be simpler to use `factory_boy` everywhere for consistency.

p2p/tools/factories.py
```diff
@@ -0,0 +1,86 @@
...
+
+from p2p import discovery
+from p2p import kademlia
+from p2p.ecies import generate_privkey
+
+
+try:
+    import factory
+except ImportError:
+    raise ImportError("The p2p.tools.factories module requires the `factory_boy` library.")
```
lithp: I think that in most situations python will show the line which caused an exception. You might get the same effect as this block by writing:

```python
import factory  # from the factory_boy library
```

This is a nitpick but even though `factory` is an annoyingly generic name it still feels weird to give factory_boy this special treatment. If a `git pull` starts causing `ImportError`s it's already pretty clear you're missing some library and a `pip install` should fix things up.

p2p/tools/factories.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def get_open_port() -> int:
```
lithp: nitpick: trinity also uses `pytest-asyncio` which [already provides](https://github.com/pytest-dev/pytest-asyncio/blob/master/pytest_asyncio/plugin.py#L219) this functionality.

There's a race condition between this function returning a port and the port being used by a test but that's probably usually fine, re-running tests is cheap.

tests/p2p/test_discovery_v5.py
```diff
@@ -1,39 +1,30 @@
...
 
 
-async def get_listening_discovery_protocol(event_loop):
-    addr = kademlia.Address('127.0.0.1', random.randint(1024, 9999))
-    proto = get_discovery_protocol(os.urandom(4), addr)
+async def get_listening_discovery_protocol(factories, event_loop):
+    addr = factories.AddressFactory.localhost()
+    proto = factories.DiscoveryProtocolFactory(address=addr)
     await event_loop.create_datagram_endpoint(
         lambda: proto, local_addr=(addr.ip, addr.udp_port), family=socket.AF_INET)
```
lithp: I don't think the race condition is very important, but to remove it we could pass 0 as the port here and read out the port we're bound to using the same trick:

```python
sock = transport.get_extra_info('socket')
port = sock.getsockname()[1]
```

This function might be another candidate for becoming a factory.

p2p/tools/factories.py
pipermerriam: My inclination to issue a custom exception message is that these `library.tools` namespaced APIs are intended for use as public APIs but they are allowed to have dependencies that require manual installation like this.  Issuing a custom import error allows us to give the user a better error message to indicate what they need to do to fix this.

I recognize it would be clear to the user why the failure happened (because of an import error) but I don't think that extends to it being clear exactly what dependency they need to install or that this isn't a library bug.  Issuing a message like this keeps us from having tons of support tickets asking why this is happening.

p2p/tools/factories.py
pipermerriam: I'm aware, however, in this case I need access to this outside of the testing environment, thus implementing it inline here.  I don't see a clean way to make use of the one provided by the `pytest-asyncio` library.

setup.py
carver: Interestingly pip is loose with the `-` and `_` so you always get both for "free". Going with the project's preferred version makes sense to me. Looks like the docs use `_`.

PR #500, opened 2019-04-13
[WIP] Support lahja#36

lithp:
### What was wrong?

Not yet ready to merge. This is just a hack and it still has ugly output when the process exits: 

```
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/home/brian/ef/lahja/lahja/endpoint.py", line 147, in _run
    self.condition.wait()
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/managers.py", line 1032, in __exit__
    return self._callmethod('release')
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/managers.py", line 795, in _callmethod
    conn.send((self._id, methodname, args, kwds))
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/connection.py", line 207, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/connection.py", line 405, in _send_bytes
    self._send(header + buf)
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/connection.py", line 369, in _send
    n = write(self._handle, buf)
BrokenPipeError: [Errno 32] Broken pipe
```

However, without this PR trinity experiences some [crazy segfaults](https://github.com/ethereum/lahja/pull/36#issuecomment-482758419) when running under lahja#36. With this PR, those segfaults are gone and everything else appears to go normally!

### How was it fixed?

Endpoints are no longer sent across process boundaries.

#### Cute Animal Picture

### DIFF
--- a/trinity/extensibility/plugin_manager.py
+++ b/trinity/extensibility/plugin_manager.py
@@ -99,7 +99,7 @@ class MainAndIsolatedProcessScope(BaseManagerProcessScope):
             # created here for API symmetry. Endpoints are pickleable *before* they are connected,
             # which means, this Endpoint will be pickled and transferred into the new process
             # together with the rest of the `PluginContext`.
-            plugin.set_context(PluginContext(TrinityEventBusEndpoint(), boot_info,))
+            plugin.set_context(PluginContext(None, boot_info,))
 
 
 class SharedProcessScope(BaseManagerProcessScope):
--- a/trinity/bootstrap.py
+++ b/trinity/bootstrap.py
@@ -144,7 +144,6 @@ def main_entry(trinity_boot: BootFn,
     if args.log_levels:
         setup_log_levels(args.log_levels)
 
-    main_endpoint.track_and_propagate_available_endpoints()
     try:
         trinity_config = TrinityConfig.from_parser_args(args, app_identifier, sub_configs)
     except AmbigiousFileSystem:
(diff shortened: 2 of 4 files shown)


trinity/extensibility/plugin_manager.py
```diff
@@ -99,7 +99,7 @@ def create_plugin_context(self,
             # created here for API symmetry. Endpoints are pickleable *before* they are connected,
             # which means, this Endpoint will be pickled and transferred into the new process
             # together with the rest of the `PluginContext`.
-            plugin.set_context(PluginContext(TrinityEventBusEndpoint(), boot_info,))
+            plugin.set_context(PluginContext(None, boot_info,))
```
cburgdorf: Now I feel bad! Because really the only reason why we pickle it here is for API symmetry (as the comment suggests) because it used to be that all endpoints had to be created in the main process and then send over. After we got rid of that architecture we still had the `BaseAsyncStopPugin` left that doesn't create its own endpoint and so I kept using the `PluginContext` for both really just to not sending `None` here or getting dragged into a bigger refactoring. 😅

lithp:
Closing in lieu of #556

trinity/extensibility/plugin_manager.py
lithp: No reason to feel bad, nobody could have known it would have caused this :sweat_smile:

Issue #237, opened 2019-01-29
Dead fixtures in the codebase


ChihChengLiang:
### What is wrong?

Found `pytest-deadfixtures` that can detect dead fixtures in the codebase.

```
$ pytest --dead-fixtures

Hey there, I believe the following fixture(s) are not being used:
Fixture name: chain_without_block_validation, location: tests/conftest.py:230
Fixture name: nodekey, location: tests/core/initialization/test_initialize_data_dir.py:27
Fixture name: beacon_chain_without_block_validation, location: tests/eth2/beacon/chains/conftest.py:60
Fixture name: chain, location: tests/eth2/beacon/chains/test_chain.py:24
Fixture name: sample_recent_proposer_record_params, location: tests/eth2/beacon/conftest.py:273
Fixture name: sample_validator_registry_delta_block_params, location: tests/eth2/beacon/conftest.py:329
Fixture name: init_shuffling_seed, location: tests/eth2/beacon/conftest.py:377
Fixture name: init_validator_privkeys, location: tests/eth2/beacon/conftest.py:392
Fixture name: chain_fixture, location: tests/json-fixtures-over-rpc/test_rpc_fixtures.py:380
Fixture name: chain, location: tests/json-fixtures-over-rpc/test_rpc_fixtures.py:388

1 error in 3.01 seconds
```

### How can it be fixed

- Check and remove the dead fixtures
- Maybe add the check to ci, if the check is reliable?

ChihChengLiang:
For eth2 part, `chain`, `beacon_chain_without_block_validation`, `sample_validator_registry_delta_block_params`, `sample_recent_proposer_record_params`, `init_shuffling_seed`, and `init_validator_privkeys` are all manually confirmed dead.

carver:
Nice, I'll remove the eth1 dead fixtures:

 - [x] chain_without_block_validation, location: tests/conftest.py:230   -     seems to be used in:
```
tests/json-fixtures-over-rpc/test_rpc_fixtures.py:388:def chain(chain_without_block_validation):jk
```                                                  
 - [x] nodekey, location: tests/core/initialization/test_initialize_data_dir.py:27                                                    
 - [x] chain_fixture, location: tests/json-fixtures-over-rpc/test_rpc_fixtures.py:380
 - [x] chain, location: tests/json-fixtures-over-rpc/test_rpc_fixtures.py:388 - this and `chain_fixture` seem to be used in:
```
[...]

Issue #292, opened 2019-02-19
Error during snappy decompression


cburgdorf:
### What is wrong?

```
b'\x1b[1m\x1b[31m   ERROR  02-19 09:50:08           ETHPeerPool  Unexpected error during auth/p2p handshake with <Node(0x865a63255b3bb68023b6bffd5095118fcc13e79dcf014fe4e47e065c350c7cc72af2e53eff895f11ba1bbb6a2b33271c1116ee870f266618eadfc2e78aa7349c@52.176.100.77:30303)>\n'
b'Traceback (most recent call last):\n'
b'  File "/home/circleci/repo/p2p/peer_pool.py", line 255, in connect\n'
b'    peer = await self.wait(handshake(remote, self.get_peer_factory()))\n'
b'  File "/home/circleci/repo/p2p/cancellable.py", line 20, in wait\n'
b'    return await self.wait_first(awaitable, token=token, timeout=timeout)\n'
b'  File "/home/circleci/repo/p2p/cancellable.py", line 42, in wait_first\n'
b'    return await token_chain.cancellable_wait(*awaitables, timeout=timeout)\n'
b'  File "/home/circleci/repo/.tox/py36-long_run_integration/lib/python3.6/site-packages/cancel_token/token.py", line 152, in cancellable_wait\n'
b'    return done.pop().result()\n'
b'  File "/home/circleci/repo/p2p/peer.py", line 128, in handshake\n'
b'    await peer.do_sub_proto_handshake()\n'
b'  File "/home/circleci/repo/p2p/peer.py", line 321, in do_sub_proto_handshake\n'
b'    cmd, msg = await self.read_msg()\n'
b'  File "/home/circleci/repo/p2p/peer.py", line 460, in read_msg\n'
b'    decoded_msg = cast(Dict[str, Any], cmd.decode(msg))\n'
b'  File "/home/circleci/repo/p2p/p2p_proto.py", line 78, in decode\n'
b'    raw_decoded = cast(Dict[str, int], super().decode(data))\n'
b'  File "/home/circleci/repo/p2p/protocol.py", line 123, in decode\n'
b'    encoded_payload = self.decompress_payload(compressed_payload)\n'
b'  File "/home/circleci/repo/p2p/protocol.py", line 130, in decompress_payload\n'
b'    return snappy.decompress(raw_payload)\n'
[...]

PR #504, opened 2019-04-16
Remove dead fixture

carver:
### What was wrong?

An unused fixture, detected in #237

### How was it fixed?

Removed it

#### Cute Animal Picture


carver:
CI failure is an unrelated bug that's tracked in #292

carver:
Hm, a couple flaky tests :(

The first one is a legitimate (but uncommon) problem. Discovery should always be able to start up within 30s.

But neither failure is related to this change, so I'll go ahead and merge.

PR #505, opened 2019-04-16
Update node connection failure tracking to use sqlalchemy

pipermerriam:
This is a pre-cursor to tracking more information about peers.

### What was wrong?

The previous logic used to persist information about previous attempts to connect to peers was built using the stdlib `sqlite` library.  This involved a decent bit of custom SQL and code to manage the database.

In attempting to write the new tracking code to track more information about peers it became cumbersome to do the SQL manually without an ORM to smooth out the database interactions. 

### How was it fixed?

Used [SQLAlchemy](https://docs.sqlalchemy.org/en/13/index.html) for database interactions and wrote a small wrapper around it to handle some basic database **interactions**.

Largely reworked all of the persistance code to be segregrated into:

1. Base database schema management tools for checking if database schema is correct and installing schema into an empty database.
2. Separate `ConnectionTracker` system that replaces the previous code for tracking connection failures.
3. New CLI flag for switching between the sqlite/memory/disabled
4. New CLI command for removing the sqlite database.

#### Cute Animal Picture

980x


pipermerriam:
TODO: add https://github.com/dropbox/sqlalchemy-stubs/

p2p/tracking/connection.py
```diff
@@ -0,0 +1,183 @@
...
+            humanize_seconds(timeout),
+            reason,
+        )
+
+    def _update_record(self, remote: Node, timeout: int, reason: str) -> None:
+        uri = remote.uri()
+        record = self._get_record(uri)
+        record.error_count += 1
+
+        adjusted_timeout = int(timeout * math.sqrt(record.error_count))
```
pipermerriam: I took some liberty here adjusting the timeout increase to be non linear.  open to feedback.

pipermerriam:
closing in favor of #534

PR #506, opened 2019-04-17
Update mypy and fix type hints

pipermerriam:
### What was wrong?

- New `eth-utils` functions that I want to use that were ported from the trinity codebase.
- New `mypy` version that is faster and fixes some things
- Bugfix in `eth-utils` for preserving type safety when using `to_dict` or `to_ordered_dict`

### How was it fixed?

Updated the two dependencies and addressed the various typing issues that `mypy` yelled about.

#### Cute Animal Picture

dims vetstreet com

### DIFF
--- a/eth2/beacon/_utils/random.py
+++ b/eth2/beacon/_utils/random.py
@@ -10,7 +10,6 @@ from eth_typing import (
     Hash32,
 )
 from eth_utils import (
-    to_tuple,
     ValidationError,
 )
 
@@ -66,10 +65,15 @@ def get_permuted_index(index: int,
     return new_index
 
 
-@to_tuple
 def shuffle(values: Sequence[TItem],
             seed: Hash32,
-            shuffle_round_count: int) -> Iterable[TItem]:
+            shuffle_round_count: int) -> Tuple[TItem, ...]:
+    return tuple(_shuffle(values, seed, shuffle_round_count))
+
+
+def _shuffle(values: Sequence[TItem],
+             seed: Hash32,
+             shuffle_round_count: int) -> Iterable[TItem]:
     """
     Return shuffled indices in a pseudorandom permutation `0...list_size-1` with
     ``seed`` as entropy.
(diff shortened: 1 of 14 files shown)


eth2/beacon/_utils/random.py
```diff
@@ -66,10 +65,15 @@ def get_permuted_index(index: int,
     return new_index
 
 
-@to_tuple
 def shuffle(values: Sequence[TItem],
             seed: Hash32,
-            shuffle_round_count: int) -> Iterable[TItem]:
+            shuffle_round_count: int) -> Tuple[TItem, ...]:
+    return tuple(_shuffle(values, seed, shuffle_round_count))
```
carver: Is this needed because of the weird `@to_tuple` mypy bug in we're seeing in eth-utils?

trinity/plugins/builtin/json_rpc/plugin.py
```diff
@@ -60,7 +59,7 @@ def configure_parser(self, arg_parser: ArgumentParser, subparser: _SubParsersAct
             help="Disables the JSON-RPC Server",
         )
 
-    def setup_eth1_modules(self, trinity_config: TrinityConfig) -> Tuple[Eth1ChainRPCModule, ...]:
+    def setup_eth1_modules(self, trinity_config: TrinityConfig) -> Tuple[BaseRPCModule, ...]:
```
carver: Since these methods are specifically returning this subset of modules, can't/shouldn't they be `Eth1`... modules? Same for beacon below.

eth2/beacon/_utils/random.py
pipermerriam: Yes, this one.  I'll update with a code comment: https://github.com/ethereum/eth-utils/issues/152

trinity/plugins/builtin/json_rpc/plugin.py
pipermerriam: I don't actually know.  I'll spend a bit more time understanding why we have the two types.

pipermerriam: looking at it, it doesn't appear to matter. Each of these functions return values are fed into `RPCServer` which only cares about getting an iterable of `BaseRPCModule`.

PR #510, opened 2019-04-17
Use RPC module class as conventional (& default) module name

carver:
### What was wrong?

#12 introduced a new requirement to explicitly name the modules. Seemed unnecessary, and it's nice to strongly encourage naming the class the same as the module in the RPC.

### How was it fixed?

Implement the default name as the lower-cased class name (like it was before #12), but implement it on the class instead of in the `RPCServer` (which I agree was a suboptimal place to name it).

#### Cute Animal Picture


carver:
(Plus a bonus dependency fix that was hiding in #497 )

PR #514, opened 2019-04-18
Specify the cryptography library as a p2p dependency since it is

pipermerriam:
### What was wrong?

https://github.com/ethereum/trinity/pull/497 migrated to a new cryptography API since the old one was deprecated.  Turns out we don't locally specify a `cryptography` dependency (it was coming from py-evm).

### How was it fixed?

Explicitely specified the dependency.

#### Cute Animal Picture

baby-Zebra-1024x682

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,8 @@ deps = {
         "asyncio-cancel-token==0.1.0a2",
         "async_lru>=0.1.0,<1.0.0",
         "eth-hash>=0.1.4,<1",
+        # cryptography does not use semver and allows breaking changes within `0.3` version bumps.
+        "cryptography>=2.6,<2.9",
         "netifaces>=0.10.7<1",
         "pysha3>=1.0.0,<2.0.0",
         "upnpclient>=0.0.8,<1",
--- a/setup.py
+++ b/setup.py
@@ -8,9 +8,9 @@ deps = {
     'p2p': [
         "asyncio-cancel-token==0.1.0a2",
         "async_lru>=0.1.0,<1.0.0",
-        "eth-hash>=0.1.4,<1",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -9,6 +9,8 @@
         "asyncio-cancel-token==0.1.0a2",
         "async_lru>=0.1.0,<1.0.0",
         "eth-hash>=0.1.4,<1",
+        # cryptography does not use semver and allows breaking changes within `0.3` version bumps.
+        "cryptography>=2.6,<2.9",
```
carver: Ah, good catch about not using semver. I think we have to pin it to <2.7 since any minor number bump can be a breaking change.

Also *nit* to move the dependency one line up (alphabetical)

carver:
I'll go ahead and remove the change from the unrelated #510

setup.py
carver: Also maybe a note that we use an API in 2.5, so that's a hard requirement. 2.6 is just to use the latest available for bugfixes, etc, AFAICT.

carver: Hm, reading the #497 follow-up, it sounded like 2.6 was a hard requirement. Can you add a comment why?

The [docs for `EllipticCurvePublicKey.from_encoded_point()`](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey.from_encoded_point) only mention that it's new in 2.5.

pipermerriam: I think 2.5 is fine. I was being overly specific (and couldn't find release notes to verify when that API was introduced)

carver: Cool, then I guess just dropping this to `>=2.5,<2.7` (since we know it works with 2.6), and moving it up a line :+1:

PR #518, opened 2019-04-19
Create log directory, if data dir is empty

### LINKED ISSUE #105: Refusing to start if `logs/` sub-folder is missing
### What is wrong?

I start up trinity in fresh directories a lot. I find it mildly annoying/surprising that I have to `mkdir -p /tmp/trindat/logs` instead of `mkdir /tmp/trindat`, when using `trinity --data-dir /tmp/trindat`. It feels like I need to know too much about the internal structure of the data directory.

_Picked up from https://github.com/ethereum/py-evm/pull/1654#discussion_r_

### How can it be fixed

I think we could do away with this check altogether...

But if other people think this is a preferred safety restriction, maybe we can add at least one special case: we allow you to launch trinity without a `logs/` folder, only if the data directory is completely empty.

carver:
### What was wrong?

Fixes #105

### How was it fixed?

In a common case, where the custom data directory is empty (and the log directory is a subdirectory), trinity will create the log directory for you.

#### Cute Animal Picture

### DIFF
--- a/tests/core/initialization/test_initialize_data_dir.py
+++ b/tests/core/initialization/test_initialize_data_dir.py
@@ -2,12 +2,18 @@ import pytest
 
 import os
 
+from trinity._utils.filesystem import (
+    is_under_path,
+)
 from trinity.initialization import (
     is_data_dir_initialized,
     initialize_data_dir,
 )
 from trinity.config import (
     TrinityConfig,
+)
+from trinity.exceptions import (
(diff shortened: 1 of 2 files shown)


carver:
@cburgdorf no rush, just for whenever you get to it

carver:
:+1: Ignoring unrelated wheel-cli test failure

PR #525, opened 2019-04-23
Validator service based on beacon plugin

mhchia:
### What was wrong?
Add validator service for https://github.com/ethereum/trinity/issues/488 

### How was it fixed?
- Add `Validator` from #465 
- Add tests for each function with fake `peer_pool`s

#### TODO
- [x] Confirm which `slot` should be used when proposing and skipping. (Thanks for @NIC619 's help)
- ~Add receive server from #465~ (In favor of opening other PRs for "receive server" and "plugin config")
- ~Refactor tests: add clean-ups and `fixture`s~

#### Cute Animal Picture


mhchia:
Considering moving `testing_config.py` and `trinity/config.py` from #465  to solve both CI and mypy errors.

mhchia:
`ReceiveServer` will be included another PR. Except for this, I think the PR is ready for review.

tests/plugins/eth2/beacon/test_validator.py
```diff
@@ -0,0 +1,312 @@
...
+    NewSlotEvent,
+)
+
+
+helpers = importlib.import_module('tests.core.p2p-proto.bcc.helpers')
+
+
+NUM_VALIDATORS = 8
+
+privkeys = tuple(int.from_bytes(
```
ChihChengLiang: Have a crazy idea that we should move the plugin test from `tests/plugins/eth2/` to `tests/eth2/beacon/plugins/`, so that we can reuse all the fixtures defined in the `conftest.py` there. Not sure what do you think. cc @pipermerriam @hwwhww @mhchia

tests/plugins/eth2/beacon/test_validator.py
```diff
@@ -0,0 +1,312 @@
...
+    new_slot = state.slot + 1
+    # test: `new_slot` should call `propose_block` if the validator get selected,
+    #   else calls `skip_block`.
+    index = _get_proposer_index(
+        state,
+        new_slot,
+        state_machine.config,
+    )
+
+    is_proposing = True
```
NIC619: Should we set it to `None` so the assertion `assert is_proposing is not None`  below takes effect?

tests/plugins/eth2/beacon/test_validator.py
```diff
@@ -0,0 +1,312 @@
...
+    # test: either `propose_block` or `skip_block` should be called.
+    assert is_proposing is not None
+    if alice.validator_index == index:
+        assert is_proposing
+    else:
+        assert not is_proposing
+
+
+@pytest.mark.asyncio
+async def test_validator_handle_new_slot(caplog, event_loop, event_bus, monkeypatch):
```
NIC619: nit: IMO for readability sake `test_validator_handle_new_slot` should comes before `test_validator_new_slot`. We test if `new_slot` is triggered as expected and then we test if `new_slot` functions as expected.

tests/plugins/eth2/beacon/test_validator.py
```diff
@@ -0,0 +1,312 @@
...
+            slot=1,
+            elapsed_time=2,
+        ),
+        BroadcastConfig(internal=True),
+    )
+    await asyncio.wait_for(
+        event_new_slot_called.wait(),
+        timeout=2,
+        loop=event_loop,
+    )
```
NIC619: How do we determine if this test failed? Does it throw exception like `TimeoutError`?

trinity/config.py
```diff
@@ -609,31 +618,25 @@ def chain_name(self) -> str:
...
     @property
     def beacon_chain_class(self) -> Type['BeaconChain']:
-        from eth2.beacon.chains.base import BeaconChain  # noqa: F811
-        return BeaconChain.configure(
-            __name__=self.chain_name,
-            sm_configuration=self.sm_configuration,
-            chain_id=self.chain_id,
-        )
+        if self._beacon_chain_class is None:
+            self._beacon_chain_class = TestnetChain.configure(
```
NIC619: Can you add a TODO here that says we should be able to customize configs for `tests/` instead of using the configs from `TestnetChain`?

trinity/plugins/eth2/beacon/validator.py
```diff
@@ -0,0 +1,167 @@
...
+
+from trinity.endpoint import TrinityEventBusEndpoint
+from trinity.protocol.bcc.peer import (
+    BCCPeer,
+    BCCPeerPool,
+)
+from trinity.plugins.eth2.beacon.slot_ticker import (
+    NewSlotEvent,
+)
+from trinity._utils.shellart import (
```
NIC619: nit: `trinity._utils` should come before `trinity.endpoint`

trinity/plugins/eth2/beacon/validator.py
```diff
@@ -0,0 +1,167 @@
...
+    bold_green,
+    bold_red,
+)
+
+from p2p.service import BaseService
```
NIC619: nit: `p2p` should come before `trinity`

tests/plugins/eth2/beacon/test_validator.py
```diff
@@ -0,0 +1,312 @@
...
+class FakePeerPool:
+    def __init__(self):
+        self.connected_nodes = {}
+
+    def add_peer(self, index):
+        self.connected_nodes[index] = FakePeer()
+
+
+def get_chain(db):
+    return chain_class.from_genesis(
```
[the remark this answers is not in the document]
mhchia: Ah I think it might be better to rename it to `get_chain_from_genesis`.

tests/plugins/eth2/beacon/test_validator.py
mhchia: Thank you for pointing it out. It should be `is_proposing = None` originally. Have changed it.

tests/plugins/eth2/beacon/test_validator.py
mhchia: Yes, after 2 seconds `TimeoutError` is raised and makes the test fail.

trinity/config.py
mhchia: Done:)

tests/plugins/eth2/beacon/test_validator.py
```diff
@@ -0,0 +1,312 @@
...
+    root_post_state = alice.skip_block(
+        slot=slot,
+        state=state,
+        state_machine=state_machine,
+    )
+    # test: confirm that no block is imported at the slot
+    with pytest.raises(BlockNotFound):
+        alice.chain.get_canonical_block_by_slot(slot)
+    # test: the state root should change after skipping the block
+    assert state.root != root_post_state
```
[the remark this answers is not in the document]
mhchia: `root_post_state` is already the root of the post state.

trinity/plugins/eth2/beacon/validator.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class Validator(BaseService):
+    """
+    Reference: https://github.com/ethereum/trinity/blob/master/eth2/beacon/tools/builder/proposer.py#L175  # noqa: E501
```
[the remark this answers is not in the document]
mhchia: Ah it is a left-over. Have removed it.

tests/plugins/eth2/beacon/test_validator.py
mhchia: +1. IMO it will help for writing tests.

PR #528, opened 2019-04-23
Cleanup logic around connecting to a peer

pipermerriam:
### What was wrong?

There were a few things about how the `PeerPool` managed connecting to new peers that were able to be improved.

### How was it fixed?

- use an `asyncio.BoundedSemephore` to properly limit the number of concurrent attempts to connect to new peers.
- early exit for the case where a backend returns zero connection candidates.
- replace `while True` with `while self.is_operational` for peer connection loop.

#### Cute Animal Picture

Little_penguins


carver:
> Just being curious: Was there some kind of issue / bug where too many concurrent connection attempts caused trouble? Or is this more of a precaution because it feels safer?

Yeah, I'm curious too. This seems like background that belongs in a comment next to the constant. I'm guessing we are just stepping on our own toes (performance drops) if we try to connect to too many peers at once, but would be good to know for sure.

pipermerriam:
@cburgdorf @carver no driving reason other than the previous implementation was only implicitly limited and that bothered me.  In theory there should be no problem bumping this number up, but it felt much better having an explicit mechanism in place.

I agree that we should do some measuring and maybe opening this up or even removing the limit is the right choice but it isn't clear to me how we'd measure what a reasonable limit would be.  Current limit is...

- at most 10 concurrently
- up to "number-of-empty-slots" x number of backends per attempt to connect (batch)
- up to 5 batches in a burst
- rate limited at 1 batch every 2 seconds when rate limit is hit. 

I'll try to remember to measure how long on average an attempt takes to fail because that's the real rate limiter right now.

pipermerriam:
Local experimentation on my local node with an average internet connection on a fast machine shows this lock only engages maybe 1/50 connection attempts and never for longer than 1 seconds with most entries being around 0.25 seconds.

PR #531, opened 2019-04-23
Make kademlia.Node pickleable

pipermerriam:
### What was wrong?

The `kademlia.Node` class was not pickleable due to the embedded `eth_keys.PublicKey` object which contains a reference to it's backend which is implemented as a module (this should be eventually be fixed upstream to make the key backend implementations pickleable)

### How was it fixed?

Used the API that python uses for handling pickle-ing of stateful objects, using the enode URI as the data transfer representation and then re-constructing it on the other side.

- [x] need to update the various places where @cburgdorf implemented workarounds for this issue to just use the `Node` classes now that they work.

#### Cute Animal Picture

2F93D1CA00000578-0-image-a-10_1450825493645

### DIFF
--- a/p2p/events.py
+++ b/p2p/events.py
@@ -8,17 +8,16 @@ from lahja import (
     BaseRequestResponseEvent,
 )
 
+from .kademlia import Node
+
 
 class BaseDiscoveryServiceResponse(BaseEvent):
-
-    def __init__(self, error: Exception) -> None:
-        self.error = error
+    pass
 
 
 class PeerCandidatesResponse(BaseDiscoveryServiceResponse):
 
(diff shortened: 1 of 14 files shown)


p2p/events.py
```diff
@@ -8,17 +8,16 @@
     BaseRequestResponseEvent,
 )
 
+from .kademlia import Node
 
-class BaseDiscoveryServiceResponse(BaseEvent):
 
-    def __init__(self, error: Exception) -> None:
-        self.error = error
```
pipermerriam: Took liberty of removing this API as it appears to not be used anywhere.

PR #532, opened 2019-04-24
Add fork choice including MVP LMD GHOST

### LINKED ISSUE #260: Implement fork choice rule
### What is wrong?

Implement LMD-GHOST
Ref:
- https://github.com/ethereum/research/blob/master/ghost/ghost.py
- https://github.com/protolambda/lmd-ghost/

### How can it be fixed

The current STUB fork choice rule is only scoring by block slot number: 
https://github.com/ethereum/trinity/blob/master/eth2/beacon/db/chain.py#L374
We can modify it to an abstract switchable fork choice rule.

ralexstokes:
### What was wrong?

Fixes #260.

### How was it fixed?

Edit: updating the previous text to reflect a more concrete direction on a mergeable PR.

Introduce a way for clients of a chain to provide their own fork choice mechanism.

Migrates the current "higher slot" fork choice to the new methodology and introduces a very minimal LMD GHOST fork choice.

### TODO
- [x] migrate current ""highest slot"" fork choice to use the new `fork_choice` module
- [ ] finish writing Store class for LMDGHOST fork choice
- [ ] add some basic tests for LMDGHOST scoring
- [ ] install LMDGHOST fork choice scoring in BeaconChainDB
- [ ] modify fork choice selection as needed to fully implement LMD GHOST algorithm
- [ ] any dangling linter fixes

### Notes

Because it has come up a few times in various conversations, I'll make a note to consider moving the fork choice out into the "vm/state machine" layer; it currently resides in the DB layer... I'm leaning towards doing this as otherwise the `DB` class, which is already the definition of a mutable artifact, becomes an even bigger "ball of mud" that happens to (increasingly) know about every other part of the stack. this outcome happens if we keep the fork choice inside the DB given the more complicated nature of the beacon chain fork choice w/ LMD GHOST. we should be able to counter this tendency by moving the fork choice process out into the `StateMachine`

#### Cute Animal Picture


ralexstokes:
continuing conversation here: https://github.com/ethereum/trinity/pull/547#pullrequestreview-233589449

eth2/beacon/chains/base.py
```diff
@@ -391,6 +391,11 @@ def import_block(
         # TODO: Now it just persists all state. Should design how to clean up the old state.
         self.chaindb.persist_state(state)
 
+        # TODO write just the state and block here
```
ralexstokes: @hwwhww this is what i'm thinking to refactor the fork choice.... it will be hoisted out of the DB layer and encapsulated into the `StateMachine`...... as @carver pointed out, it can change via `Slot` so is an attribute of our slot-changing thing :)

we can use the state machine's fork choice to find the new head and persist that data as well to the DB.

refer to the comments here for a high-level flow. would be great to get your thoughts on this as I know you were thinking about it in the earlier PR.

the only con i see is that we will have to expose some methods on the DB object that can be considered dangerous... right now, all you can really do is just import a block which constrains which things can become finalized/justified. these invariants will live across a larger surface if we move ahead w/ the refactor so its worth considering its value...

i plan to keep moving ahead w/ this approach for the time being; wanted to see if i can get your thoughts on this sooner rather than later :)

ralexstokes:
FYI this is turning out to be scratch space to sketch out some ideas -- will close once all the good stuff has been ported to other branches

PR #534, opened 2019-04-24
[WIP] move connection tracking to a plugin

pipermerriam:
### How was it fixed?

#### Cute Animal Picture

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -68,7 +68,7 @@ from p2p.events import (
     PeerCandidatesResponse,
 )
 from p2p.exceptions import AlreadyWaitingDiscoveryResponse, NoEligibleNodes, UnableToGetDiscV5Ticket
-from p2p.kademlia import to_uris, Node as KademliaNode
+from p2p.kademlia import Node as KademliaNode
 from p2p import kademlia
 from p2p import protocol
 from p2p.service import BaseService
@@ -439,9 +439,6 @@ class DiscoveryProtocol(asyncio.DatagramProtocol):
             self.receive_v5(address, cast(bytes, data))
         else:
             self.receive(address, cast(bytes, data))
-
-    def error_received(self, exc: Exception) -> None:
-        self.logger.error('error received: %s', exc)
 
     def send(self, node: kademlia.Node, message: bytes) -> None:
         self.transport.sendto(message, (node.address.ip, node.address.udp_port))
@@ -1002,9 +999,8 @@ class StaticDiscoveryService(BaseService):
             self,
             event: BaseRequestResponseEvent[PeerCandidatesResponse],
             nodes: Tuple[KademliaNode, ...]) -> None:
-        node_uris = tuple(to_uris(nodes))
         self._event_bus.broadcast(
-            event.expected_response_type()(node_uris),
+            event.expected_response_type()(nodes),
             event.broadcast_config()
         )
 
@@ -1067,7 +1063,7 @@ class DiscoveryService(BaseService):
 
             self.run_task(self.maybe_lookup_random_node())
 
-            nodes = tuple(to_uris(self.proto.get_nodes_to_connect(event.max_candidates)))
+            nodes = tuple(self.proto.get_nodes_to_connect(event.max_candidates))
 
             self.logger.debug2("Broadcasting peer candidates (%s)", nodes)
             self._event_bus.broadcast(
@@ -1078,7 +1074,7 @@ class DiscoveryService(BaseService):
     async def handle_get_random_bootnode_requests(self) -> None:
         async for event in self.wait_iter(self._event_bus.stream(RandomBootnodeRequest)):
 
-            nodes = tuple(to_uris(self.proto.get_random_bootnode()))
+            nodes = tuple(self.proto.get_random_bootnode())
 
             self.logger.debug2("Broadcasting random boot nodes (%s)", nodes)
             self._event_bus.broadcast(
--- a/p2p/events.py
+++ b/p2p/events.py
@@ -8,17 +8,16 @@ from lahja import (
     BaseRequestResponseEvent,
 )
 
+from .kademlia import Node
+
 
 class BaseDiscoveryServiceResponse(BaseEvent):
-
-    def __init__(self, error: Exception) -> None:
-        self.error = error
+    pass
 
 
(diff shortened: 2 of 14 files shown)


tox.ini
```diff
@@ -134,7 +134,7 @@ commands=
     flake8 {toxinidir}/eth2
     flake8 --exclude={toxinidir}/libp2p/p2pclient/pb {toxinidir}/libp2p
     # TODO: Drop --ignore-missing-imports once we have type annotations for eth_utils, coincurve and cytoolz
-    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p p2p -p trinity -p eth2 -p libp2p
+    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p p2p -p trinity -p eth2 -p libp2p --config-file {toxinidir}/mypy.ini
```
cburgdorf: Not important for this PR (and maybe I just go ahead and do it) but it looks like we could collapse the entire lengthy invocation into the `mypy.ini` so that this becomes just `mypy --config-file {toxinidir}/mypy.ini`.

trinity/config.py
```diff
@@ -248,12 +249,14 @@ def __init__(self,
                  port: int=30303,
                  use_discv5: bool = False,
                  preferred_nodes: Tuple[KademliaNode, ...]=None,
-                 bootstrap_nodes: Tuple[KademliaNode, ...]=None) -> None:
+                 bootstrap_nodes: Tuple[KademliaNode, ...]=None,
+                 tracking_backend: TrackingBackend=None) -> None:
```
cburgdorf: I feel that we should keep the `TrinityConfig` mostly free from stuff that comes from plugins. It looks like the plugin itself is the only place that uses `config.tracking_backend` anyway so maybe we don't really need to add it on the `TrinityConfig`?

trinity/constants.py
```diff
@@ -81,3 +83,9 @@
              Address("34.198.237.7", 30303, 30303)),
     ),
 }
+
+
+class TrackingBackend(enum.Enum):
```
cburgdorf: I also think this should live somewhere under the plugin space.

trinity/plugins/builtin/blacklist/plugin.py
```diff
@@ -0,0 +1,98 @@
...
+            choices=('sqlite3', 'memory', 'disabled'),
+            default=TrackingBackend.sqlite3,
+            type=str,
+        )
+
+        attach_parser = subparser.add_parser(
+            'remove-enode-db',
+            help='Remove the on-disk sqlite database that tracks data about node connections',
+        )
+        attach_parser.set_defaults(func=clear_node_db)
```
cburgdorf: The purist in me would make this an entire own plugin (probably in the same file though) that derives from `BaseMainProcessPlugin` and just handles the clearing. No strong feeling though.

trinity/plugins/builtin/blacklist/plugin.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def on_ready(self, manager_eventbus: TrinityEventBusEndpoint) -> None:
+        self.start()
```
cburgdorf: Maybe consider adding a `--disable-tracking-backend` and don't start if its present.  Not sure how useful it would be in this concrete case but I quite like the flexibility of being able to disable each plugin on a case by case basis (even if this means things may not work or unexpected errors are exposed)

trinity/plugins/builtin/blacklist/plugin.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _get_tracker(self) -> BaseConnectionTracker:
+        config = self.context.trinity_config
+
+        if config.tracking_backend is TrackingBackend.sqlite3:
```
cburgdorf: I think we can just read and normalize from `self.context.args` and keep `TrinityConfig` free from the `TrackingBackend`.

trinity/plugins/builtin/blacklist/server.py
```diff
@@ -0,0 +1,58 @@
...
+            )
+            self.tracker.record_blacklist(command.remote, command.timeout, command.reason)
+
+    async def _run(self) -> None:
+        self.logger.debug("Running BlacklistServer")
+
+        self.run_daemon_task(self.handle_should_connect_to_requests())
+        self.run_daemon_task(self.handle_blacklist_command())
+
+        await self.cancel_token.wait()
```
cburgdorf: Nitpick, you can do `await self.cancellation()`

trinity/plugins/registry.py
```diff
@@ -57,6 +60,7 @@ def is_ipython_available() -> bool:
 
 
 ETH1_NODE_PLUGINS: Tuple[BasePlugin, ...] = (
+    BlacklistPlugin(),
```
cburgdorf: Especially if this is considered to be Eth1 only, then I feel more strongly it should not pollute the `TrinityConfig`.

trinity/config.py
pipermerriam: How would the plugin learn about this value then?

pipermerriam: Nevermind, I see that we have access via `context.args`

trinity/plugins/builtin/blacklist/plugin.py
pipermerriam: This one feels weird (YAGNI) but I'm willing to be convinced.  Disabling it will cause the `EthPeerPool` to not be able to establish peer connections so this case is one of those where not only will things not work unexpectedly but they will **just not work**

pipermerriam:
did not intend to merge... fixing

trinity/plugins/builtin/blacklist/plugin.py
cburgdorf: Imagine someone else trying to build a replacement for this plugin which is Event-compatible. So you would need to run `trinity --disable-tracking-backend --enable-experimental-new-tracking-backend` to not have both systems interfere. I'll admit, this sounds far-fetched but I happen to like the flexibility of it. Feel free to ignore :)

pipermerriam: Good enough for me.

Issue #24, opened 2018-05-23
Maintain updated pool of recommended peers


cburgdorf:
### What is wrong?

Quick brain dump from https://github.com/ethereum/py-evm/pull/744#discussion_r190291371

Trinity currently uses a hardcoded list of recommended peers + discovery protocol. It would be nice to maintain a list of recommended peers based on the following metrics that also survives restarts (persisted in db / config file).

Metrics:

* latency
* correct responses (doesn't send unexpected data)
* num success/failed requests

### How can it be fixed

Initial idea is to have some sort of `PeerSubscriber` API which would be given a view into peer requests and responses and then to aggregate that data into some accessible form. Then the a `FancyPeerPool` class could use this data to inform what nodes/peers it returns.

carver:
This should use the new sqlite database, presumably. After a long, slow ramp-up period of collecting peers, it would be nice to attempt to reconnect to those peers right away after rebooting `trinity`.

carver:
FWIW, this should probably not add inbound peers as preferred, since it would "convert" an inbound peer to an outbound peer on restart, opening another eclipse attack avenue.

---

If we wanted to be really fancy, we could store inbound nodes in a separate list. Then we would only connect to them after we have enough outbound peers and "pretend" they were inbound for all analytical purposes. The converse of this mechanism could also be helpful: if we found a peer in the past and they re-dial us, we could treat them as outbound still for eclipse purposes. This is all potentially hairy and error prone, so probably belongs in a separate low-priority issue.

pipermerriam:
I think we should only populate the database with outbound peers, and inbound peers are strictly excluded.

As for treating a known outbound peer that dials inbound as outbound, I think we should avoid conflating these two concepts.  Inbound/Outbound connection labeling should probably be purely based on which side the connection was initiated from.  For tracking purposes, we can have meta information like `first_connection_direction: <dial-in/dial-out>` to track the origin of how we learned about the peer.

veox:
In issue #501 I argue that even having "default preferred peers" goes against the principle of least surprise.

If a node operator doesn't specify any preferred nodes, than no known peers should have elevated status.

PR #536, opened 2019-04-24
Move connection tracking to a plugin

pipermerriam:
Continuation of https://github.com/ethereum/trinity/pull/534 that got borked

### What was wrong?

The existing implementation for tracking information about failed attempts to connect to peers was ready for a few upgrades as part of the work towards #24 

- The use of stdlib `sqlite` module was good for MVP but not ideal for a more complex implementation.
- The `p2p` module is intended to not be oppinionated about as much as possible and thus limit unnecessary 3rd party dependencies.  Adding `sqlalchemy` felt like such a case.
- Passing around a *handle* to the tracker was cumbersome, requiring it to pass across/through multiple classes in order to let the `PeerPool` get access to it.

### How was it fixed?

- The SQLite based backends are now implemented under the `trinity` module.
- The database interactions are now handled by `sqlalchemy`
- A new plugin `Network Database` is responsible for managing the on disk database.
- A new plugin `Connection Blacklist Database` is responsible for tracking connection failures over an event bus based API

#### Cute Animal Picture

adorable-baby-penguins-11


p2p/peer_pool.py
```diff
@@ -298,7 +304,17 @@ def _add_peer(self,
...
             raise IneligiblePeer(f"Already connected to {remote}")
-        if not self.peer_info.should_connect_to(remote):
+
+        try:
+            should_connect = await self.wait(
+                self.connection_tracker.coro_should_connect_to(remote),
+                timeout=1,
+            )
+        except TimeoutError:
+            self.logger.warning("ConnectionTracker.coro_should_connect_to request timed out.")
```
pipermerriam: Decided to start out heavy handed here and dump warnings because this is indicative of something being wrong and if this is continually timing out then all attempts to connect to a peer will fail.

trinity/plugins/builtin/network_db/plugin.py
```diff
@@ -0,0 +1,65 @@
...
+    def name(self) -> str:
+        return "Network Database"
+
+    @property
+    def normalized_name(self) -> str:
+        return "network-db"
+
+    def configure_parser(self, arg_parser: ArgumentParser, subparser: _SubParsersAction) -> None:
+        tracking_parser = arg_parser.add_argument_group('network db')
+        tracking_parser.add_argument(
```
cburgdorf: I think the `tracking_parser` accidentally moved here. This seems to belong to the other plugin (`BlacklistPlugin`).

pipermerriam: This was intentional because this flag will effect the other trackers (which are already made but I don't have PR's open yet for them).  The idea is that all of the network-db related trackers operate on the same underlying database and you only configure that at a macro level rather than at the individual tracker level.  From the user perspective they are all the same thing, but they are split between different plugins to keep them organized.

cburgdorf: Gotcha!

pipermerriam:
CI failure appears to be real but I don't understand it yet.  It does reproduce locally and doesn't appear to be intermittent.

trinity/plugins/builtin/network_db/plugin.py
```diff
@@ -0,0 +1,65 @@
...
+)
+
+from .backends import TrackingBackend
+from .cli import NormalizeTrackingBackend
+
+
+class NetworkDBPlugin(BaseMainProcessPlugin):
+    # we access this logger from a classmethod so it needs to be available on
+    # the class (instead of as a computed property like the base class
+    # prodived.
```
lithp: `s/prodived/provided/`

cburgdorf:
@pipermerriam regarding CI fail try picking https://github.com/cburgdorf/trinity/commit/5dc02f5071c24aee995eecaf825d27669709c050 and maybe also https://github.com/cburgdorf/trinity/commit/d33ac6d05fbcf68cc1d5bf59ac73f2634c4307d5 into your branch. I think they will fix it.

p2p/tracking/connection.py
```diff
@@ -0,0 +1,68 @@
...
+
+from p2p.kademlia import Node
+from p2p.exceptions import (
+    BaseP2PError,
+    HandshakeFailure,
+    TooManyPeersFailure,
+)
+
+
+ONE_DAY = 60 * 60 * 24
```
lithp: `ONE_DAY` appears to be unused

p2p/tracking/connection.py
```diff
@@ -0,0 +1,68 @@
...
+register_error(HandshakeFailure, 10)  # 10 seconds
+register_error(TooManyPeersFailure, 60)  # one minute
+
+
+def get_timeout_for_failure(failure: BaseP2PError) -> int:
+    for cls in type(failure).__mro__:
+        if cls in FAILURE_TIMEOUTS:
+            return FAILURE_TIMEOUTS[cls]
+    failure_name = type(failure).__name__
+    raise Exception(f'Unknown failure type: {failure_name}')
```
lithp: This exception is exactly the kind of thing type systems do! The problem is that `FAILURE_TIMEOUTS.keys() ⊂ BaseP2PError`, the type signature promises too much. I think these last two lines could be dropped by adding a `register_error(BaseP2PError, 60)`.

lithp: Okay, this was harder than I expected, but I have a prototype of what it would look like to lean into the type system a little more: https://github.com/pipermerriam/trinity/pull/1

lithp: At first I thought this would require dependent types, but then I remembered python has multiple inheritance! `BlacklistError ⊂ BaseP2PError` but not a subclass of it, since they're not really the same concept: "An error which occured while dealing with a peer" vs "An error which means we should blacklist a peer".

`mypy` rejects calls which accidentally use a `BaseP2PError` instead of a `BlacklistError`, the metadata is stored with the original error. ABC doesn't play nicely with multiple inheritance, unfortunately, so I ended up overriding `__new__` myself :cry:

trinity/plugins/builtin/network_db/blacklist/tracker.py
```diff
@@ -0,0 +1,167 @@
...
+        self.config = config
+
+    def record_blacklist(self, remote: Node, timeout: int, reason: str) -> None:
+        self.event_bus.broadcast(
+            BlacklistEvent(remote, timeout, reason=reason),
+            self.config,
+        )
+
+    def should_connect_to(self, remote: Node) -> bool:
+        raise NotImplementedError("Must use coro_should_connect_to")
```
lithp: This is a bit strange, this tracker implements an abstract method but... it doesn't really implement it! This class isn't really a subclass of `BaseConnectionTracker`, it's not usable everywhere a `BaseConnectionTracker` would be, and that could lead to confusion later.

It looks like `should_connect_to` is (outside of tests) only called once, and (including tests) it's always called from inside an async function. What do you think of dropping `should_connect_to` and only having `coro_should_connect_to`?

tests/core/database/test_tracking_database_management.py
```diff
@@ -0,0 +1,173 @@
+import asyncio
+from pathlib import Path
+import uuid
+
+import pytest
+
+from p2p.exceptions import BadDatabaseError
```
lithp: Now that all the database code has been taken out of `p2p` I think it might make sense to also move this error into `trinity`. It looks like this is only consumed by `trinity.db.orm`, maybe it could be moved there, or into `trinity.exceptions`?

trinity/db/orm.py
```diff
@@ -0,0 +1,113 @@
...
+    __tablename__ = 'schema_version'
+
+    id = Column(Integer, primary_key=True)
+    version = Column(String, unique=True, nullable=False, index=True)
+
+
+#
+# SQL Based Trackers
+#
+def _get_session(path: Path) -> 'BaseSession':
```
lithp: the nittiest of nitpicks but I don't think this needs to be quoted

tests/core/database/test_tracking_database_management.py
```diff
@@ -0,0 +1,173 @@
...
+    assert _check_schema_version(session) is True
+
+    # change version to unknown value
+    schema_version = session.query(SchemaVersion).one()
+    schema_version.version = 'unknown'
+
+    session.add(schema_version)
+    session.commit()
+
+    assert _check_schema_version(session) is False
```
lithp: Neat, these tests are all great! This is a lot easier to read and write than the raw SQL would have been.

trinity/db/orm.py
```diff
@@ -0,0 +1,113 @@
...
+
+def _setup_schema(session: BaseSession) -> None:
+    Base.metadata.create_all(session.get_bind())
+    session.add(SchemaVersion(version=SCHEMA_VERSION))
+    session.commit()
+
+
+def _get_schema_version(session: BaseSession) -> str:
+    schema_version = session.query(SchemaVersion).one()
+    return schema_version.version
```
lithp: nitpick: I know the underscore means this is private, but there's a tiny chance of making a mistake here if someone confused it with `_check_schema_version` and didn't realize they needed to check for `NoResultFound`. What do you think of inlining this into `_check_schema_version`?

tests/core/tracking/test_blacklist_connection_tracking.py
```diff
@@ -0,0 +1,79 @@
...
+    SQLiteConnectionTracker,
+    MemoryConnectionTracker,
+)
+from trinity.db.orm import (
+    get_tracking_database,
+)
+
+
+def test_records_failures():
+    # where can you get a random pubkey from?
```
lithp: I think this comment is a leftover from development?

tests/core/tracking/test_blacklist_connection_tracking.py
```diff
@@ -0,0 +1,79 @@
...
+    # open a second instance
+    connection_tracker_b = MemoryConnectionTracker()
+
+    # the second instance has no memory of the failure
+    assert connection_tracker_b.should_connect_to(node) is True
+    assert connection_tracker_a.should_connect_to(node) is False
+
+
+def test_sql_does_persist(tmpdir):
+    db_path = Path(str(tmpdir.join("nodedb")))
```
lithp: nitpick: the `str` doesn't appear to be necessary here, or in `tests/core/database/test_tracking_database_management.py:31`

tests/core/tracking/test_blacklist_connection_tracking.py
```diff
@@ -0,0 +1,79 @@
...
+    assert connection_tracker_a.should_connect_to(node) is False
+    del connection_tracker_a
+
+    # open a second instance
+    connection_tracker_b = SQLiteConnectionTracker(get_tracking_database(db_path))
+    # the second instance remembers the failure
+    assert connection_tracker_b.should_connect_to(node) is False
+
+
+def test_timeout_works(monkeypatch):
```
lithp: `monkeypatch` appears to be unused, I'm surprised flake8 didn't catch this! Maybe it special-cases the tests, since invoking a fixture could legitimately change behavior? :thinking:

tests/core/tracking/test_blacklist_server.py
```diff
@@ -0,0 +1,46 @@
...
+    tracker = MemoryConnectionTracker()
+    remote_a = NodeFactory()
+    tracker.record_blacklist(remote_a, 60, "testing")
+
+    assert tracker.should_connect_to(remote_a) is False
+
+    service = BlacklistServer(event_bus, tracker)
+
+    # start the server
+    asyncio.ensure_future(service.run(), loop=event_loop)
```
lithp: Nothing to do with this PR! but:

This pattern of spawn and wait until it's running is something I wanted recently when working in lahja, lahja doesn't have Services so this wouldn't be helpful there but I think a helper could come in handy in trinity:

```python
await service.start()
```

tests/core/tracking/test_blacklist_server.py
```diff
@@ -0,0 +1,46 @@
...
+import pytest
+
+from lahja import BroadcastConfig
+
+from p2p.tools.factories import NodeFactory
+
+from trinity.constants import (
+    NETWORKING_EVENTBUS_ENDPOINT,
+)
+from trinity.plugins.builtin.network_db.blacklist.server import BlacklistServer
```
lithp: I'm not sure what my suggestion is, but this PR is adding a lot of names for the same thing. This adds:
- The network_db plugin
- The blacklist server running in that plugin
- connection tracking via ConnectionTracker

To someone browsing the codebase it's not going to be obvious that these are three layers of the same thing. I'm not sure what to suggest since I get the impression that this would be worse:
- `plugins/builtin/peer_db`, "Peer Database Plugin"
- `PeerDatabaseServer`, maybe in plugins.builtin.peer_db.server
- `SQLitePeerDatabase`

`BlacklistServer` is the name that I most think doesn't have to be it's own thing, but my only ideas are ugly compounds like `ConnectionTrackerEndpointAdapter`.

lithp: Actually, `ConnectionTrackerServer` has a nice ring, especially if `EventBusConnectionTracker` is turned into `ConnectionTrackerClient`. I don't know, just some ideas.

tests/core/tracking/test_blacklist_server.py
```diff
@@ -0,0 +1,46 @@
...
+    service = BlacklistServer(event_bus, tracker)
+
+    # start the server
+    asyncio.ensure_future(service.run(), loop=event_loop)
+    await service.events.started.wait()
+
+    config = BroadcastConfig(filter_endpoint=NETWORKING_EVENTBUS_ENDPOINT)
+    bus_tracker = EventBusConnectionTracker(event_bus, config=config)
+
+    # ensure we can read fromt he tracker over the event bus
```
lithp: s/fromt he/from the/

tests/p2p/test_connection_tracker_timeout_for_failure.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@pytest.fixture(autouse=True)
+def _prevent_global_mutation_of_registry(monkeypatch):
+    # Ensure that the tests don't result in mutation of the configuration of
+    # timeout failures.
+    from p2p.tracking import connection
+    original = copy.copy(connection.FAILURE_TIMEOUTS)
+    yield
+    keys_to_pop = set(connection.FAILURE_TIMEOUTS.keys()).difference(original.keys())
```
lithp: Why not `connection.FAILURE_TIMEOUTS = original`?

tests/p2p/test_connection_tracker_timeout_for_failure.py
```diff
@@ -0,0 +1,50 @@
...
+    HandshakeFailure,
+)
+from p2p.tracking.connection import (
+    get_timeout_for_failure,
+    register_error,
+)
+
+
+@pytest.fixture(autouse=True)
+def _prevent_global_mutation_of_registry(monkeypatch):
```
lithp: another orphaned `monkeypatch` :)

trinity/db/network.py
```diff
@@ -0,0 +1,12 @@
+from pathlib import Path
+
+from trinity.config import (
+    TrinityConfig,
+)
+
+
+def get_networkdb_path(config: TrinityConfig) -> Path:
```
lithp: What do you think of keeping it in the data dir, instead of nested away in a single-file directory? Doesn't really make a difference either way but when one of us tries to inspect the database this will make our job of navigating to it ever so slightly easier.

trinity/db/orm.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def _check_is_empty(session: BaseSession) -> bool:
+    engine = session.get_bind()
+    for table_name in Base.metadata.tables.keys():
```
lithp: I can't decide if this is dangerous or not but it makes me a little nervous. This only finds `BlacklistRecord` because `trinity.__init__` imports `main` which eventually imports the plugins. If any part of that chain is broken (or if someone later adds a model in a file which isn't automatically imported) then this method (and `_check_tables_exist`) will silently fail to look for the missing tables. I think this is unlikely but it's a weird enough error that if it does happen it'll take quite a while to debug, so it might be worth fixing now.

I think there are two options:
- bringing `BlacklistRecord` into this file, or some nearby file which is obviously where new models should be added, in much the same way `trinity.exceptions` is obviously where new exceptions should be added.
- Instead of looking for tables in the base, looking for tables in the engine. `_check_is_empty` becomes `len(session.bind.table_names()) == 0`. I'm not sure what to do about `_check_tables_exist` because you really want to look for *our* tables, so maybe the above answer is the better one.

trinity/db/orm.py
```diff
@@ -0,0 +1,113 @@
...
+            return False
+    return True
+
+
+def get_tracking_database(db_path: Path) -> BaseSession:
+    session = _get_session(db_path)
+
+    if _check_schema_version(session):
+        if not _check_tables_exist(session):
+            raise BadDatabaseError(f"Tracking database is missing tables: {db_path.resolve()}")
```
lithp: For future confused users, this might be a good place to drop a comment mentioning `remove-network-db`

trinity/db/orm.py
```diff
@@ -0,0 +1,113 @@
...
+        return session
+    elif not _check_is_empty(session):
+        raise BadDatabaseError(f"Tracking database not empty: {db_path.resolve()}")
```
lithp: *nitpick*: We want our database to not be empty! Maybe "Tracking database has tables but no schema version"

trinity/db/orm.py
```diff
@@ -0,0 +1,113 @@
...
+
+    _setup_schema(session)
+    return session
```
lithp: I played with this function a little to try to flatten the conditionals:

```python
def get_tracking_database(db_path: Path) -> BaseSession:
    session = _get_session(db_path)

    if _check_is_empty(session):
        _setup_schema(session)
        return session

    if not _check_schema_version(session):
        raise BadDatabaseError(
            f"Tracking database has wrong or missing version: {db_path.resolve()}"
        )

    if not _check_tables_exist(session):
        raise BadDatabaseError(f"Tracking database is missing tables: {db_path.resolve()}")

    return session
```

trinity/plugins/builtin/network_db/backends.py
```diff
@@ -0,0 +1,7 @@
+import enum
```
lithp: This is a selfish request :) but big files are a lot easier to navigate in vim than switching between a lot of small files. What do you think about moving this into `network_db.cli`?

trinity/plugins/builtin/network_db/blacklist/events.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class BaseBlacklistEvent(BaseEvent):
+    pass
+
+
+class BlacklistEvent(BaseBlacklistEvent):
+    def __init__(self, remote: Node, timeout: int, reason: str) -> None:
+        self.remote = remote
+        self.timeout = timeout
```
lithp: nitpick: This would be slightly more self-documenting if it was called `timeout_seconds`

trinity/plugins/builtin/network_db/blacklist/server.py
```diff
@@ -0,0 +1,58 @@
...
+        self.event_bus = event_bus
+
+    async def handle_should_connect_to_requests(self) -> None:
+        async for req in self.wait_iter(self.event_bus.stream(ShouldConnectToPeerRequest)):
+            self.logger.debug2('Received should connect to request: %s', req.remote)
+            should_connect = self.tracker.should_connect_to(req.remote)
+            self.event_bus.broadcast(
+                ShouldConnectToPeerResponse(should_connect),
+                req.broadcast_config()
+            )
```
lithp: Opened [a ticket against lahja](https://github.com/ethereum/lahja/issues/45), this is pretty formulaic and it'd be nice if lahja could somehow make this prettier.

trinity/plugins/builtin/network_db/blacklist/tracker.py
```diff
@@ -0,0 +1,167 @@
...
+    __tablename__ = 'blacklist_records'
+
+    id = Column(Integer, primary_key=True)
+    uri = Column(String, unique=True, nullable=False, index=True)
+    expires_at = Column(DateTime(timezone=True), nullable=False)
+    reason = Column(String, nullable=False)
+    error_count = Column(Integer, default=1, nullable=False)
+
+    def is_expired(self) -> bool:
+        return datetime.datetime.utcnow() < self.expires_at
```
lithp: this function is unused

trinity/plugins/builtin/network_db/blacklist/tracker.py
```diff
@@ -0,0 +1,167 @@
...
+from .events import (
+    BlacklistEvent,
+    ShouldConnectToPeerRequest,
+)
+
+
+class BlacklistRecord(Base):
+    __tablename__ = 'blacklist_records'
+
+    id = Column(Integer, primary_key=True)
```
lithp: nitpick: my personal preference is to use natural keys whenever I can. For very large or performance sensitive databases int primary keys are important, they're much easier to lookup and join against than string keys. But for something like this I think it'd be simpler for `uri` to be the primary key.

lithp: For example:

```python
return self.session.query(BlacklistRecord).filter_by(uri=uri).one()
```

becomes:

```python
return self.session.query(BlacklistRecord).get(uri)
```

trinity/plugins/builtin/network_db/cli.py
```diff
@@ -0,0 +1,27 @@
...
+        except TypeError as err:
+            raise argparse.ArgumentError(
+                self,
+                (
+                    "Unknown option for --network-tracking-backend.  Must be one of "
+                    "`sqlite3/memory/disabled`.  Got '{value}'"
+                ),
+            )
+
+        setattr(namespace, self.dest, tracking_backend)
```
lithp: This line was very weird for me. There's no way to know that it sets `args.network_tracking_backend` unless you already know how argparse works. Maybe it could hard-code `network_tracking_backend` here?

trinity/plugins/builtin/network_db/plugin.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def _get_database_session(self) -> Session:
+        if self._session is None:
+            self._session = get_tracking_database(get_networkdb_path(self.context.trinity_config))
+        return self._session
+
+    def _get_blacklist_tracker(self) -> BaseConnectionTracker:
+        backend = self.context.args.network_tracking_backend
+
+        if backend is TrackingBackend.sqlite3:
```
lithp: One last thing re `network_db.cli` and `network_db.backends`. That's two extra files and 34 extra lines but I don't think these lines would be any uglier of they were `if backend is "sqlite3"`, `if backend is "memory"`. Argparse is already limiting the choices, I think a `dest=network_tracking_backend` is all it takes to make `_get_blacklist_tracker` happy.

tox.ini
```diff
@@ -125,7 +125,7 @@ commands = {[libp2p]commands}
 
 [common-lint]
 deps = .[p2p,trinity,lint,eth2,libp2p]
-setenv=MYPYPATH={toxinidir}:{toxinidir}/stubs
+setenv=MYPYPATH={toxinidir}
```
lithp: Why did this change?

tests/core/tracking/test_blacklist_connection_tracking.py
pipermerriam: I don't think `flake8` warns about un-used function arguments, only un-used variables.  still, will remove.

tests/p2p/test_connection_tracker_timeout_for_failure.py
pipermerriam: Because the functions where it is used already have a reference to the original object.  Doing `connection.FAILURE_TIMEOUTS = original` will put a new object in it's place but won't update anything that has a reference to the original one.  It's ugly globals code....

trinity/db/network.py
pipermerriam: It isn't nested away.  The end result is a file at `./trinity/mainnet/networkdb-eth1.sqlite3`

trinity/db/orm.py
pipermerriam: I also went ahead and namespaced the database filename with the schema version since it's probably not a very big deal to leave around old ones for the time being, and, it allows for switching between versions without having to clobber your database repeatedly.

trinity/db/orm.py
pipermerriam: I had the same issue.  Now that this lives in Trinity, I think I'm more comfortable just making this code be very explicity about which models it expects to be in place.

trinity/plugins/builtin/network_db/blacklist/tracker.py
pipermerriam: I did this originally, and in general I've been in agreement about this conceptually.  This is actually how I originally implemented it and I ended up having to fight both sqlalchemy and sqlite a bit to do it **and** it made the foreign key relationships harder to setup.  This ended up being the path of least resistance.

trinity/plugins/builtin/network_db/cli.py
pipermerriam: This is standard for how the argparse `Action` classes are implemented to allow them to be re-used.

https://github.com/ethereum/trinity/blob/master/trinity/cli_parser.py#L37

You can see this is the same for all of our other custom actions.

trinity/plugins/builtin/network_db/plugin.py
pipermerriam: If we move to strings then 

- the CLI parser has to validate the value is in the known set of allowed values (enum does this automatically for us)
- we have to have a data structure that enumerates the known values to implement this validation (current this is the enum class)

Unless you feel strongly I'm inclined to keep it as-is as I like how much an Enum restricts freedom of movement and provides strong guarantees about the value and type information.

tox.ini
pipermerriam: stubs no longer exists, just leftover cruft.  I actually think we could drop this line entirely, will try.

trinity/plugins/builtin/network_db/plugin.py
lithp: Ah, I admit that this is a part I didn't play with to check that it worked before I commented, but I thought that `choices=('sqlite3', 'memory', 'do-not-track'),` was already limiting which options the user had.

I don't feel strongly about it, I have this urge to reduce line count whenever it looks like there's an alternative which does roughly the same thing with less machinery. I think that doing so makes the code easier to read and modify but I think the urge is less rational than I'm describing it here and this code works and isn't *complicated* by any means.

The part I'm most worried about is that `network_tracking_backend` comes out of nowhere. Someone trying to figure out how it's set won't be able to grep for it and find that `setattr`, unless they already know about Argparse and also trinity's conventions.

trinity/db/network.py
lithp: Huh, good point. Last night I was sure it was doing `networkdb/networkdb.sqlite3`

trinity/plugins/builtin/network_db/plugin.py
carver: > The part I'm most worried about is that `network_tracking_backend` comes out of nowhere. Someone trying to figure out how it's set won't be able to grep for it and find that `setattr`, unless they already know about Argparse and also trinity's conventions.

This is my least favorite part about argparse (or at least how we're using it?) so far. Maybe belongs in an issue to see if/how we can improve it, but probably not as a tweak to this PR.

pipermerriam:
libp2p failure unrelated.

PR #538, opened 2019-04-25
Endpoint.broadcast() is now async

lithp:
### What was wrong?

Draft of how big of a refactor it would be for `Endpoint.broadcast` to become async. I think that a lot of these `asyncio.ensure_future` calls are reasonable, but the plugin part might benefit from becoming more asyncio-aware.

(Integrates https://github.com/ethereum/lahja/pull/43)

TODO: 
- [ ] update `setup.py` with the right lahja version once it's been released.
- [x] This misses a call in `BaseIsolatedPlugin._prepare_start()`
- [x] I think that `record_blacklist` needs to become async

#### Cute Animal Picture

cute-bird


lithp:
Closing because I think it makes more sense to do it all at once with #556 (there's no reason to cut two different lahja versions)

PR #539, opened 2019-04-25
Migrate to mypy.ini and enable more checks

cburgdorf:
### What was wrong?

Our `mypy` invocation was getting rather lengthy and hence discouraged adding further checks which is bad as we all love to be punished and bow to our `mypy` master as often as we can.

### How was it fixed?

- migrated to `mypy.ini` file
- added more checks and fixed some issues in the process

#### Cute Animal Picture

### DIFF
--- a/p2p/kademlia.py
+++ b/p2p/kademlia.py
@@ -333,7 +333,7 @@ class RoutingTable:
         # nodes and sort it by distance_to.
         for bucket in self.buckets_by_distance_to(node_id):
             for n in bucket.nodes_by_distance_to(node_id):
-                if n is not node_id:
+                if n.id is not node_id:
                     nodes.append(n)
                     if len(nodes) == k * 2:
                         break
--- a/p2p/kademlia.py
+++ b/p2p/kademlia.py
@@ -8,7 +8,6 @@ import struct
 import time
 from typing import (
     Any,
-    cast,
     Dict,
     Iterable,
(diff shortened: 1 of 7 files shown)


p2p/kademlia.py
```diff
@@ -335,7 +333,7 @@ def neighbours(self, node_id: int, k: int = k_bucket_size) -> List[Node]:
         # nodes and sort it by distance_to.
         for bucket in self.buckets_by_distance_to(node_id):
             for n in bucket.nodes_by_distance_to(node_id):
-                if n is not node_id:
+                if n.id is not node_id:
```
cburgdorf: This looks to me as if it never worked before (as in always appended the node)

trinity/_utils/percentile.py
```diff
@@ -27,7 +27,7 @@ def value(self) -> float:
             raise ValueError("No data for percentile calculation")
 
         idx = (len(self.window) - 1) * self.percentile
-        if int(idx) == idx:
+        if idx.is_integer():
```
cburgdorf: @pipermerriam inspired by https://stackoverflow.com/questions/6209008/checking-if-float-is-equivalent-to-an-integer-value-in-python/22053804#22053804

p2p/kademlia.py
pipermerriam: :heart: type checking

PR #543, opened 2019-04-25
Track metadata abount peers on connection

pipermerriam:
fixes https://github.com/ethereum/trinity/issues/24

### What was wrong?

Every time the trinity client runs, it starts with zero knowledge about the network.  This means that it must bootstrap with the kademlia network and then find appropriate peers to connect to.  This process is time consuming and results in the node taking a long time to find appropriate peers and establish connections to start syncing.

### How was it fixed?

Now, each time we successfully connect to a peer, we keep a record of information about them.

- remote URI
- whether they were originally inbound or outbound
- when we first saw them
- when we last updated their record
- when we last established a connection with them that was at least X seconds long
- their genesis hash
- their network id
- the protocol name they were speaking
- the protocol version they were speaking

Now, when we start, the peer pool will query this database for nodes using the following rules.

- exclude nodes that have an active blacklist record.
- exclude nodes that were originally inbound
- exclude nodes who's network parameters do not match ours (genesis hash, network, etc)
- order nodes by the last time a good connection was established (most recent first)
- exclude nodes that we are already connected to

With this change, a node that goes offline and comes quickly back online will quickly re-establish many of their previous connections as they will be first in the queue.

#### Cute Animal Picture

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -496,6 +496,7 @@ class BasePeer(BaseService):
                 f"Reason must be an item of DisconnectReason, got {reason}"
             )
 
+        self.disconnect_reason = reason
         if reason is DisconnectReason.bad_protocol:
             self.connection_tracker.record_blacklist(
                 self.remote,
@@ -505,7 +506,6 @@ class BasePeer(BaseService):
 
         self.logger.debug("Disconnecting from remote peer %s; reason: %s", self.remote, reason.name)
         self.base_protocol.send_disconnect(reason.value)
-        self.disconnect_reason = reason
         self.transport.close()
 
     async def disconnect(self, reason: DisconnectReason) -> None:
--- a/tests/core/p2p-proto/test_peer_metadata_tracker.py
+++ b/tests/core/p2p-proto/test_peer_metadata_tracker.py
@@ -61,7 +61,6 @@ def test_track_peer_connection_metadata(remote, caplog):
     tracker.track_peer_connection(remote, *TRACK_ARGS)
 
     node = tracker._get_remote(remote.uri())
-    node = tracker._get_remote(remote.uri())
     assert node.genesis_hash == ZERO_HASH_HEX
     assert node.protocol == 'eth'
     assert node.protocol_version == 61
@@ -143,15 +142,14 @@ async def do_tracker_peer_query_test(tracker_params,
         expires_at = datetime.datetime.utcnow() + datetime.timedelta(seconds=delta_seconds)
         blacklist_tracker._create_record(remote, expires_at, 'test')
 
-    candidates = tuple(sorted(
+    candidates = tuple(
         await tracker.get_peer_candidates(
             num_requested=10,
             connected_remotes=connected_remotes or set(),
-        ),
-        key=lambda r: r.uri(),
-    ))
-    just_good_remotes = tuple(r[0] for r in sorted(good_remotes, key=lambda r: r[0].uri()))
-    just_bad_remotes = tuple(r[0] for r in sorted(bad_remotes, key=lambda r: r[0].uri()))
+        )
+    )
+    just_good_remotes = tuple(r[0] for r in good_remotes)
+    just_bad_remotes = tuple(r[0] for r in bad_remotes)
     assert len(candidates) == len(just_good_remotes)
     for remote in just_good_remotes:
         assert remote in candidates
@@ -162,25 +160,25 @@ async def do_tracker_peer_query_test(tracker_params,
 @pytest.mark.asyncio
 async def test_getting_peer_candidates_no_filter():
     await do_tracker_peer_query_test(
-        {},
-        (
+        tracker_params={},
+        good_remotes=(
             (NodeFactory(), True, SIMPLE_META),
             (NodeFactory(), True, SIMPLE_META),
             (NodeFactory(), True, SIMPLE_META),
         ),
-        (),
+        bad_remotes=(),
     )
 
 
 @pytest.mark.asyncio
 async def test_getting_peer_candidates_excludes_non_outbound():
     await do_tracker_peer_query_test(
-        {},
-        (
+        tracker_params={},
+        good_remotes=(
             (NodeFactory(), True, SIMPLE_META),
             (NodeFactory(), True, SIMPLE_META),
         ),
-        (
+        bad_remotes=(
             (NodeFactory(), False, SIMPLE_META),
         ),
     )
@@ -189,12 +187,12 @@ async def test_getting_peer_candidates_excludes_non_outbound():
 @pytest.mark.asyncio
 async def test_getting_peer_candidates_excludes_genesis_hash_mismatch():
     await do_tracker_peer_query_test(
-        {'genesis_hash': ZERO_HASH},
-        (
+        tracker_params={'genesis_hash': ZERO_HASH},
+        good_remotes=(
             (NodeFactory(), True, SIMPLE_META),
             (NodeFactory(), True, SIMPLE_META),
         ),
-        (
+        bad_remotes=(
             (NodeFactory(), True, (ZERO_ONE_HASH, 'eth', 61, 1)),
         ),
     )
@@ -203,12 +201,12 @@ async def test_getting_peer_candidates_excludes_genesis_hash_mismatch():
 @pytest.mark.asyncio
 async def test_getting_peer_candidates_excludes_protocol_mismatch():
     await do_tracker_peer_query_test(
-        {'protocols': ('eth',)},
-        (
+        tracker_params={'protocols': ('eth',)},
+        good_remotes=(
             (NodeFactory(), True, SIMPLE_META),
             (NodeFactory(), True, SIMPLE_META),
         ),
-        (
+        bad_remotes=(
             (NodeFactory(), True, (ZERO_HASH, 'les', 61, 1)),
         ),
     )
@@ -217,12 +215,12 @@ async def test_getting_peer_candidates_excludes_protocol_mismatch():
 @pytest.mark.asyncio
 async def test_getting_peer_candidates_matches_multiple_protocols():
     await do_tracker_peer_query_test(
-        {'protocols': ('eth', 'les')},
-        (
+        tracker_params={'protocols': ('eth', 'les')},
+        good_remotes=(
             (NodeFactory(), True, (ZERO_HASH, 'eth', 61, 1)),
             (NodeFactory(), True, (ZERO_HASH, 'les', 61, 1)),
         ),
-        (
+        bad_remotes=(
             (NodeFactory(), True, (ZERO_HASH, 'bcc', 61, 1)),
         ),
     )
@@ -231,12 +229,12 @@ async def test_getting_peer_candidates_matches_multiple_protocols():
 @pytest.mark.asyncio
 async def test_getting_peer_candidates_excludes_protocol_version_mismatch():
     await do_tracker_peer_query_test(
-        {'protocol_versions': (60, 61)},
-        (
+        tracker_params={'protocol_versions': (60, 61)},
+        good_remotes=(
             (NodeFactory(), True, (ZERO_HASH, 'eth', 60, 1)),
             (NodeFactory(), True, (ZERO_HASH, 'eth', 61, 1)),
         ),
-        (
+        bad_remotes=(
             (NodeFactory(), True, (ZERO_HASH, 'eth', 62, 1)),
         ),
     )
@@ -245,12 +243,12 @@ async def test_getting_peer_candidates_excludes_protocol_version_mismatch():
 @pytest.mark.asyncio
 async def test_getting_peer_candidates_excludes_network_id_mismatch():
     await do_tracker_peer_query_test(
-        {'network_id': 1},
-        (
+        tracker_params={'network_id': 1},
+        good_remotes=(
             (NodeFactory(), True, SIMPLE_META),
             (NodeFactory(), True, SIMPLE_META),
         ),
-        (
+        bad_remotes=(
             (NodeFactory(), True, (ZERO_HASH, 'eth', 62, 2)),
         ),
     )
@@ -259,17 +257,17 @@ async def test_getting_peer_candidates_excludes_network_id_mismatch():
 @pytest.mark.asyncio
 async def test_getting_peer_candidates_complex_query():
     await do_tracker_peer_query_test(
-        {
+        tracker_params={
             'genesis_hash': ZERO_HASH,
             'protocols': ['eth'],
             'protocol_versions': [61],
             'network_id': 1,
         },
-        (
+        good_remotes=(
             (NodeFactory(), True, SIMPLE_META),
             (NodeFactory(), True, SIMPLE_META),
         ),
-        (
+        bad_remotes=(
             (NodeFactory(), False, SIMPLE_META),  # inbound peer
             (NodeFactory(), True, (ZERO_HASH, 'les', 61, 1)),  # wrong protocol
             (NodeFactory(), True, (ZERO_HASH, 'eth', 60, 1)),  # wrong protocol version
@@ -282,15 +280,15 @@ async def test_getting_peer_candidates_complex_query():
 async def test_excludes_blacklisted_peers_from_candidates():
     remote_b = NodeFactory()
     await do_tracker_peer_query_test(
-        {},
-        (
+        tracker_params={},
+        good_remotes=(
             (NodeFactory(), True, SIMPLE_META),
             (NodeFactory(), True, SIMPLE_META),
         ),
-        (
+        bad_remotes=(
             (remote_b, True, SIMPLE_META),
         ),
-        (
+        blacklist_records=(
             (remote_b, 10),  # blacklisted for 10 seconds
         )
     )
@@ -300,15 +298,15 @@ async def test_excludes_blacklisted_peers_from_candidates():
 async def test_includes_expired_blacklisted_from_candidates():
     remote_a = NodeFactory()
     await do_tracker_peer_query_test(
-        {},
-        (
+        tracker_params={},
+        good_remotes=(
             (remote_a, True, SIMPLE_META),
             (NodeFactory(), True, SIMPLE_META),
             (NodeFactory(), True, SIMPLE_META),
         ),
-        (
+        bad_remotes=(
         ),
-        (
+        blacklist_records=(
             (remote_a, -10),  # expired blacklist record
         )
     )
@@ -321,18 +319,18 @@ async def test_candidate_selection_with_mixed_blacklisted_remotes():
     remote_c = NodeFactory()
     remote_d = NodeFactory()
     await do_tracker_peer_query_test(
-        {},
-        (
+        tracker_params={},
+        good_remotes=(
             (remote_a, True, SIMPLE_META),
             (NodeFactory(), True, SIMPLE_META),
             (remote_c, True, SIMPLE_META),
             (NodeFactory(), True, SIMPLE_META),
         ),
-        (
+        bad_remotes=(
             (remote_b, True, SIMPLE_META),
             (remote_d, True, SIMPLE_META),
         ),
-        (
+        blacklist_records=(
             (remote_a, -5),  # expired blacklist record
             (remote_b, 10),  # expired blacklist record
             (remote_c, -20),  # expired blacklist record
@@ -345,12 +343,12 @@ async def test_candidate_selection_with_mixed_blacklisted_remotes():
 async def test_getting_peer_candidates_excludes_already_connected():
     remote_a = NodeFactory()
     await do_tracker_peer_query_test(
-        {},
-        (
+        tracker_params={},
+        good_remotes=(
             (NodeFactory(), True, SIMPLE_META),
             (NodeFactory(), True, SIMPLE_META),
         ),
-        (
+        bad_remotes=(
             (remote_a, True, SIMPLE_META),
         ),
         connected_remotes=(remote_a,),
--- a/tests/core/p2p-proto/test_peer_metadata_tracker.py
+++ b/tests/core/p2p-proto/test_peer_metadata_tracker.py
@@ -1,0 +1,357 @@
+import datetime
+
+import pytest
+
+from p2p.tools.factories import NodeFactory
+
+from trinity.plugins.builtin.network_db.connection.tracker import (
+    SQLiteConnectionTracker,
+)
+from trinity.plugins.builtin.network_db.eth1_peer_db.tracker import (
+    MemoryEth1PeerTracker,
+)
+
+
+@pytest.fixture
+def remote():
+    return NodeFactory()
+
+
+ZERO_HASH = b'\x00' * 32
+ZERO_HASH_HEX = ZERO_HASH.hex()
+
+ZERO_ONE_HASH = b'\x01' * 32
+ZERO_ONE_HASH_HEX = ZERO_ONE_HASH.hex()
+
+
+# is_outbound, last_connected_at, genesis_hash, protocol, protocol_version, network_id
+TRACK_ARGS = (True, None, ZERO_HASH, 'eth', 61, 1)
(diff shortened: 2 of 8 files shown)


pipermerriam:
Some *sub-tasks* I'm discovering that need to be pulled out.

- It's difficult to cause trinity shutdown from within an isolated scope plugin when the shutdown condition can only be detected within `do_start`.
- It's difficult to *properly* run more than one service in an isolated scope plugin since the utils in `trinity._utils.shutdown` only support singular services.

p2p/peer_pool.py
```diff
@@ -395,11 +396,22 @@ def _add_peer(self,
...
+            end_at = time.perf_counter()
+            self.logger.debug(
+                "Failed attempt to connect to %s took %s seconds",
+                node,
+                end_at - start_at,
+            )
             return
+        else:
+            end_at = time.perf_counter()
+            self.logger.debug("Successful connect to %s took %s seconds", node, end_at - start_at)
```
pipermerriam: TODO, remove this new timing code.

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,363 @@
...
+        self.update_peer_meta(
+            remote=peer.remote,
+            genesis_hash=peer.genesis_hash,
+            protocol=peer.sub_proto.name,
+            protocol_version=peer.sub_proto.version,
+            network_id=peer.network_id,
+        )
+
+    def deregister_peer(self, peer: BasePeer) -> None:
+        # TODO: this is where we can track actual performance data about peers
```
pipermerriam: Remove `TODO`

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,363 @@
...
+        if self._meta_exists(uri):
+            self.logger.debug("Updated metadata: %s", remote)
+            eth1_meta = self._get_meta(uri)
+            eth1_meta.genesis_hash = genesis_hash.hex()
+            eth1_meta.protocol = protocol
+            eth1_meta.protocol_version = protocol_version
+            eth1_meta.network_id = network_id
+            eth1_meta.updated_at = datetime.datetime.utcnow()
+        else:
+            self.logger.debug("New metadata: %s", remote)
```
pipermerriam: Too noisy: `debug2`

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,363 @@
...
+                         network_id: int) -> None:
+        uri = remote.uri()
+        try:
+            node = self._get_remote(uri)
+        except NoResultFound:
+            self.logger.warning("Attempt to track metadata for untracked peer: %s", remote)
+            return
+
+        if self._meta_exists(uri):
+            self.logger.debug("Updated metadata: %s", remote)
```
pipermerriam: Too noisy: `debug2`

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,363 @@
...
+
+    def track_peer_connection(self, remote: Node, is_outbound: bool) -> None:
+        uri = remote.uri()
+        if self._remote_exists(uri):
+            self.logger.debug("Updated peer: %s", remote)
+            node = self._get_remote(uri)
+            node.last_connected_at = datetime.datetime.utcnow()
+            return
+        else:
+            self.logger.debug("New peer: %s", remote)
```
pipermerriam: too noisy: `debug2`

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,363 @@
...
+        else:
+            self.genesis_hash = genesis_hash
+        self.protocols = protocols
+        self.protocol_versions = protocol_versions
+        self.network_id = network_id
+
+    def track_peer_connection(self, remote: Node, is_outbound: bool) -> None:
+        uri = remote.uri()
+        if self._remote_exists(uri):
+            self.logger.debug("Updated peer: %s", remote)
```
pipermerriam: too noisy: `debug2`

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,363 @@
...
+            *filters,
+        )
+        num_candidates = base_query.count()
+        self.logger.debug2(
+            "There are %d records in the peer tracking database",
+            num_candidates,
+        )
+        candidates = base_query.order_by(
+            Remote.last_connected_at.desc(),
+        ).all()[min(10, num_requested * 4 - 10):num_requested * 4]
```
pipermerriam: TODO: remove this ugly hack.  The `get_peer_candidates` requests from the peer pool need to be updated to include the URIs for all of the peers the pool is currently connected to so that these backends can properly exclude them from their responses.

pipermerriam:
Anyone who wants to re-connect to the network almost instantly can try this branch out.  It works well and my node comes online and starts syncing pretty much immediately

pipermerriam:
I think I need help deciding the right way forward for this PR.  I *think* it's too much to do all in one go.  I'm thinking I'm going to drop the parts that source peer candidates from this API and have this PR only record the information, and do the sourcing of candidates in another PR.

Reasoning:

I'm concerned with not getting the sourcing right.  Here are some scenarios:

- Re-using the same ones too often
- Database full of mediocre peers 
- Database full of peers that were once available but no longer reachable
- Malicious/Incompetent peers in database that get blacklisted but are always eventually a valid connection candidate.

So probably inclined to release this behind two feature flags, one to turn it on at all, and another to enable sourcing of peer candidates from the recorded data.  Both of these would be off by default.

trinity/plugins/builtin/network_db/plugin.py
```diff
@@ -151,3 +209,26 @@ def do_start(self) -> None:
...
+        else:
+            yield self._get_eth1_peer_server()
+
+    def do_start(self) -> None:
+        try:
+            tracker_services = self._get_services()
+        except BadDatabaseError as err:
+            self.logger.error("Shutting down trinity: %s", str(err))
+            # TODO: no easy way to make this work....  endpoints not yet connected.
+            # self.event_bus.request_shutdown(str(err))
```
cburgdorf: Just dropping this thought here: If you can catch this condition already in `def on_ready(self, manager_eventbus: TrinityEventBusEndpoint) -> None:` then you can use the `manager_eventbus` (that's the event bus from the `main` process` to propagate the shutdown.

cburgdorf:
I just skimmed over this but responding to

>So probably inclined to release this behind two feature flags, one to turn it on at all, and another to enable sourcing of peer candidates from the recorded data. Both of these would be off by default.

I think that is the right way forward. In fact, I think we should leverage feature flags more in the future. Especially the more we stabilize, the more we probably need to get into the habit of letting some features  brew and fine tune behind feature flags for a while.

We could even try to come up with ways to promote such experimental features better, encouraging the adventurous to try them out (e.g. print out available feature flags for experimental features when Trinity boots).

p2p/peer.py
```diff
@@ -274,9 +274,15 @@ def remove_subscriber(self, subscriber: 'PeerSubscriber') -> None:
             cmd, msg = await self.read_msg()
         if isinstance(cmd, Disconnect):
             msg = cast(Dict[str, Any], msg)
-            # Peers sometimes send a disconnect msg before they send the sub-proto handshake.
-            if msg['reason'] == DisconnectReason.too_many_peers.value:
-                raise TooManyPeersFailure(f'{self} disconnected from us before handshake')
+            try:
+                reason = DisconnectReason(msg['reason'])
```
pipermerriam: TODO: extract these updates which better track/handle disconnect reasons.

p2p/peer_backend.py
```diff
@@ -37,12 +38,16 @@ def __init__(self, event_bus: Endpoint) -> None:
 
     async def get_peer_candidates(self,
                                   num_requested: int,
-                                  num_connected_peers: int) -> Iterable[Node]:
+                                  connected_remotes: Set[Node]) -> Iterable[Node]:
```
pipermerriam: TODO: this API change can be extracted independently.

pytest.ini
```diff
@@ -1,5 +1,5 @@
 [pytest]
-addopts= --showlocals --durations 50 --maxfail 10
+addopts= --showlocals --maxfail 10
```
pipermerriam: revert

trinity/config.py
```diff
@@ -246,7 +246,7 @@ def __init__(self,
                  port: int=30303,
                  use_discv5: bool = False,
                  preferred_nodes: Tuple[KademliaNode, ...]=None,
-                 bootstrap_nodes: Tuple[KademliaNode, ...]=None) -> None:
+                 bootstrap_nodes: Tuple[KademliaNode, ...]=None):
```
pipermerriam: TODO: revert

trinity/db/orm.py
```diff
@@ -27,7 +27,7 @@
 Base = declarative_base()
 
 
-SCHEMA_VERSION = '2'
+SCHEMA_VERSION = '4'
```
pipermerriam: TODO: revert to 3

trinity/plugins/builtin/network_db/connection/tracker.py
```diff
@@ -57,10 +58,16 @@ def __init__(self, session: BaseSession):
     # Core API
     #
     async def record_blacklist(self, remote: Node, timeout_seconds: int, reason: str) -> None:
-        if self._record_exists(remote.uri()):
-            self._update_record(remote, timeout_seconds, reason)
+        try:
+            record = self._get_record(remote.uri())
+        except NoResultFound:
+            expires_at = datetime.datetime.utcnow() + datetime.timedelta(seconds=timeout_seconds)
```
pipermerriam: TODO: this can be extracted to an independent PR

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,398 @@
...
+
+    eth1_meta = relationship("Eth1Meta", uselist=False, back_populates="remote")
+    blacklist = relationship(
+        "BlacklistRecord",
+        primaryjoin="Remote.uri==foreign(BlacklistRecord.uri)",
+        uselist=False,
+    )
+
+
+class Eth1Meta(Base):
```
pipermerriam: I don't think I can justify these being two separate tables anymore.

TODO: collapse to a single table with combined fields.

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,398 @@
...
+                         network_id: int) -> None:
+        pass
+
+    async def get_peer_candidates(self,
+                                  num_requested: int,
+                                  connected_remotes: Set[Node]) -> Iterable[Node]:
+        return ()
+
+
+MIN_QUALIFYING_UPTIME = 60
```
pipermerriam: TODO: 2-5 minutes my be a better minimum.  1 minute is a very short connection.

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,398 @@
...
+        uri = remote.uri()
+        now = datetime.datetime.utcnow()
+        if uptime < MIN_QUALIFYING_UPTIME:
+            # we only measure *meaningful* uptime.
+            uptime = 0
+
+        if self._remote_exists(uri):
+            self.logger.debug("Updated peer: %s", remote)
+            node = self._get_remote(uri)
+            node.last_connected_at = now if uptime >= MIN_QUALIFYING_UPTIME else None
```
pipermerriam: TODO: this is wrong,  It shouldn't revert the value to `None` in this case, but rather preserve the previous value.

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,398 @@
...
+            ~Remote.uri.in_(connected_uris),
+            *filters,
+        )
+        num_candidates = base_query.count()
+        self.logger.debug2(
+            "There are %d records in the peer tracking database",
+            num_candidates,
+        )
+        candidates = base_query.order_by(
+            Remote.last_connected_at.desc(),
```
pipermerriam: TODO:  This function needs much bigger/fatter comment that explains the peer selection mechanism in more gory detail since it's going to effect node spin-up time in a major way and it's probably something we'll want to tweak.

p2p/peer_pool.py
```diff
@@ -295,6 +295,8 @@ def _add_peer(self,
     async def stop_all_peers(self) -> None:
         self.logger.info("Stopping all peers ...")
         peers = self.connected_nodes.values()
+        for peer in peers:
+            peer.disconnect_reason = DisconnectReason.client_quitting
```
pipermerriam: I think this is wrong since it would be interpreted as the other client quitting.  Need to revert.

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,400 @@
...
+)
+
+
+class Remote(Base):
+    __tablename__ = 'remotes'
+
+    id = Column(Integer, primary_key=True)
+
+    uri = Column(String, unique=True, nullable=False, index=True)
+    is_outbound = Column(Boolean, nullable=False, index=True)
```
pipermerriam: I tinkered with updating this to include extra concepts like `was_discovered` but it required contextual information that wasn't available and this PR is already large....  So I punted.

tests/core/p2p-proto/test_peer_metadata_tracker.py
```diff
@@ -0,0 +1,357 @@
...
+    updated = tracker._get_remote(remote.uri())
+    assert updated.is_outbound is is_outbound
+
+
+def test_track_peer_connection_metadata(remote, caplog):
+    tracker = MemoryEth1PeerTracker()
+    tracker.track_peer_connection(remote, *TRACK_ARGS)
+
+    node = tracker._get_remote(remote.uri())
+    node = tracker._get_remote(remote.uri())
```
lithp: this line is duplicated

tests/core/p2p-proto/test_peer_metadata_tracker.py
```diff
@@ -0,0 +1,357 @@
...
+        tracker.track_peer_connection(remote, is_outbound, None, *meta_params)
+
+    for remote, is_outbound, meta_params in bad_remotes:
+        tracker.track_peer_connection(remote, is_outbound, None, *meta_params)
+
+    # use the same in-memory database
+    blacklist_tracker = SQLiteConnectionTracker(tracker.session)
+    for remote, delta_seconds in blacklist_records:
+        expires_at = datetime.datetime.utcnow() + datetime.timedelta(seconds=delta_seconds)
+        blacklist_tracker._create_record(remote, expires_at, 'test')
```
lithp: for clarity it might be better to keep using `tracker` here rather than create `blacklist_tracker`.

tests/core/p2p-proto/test_peer_metadata_tracker.py
```diff
@@ -0,0 +1,357 @@
...
+
+    candidates = tuple(sorted(
+        await tracker.get_peer_candidates(
+            num_requested=10,
+            connected_remotes=connected_remotes or set(),
+        ),
+        key=lambda r: r.uri(),
+    ))
+    just_good_remotes = tuple(r[0] for r in sorted(good_remotes, key=lambda r: r[0].uri()))
+    just_bad_remotes = tuple(r[0] for r in sorted(bad_remotes, key=lambda r: r[0].uri()))
```
lithp: I don't think the rest of the test doesn't rely on these being sorted, if that's right I think removing that part would make the tests easier to read. Same with `candidates` above. Although another direction might be to lean into the sorting. I think the below 5 lines could be replaced with:

`assert candidates == just_good_results`

tests/core/p2p-proto/test_peer_metadata_tracker.py
```diff
@@ -0,0 +1,357 @@
...
+            (NodeFactory(), True, SIMPLE_META),
+            (NodeFactory(), True, SIMPLE_META),
+        ),
+        (),
+    )
+
+
+@pytest.mark.asyncio
+async def test_getting_peer_candidates_excludes_non_outbound():
+    await do_tracker_peer_query_test(
```
lithp: These calls are a bit hard to read. They might be easier to read if `tracker_params` was given a default of `{}`, since quite a few tests seem to do that. They also might be easier to read if the parameters were passed in by name:

```python
await do_tracker_peer_query_test(
    good_remotes=(
        (NodeFactory(), True, SIMPLE_META),
        (NodeFactory(), True, SIMPLE_META),
    ),
    bad_remotes=(
        (NodeFactory(), False, SIMPLE_META),
    ),
)
```

tests/core/p2p-proto/test_peer_metadata_tracker.py
```diff
@@ -0,0 +1,357 @@
...
+            (NodeFactory(), True, (ZERO_HASH, 'les', 61, 1)),
+        ),
+        (
+            (NodeFactory(), True, (ZERO_HASH, 'bcc', 61, 1)),
+        ),
+    )
+
+
+@pytest.mark.asyncio
+async def test_getting_peer_candidates_excludes_protocol_version_mismatch():
```
lithp: Another alternative for making these tests easier to parse:

```python
tracker = MemoryEth1PeerTracker(protocol_versions=(60, 61))

assert is_candidate(tracker, (NodeFactory(), True, (ZERO_HASH, 'eth', 60, 1)))
assert is_candidate(tracker, (NodeFactory(), True, (ZERO_HASH, 'eth', 61, 1)))
assert is_not_candidate(tracker, (NodeFactory(), True, (ZERO_HASH, 'eth', 62, 1)))
```

trinity/plugins/builtin/network_db/eth1_peer_db/server.py
```diff
@@ -0,0 +1,61 @@
...
+    BaseEth1PeerTracker,
+)
+from .events import (
+    TrackPeerEvent,
+    GetPeerCandidatesRequest,
+    GetPeerCandidatesResponse,
+)
+
+
+class PeerDBServer(BaseService):
```
lithp: Not a criticism of this PR but this is a 61 line file (and 48 lines of event definitions) which does nothing but expose two RPCs over the event bus. Hopefully a future version of lahja or maybe trinity will have better tooling which makes this easier!

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,400 @@
...
+    genesis_hash = Column(String, nullable=False, index=True)
+    protocol = Column(String, nullable=False, index=True)
+    protocol_version = Column(Integer, nullable=False, index=True)
+    network_id = Column(Integer, nullable=False, index=True)
+
+    blacklist = relationship(
+        "BlacklistRecord",
+        primaryjoin="Remote.uri==foreign(BlacklistRecord.uri)",
+        uselist=False,
+    )
```
lithp: The [SQLAlchemy docs for one-to-one relationships](https://docs.sqlalchemy.org/en/13/orm/basic_relationships.html#one-to-one) suggest using `back_populates` or `backref`. I'm not sure how that's different from the `primaryjoin` approach but wanted to point it out in case it was better in some way.

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,400 @@
...
+        ExtendedDebugLogger,
+        logging.getLogger('trinity.protocol.common.connection.PeerTracker'),
+    )
+
+    msg_queue_maxsize = 100
+    subscription_msg_types: FrozenSet[Type[protocol.Command]] = frozenset()
+
+    def register_peer(self, peer: BasePeer) -> None:
+        # prevent circular import
+        from trinity.protocol.common.peer import BaseChainPeer
```
lithp: I never thought I would advocate for breaking things up into more files but this wouldn't be necessary if `BaseChainPeer` and `BaseChainPeerPool` were in different files.

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,400 @@
...
+            last_connected_at=None,
+            genesis_hash=peer.genesis_hash,
+            protocol=peer.sub_proto.name,
+            protocol_version=peer.sub_proto.version,
+            network_id=peer.network_id,
+        )
+
+    def deregister_peer(self, peer: BasePeer) -> None:
+        """
+        This is likely the proper place to do record statistics about peer
```
lithp: `s/to do/to`

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
lithp: 1 minute seems reasonable for now. In the future we might care more about how useful the peer was than about how long it talked to us for, so leaving it this way sounds good for now!

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -0,0 +1,400 @@
...
+            # Either they have no blacklist record or the record is expired.
+            ((Remote.blacklist == None) | (BlacklistRecord.expires_at <= now)),  # noqa: E711
+            # We are not currently connected to them
+            ~Remote.uri.in_(connected_uris),  # type: ignore
+            # They match our filters for network metadata
+            *metadata_filters,
+        ).order_by(
+            # We want the ones that we have recently connected to succesfully to be first.
+            Remote.last_connected_at.desc(),  # type: ignore
+        ).all()
```
lithp: This probably won't matter for a while but once we know about a lot of nodes this might cause problems: `all()` apparently fetches everything and throws it into a list. [`yield_per`](https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.yield_per) looks like it might also not be what we want but I'm sure there's a way to read from a cursor and only have the database return the rows we want.

lithp: Alternatively, this method could use `num_requested`, which is currently unused.

trinity/plugins/builtin/network_db/eth1_peer_db/tracker.py
pipermerriam: I don't remember all the things I tried but I believe that since the `BlacklistRecord.uri` is not a proper `ForeignKey` that the standard mechanisms ended up being in-adequate or not working in some cases.

tests/core/p2p-proto/test_peer_metadata_tracker.py
pipermerriam: These are two separate classes with different APIs.

trinity/plugins/builtin/network_db/eth1_peer_db/server.py
pipermerriam: I think a *lot* of this will get easier to slim down with the new lahja apis that are coming in since I think the new `EndpointAPI.subscribe` APIs are going to make it a lot easier for us *serve* endpoint responses without needing a `Service` to do it.

veox:
But-but-but the CAP!..



Source: Yuri Smityuk / TASS via Getty Images via Vice

tests/core/p2p-proto/test_peer_metadata_tracker.py
lithp: oops, good point!

PR #545, opened 2019-04-26
Remove cancel token awareness from low level handshake APIs

pipermerriam:
### What was wrong?

The low level handshake APIs were written to need a `CancelToken` passed in to manage cancellations.  However, all of the places that these APIs are called from use the `Cancellable.wait` wrapper which handles cancellation appropriately (i think....)

### How was it fixed?

Removed the use of cancel tokens from these APIs

#### Cute Animal Picture

Cute-Goats-Photo-01


pipermerriam:
This *appears* to be ok but I could use some help evaluating whether this is an ok change.

p2p/auth.py
```diff
@@ -87,9 +84,7 @@
     auth_init = initiator.encrypt_auth_message(auth_msg)
     writer.write(auth_init)
 
-    auth_ack = await token.cancellable_wait(
-        reader.read(ENCRYPTED_AUTH_ACK_LEN),
-        timeout=REPLY_TIMEOUT)
+    auth_ack = await asyncio.wait_for(reader.read(ENCRYPTED_AUTH_ACK_LEN), timeout=REPLY_TIMEOUT)
```
pipermerriam: From @carver comment here: https://github.com/ethereum/trinity/pull/546#discussion_r279111857

TOOD: evaluate if this is ok.  Thinking is that the function this is in is called from service which does `self.wait(...)`.  Does the cancellation of that outer task propagate down into these inner ones?

pipermerriam: brief experimentation suggests that this isn't viable and we'll end up with warnings when cancellation occurs.

carver: Hm, that actually surprises me. We've made that presumption in a few places (that we don't have to keep using `wait()` recursively). I thought that the cancellations done explicitly inside a `self.wait` if the token triggers would be enough to avoid warnings. It would potentially explain some of the warnings we see in the console now, though. Unfortunately, gsalgado was the only one who fully grokked `asyncio` I think X).

What a PITA (and potential performance hit) if we actually have to add `cancellable_wait()`s all the way down into every layer of a call.

pipermerriam: I'd rather each the curio elephant

cburgdorf: >Unfortunately, gsalgado was the only one who fully grokked asyncio I think X)

I'm happy I'm not the only one who feels that way about `asyncio`. It's like, every time I think I grokked it, I encounter something strange again that makes me realize I don't no anything at all. :joy:

voith: @cburgdorf I feel you😂

PR #546, opened 2019-04-26
[POC] Extract networking components from BasePeer into dedicated Transport class.

pipermerriam:
POC for https://github.com/ethereum/trinity/issues/90

### What was wrong?

The current implementation of the `BasePeer` has the following issues:

- Circular dependency from `Peer <---> Protocol` since `Protocol` classes need to send data over the network.
- Typing system overhead since every new protocol ends up needing it's own properly typed peer pool, peer factory.
- Very large and complex `Peer` class that contains a lot of sensitive code alongside a lot of code that we want to be able to experiment with (poor isolation of the sensitive networking code).

### How was it fixed?

Implemented a new `Transport` class (naming suggestions welcome).

This class is *purely* responsible for the lowest level networking  parts of sending and receiving data across the network.  We can now (in theory) have a `Protocol` instance without ever instantiating a `Peer` (but not really yet because the peer is still responsible for the various sub_proto handshakes.  Still working on that part)

#### Cute Animal Picture


p2p/auth.py
```diff
@@ -130,9 +124,10 @@ def pubkey(self) -> datatypes.PublicKey:
         return self.privkey.public_key
 
     async def connect(self) -> Tuple[asyncio.StreamReader, asyncio.StreamWriter]:
-        return await self.cancel_token.cancellable_wait(
+        return await asyncio.wait_for(
```
carver: Hm, so these requests are no longer cancelled if the peer is cancelled? I suspect that could cause issues at shutdown.

carver:
Seems reasonable. Maybe one of us can take a stab at actually doing #152 after this goes through.

> We can now (in theory) have a `Protocol` instance without ever instantiating a `Peer` (but not really yet because the peer is still responsible for the various sub_proto handshakes. Still working on that part)

Pretty cool that this might mean easier testing.

p2p/auth.py
pipermerriam: That is actually part of #545 and is orthogonal to this PR.

As for the cancellation question... I actually don't know.  At the service level, this is all wrapped in a `self.wait` using the service's cancel token, but it isn't intuitive to me whether cancellations propagate in a chain of nested `await` calls and if they do, if that propagation will properly cancel this.

pipermerriam:
@carver take a look at https://github.com/pipermerriam/trinity/pull/2 which is in the direction of what you linked with configurable sub-protocols.  It's incomplete but if you look at the `p2p.connection.Connection` class it should give you an idea of what is going on.  The `BasePeer` class gets *extremely* thin at that point which I  like a lot.

cburgdorf:
>The `BasePeer` class gets extremely thin at that point which I like a lot.

This is great and the direction goes well together with the general `ProxyPeerPool` work that I've been doing. :+1:  Some of our work may overlap but I currently have put my work there on hold until we have the lahja changes in place that support high traffic event bus usage like the one that is needed for that. In general, the work that I've been doing there shouldn't be too hard to adapt to any changes because my solution basically just replays actions on the actual peers and propagates back the responses. And maybe it turns out that your work even opens up other alternative ways forward regarding the isolated process sync so I'm all for it :+1:

Issue #523, opened 2019-04-20
Justification is not persisted in the `BeaconChainDB`


ralexstokes:
This is a bit of an issue meant for discussion as I figure out the best way to introduce the fork choice rule to the beacon chain, which touches on a number of issues relating to getting a "minimum viable testnet" node.

### What is wrong?

An instance of the Beacon node will receive blocks from the network (and possibly ones we create as a proposer). These blocks are passed to an instance of the `BeaconChain` class and processed by its method `import_block`.

This method delegates to some internal machinery, producing the state after applying the operations in this block and persisting the block, the new state and running the fork choice to yield the extension to the canonical chain.

It would seem that there is important data in the resulting state that we are not incorporating into the `BeaconChainDB`, namely the resulting justification data produced as a result of the beacon chain state transition.

We will need to incorporate this data if we expect the node to function properly :)

### How can it be fixed

My recommendation is to add another property we track like the `finalized_head`, namely the `justified_head`, which is the justified block of the greatest epoch we have produced via the state transition of valid blocks (starting from genesis). It seems feasible to compute the justified head on demand via the appropriate state but unless we are worried about the volume of DB writes I think it will make it simpler to go ahead and materialize this field in the DB.

When we import a block, we produce the resulting post-block state and both data should be saved to the DB. Right now (pending #522), we will only save data we know are valid to the DB. In this case, we can write the data unconditionally.

I would suggest we go ahead and at this time compute any updates to the latest {finalized, justified} head and also write them to the database.

pipermerriam:
With journaling/batching mechanisms you can make some trade-offs between having data "appear" to be written to the database in exchange for some level of boosted complexity in reasoning about the database itself.

I.E. write information to the database that you may not be entirely sure about yet but only actually persisting through to the underlying database at specified checkpoints once you know that data is *good*.

carver:
Yeah, this sounds a whole lot like how you sometimes want to generate/access the state root without necessarily persisting it to the DB.

ralexstokes:
thanks @pipermerriam @carver 

it sounds like we can look into these more complex mechanisms if we have issues around performance, like in the case that we want to track some head but not actually write it until we know more about the network state.

in this case, it seems that the justification property i'm referring to is monotonic in the sense that we are fine to go w/ the justified block at the highest slot...

[...]

PR #547, opened 2019-04-27
Update finalization and justification in the DB

ralexstokes:
### What was wrong?

Addresses #523.

This PR serves as preliminary work for the implementation of the LMD GHOST fork choice in #532.

### How was it fixed?

- [x] When a state is persisted, we check if there is a new finalized head and update it accordingly in the DB if there is one.

- [x] When a state is persisted, we check if there is a newly justified block that has been justified for at least one epoch. If so, then we set it as the new justified head.

Additionally, adds some attributes to the `BeaconChainDB` to avoid what would otherwise result in a gratuitous number of database reads.

### Questions

* ~~Do you think it is cleaner to return `ZERO_HASH32` instead of using an `Optional[Hash32]` in `_get_last_finalized_root_if_present`? The logic should all work the same and some may consider it better stylistically to avoid `None` if we can.~~

#### Cute Animal Picture


eth2/beacon/db/chain.py
```diff
@@ -150,9 +164,26 @@ def get(self, key: bytes) -> bytes:
...
+        except FinalizedHeadNotFound:
+            return None
+
+    def _get_highest_justified_epoch(self, db: BaseDB, config: Eth2Config) -> Epoch:
+        try:
+            justified_head_root = self._get_justified_head_root(db)
+            slot = self.get_slot_by_root(justified_head_root)
+            return slot_to_epoch(slot, config.SLOTS_PER_EPOCH)
+        except JustifiedHeadNotFound:
+            return Epoch(0)
```
hwwhww: I suppose we should return `config.GENESIS_EPOCH`?

eth2/beacon/db/chain.py
```diff
@@ -150,9 +164,26 @@ def get(self, key: bytes) -> bytes:
...
         self.db = db
 
+        self._last_finalized_root = self._get_last_finalized_root_if_present(db)
+        self._highest_justified_epoch = self._get_highest_justified_epoch(db, config)
+
+    def _get_last_finalized_root_if_present(self, db: BaseDB) -> Optional[Hash32]:
+        try:
+            return self._get_finalized_head_root(db)
+        except FinalizedHeadNotFound:
+            return None
```
hwwhww: > Do you think it is cleaner to return `ZERO_HASH32` instead of using an `Optional[Hash32]` in `_get_last_finalized_root_if_present`? The logic should all work the same and some may consider it better stylistically to avoid `None` if we can.

Since in genesis state, `finalized_epoch` is `GENESIS_EPOCH` and `finalized_root` is `ZERO_HASH32`, I think it makes sense to return `ZERO_HASH32` here. 👍

eth2/beacon/db/chain.py
```diff
@@ -150,9 +164,26 @@ def get(self, key: bytes) -> bytes:
 
 
 class BeaconChainDB(BaseBeaconChainDB):
-    def __init__(self, db: BaseAtomicDB) -> None:
+    def __init__(self, db: BaseAtomicDB, config: Eth2Config) -> None:
         self.db = db
 
+        self._last_finalized_root = self._get_last_finalized_root_if_present(db)
```
hwwhww: Maybe `_last_finalized_root` -> `_finalized_root`?

eth2/beacon/db/chain.py
ralexstokes: ah yes -- thank you, i guess this is why we have code review :)

eth2/beacon/db/chain.py
ralexstokes: yeah it seems much cleaner to me now, as well

eth2/beacon/db/chain.py
ralexstokes: yeah that works

trinity/config.py
```diff
@@ -623,6 +633,11 @@ def beacon_chain_class(self) -> Type['BeaconChain']:
 
     def initialize_chain(self,
                          base_db: BaseAtomicDB) -> 'BeaconChain':
+        # TODO(ralexstokes):
```
pipermerriam: I like this

ralexstokes:
Note: there are some things in this PR i'm punting on that I intend to improve in #532. I decided to go ahead and make this a separate PR so that #532 does not get too unwieldly.

ralexstokes:
ok -- i wanted to add the tests that i have done today -- perhaps ill wait for an approving PR and then we can merge this in?

i'll be working on top of it for #532.

ralexstokes:
@hwwhww i agree we should seek simpler interfaces where we can....

the main use of it here is actually for `SLOTS_PER_EPOCH` which i think is pretty helpful to be able to parameterize; i'll keep the existing interface for `BeaconChainDB` unless you think we should go another route (feel free to open an issue to this effect and i'll handle it)

ralexstokes:
this PR is blocked on https://github.com/ethereum/py-ssz/pull/71

eth2/beacon/db/chain.py
```diff
@@ -589,18 +642,97 @@ def persist_state(self,
...
+                                                       db: BaseDB,
+                                                       genesis_block: BaseBeaconBlock) -> None:
+        """
+        The genesis ``BeaconState`` lacks the correct justification and finality
+        data in the early epochs. The invariants of this class require an exceptional
+        handling to mark the genesis block's root and the genesis epoch as
+        finalized and justified.
+        """
+        genesis_root = genesis_block.signing_root
+        genesis_epoch = slot_to_epoch(genesis_block.slot, self.config.SLOTS_PER_EPOCH)
```
hwwhww: Remove it and use `config.GENESIS_EPOCH` at below.
```suggestion
```

eth2/beacon/db/chain.py
```diff
@@ -589,18 +642,97 @@ def persist_state(self,
...
+        self._update_finalized_head(genesis_root)
+        self._update_justified_head(genesis_root, genesis_epoch)
```
hwwhww:
```suggestion
        self._update_justified_head(genesis_root, config.GENESIS_EPOCH)
```

eth2/beacon/db/chain.py
ralexstokes: nice catch

ralexstokes:
@hwwhww yeah i've been talking w/ some of the other snake charmers about where to put the fork choice -- we aren't tied to it being in the DB, although it will take a bit of work to pull out somewhere else... this is something to consider in #532... @carver did point out we want, if anything, move the fork choice to the `StateMachine` layer (as a fork could change the fork choice)

ralexstokes:
@hwwhww i'll wait for a confirming 👍 from you and then this can be merged in :)

setup.py
```diff
@@ -84,7 +84,7 @@
         "py-ecc>=1.6.0,<2.0.0",
         "rlp>=1.1.0,<2.0.0",
         "py-evm==0.2.0a42",
-        "ssz==0.1.0a7",
+        "ssz @ https://github.com/ethereum/py-ssz/tarball/master",
```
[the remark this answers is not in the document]
ralexstokes: yep, i'll revert and merge

PR #549, opened 2019-04-28
Fix RANDAO message_hash

hwwhww:
### What was wrong?

In spec [v0.5.1](https://github.com/ethereum/eth2.0-specs/blob/v0.5.1/specs/core/0_beacon-chain.md#randao), we use `hash_tree_root` to generate `message_hash`.

### How was it fixed?
Use `ssz.hash_tree_root(epoch, sedes=ssz.sedes.uint64)`

#### Cute Animal Picture

barn-owl-1107397_640


ralexstokes:
@hwwhww looks like there was a (unrelated) CI failure?

if you want to verify that's not a problem, then go ahead and merge when ready!

the logic looks good (and nice catch -- we are no longer just using the serialization...)

PR #551, opened 2019-04-29
Fix `_compute_inactivity_leak_deltas`

hwwhww:
### What was wrong?
See https://github.com/ethereum/eth2.0-specs/blob/v0.5.1/specs/core/0_beacon-chain.md#rewards-and-penalties: `def compute_inactivity_leak_deltas`. It should be `base_rewards[index] * config.MIN_ATTESTATION_INCLUSION_DELAY // inclusion_infos[index].inclusion_distance`.

### How was it fixed?
Fix the wrong operator.

#### Cute Animal Picture

short-eared-owl-1225961_640


hwwhww:
I've rebuilt and rerun many times but still failed to make the irrelevant tests pass. :'(

PR #552, opened 2019-04-29
v0.5.1 state tests

hwwhww:
### What was wrong?

v0.5.1 state-tests: https://github.com/ethereum/eth2.0-tests/tree/master/state
Add tests/eth2/beacon/state-fixtures/test_minimal_state.py for testing it.

### How was it fixed?
1. Add submodule `ethereum/eth2.0-tests`
2. Add tests/eth2/beacon/state-fixtures/test_minimal_state.py
3. Update `BeaconStateMachine.__init__` for supporting setting `pre_state` in state machine.
4. Sync the spec constants
5. Simplify `validate_proposer_signature` interface

### How to test it:
```
pytest tests/eth2/beacon/state-fixtures/ -s
```

(And v0.6.0 tests are coming...)

#### Cute Animal Picture

racoon-1414027_640

### DIFF
--- a/tests/eth2/beacon/state-fixtures/test_minimal_state.py
+++ b/tests/eth2/beacon/state-fixtures/test_minimal_state.py
@@ -1,12 +1,20 @@
-import os
+from pathlib import Path
 import yaml
+try:
+    from yaml import CLoader as Loader
+except ImportError:
+    from yaml import Loader
 import pytest
 
-from py_ecc import bls
+from eth_utils import (
+    to_tuple,
+)
+from py_ecc import bls  # noqa: F401
 from ssz.tools import (
     from_formatted_dict,
     to_formatted_dict,
 )
+
 
 from eth2.configs import (
     Eth2Config,
@@ -22,21 +30,19 @@ from eth2.beacon.state_machines.forks.serenity import (
 )
 
 # Test files
-ROOT_PROJECT_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+ROOT_PROJECT_DIR = Path(__file__).cwd()
 
-BASE_FIXTURE_PATH = os.path.join(ROOT_PROJECT_DIR, '../../', 'eth2-fixtures', 'state')
+BASE_FIXTURE_PATH = ROOT_PROJECT_DIR / 'eth2-fixtures' / 'state'
 
-FILE_NAMES = [
+FIXTURE_FILE_NAMES = [
     "sanity-check_small-config_32-vals.yaml",
-    "sanity-check_default-config_100-vals.yaml",
+    # "sanity-check_default-config_100-vals.yaml",
 ]
+
 
 #
 # Mock bls verification for these tests
 #
-bls = bls
-
-
 def mock_bls_verify(message_hash, pubkey, signature, domain):
     return True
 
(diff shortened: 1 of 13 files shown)


tests/eth2/beacon/state-fixtures/test_minimal_state.py
```diff
@@ -0,0 +1,138 @@
...
+        return
+
+    mocker.patch('py_ecc.bls.verify', side_effect=mock_bls_verify)
+    mocker.patch('py_ecc.bls.verify_multiple', side_effect=mock_bls_verify_multiple)
+
+
+def get_test_case(file_names):
+    all_test_cases = []
+    for file_name in file_names:
+        with open(BASE_FIXTURE_PATH + '/' + file_name, 'U') as f:
```
ChihChengLiang: Tip: In py-ssz the file path and line number of yaml, supported by `ruamel.yaml`, are passed into test_ids, so that devs can jump to the exact line of yaml file in the failing cases. 
```
def make_test_id(filename, test_case):
    return f"{filename}:{test_case.lc.line}"
```

tests/eth2/beacon/state-fixtures/test_minimal_state.py
```diff
@@ -0,0 +1,138 @@
...
+)
+
+# Test files
+ROOT_PROJECT_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+
+BASE_FIXTURE_PATH = os.path.join(ROOT_PROJECT_DIR, '../../', 'eth2-fixtures', 'state')
+
+FILE_NAMES = [
+    "sanity-check_small-config_32-vals.yaml",
+    "sanity-check_default-config_100-vals.yaml",
```
ChihChengLiang: It looks like the test is slow because loading 100-vals tests yaml file is too slow. I tried `CLoader` and it can make loading 100-vals significantly faster.
```python
from yaml import CLoader as Loader
with open("eth2-fixtures/state/sanity-check_default-config_100-vals.yaml") as f:
    yaml.load(f, Loader=Loader)
```

tests/eth2/beacon/state-fixtures/test_minimal_state.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+test_cases = get_test_case(FILE_NAMES)
+
+
+@pytest.mark.parametrize("test_case", test_cases)
+def test_state(base_db, test_case):
+    test_name = test_case['name']
+    if test_name == 'test_transfer':
+        print('skip')
```
ChihChengLiang: maybe do `pytest.skip(reason)`?

tests/eth2/beacon/state-fixtures/test_minimal_state.py
ChihChengLiang: In `ruamel.yaml`, could speed up with `yaml = YAML(typ='unsafe')`

tests/eth2/beacon/state-fixtures/test_minimal_state.py
hwwhww: It seems that if I use `yaml = YAML(typ='unsafe')`, then I cannot use `lc` feature. 😮 
I chose `lc` for debugging!

PR #553, opened 2019-04-29
Update shutdown utils to accept multiple services

pipermerriam:
### What was wrong?

A plugin which runs multiple services will want to use the shutdown utilities to manage the clean shutdown of each of those services. These utilities however only support a single service.

### How was it fixed?

Extended the utilities in` trinity._utils.shutdown` to allow for many services to be passed in.

#### Cute Animal Picture

Baby miniature goats

### DIFF
--- a/tests/core/test_service_and_endpoint_shutdown_utils.py
+++ b/tests/core/test_service_and_endpoint_shutdown_utils.py
@@ -1,0 +1,51 @@
+import asyncio
+import multiprocessing
+import os
+import signal
+
+import pytest
+
+from p2p.service import BaseService
+
+from trinity.endpoint import TrinityEventBusEndpoint
+from trinity._utils.shutdown import (
+    exit_with_endpoint_and_services,
+)
+
+
+class SimpleService(BaseService):
+    def __init__(self, ready_to_kill_event, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.ready_to_kill_event = ready_to_kill_event
+
+    async def _run(self):
+        self.ready_to_kill_event.set()
+        await self.cancellation()
+
+
+def run_service(ready_to_kill_event):
+    loop = asyncio.get_event_loop()
+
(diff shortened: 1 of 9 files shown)


tests/core/test_service_and_endpoint_shutdown_utils.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@pytest.mark.parametrize('sig', (signal.SIGINT, signal.SIGTERM))
+@pytest.mark.asyncio
+async def test_exit_with_endpoind_and_services_facilitates_clean_shutdown(sig):
+    ready_to_kill_event = multiprocessing.Event()
+    proc = multiprocessing.Process(target=run_service, args=(ready_to_kill_event,))
+    proc.start()
+
+    ready_to_kill_event.wait()
```
cburgdorf: I think you can just do the following here:

```python
await run_service.events.started.wait()
```

Or is there a reason this needs to be a `multiprocessing.Event` ?

tests/core/test_service_and_endpoint_shutdown_utils.py
```diff
@@ -0,0 +1,51 @@
...
+)
+
+
+class SimpleService(BaseService):
+    def __init__(self, ready_to_kill_event, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.ready_to_kill_event = ready_to_kill_event
+
+    async def _run(self):
+        self.ready_to_kill_event.set()
```
cburgdorf: `BaseService` already exposes `service.events.started` which is an `asyncio.Event` but I *think*  that should do it as well, no?

pipermerriam: This is specifically a `multiprocessing.Event` so that I can pass it across the process boundary.  Note that the `wait()` call happen in the main process.

cburgdorf: Ah, of course :see_no_evil:

PR #554, opened 2019-04-29
Update token bucket API

pipermerriam:
### What was wrong?

The `token_bucket` API that was implemented in the `p2p` module was done so in `p2p._utils` which is private.  I'd like to use this API in `triinity` **and** I need some additional functionality too allow failures when taking tokens from the bucket.

### How was it fixed?

Converted from a pure async generator to a simple class that has a `take_nowait` method which can be used to force an immediate success or failure when taking from the bucket.

#### Cute Animal Picture

miryah_kid_feb282012_00261

### DIFF
--- a/p2p/token_bucket.py
+++ b/p2p/token_bucket.py
@@ -1,0 +1,91 @@
+import asyncio
+import time
+from typing import (
+    Union,
+    AsyncGenerator,
+)
+
+
+class NotEnoughTokens(Exception):
+    """
+    Raised if the token bucket is empty when trying to take a token in blocking
+    mode.
+    """
+    pass
+
+
+class TokenBucket:
+    def __init__(self,
+                 rate: Union[int, float],
+                 capacity: Union[int, float]) -> None:
+        self._rate = rate
+        self._capacity = capacity
+        self._num_tokens = self._capacity
+        self._last_refill = time.perf_counter()
+        self._seconds_per_token = 1 / self._rate
+        self._take_lock = asyncio.Lock()
+
+    async def __aiter__(self) -> AsyncGenerator[None, None]:
+        """
+        Can be used as an async iterator to limit the rate at which a loop can
+        run.
+        """
+        while True:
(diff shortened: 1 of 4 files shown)


p2p/token_bucket.py
```diff
@@ -0,0 +1,91 @@
...
+    def get_num_tokens(self) -> float:
+        """
+        Return the number of tokens current in the bucke
+        """
+        return max(0, self._get_num_tokens(time.perf_counter()))
+
+    def _get_num_tokens(self, when: float) -> float:
+        return min(
+            self._capacity,
+            self._num_tokens + (self._rate * (when - self._last_refill)),
```
carver: Seems like the `max(0` part could move in here, then this could use that new `clamp()` tool.

p2p/token_bucket.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def take_nowait(self, num: Union[int, float] = 1) -> None:
+        # we calculate this value locally to ensure that in the case of not
+        # having enough tokens the error message is accurate due to race
+        # condition between calculating capacity and raising the error message.
+        num_tokens = self.get_num_tokens()
+        if num_tokens >= num:
+            self._take(num)
+        else:
+            raise NotEnoughTokens(f"Insufficient capacity.  Needed {num} but only has {num_tokens}")
```
carver: Not sure what the right precision is here, but maybe:
```suggestion
            raise NotEnoughTokens(f"Insufficient capacity.  Needed {num:.2f} but only has {num_tokens:.2f}")
```

p2p/token_bucket.py
pipermerriam: I'll add a comment.

The implementation is optimized a bit, part of which is allowing the internal representation of the balance to go negative (which is how we determine if a sleep is needed)

pipermerriam:
I accidentally merged this by merging #557 

@carver leave me any review you have and I'll clean it up.

PR #555, opened 2019-04-29
Extend blacklisting to peer behavior to peer request timeouts.

pipermerriam:
- Part of https://github.com/ethereum/trinity/issues/520
- Builds on https://github.com/ethereum/trinity/pull/557

### What was wrong?

When we encounter a timeout when interacting with the peer request/response API (**not** the core peer API for reading messages), we tend to simply log that the request timed out and continue on.  I.E. a peer that always times out (but that is not unresponsive) will remain in our peer pool indefinitely.

### How was it fixed?

The `Manager` class for the request/response API will now disconnect from a peer that times out too often.  This is measured using a token bucket which allows for "on average" one timeout every 5 minutes with a maximum burst of 5 timeouts.  Peers who hit this limit are disconnected and blacklisted for 5 minutes.

> **NOTE** the the internal API for connection tracker will scale this timeout up automatically as it happens more often.

This also removes the 99th percentile dynamic tuning of round trip timeouts as was discussed in another issue.   The reason is in the event that our network connection to a peer changes a little for the worse, they start timing out all of their requests.  This is most prone to happening to our most performant peers.  I believe we decided that without aggregate statistics about peer response times across the pool, any form of dynamic tuning of this value was likely to cause more problems than it solved.  (going to extract this to a separate PR tomorrow)

#### Cute Animal Picture

goatbucklingadrenaline

### DIFF
--- a/p2p/constants.py
+++ b/p2p/constants.py
@@ -131,5 +131,5 @@ MAX_CONCURRENT_CONNECTION_ATTEMPTS = 10
 # `DisconnectReason.bad_protocol`
 BLACKLIST_SECONDS_BAD_PROTOCOL = 60 * 10  # 10 minutes
 
-# Amount of time a peer will be blacklisted when they timeout too freequently
+# Amount of time a peer will be blacklisted when they timeout too frequently
 BLACKLIST_SECONDS_TOO_MANY_TIMEOUTS = 60 * 5  # 5 minutes
--- a/p2p/constants.py
+++ b/p2p/constants.py
@@ -127,5 +127,9 @@ REQUEST_PEER_CANDIDATE_TIMEOUT = 1
 # The maximum number of concurrent attempts to establis new peer connections
 MAX_CONCURRENT_CONNECTION_ATTEMPTS = 10
 
-# Time values in seconds
-MINUTES_10 = 60 * 10
+# Amount of time a peer will be blacklisted when they are disconnected as
+# `DisconnectReason.bad_protocol`
+BLACKLIST_SECONDS_BAD_PROTOCOL = 60 * 10  # 10 minutes
+
+# Amount of time a peer will be blacklisted when they timeout too freequently
+BLACKLIST_SECONDS_TOO_MANY_TIMEOUTS = 60 * 5  # 5 minutes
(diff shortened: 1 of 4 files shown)


pipermerriam:
Builds upon https://github.com/ethereum/trinity/pull/554

carver:
> blacklisted for 60 seconds.

Seemed surprisingly short when the trigger is based on one timeout allowed per 5 min. Maybe that timeout rate and the blacklist could be the same? So a 5-minute blacklist for a 1 timeout/5min window.

trinity/protocol/common/managers.py
```diff
@@ -68,6 +75,11 @@ def __init__(
         self.response_msg_type = response_msg_type
         self._lock = asyncio.Lock()
 
+        # token bucket for limiting timeouts.
+        # - Refills at 1-token every 5 minutes
+        # - Max capacity of 5 tokens
```
pipermerriam: defined capacity is actually `3`

pipermerriam:
Bumped to 5 minutes and did the other cleanup from #558

p2p/constants.py
```diff
@@ -127,5 +127,9 @@
 # The maximum number of concurrent attempts to establis new peer connections
 MAX_CONCURRENT_CONNECTION_ATTEMPTS = 10
 
-# Time values in seconds
-MINUTES_10 = 60 * 10
+# Amount of time a peer will be blacklisted when they are disconnected as
+# `DisconnectReason.bad_protocol`
+BLACKLIST_SECONDS_BAD_PROTOCOL = 60 * 10  # 10 minutes
+
+# Amount of time a peer will be blacklisted when they timeout too freequently
```
carver:
```suggestion
# Amount of time a peer will be blacklisted when they timeout too frequently
```

trinity/protocol/common/managers.py
```diff
@@ -68,6 +78,11 @@ def __init__(
         self.response_msg_type = response_msg_type
         self._lock = asyncio.Lock()
 
+        # token bucket for limiting timeouts.
+        # - Refills at 1-token every 5 minutes
+        # - Max capacity of 3 tokens
+        self.timeout_bucket = TokenBucket(1 / 300, 3)
```
carver: Looks like these two numbers belong in `.constants`

PR #556, opened 2019-04-29
Use lahja v0.12.0

lithp:
This is a version of trinity which uses the new asyncio-powered lahja (https://github.com/ethereum/lahja/pull/39). Previously the `*_nowait()` methods would block the main thread and do their work. Since most of them now call async methods they now schedule their work to be done in the future. This caused problems inside trinity, which for example assumes that it can call `broadcast_nowait` immediately after it calls `connect_to_endpoints_nowait`. To account for this a couple parts of trinity were turned into async methods so that the blocking versions of those `Endpoint` methods could be called.

#### Cute Animal Picture

seal

### DIFF
--- a/trinity/bootstrap.py
+++ b/trinity/bootstrap.py
@@ -1,12 +1,15 @@
+import asyncio
 from argparse import ArgumentParser, Namespace
 import logging
 import multiprocessing
 import os
+import signal
 from typing import (
     Any,
     Callable,
     Dict,
     Iterable,
+    Tuple,
     Type,
 )
 
@@ -43,6 +46,9 @@ from trinity.extensibility import (
     BaseManagerProcessScope,
     MainAndIsolatedProcessScope,
     PluginManager,
+)
+from trinity.events import (
+    ShutdownRequest,
 )
 from trinity._utils.ipc import (
     kill_process_gracefully,
@@ -95,7 +101,7 @@ BootFn = Callable[[
     logging.handlers.QueueListener,
     TrinityMainEventBusEndpoint,
     logging.Logger
-], None]
+], Tuple[multiprocessing.Process, ...]]
 
 
 def main_entry(trinity_boot: BootFn,
@@ -144,7 +150,6 @@ def main_entry(trinity_boot: BootFn,
     if args.log_levels:
         setup_log_levels(args.log_levels)
 
-    main_endpoint.track_and_propagate_available_endpoints()
     try:
         trinity_config = TrinityConfig.from_parser_args(args, app_identifier, sub_configs)
     except AmbigiousFileSystem:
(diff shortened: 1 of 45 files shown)


carver:
One more todo suggestion (before merging):

- [ ] update to a released version of lahja

lithp:
Good point, added!

lithp:
The tests work correctly when I run them locally, using both py36 and py37. It looks like circleci's tox is still trying to use `lahja=0.11.2` and I don't know why, given that setup.py clearly specifies which branch to use.

trinity/extensibility/plugin.py
```diff
@@ -287,38 +287,45 @@ def start(self) -> None:
...
         level = self.context.boot_kwargs.get('log_level', logging.INFO)
         setup_queue_logging(log_queue, level)
         if self.context.args.log_levels:
             setup_log_levels(self.context.args.log_levels)
+
+        with self.context.trinity_config.process_id_file(self.normalized_name):
+            loop = asyncio.get_event_loop()
+            asyncio.ensure_future(self._prepare_start())
+            loop.run_forever()
+            loop.close()
```
cburgdorf: One thing I like to point out is that originally the `BaseIsolatedPlugin` worked under the assumption that plugins are free to not use `asyncio` at all and just use entirely synchronous APIs. Based on the fact that these plugins run in their own process, there shouldn't be anything fundamentally wrong about it (yet discouraged).

However, now that I think about it, I guess that assumption wasn't even true anymore from that moment on where plugins started being equipped with the event bus. Or actually, I'm not sure what happens to an `BaseIsolatedPlugin` today that doesn't start a loop. Do the attempts to connect the event bus just go unnoticed or do errors show up?

Anyway, this PR puts the creation of the loop out of the plugin author land and I just wanted to point out what the original motivation for having it in the plugin author land was. Based on the fact how integral the event bus architecture is today, I'm ok to just entirely drop support for isolated plugins that want to have full control about creating/not-creating a loop.

trinity/bootstrap.py
```diff
@@ -195,28 +199,66 @@ def main_entry(trinity_boot: BootFn,
...
-        # the `main_endpoint` which is why we connect to our own endpoint here
-        main_endpoint.connect_to_endpoints_blocking(main_connection_config)
-        trinity_boot(
-            args,
+    def kill_trinity_with_reason(reason: str) -> None:
+        kill_trinity_gracefully(
             trinity_config,
-            extra_kwargs,
+            stderr_logger,
+            (),  # TODO: networking_process, database_server_process
```
cburgdorf: The reason this wasn't generic so far is that the `beacon` node does not have a hardcoded `networking` process (which eventually should be the case for the eth1 node, too)

trinity/extensibility/plugin.py
carver: Hm, the ability to start plugins with different options seems like it will be important in not too long. (wanting to test with `curio`, `trio`, etc) I'm not saying that requiring asyncio support is a dealbreaker for this PR, but if it's easy to support other options, let's keep it.

pipermerriam: Relevant: https://github.com/lithp/lahja/pull/2

lithp: I agree that it'll soon be important for plugins to be able to switch between curio / trio / asyncio. In fact, I'm quite excited to see how curio works!

However, I still think this PR is a step in the right direction. Every plugin needs an event bus, and setting up the event bus requires having entered the async part of the program (I assume that'll also be true with a curio eventbus). Currently this is called `BaseIsolatedPlugin` but I think it'll eventually become `BaseAsyncioIsolatedPlugin`, with an equivalent `BaseCurioIsolatedPlugin` which sets up a curio event bus and passes it to the plugin.

In addition, by centralizing event loop creation this changeset drops some code which was duplicated across multiple files, which is usually a good sign :) If we ever decide to switch to using curio entirely that'll now be easier!

pipermerriam: I don't think you guys need to worry to much about the curio part.  Just make this work for now.  Migrating something over to a curio event bus is going to have a decent freedom of movement.  We don't need 1:1 API parity, just equivalent functionality.  In fact, most of the `_nowait` APIs aren't going to make much sense in the curio model but I'm still sorting that out.

lithp:
@cburgdorf aaand a last one for you to look at :)

trinity/extensibility/plugin_manager.py
```diff
@@ -99,7 +99,7 @@ def create_plugin_context(self,
             # created here for API symmetry. Endpoints are pickleable *before* they are connected,
             # which means, this Endpoint will be pickled and transferred into the new process
             # together with the rest of the `PluginContext`.
-            plugin.set_context(PluginContext(TrinityEventBusEndpoint(), boot_info,))
+            plugin.set_context(PluginContext(None, boot_info,))
```
cburgdorf: All plugins that derive from `BaseIsolatedPlugin` need to change event bus access from `self.context_event_bus` to `self.event_bus`.

I ran `trinity trinity --ethstats` on this PR which will cause a shutdown because of missing CLI flags and then I hit this error:

```
   ERROR  05-06 09:59:13               asyncio  Task exception was never retrieved
future: <Task finished coro=<exit_with_endpoint_and_services() done, defined at /home/cburgdorf/Documents/hacking/ef/trinity/trinity/_utils/shutdown.py:19> exception=AttributeError("'NoneType' object has no attribute 'stop'")>
  File "/home/cburgdorf/Documents/hacking/ef/trinity/trinity/_utils/shutdown.py", line 22, in exit_with_endpoint_and_services
    endpoint.stop()
AttributeError: 'NoneType' object has no attribute 'stop'
    INFO  05-06 09:59:14         PluginManager  Process 22
```

That's because of this line that still tries to read the event bus from `self.context.event_bus`

`asyncio.ensure_future(exit_with_endpoint_and_services(self.context.event_bus, service))

lithp: I tried to fix this and things got a little out of hand: https://github.com/ethereum/trinity/pull/592

PR #557, opened 2019-04-29
Piper/expose event bus on peer

pipermerriam:
- Part of https://github.com/ethereum/trinity/issues/520
- builds on https://github.com/ethereum/trinity/pull/554

### What was wrong?

In order to implement blacklisting for various peer interactions, we need access to the connection tracker from the peer instance.

### How was it fixed?

Adds a handle to the event bus to each peer.

Also includes these cleanups:

- Note about the round trip timeout
- Import ordering in `p2p.peer`
- rename `PeerPool.event_bus` to method `PeerPool.get_event_bus()` to make it a little clearer that something is going on under the hood.  (I don't like this API but it's simple and I figure we'll iterate on it)
- add version to the network-db sqlite database filename.  Makes for smoother upgrades and ease of development across different database versions.

#### Cute Animal Picture

28d3ebe00d1c57f39fcb204020b29159

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -1,3 +1,4 @@
+from abc import ABC, abstractmethod
 import asyncio
 import collections
 import contextlib
@@ -6,11 +7,6 @@ import functools
 import logging
 import operator
 import struct
-from abc import (
-    ABC,
-    abstractmethod
-)
-
 from typing import (
     Any,
     cast,
@@ -23,6 +19,8 @@ from typing import (
     Type,
     TYPE_CHECKING,
 )
+
+from lahja import Endpoint
 
 import sha3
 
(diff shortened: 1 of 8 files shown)


p2p/peer.py
```diff
@@ -247,6 +253,23 @@ def __init__(self,
...
         self.boot_manager = self.get_boot_manager()
+        self.connection_tracker = self.setup_connection_tracker()
+
+    @property
+    def has_event_bus(self) -> bool:
+        return self._event_bus is not None
+
+    def get_event_bus(self) -> Endpoint:
+        if self._event_bus is None:
+            raise AttributeError("No event bus configured for this peer pool")
```
cburgdorf: >No event bus configured for this peer ~~pool~~

PR #558, opened 2019-04-29
Piper/blacklist peers who send malformed messages

pipermerriam:
- builds on: https://github.com/ethereum/trinity/pull/557
- part of: #520

### What was wrong?

Currently if a peer sends us a malformed message we simply disconnect from them.  With the upcoming change that has us persist information about peers that we can connect to, this ends up resulting in us re-connecting to bad/broken peers over and over again.

### How was it fixed?

When a peer sends us a malformed message, blacklist them for 5 minutes.

#### Cute Animal Picture

4af43725bc906e8d02dffacd8b453a84--baby-pygmy-goats-baby-goats

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -430,6 +430,11 @@ class BasePeer(BaseService):
                 )
                 return
             except MalformedMessage as err:
+                self.connection_tracker.record_blacklist(
+                    self.remote,
+                    timeout_seconds=300,  # 5 minutes
+                    reason=f"Malformed message: {err}",
+                )
                 await self.disconnect(DisconnectReason.bad_protocol)
                 return
 
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -84,6 +84,7 @@ from .constants import (
     CONN_IDLE_TIMEOUT,
     HEADER_LEN,
     MAC_LEN,
+    MINUTES_10,
     SNAPPY_PROTOCOL_VERSION,
 )
 
@@ -430,11 +431,6 @@ class BasePeer(BaseService):
                 )
                 return
             except MalformedMessage as err:
-                self.connection_tracker.record_blacklist(
-                    self.remote,
-                    timeout_seconds=300,  # 5 minutes
-                    reason=f"Malformed message: {err}",
-                )
                 await self.disconnect(DisconnectReason.bad_protocol)
                 return
 
@@ -641,6 +637,14 @@ class BasePeer(BaseService):
             raise ValueError(
                 f"Reason must be an item of DisconnectReason, got {reason}"
             )
+
+        if reason is DisconnectReason.bad_protocol:
+            self.connection_tracker.record_blacklist(
+                self.remote,
+                timeout_seconds=MINUTES_10,
+                reason=f"Bad protocol",
+            )
+
         self.logger.debug("Disconnecting from remote peer %s; reason: %s", self.remote, reason.name)
         self.base_protocol.send_disconnect(reason.value)
         self.disconnect_reason = reason
--- a/p2p/constants.py
+++ b/p2p/constants.py
@@ -126,3 +126,6 @@ REQUEST_PEER_CANDIDATE_TIMEOUT = 1
 
 # The maximum number of concurrent attempts to establis new peer connections
 MAX_CONCURRENT_CONNECTION_ATTEMPTS = 10
+
+# Time values in seconds
+MINUTES_10 = 60 * 10


p2p/peer.py
```diff
@@ -424,6 +424,11 @@ def is_closing(self) -> bool:
                 )
                 return
             except MalformedMessage as err:
+                self.connection_tracker.record_blacklist(
+                    self.remote,
+                    timeout_seconds=300,  # 5 minutes
```
carver: Magic number seems like it belongs in `p2p.constants`.

p2p/peer.py
```diff
@@ -424,6 +424,11 @@ def is_closing(self) -> bool:
                 )
                 return
             except MalformedMessage as err:
+                self.connection_tracker.record_blacklist(
```
carver: Actually, maybe this whole call belongs inside `peer.disconnect`, so that we always blacklist on every `DisconnectReason.bad_protocol`. Otherwise, every different type of disconnect has to remember to blacklist and decide how long to blacklist for.

It does maybe mean passing in an optional message to `disconnect()` describing the reason, which seems a better tradeoff than having to remember to call `record_blacklist()` at every disconnect call.

pipermerriam: I'm of the opposite opinion.  My reasoning.

I'm a little uncomfortable at this stage assuming that every `bad_protocol` disconnect means we should blacklist. My doubt comes from fear of being too heavy handed with blacklisting such that it negatively effects peer performance.  This may be too cautious or paranoid.

Also, while we do have places in the `Peer` where we have peer functionality coupled with blacklisting, the `disconnect()` method feels like it shouldn't have the blacklisting business logic coupled.

Thoughts?

carver: > I'm a little uncomfortable at this stage assuming that every `bad_protocol` disconnect means we should blacklist. My doubt comes from fear of being too heavy handed with blacklisting such that it negatively effects peer performance. This may be too cautious or paranoid.

I'm having trouble imagining a case in which a peer has violated the network protocol and it's important to reconnect to them again within 5 minutes. If it's a bad enough violation to disconnect, it seems bad enough to stay disconnected for a little while.

(If it's *not* bad enough to disconnect, then maybe we should do something else instead, like drop them from the queue of `WaitingPeer`s for that particular type of request).

> Also, while we do have places in the `Peer` where we have peer functionality coupled with blacklisting, the `disconnect()` method feels like it shouldn't have the blacklisting business logic coupled.

Because I'm having a hard time imagining wanting to disconnect for a protocol violation but *preferring* that we can reconnect within 5 minutes, it seems like the logic is inherently coupled.

There's already one example of a disconnect that deserves a blacklist that wasn't included in this PR yet:
https://github.com/ethereum/trinity/blob/c589b8ec17df51e4fa3b5be4e6582f645d6e2a92/trinity/sync/full/chain.py#L818-L827

I only found the one, but it seems very likely that the next `disconnect()` we add will forget too.

In some ways, I would find it more surprising that we would disconnect from a peer for being naughty and then allow them to reconnect right away, than that we would "automatically" blacklist for you.

So I guess I feel moderately strongly that this is leaving room for naughty peers to keep reconnecting to us and wasting our time, when we forget to blacklist.

---

Perhaps an alternative is to split `disconnect` into two methods:
- `.hang_up(DisconnectReason)` -- just disconnect, because you want to end the conversation. I think it's reasonably to `raise` on a `bad_protocol` reason here.
- `.ban(DisconnectReason)` -- disconnect and add to blacklist, maybe `raise` if a silly reason like `disconnect_requested`.

pipermerriam: Alright, I'm convinced enough.

pipermerriam:
@carver made requested changes, if you can take another :eyes: I'd like to get this resolved so I can build the other PR's that ban peers off of it.

p2p/peer.py
```diff
@@ -84,6 +84,7 @@
     CONN_IDLE_TIMEOUT,
     HEADER_LEN,
     MAC_LEN,
+    MINUTES_10,
```
carver: I was imagining something semantically linked, like `BAD_PROTOCOL_PENALTY`

p2p/peer.py
```diff
@@ -636,6 +637,14 @@ def _disconnect(self, reason: DisconnectReason) -> None:
             raise ValueError(
                 f"Reason must be an item of DisconnectReason, got {reason}"
             )
+
+        if reason is DisconnectReason.bad_protocol:
+            self.connection_tracker.record_blacklist(
+                self.remote,
+                timeout_seconds=MINUTES_10,
+                reason=f"Bad protocol",
```
carver: Doesn't need to be an f-string anymore, I guess.

pipermerriam:
doing cleanup in the other PRs

PR #561, opened 2019-04-30
Add test that scans Trinity shutdown for errors

cburgdorf:
### What was wrong?

We currently do not have a test that ensures trinity exists cleanly.

### How was it fixed?

Added a test that does the following:

1. Run Trinity
2. Wait until the syncer starts its job
3. Send a `SIGINT`
4. Scan for errors in the following output

Notice that the test is currently marked as `xfail`. When we get rid of the failures, we'll remove the `xfail` so that the test can ensure ongoing clean exits.

#### Cute Animal Picture

### DIFF
--- a/trinity/tools/async_process_runner.py
+++ b/trinity/tools/async_process_runner.py
@@ -65,8 +65,8 @@ class AsyncProcessRunner():
         self.kill()
         raise TimeoutError(f'Killed process after {timeout_sec} seconds')
 
-    def kill(self) -> None:
+    def kill(self, sig: int = signal.SIGKILL) -> None:
         try:
-            os.killpg(os.getpgid(self.proc.pid), signal.SIGKILL)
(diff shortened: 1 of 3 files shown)


trinity/tools/async_process_runner.py
```diff
@@ -65,8 +65,14 @@ def __init__(self, debug_fn: Callable[[bytes], None] = None) -> None:
         self.kill()
         raise TimeoutError(f'Killed process after {timeout_sec} seconds')
 
+    def interupt(self) -> None:
+        self.signal_process_group(signal.SIGINT)
+
     def kill(self) -> None:
```
pipermerriam: we *could* mirror the `os.kill` API and just let this function take an optional `signal` parameter that defaults to `signal.SIGKILL`

cburgdorf: That makes sense :+1:  Updated!

PR #564, opened 2019-04-30
Extract sub proto selection to utility

pipermerriam:
### What was wrong?

The mechanism for selecting sub protocols is burred within the `BasePeer` class, requiring a lot of scaffolding to test it.

### How was it fixed?

Extracted it into a utility function that can be tested in isolation.

- [x] TODO: update tests (and write some more)

#### Cute Animal Picture

85cccf0bee1fb44fe8745079fb83887c


p2p/protocol.py
```diff
@@ -175,6 +178,9 @@ class BaseRequest(ABC, Generic[TRequestPayload]):
     response_type: Type[Command]
 
 
+TCapability = Tuple[str, int]
```
lithp: Same thing here, I *think* it's convention for `T*` to mean `TypeVar`, and `CapabilityType` might be more conventional

p2p/protocol.py
```diff
@@ -210,10 +216,61 @@ def send_request(self, request: BaseRequest[PayloadType]) -> None:
...
+    self.sub_proto.
+
+    Raise NoMatchingPeerCapabilities if none of our supported protocols match one of the
+    remote's protocols.
+
+    Raise NotImplementedError if all protocols do not share the same name.
+    """
+    proto_names = set(proto.name for proto in protocols)
+    if len(proto_names) != 1:
+        raise NotImplementedError(
```
pipermerriam: Note: this will change when we support multiple simultaneously connected protocols.

p2p/protocol.py
```diff
@@ -210,10 +216,61 @@ def send_request(self, request: BaseRequest[PayloadType]) -> None:
...
+
+def select_sub_protocol(protocols: Sequence[Type[Protocol]],
+                        capabilities: CapabilitiesType) -> Type[Protocol]:
+    """
+    Return the highest version protocol from the provided `protocols` based on
+    the provided remote `capabilities`.
+
+    Return the protocol with the highest version that is present in the
+    remote and stores an instance of it (with the appropriate cmd_id offset) in
+    self.sub_proto.
```
lithp: The part about storing an instance in `self.sub_proto` is no longer true.

p2p/protocol.py
```diff
@@ -210,10 +216,61 @@ def send_request(self, request: BaseRequest[PayloadType]) -> None:
...
+            f" - {tuple(sorted(supported_capabilities))}\n"
+            f" - {tuple(sorted(capabilities))}"
+        )
+
+    for match in matching_capabilities:
+        for proto in protocols:
+            if proto.as_capability() == match:
+                return proto
+
+    raise NoMatchingPeerCapabilities()
```
lithp: This function would be so easily expressed with SQL:

```sql
SELECT max(local.as_capability.version)
FROM local_protocols AS local
  JOIN remote_capabilities AS remote ON (local.as_capability.version = remote.version);
```

What do you think of an implementation which emulates the SQL more directly?

```python
    proto_names = set(proto.name for proto in protocols)
    if len(proto_names) != 1:
        raise NotImplementedError(
            "Negotiation of capabilities across multiple protocol names not "
            "implemented"
        )

    for proto in sorted(protocols, key=operator.attrgetter('version'), reverse=True):
        if proto.as_capability() in capabilities:
            return proto

    supported_capabilities = (proto.as_capability() for proto in protocols)
    raise NoMatchingPeerCapabilities(
        f"No matching capabilities:\n"
        f" - {tuple(sorted(supported_capabilities))}\n"
        f" - {tuple(sorted(capabilities))}"
    )
```

In particular, it doesn't seem necessary to throw away the protocol only to look for it again later!

p2p/protocol.py
```diff
@@ -210,10 +216,61 @@ def send_request(self, request: BaseRequest[PayloadType]) -> None:
...
+
     def __repr__(self) -> str:
         return "(%s, %d)" % (self.name, self.version)
 
 
+CapabilitiesType = Tuple[CapabilityType, ...]
+
+
+def select_sub_protocol(protocols: Sequence[Type[Protocol]],
+                        capabilities: CapabilitiesType) -> Type[Protocol]:
```
lithp: I'm having a hard time putting into words why I think this signature feels weird. I really like how it's invoked:

```python
            sub_proto_class = select_sub_protocol(
                self.supported_sub_protocols,
                remote_capabilities,
            )
```

But once you're inside the method that information is thrown away, it's not clear what the difference between the protocols and capabilities are. Maybe this function would read easier if the parameters were `local_protocols` and `remote_capabilities`?

Alternatively, it doesn't *really* matter which one is remote and which is local, the operation is symmetric! This function is just looking for the highest version number the two have in common. Instead of hiding that information behind `select_sub_protocol`, maybe the interface could look like:

```python
negotiated_capability = highest_matching_capability(
    remote_capabilities,
    self.supported_capabilities,
)
sub_proto_class = Protocol.for_capability(negotiated_capability)
```

This would break up the work that `select_sub_protocol` does into three places:
- `@property Protocol.supported_capabilities`
- `highest_matching_capability`
- and `Protocol.for_capability`

Which is more moving parts, but maybe that's worth it if each part does something straightforward?

p2p/protocol.py
```diff
@@ -210,10 +216,61 @@ def send_request(self, request: BaseRequest[PayloadType]) -> None:
...
+    the provided remote `capabilities`.
+
+    Return the protocol with the highest version that is present in the
+    remote and stores an instance of it (with the appropriate cmd_id offset) in
+    self.sub_proto.
+
+    Raise NoMatchingPeerCapabilities if none of our supported protocols match one of the
+    remote's protocols.
+
+    Raise NotImplementedError if all protocols do not share the same name.
```
lithp: *nitpick*: "If all protocols do not share the same name" sounds like it'll throw an exception if the names are (A, B, C), but also sounds like it won't throw an exception if the names are (A, B, B), they're not all different! Maybe "unless all protocols share the same name"

p2p/protocol.py
pipermerriam: I went with a different approach.  Tell me what you think.  After looking it over it actually felt less complex to make this function simpler and only handle the correct matching rules and to move the error case of no  match or multiple matches up to the `Protocol` class.

Tell me if you think the new implementation is adequately legible.

p2p/protocol.py
```diff
@@ -236,48 +239,34 @@ def __repr__(self) -> str:
...
-        raise NotImplementedError(
-            "Negotiation of capabilities across multiple protocol names not "
-            "implemented"
+    # make a set for faster inclusion checks
+    capabilities_set = set(capabilities)
+
+    # group the protocols by name
+    proto_groups = groupby(operator.attrgetter('name'), protocols)
+    for _, homogenous_protocols in sorted(proto_groups.items()):
+        # sort the commonly named protocols by decreasing version number.
```
lithp: Cool! A couple thoughts:
- This is better than the original and also better than the factorings I suggested, I think it makes a lot of sense for `process_p2p_handshake` to handle all the exceptions.
- I had never seen `@to_tuple` before using this code base but I think that from now on it'll play a part in all my python projects, this style of yielding your results leads to such great code.
- I read through the rlpx spec and think this is an eminently reasonable interpretation.
- My only remaining complaint is this comment, I'm not sure what "commonly named protocols" means, and think the code is clear enough that it might be okay to drop this comment entirely.

pipermerriam: Will convert `commonly-named` to be `same-named` or some facsimile.

lithp:
Wait!

lithp:
I just realized there's no cute animal picture. We can't merge without one :)

pipermerriam:
that was almost a disaster....good catch

PR #565, opened 2019-04-30
Remove mutable data structures from p2p.protocol classes

pipermerriam:
### What was wrong?

The `p2p.protocol.Command` and `p2p.protocol.Protocol` classes were defined with mutable data structures.

### How was it fixed?

Replaced them with immutable data structures.

#### Cute Animal Picture

e99379dd8b2cc14ec34e75988feac46d

### DIFF
--- a/p2p/protocol.py
+++ b/p2p/protocol.py
@@ -53,16 +53,15 @@ TRequestPayload = TypeVar('TRequestPayload', bound=PayloadType, covariant=True)
 _DecodedMsgType = PayloadType
 
 
-TStructure = Union[
+StructureType = Union[
     Tuple[Tuple[str, Any], ...],
-    sedes.CountableList,
 ]
 
 
 class Command:
     _cmd_id: int = None
     decode_strict = True
-    structure: TStructure
+    structure: StructureType
 
     _logger: logging.Logger = None
 
@@ -85,7 +84,7 @@ class Command:
         return f"{type(self).__name__} (cmd_id={self.cmd_id})"
 
     def encode_payload(self, data: Union[PayloadType, sedes.CountableList]) -> bytes:
-        if isinstance(data, dict):  # convert dict to ordered tuple
+        if isinstance(data, dict):
             if not isinstance(self.structure, tuple):
                 raise ValueError(
                     "Command.structure must be a list when data is a dict.  Got "
@@ -97,7 +96,8 @@ class Command:
                 raise ValueError(
                     f"Keys in data dict ({data_keys}) do not match expected keys ({expected_keys})"
                 )
-            data = [data[name] for name, _ in self.structure]
+            data = tuple(data[name] for name, _ in self.structure)
+
         if isinstance(self.structure, sedes.CountableList):
             encoder = self.structure
         else:
--- a/p2p/protocol.py
+++ b/p2p/protocol.py
@@ -53,10 +53,16 @@ TRequestPayload = TypeVar('TRequestPayload', bound=PayloadType, covariant=True)
 _DecodedMsgType = PayloadType
 
 
+TStructure = Union[
+    Tuple[Tuple[str, Any], ...],
+    sedes.CountableList,
+]
+
+
 class Command:
     _cmd_id: int = None
     decode_strict = True
-    structure: List[Tuple[str, Any]] = []
+    structure: TStructure
 
     _logger: logging.Logger = None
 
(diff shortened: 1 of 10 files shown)


p2p/protocol.py
```diff
@@ -53,10 +53,16 @@ class TypedDictPayload(TypedDict):
 _DecodedMsgType = PayloadType
 
 
+TStructure = Union[
+    Tuple[Tuple[str, Any], ...],
+    sedes.CountableList,
+]
```
lithp: I've usually seen the `T*` convention used to indicate that this is or contains a `TypeVar`. `StructureType` might be better.

lithp: Also, frustratingly enough, mypy doesn't seem to be doing its job here. If you remove `Union` and just use `Tuple[Tuple[str, Any], ...]`, then you'll discover that `NewBeaconBlock` in `trinity/protocol/bcc/commands.py` still uses a list where it should use a tuple. Once you fix that, you'll discover that mypy shows success, even though some classes definitely use `sedes.CountableList`, which is definitely not a tuple!

Well, the immediate fix is to also change `NewBeaconBlock`. I'm experimenting with doing PR reviews faster so I won't look into why mypy is messing up here, but that might be worth doing eventually, and the long-term fix might be to encode `structure` differently:

```python
structure = sedes.Elements(
  ('encoded_block', sedes.binary),
)
```

So that we don't need to use `Union` any more.

p2p/protocol.py
```diff
@@ -79,9 +85,12 @@ def __str__(self) -> str:
...
     def encode_payload(self, data: Union[PayloadType, sedes.CountableList]) -> bytes:
-        if isinstance(data, dict):  # convert dict to ordered list
-            if not isinstance(self.structure, list):
-                raise ValueError("Command.structure must be a list when data is a dict")
+        if isinstance(data, dict):  # convert dict to ordered tuple
+            if not isinstance(self.structure, tuple):
+                raise ValueError(
+                    "Command.structure must be a list when data is a dict.  Got "
+                    f"{self.structure}"
+                )
```
lithp: Since this wasn't triggered it might also be worth adding a test that `NewBeaconBlock` can be successfully encoded.

p2p/protocol.py
```diff
@@ -79,9 +85,12 @@ def __str__(self) -> str:
         return f"{type(self).__name__} (cmd_id={self.cmd_id})"
 
     def encode_payload(self, data: Union[PayloadType, sedes.CountableList]) -> bytes:
-        if isinstance(data, dict):  # convert dict to ordered list
-            if not isinstance(self.structure, list):
-                raise ValueError("Command.structure must be a list when data is a dict")
+        if isinstance(data, dict):  # convert dict to ordered tuple
```
lithp: *nitpick*: "convert dict to ordered tuple", all tuples are ordered and it looks like this is still converting `data` into a list (line 97)

p2p/protocol.py
```diff
@@ -180,8 +189,8 @@ class Protocol:
     name: str = None
     version: int = None
     cmd_length: int = None
-    # List of Command classes that this protocol supports.
-    _commands: List[Type[Command]] = []
+    # Command classes that this protocol supports.
+    _commands: Tuple[Type[Command], ...]
```
lithp: Was it intentional that this no longer defaults to the empty list? I think that the `self.commands` line in `__init__` means that any subclasses which don't define `_commands` will throw an exception upon initialization.

p2p/protocol.py
pipermerriam: I believe the mypy failure is because the rlp library doesn't expose type hints so it interprets `sedes.CountableList` as an `Any` type (just intuition talking here).  I'll get this stuff cleaned up best I can.

p2p/protocol.py
pipermerriam: Yes, I did intentionally remove the default.  It seems faulty to define a protocol without any commands and having a default seems foot-gun-y

p2p/protocol.py
pipermerriam: Future work, I'd like to update the rlp library to have a firm base class for `BaseSedes` and then we can use that.  Right now it uses inference about whether the `sedes` object has the appropriate methods on it which is sloppy.

p2p/protocol.py
pipermerriam: Going to skip this.  This is an area of code that I've wanted to fix for a while, removing the flexibility of input values and better type safety in general (probably requiring an RLP library update).

p2p/protocol.py
lithp: Sure thing, just wanted to check :+1: It might be nice to have a better error message than `'NoneType' object is not iterable` but that could be a future-PR thing.

p2p/protocol.py
cburgdorf: >type (just intuition talking here)

That is correct. In fact, there is a rule `--disallow-subclassing-any` that would reject this. In general, `mypy` assistance in Trinity is still far from what `mypy` is actually capable of. There are many checks that we can't turn on yet. E.g. we have a lot of decorator (*cough*, `curry`) that take beautifully typed functions and turn them into `def lalala(*args: Any) -> Any)`.

p2p/protocol.py
pipermerriam: *nit* it'll actually be an `AttributeError` since the property will be missing entirely.  Future PR though

PR #566, opened 2019-04-30
Extract low level peer networking into transport class

### LINKED ISSUE #90: Move functionality from `Peer` class down into `Protocol` class.

pipermerriam:
Replaces https://github.com/ethereum/trinity/pull/546

fixes #90

Based on: #563 #564 #565 which all must be merged first

### What was wrong?

The `p2p.BasePeer` class has been shown to contain *too much logic*.  This has had many resulting difficulties such as:

- harder to test
- harder to type
- unable to support multiple protocols

### How was it fixed?

This PR extracts the lowest level networking components from the `BasePeer` class out into a `Transport` class which has well constrained responsibilities

- reading and writing to the streams
- encryption and decryption

#### Cute Animal Picture

Gusbypail jpg w560h702

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -88,7 +88,7 @@ async def handshake(remote: Node, factory: 'BasePeerFactory') -> 'BasePeer':
     handshake or if none of the sub-protocols supported by us is also
     supported by the remote.
     """
-    transport = await Transport.handshake(
+    transport = await Transport.connect(
         remote,
         factory.privkey,
         factory.cancel_token,
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -75,7 +75,8 @@ if TYPE_CHECKING:
 
 
 async def handshake(remote: Node, factory: 'BasePeerFactory') -> 'BasePeer':
-    """Perform the auth and P2P handshakes with the given remote.
+    """
+    Perform the auth and P2P handshakes with the given remote.
 
     Return an instance of the given peer_class (must be a subclass of
     BasePeer) connected to that remote in case both handshakes are
@@ -87,7 +88,7 @@ async def handshake(remote: Node, factory: 'BasePeerFactory') -> 'BasePeer':
     handshake or if none of the sub-protocols supported by us is also
     supported by the remote.
     """
-    transport = await Transport.open_connection(
+    transport = await Transport.handshake(
         remote,
         factory.privkey,
         factory.cancel_token,
@@ -205,9 +206,6 @@ class BasePeer(BaseService):
 
     def __repr__(self) -> str:
         return f"{self.__class__.__name__} {self.remote!r}"
-
-    def __hash__(self) -> int:
-        return hash(self.remote)
 
     #
     # Proxy Transport attributes
@@ -357,7 +355,6 @@ class BasePeer(BaseService):
                 return
 
     async def read_msg(self) -> Tuple[Command, PayloadType]:
-        # TODO: Transport class needs a cancel token for proper cancellation
         msg = await self.transport.recv(self.cancel_token)
         cmd = self.get_protocol_command_for(msg)
         # NOTE: This used to be a bottleneck but it doesn't seem to be so anymore. If we notice
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -4,7 +4,6 @@ import collections
 import contextlib
 import functools
 import logging
-import struct
 from typing import (
     Any,
     cast,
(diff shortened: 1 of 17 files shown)


pipermerriam:
There are a few dependent PR's that need to be merged but this branch is effectively complete if anyone wants to look over it.  ONly last commit is relevant if you want to ignore all the other changes that are accounted for in other pull requests.

pipermerriam:
Tests are :green_heart: !

p2p/peer.py
```diff
@@ -430,28 +357,8 @@ def is_closing(self) -> bool:
...
-        frame_data = await self.read(read_size + MAC_LEN)
-        try:
-            msg = self.decrypt_body(frame_data, frame_size)
-        except DecryptionError as err:
-            self.logger.debug(
-                "Bad message body from peer %s: Error: %r",
-                self, err,
-            )
-            raise MalformedMessage from err
+        # TODO: Transport class needs a cancel token for proper cancellation
```
carver: Could tack that onto this PR now, if you want. Just seeing if we can avoid adding the `TODO` to master.

p2p/transport.py
```diff
@@ -0,0 +1,244 @@
...
+        """
+        if not self._reader.at_eof():
+            self._reader.feed_eof()
+        self._writer.close()
+
+    @property
+    def is_closing(self) -> bool:
+        return self._writer.transport.is_closing()
+
+    def _encrypt(self, header: bytes, frame: bytes) -> bytes:
```
carver: I'm assuming these methods were all a straight copy/paste from the old `Peer` methods, so I didn't really look. If there was a more substantive change, let me know and I'll take a deeper look.

trinity/protocol/common/validators.py
```diff
@@ -111,7 +111,7 @@ def _is_numbered(self) -> bool:
     def _get_formatted_params(self) -> str:
         identifier = self.block_number_or_hash
         return (
-            f'ident: {identifier if self._is_numbered else humanize_hash(identifier)}  '  # type: ignore  # mypy doesn't know that identifier is of the right type  # noqa: E501
+            f'ident: {identifier if self._is_numbered else humanize_hash(identifier)}  '  # type: ignore  # mypy doesn't know `identifier` is a Hash32  # noqa: E501
```
carver: One option would be to pass in the block hash as a `Hash32` parameter to the method, since the caller does the cast.

That seems a tiny bit safer, since someone might try to use `_get_formatted_params` elsewhere without pre-checking that the identifier is a hash.

p2p/peer.py
```diff
@@ -195,34 +140,31 @@ class BasePeer(BaseService):
...
-        self.reader = connection.reader
-        self.writer = connection.writer
         # Initially while doing the handshake, the base protocol shouldn't support
         # snappy compression
-        self.base_protocol = P2PProtocol(self, snappy_support=False)
+        self.base_protocol = P2PProtocol(
+            transport=self.transport,
+            snappy_support=False,
+            capabilities=self.capabilities,
+            listen_port=self.listen_port,
```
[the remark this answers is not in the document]
pipermerriam: @carver maybe... though it will have to live at whatever *thing* lives just above the protocol connections since capabilities spans across all supported protocols....

p2p/peer.py
```diff
@@ -271,6 +200,26 @@ def setup_connection_tracker(self) -> BaseConnectionTracker:
...
         return NoopConnectionTracker()
 
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__} {self.remote}"
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__} {self.remote!r}"
+
+    def __hash__(self) -> int:
+        return hash(self.remote)
```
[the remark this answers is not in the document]
pipermerriam: I'm not familiaar with the bug you're referring to but I do recall us learning that `hash` isn't stable across pickling (and thus the hashing of the `remote` won't be stable across pickling and thus... I don't know.

p2p/peer.py
pipermerriam: I actually already addressed this, just need to drop the `TODO`

p2p/transport.py
```diff
@@ -0,0 +1,244 @@
...
+
+        mac_cipher = Cipher(algorithms.AES(mac_secret), modes.ECB(), default_backend())
+        self._mac_enc = mac_cipher.encryptor().update
+
+    @classmethod
+    async def open_connection(cls,
+                              remote: Node,
+                              private_key: datatypes.PrivateKey,
+                              token: CancelToken) -> 'Transport':
+        # TODO: what do we name this method....
```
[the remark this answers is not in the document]
pipermerriam: maybe just `connect()`?  That is what it does, establish the connection to the remote.

p2p/transport.py
pipermerriam: copy/pasta

trinity/protocol/common/validators.py
pipermerriam: will clean up

p2p/transport.py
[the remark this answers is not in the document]
pipermerriam: now I want to call it `handshake`.... anyone have a strong opinion, otherwise it's going to be `handshake` or `connect` depending on which way the wind shifts.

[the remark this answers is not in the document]
carver: I don't love having the method name conflict with `auth.handshake()`, so I guess a slight preference for `connect()`, but not a strong opinion.

PR #571, opened 2019-04-30
Move uptime tracking to base service class

pipermerriam:
### What was wrong?

The peer class implemented an `uptime` property.  It was somewhat awkwardly implemented to return a string since it was only used for logging purposes.

### How was it fixed?

Maybe YAGNI but this felt cleaner, moving this onto `BaseService.uptime` and converting it to be a `datetime.timedelta`, then leveraging the `humanize_seconds` utility in the actual logging statement.

#### Cute Animal Picture

0b16cc1c64e61433e5d351c34ca94c81

### DIFF
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -1,9 +1,9 @@
 from abc import ABC, abstractmethod
 import asyncio
 import concurrent
-import datetime
 import functools
 import logging
+import time
 from typing import (
     Any,
     Awaitable,
@@ -47,6 +47,8 @@ class BaseService(ABC, CancellableMixin):
 
     _logger: ExtendedDebugLogger = None
 
+    _start_time: float = None
+
     def __init__(self,
                  token: CancelToken=None,
                  loop: asyncio.AbstractEventLoop = None) -> None:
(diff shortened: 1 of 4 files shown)


p2p/service.py
```diff
@@ -98,6 +103,7 @@ def get_event_loop(self) -> asyncio.AbstractEventLoop:
         try:
             async with self._run_lock:
                 self.events.started.set()
+                self.start_time = datetime.datetime.utcnow()
```
carver: Probably should use `time.monotonic()` to avoid weird results on system clock changes.

Also, I'd say make this `_start_time` so that people aren't as tempted to access directly.

p2p/service.py
```diff
@@ -73,6 +74,10 @@ def logger(self) -> ExtendedDebugLogger:
             )
         return self._logger
 
+    @property
+    def uptime(self) -> datetime.timedelta:
+        return datetime.datetime.utcnow() - self.start_time
```
carver: Looks like calling `service.uptime` before running causes an attribute error.

pipermerriam: :+1: will handle falling back to `0` if the service hasn't started.

PR #579, opened 2019-05-02
Remove `Record` suffix

### LINKED ISSUE #399: Remove `Record` from the various data types that still have it in the name.
### What is wrong?

Some types in our wonderful codebase have the suffix `Record` to indicate they are a bundle of data. This pattern IIRC was inherited from the specs repo. The suffix has since been dropped from the corresponding names in the spec and I would suggest we do so as well in `trinity`.

I see:

- [ ] `CrosslinkRecord`
- [ ] `PendingAttestationRecord`
- [ ] `ValidatorRecord`

Please add any others I've missed!

### How can it be fixed

Should be a search and replace operation.

hwwhww:
### What was wrong?
Fix #399

### How was it fixed?
Find and replace them.

#### Cute Animal Picture

🦑


hwwhww:
Thank you @mhchia!
CI record in my branch: https://github.com/hwwhww/trinity/commits/remove_record

ralexstokes:
yay!

PR #580, opened 2019-05-02
Add receive server for validator service

mhchia:
### What was wrong?
Our validator service has no corresponding handler for the command `NewBeaconBlock`.

### How was it fixed?
Based on `BCCReceiveServer`, the work of @NIC619 from https://github.com/ethereum/trinity/pull/465.
- Add `BCCReceiveServer` to define the handler for `NewBeaconBlock`.
- Add method `BCCReceiveServer._try_import_or_handle_orphan` to check and maintain orphan blocks in `BCCReceiveServer.orphan_block_pool`. Currently, it is a in-memory set. We can further optimize it for purposes through changing `orphan_block_pool._pool`.
- Add method `BCCReceiveServer._request_block_by_root` used by `BCCReceiveServer._try_import_or_handle_orphan` to request the parents of the orphans from peers.
- Add a handler for `BeaconBlocks` to handle the response of our sent requests.

#### Cute Animal Picture

### DIFF
--- a/tests/core/p2p-proto/bcc/test_receive_server.py
+++ b/tests/core/p2p-proto/bcc/test_receive_server.py
@@ -107,11 +107,12 @@ async def get_peer_and_receive_server(request, event_loop) -> Tuple[
         bob_chain_db=bob_chain.chaindb,
     )
 
-    # Add a queue to ensure messages put after every msg handler finishes.
-    # This is crucial to synchronize the test with the BCCReceiveServer.
     msg_queue = asyncio.Queue()
     orig_handle_msg = BCCReceiveServer._handle_msg
 
+    # Inject a queue to each `BCCReceiveServer`, which puts the message passed to `_handle_msg` to
+    # the queue, right after every `_handle_msg` finishes.
+    # This is crucial to make the test be able to wait until `_handle_msg` finishes.
     async def _handle_msg(self, base_peer, cmd, msg):
         task = asyncio.ensure_future(orig_handle_msg(self, base_peer, cmd, msg))
 
@@ -149,18 +150,23 @@ def test_orphan_block_pool():
     # test: add
     pool.add(b1)
     assert b1 in pool._pool
+    assert len(pool._pool) == 1
     # test: add: no side effect for adding twice
     pool.add(b1)
+    assert len(pool._pool) == 1
     # test: add: two blocks
     pool.add(b2)
+    assert len(pool._pool) == 2
     # test: get
     assert pool.get(b1.signing_root) == b1
     assert pool.get(b2.signing_root) == b2
     # test: pop_children
     b2_children = pool.pop_children(b2)
     assert len(b2_children) == 0
+    assert len(pool._pool) == 2
     b0_children = pool.pop_children(b0)
     assert len(b0_children) == 2 and (b1 in b0_children) and (b2 in b0_children)
+    assert len(pool._pool) == 0
 
 
 @pytest.mark.asyncio
--- a/tests/core/p2p-proto/bcc/test_receive_server.py
+++ b/tests/core/p2p-proto/bcc/test_receive_server.py
@@ -77,8 +77,26 @@ async def get_fake_chain() -> FakeChain:
     return FakeChain(chain_db.db)
 
 
+def get_blocks(
+        receive_server: BCCReceiveServer,
+        parent_block: SerenityBeaconBlock = None,
+        num_blocks: int = 3) -> Tuple[SerenityBeaconBlock, ...]:
+    chain = receive_server.chain
+    if parent_block is None:
+        parent_block = chain.get_canonical_head()
+    blocks = []
+    for _ in range(num_blocks):
+        block = chain.create_block_from_parent(
+            parent_block=parent_block,
+            block_params=FromBlockParams(),
+        )
+        blocks.append(block)
+        parent_block = block
+    return tuple(blocks)
+
+
 async def get_peer_and_receive_server(request, event_loop) -> Tuple[
-        BCCPeer, BCCReceiveServer]:
+        BCCPeer, BCCRequestServer, BCCReceiveServer, asyncio.Queue]:
     alice_chain = await get_fake_chain()
     bob_chain = await get_fake_chain()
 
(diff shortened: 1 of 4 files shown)


mhchia:
Tests failed in https://circleci.com/gh/mhchia/trinity/1857. It seems to be related to the change from `signed_root` to `signing_root`. Will investigate it tomorrow.

trinity/protocol/bcc/servers.py
```diff
@@ -133,3 +151,162 @@ def __init__(self,
...
+        # TODO: relay the block if it is valid
+
+    def _try_import_or_handle_orphan(self, block: BeaconBlock) -> None:
+        blocks_to_be_imported: List[BeaconBlock] = []
+
+        blocks_to_be_imported.append(block)
+        while len(blocks_to_be_imported) != 0:
+            block = blocks_to_be_imported.pop()
+            # try to import the block
+            if not self._is_block_root_in_db(block.previous_block_root):
```
NIC619: We can also check `if block.previous_block_root in self._is_block_root_in_orphan_block_pool`

trinity/protocol/bcc/servers.py
```diff
@@ -133,3 +151,162 @@ def __init__(self,
...
+                self.logger.debug(
+                    f"blocks {matched_orphan_blocks} match their parent {block}"
+                )
+            blocks_to_be_imported.extend(matched_orphan_blocks)
+
+    def _request_block_by_root(self, block_root: Hash32) -> None:
+        for i, peer in enumerate(self._peer_pool.connected_nodes.values()):
+            peer = cast(BCCPeer, peer)
+            self.logger.debug(
+                bold_red(f"send block request to: request_id={i}, peer={peer}")
```
NIC619: Shouldn't `request_id` be `req_request_id`?

trinity/protocol/bcc/servers.py
```diff
@@ -133,3 +151,162 @@ def __init__(self,
...
+            )
+            req_request_id = random.randint(0, 32768)
```
NIC619: Why not just `req_request_id`?

NIC619: And perhaps we can just reuse `gen_request_id` in `/trinity/trinity/_utils/les.py`.

trinity/protocol/bcc/servers.py
```diff
@@ -133,3 +151,162 @@ def __init__(self,
...
+            raise Exception(f"Invariant: Only subscribed to {self.subscription_msg_types}")
+
+    async def _handle_beacon_blocks(self, peer: BCCPeer, msg: BeaconBlocksMessage) -> None:
+        if not peer.is_operational:
+            return
+        request_id = msg["request_id"]
+        if request_id not in self.map_requested_id_block_root:
+            raise Exception(f"request_id={request_id} is not found")
+        encoded_blocks = msg["encoded_blocks"]
+        if len(encoded_blocks) != 1:
```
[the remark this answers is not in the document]
mhchia: Sure. Added TODO both here and the check in the test.

trinity/protocol/bcc/servers.py
```diff
@@ -133,3 +151,162 @@ def __init__(self,
...
+            raise Exception("should only receive 1 block from our requests")
+        resp_block = ssz.decode(encoded_blocks[0], BeaconBlock)
+        if resp_block.signed_root != self.map_requested_id_block_root[request_id]:
+            raise Exception(
+                f"block root {resp_block.signed_root} does not correpond to the one we requested"
+            )
+        self.logger.debug(f"received request_id={request_id}, resp_block={resp_block}")  # noqa: E501
```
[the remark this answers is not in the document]
mhchia: Nice catch!

trinity/protocol/bcc/servers.py
mhchia: Yes, you're right. Have changed it.

trinity/protocol/bcc/servers.py
mhchia: Good idea. Changed it.

trinity/protocol/bcc/servers.py
mhchia: Sorry, I don't quite understand. Could you elaborate more on the advantage to check `if block.previous_block_root in self._is_block_root_in_orphan_block_pool` over `if not self._is_block_root_in_db(block.previous_block_root)`?

tests/core/p2p-proto/bcc/test_receive_server.py
```diff
@@ -0,0 +1,396 @@
...
+        request,
+        event_loop,
+    )
+    alice_msg_buffer = MsgBuffer()
+    alice.add_subscriber(alice_msg_buffer)
+    bob_chain = bob_recv_server.chain
+    head = bob_chain.get_canonical_head()
+    block_0 = bob_chain.create_block_from_parent(
+        parent_block=head,
+        block_params=FromBlockParams(),
```
ChihChengLiang: This 3 blocks pattern can be extracted as a function for readability.

mhchia: Good idea. Added `get_blocks` to do this.

trinity/protocol/bcc/servers.py
```diff
@@ -133,3 +151,162 @@ def __init__(self,
...
+        if not peer.is_operational:
+            return
+        request_id = msg["request_id"]
+        if request_id not in self.map_requested_id_block_root:
+            raise Exception(f"request_id={request_id} is not found")
+        encoded_blocks = msg["encoded_blocks"]
+        if len(encoded_blocks) != 1:
+            raise Exception("should only receive 1 block from our requests")
+        resp_block = ssz.decode(encoded_blocks[0], BeaconBlock)
+        if resp_block.signed_root != self.map_requested_id_block_root[request_id]:
```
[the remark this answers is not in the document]
mhchia: I changed to compare `self.map_requested_id_block_root[request_id]` with `resp_block.signing_root`, because `signing_root` is used as the block hash in db.

Reference: https://github.com/ethereum/eth2.0-specs/pull/816

mhchia:
Updated the PR according to the review feedback.

tests/core/p2p-proto/bcc/test_receive_server.py
```diff
@@ -0,0 +1,369 @@
...
+def test_orphan_block_pool():
+    pool = OrphanBlockPool()
+    b0 = create_test_block()
+    b1 = create_test_block(parent=b0)
+    b2 = create_test_block(parent=b0, state_root=b"\x11" * 32)
+    # test: add
+    pool.add(b1)
+    assert b1 in pool._pool
+    # test: add: no side effect for adding twice
+    pool.add(b1)
```
NIC619: Add a check on length of `pool._pool`?

trinity/protocol/bcc/servers.py
NIC619: I meant we can first check if it's parent is in the orphan peer pool then next check if it's parent is in the db. This helps by not having to lookup the db in case where it's parent is already in the pool.
But after a second thought this doesn't really help much(or even worse this adds another lookup into orphan pool) since case like this is probably rare.

PR #581, opened 2019-05-02
Add `black` code fmt

ralexstokes:
### What was wrong?

I was talking to @pipermerriam some time ago and was lamenting the process where I have to manually conform to some code style.

He mentioned an option to get around this is `black`: https://github.com/python/black.

The intent w/ this PR is to start a discussion around if we want to start incorporating a syntax formatter into our repos or not.

### How was it fixed?

1. Runs `black` over all the code in this repo.
2. Adds `black --check` to the CI so that if code is committed that does not follow the formatter, it will be rejected by the CI (ensures uniformity).

It looks like the `black` conventions clash with a number of our existing linting rules (and mypy gets upset...).

I'm happy to take the charge on finding a nice union of these tools but wanted to gauge if we have more support before I move ahead on that work. 

### Notes

If we move ahead here, I would suggest either adding a git hook to run `black` at some point in your workflow, or possibly even better, integrating it into some "format on save" feature in your lovely editor. There are some common editor integrations in the `black` readme.

#### Cute Animal Picture


ralexstokes:
i'll add a `blocked` label to indicate that effectively this 'should not be merged'.

we can change this after some discussion.

pipermerriam:
I'm :+1: on this, but going to go through the actual changes now and see if there are any sections that seem to justify adding `# fmt: off/on` to maintain style for readability.

eth2/_utils/blobs.py
```diff
@@ -53,9 +36,9 @@ def calc_chunk_root(collation_body: bytes) -> Hash32:
 
 def check_body_size(body: bytes) -> bytes:
     if len(body) != COLLATION_SIZE:
-        raise ValidationError("{} byte collation body exceeds maximum allowed size".format(
-            len(body)
-        ))
+        raise ValidationError(
+            "{} byte collation body exceeds maximum allowed size".format(len(body))
```
pipermerriam: *note* can be updated to f-string

eth2/_utils/blobs.py
```diff
@@ -68,7 +51,9 @@ def serialize_blobs(blobs: Iterable[bytes]) -> Iterator[bytes]:
         if len(blob) == 0:
             raise ValidationError("Cannot serialize blob {} of length 0".format(i))
         if len(blob) > MAX_BLOB_SIZE:
-            raise ValidationError("Cannot serialize blob {} of size {}".format(i, len(blob)))
+            raise ValidationError(
+                "Cannot serialize blob {} of size {}".format(i, len(blob))
```
pipermerriam: *note* can be updated to f-string

eth2/_utils/merkle/common.py
```diff
@@ -71,22 +57,22 @@ def get_merkle_proof(tree: MerkleTree, item_index: int) -> Iterable[Hash32]:
         return ()
 
     branch_indices = get_branch_indices(item_index, len(tree))
-    proof_indices = [i ^ 1 for i in branch_indices][:-1]  # get sibling by flipping rightmost bit
+    proof_indices = [i ^ 1 for i in branch_indices][
+        :-1
+    ]  # get sibling by flipping rightmost bit
```
pipermerriam: Icky formatting, maybe relocate comment to previous line so black doesn't do this?

eth2/_utils/ssz.py
```diff
@@ -36,65 +30,52 @@ def diff_ssz_object(left: BaseBeaconBlock,
...
-                obj_b,
+                obj_a_name, obj_a, obj_b_name, obj_b
             )
         )
     longest_field_name = max(len(field_name) for field_name, _, _ in diff)
-    error_message = (
-        "Mismatch between {obj_a_name} and {obj_b_name} on {0} fields:\n - {1}".format(
-            len(diff),
-            "\n - ".join(tuple(
+    error_message = "Mismatch between {obj_a_name} and {obj_b_name} on {0} fields:\n - {1}".format(
```
pipermerriam: Maybe justifies f-string cleanup.

eth2/beacon/chains/base.py
```diff
@@ -273,15 +248,17 @@ def get_block_class(self, block_root: Hash32) -> Type[BaseBeaconBlock]:
...
+        self, parent_block: BaseBeaconBlock, block_params: FromBlockParams
+    ) -> BaseBeaconBlock:
         """
         Passthrough helper to the ``StateMachine`` class of the block descending from the
         given block.
         """
         return self.get_state_machine_class_for_block_slot(
-            slot=parent_block.slot + 1 if block_params.slot is None else block_params.slot,
+            slot=parent_block.slot + 1
+            if block_params.slot is None
```
pipermerriam: maybe extract ternary statement since this is kinda weird readability

eth2/beacon/committee_helpers.py
```diff
@@ -56,25 +25,31 @@
...
-            active_validator_count // slots_per_epoch // target_committee_size,
+    active_validator_count: int,
+    shard_count: int,
+    slots_per_epoch: int,
+    target_committee_size: int,
+) -> int:
+    return (
+        max(
+            1,
+            min(
```
pipermerriam: could be re-written to use `eth_utils.clamp` and probably be more readable.

eth2/beacon/helpers.py
```diff
@@ -101,60 +78,52 @@ def _get_historical_root(
         )
     if slot >= state_slot:
         raise ValidationError(
-            "slot ({}) should be less than state.slot ({})".format(
-                slot,
-                state_slot,
-            )
+            "slot ({}) should be less than state.slot ({})".format(slot, state_slot)
```
pipermerriam: f-string

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -378,39 +337,41 @@ def validate_attestation_source_epoch_and_root(state: BeaconState,
             raise ValidationError(
                 "Current epoch attestation that "
                 "`source_epoch` is not targeting the `state.current_justified_epoch`:\n"
-                "\tFound: %s, Expected %s" %
-                (attestation_data.source_epoch, state.current_justified_epoch)
+                "\tFound: %s, Expected %s"
+                % (attestation_data.source_epoch, state.current_justified_epoch)
```
pipermerriam: f-string

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -378,39 +337,41 @@ def validate_attestation_source_epoch_and_root(state: BeaconState,
...
             )
 
         if attestation_data.source_root != state.current_justified_root:
             raise ValidationError(
                 "Current epoch attestation that "
                 "`source_root` is not equal to `state.current_justified_root`:\n"
-                "\tFound: %s, Expected %s" %
-                (attestation_data.source_root, state.current_justified_root)
+                "\tFound: %s, Expected %s"
+                % (attestation_data.source_root, state.current_justified_root)
```
pipermerriam: f-string

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -378,39 +337,41 @@ def validate_attestation_source_epoch_and_root(state: BeaconState,
...
     else:
         # Case 2: previous epoch attestations
         if attestation_data.source_epoch != state.previous_justified_epoch:
             raise ValidationError(
                 "Previous epoch attestation that "
                 "`source_epoch`` is not targeting the `state.previous_justified_epoch`:\n"
-                "\tFound: %s, Expected %s" %
-                (attestation_data.source_epoch, state.previous_justified_epoch)
+                "\tFound: %s, Expected %s"
+                % (attestation_data.source_epoch, state.previous_justified_epoch)
```
pipermerriam: f-string

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -378,39 +337,41 @@ def validate_attestation_source_epoch_and_root(state: BeaconState,
...
             )
 
         if attestation_data.source_root != state.previous_justified_root:
             raise ValidationError(
                 "Previous epoch attestation that "
                 "`source_root` is not equal to `state.previous_justified_root`:\n"
-                "\tFound: %s, Expected %s" %
-                (attestation_data.source_root, state.previous_justified_root)
+                "\tFound: %s, Expected %s"
+                % (attestation_data.source_root, state.previous_justified_root)
```
pipermerriam: f-string

eth2/beacon/state_machines/forks/serenity/operation_processing.py
```diff
@@ -144,7 +130,10 @@ def process_attestations(state: BeaconState,
     new_previous_epoch_pending_attestations = []
     new_current_epoch_pending_attestations = []
     for attestation in block.body.attestations:
-        if slot_to_epoch(attestation.data.slot, config.SLOTS_PER_EPOCH) == current_epoch:
+        if (
+            slot_to_epoch(attestation.data.slot, config.SLOTS_PER_EPOCH)
```
pipermerriam: Looks like the return value of this could be cached (pre-computed) which would allow these lines to be shorter and not end up being broken across multiple lines.

eth2/beacon/tools/misc/ssz_vector.py
```diff
@@ -26,6 +24,6 @@ def override_vector_lengths(config: Eth2Config) -> None:
     for key, value in historical_batch_vector_dict.items():
         HistoricalBatch._meta.container_sedes.field_name_to_sedes[key].length = value
 
-    Deposit._meta.container_sedes.field_name_to_sedes['proof'].length = (
-        config.DEPOSIT_CONTRACT_TREE_DEPTH
-    )
+    Deposit._meta.container_sedes.field_name_to_sedes[
+        "proof"
+    ].length = config.DEPOSIT_CONTRACT_TREE_DEPTH
```
pipermerriam: *not part of this PR but...*  

This appears to mutate the `Deposit` class's definition.  That seems very dangerous.  At minimum, we should update it to use something like pytest's `monkeypatch` which handles reversion back to the original value...

libp2p/p2pclient/pb/crypto_pb2.py
```diff
@@ -60,101 +58,146 @@
...
+    ),
+)
 _sym_db.RegisterMessage(PublicKey)
 
-PrivateKey = _reflection.GeneratedProtocolMessageType('PrivateKey', (_message.Message,), dict(
-  DESCRIPTOR = _PRIVATEKEY,
-  __module__ = 'crypto_pb2'
-  # @@protoc_insertion_point(class_scope:crypto.pb.PrivateKey)
-  ))
+PrivateKey = _reflection.GeneratedProtocolMessageType(
```
pipermerriam: I think these files should be added to the --ignore configuration for black since they are generated by protobuf.

p2p/auth.py
```diff
@@ -88,21 +89,19 @@
...
-        timeout=REPLY_TIMEOUT)
+        reader.read(ENCRYPTED_AUTH_ACK_LEN), timeout=REPLY_TIMEOUT
+    )
 
     if reader.at_eof():
         # This is what happens when Parity nodes have blacklisted us
         # (https://github.com/ethereum/py-evm/issues/901).
-        raise HandshakeFailure("%s disconnected before sending auth ack", repr(initiator.remote))
+        raise HandshakeFailure(
+            "%s disconnected before sending auth ack", repr(initiator.remote)
```
pipermerriam: f-string

p2p/discovery.py
```diff
@@ -344,27 +374,34 @@ def _send_find_node(self, node: kademlia.Node, target_node_id: int) -> None:
...
             self.logger.debug2("node lookup; querying %s", nodes_to_ask)
             nodes_asked.update(nodes_to_ask)
-            results = await asyncio.gather(*(
-                _find_node(node_id, n)
-                for n
-                in nodes_to_ask
-                if not self.neighbours_callbacks.locked(n)
-            ))
+            results = await asyncio.gather(
+                *(
```
pipermerriam: This is a little opaque.  Maybe we can create the iterator and then put it into `asyncio.gather` for readability?

davesque:
I wish I could feel 100% on black.  It just feels so painfully less than perfect.  I should take another look at it and try and brainwash myself into being okay with it.

p2p/discovery.py
```diff
@@ -410,27 +447,35 @@ def connection_made(self, transport: asyncio.BaseTransport) -> None:
...
 
         try:
-            bonded = await asyncio.gather(*(
-                self.bond(n)
-                for n
-                in self.bootstrap_nodes
-                if (not self.ping_callbacks.locked(n) and not self.pong_callbacks.locked(n))
-            ))
+            bonded = await asyncio.gather(
+                *(
```
pipermerriam: Same here with converting two two parts, make generator then feed to `asyncio.gather`

p2p/discovery.py
```diff
@@ -455,52 +500,62 @@ def receive(self, address: kademlia.Address, message: bytes) -> None:
...
 
     def recv_ping_v4(self, node: kademlia.Node, _: Any, message_hash: Hash32) -> None:
-        self.logger.debug2('<<< ping(v4) from %s', node)
+        self.logger.debug2("<<< ping(v4) from %s", node)
         self.process_ping(node, message_hash)
         self.send_pong_v4(node, message_hash)
 
-    def recv_find_node_v4(self, node: kademlia.Node, payload: Tuple[Any, ...], _: Hash32) -> None:
+    def recv_find_node_v4(
+        self, node: kademlia.Node, payload: Tuple[Any, ...], _: Hash32
```
pipermerriam: changing the import strategy to do `from kademlia import Node` might allow many of these function definitions to fit on a single line.

p2p/discovery.py
```diff
@@ -1188,14 +1340,15 @@ def __init__(self, node: kademlia.Node, pong: bytes, topics: List[bytes],
         self.registration_times = [now + wait_period for wait_period in wait_periods]
 
     def __repr__(self) -> str:
-        return 'Ticket(%s:%s)' % (self.node, self.topics)
+        return "Ticket(%s:%s)" % (self.node, self.topics)
```
pipermerriam: f-string?

p2p/kademlia.py
```diff
@@ -112,20 +97,24 @@ def from_uri(cls, uri: str) -> 'Node':
     def uri(self) -> str:
         hexstring = self.pubkey.to_hex()
         hexstring = remove_0x_prefix(hexstring)
-        return f'enode://{hexstring}@{self.address.ip}:{self.address.tcp_port}'
+        return f"enode://{hexstring}@{self.address.ip}:{self.address.tcp_port}"
 
     def __str__(self) -> str:
-        return '<Node(%s@%s)>' % (self.pubkey.to_hex()[:6], self.address.ip)
+        return "<Node(%s@%s)>" % (self.pubkey.to_hex()[:6], self.address.ip)
```
pipermerriam: f-string

p2p/kademlia.py
```diff
@@ -302,7 +292,9 @@ def add_node(self, node: Node) -> Node:
             # Split if the bucket has the local node in its range or if the depth is not congruent
             # to 0 mod k_b
             depth = _compute_shared_prefix_bits(bucket.nodes)
-            if bucket.in_range(self.this_node) or (depth % k_b != 0 and depth != k_id_size):
+            if bucket.in_range(self.this_node) or (
+                depth % k_b != 0 and depth != k_id_size
```
pipermerriam: Maybe pull statement out to `should_split = ...` so `if` statement fits on one line?

cburgdorf:
I fully support using black for Trinity 👍

p2p/peer_pool.py
```diff
@@ -290,11 +257,13 @@ def _add_peer(self,
     async def stop_all_peers(self) -> None:
         self.logger.info("Stopping all peers ...")
         peers = self.connected_nodes.values()
-        await asyncio.gather(*(
-            peer.disconnect(DisconnectReason.client_quitting)
-            for peer in peers
-            if peer.is_running
-        ))
+        await asyncio.gather(
+            *(
```
pipermerriam: same suggestion to convert to two steps

p2p/service.py
```diff
@@ -259,10 +268,15 @@ def call_later(self, delay: float, callback: 'Callable[..., None]', *args: Any)
...
         if self._child_services:
-            self.logger.debug("Waiting for child services: %s", list(self._child_services))
-            await asyncio.gather(*[
-                child_service.events.cleaned_up.wait()
-                for child_service in self._child_services])
+            self.logger.debug(
+                "Waiting for child services: %s", list(self._child_services)
+            )
+            await asyncio.gather(
+                *[
```
pipermerriam: same, convert to two steps, and maybe use `(...)` here to make it lazy instead of evaluating to a list.

trinity/_utils/filesystem.py
```diff
@@ -22,26 +22,29 @@ class PidFile:
...
-                "File %s already exists -- cannot run the same process twice" % self.filepath
+                "File %s already exists -- cannot run the same process twice"
+                % self.filepath
             )
         elif self._running:
             raise RuntimeError("Cannot enter the same PidFile context twice")
         if not self.filepath.parent.is_dir():
             raise IOError(
-                "Cannot create pidfile in base directory that doesn't exist: %s" % self.filepath
+                "Cannot create pidfile in base directory that doesn't exist: %s"
```
pipermerriam: f-string

trinity/_utils/filesystem.py
```diff
@@ -22,26 +22,29 @@ class PidFile:
...
     def __init__(self, process_name: str, path: Path) -> None:
-        self.filepath = path / (process_name + '.pid')
+        self.filepath = path / (process_name + ".pid")
         self._running = False
 
     def __enter__(self) -> None:
         if self.filepath.exists():
             raise FileExistsError(
-                "File %s already exists -- cannot run the same process twice" % self.filepath
+                "File %s already exists -- cannot run the same process twice"
```
pipermerriam: f-string

trinity/_utils/mp.py
```diff
@@ -85,9 +78,10 @@ def __str__(self) -> str:
 
 
 class ChainedExceptionWithTraceback(Exception):
-
     def __init__(self, exc: Exception, tb: TracebackType) -> None:
-        self.tb = '\n"""\n%s"""' % ''.join(traceback.format_exception(type(exc), exc, tb))
+        self.tb = '\n"""\n%s"""' % "".join(
```
pipermerriam: f-string

trinity/bootstrap.py
```diff
@@ -2,76 +2,49 @@
...
-    r"      ______     _       _ __       ",
-    r"     /_  __/____(_)___  (_) /___  __",
-    r"      / / / ___/ / __ \/ / __/ / / /",
-    r"     / / / /  / / / / / / /_/ /_/ / ",
-    r"    /_/ /_/  /_/_/ /_/_/\__/\__, /  ",
-    r"                           /____/   ",
-))
+TRINITY_HEADER = "\n".join(
+    (
+        "\n" r"      ______     _       _ __       ",
```
pipermerriam: maybe use a `# fmt: off/on` for this block to keep it lined up or add a comment after te `"\n"` to make that line long enough so it doesn't combine these.

trinity/plugins/builtin/ethstats/ethstats_service.py
```diff
@@ -64,38 +47,39 @@ def __init__(
...
 
                     await self.wait_first(
-                        self.server_handler(client),
-                        self.statistics_handler(client),
+                        self.server_handler(client), self.statistics_handler(client)
                     )
             except websockets.ConnectionClosed as e:
-                self.logger.info('Connection to %s is closed: %s' % (self.server_url, e))
+                self.logger.info(
+                    "Connection to %s is closed: %s" % (self.server_url, e)
```
pipermerriam: Should not be formatting string before passing into logging, adjust args to be positional.

trinity/plugins/builtin/fix_unclean_shutdown/plugin.py
```diff
@@ -1,60 +1,60 @@
...
-        self.logger.info("Searching for process id files in %s..." % trinity_config.data_dir)
-        pidfiles = tuple(trinity_config.pid_dir.glob('*.pid'))
+        self.logger.info(
+            "Searching for process id files in %s..." % trinity_config.data_dir
+        )
+        pidfiles = tuple(trinity_config.pid_dir.glob("*.pid"))
         if len(pidfiles) > 1:
-            self.logger.info('Found %d processes from a previous run. Closing...' % len(pidfiles))
+            self.logger.info(
+                "Found %d processes from a previous run. Closing..." % len(pidfiles)
```
pipermerriam: Most of the logging statements in this file are doing it wrong....  not commenting on all of them.  Should not be pre-formatting the string.

trinity/plugins/builtin/tx_pool/validators.py
```diff
@@ -41,8 +32,7 @@ def __init__(self,
         self.vm_configuration = self.chain.vm_configuration
 
         self._ordered_tx_classes = [
```
pipermerriam: Noting the use of a mutable `list` here.

trinity/protocol/common/trackers.py
```diff
@@ -99,8 +92,8 @@ def get_stats(self) -> str:
         # missing: total number of missing response items
         # quality: 0-100 for how complete responses are
         return (
-            'msgs=%d  items=%d  rtt=%.2f/%.2f/%.2f  ips=%.5f  '
-            'timeouts=%d  quality=%d'
+            "msgs=%d  items=%d  rtt=%.2f/%.2f/%.2f  ips=%.5f  "
+            "timeouts=%d  quality=%d"
```
pipermerriam: f-string

trinity/protocol/eth/peer.py
```diff
@@ -43,7 +28,7 @@ class ETHPeer(BaseChainPeer):
 
     def get_extra_stats(self) -> List[str]:
         stats_pairs = self.requests.get_stats().items()
-        return ['%s: %s' % (cmd_name, stats) for cmd_name, stats in stats_pairs]
+        return ["%s: %s" % (cmd_name, stats) for cmd_name, stats in stats_pairs]
```
pipermerriam: f-string

trinity/protocol/les/peer.py
```diff
@@ -55,7 +30,7 @@ class LESPeer(BaseChainPeer):
 
     def get_extra_stats(self) -> List[str]:
         stats_pairs = self.requests.get_stats().items()
-        return ['%s: %s' % (cmd_name, stats) for cmd_name, stats in stats_pairs]
+        return ["%s: %s" % (cmd_name, stats) for cmd_name, stats in stats_pairs]
```
pipermerriam: f-string

trinity/rpc/ipc.py
```diff
@@ -48,28 +39,30 @@
...
-                writer,
-                "reached limit: %d bytes, starting with '%s'" % (
-                    e.consumed,
-                    request_bytes[:20],
-                ),
-            ))
+            await cancel_token.cancellable_wait(
+                write_error(
+                    writer,
+                    "reached limit: %d bytes, starting with '%s'"
```
pipermerriam: f-string

trinity/server.py
```diff
@@ -268,12 +250,9 @@ def _make_request_server(self) -> BaseRequestServer:
...
 
         total_peers = len(self.peer_pool)
-        inbound_peer_count = len([
-            peer
-            for peer
-            in self.peer_pool.connected_nodes.values()
-            if peer.inbound
-        ])
+        inbound_peer_count = len(
+            [peer for peer in self.peer_pool.connected_nodes.values() if peer.inbound]
```
pipermerriam: *nit* me-no-likey-`list`

trinity/sync/light/service.py
```diff
@@ -196,26 +170,25 @@ def callback(r: protocol._DecodedMsgType) -> None:
...
         try:
             rlp_account = HexaryTrie.get_from_proof(header.state_root, key, proof)
         except BadTrieProof as exc:
-            raise BadLESResponse("Peer %s returned an invalid proof for account %s at block %s" % (
-                peer,
-                encode_hex(address),
-                encode_hex(block_hash),
-            )) from exc
+            raise BadLESResponse(
+                "Peer %s returned an invalid proof for account %s at block %s"
```
pipermerriam: f-string

trinity/sync/light/service.py
```diff
@@ -254,33 +226,35 @@ def callback(r: protocol._DecodedMsgType) -> None:
...
         else:
             # a bad-acting peer sent an invalid non-empty bytecode
-            raise BadLESResponse("Peer %s sent code %s that did not match hash %s in account %s" % (
-                peer,
-                encode_hex(bytecode),
-                encode_hex(code_hash),
-                encode_hex(address),
-            ))
+            raise BadLESResponse(
+                "Peer %s sent code %s that did not match hash %s in account %s"
```
pipermerriam: f-string

trinity/sync/light/service.py
```diff
@@ -296,45 +270,48 @@ def callback(r: protocol._DecodedMsgType) -> None:
         except HeaderNotFound:
             # We presume that the current peer is the best peer. Because
             # our best peer doesn't have the header we want, there are no eligible peers.
-            raise NoEligiblePeers("Our best peer does not have the header %s" % block_hash)
+            raise NoEligiblePeers(
+                "Our best peer does not have the header %s" % block_hash
```
pipermerriam: f-string

trinity/sync/light/service.py
```diff
@@ -296,45 +270,48 @@ def callback(r: protocol._DecodedMsgType) -> None:
...
             raise NoEligiblePeers(
                 "Peer %s claims to be ahead of %s, but returned empty code with hash %s. "
-                "It is on number %d, maybe an uncle. Retry with an older block hash." % (
-                    peer,
-                    header,
-                    code_hash,
-                    head_number,
-                )
+                "It is on number %d, maybe an uncle. Retry with an older block hash."
+                % (peer, header, code_hash, head_number)
```
pipermerriam: not commenting on rest, lots of old `%` based formatting in this file.

carver:
> I wish I could feel 100% on black. It just feels so painfully less than perfect. I should take another look at it and try and brainwash myself into being okay with it.

Yeah, this is kinda where I'm at. I think I'd be *more* on board if it was ugly **and** stable. It's still in beta, I expect the format to change at every release, which means we have to pin it exactly, and then deal with upgrading to the latest black from time to time.

I'm super familiar and comfortable with the current formatting, and more or less output the current style into my editor now, so I'm not feeling the pain as much. So I guess it's not "over my dead body" but I'm not happy with it.

davesque:
Screen Shot 2019-05-02 at 5 13 44 PM
😱

pipermerriam:
I somewhat view it as the lesser of two evils but I think I'm going to put myself in the +0 camp meaning I'm for it but not willing enough at this point to hold my ground against anyone who's vocally opposed.

ralexstokes:
@pipermerriam wow thanks for reviewing all that stuff!

i'll go thru and address most of them... unclear what motivation i will have for all the `f-string` stuff...

i'm personally of the mind that the most important thing here is to remove the concern of formatting from the programmer... i would rather spend that energy/time doing other things. i can see where the output of black may not be what we like or are used to. i think this type of tooling works very well in e.g. go or rust with the caveat that the languages got to (at least somewhat) co-evolve w/ the lang formatter. from skimming the output of the diff, it seems like a lot of the more sub-optimal parts can be fixed by changing _how_ that particular piece of code is written... this implies a bit of a learning curve optimizing for what you like, what black is good at formatting and what is valid python... (aside/random thought: perhaps this task is easier in a static language bc the space of valid programs is somewhat smaller??)

this being said, i don't think we should switch to a tool if the majority of us aren't going to like it :) i would be open to looking at other formatters -- `yapf`? or we can just drop the issue and spend our time on the formatting. i agree that after a while, the expected style becomes learned and it is less of an issue.

njgheorghita:
> i'm personally of the mind that the most important thing here is to remove the concern of formatting from the programmer...

Yup, completely agree on this. I use black in all the ethpm libraries, and I'm a big fan. It saves me a significant amount of time & mental energy. I found that any particular nitpicks I have about a certain formatting rule that black imposes disappeared pretty quickly after I started using the tool. But, it might not be the right tool to be the standard for all of our formatting guidelines

cburgdorf:
>i'm personally of the mind that the most important thing here is to remove the concern of formatting from the programmer...

I'm on the same page. :+1:

carver:
> this being said, i don't think we should switch to a tool if the majority of us aren't going to like it :) i would be open to looking at other formatters -- `yapf`? or we can just drop the issue and spend our time on the formatting.

I haven't formed any opinions on alternatives. AIUI, `black` is progressing at a quick clip. I suspect that when it settles down a bit (and comes out of beta), that I would go from a `-0` (to riff on Piper's position), to at least a `+0`.

carver:
So punting for a bit seems like a very reasonable plan.

pipermerriam:
@carver could we possibly take the approach of putting it in place and talking it over in June when we're all together and see if it's working or not?  Fear of unknown vs try it and see what it's like.

carver:
> @carver could we possibly take the approach of putting it in place and talking it over in June when we're all together and see if it's working or not? Fear of unknown vs try it and see what it's like.

Hm, what does the path look like if we decide we don't want `black` after using it for a while? I'm not sure how we recover. Reverting the merge commit would be a mess after a bunch of new code was added. There isn't a tool to apply flake8 rules that I know about. So it looks like a one-way lock-in.

pipermerriam:
> Hm, what does the path look like if we decide we don't want `black` after using it for a while?

I think we just turn it off in CI and new code still gets all of our existing lint rules enforced but goes in the way it does today, however the programmer wrote it.  It leaves us in a different state than we started off in, but it'll still be well formatted code.

carver:
> If the major blocker is it not being stable

Yeah, I think this is the biggest issue for me. I don't love blowing up the git history with the re-format, but I can live with one. I'd really rather avoid blowing up the git history at every point release.

pipermerriam:
So on hold until they hit a non-beta.  I'm good with that.  Trinity is big.

I'd still like to do a slow roll-out across our smaller stuff but it's not a priority.

eth2/beacon/tools/misc/ssz_vector.py
ralexstokes: @pipermerriam we need to set a class-wide parameter at runtime (after we know what the Eth2 configuration is)... do you have any thoughts on how this is best done in Python?

pipermerriam: We *sort-of* have this in `Chain.vm_configuration` which is a class level configuration that gets set at runtime.  We do this with dynamic sub-classing which isn't an ideal solution but it does work.

ralexstokes:
given that we want to block adoption of `black` until non-beta release status, i'll go ahead and close this for now.

PR #584, opened 2019-05-04
Fix CI failure: Add config to `FakeAsyncBeaconChainDB.__init__`

mhchia:
### What was wrong?
`config` is added in `BaseBeaconChainDB`. Sorry I made the CI [fail](https://circleci.com/gh/ethereum/trinity/33433) because I didn't rebase before merging my PR.

### How was it fixed?
Add config in `FakeAsyncBeaconChainDB`, and pass the corresponding configs in the tests.

#### Cute Animal Picture

### DIFF
--- a/tests/core/p2p-proto/bcc/helpers.py
+++ b/tests/core/p2p-proto/bcc/helpers.py
@@ -14,13 +14,14 @@ from eth.constants import (
 )
 from eth.db.atomic import AtomicDB
 
+from eth2.configs import Eth2Config
 from eth2.beacon.db.chain import BeaconChainDB
-from trinity.db.beacon.chain import BaseAsyncBeaconChainDB
 from eth2.beacon.types.blocks import (
     BeaconBlock,
     BeaconBlockBody,
 )
 
+from trinity.db.beacon.chain import BaseAsyncBeaconChainDB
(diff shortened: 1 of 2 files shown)


tests/core/p2p-proto/bcc/helpers.py
```diff
@@ -101,7 +103,7 @@ def create_branch(length, root=None, **start_kwargs):
 
 async def get_chain_db(blocks=()):
     db = AtomicDB()
-    chain_db = FakeAsyncBeaconChainDB(db)
+    chain_db = FakeAsyncBeaconChainDB(db=db, config=SERENITY_CONFIG)
```
hwwhww: Should `config` be configurable (as a parameter of `get_chain_db`) so that we can use this function with `XIAO_LONG_BAO_CONFIG`?

mhchia: That sounds good to me! How about I change it to the following, so that we can pass the config we want?
```python
async def get_chain_db(blocks=(), config=SERENITY_CONFIG):
    db = AtomicDB()
    chain_db = FakeAsyncBeaconChainDB(db)
    chain_db = FakeAsyncBeaconChainDB(db=db, config=config)
```

mhchia:
Thank you for the review:)

PR #585, opened 2019-05-05
Add second tick to SlotTicker

NIC619:
### How was it fixed?

- [x] Relocate `test_receive_server.py`
- [x] Add `tests/plugin/eth2/beacon/helpers.py`
- [x] Add `_finalized_root` and `_highest_justified_epoch` to `FakeAsyncBeaconChainDB`
- [x] Add second tick to `SlotTicker` which ticks when half the time(i.e., `seconds_per_slot/2` seconds) passed
    - [x] not-proposing validator skips the slot only when a second tick is received

#### Cute Animal Picture


mhchia:
> fix missing config error after #580 is merged

I made a PR https://github.com/ethereum/trinity/pull/584, which should address this issue.

> would it be better to move this test to tests/plugins/eth2/beacon/

I intended to put it beside the test for `BCCRequestServer`. But it is dedicated used in the beacon node plugin, I agree that it might be better to move it to `tests/plugins/eth2/beacon/`.

NIC619:
I realized it doesn't make much sense to check if receive server will reject a block with wrong signature or data. After all there are already tests for rejecting bad blocks.
So I repurpose this PR to adding a second tick to `SlotTicker` as described in the PR comment.

tests/plugins/eth2/beacon/helpers.py
```diff
@@ -0,0 +1,61 @@
...
+    for i in range(NUM_VALIDATORS)
+)
+index_to_pubkey = {}
+keymap = {}  # pub -> priv
+for i, k in enumerate(privkeys):
+    pubkey = bls.privtopub(k)
+    index_to_pubkey[i] = pubkey
+    keymap[pubkey] = k
+
+genesis_time = int(time.time())
```
ChihChengLiang: nip: maybe we don't need this variable. It's ok to use 0 in `create_mock_genesis` and `BeaconGenesisData`

tests/plugins/eth2/beacon/helpers.py
```diff
@@ -0,0 +1,61 @@
...
+genesis_data = BeaconGenesisData(
+    genesis_time=genesis_time,
+    genesis_slot=XIAO_LONG_BAO_CONFIG.GENESIS_SLOT,
+    keymap=keymap,
+    num_validators=NUM_VALIDATORS,
+)
+beacon_chain_config = BeaconChainConfig(chain_name='TestTestTest', genesis_data=genesis_data)
+chain_class = beacon_chain_config.beacon_chain_class
+
+override_vector_lengths(XIAO_LONG_BAO_CONFIG)
```
ChihChengLiang: Not sure how would the side effect of `override_vector_lengths` work here. I recommend moving this fixture to a higher level https://github.com/NIC619/trinity/blob/aecb15058a7d678e62121887b5971ec5d00e4b7e/tests/eth2/beacon/conftest.py#L62
cc @hwwhww

trinity/plugins/eth2/beacon/slot_ticker.py
```diff
@@ -23,13 +23,14 @@
     TrinityEventBusEndpoint,
 )
 
-DEFAULT_CHECK_FREQUENCY = 5
+DEFAULT_CHECK_FREQUENCY = 6
 
 
 class NewSlotEvent(BaseEvent):
```
ChihChengLiang: Proposer rename `NewSlotEvent` to `SlotTickerEvent`, since it now handles the `on the new slot` and `on the middle of the slot` cases.

trinity/plugins/eth2/beacon/slot_ticker.py
```diff
@@ -23,13 +23,14 @@
...
-    def __init__(self, slot: Slot, elapsed_time: Second):
+    def __init__(self, slot: Slot, elapsed_time: Second, is_second_signal: bool):
```
ChihChengLiang: Proposer rename `is_second_signal` to `is_half_slot`.

trinity/plugins/eth2/beacon/validator.py
```diff
@@ -100,7 +100,7 @@ def __init__(
                 state_machine=state_machine,
                 head_block=head,
             )
-        else:
+        elif is_second_signal:
```
ChihChengLiang: Add a comment:
```
# skip the block at half of the slot
```

trinity/plugins/eth2/beacon/validator.py
```diff
@@ -79,9 +79,9 @@ def __init__(
         The callback for `SlotTicker`, to be called whenever new slot is ticked.
         """
         async for event in self.event_bus.stream(NewSlotEvent):
-            await self.new_slot(event.slot)
+            await self.new_slot(event.slot, event.is_second_signal)
```
ChihChengLiang: Maybe do 
```
if event.is_second_signal:
    self.handle_new_slot(event.slot)
else:
    self.handle_half_slot(event.slot)
```
rename new_slot to handle_new_slot
create handle_half_slot

trinity/plugins/eth2/beacon/slot_ticker.py
NIC619: I rename it to `is_second_half_slot` to indicate we are in the second half of the slot. Let me know if it doesn't make sense.

ChihChengLiang: Think about this more, `is_second_signal` or `is_second_tick` sounds better. Since the second tick is essentially a concept of timeout and does not happen at the half of the slot.

trinity/plugins/eth2/beacon/validator.py
NIC619: Handling `new_slot` and `second_half_slot` are not entirely mutually exclusive and they both need to compute `proposer_index` to determine if they should `propose`/`skip` block. So I refactor a bit but still keep them in the same function. Let me know if that makes sense.

tests/core/p2p-proto/bcc/helpers.py
```diff
@@ -51,6 +51,9 @@ def __init__(self, db: BaseAtomicDB, config: Eth2Config) -> None:
         self.db = db
         self.config = config
 
+        self._finalized_root = self._get_finalized_root_if_present(db)
+        self._highest_justified_epoch = self._get_highest_justified_epoch(db)
```
mhchia: It seems we can just change to
```python
def __init__(self, db: BaseAtomicDB, config: Eth2Config) -> None:
    super().__init__(self, db, config)
```

trinity/plugins/eth2/beacon/validator.py
```diff
@@ -60,28 +66,33 @@ def __init__(
...
+            genesis_epoch: Epoch,
+            slots_per_epoch: int,
             token: CancelToken = None) -> None:
         super().__init__(token)
         self.validator_index = validator_index
         self.chain = chain
         self.peer_pool = peer_pool
         self.privkey = privkey
         self.event_bus = event_bus
+        # TODO: `latest_proposed_epoch` should be written into/read from validator's own db
```
mhchia: Does this TODO mean we need a dedicated db for a `Validator`?

NIC619: Yes!

PR #586, opened 2019-05-06
ENR

### LINKED ISSUE #583: Implement ENRs
### What is wrong?

First step of implementing discv5 is implementing the ENR format. The spec can be found [here](http://eips.ethereum.org/EIPS/eip-778).

### How can it be fixed

Add a class representing ENRs with methods to

- create them based on a set of key-value pairs
- sign them
- validate their signature
- get values by key
- compare sequence numbers of two records

Add an RLP sedes object to serialize/deserialize them. We can probably not just use `rlp.Serializable` as the number of fields isn't fixed.

jannikluhn:
Closes #583. Still WIP: A couple of additional tests are needed and we have to use compressed public keys (working on https://github.com/ethereum/eth-keys/issues/55 now).

### How was it fixed?

- There's an ENR class that exposes the ENR key value pairs via a dict interface, along the sequence number and signature. If the signature is `None` it is considered unsigned. A signed copy can be created with `to_signed_enr`.
- There are special sedes classes `ENRSedes` and `ENRContentSedes` to serialize/deserialize an ENR with or without the signature. Values are (de)serialized as binaries, unless there's something better in `ENR_KEY_SEDES_MAPPING`. 
- Identity schemes (responsible for signing and validating signatures) are represented by children of `ENRIdentityScheme` which define a method to check if they can be used for a given ENR or not (usually by checking if the ENR contains a public key in the right format). The ENR will use the first matching identity scheme it can find (there will probably only be a few).

#### Cute Animal Picture


jannikluhn:
Updated to use the new eth-keys API introduced in https://github.com/ethereum/eth-keys/pull/56 and https://github.com/ethereum/eth-keys/pull/58. I still have to add the new version in `setup.py`, but I can only do this once the new changes have been released.

I also restructured the code a little in order to have two separate classes for signed and unsigned ENRs (instead of an `is_signed` property and a `signature` that may be `None`).

p2p/discv5/identity_schemes.py
```diff
@@ -0,0 +1,91 @@
...
+class IdentityScheme(ABC):
+
+    @classmethod
+    @abstractmethod
+    def is_applicable(cls, enr: "BaseENR") -> bool:
+        pass
+
+    @classmethod
+    @abstractmethod
+    def create_signature(cls, enr: "BaseENR", private_key: Any) -> bytes:
```
hwwhww:
```suggestion
    def create_signature(cls, enr: "BaseENR", private_key: PrivateKey) -> bytes:
```

p2p/discv5/enr.py
```diff
@@ -0,0 +1,236 @@
...
+        return self._kv_pairs.get(key, default)
+
+    @abstractmethod
+    def __eq__(self, other: Any) -> bool:
+        pass
+
+
+class UnsignedENR(BaseENR, ENRContentSedes):
+
+    def to_signed_enr(self, private_key: Any) -> "ENR":
```
hwwhww:
```suggestion
    def to_signed_enr(self, private_key: PrivateKey) -> "ENR":
```

p2p/discv5/identity_schemes.py
```diff
@@ -0,0 +1,91 @@
...
+
+from typing import (
+    Any,
+    TYPE_CHECKING,
+)
+
+from eth_keys.datatypes import (
+    PrivateKey,
+    PublicKey,
+    NonRecoverableSignature,
```
hwwhww: ImportError: cannot import name 'NonRecoverableSignature' from 'eth_keys.datatypes' (/home/circleci/repo/.tox/py37-p2p/lib/python3.7/site-packages/eth_keys/datatypes.py)

p2p/discv5/enr.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+class ENRContentSedes:
+
+    @classmethod
+    def serialize(cls, enr: "BaseENR") -> Tuple[bytes, ...]:
+        serialized_sequence_number = big_endian_int.serialize(enr.sequence_number)
+        serialized_keys = tuple(binary.serialize(key) for key in enr.keys())
+        value_serializers = tuple(
+            ENR_KEY_SEDES_MAPPING.get(key, binary)
```
pipermerriam: This probably deserves a comment.  Not intuitive to me (though I'm not familiar to the protocol) why we'd fall back to the `binary` sedes.  Maybe something explaining the basic structure of the record and why we do this fallback?

p2p/discv5/enr.py
```diff
@@ -0,0 +1,236 @@
...
+            )
+
+
+class ENRSedes:
+
+    @classmethod
+    def serialize(cls, enr: "ENR") -> Tuple[bytes, ...]:
+        serialized_signature = binary.serialize(enr.signature)
+        serialized_content = ENRContentSedes.serialize(enr)
+        return (serialized_signature,) + serialized_content
```
pipermerriam: Not necessarily better but I found [`toolz.cons`](https://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.cons) recently which does exactly this.  I believe it is lazy meaning that it has to be combined with an outer `tuple` cast in this scenario which isn't exactly ideal.

p2p/discv5/enr.py
```diff
@@ -0,0 +1,236 @@
...
+        applicable_identity_schemes = tuple(
+            identity_scheme
+            for identity_scheme in identity_schemes
+            if identity_scheme.is_applicable(self)
+        )
+
+        if not applicable_identity_schemes:
+            raise ValidationError("No known identity scheme used")
+
+        return applicable_identity_schemes[0]
```
pipermerriam: Is it intentional that this just blindly takes the first one?  If so I think that deserves a comment indicating that we're aware that this discards the extra ones and that is an intentional design decision.

Second, this function could be re-structured as:

```python
for scheme in identity_schemes:
    if scheme.is_applicable(self):
        return scheme
else:
    raise ValidationError(...)
```

This should be slightly more performant as well.

p2p/discv5/enr.py
```diff
@@ -0,0 +1,236 @@
...
+        identity_scheme.validate_signature(self)
+
+    def extract_node_address(self) -> bytes:
+        identity_scheme = self.get_identity_scheme()
+        return identity_scheme.extract_node_address(self)
+
+    def __eq__(self, other: Any) -> bool:
+        return (
+            other.__class is self.__class__ and
+            dict(other) == dict(self) and
```
pipermerriam: Since it appears that ENR records are immutable mappings, what do you think about implementing `__hash__` on the class and implementing this as `type(other) is type(self) and hash(other) == hash(self)` and capturing all of the value equality in the `__hash__` implementation.

p2p/discv5/enr.py
```diff
@@ -0,0 +1,236 @@
...
+                f"ENRs must not be larger than {MAX_ENR_SIZE} bytes"
+            )
+
+
+class BaseENR(Mapping[bytes, Any], ABC):
+    def __init__(self,
+                 sequence_number: int,
+                 kv_pairs: Mapping[bytes, Any]) -> None:
+        self._sequence_number = sequence_number
+        self._kv_pairs = kv_pairs
```
pipermerriam: Since it appears the dict API implemented on the ENR is intended to make it immutable, maybe this value should be isolated by casting it to `dict` to ensure any ties with the outside object dont't result in mutation of the ENR values.

p2p/discv5/identity_schemes.py
```diff
@@ -0,0 +1,91 @@
...
+        signature = NonRecoverableSignature(enr.signature)
+        message = enr.get_signing_message()
+
+        try:
+            is_valid = signature.verify_msg(message, public_key)
+        except BadSignature:
+            is_valid = False
+
+        if not is_valid:
+            raise ValidationError("Invalid signature")
```
pipermerriam: Seems like you could just raise this from within the `except` block above to avoid the need to store an `is_valid` variable.

tests/p2p/discv5/test_enr.py
```diff
@@ -0,0 +1,138 @@
...
+    ValidationError,
+)
+
+from p2p.discv5.enr import (
+    ENR,
+    UnsignedENR,
+)
+
+
+OFFICIAL_ENR_RLP = decode_hex(
```
pipermerriam: Can you add a comment linking to the source of this test vector?

p2p/discv5/enr.py
carver: Hash is not guaranteed to be unique so that equality test gets you some false positives (and prevents subclasses from matching, which means false negatives, I suppose). Implementing hash is still worthwhile performance-wise, since most equality checks that matter will check the hash first anyways. Though I don't think it is strictly necessary for this pr.

p2p/discv5/identity_schemes.py
jannikluhn: I'd actually like to make it use `bytes` instead. The reason is that different identity schemes will use different private key formats and `bytes` would be the lowest most common denominator.

jannikluhn:
Opened a meta issue here (I guess I should have started with that): https://github.com/ethereum/trinity/issues/668

p2p/discv5/enr.py
jannikluhn: It doesn't really matter as we don't use the value in any way, we could even use `big_endian_int`. `binary` seemed to make the most sense as it doesn't try to interpret, it's basically an identity function. I put it in a constant and added a comment.

p2p/discv5/enr.py
jannikluhn: Yes, the idea is that `identity_schemes` is ordered by preference, so we just pick the one that we like the most. I don't expect there to be many, probably not more than two.

I changed the name to `get_preferred_identity_scheme` and added a comment to the definition of `identity_schemes` about the order. I hope that makes it more clear.

p2p/discv5/identity_schemes.py
jannikluhn: `verify_signature` can either return `False` or raise a `BadSignature`, depending on the way the signature is invalid. So there isn't away to avoid the variable without copying the `raise ValidationError` line which I'd like to avoid.

p2p/discv5/enr.py
```diff
@@ -0,0 +1,236 @@
...
+            other.__class is self.__class__ and
+            dict(other) == dict(self) and
+            other.signature == self.signature
+        )
+
+
+ENR_KEY_SEDES_MAPPING = {
+    b"id": binary,
+    b"secp256k1": Binary(min_length=33, max_length=33),
+    b"ip": binary,
```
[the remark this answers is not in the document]
jannikluhn: Did it the proper way and added a special sedes object for IP addresses.

p2p/discv5/enr.py
jannikluhn: Added a hash function, but stayed away from using it in `__eq__` as false positives sound dangerous.

jannikluhn:
I've addressed the feedback and now I'm waiting for ethereum/eth-keys#58 and a subsequent eth-keys release. Then I'll have a look at the outstanding CI errors.

pipermerriam:
That suggests the verify_signature api should be changed if possible to either convert the exception case to return false. I'm on mobile so haven't really done my homework on feasibility or why the api was designed that way.

Sent from ProtonMail mobile

p2p/discv5/identity_schemes.py
jannikluhn: > That suggests the verify_signature api should be changed if possible to either convert the exception case to return false. I'm on mobile so haven't really done my homework on feasibility or why the api was designed that way.

Looks like what I said is not accurate, at least not after the changes in https://github.com/ethereum/eth-keys/pull/58 (as we don't try to recover the signer anymore). So `verify_msg` should never fail as far as I can see. However, the signature can still be broken, but this would raise an exception when creating the signature. I updated the function accordingly to cover the two failure cases. It still needs the `is_valid` variable, but it looks less weird now.

p2p/discv5/enr.py
jannikluhn: Spec has just been updated to explicify the used/supported identity scheme: https://github.com/ethereum/EIPs/commit/9b1c22d2850cc8efb7aabb2aba4098873ed0d5a9

As the change doesn't appear to be final yet, I'm gonna wait a little before updating, but definitely do it before merging.

fjl:
Very nice to see Trinity will have ENR support :).
Note that EIP-778 was [changed one more time](https://github.com/ethereum/EIPs/commit/5a0665ca9cece6172087fe9eadcfd1c6d22ba9c9) to add "tcp6", "udp6" keys. They allow
relaying IPv6-specific port numbers. 

We have also decided to use base64url encoding (without padding) instead of standard base64.
Base64 without padding is not supported in the Python standard library, but you can use these
functions to do it: https://github.com/fjl/p2p-drafts/commit/23ed3427007e19a3e6f81202e3e967ddd74ef9e7#diff-09ba11ec95dfcee8269304dd2cc0abf3R156

jannikluhn:
- Updated to latest version of ENR spec (thanks @fjl)
- reworked tests: clear separation between ENR and identity scheme
- cleaned up commit history

Unfortunately, now everything breaks due to a version conflict. Pretty much all `eth-*` libraries require `eth_keys < 3.0.0`, but here we need `eth_keys >= 3.0.0`. I guess there's no other way but going through all of them and make a new release? `3.0.0` *should* be backwards compatible.

carver:
> Unfortunately, now everything breaks due to a version conflict. Pretty much all `eth-*` libraries require `eth_keys < 3.0.0`, but here we need `eth_keys >= 3.0.0`. I guess there's no other way but going through all of them and make a new release? `3.0.0` _should_ be backwards compatible.

Yeah, sorry, this is fixed in #699 and was released so it doesn't affect fresh installs. But it hasn't been merged in yet. I'll ping you when it gets merged in, and the rebase should be painless.

carver:
@jannikluhn Okay, a rebase should fix the eth-keys issue now.

jannikluhn:
@carver that didn't do the trick. The problem is that this PR needs `eth-keys >= 3.0.0`, everything else `eth-keys < 3.0.0`. So I think there's no other way than updating all of them to use `< 4.0.0`.

carver:
Ah, right. @pipermerriam instead of propagating a chain of eth-keys releases up to trinity (I think at least these would need new releases: eth-account, web3, py-evm), maybe we can just do a patch release that includes everything that's in 0.3.0?

pipermerriam:
`v0.2.3` is now available and equal to `v0.3.0`

setup.py
```diff
@@ -14,6 +14,7 @@
         # cryptography does not use semver and allows breaking changes within `0.3` version bumps.
         "cryptography>=2.5,<2.7",
         "eth-hash>=0.1.4,<1",
+        "eth-keys>=0.3.0,<0.4.0",
```
carver: @jannikluhn Can you try again with

```suggestion
        "eth-keys>=0.2.3,<0.3.0",
```

carver:
Hm, interesting, I think this is a legit issue. The new `eth-keys` is treating a valid signature as invalid?

```
_ test_rpc_against_fixtures[/home/circleci/repo/fixtures/BlockchainTests/bcValidBlockTest/SimpleTx3.json:SimpleTx3_Frontier:Frontier] _
[gw1] linux -- Python 3.6.8 /home/circleci/repo/.tox/py36-rpc-blockchain/bin/python

event_bus = <trinity.endpoint.TrinityEventBusEndpoint object at 0x7f2f8c0a39b0>
chain_fixture = {'_info': {'comment': '', 'filledwith': 'testeth 1.5.0.dev2-68+commit.4a43125d.dirty', 'lllcversion': 'Version: 0.5.0-...80832fefd8808454c98c8142a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421880102030405060708c0c0', ...}
fixture_data = ('/home/circleci/repo/fixtures/BlockchainTests/bcValidBlockTest/SimpleTx3.json', 'SimpleTx3_Frontier', 'Frontier')

    @pytest.mark.asyncio
    async def test_rpc_against_fixtures(event_bus, chain_fixture, fixture_data):
        rpc = RPCServer(
            initialize_eth1_modules(MainnetFullChain(None), event_bus)
        )
    
        setup_result, setup_error = await call_rpc(rpc, 'evm_resetToGenesisFixture', [chain_fixture])
        # We need to advance the event loop for modules to be able to pickup the new chain
        await asyncio.sleep(0)
        assert setup_error is None and setup_result is True, "cannot load chain for {0}".format(fixture_data)  # noqa: E501
    
        await validate_accounts(rpc, chain_fixture['pre'])
    
        for block_fixture in chain_fixture['blocks']:
            should_be_good_block = 'blockHeader' in block_fixture
    
            if 'rlp_error' in block_fixture:
                assert not should_be_good_block
                continue
    
            block_result, block_error = await call_rpc(rpc, 'evm_applyBlockFixture', [block_fixture])
    
            if should_be_good_block:
>               assert block_error is None
E               AssertionError: assert 'Invalid Signature' is None

block_error = 'Invalid Signature'
block_fixture = {'blockHeader': {'bloom': '0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000...ce': '0x00', ...}, {'data': '0x', 'gasLimit': '0x5208', 'gasPrice': '0x01', 'nonce': '0x03', ...}], 'uncleHeaders': []}
block_result = None
chain_fixture = {'_info': {'comment': '', 'filledwith': 'testeth 1.5.0.dev2-68+commit.4a43125d.dirty', 'lllcversion': 'Version: 0.5.0-...80832fefd8808454c98c8142a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421880102030405060708c0c0', ...}
event_bus  = <trinity.endpoint.TrinityEventBusEndpoint object at 0x7f2f8c0a39b0>
fixture_data = ('/home/circleci/repo/fixtures/BlockchainTests/bcValidBlockTest/SimpleTx3.json', 'SimpleTx3_Frontier', 'Frontier')
rpc        = <trinity.rpc.main.RPCServer object at 0x7f2f68721ba8>
setup_error = None
setup_result = True
should_be_good_block = True
```

jannikluhn:
yeah, that's my fault, this PR should fix it: https://github.com/ethereum/eth-keys/pull/62

jannikluhn:
Finally everything is green :tada:

I think I've addressed all comments so far, but since the last review I've restructured and extended the tests a bit. Anyone wants to have another look? Otherwise this is ready from my side.

p2p/discv5/enr.py
```diff
@@ -0,0 +1,336 @@
...
+)
+
+
+class ENRContentSedes:
+
+    @classmethod
+    def serialize(cls, enr: "BaseENR") -> Tuple[bytes, ...]:
+        serialized_sequence_number = big_endian_int.serialize(enr.sequence_number)
+
+        sorted_key_value_pairs = sorted(enr.items(), key=lambda key_value: key_value[0])
```
pipermerriam: I believe historically we've measured `operator.itemgetter(0)` to be faster and slightly more preferred.

p2p/discv5/enr.py
```diff
@@ -0,0 +1,336 @@
...
+                serialized_enr,
+            )
+
+    @classmethod
+    def _validate_key_uniqueness(cls,
+                                 keys: Sequence[bytes],
+                                 serialized_enr: Sequence[bytes]) -> None:
+        duplicates = set(
+            key for index, key in enumerate(keys)
+            if key in keys[index + 1:]
```
pipermerriam: I think this has some unfortunate time complexity.  An `itertools.Counter` does a pretty nice job of both detecting duplicates and allowing easy reporting of which ones are duplicate.

```python
duplicates = {key for key, num in itertools.Counter(keys).items() if num > 1}
```

That one-liner does the trick pretty concisely.

p2p/discv5/enr.py
```diff
@@ -0,0 +1,336 @@
...
+
+        byte_size = sum(len(element) for element in serialized_enr)
+        if byte_size > MAX_ENR_SIZE:
+            raise DeserializationError(
+                f"ENRs must not be larger than {MAX_ENR_SIZE} bytes",
+                serialized_enr,
+            )
+
+
+class BaseENR(Mapping[bytes, Any], ABC):
```
pipermerriam: Are the value types truly `Any` or can that be constrained more?

p2p/discv5/enr.py
```diff
@@ -0,0 +1,336 @@
...
+    @property
+    def sequence_number(self) -> int:
+        return self._sequence_number
+
+    def get_signing_message(self) -> bytes:
+        return rlp.encode(self, ENRContentSedes)
+
+    def get_identity_scheme(self) -> Type[IdentityScheme]:
+        try:
+            identity_scheme_id = self[b"id"]
```
pipermerriam: Maybe instead of having to do this `try/except` checking at runtime it should happen during initialization and be memoized somewhere private since it's probably very natural to do that at whatever place the `b'id'` key is validated to be present.

p2p/discv5/identity_schemes.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+if TYPE_CHECKING:
+    from p2p.discv5.enr import (  # noqa: F401
+        BaseENR,
+        ENR,
+    )
+
+
+identity_scheme_registry: Dict[bytes, Type["IdentityScheme"]] = {}
```
pipermerriam: We probably still use this pattern in a number of places but I've grown to think that we should be using a registry pattern closer to something like this.

```
class Registry(collections.UserDict):
    def register_thing(self, thing) -> None:
        ...  # business logic here
        self[thing_key] = thing

default_registry = Registry()
register_thing = default_registry.register_thing
```

Note that `collections.UserDict` gives us an easy mechanism for simple registries like this.

https://docs.python.org/3/library/collections.html#collections.UserDict

Note that this requires ALL apis that make use of the registry to have some mechanism for passing in a custom registry.  This pattern should result in easier isolation during testing as well as more flexibility when a library needs to override the base registry.  I also think this approach has minimal extra overhead with respect to implementation and user facing API since the actualy underlying `Registry` API can be maintained as private.

tests/p2p/discv5/test_enr.py
```diff
@@ -0,0 +1,312 @@
...
+            b"key1",
+        ]),
+        rlp.encode([
+            b"signature",
+            0,
+            b"key1",
+            b"value1",
+            b"id",
+        ]),
+    )
```
pipermerriam: This should be parametrized using `pytest.mark.parametrize` so it runs as 4 different tests.

p2p/discv5/enr.py
jannikluhn: For now it's just `int` and `bytes`, but in theory it can be anything -- we might want to add special types for, e.g., libp2p multiaddrs or public key objects in the future. So to keep it easily extendable I'd prefer to keep it `Any`.

p2p/discv5/enr.py
jannikluhn: neat! changed it

p2p/discv5/identity_schemes.py
jannikluhn: Sounds reasonable. I've implemented something like this, but I'm not sure if I fully understood. Please have a look if that's how you imagined it.

p2p/discv5/identity_schemes.py
```diff
@@ -0,0 +1,115 @@
...
+    def validate_signature(cls, enr: "ENR") -> None:
+        pass
+
+    @classmethod
+    @abstractmethod
+    def extract_node_address(cls, enr: "ENR") -> bytes:
+        pass
+
+
+@default_identity_scheme_registry.register
```
pipermerriam: You might expose the `register` function as a top level module item that users can import to register a new identity scheme but that's only relevant if you want to expose this as a public API.  For now I think it's fine to keep it private.

tests/p2p/discv5/test_enr.py
```diff
@@ -0,0 +1,345 @@
...
+        return enr.signature[:cls.private_key_size]
+
+
+@pytest.fixture
+def mock_identity_scheme():
+    return MockIdentityScheme
+
+
+@pytest.fixture
+def test_identity_scheme_registry(mock_identity_scheme):
```
pipermerriam: because pytest is *greedy* I've tended to avoid naming anything in a module that pytest scrapes with `test_` (or classes that start with `Test`) because it sometimes tries to run them as tests.  Naming schemes to get around this that seem appropriate are just to call this `identity_scheme_registry` or `identity_scheme_registry_for_testing`

PR #588, opened 2019-05-06
Add plugin that helps with performance benchmarking

cburgdorf:
### What was wrong?

We don't have good tooling to measure Trinitys run time performance. For things that aren't easily quantifiable in isolation (e.g. with `timeit`), our best option so far is to just run Trinity and manually take some numbers. This PR tries to improve that a little.

### How was it fixed?

1. The `PerformanceTrackerPlugin` roughly does the following:
1.1 Measure the sync time from boot up to some target block
1.2 Also measure the time from the start of sync up to the target block (in other words don't include the upfront time until syncing has startet)
1.3 Write report to log and append it to the `performance_report.txt` file
1.4 Shutdown Trinity

2. The `TrackerRunnerPlugin` plugin is a `BaseMainProcessPlugin` plugin is a minimal viable runner that runs Trinity up to `n` runs with performance tracking enabled.
2.1 E.g. `trinity track-perf --track-perf-target-block 100000 --trag-perf-run-count 5`
2.2 After all runs are finished one can look at the results in `performance_report.txt` which should look like this:

```
Performance tracking finished run: Unnamed run #1
Hit target block 100
Total duration: 15s
Actual duration: 0s
#####################
Performance tracking finished run: Unnamed run #2
Hit target block 100
Total duration: 15s
Actual duration: 5s
#####################
Performance tracking finished run: Unnamed run #3
Hit target block 100
Total duration: 20s
Actual duration: 5s
#####################
```

With the default configuration, the tracker looks up the current head every 5 seconds which means that it won't be accurate to the very second. However, the general idea is that this should make it very easy to e.g. run Trinity ten times in a row up to block `1m` or something and hence give one a rough measure about its performance.

-  [ ] add changelog entry

#### Cute Animal Picture

### DIFF
--- a/docs/guides/index.rst
+++ b/docs/guides/index.rst
@@ -10,3 +10,4 @@ This section aims to provide hands-on guides to demonstrate how to use Trinity. 
    quickstart
    architecture
    writing_plugins
+   measuring_runtime_sync_performance
--- a/docs/guides/measuring_runtime_sync_performance.rst
+++ b/docs/guides/measuring_runtime_sync_performance.rst
@@ -1,0 +1,93 @@
+Measuring runtime sync performance
+==================================
+
+Trinity comes with a built-in plugin that allows us to measure its runtime performance.
+The plugin is currently limited to a single use case which is measuring the time it takes to
+sync from a blank slate to some target block.
+
+
+Measuring runtime performance
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To start performance testing with default settings simply run ``trinity track-perf``. This will
+kick off a series of tasks:
+
+1. Boot the performance runner
+2. Rename Trinity's currently used database with a timestamp suffix
+3. Run ``trinity`` and monitor its performance until block ``100000`` is hit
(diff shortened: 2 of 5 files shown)


cburgdorf:
Just dumping another idea here. We could add flags such as `--bin-a /path/to/trinity-a --bin-b path/to/trinity-b` so that the runner alternates between launching Trinity from two different locations, essentially A/B testing them. This might be a bit more accurate compared to first completing 10 runs on version x and then completing 10 runs on version y. Because it might be that network conditions have changed in between.

cburgdorf:
A/B testing (as in continously toggling between two different programs/branches) is now implemented. One thing that would be super cool is to automatically output graphics such as those seen here https://www.parity.io/performance-analysis/

cburgdorf:
@pipermerriam I added a bit of docs to this today. The code isn't really sophisticated but it's also 100 % plugin code that doesn't affect anything else in trinity and I think having this in `master` would be helpful (otherwise, testing branches does always mean I have to cherry-pick this in first).

Issue #590, opened 2019-05-07
Node crashes during handshake


cburgdorf:
### What is wrong?

```
   ERROR  05-07 11:56:28           ETHPeerPool  Unexpected error during auth/p2p handshake with <Node(0xc8ec321c76436354673931caff416e95cf0d268eab2ac1e73cceaf2ea0e12f492b6a22905627472bb02ff6fdb0085a7fe6139a9cafa8ed4d0fadf190a39102e9@45.112.205.131:7841)>
Traceback (most recent call last):
  File "/home/ubuntu/trinity/p2p/peer_pool.py", line 327, in connect
    peer = await self.wait(handshake(remote, self.get_peer_factory()))
  File "/home/ubuntu/trinity/p2p/cancellable.py", line 20, in wait
    return await self.wait_first(awaitable, token=token, timeout=timeout)
  File "/home/ubuntu/trinity/p2p/cancellable.py", line 42, in wait_first
    return await token_chain.cancellable_wait(*awaitables, timeout=timeout)
  File "/home/ubuntu/trinity/venv/lib/python3.6/site-packages/cancel_token/token.py", line 152, in cancellable_wait
    return done.pop().result()
  File "/home/ubuntu/trinity/p2p/peer.py", line 113, in handshake
    ) = await auth.handshake(remote, factory.privkey, factory.cancel_token)
  File "/home/ubuntu/trinity/p2p/auth.py", line 62, in handshake
    initiator, reader, writer, token)
  File "/home/ubuntu/trinity/p2p/auth.py", line 88, in _handshake
    writer.write(auth_init)
  File "/usr/lib/python3.6/asyncio/streams.py", line 300, in write
    self._transport.write(data)
  File "uvloop/handles/stream.pyx", line 671, in uvloop.loop.UVStream.write
  File "uvloop/handles/handle.pyx", line 159, in uvloop.loop.UVHandle._ensure_alive
RuntimeError: unable to perform operation on <TCPTransport closed=True reading=False 0x2d2592e8>; the handler is closed
 WARNING  05-07 11:56:28           ETHPeerPool  Task <coroutine object BasePeerPool.maybe_connect_more_peers at 0x7f79b966cc50> finished unexpectedly: unable to perform operation on <TCPTransport closed=True reading=False 0x2d2592e8>; the handler is closed
```

### How can it be fixed

Looks like we need to catch the `RuntimeError` there.

PR #591, opened 2019-05-07
Catch RuntimeError during handshake

cburgdorf:
### What was wrong?

A runtime error inside the event loop may happen during handshake and cause the node to crash (#590)

### How was it fixed?

Catch the `RuntimeError` and raise a `HanshakeFailure`

- [ ] add release notes entry

#### Cute Animal Picture

### DIFF
--- a/p2p/auth.py
+++ b/p2p/auth.py
@@ -85,7 +85,11 @@ async def _handshake(initiator: 'HandshakeInitiator', reader: asyncio.StreamRead
     initiator_nonce = keccak(os.urandom(HASH_LEN))
     auth_msg = initiator.create_auth_message(initiator_nonce)
     auth_init = initiator.encrypt_auth_message(auth_msg)
-    writer.write(auth_init)
+
+    try:
+        writer.write(auth_init)
+    except RuntimeError as err:
+        raise HandshakeFailure("Error during handshake with {initiator.remote!r}: {err}") from err
 
     auth_ack = await token.cancellable_wait(
         reader.read(ENCRYPTED_AUTH_ACK_LEN),
@@ -94,7 +98,7 @@ async def _handshake(initiator: 'HandshakeInitiator', reader: asyncio.StreamRead
     if reader.at_eof():
         # This is what happens when Parity nodes have blacklisted us
         # (https://github.com/ethereum/py-evm/issues/901).
-        raise HandshakeFailure("%s disconnected before sending auth ack", repr(initiator.remote))
+        raise HandshakeFailure(f"{initiator.remote!r} disconnected before sending auth ack")
 
     ephemeral_pubkey, responder_nonce = initiator.decode_auth_ack_message(auth_ack)
     aes_secret, mac_secret, egress_mac, ingress_mac = initiator.derive_secrets(


pipermerriam:
Any chance you have the stacktrace laying around for this?

p2p/auth.py
```diff
@@ -85,7 +85,11 @@
     initiator_nonce = keccak(os.urandom(HASH_LEN))
     auth_msg = initiator.create_auth_message(initiator_nonce)
     auth_init = initiator.encrypt_auth_message(auth_msg)
-    writer.write(auth_init)
+
+    try:
+        writer.write(auth_init)
+    except RuntimeError:
+        raise HandshakeFailure(f"Runtime error during handshake with {initiator.remote!r}")
```
pipermerriam: Maybe this should be `raise HandshakeFailure("Error during handshake with {initiator.remote!r}: {err}") from err` to re-raise from original location of `RuntimeError`?

carver:
Hm, I don't love catching such a generic exception and assuming we know why it was raised. At least we keep the underlying message in the `HandshakeFailure`, which is good. The only other option I can think of is to check if the writer transport is open before trying to write. I think that's better, but not enough to go make sure that it actually handles all the cases that are caught by the `RuntimeError` :) ... so I guess nothing to see here :see_no_evil:

pipermerriam:
How about we check that the writer is closed in the exception block to try and *verify* that we're in the expected error state?  If somehow the write isn't closed we could just re-raise the original exception.

cburgdorf:
@carver 

>Hm, I don't love catching such a generic exception and assuming we know why it was raised

Yes, I feel the same but couldn't think of a better way to fix it (and a bit of searching around revealed [aio-libs](https://github.com/aio-libs/) where applying the same strategy even though I seem unable to look it up again). 

> The only other option I can think of is to check if the writer transport is open before trying to write. I think that's better

Yes, maybe that's better.

>but not enough to go make sure that it actually handles all the cases that are caught by the `RuntimeError` :)

Right, but that's probably a good thing. So, if we happen to see another `RuntimeError` that is not related to the writer being closed, we can decide again on a case by case basis what to do.

@pipermerriam 

>How about we check that the writer is closed in the exception block to try and verify that we're in the expected error state? If somehow the write isn't closed we could just re-raise the original exception.

What's the main benefit of that? Not having to check `writer.transport.is_closing()` ([writer.is_closing()](https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamWriter.is_closing) is not available in Python 3.6) for the happy path? I don't think this is a performance critical path so not much to win here by avoiding that check imho.

So, I think, if we don't want to catch all `RuntimeError` at once, which I now feel convinced we shouldn't (thanks @carver for bringing it up), I think I would prefer to check `is_closing` upfront and not catch any `RuntimeError` at all and just see where that would take us :)

pipermerriam:
@cburgdorf you can ignore my comment.  If there is a *pre-check* we can do instead of an *after the error occured* check, that's better.  I was just hastily reviewing this thread.

cburgdorf:
Alright, send a PR to correct this https://github.com/ethereum/trinity/pull/614

PR #592, opened 2019-05-08
Refactor Plugin management

lithp:
This PR started out as an attempt to address [this comment](https://github.com/ethereum/trinity/pull/556#discussion_r281092960). In short: as of #556, `BaseIsolatedPlugin`s receive a context where `context.event_bus == None`. This caused problems: some plugins directly accessed `self.context.event_bus` (which did not exist) instead of accessing `self.event_bus` (which does exist). The simple fix would have been to fix those accesses, but this PR is the result of an experiment where I tried to make that bug impossible.

This resulted in me pulling at a thread for a while, after each change another small change appeared. For example: First I removed `event_bus` from `PluginContext`, but that left `PluginContext` holding nothing but `TrinityBootInfo`, so the next step was to remove `PluginContext`. I'm not married to any of these changes, I'm happy to walk out one that you disagree with! But I *think* that most of these help a little with clarity, and it's nice that this ends up removing 64 lines in exchange for almost no functionality.

Changes:

- Plugins are no longer instantiated during import, before any startup.
- `BasePlugin.configure_parser` is now a classmethod, because the parser is configured before plugin instantiation.
- `BasePlugin.__init__()` accepts a `TrinityBootInfo`, `BaseAsyncStopPlugin.__init__()` accepts an additional event_bus parameter. Previously plugins started in an uninitialized state until `set_context` had been called on them. They now accept their context during construction, such that it's now impossible to create an uninitialized Plugin.

    This required changes to a few of the plugins:
    - `AttachPlugin` and `DbShellPlugin` now always use ipython, when it's available.
    - `SyncerPlugin` now hard-codes the strategies which it uses, although a subclass could easily change them.
   - In general, it's no longer possible to configure plugins by passing information into their constructor.

- `PluginContext` no longer exists. It was a wrapper around `(boot_info, endpoint)`, now each of those parameters are passed in separately. This is done because `BaseAsyncStopPlugin` is the only subclass which uses an event bus so it can now be the only subclass which accepts one.
- `PluginManager.register()` was dropped. Previously `PluginManager()` returned an object that was uninitialized until `.register(Iterable[BasePlugin])` was called. Now those plugins are passed into the constructor and it's impossible to create an uninitialized `PluginManager`.
- `BaseManagerProcessScope.create_plugin_context` is now `create_plugin`. Now that the implementation is so simple there's probably room to simplify this class further, but I think this PR might be at a good checkpoint and that can be saved for later.

#### Cute Animal Picture

bird-stick

### DIFF
--- a/trinity/extensibility/plugin_manager.py
+++ b/trinity/extensibility/plugin_manager.py
@@ -12,11 +12,13 @@ import logging
 from typing import (
     Any,
     Awaitable,
+    cast,
     Dict,
     Iterable,
     List,
     Optional,
-    Union,
+    Type,
+    TypeVar,
 )
 
 from trinity.config import (
@@ -34,7 +36,6 @@ from trinity.extensibility.plugin import (
     BaseIsolatedPlugin,
     BaseMainProcessPlugin,
     BasePlugin,
-    PluginContext,
     TrinityBootInfo,
 )
 from trinity._utils.ipc import (
@@ -42,19 +43,21 @@ from trinity._utils.ipc import (
 )
 
 
+TPlugin = TypeVar('TPlugin', bound=BasePlugin)
+
+
 class BaseManagerProcessScope(ABC):
     """
     Define the operational model under which a
     :class:`~trinity.extensibility.plugin_manager.PluginManager` works. Subclasses
     define whether a :class:`~trinity.extensibility.plugin_manager.PluginManager` is
-    responsible to manage a specific plugin and how its
-    :class:`~trinity.extensibility.plugin.PluginContext` is created.
+    responsible to manage a specific plugin and how it is created.
     """
 
     endpoint: TrinityEventBusEndpoint
 
     @abstractmethod
-    def is_responsible_for_plugin(self, plugin: BasePlugin) -> bool:
+    def is_responsible_for_plugin(self, plugin: Type[BasePlugin]) -> bool:
         """
         Define whether a :class:`~trinity.extensibility.plugin_manager.PluginManager` operating
         under this scope is responsible to manage the given ``plugin``.
@@ -62,11 +65,11 @@ class BaseManagerProcessScope(ABC):
         pass
 
     @abstractmethod
-    def create_plugin_context(self,
-                              plugin: BasePlugin,
-                              boot_info: TrinityBootInfo) -> None:
+    def create_plugin(self,
+                      plugin_type: Type[TPlugin],
+                      boot_info: TrinityBootInfo) -> TPlugin:
         """
-        Create the :class:`~trinity.extensibility.plugin.PluginContext` for the given ``plugin``.
+        Instantiate the given plugin.
         """
         pass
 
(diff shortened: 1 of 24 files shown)


trinity/extensibility/plugin_manager.py
```diff
@@ -162,45 +151,36 @@ def event_bus_endpoint(self) -> TrinityEventBusEndpoint:
         """
         return self._scope.endpoint
 
-    def register(self, plugins: Union[BasePlugin, Iterable[BasePlugin]]) -> None:
```
cburgdorf: I like that this PR allows plugins to get passed right at the instantiation of the `PluginManager` but I just wanted to point out that in the future a `register` API might be valuable to allow hot plugin loading into a running Trinity instance.

trinity/plugins/registry.py
```diff
@@ -41,43 +38,29 @@
...
-
-
-BASE_PLUGINS: Tuple[BasePlugin, ...] = (
-    AttachPlugin(use_ipython=is_ipython_available()),
-    FixUncleanShutdownPlugin(),
-    JsonRpcServerPlugin(),
-    NetworkDBPlugin(),
-    PeerDiscoveryPlugin(),
-    BeaconNodePlugin(),
+BASE_PLUGINS: Tuple[Type[BasePlugin], ...] = (
```
cburgdorf: :heart: I like that. I always thought it was a bit annoying (and wasteful!) to create instances of all plugins even they never actually end up running :+1:

trinity/extensibility/plugin_manager.py
lithp: Absolutely! I don't think there's anything wrong with adding a `register()` method if it's needed. `PluginManager` previously had a few states: 
1. uninitialized, you must call register()
2. ready for plugins to be started, please call prepare()
3. plugins running, call shutdown_blocking()
4. plugins running, call shutdown()
5. plugins have been shutdown

I've always been a little uncomfortable with state machines, I'm never sure how to model them in OOP, but I noticed that the first state, the uninitialized state, wasn't being used, and removing it makes the chance of making a mistake (like forgetting to call `register`) slightly smaller.

trinity/plugins/builtin/fix_unclean_shutdown/plugin.py
```diff
@@ -18,43 +19,48 @@
 
 
 class FixUncleanShutdownPlugin(BaseMainProcessPlugin):
+    logger = logging.getLogger(f'trinity.extensibility.plugin.NetworkDBPlugin')
```
cburgdorf: Do you think it would be reasonable to have a `classmethod` `get_logger` directly on the `BasePlugin` so that plugins that need a logger on the class level do not have to roll their own?

lithp: Definitely, I'll look into it!

lithp: I've tried adding it [here](https://github.com/ethereum/trinity/pull/592/commits/aa08da215e7b716c23a229ea1a8ceea91bbe0c33) but had some trouble with giving each subclass it's own name. What do you think of this approach?

lithp: Okay, after some more thought I think this method is too magical, I'll try to go with the `@classmethod get_logger()` you suggested

lithp:
I'm pretty sure the tests are failing because https://github.com/ethereum/trinity/commit/a9d69d37fc860aab9139bf179deaca6d6ee2434a#diff-bcb3f875bb745a964ec62cfb0d9ac188 added a plugin.

lithp:
I had a lot of fun here, this PR started failing today after a logging change because in python 3.6 it's not possible to pickle a logger. It took me a while to figure out what was causing the problem (The exception complained that `RLock`s couldn't be pickled, but there was no indication of where that `RLock` was coming from). In order to diagnose the problem I found the `Pickler` which multiprocessing uses and changed it's dispatch table to print when the Pickler did any lookups. This gave me enough information to figure out that loggers were causing the issue. Dropping a note here for future reference!

lithp:
@ChihChengLiang, this PR changes `NetworkGeneratorPlugin`, which you recently added. I think that it still works but I didn't test it manually; if this change causes you any problems please tell me and I'll make it right.

PR #595, opened 2019-05-09
Handle corrupt snappy data

pipermerriam:
Fixes https://github.com/ethereum/trinity/issues/292

### What was wrong?

If a peer sends us a data payload that snappy can't decode, it blows the whole node up.

### How was it fixed?

The actual exception being raised is being raised from C code so it doesn't appear we can properly catch it.   This code catches any error out of `snappy.descompress`, logs enough info for us to know what happened, and then raises a `MalformedMessage` exception in it's place.

This also updates the logic of the handshake to not treat `MalformedMessage` as a simple `HandshakeError` since we want to distinguish between *soft* handshake errors like mismatched network id's and hard ones where a peer is sending us bad messages.  The blacklisting code treats malformed messages more severely which I think is what we want here.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

7c5df7d2cd45e5ff9c70d6105dfc3986

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -292,8 +292,12 @@ class BasePeer(BaseService):
 
         try:
             cmd, msg = await self.read_msg()
-        except rlp.DecodingError:
-            raise HandshakeFailure("Got invalid rlp data during handshake")
+        except rlp.DecodingError as err:
+            # We dump the exception trace here to ensure that when/if this
+            # happens we have enough information to better diagnose if this is
+            # an error in our handshake code or just a broken peer.
+            self.logger.debug("Got RLP decoding error during P2P handshake", exc_info=True)
+            raise MalformedMessage(f"Got invalid rlp data during handshake: {err}") from err
 
         if isinstance(cmd, Disconnect):
             msg = cast(Dict[str, Any], msg)
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -294,8 +294,6 @@ class BasePeer(BaseService):
             cmd, msg = await self.read_msg()
         except rlp.DecodingError:
             raise HandshakeFailure("Got invalid rlp data during handshake")
-        except MalformedMessage as e:
-            raise HandshakeFailure("Got malformed message during handshake") from e
 
         if isinstance(cmd, Disconnect):
             msg = cast(Dict[str, Any], msg)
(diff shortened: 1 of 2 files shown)


p2p/peer.py
```diff
@@ -294,8 +294,6 @@ def remove_subscriber(self, subscriber: 'PeerSubscriber') -> None:
             cmd, msg = await self.read_msg()
         except rlp.DecodingError:
             raise HandshakeFailure("Got invalid rlp data during handshake")
```
pipermerriam: This one is also questionable.  We might should treat this like a `MalformedMessage` as well since a peer sending invalid RLP is likely not someone we want to reconnect to right away.

p2p/peer.py
```diff
@@ -292,10 +292,12 @@ def remove_subscriber(self, subscriber: 'PeerSubscriber') -> None:
...
-        except rlp.DecodingError:
-            raise HandshakeFailure("Got invalid rlp data during handshake")
-        except MalformedMessage as e:
-            raise HandshakeFailure("Got malformed message during handshake") from e
+        except rlp.DecodingError as err:
+            # We dump the exception trace here to ensure that when/if this
+            # happens we have enough information to better diagnose if this is
+            # an error in our handshake code or just a broken peer.
+            self.logger.debug("Got RLP decoding error during P2P handshake", exc_info=True)
+            raise MalformedMessage(f"Got invalid rlp data during handshake: {err}") from err
```
pipermerriam: @cburgdorf I went ahead and took the liberty of doing this.  :+1: / :-1:

cburgdorf: I follow your reasoning to raise `MalformedMessage` instead of `HandshakeFailure` but I will add that I don't consider me as someone who has the kind of deep insight to spot subtle issues with that assumption in case there exist any. :sweat_smile:

pipermerriam: Going to hold till Monday for merge just in case there's a reason I'm missing not to do it this way.

carver: Hm, yeah, I think there's a bug here. Looks like the `DecodingError` is already caught and transformed down here as well:
https://github.com/ethereum/trinity/blob/4b2d53c330d267f97c80136029980d62c473d435/p2p/protocol.py#L117-L118

Once that's resolved, I don't see any more problems :+1:

pipermerriam: Good catch

PR #604, opened 2019-05-14
 Make test_demo.py to another CI job and mock the BLS verification

### LINKED ISSUE #354: Make `test_demo.py` to a single CI job

hwwhww:
### What was wrong?
Fix #354 and close #428 

### How was it fixed?
@ralexstokes I can't help but go mocking the BLS ahead for better testing experience! 😄 

1. Move `tests/eth2/core/beacon/state_machines/test_demo.py` -> `tests/eth2/integration/test_demo.py`
2. Mock BLS verification in `test_demo.py`. Execution time goes down from `309.57`s to `155.11`s!

#### Cute Animal Picture
PIXNIO-234578-900x596


hwwhww:
CI passed here: https://circleci.com/gh/hwwhww/trinity/7547

ralexstokes:
works for me :)

PR #605, opened 2019-05-14
NamedTuple for `CommitteeConfig`

hwwhww:
### What was wrong?
Using simple `class CommitteeConfig()` created unnecessary hashing overhead.  

### How was it fixed?
1. Use `NamedTuple` in `CommitteeConfig`
2. New NamedTuple writing style for Eth2Config

#### Cute Animal Picture

PIXNIO-2104830-600x400

### DIFF
--- a/eth2/configs.py
+++ b/eth2/configs.py
@@ -14,75 +14,85 @@ from eth2.beacon.typing import (
 )
 
 
-Eth2Config = NamedTuple(
-    'Eth2Config',
-    (
-        # Misc
-        ('SHARD_COUNT', int),
-        ('TARGET_COMMITTEE_SIZE', int),
-        ('MAX_BALANCE_CHURN_QUOTIENT', int),
-        ('MAX_INDICES_PER_SLASHABLE_VOTE', int),
-        ('MAX_EXIT_DEQUEUES_PER_EPOCH', int),
-        ('SHUFFLE_ROUND_COUNT', int),
-        # Deposit contract
-        ('DEPOSIT_CONTRACT_ADDRESS', Address),
-        ('DEPOSIT_CONTRACT_TREE_DEPTH', int),
-        # Gwei values,
-        ('MIN_DEPOSIT_AMOUNT', Gwei),
-        ('MAX_DEPOSIT_AMOUNT', Gwei),
-        ('FORK_CHOICE_BALANCE_INCREMENT', Gwei),
-        ('EJECTION_BALANCE', Gwei),
-        # Initial values
-        ('GENESIS_FORK_VERSION', int),
-        ('GENESIS_SLOT', Slot),
-        ('GENESIS_EPOCH', Epoch),
-        ('GENESIS_START_SHARD', Shard),
-        # `FAR_FUTURE_EPOCH`, `EMPTY_SIGNATURE` `ZERO_HASH (ZERO_HASH32)`
-        # are defined in constants.py
-        ('BLS_WITHDRAWAL_PREFIX_BYTE', bytes),
-        # Time parameters
-        ('SECONDS_PER_SLOT', Second),
-        ('MIN_ATTESTATION_INCLUSION_DELAY', int),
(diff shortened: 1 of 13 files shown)


hwwhww:
@ChihChengLiang 
I'm still in favor of using `NamedTuple` to be consistent with `Eth2Config`. 😅 

Ping a tie-breaker @ralexstokes @pipermerriam!

eth2/configs.py
```diff
@@ -14,75 +14,85 @@
...
-        ('MAX_PROPOSER_SLASHINGS', int),
-        ('MAX_ATTESTER_SLASHINGS', int),
-        ('MAX_ATTESTATIONS', int),
-        ('MAX_DEPOSITS', int),
-        ('MAX_VOLUNTARY_EXITS', int),
-        ('MAX_TRANSFERS', int),
-    )
-)
+class Eth2Config(NamedTuple):
+    # Misc
```
ralexstokes: this definitely reads better than what was before :)

ralexstokes:
@ChihChengLiang @hwwhww can we not override the `__hash__` of the `CommitteeConfig` and still inherit from `NamedTuple`?

seems like that gives us the best of both worlds :)

looks like this needs a rebase but i would be in favor of merging this in and then we can make another pass down the road with respect to performance... of which there will be much to do :)

ralexstokes:
if there is an issue w/ subclassing `NamedTuple`, could be worthwhile to consider a `dataclass`: https://docs.python.org/3/library/dataclasses.html

ralexstokes:
did we want to do anything to move this along?

ChihChengLiang:
The original motivation for this PR is to fix the lru cache issue (#606).  lru caches the  __hash__ of the CommitteeConfig, so we want the `__hash__` not to change for the same CommitteeConfig. It looks like the dataclass approach also works, but I'm down with any approaches.

```python
from eth2.beacon.state_machines.forks.serenity.configs import Eth2Config
from eth2.configs import CommitteeConfig
from dataclasses import dataclass

a = CommitteeConfig(Eth2Config)
b = CommitteeConfig(Eth2Config)


print(a.__hash__ == b.__hash__)  # False


@dataclass
class CommitteeConfig2:
    SHARD_COUNT: int

    @classmethod
    def from_eth2_config(cls, config):
        cls(**config)


a = CommitteeConfig2(Eth2Config)
b = CommitteeConfig2(Eth2Config)

print(a.__hash__ == b.__hash__)  # True
```

ralexstokes:
@ChihChengLiang does `Eth2Config` have this same issue? i'd even support just removing `CommitteeConfig` and i'd think it is worth revisiting performance hotspots with the newest state transition code

ChihChengLiang:
`SERENITY_CONFIG` (Eth2Config) should be fine, we always refer to the same `SERENITY_CONFIG`. 

I'm so sorry the above python script was so wrong 🙈. Turns out that `dataclass` is unhashable, so lru cache doesn't like it.

> Since a dictionary is used to cache results, the positional and keyword arguments to the function must be hashable. [doc](https://docs.python.org/3/library/functools.html#functools.lru_cache)

```python
from eth2.beacon.state_machines.forks.serenity.configs import SERENITY_CONFIG
from eth2.configs import CommitteeConfig
from dataclasses import dataclass

a = CommitteeConfig(SERENITY_CONFIG)
b = CommitteeConfig(SERENITY_CONFIG)


print(hash(a) == hash(b))  # False


@dataclass
class CommitteeConfig2:
    SHARD_COUNT: int

    @classmethod
    def from_eth2_config(cls, config):
        return cls(SHARD_COUNT=config.SHARD_COUNT)


a = CommitteeConfig2.from_eth2_config(SERENITY_CONFIG)
b = CommitteeConfig2.from_eth2_config(SERENITY_CONFIG)

print(hash(a) == hash(b))  # TypeError: unhashable type: 'CommitteeConfig2'

```

PR #445, opened 2019-03-21
Release notes for v0.1.0-alpha.24

carver:
This is the long-running PR for release notes to be included in the next release.

Community contributors are welcome to comment below with your notes, or open a PR to merge into this branch.

Coders with write access are welcome to add commits on top. Please don't force push, though: it would be sad to accidentally blow away someone else's notes.

PR #608, opened 2019-05-14
tests: update to match `py-evm`: don't skip Constantinople, add Petersburg, use `--fork`; break out long-running categories

veox:
### What was wrong?

Tests for `network=='Constantinople'` are being skipped.

The pinned git submodule is for https://github.com/ethereum/tests/tree/420f443477caa8516f1f9ee8122fafc3415c0f34 (same as tag `v6.0.0-beta.2`, which is not even latest), whereas `py-evm` has https://github.com/ethereum/tests/tree/6b85703b568f4456582a00665d8a3e5c3b20b484 (untagged commit slightly after `v6.0.0-beta.3`, which itself does not include `ConstantinopleFix` - which why we're using something "slightly after").

### How was it fixed?

- [x] Skip `network=='Istanbul'` (leave placeholder).
- [x] Use same `fixtures` commit as `py-evm`.
- [x] Add `petersburg` jobs to CI.
- [x] Use `pytest --fork` arg (copy-paste! from `py-evm`).
- [x] Address test failures (the usual disparity on "synthetic state").
- [x] Address long CI run times:
  * break out categories into their own CI jobs:
    * zero-knowledge;
    * `SSTORE`;
  * switch these break-out categories to use Python 3.7 (a tiny bit more performant than 3.6).

### To-Do

- [x] Clean up commit history

- [x] Add entry to the [release notes PR](https://github.com/ethereum/trinity/pulls?q=is%3Aopen+is%3Apr+label%3A%22Release+Notes%22): https://github.com/ethereum/trinity/pull/445#issuecomment-493034038

### Cute Animal Picture

Felicia the ferret slides out of a wave guide

Source: Felicia the ferret of [Fermilab; via Atlas Obscura](https://www.atlasobscura.com/articles/felicia-ferret-particle-accelerator-fermilab)

### DIFF
--- a/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
+++ b/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
@@ -105,6 +105,7 @@ SLOW_TESTS = (
 # at the commit currently checked out in submodule `fixtures`.
 # Ideally, this list should be empty.
 # WHEN ADDING ENTRIES, ALWAYS PROVIDE AN EXPLANATION!
+# TODO: import from `py-evm` if possible
 INCORRECT_UPSTREAM_TESTS = {
     # The test considers a "synthetic" scenario (the state described there can't
     # be arrived at using regular consensus rules).
@@ -112,6 +113,28 @@ INCORRECT_UPSTREAM_TESTS = {
     # The result is in conflict with the yellow-paper:
     # * https://github.com/ethereum/py-evm/pull/1224#issuecomment-418800369
     ('GeneralStateTests/stRevertTest/RevertInCreateInInit_d0g0v0.json', 'RevertInCreateInInit_d0g0v0_Byzantium'),  # noqa: E501
+    ('GeneralStateTests/stRevertTest/RevertInCreateInInit_d0g0v0.json', 'RevertInCreateInInit_d0g0v0_Constantinople'),  # noqa: E501
+    ('GeneralStateTests/stRevertTest/RevertInCreateInInit_d0g0v0.json', 'RevertInCreateInInit_d0g0v0_ConstantinopleFix'),  # noqa: E501
+
+    # The CREATE2 variant seems to have been derived from the one above - it, too,
+    # has a "synthetic" state, on which py-evm flips.
+    # * https://github.com/ethereum/py-evm/pull/1181#issuecomment-446330609
+    ('GeneralStateTests/stCreate2/RevertInCreateInInitCreate2_d0g0v0.json', 'RevertInCreateInInitCreate2_d0g0v0_Constantinople'),  # noqa: E501
+    ('GeneralStateTests/stCreate2/RevertInCreateInInitCreate2_d0g0v0.json', 'RevertInCreateInInitCreate2_d0g0v0_ConstantinopleFix'),  # noqa: E501
+
+    # Four variants have been specifically added to test a collision type
+    # like the above; therefore, they fail in the same manner.
+    # * https://github.com/ethereum/py-evm/pull/1579#issuecomment-446591118
+    # Interestingly, d2 passes in Constantinople after a py-evm refactor of storage handling,
+    # the same test is already passing in ConstantinopleFix. Since the situation is synthetic,
+    # not much research went into why, yet.
+    ('GeneralStateTests/stSStoreTest/InitCollision_d0g0v0.json', 'InitCollision_d0g0v0_Constantinople'),  # noqa: E501
+    ('GeneralStateTests/stSStoreTest/InitCollision_d1g0v0.json', 'InitCollision_d1g0v0_Constantinople'),  # noqa: E501
+    ('GeneralStateTests/stSStoreTest/InitCollision_d2g0v0.json', 'InitCollision_d2g0v0_Constantinople'),  # noqa: E501
+    ('GeneralStateTests/stSStoreTest/InitCollision_d3g0v0.json', 'InitCollision_d3g0v0_Constantinople'),  # noqa: E501
+    ('GeneralStateTests/stSStoreTest/InitCollision_d0g0v0.json', 'InitCollision_d0g0v0_ConstantinopleFix'),  # noqa: E501
+    ('GeneralStateTests/stSStoreTest/InitCollision_d1g0v0.json', 'InitCollision_d1g0v0_ConstantinopleFix'),  # noqa: E501
+    ('GeneralStateTests/stSStoreTest/InitCollision_d3g0v0.json', 'InitCollision_d3g0v0_ConstantinopleFix'),  # noqa: E501
 }
 
 RPC_STATE_NORMALIZERS = {
@@ -376,8 +399,8 @@ async def validate_uncles(rpc, block_fixture, at_block):
 @pytest.fixture
 def chain_fixture(fixture_data):
     fixture = load_fixture(*fixture_data)
-    if fixture['network'] == 'Constantinople':
-        pytest.skip('Constantinople VM rules not yet supported')
+    if fixture['network'] == 'Istanbul':
+        pytest.skip('Istanbul VM rules not yet supported')
     return fixture
 
 
--- a/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
+++ b/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
@@ -207,10 +207,10 @@ def blockchain_fixture_mark_fn(fixture_path, fixture_name, fixture_fork):
 
 def generate_ignore_fn_for_fork(passed_fork):
     if passed_fork:
-        passed_fork = passed_fork.lower()
+        normalized_fork = passed_fork.lower()
 
         def ignore_fn(fixture_path, fixture_key, fixture_fork):
-            return fixture_fork.lower() != passed_fork
+            return fixture_fork.lower() != normalized_fork
 
         return ignore_fn
 
--- a/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
+++ b/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
@@ -48,16 +48,12 @@ ROOT_PROJECT_DIR = Path(__file__).parent.parent.parent
 BASE_FIXTURE_PATH = os.path.join(ROOT_PROJECT_DIR, 'fixtures', 'BlockchainTests')
 
 SLOW_TESTS = (
-    'Call1024PreCalls_d0g0v0_Byzantium',
-    'Call1024PreCalls_d0g0v0_EIP150',
-    'Call1024PreCalls_d0g0v0_EIP158',
+    'Call1024PreCalls_d0g0v0',
     'ContractCreationSpam_d0g0v0_Homestead',
     'ContractCreationSpam_d0g0v0_Frontier',
-    'ForkStressTest_EIP150',
-    'ForkStressTest_EIP158',
-    'ForkStressTest_Homestead',
-    'ForkStressTest_Frontier',
-    'ForkStressTest_Byzantium',
+    'Create2Recursive_d0g0v0',
+    'Create2Recursive_d0g1v0',
+    'ForkStressTest',
     'stQuadraticComplexityTest/Call50000_d0g1v0.json',
     'stQuadraticComplexityTest/QuadraticComplexitySolidity_CallDataCopy_d0g1v0.json',
     'stQuadraticComplexityTest/Return50000_2_d0g1v0.json',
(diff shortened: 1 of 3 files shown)


veox:
Tests are running, but for some reason aren't showing here. :/

veox:
Failure is in the usual "incorrect upstream" disagreement.

veox:
Enabled `ConstantinopleFix`, too, out of curiosity.

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
```diff
@@ -115,6 +115,24 @@
     # The result is in conflict with the yellow-paper:
     # * https://github.com/ethereum/py-evm/pull/1224#issuecomment-418800369
     ('GeneralStateTests/stRevertTest/RevertInCreateInInit_d0g0v0.json', 'RevertInCreateInInit_d0g0v0_Byzantium'),  # noqa: E501
+    ('GeneralStateTests/stRevertTest/RevertInCreateInInit_d0g0v0.json', 'RevertInCreateInInit_d0g0v0_Constantinople'),  # noqa: E501
```
veox: I think this `INCORRECT_UPSTREAM_TESTS` blob is essentially the same as the one in `py-evm`; wonder if we should make it more accessible in `py-evm` so it can be imported here instead of copy-pasting.

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
```diff
@@ -115,6 +115,24 @@
...
+
+    # The CREATE2 variant seems to have been derived from the one above - it, too,
+    # has a "synthetic" state, on which py-evm flips.
+    # * https://github.com/ethereum/py-evm/pull/1181#issuecomment-446330609
+    ('GeneralStateTests/stCreate2/RevertInCreateInInitCreate2_d0g0v0.json', 'RevertInCreateInInitCreate2_d0g0v0_Constantinople'),  # noqa: E501
+
+    # Four variants have been specifically added to test a collision type
+    # like the above; therefore, they fail in the same manner.
+    # * https://github.com/ethereum/py-evm/pull/1579#issuecomment-446591118
+    # Interestingly, for py-evm - but not trinity! -
```
veox: ... likely because we're using a tagged version of `py-evm` that's older than `master`?..

carver:
> Skip `network=='ConstantinopleFix'` instead.

Wait, shouldn't we be running `Constantinople(Fix)` tests? I thought everything was passing now.

veox:
We've not even pinned the `git submodule` for `fixtures` that has `ConstantinopleFix`. That happened in `py-evm` only.

veox:
6ccfb24 failed locally for me, but it shows to have passed CI here. :/

Possibly need to check if the CI machine _really_ updated its `git submodule`.

tox.ini
```diff
@@ -32,7 +32,9 @@ commands=
     rpc-state-spurious_dragon: pytest -n 4 {posargs:tests/json-fixtures-over-rpc/test_rpc_fixtures.py -k 'GeneralStateTests and not stQuadraticComplexityTest and EIP158'}
     # The following test seems to consume a lot of memory. Restricting to 3 processes reduces crashes
     rpc-state-byzantium: pytest -n 3 {posargs:tests/json-fixtures-over-rpc/test_rpc_fixtures.py -k 'GeneralStateTests and not stQuadraticComplexityTest and Byzantium'}
-    rpc-state-constantinople: pytest -n 3 {posargs:tests/json-fixtures-over-rpc/test_rpc_fixtures.py -k 'GeneralStateTests and not stQuadraticComplexityTest and Constantinople'}
+    # Must match Constantinople but not ConstantinopleFix!
+    rpc-state-constantinople: pytest -n 3 {posargs:tests/json-fixtures-over-rpc/test_rpc_fixtures.py -k 'GeneralStateTests and not stQuadraticComplexityTest and not ConstantinopleFix and Constantinople'}
```
veox: Important: I _think_ this would previously match both `Constantinople` and `ConstantinopleFix`; `py-evm` has a `pytest` helper arg `--fork`, which (I'd guess, haven't checked) matches strictly.

cburgdorf: >I think that without this change, it would now match both `Constantinople` and `ConstantinopleFix`. Therefore this logic contraption

Correct!

>`py-evm` has a pytest helper arg `--fork`, which (I'm hoping without checking :sweat_smile: ) matches strictly.

Yes, that's what it does. I would suggest we copy that over into this repo to avoid such issues. I actually wasn't aware that this happened *before* the repos were split apart.

veox: Note: Added in https://github.com/ethereum/py-evm/pull/1470 there.

veox:
Constantinople and Petersburg tests take >30 minutes on my free CircleCI account. Adding entries to `SLOW_TESTS`...

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
```diff
@@ -70,28 +71,22 @@
...
-    'static_Call50000bytesContract50_1_d1g0v0_Byzantium',
-    'static_Call50000bytesContract50_2_d1g0v0_Byzantium',
-    'static_LoopCallsThenRevert_d0g0v0_Byzantium',
-    'static_LoopCallsThenRevert_d0g1v0_Byzantium',
-    'Call1024PreCalls_d0g0v0_Byzantium',
-    'Call1024PreCalls_d0g0v0_EIP158',
-    'Call1024PreCalls_d0g0v0_EIP150',
-    'Call1024PreCalls_d0g0v0_Byzantium',
-    'Call1024PreCalls_d0g0v0_EIP150',
-    'Call1024PreCalls_d0g0v0_EIP158',
```
veox: The above six lines are duplicates of the first three lines in the list.

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
```diff
@@ -50,16 +51,12 @@
...
     'ContractCreationSpam_d0g0v0_Homestead',
     'ContractCreationSpam_d0g0v0_Frontier',
-    'ForkStressTest_EIP150',
-    'ForkStressTest_EIP158',
-    'ForkStressTest_Homestead',
-    'ForkStressTest_Frontier',
-    'ForkStressTest_Byzantium',
+    'Create2Recursive_d0g0v0',
+    'Create2Recursive_d0g1v0',
+    'ForkStressTest',
```
veox: This test is slow across all forks - however, removal of fork qualifier means that _any_ test that might happen to have this string will be marked `SLOW`.

veox:
PR ready for review, I guess. (Test runs might still take very long, though.)

carver:
Hm, yeah, 30min is pretty long. Byzantium only takes 6min:
```
============ 4698 passed, 4751 skipped, 1 xfailed in 373.23 seconds ============
```

Constantinople:
```
========== 10323 passed, 10384 skipped, 5 xfailed in 1920.96 seconds ===========
```

I just checked the native runs in py-evm, and the numbers of tests look about right. (wow 2x more tests!) Interestingly, Constantinople is less than 2x slower in native, but >5x slower over RPC.

So I don't see an obvious way to quickly shorten the long run. :(  It should help a bit when we release & import the latest py-evm. But since most of the time seems to be RPC overhead, I guess it won't help that much.

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
carver: Are you up for making the changes and/or making an issue in py-evm to expose this API?

cburgdorf:
>My first guess is that it's somewhere in lahja, since I don't remember RPC adding that much overhead before it was handled by events. cc @cburgdorf @lithp

If I'm not mistaken, the only thing that the JSON-RPC server uses the event bus for is to make requests to the PeerPool (e.g. to get the current peer count) but that shouldn't come into play here I guess. Worth taking a closer look into this.

carver:
Yeah, you're right I was conflating two things: the event bus, and rpc running as an isolated plugin. I should have said the latter. So the overhead would be about sending the chain requests over multiprocessing.

cburgdorf:
>So the overhead would be about sending the chain requests over multiprocessing

Mmh..that should not have changed the tests. It's always been that way that the RPC tests would connect via IPC and make the request. In fact, the RPC tests do not even know about the plugin. They just spin up the `IPCServer` directly (nothing changed about that).

https://github.com/ethereum/trinity/blob/a37f26cccc0dead82e94ffa1bbc0ab34607410df/tests/conftest.py#L298-L320

carver:
> > So the overhead would be about sending the chain requests over multiprocessing
> 
> Mmh..that should not have changed the tests. It's always been that way that the RPC tests would connect via IPC and make the request. In fact, the RPC tests do not even know about the plugin. They just spin up the `IPCServer` directly (nothing changed about that).
> 
> https://github.com/ethereum/trinity/blob/a37f26cccc0dead82e94ffa1bbc0ab34607410df/tests/conftest.py#L298-L320

I'm talking about the time before RPCServer was moved to a `BaseIsolatedPlugin` and just ran in the main process. Quite a while ago, now.

cburgdorf:
>I'm talking about the time before `RPCServer` was moved to a `BaseIsolatedPlugin` and just ran in the main process. Quite a while ago, now.

Yes, but what I mean is: From the perspective of these tests, nothing has changed when the `RPCServer` moved into an `BaseIsolatedPlugin`. Afaik, the tests do not spin up a Trinity instance but instead just manually instantiate the `RPCServer`. In fact, I think the `RPCServer` in the tests just runs in the same process all the other tests run.

carver:
> > I'm talking about the time before `RPCServer` was moved to a `BaseIsolatedPlugin` and just ran in the main process. Quite a while ago, now.
> 
> Yes, but what I mean is: From the perspective of these tests, nothing has changed when the `RPCServer` moved into an `BaseIsolatedPlugin`. Afaik, the tests do not spin up a Trinity instance but instead just manually instantiate the `RPCServer`. In fact, I think the `RPCServer` in the tests just runs in the same process all the other tests run.

Ah right, which also explains why there wasn't much of a hit from https://github.com/ethereum/py-evm/pull/1212

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
veox: I'll make issues, as you outlined in:

> Yeah, to avoid committing the TODO's to master, how about an issue in trinity that links to an issue in py-evm which lists the methods to expose in the API.

veox:
> 30+ minute test jobs make me sad, but I'm not sure where to start on optimizing those yet.

We could switch those CI jobs to run on Python 3.7 instead of 3.6...

That said, it might not be that big of a deal for someone not using CircleCI's free tier. 

-----

Another approach is to run some classes of tests separately, similar to how `stQuadraticComplexityTest` are already done; candidates:

* `ZeroKnowledge` (>1300);
* recursive and stack exhaustion (~100?.. seconds each).

carver:
> > 30+ minute test jobs make me sad, but I'm not sure where to start on optimizing those yet.
> 
> We could switch those CI jobs to run on Python 3.7 instead of 3.6...

:+1: 

> 
> That said, it might not be that big of a deal for someone not using CircleCI's free tier.

Eh, `max(len(job) for job in ci_jobs)` still matters, even with high concurrency. So I like your next approach...

> 
> Another approach is to run some classes of tests separately, similar to how `stQuadraticComplexityTest` are already done; candidates:
> 
> * `ZeroKnowledge` (>1300 items, many last 0.15 seconds each; easily 200 seconds total);
> * recursive and stack exhaustion (~100 (orly?) items, up to a few seconds each; easily 120 seconds total).
> 
> I've made [a separate branch](https://github.com/veox/trinity/commits/dont-skip-constantinople-tests%2Btest-zk-separately) to test out the latter hypothesis.
> 
> EDIT:
> 
> * ZK tests [took 5:33 to run](https://circleci.com/gh/veox/trinity/2132) (06:11 for entire CI job);
> * [25:07 for Constantinople](https://circleci.com/gh/veox/trinity/2129) with ZK tests ignored via `pytest -k`;
> * [24:55 for Petersburg](https://circleci.com/gh/veox/trinity/2112) with ZK tests ignored via `pytest -k`.
> 
> So, a matching reduction from the previous 30+ minutes. It's a viable strategy...

Yeah, I'm definitely in favor of this direction.

veox:
Identified `SSTORE` as the largest offender. On a check-everything local `pytest tests` run, fixtures from `stSStoreTest` made up 5722 of 10000 slowest test runs.

Breaking them out further reduced [Petersburg job](https://circleci.com/gh/veox/trinity/2484) run time to 04:10 (4:41 job total).

[`SSTORE` job](https://circleci.com/gh/veox/trinity/2473) took 19:20 to run, though (20:03 total), because it runs the tests for all forks. PR #623 is most likely to help with this.

veox:
Rebased on `master` to get PR #623; `sstore` job now takes 8-9 minutes (jobs [2675](https://circleci.com/gh/veox/trinity/2675), [2684](https://circleci.com/gh/veox/trinity/2684), [2685](https://circleci.com/gh/veox/trinity/2685)):

```
================== 11253 passed, 7 xfailed in 525.26 seconds ===================
================== 11253 passed, 7 xfailed in 506.88 seconds ===================
================== 11253 passed, 7 xfailed in 516.59 seconds ===================
```

Switching the break-out categories to test with Python 3.7; will post results here.

veox:
Cleaned up commits; unmarking WIP; ready for review.

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
```diff
@@ -188,13 +205,33 @@ def blockchain_fixture_mark_fn(fixture_path, fixture_name):
         return pytest.mark.xfail(reason="Listed in INCORRECT_UPSTREAM_TESTS.")
 
 
+def generate_ignore_fn_for_fork(passed_fork):
+    if passed_fork:
+        passed_fork = passed_fork.lower()
```
pipermerriam: *nit* we'd normally do this as `normalized_fork = passed_fork.lower()` to avoid destroying the initial data.

veox: I'll fix that up real quick; but note it's a copy-paste from https://github.com/ethereum/py-evm/blob/ddefa2fbad378aad6562239eb087e8eb27794fc1/tests/json-fixtures/test_blockchain.py#L231.

pipermerriam:
will merge when green.  Thanks a bunch.

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
pipermerriam: I'll admit we aren't always perfectly consistent about these things.  Thanks for the cleanup.

PR #610, opened 2019-05-15
Add Firehose protocol

lithp:
Opening a PR so I can track CircleCI and watch out for merge conflicts.

### How was it fixed?

- Created stub protocol implementation, work through most of the boilerplate
- Give the request sever a chaindb it can read from
- Add a test that you can read from the chaindb and send responses
- Implemented `GetNodeChunk` command, which returns all the nodes up to 2 layers deep from the requested node.
- Implemented a simple syncer which only uses the `GetNodeChunk` command and slurps over the entire account trie.

### To-Do

- [x] Implement some Firehose methods
- [x] Add a `GetLeaves` command?
- [x] Add request_id aided request pipelining, that'll make the sync a lot faster.
- [ ] Fix merge conflicts
- [ ] It took me about 90 lines to add a new type of command. Is there a way to reduce some of this boilerplate?
- [ ] Update this PR once https://github.com/ethereum/trinity/pull/617 is merged?
- [ ] What should be in the `Status` message?
- [ ] It might be nice to have a `make_nibbles` helper.
- [ ] Add some kind of timeout to requests so peers can't ask for too much
- [ ] Add validation to requests, what happens if peers ask for state roots we don't have?
- [ ] Test trie iteration with a larger database, the current tests are tiny and likely missing edge-cases
- [ ] The simpe syncer doens't recurse into storage tries but should
- [ ] We shouldn't need to specify `request_id` when defining each command.
- [x] Start exercising request ids!
  - [x] Send a random `request_id` with each request
  - [x] Add support for sending multiple requests at a time
  - [x] Correlate incoming responses with outbound requests
- [ ] The first request the simple syncer returns a lot of data, but it has a hard time dealing with the bottom of the trie, where "2 layers" only constitutes a few nodes. This protocol would be better at traversing the lower levels if `GetNodeChunk` returned something like "2 layers or 100 nodes, whichever is larger".

- [ ] Clean up commit history

#### Cute Animal Picture


trinity/protocol/firehose.py
```diff
@@ -233,10 +238,18 @@ def __init__(self, db: BaseAsyncChainDB,
...
+            state_root = cast(Hash32, msg['state_root'])  # TODO: is this cast correct?
+            print(f'state_root: {state_root}')
+            await self.handle_get_leaf_count(peer, request_id, state_root)
+
+    async def handle_get_leaf_count(self, peer: FirehosePeer,
+                                    request_id: int, state_root: Hash32) -> None:
+
+        count = 0
+        for _ in trie_iterator(self.db, state_root):
+            count += 1
```
carver: Ugh, this is why I don't like this part of the protocol. *Maybe* it's okay, if it's only an estimate.

lithp: Heh, what don't you like about this?

carver: That it would take ages to iterate over all the leaves.

We don't want to expose network protocols that cause us to do huge amounts of work. Especially if a partial answer is unhelpful. If it's okay to give an estimate, like a 4x order-of-magnitude guess at the total leaf count, then we can just probe in a couple times.

pipermerriam: Is this red queen stuff?

carver: Yeah, there is an attempt to modify the command so we ask the caller to figure out how big of a bucket to ask for (by letting them request the total leaf size up front) rather than have the server respond with either nodes or leaf buckets.

carver:
Hooray for showing that disk I/O and network latency are the top two contributors to slowness! (to be resolved with a database re-arch and pipelining) :tada:

trinity/protocol/firehose.py
```diff
@@ -0,0 +1,2407 @@
...
+        leaves, proof = pickle.loads(value)
+
+        return {
+            'prefix': tuple(key),
+            'leaves': leaves,
+            'proof': proof,
+        }
+
+
+class FirehoseRequestServer(BaseRequestServer):
```
cburgdorf: This is currently build on the `BaseRequestServer`  which (I believe) is on a path to removal. The `ETHRequestServer` and `LESRequestServer` are already built on top of the `BaseIsolatedRequestServer` which can be run in its own process. The reason only reason that kept me from killing (or replacing) the `BaseRequestServer` already was that the beacon node still uses it for something that needs direct exposure to the peer pool (iterating over the peers) which we currently can't handle yet outside of the peer pool process.

But I *think* the `FirehoseRequestServer` could follow the `ETHRequestServer` and `LESRequestServer` and run in its own process already.

That said, it adds further friction during development so it's totally understandable to delay that part while things are still in flux.

PR #613, opened 2019-05-15
[POC] Trio based Service implementation

pipermerriam:
A proof of concept I want to show off for a re-write of our `Service` concept in `trio`.  There are things here that aren't necessarily trio specific and that we could backport to the legacy `BaseService`.

This implementation focuses on keeping the service logic and the logic of running the service separated.  Here is what a really simple service that just waits for a `KeyboardInterrupt` would look like.

```python
@Service.from_function
async def SimpleService(service):
    try:
        await service.wait_cancelled()
    except KeyboardInterrupt:
        ...  # could do cleanup here
```

This is equivalent to:

```python
class SimpleService(ServiceLogic):
    async def run(self, service):
        try:
            await service.wait_cancelled()
        except KeyboardInterrupt:
            ...  # could do cleanup here
```

You could run either of these like this:

```python
>>> service = SimpleService()
>>> await service.run()
```

The biggest change is that the logic part of the service is handed a *handle* to the API for managing the service so that it can do things like triggering cancellation, waiting for cancellation, checking status.

I'm particularly proud of the way error handling is done.  When the logic function exits by throwing an exception, we hold onto that, do all the cleanup, and then re-raise it as if it came from the logic function itself, allowing errors that occured in the background where the service was running to propagate to the process it was run in.

You can see in the tests that the lifecycle of the service is always maintained across multiple different exit conditions.

### DIFF
--- a/p2p/trio_service.py
+++ b/p2p/trio_service.py
@@ -1,0 +1,196 @@
+from abc import ABC, abstractmethod
+import logging
+import sys
+from types import TracebackType
+from typing import Any, Callable, Awaitable, Optional, Tuple, Type
+
+import trio
+
+import trio_typing
+
+
+class ServiceAPI(ABC):
+    @property
+    @abstractmethod
+    def has_started(self) -> bool:
+        ...
+
+    @property
+    @abstractmethod
+    def is_running(self) -> bool:
+        ...
+
+    @property
+    @abstractmethod
+    def is_cancelled(self) -> bool:
+        ...
+
+    @property
+    @abstractmethod
+    def is_stopped(self) -> bool:
+        ...
+
+    @property
+    @abstractmethod
+    def did_error(self) -> bool:
+        ...
+
+    @abstractmethod
+    def cancel(self) -> None:
+        ...
+
+    @abstractmethod
+    async def wait_started(self) -> None:
+        ...
+
+    @abstractmethod
+    async def wait_cancelled(self) -> None:
+        ...
+
+    @abstractmethod
+    async def wait_stopped(self) -> None:
+        ...
+
+
+LogicFnType = Callable[[ServiceAPI], Awaitable[Any]]
+
+
+class ServiceLogic(ABC):
+    @abstractmethod
+    async def run(self, service: ServiceAPI) -> None:
+        ...
+
+
+class Service(ServiceAPI):
+    logger = logging.getLogger('p2p.trio_service.Service')
+
+    logic_class: ServiceLogic
+
+    _run_error: Optional[Tuple[
+        Optional[Type[BaseException]],
+        Optional[BaseException],
+        Optional[TracebackType],
+    ]] = None
+
+    def __init__(self, *args, **kwargs) -> None:
+        # initialize the service logic
+        self.logic = self.logic_class(*args, **kwargs)
+
+        # events
+        self._started = trio.Event()
+        self._cancelled = trio.Event()
+        self._stopped = trio.Event()
+
+        # locks
+        self._run_lock = trio.Lock()
+
+    async def _handle_cancelled(self, nursery: trio_typing.Nursery):
+        """
+        Handles the case where cancellation occurs because the
+        `event.set_cancelled()` has been called, this propagates that to force
+        the nursery to be cancelled.
+        """
+        self.logger.debug('_handle_cancelled waiting for cancellation')
+        await self._cancelled.wait()
+        self.logger.debug('_handle_cancelled triggering nursery cancellation')
+        nursery.cancel_scope.cancel()
+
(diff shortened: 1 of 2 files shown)


carver:
> I'm particularly proud of the way error handling is done. When the logic function exits by throwing an exception, we hold onto that, do all the cleanup, and then re-raise it as if it came from the logic function itself, allowing errors that occured in the background where the service was running to propagate to the process it was run in.

Yeah, that sounds great!

A thing I haven't seen a good pattern for yet is:

```py
class InnerService(ServiceLogic):
    async def do_thing(self):
        # do thing that takes a long time, and might be cancelled
    async def run(self, service):
        ...

class OuterService(ServiceLogic):
    async def sub_task_that_is_allowed_to_fail(self):
        self.service.loop.call_later(0.5, self.inner_service.cancel_nowait())  # or however trio does it
        while True:
            await self.wait(inner_service.do_thing())

outer = OuterService(InnerService())
await outer.launch()  # wait for the start event
await outer.sub_task_that_is_allowed_to_fail()
```

Right now, the inner service is raising an `OperationCancelled` and it's hard (impossible?) to tell the difference between which service raised it. I'm curious what this might look like in the new approach.

pipermerriam:
@carver trio handles this natively!

The cancellations are tied to their cancel scopes, I'm going to write a test that demonstrates this.

p2p/trio_service.py
```diff
@@ -0,0 +1,196 @@
...
+
+    @property
+    @abstractmethod
+    def is_cancelled(self) -> bool:
+        ...
+
+    @property
+    @abstractmethod
+    def is_stopped(self) -> bool:
+        ...
```
cburgdorf: Just curious. Is there any advantage of the `...` syntax over `pass`?

p2p/trio_service.py
```diff
@@ -0,0 +1,196 @@
...
+        await self._cancelled.wait()
+
+    async def wait_stopped(self) -> None:
+        await self._stopped.wait()
+
+    #
+    # Service from simple function API
+    #
+    @classmethod
+    def from_function(cls, logic_fn: LogicFnType) -> 'Service':
```
cburgdorf: Fancy helper!

cburgdorf:
This looks really nice and clean! I'm getting excited for the switch :)

p2p/trio_service.py
pipermerriam: I was taking it out for a test spin.  I think it stands out more than `pass` to indicate that there the function is not implemented.

pipermerriam:
@cburgdorf what do you think about adopting the `logic/manager` pattern for plugins?

p2p/trio_service.py
```diff
@@ -0,0 +1,196 @@
...
+LogicFnType = Callable[[ServiceAPI], Awaitable[Any]]
+
+
+class ServiceLogic(ABC):
+    @abstractmethod
+    async def run(self, service: ServiceAPI) -> None:
+        ...
+
+
+class Service(ServiceAPI):
```
cburgdorf: We probably still want things like `add_child_service` in this API?

cburgdorf:
>what do you think about adopting the logic/manager pattern for plugins?

Yep, makes a lot of sense to me. So far, when it comes to implementing the actual logic part of a plugin we don't have anything streamlined yet which comes with some annoying consequences. These are my favorite two:

- There is no well defined pattern to choose between plugin logic which in the case of errors should take the entire Trinity node down or keep it running (we do have such cases but the implementation feels more ad hoc)

- We have this [clunky thing where we create a service solely to create a parent CancelToken](https://github.com/ethereum/trinity/blob/15958c31b928a5a8b6c95dcd9be91eee8b815224/trinity/plugins/builtin/peer_discovery/plugin.py#L91)

I guess with the logic/manager pattern these things could become easier.

If I understand your prototype code correctly you basically create services either by passing the logic into the `Service` constructor or just use the `from_function` classmethod that you can even just use as a decorator on a simple function to "upgrade" it to a fully fledged service. How do you think we would handle child services?

So, let me just think out loud here: I imagine the plugin itself would have all the manager functionality in itself and then expose APIs such as `run_function` and `run_daemon_function` for the user to run simple functions which behind the scenes are upgraded to proper services similar to what your `from_function` method is doing. And depending on whether an error should take down the whole node or not you'd choose to launch it with `run_deamon_function` or just `run_function`.

And from the perspective  of the `PluginManager` each plugin would then simply be a top level service that may wrap several child services but that would be transparent from that angle.

Is this going into the direction you were thinking?

pipermerriam:
@cburgdorf see p2p/trio_service.py for an updated version of this.  Closing this as I don't intend to maintain it right now and the service logic is evolving as I'm using it to implement discovery.

PR #615, opened 2019-05-15
[POC/WIP] - Trio based discovery service

pipermerriam:
PR open for my own tracking purposes.  Nothing to see here....


pipermerriam:
Current thoughts on isolating this into smaller pieces.

- Service implementation and tests
- TrioIsolatedProcessPlugin implementation and tests
- Extract "preferred peers" concept from discovery, convert to use peer backend system, convert to trio based plugin.

Nasty part is going to be unraveling `p2p/discovery.py` but it looks doable.

pipermerriam:
Extractable piece that has bubbled out of this is elimination of the `TrinityEventBusEndpoint` subclass via a new `BaseService` that does all of the fancy announcement stuff that is needed.

PR #617, opened 2019-05-16
Move RequestServer into isolated plugin

cburgdorf:
### What was wrong?

It is our long term plan to have the RequestServer, the PeerPool and the Sync run in their dedicated isolated processes. The benefits are:

- The `PeerPool` will be free from any CPU intensive work (e.g. block verification etc)
- Leverage multiple CPUs better
- Have pieces be more reusable (e.g RequestServer plugin can be used for beacon node)
- Have each process be a better performance analysis target via `py-spy`
- ???

Currently, these three things are still together in one process that is so far know as the `networking` process.

### How was it fixed?

- Upgraded to unreleased Lahja version that enables us to broadcast on a higher frequency (by reducing unnecessary cross process propagation when it isn't needed)
- Exposed the relevant peer messages on the event bus
- Created ProxyPeer classes that allow to forward actions to the actual peers in the peer pool
- Created a new IsolatedRequestServer that consumes relevant peer pool messages via event bus and uses proxy peers to communicate back responses
- Created a new isolated plugin that runs the isolated request server

### To-Do

- [x] Clean up commit history

- [ ] Add entry to the [release notes PR](https://github.com/ethereum/trinity/pulls?
q=is%3Aopen+is%3Apr+label%3A%22Release+Notes%22)
- [x] Run `BCCRequestServer` via the same plugin
- [ ] Release new Lahja and update `setup.py`

#### Cute Animal Picture


trinity/protocol/eth/peer.py
```diff
@@ -95,9 +120,95 @@ def handle_sub_proto_msg(self, cmd: Command, msg: _DecodedMsgType) -> None:
...
+        async for ev in self.wait_iter(self.event_bus.stream(SendBlockHeadersEvent)):
+            try:
+                peer = self.get_peer(ev.remote)
+            except PeerConnectionLost:
+                pass
+            else:
+                peer.sub_proto.send_block_headers(ev.headers)
+
+    async def handle_send_block_bodies_events(self) -> None:
+        async for ev in self.wait_iter(self.event_bus.stream(SendBlockBodiesEvent)):
```
pipermerriam: These all look so similar.  Maybe replace with

```python
async def handle_event(self, StreamEventType, event_handler_fn):
    async for ev in self.wait_iter(self.event_bus.stream(StreamEventType)):
        try:
            peer = self.get_peer(ev.remote)
        except PeerConnectionLost:
            pass
        else:
            event_handler_fn(peer, ev)
```

cburgdorf: Yep, I cleaned this up a bit more. Not sure what you think about the lambdas. I know you are not a fan but I couldn't figure out an alternative that I was satisfied with.

cburgdorf:
@pipermerriam This can be given a full review now. It is still blocked on the upcoming lahja release but other than that it has passed my own judgement :)

trinity/protocol/eth/servers.py
```diff
@@ -148,42 +163,35 @@ def __init__(self, db: BaseAsyncChainDB, token: CancelToken) -> None:
...
+                          remote: Node,
+                          cmd: Command,
                           msg: protocol._DecodedMsgType) -> None:
-        peer = cast(ETHPeer, base_peer)
 
-        ignored_commands = (
-            commands.Transactions,
-            commands.NewBlockHashes,
-        )
+        self.logger.debug("Peer %s requested %s", remote, cmd)
```
carver: This should probably be `debug2` before merging, otherwise I expect it to be quite noisy.

trinity/protocol/les/servers.py
```diff
@@ -39,28 +50,34 @@ class LESPeerRequestHandler(BasePeerRequestHandler):
...
+
+        self.logger.debug("Peer %s requested %s", remote, cmd)
+        peer = LESProxyPeer.from_node(remote, self.event_bus, self.broadcast_config)
         if isinstance(cmd, commands.GetBlockHeaders):
-            block_request_kwargs = cast(Dict[str, Any], msg)
-            await self._handler.handle_get_block_headers(peer, block_request_kwargs)
+            await self._handler.handle_get_block_headers(peer, cast(Dict[str, Any], msg))
         else:
-            self.logger.debug("%s msg from %s not implemented", cmd, peer)
+            self.logger.debug("%s msg not handled yet, need to be implemented", cmd)
```
carver: `s/need/needs/`

cburgdorf:
Just in case anyone wants to review this, hold back, I'm testing some things out to fine tune this.

tests/core/json-rpc/test_ipc.py
```diff
@@ -87,6 +87,8 @@ def can_decode_json(potential):
         request_msg,
         event_loop):
 
+    # Give the event subscriptions a moment to propagate
+    await asyncio.sleep(0.01)
```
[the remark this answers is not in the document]
cburgdorf: Check out the last commit. It gets rid of the sleeps but depends on https://github.com/ethereum/lahja/pull/72 getting merged before.

Btw, I don't think that will bite us in real code. In general, the code should be written in a way that it can handle cases of the other end not being fully booted yet. E.g. we already have things like the `PeerPool` reaching out to the discovery to get peers but then timing out on the first request if the discovery isn't yet up and running to answer it.

I think this is more an issue of the test cases where we wire things together manually and have strict requirements about the other side being fully up and running yet. 

I know this answer is kinda fuzzy and it might be wishful thinking but I *think* we are safe :sweat_smile:

tests/conftest.py
```diff
@@ -126,7 +126,7 @@ def async_process_runner():
         pass
 
 
-@pytest.fixture(scope='module')
+@pytest.fixture(scope='function')
```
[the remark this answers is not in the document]
cburgdorf: Ah! Didn't know. Sure!

trinity/protocol/common/events.py
```diff
@@ -41,11 +54,24 @@ def expected_response_type() -> Type[PeerCountResponse]:
...
 
 
-class DisconnectPeerEvent(BaseEvent):
+class DisconnectPeerEvent(HasRemoteEvent):
     """
     Event broadcasted when we want to disconnect from a peer
     """
 
     def __init__(self, remote: Node, reason: DisconnectReason) -> None:
         self.remote = remote
```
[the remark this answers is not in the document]
cburgdorf: We need that event for the `bound` in https://github.com/ethereum/trinity/pull/617/files#diff-8067591e62bc9b699e77ac5f8671eb6eR47

Not sure how we would write the `run_event_daemon` otherwise?  Maybe with a `typing.Protocol` but that would mean it would work with *anything*  that has a `remote` property even if it isn't a `BaseEvent`.

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -50,38 +70,44 @@ def __init__(self,
...
+    async def handle_native_peer_message(self,
+                                         remote: Node,
+                                         cmd: Command,
+                                         msg: PayloadType) -> None:
+        """
+        Process every native peer message. Subclasses should overwrite this to forward specific
+        peer messages on the event bus. The handler is called for every message that is defined in
+        ``self.subscription_msg_types``.
+        """
+        pass
```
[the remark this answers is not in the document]
cburgdorf: I can make this an `abstractmethod` :+1:

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -94,5 +120,32 @@ def get_peer(self, remote: Node) -> TPeer:
             else:
                 return cast(TPeer, peer)
 
+    async def accept_connect_commands(self) -> None:
```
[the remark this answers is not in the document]
cburgdorf: It's already called `accept_connect_commands` in `master` but I agree that the naming doesn't conform to the others. Will change :+1:

trinity/protocol/common/events.py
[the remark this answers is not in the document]
carver: Maybe `HasRemoteEvent` turns into an abstract class with an abstract property `remote`?

pipermerriam:
(sorry for the too-soon review.  I just saw your request to hold off)

trinity/protocol/common/events.py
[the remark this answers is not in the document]
cburgdorf: Unfortunately `mypy` goes absolutely crazy with abstract properties. Even with the suggest solution of using `type: ignore` I couldn't get it to work. https://github.com/python/mypy/issues/1362

So, I've gone back to the non-abstract version and just properly using `super().__init__(remote)` everywhere.

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -50,38 +70,44 @@ def __init__(self,
...
         self.run_daemon_task(self.handle_peer_count_requests())
         self.run_daemon_task(self.accept_connect_commands())
-        self.run_daemon_task(self.handle_disconnect_peer_events())
-
-        await self.cancel_token.wait()
+        self.run_daemon_task(self.handle_native_peer_messages())
+
+        await self.cancellation()
+
+    def run_event_daemon(self,
```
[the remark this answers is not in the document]
cburgdorf: SGTM :+1:

tests/core/json-rpc/test_ipc.py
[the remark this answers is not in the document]
cburgdorf: I was able to resolve all sleeps except this one with the existing `wait_for` APIs that are already in lahja/master.

For this one, we can not wait for a specific event but would rather like for *any* event subscription to be propagated. This would be possible with the upcoming `wait_for_x` APIs proposed in https://github.com/ethereum/lahja/pull/74 but I would like to base this PR on top of lahja/master to land it sooner.

So I hope we can live with this one single sleep until we have figured out the best API to wait on subscription propagations.

tests/core/integration_test_helpers.py
```diff
@@ -174,14 +180,30 @@ def load_fixture_db(db_fixture, db_class=LevelDB):
...
-    await peer_pool_event_bus_request_handler.events.started.wait()
+    asyncio.ensure_future(event_server.run())
+    # Give event subscriptions a moment to propagate
+    #await asyncio.sleep(0.01)
+    #await event_bus.wait_until_any_connection_subscribed_to(BaseEvent)
+    await event_server.events.started.wait()
+    return event_server
+
+
+async def run_request_server(event_bus, chaindb, server_type=None):
```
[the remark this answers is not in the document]
cburgdorf: Sounds good in general. I just wonder how to deal with the rightward drift when you need multiple of them e.g.

```python:
async with run_event_server():
    async with run_request_server():
        async with run_foo():
            pass
```

It would be nice to be able to say:

```python:

async with (
    run_event_server(),
    run_request_server(),
    run_foo()
):
    pass
```

setup.py
```diff
@@ -29,7 +29,8 @@
         "plyvel==1.0.5",
         PYEVM_DEPENDENCY,
         "web3==4.4.1",
-        "lahja==0.12.0",
+        "lahja@git+https://github.com/ethereum/lahja.git@22396a8fe79809357438e917e66a7101e6e3ac01",  # noqa: E501
```
cburgdorf: @carver @pipermerriam I wonder what's the best thing to do here. Ideally I'd like to get this change in sooner than later because it touches many files and especially the eth2 code changes on a such rapid pace (rightfully so :clap: ) that I have to resolve conflicts with every rebase.

So I *could* rebase lahja from master. However, I feel that it would just be a matter of some more days until we want to cut a new release soon after. So, if you agree that this PR could land soon, do you think it would be acceptable to keep lahja pinned to this git ref from current lahja/master just for some days?

trinity/protocol/common/events.py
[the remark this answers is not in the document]
carver: Hm, we do this in a number of places. What's the difference here?
https://github.com/ethereum/trinity/blob/675c583bccac8ecfbdfc0ca33be5c40d56f1fd27/trinity/extensibility/plugin.py#L80-L83

setup.py
carver: Yeah, I think it's okay temporarily. As long as getting lahja released with the needed changes becomes top priority :). Don't want it to hold up the next trinity release.

tests/core/integration_test_helpers.py
```diff
@@ -174,14 +178,28 @@ def load_fixture_db(db_fixture, db_class=LevelDB):
...
+    event_server = handler_type(
         event_bus,
         peer_pool,
         peer_pool.cancel_token
     )
-    asyncio.ensure_future(peer_pool_event_bus_request_handler.run())
-    await peer_pool_event_bus_request_handler.events.started.wait()
+    asyncio.ensure_future(event_server.run())
+
+    await event_server.events.started.wait()
```
carver: Note to self: These two lines are so common that we probably want a one-liner, like `await event_server.launch()` that starts the `BaseService` in the background, but doesn't return until it's initialized. Piper's got bigger ideas for the service refactor, so we'll probably delay that syntactic sugar until after a refactor.

tests/core/integration_test_helpers.py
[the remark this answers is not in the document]
carver: Looking for something like this? https://stackoverflow.com/a/1073814/8412986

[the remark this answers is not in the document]
carver: Planning to convert this to the `asynccontextmanager` in this PR?

tests/core/json-rpc/test_ipc.py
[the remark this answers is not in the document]
carver: Maybe add a comment with the proposed replacement code, to be used when the new API is merged/released?

tests/integration/test_trinity_cli.py
```diff
@@ -194,7 +194,7 @@
 async def test_shutdown(command, async_process_runner):
 
     async def run_then_shutdown_and_yield_output():
-        await async_process_runner.run(command, timeout_sec=30)
+        await async_process_runner.run(command, timeout_sec=40)
```
carver: Was this local tweaking, or is it intended to be merged? Could use some more commentary about why 40 seems necessary.

trinity/plugins/builtin/request_server/plugin.py
```diff
@@ -0,0 +1,116 @@
...
+                trinity_config.get_app_config(Eth1AppConfig),
+                db_manager,
+            )
+        elif trinity_config.has_app_config(BeaconAppConfig):
+            server = self.make_beacon_request_server(
+                trinity_config.get_app_config(BeaconAppConfig),
+                db_manager
+            )
+        else:
+            raise Exception("Invariant")
```
carver: Just a bit more info here would be good, like: "Trinity config must have either eth1 or beacon chain config"

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -1,56 +1,131 @@
...
         self.peer_pool = peer_pool
         self.event_bus = event_bus
 
-    async def accept_connect_commands(self) -> None:
+    async def _run(self) -> None:
+        self.logger.debug("Running %s", self.__class__.__name__)
+
+        self.run_daemon_event(
+            DisconnectPeerEvent,
+            lambda peer, event: peer.disconnect(event.reason)
```
carver: This is a regular `peer` right? (meaning `peer.disconnect()` is an awaitable)

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -62,37 +137,29 @@ def __init__(self,
                 req.broadcast_config()
             )
 
-    async def handle_disconnect_peer_events(self) -> None:
-        async for command in self.wait_iter(self.event_bus.stream(DisconnectPeerEvent)):
+    async def handle_event(self,
```
carver: This name confused me on the first readthrough (because it sounds like it's handling a single event once).

Maybe:
- `event_handler`
- `event_responder`
- `event_daemon`
- `handle_stream`
- `event_callback`

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -62,37 +137,29 @@ def __init__(self,
...
+                           event_type: Type[TStreamEvent],
+                           event_handler_fn: Callable[[TPeer, TStreamEvent], Any]) -> None:
+
+        async for ev in self.wait_iter(self.event_bus.stream(event_type)):
```
carver: `ev` -> `event`?

trinity/protocol/common/servers.py
```diff
@@ -77,6 +88,57 @@ def __init__(
...
+        self._subscribed_events = subscribed_events
+
+    async def _run(self) -> None:
+
+        for event_type in self._subscribed_events:
+            self.run_daemon_task(self.handle_event(event_type))
+
+        await self.cancellation()
+
+    async def handle_event(self, event_type: Type[PeerPoolMessageEvent]) -> None:
```
carver: same name surprise here

trinity/protocol/common/servers.py
```diff
@@ -77,6 +88,57 @@ def __init__(
...
+    async def _quiet_handle_msg(
+            self,
+            remote: Node,
+            cmd: protocol.Command,
+            msg: protocol._DecodedMsgType) -> None:
+        try:
+            await self._handle_msg(remote, cmd, msg)
+        except OperationCancelled:
+            # Silently swallow OperationCancelled exceptions because otherwise they'll be caught
+            # by the except below and treated as unexpected.
```
carver: Maybe add a debug log just so we notice if this is happening unexpectedly.

trinity/protocol/eth/events.py
```diff
@@ -0,0 +1,88 @@
...
+
+class GetReceiptsEvent(PeerPoolMessageEvent):
+    """
+    Event to carry a ``GetReceipts`` command from the peer pool to any process that
+    subscribes the event through the event bus.
+    """
+    pass
+
+
+class GetNodeDataEvent(PeerPoolMessageEvent):
```
carver: cc to make sure @lithp notices that the firehose request server will probably need some changes after this merges. (Like adding a new event)

trinity/protocol/eth/peer.py
```diff
@@ -97,9 +123,64 @@ def handle_sub_proto_msg(self, cmd: Command, msg: _DecodedMsgType) -> None:
             )
 
 
+class ETHProxyPeer:
+    """
+    A ``ETHPeer`` that can be used from any process as a drop-in replacement for the actual
+    peer that sits in the peer pool. Any action performed on the ``ETHProxyPeer`` is delegated
+    to the actual peer in the pool.
```
carver: Is this comment correct? There are a lot of methods on a peer. What about `disconnect`, for example? I only see `sub_proto` implemented.

trinity/protocol/les/servers.py
```diff
@@ -39,28 +50,34 @@ class LESPeerRequestHandler(BasePeerRequestHandler):
...
+
+        self.logger.debug2("Peer %s requested %s", remote, cmd)
+        peer = LESProxyPeer.from_node(remote, self.event_bus, self.broadcast_config)
         if isinstance(cmd, commands.GetBlockHeaders):
-            block_request_kwargs = cast(Dict[str, Any], msg)
-            await self._handler.handle_get_block_headers(peer, block_request_kwargs)
+            await self._handler.handle_get_block_headers(peer, cast(Dict[str, Any], msg))
         else:
-            self.logger.debug("%s msg from %s not implemented", cmd, peer)
+            self.logger.debug("%s msg not handled yet, needs to be implemented", cmd)
```
carver: The peer info could theoretically be interesting (here and the eth server). If some peer is requesting unexpected/non-standard commands, it might be nice to identify the type of client that's running by scanning the logs for the peer hash and its related corresponding client string.

trinity/protocol/eth/peer.py
cburgdorf: That comment is a bit ahead of time haha. In the follow up PR that also touches the sync, this becomes a proper `BaseService` derived class and I also had a `typing_extensions.Protocol` `ETHPeerLike` that abstracted away the usage of the actual peer or the proxy peer.

I will adjust the comment for this PR so that it properly reflects the status quo.

tests/integration/test_trinity_cli.py
cburgdorf: Yep, will add a comment. The reason is that this test first boots the node until it starts syncing, and then shuts it down and tries to follow all the output until it is fully shut down. So it naturally needs more time compared to the rest of these tests.

trinity/protocol/common/peer_pool_event_bus.py
cburgdorf: Good catch!

trinity/protocol/les/servers.py
cburgdorf: Notice that the only reason we could run into this, is when we mistakenly add a subscription for an event type that we then forget to handle. In other words, if we see something here, it is always our own misconfiguration.

tests/core/integration_test_helpers.py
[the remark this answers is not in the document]
cburgdorf: Woah, I had no idea this was actually possible! Thanks @carver for bringing it up in that other comment.

This works just fine.
```python
import asyncio
from contextlib import asynccontextmanager

@asynccontextmanager
async def foo():
    try:
        print("foo")
        yield 1
    finally:
        print("cleaning foo")

@asynccontextmanager
async def bar():
    try:
        print("bar")
        yield 2
    finally:
        print("cleaning bar")

async def meh():
    async with foo(), bar():
        print("wohoo")

asyncio.run(meh())
```

trinity/protocol/common/peer_pool_event_bus.py
cburgdorf: I like `handle_stream` :+1:

trinity/protocol/common/servers.py
cburgdorf: Wait? When would that be unexpected and how would I recognize that? Also notice, that this code did not diverge from master.

https://github.com/ethereum/trinity/blob/675c583bccac8ecfbdfc0ca33be5c40d56f1fd27/trinity/protocol/common/servers.py#L58-L70

tests/core/json-rpc/test_ipc.py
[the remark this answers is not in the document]
cburgdorf: The syntax isn't yet entirely finalized but I promise to come back to this and update it as soon as it lands!

trinity/protocol/common/events.py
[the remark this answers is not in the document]
cburgdorf: Oh yeah, I was doing something stupid. Updated it. It does feel a bit boilerplatey but could be worse...

cburgdorf:
Ok, still missing some `async with`...on it

trinity/protocol/eth/events.py
lithp: Thanks for the ping! I think what you're saying is that I should race to merge Firehose before this PR lands ;)

cburgdorf: Hahaha! This PR is like a black whole. Eventually it sucks in the entire code base...at this point my whole job will be about rebasing this PR until eternity :sweat_smile:

trinity/protocol/common/servers.py
carver: Yeah, I guess it's fine for now. It will be handled differently after switching to `trio` eventually, so we can handle it then. (To answer your question, we wouldn't recognize that anything was wrong in the code. We might recognize it by browsing the logs, and noticing an unusual number or location of the logs.)

tests/core/p2p-proto/test_server.py
```diff
@@ -173,21 +169,21 @@ def get_server(privkey, address, event_bus):
...
-        TO_NETWORKING_BROADCAST_CONFIG
-    )
+        await event_bus.wait_until_any_connection_subscribed_to(ConnectToNodeCommand)
+        await event_bus.broadcast(
+            ConnectToNodeCommand(RECEIVER_REMOTE),
+            TO_NETWORKING_BROADCAST_CONFIG
+        )
 
-    await asyncio.sleep(0.5)
+        await asyncio.sleep(0.5)
```
carver: I know this wasn't you, but wow 0.5s is kind of a long time for a single test. Can we get away with trimming this down?

cburgdorf:
>Looks like you are debugging something, but it seems good to go. Feel free to re-ping me if the solution you end up with needs review, too.

I think, the test that scans the shutdown for errors (which is currently a `xfail`) got more flaky because, somewhat ironically, the shutdown became more stable with the upgrade to a newer lahja release.

I marked the test as `skip` for now just because I really want to get this in now and it is 1 AM here. I swear it's the first thing I pickup tomorrow morning to resolve it properly :pray:

tests/core/p2p-proto/test_server.py
cburgdorf: This is going to be the second thing I try tomorrow morning :sweat_smile:

tests/core/json-rpc/test_ipc.py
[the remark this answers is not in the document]
pipermerriam: Tracking issue?

tests/core/p2p-proto/test_server.py
pipermerriam: Reminder about shortening this wait if we can.

trinity/protocol/eth/events.py
```diff
@@ -0,0 +1,104 @@
...
+class SendNodeDataEvent(HasRemoteEvent):
+    """
+    Event to proxy a ``ETHPeer.sub_proto.send_node_data`` call from a proxy peer to the actual
+    peer that sits in the peer pool.
+    """
+    def __init__(self, remote: Node, nodes: Tuple[bytes, ...]) -> None:
+        self._remote = remote
+        self.nodes = nodes
+
+    @property
```
pipermerriam: I *think* that you can define these classes without inheriting from `HasRemoteEvent` and set this property normally and still get away with telling `mypy` that something takes a `TypeVar('Something', bound=HasRemoteEvent)` and it will still *just work*)

trinity/protocol/eth/peer.py
```diff
@@ -97,9 +125,80 @@ def handle_sub_proto_msg(self, cmd: Command, msg: _DecodedMsgType) -> None:
...
+                                         cmd: Command,
+                                         msg: PayloadType) -> None:
+
+        ignored_commands = (
+            Transactions,
+            NewBlockHashes,
+        )
+
+        if isinstance(cmd, ignored_commands):
+            pass
```
pipermerriam: Maybe a log message at debug level so this doesn't get completely lost?

tests/core/p2p-proto/test_server.py
carver: https://github.com/ethereum/trinity/pull/656

trinity/protocol/eth/events.py
cburgdorf: Only if `HasRemoteEvent` were a `typing_extensions.Protocol` so that *structural typing*  is used.  But that would allow *any kind of thing* with a `remote` property to be used here and not enforce proper `BaseEvent` checking. Unfortunately mypy doesn't allow multiple bounds ([yet?](https://github.com/python/typing/issues/213#issuecomment-479039371)) so that we can not do something like `TypeVar('something', bound=[Intersection[BaseEvent, SupportsRemote]])`

trinity/protocol/eth/peer.py
cburgdorf: Notice that this code just moved, we've had this construct before. But I will send a follow up PR to just forward these two events because we will need that anyway when we start moving the transaction pool into a `BaseIsolatedPlugin` as well.

setup.py
lithp: Don't know what the plan is here but in case this got dropped a friendly reminder that master is still using this git revision.

PR #618, opened 2019-05-16
Connect to unsynced peers, and trinity peers

carver:
Trinity peers were not communicating DAO headers during handshake, so trinity would not connect to another trinity node. (lol, am I the first one to try this?)

Why? I think it's because all incoming requests were buffered, including the request for the DAO headers. So they both asked each other for the DAO headers and never answered because they were waiting for the other to respond.

Current solution is to immediately add peer to pool during boot, and then kick it back out if it fails the DAO check. At least, that's the solution when I connected my two trinity peers together.

Bonus: We were previously failing any peer who couldn't answer with the DAO
fork header. That means that unsynced peers would be rejected. This modifies the peer boot to accept unsynced peers for now.

### To-Do

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/common/boot.py
+++ b/trinity/protocol/common/boot.py
@@ -62,9 +62,11 @@ class DAOCheckBootManager(BasePeerBootManager):
                 ) from err
 
             if len(headers) != 2:
-                raise DAOForkCheckFailure(
-                    f"{self.peer} failed to return DAO fork check headers"
+                self.logger.debug(
+                    f"{self.peer} returned only {headers!r} at DAO fork. "
+                    "Assuming peer is syncing, for now..."
                 )
+                return
             else:
                 parent, header = headers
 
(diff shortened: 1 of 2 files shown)


trinity/protocol/common/boot.py
```diff
@@ -62,9 +62,11 @@ class DAOCheckBootManager(BasePeerBootManager):
                 ) from err
 
             if len(headers) != 2:
-                raise DAOForkCheckFailure(
-                    f"{self.peer} failed to return DAO fork check headers"
+                self.logger.debug(
+                    f"{self.peer} returned only {headers!r} at DAO fork. "
+                    "Assuming peer is syncing, for now..."
```
pipermerriam: I'd advocate for this to fall back to request the peer's announced `HEAD` block and to check whether the returned block number is less than the DAO headers.  It seems we should be able to actually verify whether a peer responded because they don't have the headers.

carver: How about something like 7ac9344 ?

PR #619, opened 2019-05-16
Header skipping unification

carver:
### What was wrong?

Two issues:
1. The header skipping log was *very* noisy, spitting out a line for every skipped header
2. The full syncer was using the `skip_complete_headers` utility, but not header sync, so fixing ^ didn't apply everywhere

### How was it fixed?

1. Combine the skipped headers, then log a single line about all skipped headers
2. Use `skip_complete_headers()` in the header syncer

#### Cute Animal Picture

### DIFF
--- a/trinity/_utils/headers.py
+++ b/trinity/_utils/headers.py
@@ -75,13 +75,28 @@ async def _skip_complete_headers_iterator(
     the remaining headers.
     """
     iter_headers = iter(headers)
+    # for logging:
+    first_discarded = None
+    last_discarded = None
+    num_discarded = 0
     for header in iter_headers:
         is_present = await completion_check(header)
         if is_present:
-            logger.debug("Discarding header that we already have: %s", header)
+            if first_discarded is None:
+                first_discarded = header
+            else:
+                last_discarded = header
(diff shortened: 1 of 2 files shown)


trinity/_utils/headers.py
```diff
@@ -75,13 +75,28 @@ def sequence_builder(start_number: T,
...
+            if first_discarded is None:
+                first_discarded = header
+            else:
+                last_discarded = header
+            num_discarded += 1
         else:
             yield header
             break
 
+    logger.debug(
```
pipermerriam: I've started viewing APIs that pass around loggers as code smell.  Similarly, this utility function being in charge of the log message also feels a little off.  I'm like -0.25 on this.

Thoughts on adjusting the API to instead return two tuples, `(kept_headers, discarded_headers)`.  Then the call-site can be in charge of printing the message.  You could then use [`trinity._utils.humanize.humanize_integer_sequence`](https://github.com/ethereum/trinity/blob/432cfd2019cad2c925f74e5a3834d31976b390f0/trinity/_utils/humanize.py#L38) to display the header numbers in a single line in their most concise representation.

PR #625, opened 2019-05-17
Candidate backend API now provides all connected remotes

pipermerriam:
### What was wrong?

The peer backend API was only getting a count of currently connected peers.  This doesn't allow the backends to exclude peers that we are already connected to.

### How was it fixed?

The request now comes with a full list of the remotes that we are currently connected to.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

alpaca04

### DIFF
--- a/p2p/peer_backend.py
+++ b/p2p/peer_backend.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 from typing import (
-    Iterable,
+    Set,
+    Tuple,
 )
 
 from lahja import (
@@ -24,7 +25,7 @@ class BasePeerBackend(ABC):
     @abstractmethod
     async def get_peer_candidates(self,
                                   num_requested: int,
-                                  num_connected_peers: int) -> Iterable[Node]:
+                                  connected_remotes: Set[Node]) -> Tuple[Node, ...]:
         pass
 
 
@@ -37,12 +38,16 @@ class DiscoveryPeerBackend(BasePeerBackend):
 
     async def get_peer_candidates(self,
                                   num_requested: int,
-                                  num_connected_peers: int) -> Iterable[Node]:
+                                  connected_remotes: Set[Node]) -> Tuple[Node, ...]:
         response = await self.event_bus.request(
             PeerCandidatesRequest(num_requested),
             TO_DISCOVERY_BROADCAST_CONFIG,
         )
-        return response.candidates
+        return tuple(
+            candidate
+            for candidate in response.candidates
+            if candidate not in connected_remotes
+        )
 
 
 class BootnodesPeerBackend(BasePeerBackend):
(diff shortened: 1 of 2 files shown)


p2p/peer_backend.py
```diff
@@ -37,12 +38,16 @@ def __init__(self, event_bus: Endpoint) -> None:
 
     async def get_peer_candidates(self,
                                   num_requested: int,
-                                  num_connected_peers: int) -> Iterable[Node]:
+                                  connected_remotes: Set[Node]) -> Iterable[Node]:
```
cburgdorf: nitpick: I think the return type should be `Tuple[Node, ...]` here which would also be more in line with our general [recommendation on function variance](https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#function-variance): Use most abstract type for *parameters*, use *concrete type* for return value. There's little to gain from *hiding*  a tuple behind a `Iterable[Node]` if what we actually return is `Tuple[Node, ...]`. A method that expects an `Iterable[Node]` such as:

```python
def do_something_with_nodes(nodes: Iterable[Node]) -> None:
    ...
```

is satisfied with `do_something_with_nodes(get_peer_candidates())` just fine if it returns a `Tuple[Node, ...] ` because that implicitly qualifies as `Iterable[Node]`.

And what's more, if you annotate `get_peer_candiadtes(...)` with `Iterable[Node]` then mypy will yell at you if you try to do `len(get_peer_candidates(...))` because it *thinks*  it's an `Iterable[Node]` when in reality it is a `Tuple[Node]` and does support `len(...)` just fine.

pipermerriam: I think the `Iterable[Node]` is a leftover from when this was a singular API that was consumed lazily.  Changing.

p2p/peer_backend.py
```diff
@@ -51,13 +56,17 @@ def __init__(self, event_bus: Endpoint) -> None:
...
             response = await self.event_bus.request(
                 RandomBootnodeRequest(),
                 TO_DISCOVERY_BROADCAST_CONFIG
             )
 
-            return response.candidates
+            return tuple(
+                candidate
+                for candidate in response.candidates
+                if candidate not in connected_remotes
```
carver: Why implement this in every `BasePeerBackend` instead of implement it once at the call site? It looks like we always want the filtering to happen.

pipermerriam: Compelling reason is here:

https://github.com/ethereum/trinity/pull/543/files#diff-f20890a2a306bb28e0b1315efc298f2eR305

But general reason is that any backend that *can* do a pre-filter on a set of potential candidates larger than the number requested needs to be able to do so, especially if the number of potential candidates is large like it can/will be when we track them in a database.

PR #627, opened 2019-05-17
remove async from blacklist tracker methods

pipermerriam:
### What was wrong?

When we disconnect from a peer we *may* want to also blacklist them.  The `tracker.record_blacklist` method is `async` which means that it can only be used from `async` methods.

When we are within something like the `BootManager` or `RequestManager` for a peer and we want to disconnect we **must** use `Peer.disconnect_nowait` because using `await Peer.disconnect` throws a cancellation error because of the inner call to `Peer.cancel()`.

This means we **must** be able to issue a blacklist from a sync method.

### How was it fixed?

Changed to using `broadcast_nowait` and removed the `async` nature of the methods on the blacklist tracker.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

12-06-19 Sugar (9)C750

### DIFF
--- a/tests/core/network-db/test_connection_tracker_server.py
+++ b/tests/core/network-db/test_connection_tracker_server.py
@@ -19,7 +19,7 @@ from trinity.plugins.builtin.network_db.connection.tracker import (
 async def test_connection_tracker_server_and_client(event_loop, event_bus):
     tracker = MemoryConnectionTracker()
     remote_a = NodeFactory()
-    await tracker.record_blacklist(remote_a, 60, "testing")
+    tracker.record_blacklist(remote_a, 60, "testing")
 
     assert await tracker.should_connect_to(remote_a) is False
 
@@ -40,7 +40,9 @@ async def test_connection_tracker_server_and_client(event_loop, event_bus):
 
     assert await bus_tracker.should_connect_to(remote_b) is True
 
-    await bus_tracker.record_blacklist(remote_b, 60, "testing")
+    bus_tracker.record_blacklist(remote_b, 60, "testing")
+    # let the underlying broadcast_nowait execute
+    await asyncio.sleep(0.01)
 
     assert await bus_tracker.should_connect_to(remote_b) is False
     assert await tracker.should_connect_to(remote_b) is False
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -253,7 +253,7 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
             await peer.disconnect(DisconnectReason.timeout)
             return
         except HandshakeFailure as err:
-            await self.connection_tracker.record_failure(peer.remote, err)
+            self.connection_tracker.record_failure(peer.remote, err)
             raise
         else:
             if peer.is_operational:
@@ -348,7 +348,7 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
             raise
         except HandshakeFailure as e:
             self.logger.debug("Could not complete handshake with %r: %s", remote, repr(e))
-            await self.connection_tracker.record_failure(remote, e)
+            self.connection_tracker.record_failure(remote, e)
             raise
         except COMMON_PEER_CONNECTION_EXCEPTIONS as e:
             self.logger.debug("Could not complete handshake with %r: %s", remote, repr(e))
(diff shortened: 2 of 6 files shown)


tests/core/network-db/test_connection_tracker_server.py
```diff
@@ -40,7 +40,9 @@
 
     assert await bus_tracker.should_connect_to(remote_b) is True
 
-    await bus_tracker.record_blacklist(remote_b, 60, "testing")
+    bus_tracker.record_blacklist(remote_b, 60, "testing")
+    # let the underlying broadcast_nowait execute
+    await asyncio.sleep(0.01)
```
lithp: I'm trying to reason through why this doesn't look great to me:

It seems easy to forget this `sleep()` call. The beauty of `await` is that you can write code which by and large assumes everything is single-threaded and your code still works even when things are happening concurrently. These regular methods which schedule operations to happen in the future break that abstraction and now we're back to the world of reasoning about threads. 

Maybe it's okay in this situation because few things (nothing?) but the tests will assume that blacklisted nodes are immediately blacklisted. Also, if we called it `record_blacklist_nowait` that would make the scheduling aspect explicit so that it's harder to assume that the ban is immediately recorded, and probably that's enough?

lithp:
This async/sync thing is frustrating. I don't fully understand the problem here, you're saying that whenever we call `Peer.cancel()` a `CancellationException` might be thrown? And this is important because we sometimes call `Peer.disconnect{,_nowait}` from contexts where killing a peer shouldn't result in cancelling the calling service?

If you want to move on to working on something else I'm okay with approving this! But what about one of these solutions instead:

- I'm pretty sure I'm understanding wrong, but if I got it right `Peer.disconnect()` could swallow the `CancellationException`, since throwing it sounds like behavior you'd rarely if ever want. It looks like all the calls to `disconnect_nowait` are from inside async contexts, so somehow fixing this would allow simplifying the interface by dropping `disconnect_nowait` entirely.

- Maybe `disconnect()` can not call `record_blacklist`, and leave calling it to the callers? This would add some code but `disconnect(DisconnectReason.bad_protocol)` is only called in 5 places, each of those call sites could call `record_blacklist` themselves.

carver:
> * Maybe `disconnect()` can not call `record_blacklist`, and leave calling it to the callers? This would add some code but `disconnect(DisconnectReason.bad_protocol)` is only called in 5 places, each of those call sites could call `record_blacklist` themselves.

Hah, that's how it started a little while ago. My concern was that it's too easy for callers to forget to blacklist (IIRC it already happened once in the PR). Plus it forces the caller to make some decisions about how long to blacklist, etc. So I'm not a fan of that particular direction.

pipermerriam:
I think part of the complexity here is the shortfalls of the `CancelToken` API.  I'd consider the current *dirtyness* of the API a direct result of that and something that is likely to be able to be cleaned up significantly under the `trio` model where we get better control and more powerful tools to deal with cancellations.

As it stands, I tried getting rig of `disconnect_nowait` and the result was being unable to cleanly do an asynchronous disconnection from within the places where we currently use `disconnect_nowait` due to the cascading cancellation causing the call-site to blow up from being cancelled...

carver:
Since neither of you seem super keen on joining blacklist into the `peer.disconnect()`, what about the reverse? So blacklisting a peer would automatically disconnect them, if they are currently connected. That seems relatively reasonable: when would you want to blacklist a peer and stay connected to them?

pipermerriam:
@carver https://github.com/ethereum/trinity/issues/662

pipermerriam:
doh, i keep doing this.  merging when I *think* someone actually clicked "approve" but quickly realizing they were all just comments.... happy to cleanup anything post-merge or revert if anyone has comments they would like addressed.

lithp:
> As it stands, I tried getting rig of disconnect_nowait and the result was being unable to cleanly do an asynchronous disconnection from within the places where we currently use disconnect_nowait due to the cascading cancellation causing the call-site to blow up from being cancelled...

I'm still not sure I understand the problem, I guess I should play with this code, it sounds like a bug that calling disconnect would blow up the caller. That sounds like the root problem here!

> Since neither of you seem super keen on joining blacklist into the peer.disconnect(), what about the reverse? So blacklisting a peer would automatically disconnect them, if they are currently connected. That seems relatively reasonable: when would you want to blacklist a peer and stay connected to them?

This sounds like it's worth a shot, but since this PR has been merged and it's not a huge deal that this method calls async from inside sync maybe this is something worth opening a ticket for and coming back to later.

pipermerriam:
@lithp switch any of the `disconnect_nowait` calls to be `await disconnect(` and run your client for a bit.  You might save some time just by putting an `await disconnect` somewhere in `common.managers` where it'll get hit reliably.

PR #628, opened 2019-05-17
refactor blacklist tracking API to allow explicit expiration

pipermerriam:
### What was wrong?

The business logic for adjusting the total timeout based on how many times a peer has errored out was at the wrong level, not allowing us to do explicit setting of the expiration via the core API.  This made testing harder in #543

### How was it fixed?

Moved the business logic up a level into the `record_blacklist` method.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

16-05-28 Honeysuckle Rose (7)C1000

### DIFF
--- a/trinity/plugins/builtin/network_db/connection/tracker.py
+++ b/trinity/plugins/builtin/network_db/connection/tracker.py
@@ -49,6 +49,15 @@ class BlacklistRecord(Base):
     error_count = Column(Integer, default=1, nullable=False)
 
 
+def adjust_repeat_offender_timeout(base_timeout: float, error_count: int) -> datetime.datetime:
+    """
+    sub-linear scaling based on number of errors recorded against the offender.
+    """
+    adjusted_timeout_seconds = int(base_timeout * math.sqrt(error_count + 1))
+    delta = datetime.timedelta(seconds=adjusted_timeout_seconds)
+    return datetime.datetime.utcnow() + delta
+
+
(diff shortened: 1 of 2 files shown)


trinity/plugins/builtin/network_db/connection/tracker.py
```diff
@@ -57,10 +57,16 @@ def __init__(self, session: BaseSession):
...
+        try:
+            record = self._get_record(remote.uri())
+        except NoResultFound:
+            expires_at = datetime.datetime.utcnow() + datetime.timedelta(seconds=timeout_seconds)
+            self._create_record(remote, expires_at, reason)
         else:
-            self._create_record(remote, timeout_seconds, reason)
+            adjusted_timeout_seconds = int(timeout_seconds * math.sqrt(record.error_count + 1))
+            delta = datetime.timedelta(seconds=adjusted_timeout_seconds)
+            scaled_expires_at = datetime.datetime.utcnow() + delta
```
lithp: nitpick: I think I would pull these three lines into a helper, maybe something like `compute_expires_at(timeout_seconds, record)`, then it's easier to glance at this function and know it's an upsert, and it's easier to test the upserting part and the expiration computation independently.

lithp:
Not part of this PR but I think that for some errors even super-linear makes sense. If the network is wrong the network is probably always going to be wrong for this ip, and it's definitely always going to be wrong for this public key, probably some kind of exponential backoff is appropriate.

PR #629, opened 2019-05-17
Blacklist peers who disconnect too quickly

pipermerriam:
### What was wrong?

While working on #543 I ran into a problem.  When looking at a set of historical peer connections we choose peers that are not currently blacklisted, and then order them by how recently we successfully connected to them.

Under this rule, a peer could allow our connection for a short duration, and then disconnect, claiming they are shutting down.  They would still be very close to the front of the queue when we query our database, which would result in us attempting to connect to them again.  They could repeat this for a very long time, constantly disrupting our sync but always being very close to the front of the line for peers we choose to connect to.

### How was it fixed?

If a peer disconnects from us after too short a duration (currently set to 60 seconds but I think 2-5 minutes might be better), we put them on the blacklist for some duration, allowing other peers to make their way to the front of the line.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

13-09-28 Breeze Boy Cria (34)C750

### DIFF
--- a/p2p/constants.py
+++ b/p2p/constants.py
@@ -133,3 +133,8 @@ BLACKLIST_SECONDS_BAD_PROTOCOL = 60 * 10  # 10 minutes
 
 # Amount of time a peer will be blacklisted when they timeout too frequently
 BLACKLIST_SECONDS_TOO_MANY_TIMEOUTS = 60 * 5  # 5 minutes
+
+# Both the amount of time that we consider to be a peer disconnecting from us
+# too quickly as well as the amount of time they will be blacklisted for doing
+# so.
+BLACKLIST_SECONDS_QUICK_DISCONNECT = 60
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -67,6 +67,7 @@ from p2p.tracking.connection import (
 
 from .constants import (
     BLACKLIST_SECONDS_BAD_PROTOCOL,
+    BLACKLIST_SECONDS_QUICK_DISCONNECT,
     SNAPPY_PROTOCOL_VERSION,
 )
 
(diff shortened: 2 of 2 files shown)


p2p/constants.py
```diff
@@ -133,3 +133,8 @@
 
 # Amount of time a peer will be blacklisted when they timeout too frequently
 BLACKLIST_SECONDS_TOO_MANY_TIMEOUTS = 60 * 5  # 5 minutes
+
+# Both the amount of time that we consider to be a peer disconnecting from us
+# too quickly as well as the amount of time they will be blacklisted for doing
+# so.
+BLACKLIST_SECONDS_QUICK_DISCONNECT = 60
```
cburgdorf: Just pointing out that quick connecting / disconnecting is also quite rational behavior under development. But since the penalty isn't that hard I guess it is ok. I guess if you develop explicitly against a Trinity node you'll just want to start the other node with `--network-tracking-backend do-not-track` anyway. And if you develop and just randomly hit a Trinity node that blacklists you for this behavior then it's just a 60s penalty so not the end of the world.

Just wanted to bring it up.

pipermerriam: Good point.  Probably worth putting in development docs at some point in a `tips-and-tricks` section as well as actually writing up some details on the blacklisting logic in another section so that it's not so opaque

cburgdorf: Yeah...docs :sweat_smile: I have a tiny little man in the back of my head reminding me that I wanted to do a *docs sprint* some time soon™ again.

pipermerriam: Good workshop to organize at the summit.  If you had a list of areas/topics that were good targets we could breakout into groups and write docs for an hour one day.

pipermerriam:
murrr, merged before realizing it was just a comment review.  @carver ping me if you have anything you want cleaned up or if you think this should be rolled back.

carver:
Heh, np with it. I think it should probably eventually be replaced with a selection algorithm that takes into account how long the average session with a peer is. This solution can just be gamed by waiting 61s, and in general preferring longer-lasting peers seems valuable. Figuring out the right tradeoff between recency and longevity seems subjective, of course.

PR #631, opened 2019-05-18
Avoid logging in skip_complete_headers() utility

carver:
Suggested in https://github.com/ethereum/trinity/pull/619#discussion_r285304221

By changing the result from "return incomplete headers" to "return tuple of completed and tuple of incompleted headers". Added some tests as a gut-check.

#### Cute Animal Picture

### DIFF
--- a/trinity/_utils/humanize.py
+++ b/trinity/_utils/humanize.py
@@ -1,4 +1,4 @@
-from typing import Iterable, Sequence, Tuple
+from typing import Iterable, Tuple
 
 from eth_utils.toolz import sliding_window
 
@@ -35,7 +35,7 @@ def _humanize_range(bounds: Tuple[int, int]) -> str:
         return f'{left}-{right}'
 
 
-def humanize_integer_sequence(values: Sequence[int]) -> str:
+def humanize_integer_sequence(values: Iterable[int]) -> str:
     """
     Return a human readable string that attempts to concisely define a sequence
     of integers.
--- a/trinity/_utils/humanize.py
+++ b/trinity/_utils/humanize.py
@@ -35,7 +35,7 @@ def _humanize_range(bounds: Tuple[int, int]) -> str:
         return f'{left}-{right}'
 
 
-def humanize_integer_sequence(values: Iterable[int]) -> str:
+def humanize_integer_sequence(values_iter: Iterable[int]) -> str:
     """
     Return a human readable string that attempts to concisely define a sequence
     of integers.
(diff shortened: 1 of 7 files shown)


trinity/_utils/humanize.py
```diff
@@ -35,7 +35,7 @@ def _humanize_range(bounds: Tuple[int, int]) -> str:
         return f'{left}-{right}'
 
 
-def humanize_integer_sequence(values: Sequence[int]) -> str:
+def humanize_integer_sequence(values: Iterable[int]) -> str:
```
pipermerriam: I'm not 100% sure this api actually works with an iterable (generator?).  The first line `if not values` suggests it wont.

carver: Haha I missed that check, thanks. :man_facepalming:

carver:
Weird, I guess py3.6 lets you do `call_func(*( __empty_generator))` but not py3.7...

trinity/sync/common/headers.py
```diff
@@ -297,7 +297,7 @@ def __init__(self,
             self.logger.debug(
                 "During header sync launch, skipping over (%d) already stored headers %s: %s..%s",
                 len(completed_headers),
-                humanize_integer_sequence(h.block_number for h in completed_headers),
+                humanize_integer_sequence([h.block_number for h in completed_headers]),
```
pipermerriam: *nitpick*

I know it's less concise but I feel like using list comprehensions might encourage other contributors who are less indoctrinated into the immutability camp to use lists elsewhere.  Using `tuple(<generator>)` doesn't seem to have this issue.

carver: :( it's more than just a few extra characters, it means breaking up the line because we would go over the 100 char limit. Also, if you build a list in-line, there's not as much of a problem with it being mutable, since there aren't any handles to the list in the parent scope.

I think I'd rather go back to having `humanize_integer_sequence` accept a generator, and just convert it to a tuple internally, than to add another indentation to this line.

pipermerriam: I'll point out that `black` allows us to stop thinking about these things.  You'd just add the `tuple(...)` and the black would move things around for you.

Admittedly you/we/I could still do some personal thrash trying to get the line written in a way that the black formatted results in a format we like so it's not pain free, but it does allow for a no-hassle option where you don't feel obligated to tweak the line length.

Or you could add a `# noqa: E501` and then go take a shower to wash off the shame....

(do as you wish :smiling_imp: )

carver: So far I think I like the generator-support option best.

pipermerriam: It's like freezing to death.  At first it's really uncomfortable and doesn't feel very nice but eventually a nice calm takes over and you stop feeling cold and drift into a dreamless sleep......

carver: > It's like freezing to death.

Haha, `it` being `black`? Yeah, since Django announced the intention to support black (once it goes stable), I figure that's where we're headed too. I've accepted my fate :P

PR #633, opened 2019-05-18
Add attestation handling to validator and receive server

NIC619:
### How was it fixed?
- `Validator` plugin:
    - [x] add `Validator.this_epoch_assignment`: cache the validators' assignment of this epoch
    - [x] add `Validator.attest`: when slot ticked, validators check if they are supposed to attest during this slot and attest if they are
    - [x] broadcast the attestation
- `ReceiveServer`:
    - [x] add `_handle_attestations`
        - [x] validate the received attestations
        - [x] check if the received attestations are already in database or attestation pool
            - currently make this a stub
        - [x] broadcast the new and valid attestations
- [x] Update p2p proto `Attestation` command and `AttestationMessage`

- TODOs in later PRs:
    - store included-in-block attestations reference into database
    - attestation pool
        - store not yet included-in-block attestations in attestation pool
        - when receiving a valid block, remove the included-in-block attestations from attestation pool 
        - `Validator` request ready-to-be-included attestations from `ReceiveServer` when proposing 
### To-Do

- [ ] Clean up commit history

### DIFF
--- a/eth2/beacon/tools/builder/validator.py
+++ b/eth2/beacon/tools/builder/validator.py
@@ -538,12 +538,11 @@ def create_signed_attestation_at_slot(
         state_machine: BaseBeaconStateMachine,
         attestation_slot: Slot,
         beacon_block_root: Hash32,
-        validator_index: ValidatorIndex,
+        validator_privkeys: Dict[ValidatorIndex, int],
         committee: Tuple[ValidatorIndex, ...],
-        shard: Shard,
-        privkey: int) -> Attestation:
+        shard: Shard) -> Attestation:
     """
-    Create the attestations of the given ``attestation_slot`` slot with ``privkey``.
+    Create the attestations of the given ``attestation_slot`` slot with ``validator_privkeys``.
     """
     state_transition = state_machine.state_transition
     state = state_transition.apply_state_transition_without_block(
@@ -581,14 +580,19 @@ def create_signed_attestation_at_slot(
             signature_domain=SignatureDomain.DOMAIN_ATTESTATION,
             slots_per_epoch=config.SLOTS_PER_EPOCH,
         )
+        for _, privkey in validator_privkeys.items()
     ]
 
+    voting_committee_indices = [
+        CommitteeIndex(committee.index(validator_index))
+        for validator_index in validator_privkeys
+    ]
     # aggregate signatures and construct participant bitfield
     aggregation_bitfield, aggregate_signature = aggregate_votes(
         bitfield=get_empty_bitfield(len(committee)),
         sigs=(),
         voting_sigs=signatures,
-        voting_committee_indices=[CommitteeIndex(committee.index(validator_index))],
+        voting_committee_indices=voting_committee_indices,
     )
 
     # create attestation from attestation_data, particpipant_bitfield, and signature
--- a/eth2/beacon/tools/builder/validator.py
+++ b/eth2/beacon/tools/builder/validator.py
@@ -507,7 +507,6 @@ def create_mock_signed_attestations_at_slot(
     for crosslink_committee in crosslink_committees_at_slot:
         committee, shard = crosslink_committee
 
-        num_voted_attesters = int(len(committee) * voted_attesters_ratio)
         previous_crosslink = state.latest_crosslinks[shard]
 
         attestation_data = AttestationData(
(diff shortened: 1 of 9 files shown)


trinity/plugins/eth2/beacon/validator.py
```diff
@@ -64,17 +78,22 @@ def __init__(
             peer_pool: BCCPeerPool,
             validator_privkeys: Dict[ValidatorIndex, int],
             event_bus: TrinityEventBusEndpoint,
-            genesis_epoch: Epoch,
-            slots_per_epoch: int,
             token: CancelToken = None) -> None:
         super().__init__(token)
         self.chain = chain
         self.peer_pool = peer_pool
```
ChihChengLiang: (Not in this PR) Realized that our validator knows too much. If we are pursuing Beacon Node <--> Validator client separation in the future, the validator client can only access all kind of information through RPC. I'm wondering, for example, peer_pool is the information that the validator client shouldn't know. We are using it in validator client to broadcast blocks and attestations, but these should be done in the server of beacon node.

mhchia: Agree. I think peer pool, receive server, and request server should reside in beacon node. The validator service can be the consumer of these services through RPC.

trinity/plugins/eth2/beacon/validator.py
```diff
@@ -183,3 +226,83 @@ def skip_block(self,
         # will run the state transition which also includes the state transition for skipped slots.
         self.chain.chaindb.persist_state(post_state)
         return post_state.root
+
+    @curry
+    def _is_attesting(self,
+                      validator_index: ValidatorIndex,
+                      slot: Slot,
+                      epoch: Epoch) -> Tuple[bool, ValidatorIndex, Shard]:
```
mhchia: Also, IMO it is a little bit non-trivial that what the other fields mean in the `Tuple` returned by `_is_attesting`. Probably it would be better to makes this function simpler, or inline these logics in the callers?(It seems only called by `attest`.)

trinity/plugins/eth2/beacon/validator.py
```diff
@@ -183,3 +226,83 @@ def skip_block(self,
...
+        sorted_attesting_validators = sorted(
+            attesting_validators,
+            key=lambda attesting_data: attesting_data[2],
+        )
+        # Group the attesting validators by shard
+        from itertools import groupby
+        attesting_validators_groups = groupby(
+            sorted_attesting_validators,
+            lambda attesting_data: attesting_data[2],
+        )
```
mhchia: Consider putting this to the top?

trinity/plugins/eth2/beacon/validator.py
```diff
@@ -64,17 +78,22 @@ def __init__(
...
+        config = self.chain.get_state_machine().config
+        self.slots_per_epoch = config.SLOTS_PER_EPOCH
         # TODO: `latest_proposed_epoch` should be written into/read from validator's own db
-        self.latest_proposed_epoch = genesis_epoch
-        self.slots_per_epoch = slots_per_epoch
+        self.latest_proposed_epoch = {}
+        self.latest_attested_epoch = {}
+        self.this_epoch_assignment = {}
+        for validator_index in validator_privkeys:
+            self.latest_proposed_epoch[validator_index] = Epoch(-1)
```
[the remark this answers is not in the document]
NIC619: Because we might be proposing in the genesis epoch, this is a workaround to avoid needing extra logic for handling edge case in genesis epoch(i.e., extra check to prevent proposing twice in genesis epoch).

trinity/protocol/bcc/servers.py
```diff
@@ -238,13 +245,39 @@ def __init__(
...
         elif isinstance(cmd, BeaconBlocks):
             await self._handle_beacon_blocks(peer, cast(BeaconBlocksMessage, msg))
         else:
             raise Exception(f"Invariant: Only subscribed to {self.subscription_msg_types}")
 
+    async def _handle_attestations(self, peer: BCCPeer, msg: AttestationsMessage) -> None:
+        if not peer.is_operational:
+            return
+        encoded_attestations = msg["encoded_attestations"]
+        attestations = (
```
ChihChengLiang:
```suggestion
        attestations = tuple(
```

trinity/plugins/eth2/beacon/validator.py
```diff
@@ -90,6 +109,28 @@ def __init__(
         """
         async for event in self.event_bus.stream(SlotTickEvent):
```
ChihChengLiang: Ah, sorry I missed this in your previous PR. This is where we can determine what to do in each tick.
```python
async for event in self.event_bus.stream(SlotTickEvent):
    if not event.is_second_tick:
        # do something in the first tick
    else:
        # do something in the second tick
```

trinity/plugins/eth2/beacon/validator.py
```diff
@@ -104,17 +145,19 @@ def __init__(
             state_machine.config,
         )
         # Since it's expected to tick twice in one slot, `latest_proposed_epoch` is used to prevent
```
ChihChengLiang: So maybe we don't need `latest_proposed_epoch` if `is_second_tick` has been determined in `handle_slot_tick`

trinity/plugins/eth2/beacon/validator.py
NIC619: There's an extreme case(depending on how reliable `SlotTicker` is) where it didn't tick in the first half of the slot so in this case we still need to propose/attest when the `is_second_tick == True`.
But I guess if `SlotTicker` is reliable maybe we can disregard this case.

trinity/plugins/eth2/beacon/validator.py
NIC619: `latest_proposed_epoch` is still needed because we need to handle the case where `Validator` service crashed and restarted. Though we currently have not handled this properly since `latest_proposed_epoch` is not persisted into database yet.

eth2/beacon/tools/builder/validator.py
```diff
@@ -533,6 +532,78 @@ def create_mock_signed_attestations_at_slot(
...
+        sigs=(),
+        voting_sigs=signatures,
+        voting_committee_indices=voting_committee_indices,
+    )
+
+    # create attestation from attestation_data, particpipant_bitfield, and signature
+    return Attestation(
+        aggregation_bitfield=aggregation_bitfield,
+        data=attestation_data,
+        custody_bitfield=Bitfield(b'\x00' * len(aggregation_bitfield)),
```
ChihChengLiang: nitpick
```suggestion
        custody_bitfield=get_empty_bitfield(len(aggregation_bitfield)),
```

PR #635, opened 2019-05-20
Tidy up with Eth2GenesisConfig

ChihChengLiang:
### What was wrong?

Some genesis slots are accessed from constants like SERENITY_CONFIG, this makes it difficult to override them for various use cases.

Some configs are
- defined differently between spec versions. e.g. GENESIS_SLOT is sometimes 0 and sometimes 2**32 
- defined as a config of a single state machine. 

This is confusing because it's difficult to know when to access them as a constant or a variable.

### How was it fixed?

- Wrap configs that are unlikely to change with Eth2GenesisConfig.
- For classes who access these configs, rename the `__init__` parameters from `config` to `genesis_config`. So it is clear that we are accessing a genesis config rather than a specific state machine config.

#### Cute Animal Picture


eth2/configs.py
```diff
@@ -86,3 +86,18 @@ def __init__(self, config: Eth2Config):
...
+
+class Eth2GenesisConfig:
+    """
+    Genesis parameters that might lives in
+    a state or a state machine config
+    but is assumed unlikely to change between forks.
+    Pass this to the chains, chain_db, or other objects that need them.
+    """
+
+    def __init__(self, config: Eth2Config):
```
hwwhww: nitpick: missing type hint ` -> None`

eth2/beacon/db/chain.py
```diff
@@ -160,9 +162,9 @@ def get(self, key: bytes) -> bytes:
 
 
 class BeaconChainDB(BaseBeaconChainDB):
-    def __init__(self, db: BaseAtomicDB, config: Eth2Config) -> None:
+    def __init__(self, db: BaseAtomicDB, genesis_config: Eth2GenesisConfig) -> None:
```
hwwhww: I found there are some `trinity/tests/eth2/core/beacon/` tests are still using `chaindb = BeaconChainDB(self, db: BaseAtomicDB, config: Eth2Config)`; for example: https://github.com/ethereum/trinity/blob/b01e398dec4ac5db40210a4ab1d30cd884c0f05b/tests/eth2/core/beacon/db/test_beacon_chaindb.py#L38-L40

I think we can add new fixtures in `tests/eth2/core/beacon/conftest.py` like:

```python
@pytest.fixture
def genesis_config(config):
    return Eth2GenesisConfig(config)


@pytest.fixture
def beacon_chaindb(base_db, genesis_config):
    return BeaconChainDB(base_db, genesis_config)
```

eth2/configs.py
```diff
@@ -86,3 +86,18 @@ def __init__(self, config: Eth2Config):
...
+        self.GENESIS_SLOT = config.GENESIS_SLOT
+        self.GENESIS_EPOCH = config.GENESIS_EPOCH
+        self.SECONDS_PER_SLOT = config.SECONDS_PER_SLOT
+        self.SLOTS_PER_EPOCH = config.SLOTS_PER_EPOCH
```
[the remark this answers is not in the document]
ChihChengLiang: No AFAIK, but would like to make their hash consistent once #605 resolves.

ChihChengLiang:
Good catch! Thanks for the review.

PR #636, opened 2019-05-20
Move upnp service into an isolated plugin

cburgdorf:
### What was wrong?

The `networking` process is fading out and the upnp service is still one of the remaining things that is currently hosted by the `networking` process. Also the `upnp` service has some blocking code that we try to smooth out by running it in the thread pool but that still takes resources that are shared with other networking/sync.

### How was it fixed?

This doesn't actually make the Upnpservice usable (yet) but I wanted to keep this PR simple for now and send more changes in a follow up PR.

This:

1. **Does not** remove any current functionality (because we never use a discovered external IP anyway)
2. Speeds up the boot because we don't wait on upnp discovery anymore
3. Moves the `UpnpService` into its own isolated plugin
4. Improves logging inside the `UpnpService` a bit
5. Moves the service out of the `p2p` module.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


carver:
It's been a little while since I poked around, but I think we'll need some kind of way to update the IP/port we broadcast during discovery, so that the node can be contacted externally after updating the NAT.

cburgdorf:
>but I think we'll need some kind of way to update the IP/port we broadcast during discovery

Yep, there are multiple things here:

1. Currently, we do not use the external ip that the upnp service discovers at all and instead use `0.0.0.0`.

https://github.com/ethereum/trinity/blob/0a7e12179e40c5a9af36217a38f2f97b8f28de78/trinity/server.py#L145-L151

We once had this but it got lost during a refactoring. That should be easy to bring back though. I'm thinking that the new upnp support will just raise an event continuously every time it discovers a new ip / mapping.

When Trinity boots, we can just wait `n` seconds for an ip to be resolved and move on with `0.0.0.0` if that doesn't happen within our time budget.

2. But I'm still not sure what it means if the external IP then changes later. Does it mean we have to tear down the server and restart it with the `host` set to the new external ip?

3. Or does our started TCP listener stay static during the lifecycle of one run and the only thing we need to take care of is to ensure we advertise our external ip / port correctly to other peers?

pipermerriam:
I think the `0.0.0.0` is bad mojo that we need to get rid of soon.  My understanding is that it can lead to unwanted externally exposed servers.

Maybe we should launch the listener on `127.0.0.1` and then launch additional listeners on discovered external ports from `UPNP` and prioritize broadcasting the *latest* one from UPNP, falling back to the localhost one last?

trinity/plugins/builtin/upnp/plugin.py
```diff
@@ -0,0 +1,50 @@
+from argparse import (
+    ArgumentParser,
+    _SubParsersAction,
+)
+import asyncio
+
+
+from p2p.nat import (
+    UPnPService,
```
pipermerriam: Does it still make sense for `UPNP` to live in the `p2p` codebase?

cburgdorf: I was also thinking of moving it out there because it is fairly decoupled from the rest of the `p2p` module.

carver:
> * But I'm still not sure what it means if the external IP then changes later. Does it mean we have to tear down the server and restart it with the `host` set to the new external ip?

It's less common to change external IP, more common to change external port. Nope, we can leave the server running.

> * Or does our started TCP listener stay static during the lifecycle of one run and the only thing we need to take care of is to ensure we advertise our external ip / port correctly to other peers?

Right, this one ^.

The UPnP server is responsible for redirecting requests coming into the external IP/port toward the internal IP/port. So as the external one changes, the only thing we have to update is where we tell peers to find us.

carver:
> I think the `0.0.0.0` is bad mojo that we need to get rid of soon. My understanding is that it can lead to unwanted externally exposed servers.
> 
> Maybe we should launch the listener on `127.0.0.1` and then launch additional listeners on discovered external ports from `UPNP` and prioritize broadcasting the _latest_ one from UPNP, falling back to the localhost one last?

AIUI, binding to `127.0.0.1` would only let peers on the local computer make inbound requests.

We could bind to a particular IP address that we discover is alive, but I'm not convinced that provides any benefit. It does have a cost, of the server becoming inaccessible if the internet connection changes (like switching from wireless to LAN).

pipermerriam:
Am I incorrect about `0.0.0.0` being too permissive/broad?

carver:
> Am I incorrect about `0.0.0.0` being too permissive/broad?

We're trying to open a publicly reachable server on the computer, so yeah, I think it's the desired amount of permissive in this case.

trinity/plugins/builtin/upnp/nat.py
```diff
@@ -106,6 +106,7 @@ def __init__(self, port: int, token: CancelToken = None) -> None:
                     )
                     continue
                 else:
+                    self.logger.info("Found external IP address: %s", external_ip)
```
carver: This log looks redundant with the log at line 151. At some point, this method should also return the external port used, but I'm fine with punting that a bit.

trinity/plugins/builtin/upnp/nat.py
```diff
@@ -176,5 +177,6 @@ def current_mapping(self) -> PortMapping:
             try:
                 device.WANIPConn1
             except AttributeError:
+                self.logger.debug2("Device %s does not have an external IP. Skipping.", device)
```
carver: This log is confusing. These devices aren't missing external IPs, they are missing the interface needed to register the NAT mapping. So it might be a router with UPnP disabled, for example.

trinity/plugins/builtin/upnp/nat.py
cburgdorf: :see_no_evil:

trinity/plugins/builtin/upnp/nat.py
cburgdorf: :see_no_evil: This is a great way of making forward progress. I'll just keep on adding stupid comments to have things explained to me haha :)

To my defense, my router doesn't support upnp (at all) which makes it harder to work on this.

cburgdorf:
Thanks for your review @carver! I turned your comment into an issue https://github.com/ethereum/trinity/issues/691

PR #646, opened 2019-05-21
Rate limit the full sync import logs

carver:
### What was wrong?

Usually regular sync can just spam all the import block logs to `INFO`, because they're only coming once every 12s. But if you run regular sync from block 0, you get a screen full of block import logs relatively quickly.

### How was it fixed?

Added a token bucket to rate-limit the number of logs spit out to screen. If there's been a gap, show up to 5 in a row. Otherwise, only show one every 3s.

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/full/chain.py
+++ b/trinity/sync/full/chain.py
@@ -6,6 +6,7 @@ from functools import (
     partial,
 )
 from operator import attrgetter
+import time
 from typing import (
     AsyncIterator,
     Awaitable,
@@ -51,6 +52,7 @@ from p2p.exceptions import BaseP2PError, PeerConnectionLost
 from p2p.peer import BasePeer, PeerSubscriber
 from p2p.protocol import Command
 from p2p.service import BaseService
+from p2p.token_bucket import TokenBucket
 
 from trinity.chains.base import BaseAsyncChain
 from trinity.db.eth1.chain import BaseAsyncChainDB
@@ -976,6 +978,12 @@ class RegularChainBodySyncer(BaseBodyChainSyncer):
         # Track if any headers have been received yet
         self._got_first_header = asyncio.Event()
 
+        # Rate limit the block import logs
+        self._import_log_limiter = TokenBucket(
+            0.33,  # show about one log per 3 seconds
+            5,  # burst up to 5 logs after a lag
+        )
+
     async def _run(self) -> None:
         head = await self.wait(self.db.coro_get_canonical_head())
         self._block_import_tracker.set_finished_dependency(head)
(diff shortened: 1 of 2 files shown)


trinity/sync/full/chain.py
```diff
@@ -1069,8 +1076,19 @@ def _mark_body_download_complete(
...
+                if self._import_log_limiter.can_take(1):
+                    logger = self.logger.info
+                    self._import_log_limiter.take_nowait(1)
+                else:
+                    logger = self.logger.debug
+                logger(
+                    "Imported block %d (%d txs) in %.2f seconds",
+                    block.number,
+                    len(block.transactions),
+                    timer.elapsed,
```
pipermerriam: Thoughts on the actual log entry being updated to aggregate the log data?  Would result in two different messages, one that would trigger during fast imports, displaing aggregate info for multiple logs and then another for once things slow down that shows singular block data.

carver: You mean like if the token bucket has capacity then you display a log immediately, but if it is out of capacity then you start accumulating stats for the next log?

That gets a little ugly, adding a few more instance variables to keep track of the log data. I'll open a tracking issue.

carver: #701

trinity/sync/full/chain.py
```diff
@@ -1069,8 +1077,20 @@ def _mark_body_download_complete(
...
+                    logger = self.logger.info
+                    self._import_log_limiter.take_nowait(1)
+                else:
+                    logger = self.logger.debug
+                logger(
+                    "Imported block %d (%d txs) in %.2f seconds, with %s lag",
+                    block.number,
+                    len(block.transactions),
+                    timer.elapsed,
+                    humanize_seconds(time.time() - block.header.timestamp),
```
carver: I also added this to show how far back the imported header is from the current time. This is useful during beam sync, to show how far behind you're falling.

trinity/sync/full/chain.py
```diff
@@ -1069,8 +1077,20 @@ def _mark_body_download_complete(
...
                 # simple import of a single new block.
-                self.logger.info("Imported block %d (%d txs) in %.2f seconds",
-                                 block.number, len(block.transactions), timer.elapsed)
+
+                # decide whether to log to info or debug, based on log rate
+                if self._import_log_limiter.can_take(1):
+                    logger = self.logger.info
+                    self._import_log_limiter.take_nowait(1)
+                else:
+                    logger = self.logger.debug
```
pipermerriam: might be worth calling this `log_fn` or something other than `logger` to disambiguate.

PR #649, opened 2019-05-21
Drop git dependency for pypi release

carver:
### What was wrong?

pypi (reasonably) would not allow deployment with an `@git` dependency in `libp2p`

### How was it fixed?

Switched back to the standard release (0.0.7 was released in May, and the `@git` dependency was added in March, so maybe the feature made it upstream?)

#### Cute Animal Picture

### DIFF
--- a/p2p/_utils.py
+++ b/p2p/_utils.py
@@ -117,5 +117,8 @@ def trim_middle(arbitrary_string: str, max_length: int) -> str:
         half_len, is_odd = divmod(max_length, 2)
         first_half = arbitrary_string[:half_len - 1]
         last_half_len = half_len - 2 + is_odd
-        last_half = arbitrary_string[last_half_len * -1:]
+        if last_half_len > 0:
+            last_half = arbitrary_string[last_half_len * -1:]
+        else:
+            last_half = ''
         return f"{first_half}✂✂✂{last_half}"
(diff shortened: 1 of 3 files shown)


carver:
Contains an unrelated bugfix for #642

carver:
As soon as this turns green, I will merge and release. @ralexstokes if this causes problems, feel free to revert 3b4a206 in master after the release. (although this will be an issue at every release going forward)

carver:
Oh I cc'd Alex because he touched that line last, but @mhchia probably makes more sense to tag, since it was his fork.

ralexstokes:
@carver, i would guess the problem is related to this: https://github.com/multiformats/py-multiaddr/issues/47

once we get that resolved then we should be able to update to the new version of py-multiaddr.

thanks for patching this up for the release!

ralexstokes:
oh i see, actually we probably just need to update to the latest release of `py-multiaddr` (rather than specify a remote)... once the issue linked above is fixed i'll get a release on that repo and we can patch this up

ralexstokes:
@mhchia is there anything on your `feature/add-unix-proto` branch that hasn't been upstreamed yet?

mhchia:
@carver 
Thank you for the fix!

@ralexstokes 
Everything in `feature/add-unix-proto` is merged into the master. However, since there are breaking changes in later releases, to use the latest version I will make a PR to our libp2p bindings code.  And after the fix in py-multiaddr is merged, there should be no issue anymore.

PR #650, opened 2019-05-22
Discv5 Messages

jannikluhn:
### What was wrong?

This implements the various message types used in discv5. They are RLP objects with an integer ID. All message types are added to a registry so that it's easy to select the corresponding class if a packet specifying the message id is received.

I put this into it's own PR to make reviewing easier. It builds on top of #586, I'll rebase once that one is merged. Probably only worth reviewing after that.

#### Cute Animal Picture


jannikluhn:
This is now rebased and ready for review.

p2p/discv5/messages.py
```diff
@@ -0,0 +1,182 @@
...
+
+class MessageTypeRegistry(MessageTypeRegistryBaseType):
+
+    def register(self,
+                 message_data_class: Type["MessageData"]
+                 ) -> Type["MessageData"]:
+        """Class Decorator to register MessageData classes."""
+        message_type = message_data_class.message_type
+        if message_type is None:
+            raise ValueError("Message type must be defined")
```
pipermerriam: Do you think using our `eth_utils.ValidationError` would be more appropriate/consistent here?  Since these seem to all occur at module loading time I'm not sure how big a difference it makes but figured I'd toss the idea out.

jannikluhn: My rough guideline so far was to use `ValidationError` for everything that's ultimately based on external input, whereas here the exception would be caused by a bug in the implementation. So I think `ValueError` is more appropriate, but I'm happy to change it if `ValidationError` is used more broadly in the rest of the codebase.

pipermerriam: That reasoning works for me.  `ValueError` works.

PR #651, opened 2019-05-22
Drop usage of connect_to_endpoints_nowait API

cburgdorf:
### What was wrong?

Looks like we want to drop the `connect_to_endpoints_nowait` API soon so I removed the usage.

### How was it fixed?

Migrated to async API and used `asyncio.ensure_future`. Two alternatives that I considered:

- move the `asyncio.ensure_future`  into the `auto_connect_new_announced_endpoints` method
- keep using the `subscribe` API but with an `async` handler.

The latter isn't available yet but will be possible with a future lahja version. The former would mean to introduce yet another (maybe nested) function to feed into the `asyncio.ensure_future`.

In the end, I thought pushing the `asyncio.ensure_future` to the actual usage level is inline with the rest of the code and also more explicit. 

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/main.py
+++ b/trinity/main.py
@@ -156,7 +156,7 @@ async def launch_node_coro(args: Namespace, trinity_config: TrinityConfig) -> No
     )
 
     await endpoint.start_serving(networking_connection_config)
-    endpoint.auto_connect_new_announced_endpoints()
+    asyncio.ensure_future(endpoint.auto_connect_new_announced_endpoints())
     await endpoint.connect_to_endpoints(
         ConnectionConfig.from_name(MAIN_EVENTBUS_ENDPOINT, trinity_config.ipc_dir),
         # Plugins that run within the networking process broadcast and receive on the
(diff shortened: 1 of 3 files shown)


trinity/main.py
```diff
@@ -156,7 +156,7 @@ def launch_node(args: Namespace, trinity_config: TrinityConfig) -> None:
     )
 
     await endpoint.start_serving(networking_connection_config)
-    endpoint.auto_connect_new_announced_endpoints()
+    asyncio.ensure_future(endpoint.auto_connect_new_announced_endpoints())
```
pipermerriam: Not needed in this PR but I'd love to move this functionality out of the endpoint class entirely and have it be managed by a service or something like that so that we don't need to subclass the `lahja` endpoint classes locally.

cburgdorf: Yes, once we have `TrioEndpoint` and `AsyncioEndpoint` we want to share the functionality that is currently provided here across both. :+1:

PR #652, opened 2019-05-22
Implement encryption and decryption functions

jannikluhn:
Another small discv5 PR. This one implements the session encryption and decryption function. It uses the `cryptography` library that's used in the existing discovery code already.

It's finished from my side and doesn't depend on any other PR.

See also the [discv5 wire protocol spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md).

#### Cute Animal Picture


tests/p2p/discv5/test_encryption.py
```diff
@@ -0,0 +1,82 @@
...
+plain_text_st = st.binary(min_size=0, max_size=10)
+cipher_text_st = st.binary(min_size=0, max_size=10)
+aad_st = st.binary(min_size=0, max_size=10)
+
+
+def test_key_validation_invalid():
+    for length in (0, 12, 15, 17, 32):
+        with pytest.raises(ValidationError):
+            validate_aes128_key(AES128Key(b"\x00" * length))
+    validate_aes128_key(AES128Key(b"\x00" * 16))
```
pipermerriam: This line seems odd/out-of-place.

tests/p2p/discv5/test_encryption.py
```diff
@@ -0,0 +1,82 @@
...
+@given(key_st)
+def test_key_validation_valid(key):
+    validate_aes128_key(AES128Key(key))
+
+
+def test_nonce_validation_invalid():
+    for length in (0, 11, 13, 16):
+        with pytest.raises(ValidationError):
+            validate_nonce(Nonce(b"\x00" * length))
+    validate_nonce(Nonce(b"\x00" * 12))
```
pipermerriam: seems out of place or like it should be in it's own test.

tests/p2p/discv5/test_encryption.py
```diff
@@ -0,0 +1,82 @@
...
+from p2p.discv5.constants import (
+    AES128_KEY_SIZE,
+    NONCE_SIZE,
+)
+
+
+key_st = st.binary(min_size=AES128_KEY_SIZE, max_size=AES128_KEY_SIZE)
+nonce_st = st.binary(min_size=NONCE_SIZE, max_size=NONCE_SIZE)
+plain_text_st = st.binary(min_size=0, max_size=10)
+cipher_text_st = st.binary(min_size=0, max_size=10)
```
pipermerriam: This doesn't appear to be used.

p2p/discv5/encryption.py
```diff
@@ -0,0 +1,65 @@
...
+                   authenticated_data: bytes
+                   ) -> bytes:
+    validate_aes128_key(key)
+    validate_nonce(nonce)
+
+    aesgcm = AESGCM(key)
+    try:
+        plain_text = aesgcm.decrypt(nonce, cipher_text, authenticated_data)
+    except InvalidTag as error:
+        raise ValidationError("Decryption failed") from error
```
pipermerriam: Wondering if `ValidationError` is right.  Something like our own `DecryptionError` seem like it would be appropriate here.

veox:
`py3{6,7}-wheel-cli` jobs fail, because the `trinity -l=DEBUG2` runs do not produce expected log lines within 30 seconds.

@jannikluhn Could you re-run those jobs?.. To see if this is a heisenbug.

jannikluhn:
Reran the jobs (for the fourth time actually), this time only the py37 test failed. So there definitely is some luck involved. Maybe CircleCI is just too slow today?

veox:
Well, the [last test failing](https://circleci.com/gh/ethereum/trinity/43000?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) tries to connect to actual main-net bootnodes; I wouldn't be surprised it fails because the remotes are overloaded, or started blocking useless CI peers (I do, on my main-net nodes).

That's a blind guess, though, and it could as well be CircleCI's end, as you say.

I'd say we should be running these tests on a testnet anyway, at least those that aren't testing for main-net specifically... But that should probably be a separate PR (out of scope here).

jannikluhn:
@veox thank you! Once your PR was merged, the tests were passing.

PR #656, opened 2019-05-23
Reduce some duration of sleep(n) in tests

cburgdorf:
### What was wrong?

Timeout of `0.5` in tests is often more than what is actually needed.

### How was it fixed?

Reduced to `0.1`.

#### Cute Animal Picture


cburgdorf:
@carver yep poking around :)

PR #659, opened 2019-05-23
Clarify log level usage

cburgdorf:
### What was wrong?

Log level usage isn't explained anywhere.

### How was it fixed?

Tada!

image

Notice that the 5th footnote is missing on the screenshot but it is actually there (not enough space) :)

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/api/api.cli.rst
+++ b/docs/api/api.cli.rst
@@ -1,5 +1,5 @@
 Command Line Interface (CLI)
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+============================
 
 .. code-block:: shell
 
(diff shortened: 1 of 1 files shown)


docs/api/api.cli.rst
```diff
@@ -43,3 +43,34 @@ Command Line Interface (CLI)
...
                             The filesystem path to the file which contains the
                             nodekey
+
+
+
+Per-module logging
+~~~~~~~~~~~~~~~~~~
+
+Trinity provides rich logging output that can be of tremendous help during debugging. By default,
+only logging output that is classified as ``INFO`` is printed. This can be adjusted to other log
```
carver: Maybe specify "INFO or higher"?

docs/api/api.cli.rst
```diff
@@ -43,3 +43,34 @@ Command Line Interface (CLI)
...
+only interested in logging output of a specific module (e.g. ``trinity.sync``).
+
+Fortunately, Trinity allows us to configure the logging on a per-module basis by using the ``-l``
+flag multiple times such as in: ``trinity -l DEBUG2 -l p2p.discovery=ERROR``. In this case, we'd
+see logging output of the category ``DEBUG2`` for all modules except ``p2p.discovery`` for which
+the category is set to ``ERROR``.
+
+It is important to understand though, that additional provided ``-l`` flags do only work as filters
+on top of the global log level. In other words, if we are only interested in logging output of the
+``p2p.discovery`` module, the following **does not work**: ``trinity -l ERROR -l p2p.discovery=DEBUG2``.
```
carver: Oooh, that's what happening. That's really sad because this is almost always what you want: to have "normal" logs for all module by default, and show deep logs for one particular module.

docs/api/api.cli.rst
```diff
@@ -43,3 +43,34 @@ Command Line Interface (CLI)
...
+
+We can still achieve what we want by configuring the global log level to ``DEBUG2`` and reducing the
+log level for all top level modules to ``ERROR`` and then only setting the single module that we are
+interested in to ``DEBUG2``.
+
+``trinity -l DEBUG2 -l lahja=ERROR -l trinity=ERROR -l p2p=ERROR -l p2p.discovery=DEBUG2``
```
carver: Also `eth`? Also... I guess other things as we continue. Maybe an `eth2` top level? I think we'll want to come up with another solution before too long.

docs/api/api.cli.rst
```diff
@@ -43,3 +43,34 @@ Command Line Interface (CLI)
...
                             nodekey
+
+
+
+Per-module logging
+~~~~~~~~~~~~~~~~~~
+
+Trinity provides rich logging output that can be of tremendous help during debugging. By default,
+only logging output that is classified as ``INFO`` is printed. This can be adjusted to other log
+level such as ``DEBUG`` or ``DEBUG2`` by using the ``-l`` flag.
```
pipermerriam: for docs, maybe we should use the long-form of these flags.  They are more self documenting that way which in theory, also means that by encouraging the use of them in that form, secondary readers of scripts and things will have a clearer understanding of what is happening.

docs/api/api.cli.rst
pipermerriam: The only way I know to get around this would be to scrape for explicitly declared log levels that are below the main level, and then to manually attach handlers to each of those specific loggers.  I'd be :+1: for this change.

docs/api/api.cli.rst
cburgdorf: Yep, it just gets super lengthy with the compound flags used further down here. But maybe I can get rid of that :)

docs/api/api.cli.rst
cburgdorf: I think what we can do is just get a computed list of all top level modules and then the global log level would behind the scenes just set these individual levels rather than setting the actual global handler level.

I'm going to try this out and send a PR.

docs/api/api.cli.rst
cburgdorf: Oh yes, `eth` and `eth2` as well. Let me see if I can send an actual fix to make this more ergonomic. It's getting out of hand haha.

docs/api/api.cli.rst
pipermerriam: @cburgdorf I think that will work but it feels roughly equal in complexity but like a less precise solution since it still only makes it work through indirection and still requires the global level be set to the very low level.  Also, that means that other low level logs from other libraries will also make their way through which is very likely going to produce extra noise...

pipermerriam: Said a different way, I think your proposed solution would not fix the following situation since it doesn't set the global level where-as if we assign a handler directly to that logger I think it would work.

```python
trinity -l p2p.discovery=DEBUG2`
```

cburgdorf: For an invocation such as `trinity -l p2p.discovery=DEBUG2` it will set every logger to `logging.INFO` except `p2p.discovery` which is set to `DEBUG2`.

I have this working locally but the somehow maintaining a stable picture of the current logging setup with all its nuances consumes the entire available 640kb of my brain memory so I'm going just ignoring any other ideas for a moment :hear_no_evil: :see_no_evil: 

I'm ok to throw it away afterwards if yours proves to be better. :smile:

cburgdorf: Here we go https://github.com/ethereum/trinity/pull/664

cburgdorf:
@pipermerriam I updated this based on the latest version on master. I also added a screenshot for to assist a visual review. Care to look at it again?

pipermerriam:
:heart: :shipit:

PR #660, opened 2019-05-23
Fix #654, get attestation by root

### LINKED ISSUE #654: Store Attestation.root -> (Block_root, index)
### What is wrong?

This is a part of the attestation pool.

- When a block is processed, store Attestation.root -> (Block_root, index of Attestation)
- When received a new Attestation, discard if the root is seen before.

### How can it be fixed

Add API in the chaindb

ChihChengLiang:
### What was wrong?

#654

### How was it fixed?

- Add get_attestation_by_root 
- TODO: chain object test is somewhat non-trivial...

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/db/chain.py
+++ b/eth2/beacon/db/chain.py
@@ -167,6 +167,11 @@ class BaseBeaconChainDB(ABC):
     #
     # Attestation API
     #
+
+    @abstractmethod
+    def attestation_exists(self, attestation_root: Hash32) -> bool:
+        pass
+
     @abstractmethod
     def get_attestation_by_root(self,
                                 attestation_root: Hash32,
@@ -766,6 +771,10 @@ class BeaconChainDB(BaseBeaconChainDB):
             )
         attestation_key = ssz.decode(encoded_key, sedes=AttestationKey)
         return attestation_key.block_root, attestation_key.index
+
+    def attestation_exists(self, attestation_root: Hash32) -> bool:
+        lookup_key = SchemaV1.make_attestation_root_to_block_lookup_key(attestation_root)
+        return self.exists(lookup_key)
 
     def get_attestation_by_root(self,
                                 attestation_root: Hash32,
--- a/eth2/beacon/db/chain.py
+++ b/eth2/beacon/db/chain.py
@@ -46,6 +46,9 @@ from eth2.beacon.typing import (
     Epoch,
     Slot,
 )
+from eth2.beacon.types.attestations import (
+    Attestation,
+)
 from eth2.beacon.types.states import BeaconState  # noqa: F401
 from eth2.beacon.types.blocks import (  # noqa: F401
     BaseBeaconBlock,
@@ -64,6 +67,13 @@ from eth2.beacon.db.schema import SchemaV1
 from eth2.configs import (
     Eth2GenesisConfig,
 )
+
+
+class AttestationKey(ssz.Serializable):
+    fields = [
+        ('block_root', ssz.sedes.bytes32),
+        ('index', ssz.sedes.uint8),
+    ]
 
 
 class BaseBeaconChainDB(ABC):
@@ -151,6 +161,15 @@ class BaseBeaconChainDB(ABC):
     @abstractmethod
     def persist_state(self,
                       state: BeaconState) -> None:
+        pass
+
+    #
+    # Attestation API
+    #
+    @abstractmethod
+    def get_attestation_by_root(self,
+                                attestation_root: Hash32,
+                                block_class: Type[BaseBeaconBlock]) -> Attestation:
         pass
 
     #
@@ -468,6 +487,7 @@ class BeaconChainDB(BaseBeaconChainDB):
         )
         cls._add_block_root_to_slot_lookup(db, curr_block_head)
         cls._set_block_scores_to_db(db, curr_block_head)
+        cls._add_attestations_root_to_block_lookup(db, curr_block_head)
 
         orig_blocks_seq = concat([(first_block,), blocks_iterator])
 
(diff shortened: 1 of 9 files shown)


trinity/db/beacon/chain.py
```diff
@@ -104,10 +107,19 @@ class BaseAsyncBeaconChainDB(ABC):
...
 
+    @abstractmethod
+    def coro_get_attestation_by_root(self,
+                                     attestation_root: Hash32,
+                                     block_class: Type[BaseBeaconBlock]) -> Attestation:
+        pass
     #
     # Raw Database API
     #
+
```
NIC619:
```suggestion

```

trinity/db/beacon/chain.py
```diff
@@ -104,10 +107,19 @@ class BaseAsyncBeaconChainDB(ABC):
     async def coro_persist_state(self,
                                  state: BeaconState) -> None:
         pass
+    #
+    # Attestation API
+    #
```
NIC619:
```suggestion

```

hwwhww:
The requirements of #660 is having a function to check if the attestation is already in the pool, but this PR is adding `get_attestation_by_root(self, attestation_root: Hash32)-> Attestation`. From the requirements, I think we don't need to decode the block to and get the `Attestation`.

I mean, adding  `get_attestation_by_root(self, attestation_root: Hash32)-> Attestation` may be useful and I'd say we should keep it, but for the Trinity side, to "check if attestation is already in the pool", we can add another simpler API like `attestation_exists(self, attestation_root: Hash32)-> bool`.

NIC619:
@hwwhww @ChihChengLiang I apologize for the confusion, by attestation pool I mentioned in #654 I mean the in-memory attestation pool of `ReceiveServer`, not the attestations stored in database.

eth2/beacon/db/chain.py
```diff
@@ -719,6 +740,34 @@ def _handle_exceptional_justification_and_finality(self,
...
+    def _get_attestation_key_by_root(db: BaseDB, attestation_root: Hash32) -> Tuple[Hash32, int]:
+        encoded_key = db[SchemaV1.make_attestation_root_to_block_lookup_key(attestation_root)]
+        attestation_key = ssz.decode(encoded_key, sedes=AttestationKey)
+        return attestation_key.block_root, attestation_key.index
+
+    def get_attestation_by_root(self,
+                                attestation_root: Hash32,
+                                block_class: Type[BaseBeaconBlock]) -> Attestation:
+        block_root, index = self._get_attestation_key_by_root(self.db, attestation_root)
+        block = self._get_block_by_root(self.db, block_root, block_class)
```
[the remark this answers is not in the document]
ChihChengLiang: This makes perfect sense. Since `get_block_class` is a method of `chain`, the `get_attestation_by_root` should only be a method of `chain` too.

tests/eth2/core/beacon/chains/test_beacon_chain.py
```diff
@@ -140,3 +144,58 @@ def test_from_genesis(base_db,
...
+
+    def build_block(state, blocks, attestations):
+        return create_mock_block(
+            state=state,
+            config=config,
+            state_machine=valid_chain.get_state_machine(blocks[-1]),
+            block_class=genesis_block.__class__,
+            parent_block=blocks[-1],
+            keymap=keymap,
+            slot=state.slot + 2,
```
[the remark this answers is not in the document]
ChihChengLiang: Good catch, should be `state.slot + 1`.

ChihChengLiang:
Good catches and thanks for the review. `attestation_exists` API is added. `get_attestation_by_root` is now handled solely by `chain`, ensuring the correct block class at the block root is applied.
The PR is ready to be reviewed again.

trinity/db/beacon/chain.py
```diff
@@ -105,6 +105,18 @@ class BaseAsyncBeaconChainDB(ABC):
                                  state: BeaconState) -> None:
         pass
 
+    #
+    # Attestation API
+    #
+
```
NIC619: nit:
```suggestion

```

tests/eth2/core/beacon/chains/test_beacon_chain.py
```diff
@@ -140,3 +144,58 @@ def test_from_genesis(base_db,
...
+            parent_block=blocks[-1],
+            keymap=keymap,
+            slot=state.slot + 1,
+            attestations=attestations,
+        )
+    for _ in range(min_attestation_inclusion_delay + 1):
+        block = build_block(state, blocks, tuple())
+        valid_chain.import_block(block)
+        state = valid_chain.get_state_machine(block).state
+        blocks += (block,)
```
hwwhww: I think we can just test *one block + one attestation*, and skip `min_attestation_inclusion_delay` blocks in order to decrease the computation in this test.

Speaking of that, not necessary in this PR but we may need new APIs in `BaseBeaconChain` -> `StateTransition` -> `StateMachine` to perform `apply_state_transition_without_block` (similar to `import_block` for `apply_state_transition`), instead of calling `post_state = state_machine.state_transition.apply_state_transition_without_block(...)`. What do you think about it?

NIC619: This would definitely be helpful!

ChihChengLiang: Set `min_attestation_inclusion_delay =0`, reducing test time from 29.22 sec 🐢 to 11.19 🐇.  

For the new API, just realized the need is already reflected in the validator client and I believe the name of it is `skip_block`.
https://github.com/ethereum/trinity/blob/master/trinity/plugins/eth2/beacon/validator.py#L260-L279
Will create an issue for this.

ChihChengLiang: Move the discussion in #677

PR #664, opened 2019-05-24
Make logging more ergonomic and flexible

cburgdorf:
### What was wrong?

Configuring the individual loggers is cumbersome (see https://github.com/ethereum/trinity/pull/659#discussion_r287093201)

### How was it fixed?

Let's go though a bunch of examples:

1. Run `trinity` -> stderr prints exactly the same logging output as current master (`INFO` only) 

2. Run `trinity -l p2p.discovery=DEBUG2` -> stderr prints only `INFO` logs except for `p2p.discovery` where `DEBUG2` or higher is shown

3. Run `trinity -l ERROR -l p2p.discovery=DEBUG2` -> stderr prints only `ERROR` logs except for `p2p.discovery` where `DEBUG2` or higher is shown

4. Run `trinity -l DEBUG2 --stderr-log-level INFO` -> stderr prints only `INFO` logs but file log prints `DEBUG2` or higher logs

(^--This combination of CLI flags was previously illegal but I *think*  you were still able to get the same result using different CLI flags?)

5. Run `trinity -l DEBUG2 --file-log-level INFO` -> stderr prints `DEBUG2` or higher logs but file log prints only `INFO` logs

(The reverse of 4)

### Additional notes

1. The `sqlalchemy` library prints lots of logs directly on `INFO` which is why I silenced it manually. I'm wondering why we don't see these logs on current master? We do see logs from `lahja` which just happens to be another 3rd party library that we use. I don't see the difference here.

2. As @pipermerriam pointed out in #659 this approach may lead us to show more `DEBUG` logs from 3rd party libraries when we run Trinity with `-l DEBUG` but kinda related to my previous point: How do we differentiate between a 3rd party library like lahja where we **do** want to see the logs from another one where we don't? From a technical standpoint these just happen to be equal 3rd party libraries.

3. Previously the `--log-level` and `--stderr-log-level` flags had overlapping responsibilities but the `--log-level` and the `--file-log-level` flags did not. I found that really confusing and this PR treats them entirely the same way.

At least this PR gives us a way to configure the noise on a per module basis.

### To-Do

- [ ] Clean up commit history
- [ ] write tests to proof all the claims this PR makes 

#### Cute Animal Picture

### DIFF
--- a/trinity/_utils/logging.py
+++ b/trinity/_utils/logging.py
@@ -12,6 +12,7 @@ from logging.handlers import (
 )
 import os
 from pathlib import Path
+import pkgutil
 import sys
 from typing import (
     Any,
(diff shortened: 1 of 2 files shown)


pipermerriam:
I am :-1: on this approach.  It feels far too heavy handed to require enumeration of a module's loggers and set all of their levels, especially when I'm pretty sure we have a path to do this much more simply.

I'm also not convinced that this won't have un-expected side effects.

cburgdorf:
I just wanted to put this together to have something that I know works so I can compare it against your proposed approach. I'll check that one next.

pipermerriam:
This seems to do exactly what we want (except that I didn't fix the logfile level stuff...)

https://github.com/ethereum/trinity/pull/665

cburgdorf:
This does in fact look a whole lot simpler! Do you plan to clean this up and send a PR or do you want me to do that? Either way is fine.

pipermerriam:
If you have capacity it is all yours

Sent from ProtonMail mobile

cburgdorf:
Sure, I get it in on Monday!

Am Fr., 24. Mai 2019 um 17:54 Uhr schrieb Piper Merriam <
notifications@github.com>:

cburgdorf:
Closing this in favor for #682

PR #665, opened 2019-05-24
Make logging work like I want it to...

pipermerriam:
Make it work...

### DIFF
--- a/trinity/_utils/logging.py
+++ b/trinity/_utils/logging.py
@@ -71,16 +71,28 @@ class HasExtendedDebugLogger:
 
 def setup_log_levels(log_levels: Dict[str, int]) -> None:
     for name, level in log_levels.items():
+        formatter = TrinityLogFormatter(
+            fmt='%(levelname)8s  %(asctime)s  %(shortname)20s  %(message)s',
+            datefmt='%m-%d %H:%M:%S'
+        )
+
+        handler_stream = logging.StreamHandler(sys.stderr)
+        handler_stream.setLevel(level)
+        handler_stream.setFormatter(formatter)
+
         logger = logging.getLogger(name)
         logger.setLevel(level)
+        logger.addHandler(handler_stream)
+
+        logger.info('SETUP LOGGER FOR: %s=%s', name, logging.getLevelName(level))
 
 
 def setup_trinity_stderr_logging(level: int=None,
                                  ) -> Tuple[Logger, Formatter, StreamHandler]:
     if level is None:
         level = logging.INFO
-    logger = logging.getLogger('trinity')
-    logger.setLevel(logging.DEBUG)
+    logger = logging.getLogger()
+    logger.setLevel(0)
 
     handler_stream = logging.StreamHandler(sys.stderr)
     handler_stream.setLevel(level)


trinity/_utils/logging.py
```diff
@@ -71,16 +71,28 @@ def logger(self) -> ExtendedDebugLogger:
 
 def setup_log_levels(log_levels: Dict[str, int]) -> None:
     for name, level in log_levels.items():
+        formatter = TrinityLogFormatter(
```
pipermerriam: In theory we can re-use the same formatter instance across all loggers if we want to be efficient.

trinity/_utils/logging.py
```diff
@@ -71,16 +71,28 @@ def logger(self) -> ExtendedDebugLogger:
...
+            fmt='%(levelname)8s  %(asctime)s  %(shortname)20s  %(message)s',
+            datefmt='%m-%d %H:%M:%S'
+        )
+
+        handler_stream = logging.StreamHandler(sys.stderr)
```
pipermerriam: in theory we can cache this and only ever make one of these per log level and re-use it across all loggers at that level.

cburgdorf: I guess we might have to. I noticed that when you run `trintiy -l DEBUG` you get every message printed twice. So it probably means we are accidentally setting up duplicate handler.

cburgdorf: Ok, I'm a bit lost here. I tried adding a `logger.handlers.clear()` in `setup_log_levels` (which probably isn't the right thing to do) and this does remove the duplicate log entries for when you just run `trinity`. But if you run `trinity -l trinity=DEBUG` you'll start seeing duplicate entries again.

```
    INFO  05-27 14:26:56               trinity  
      ______     _       _ __       
     /_  __/____(_)___  (_) /___  __
      / / / ___/ / __ \/ / __/ / / /
     / / / /  / / / / / / /_/ /_/ / 
    /_/ /_/  /_/_/ /_/_/\__/\__, /  
                           /____/   
    INFO  05-27 14:26:56               trinity  
      ______     _       _ __       
     /_  __/____(_)___  (_) /___  __
      / / / ___/ / __ \/ / __/ / / /
     / / / /  / / / / / / /_/ /_/ / 
    /_/ /_/  /_/_/ /_/_/\__/\__, /  
                           /____/   
    INFO  05-27 14:26:56               trinity  Started main process (pid=19297)
    INFO  05-27 14:26:56               trinity  Started main process (pid=19297)
```

My conclusion goes as follows: The intention of this solution is to add **additional** handler to selectively pick logs that would otherwise be muted (e.g. `trinity -l ERROR -l p2p=DEBUG`). But at the same time, **if it isn't muted`, `-l=<module>` will just end up duplicating log output by handling some module's log explicitly if it is also already covered through a logger higher up the stack.

This is why `trinity -l trinity=DEBUG` will cause duplicates for every trinity log and `trinity -l p2p=DEBUG2` will **not only** give you additional `DEBUG2` logs for the `p2p` module but will cause every single `INFO` of the `p2p` module to also get printed twice.

I guess what would be needed with this solution is an additional filter that ensures that **additional** handler do only let logs go through that are not already going through based on the root logger's configuration. :thinking:

cburgdorf: Ok, if we set `logger.propagate = False` it seems we get the behavior we want :)

cburgdorf:
@pipermerriam yes, and it's ready for review! https://github.com/ethereum/trinity/pull/682

PR #666, opened 2019-05-24
Free the fork choice

ralexstokes:
**EDIT: update description to reflect current state of PR, may be helpful to start w/ `Summary of changes` below**

### What was wrong?

Big part of implementing LMG GHOST fork choice.

Pulling this out of #532 as it is substantial enough to stand on its own.

In Eth1, the fork scoring function is just a function of the block header. This fact makes it easy to calculate the score as we import new headers, adjusting the canonical head along the way.

In Eth2, the fork scoring function is more complex (a function of the block, the state, and off-chain attestations); moreover, we want to run the fork choice in more situations than just importing a new valid block.

The current design of the fork choice calculation does not readily facilitate these tasks.

### How was it fixed?

Summary of changes:
- Extend API of `persist_block` to allow arbitrary scoring
- Add `set_score` API (not technically required for this PR but will need for future fork choice work. easier to include here than patch up the git history... forgive me for the lack of precision :) )
- Update the tests!

To address these limitations, we allow the caller of `persist_block` to provide their own scoring logic.
~The DB class now exposes methods to persist the fork choice but does not actively compute it (at least via the new API `persist_block_without_scoring`). The particular fork choice lives to a particular variation of the `StateMachine`. For example, the `SerenityStateMachine` will have the `lmd_ghost_fork_scoring` rule after it lands in a subsequent PR. For now, they just use the existing `higher_slot_scoring` rule.~

### Rationale for putting the fork choice scoring in the StateMachine and using it in the Chain class

The Chain class is what ultimately orchestrates validating incoming data, persisting valid data, running the fork choice (based on the current state machine's scoring rule) and persisting the results of this fork choice (i.e. writing the new block's score and updating the canonical head if necessary).

This configuration is more flexible, exposing a direct path to inject the AttestationPool necessary for LMD GHOST, and avoids the situation where the DB knows about every other object in the system (single responsibility and all that).

~I have some concerns mainly around the new API(s). It might be dangerous to expose something like `set_score` for anyone to call and `update_canonical_head_if_needed` feels a little clunky still... I would suggest we merge this in and as we gain confidence w/ the new stuff, we can look at deprecating the old stuff (namely, `persist_block` and its callers).~

### Other comments

We could make the argument to pull the actual process of selecting which fork to follow (given a "score chain") but this seems low value given that we can map any fork choice to "follow the monotonically increasing score" by warping the fork choice scoring function).

There is another concern around "overloading" the state machine w/ the fork choice -- it seems like the best place to put it for now but readers should note that the state machine is no longer just a transformation from block to block. To justify this change, let's answer the question: if we change the fork choice (at some slot), does that imply a new state machine?

### To-Do

I'll add more tests but given there are some design questions up in the air, I'll get this up so you can make a pass at reviewing it if you want...

- [x] Add more tests to cover new APIs

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/fork_choice/fork_choice_scoring.py
+++ b/eth2/beacon/fork_choice/fork_choice_scoring.py
@@ -1,0 +1,5 @@
+from typing import Callable
+
+from eth2.beacon.types.blocks import BaseBeaconBlock
+
+ForkChoiceScoring = Callable[[BaseBeaconBlock], int]
--- a/eth2/beacon/state_machines/forks/serenity/__init__.py
+++ b/eth2/beacon/state_machines/forks/serenity/__init__.py
@@ -22,12 +22,12 @@ from .state_transitions import SerenityStateTransition
 class SerenityStateMachine(BeaconStateMachine):
     # fork name
     fork = 'serenity'  # type: str
+    config = SERENITY_CONFIG
 
     # classes
     block_class = SerenityBeaconBlock  # type: Type[BaseBeaconBlock]
     state_class = SerenityBeaconState  # type: Type[BeaconState]
     state_transition_class = SerenityStateTransition  # type: Type[BaseStateTransition]
-    config = SERENITY_CONFIG
 
     # methods
     @staticmethod
--- a/eth2/beacon/state_machines/forks/serenity/__init__.py
+++ b/eth2/beacon/state_machines/forks/serenity/__init__.py
@@ -1,6 +1,9 @@
 from typing import Type  # noqa: F401
 
-from eth2.beacon.fork_choice import higher_slot_scoring
+from eth2.beacon.fork_choice import (
+    ForkChoiceScoring,
+    higher_slot_scoring,
+)
 from eth2.beacon.typing import (
     FromBlockParams,
 )
@@ -29,10 +32,12 @@ class SerenityStateMachine(BeaconStateMachine):
     block_class = SerenityBeaconBlock  # type: Type[BaseBeaconBlock]
     state_class = SerenityBeaconState  # type: Type[BeaconState]
     state_transition_class = SerenityStateTransition  # type: Type[BaseStateTransition]
-    fork_choice_scoring = staticmethod(higher_slot_scoring)
 
     # methods
     @staticmethod
     def create_block_from_parent(parent_block: BaseBeaconBlock,
                                  block_params: FromBlockParams) -> BaseBeaconBlock:
         return create_serenity_block_from_parent(parent_block, block_params)
+
+    def get_fork_choice_scoring(self) -> ForkChoiceScoring:
+        return higher_slot_scoring
--- a/eth2/beacon/state_machines/forks/serenity/__init__.py
+++ b/eth2/beacon/state_machines/forks/serenity/__init__.py
@@ -1,5 +1,6 @@
 from typing import Type  # noqa: F401
 
+from eth2.beacon.fork_choice import higher_slot_scoring
 from eth2.beacon.typing import (
     FromBlockParams,
 )
@@ -28,6 +29,7 @@ class SerenityStateMachine(BeaconStateMachine):
     block_class = SerenityBeaconBlock  # type: Type[BaseBeaconBlock]
     state_class = SerenityBeaconState  # type: Type[BeaconState]
     state_transition_class = SerenityStateTransition  # type: Type[BaseStateTransition]
+    fork_choice_scoring = higher_slot_scoring
 
     # methods
     @staticmethod
--- a/eth2/beacon/state_machines/forks/serenity/__init__.py
+++ b/eth2/beacon/state_machines/forks/serenity/__init__.py
@@ -29,7 +29,7 @@ class SerenityStateMachine(BeaconStateMachine):
     block_class = SerenityBeaconBlock  # type: Type[BaseBeaconBlock]
     state_class = SerenityBeaconState  # type: Type[BeaconState]
     state_transition_class = SerenityStateTransition  # type: Type[BaseStateTransition]
-    fork_choice_scoring = higher_slot_scoring
+    fork_choice_scoring = staticmethod(higher_slot_scoring)
 
     # methods
     @staticmethod
--- a/tests/core/p2p-proto/bcc/helpers.py
+++ b/tests/core/p2p-proto/bcc/helpers.py
@@ -38,6 +38,7 @@ from eth2.beacon.constants import (
 from tests.core.integration_test_helpers import (
     async_passthrough,
 )
+from eth2.beacon.fork_choice import higher_slot_scoring
 from eth2.beacon.state_machines.forks.serenity import SERENITY_CONFIG
 from eth2.configs import (
     Eth2GenesisConfig,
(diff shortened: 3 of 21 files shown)


ralexstokes:
@hwwhww given that you mentioned you would make a pass at this -- another option here is to keep `persist_block` and just add another parameter which is the fork choice.

so `persist_block(block, block_class)` becomes `persist_block(block, block_class, fork_choice_scoring)`

as i'm writing these tests, i'm having trouble thinking of cases where you would really want to persist a block w/o scoring it and finding the new head... the goal w/ this entire stream of work is so that we can execute arbitrary logic when running the fork choice.

one way to do this is how this PR currently is (expose the pieces and let the client of this code 'deal w/ it'). the other way i'm alluding to in this comment is that we just pass in this code as the `fork_choice_scoring`.

e..g for LMD GHOST, something like:

```python
def lmd_ghost_fork_choice_scoring(attestation_pool):
    return lambda block, chaindb: run_lmd_ghost(block, chaindb, attestation_pool)
```

you have a preference for one way or the other? i think this latest way avoids some of the security/usability concerns of the way that this PR currently proposes...

tests/eth2/core/beacon/db/test_beacon_chaindb.py
```diff
@@ -33,13 +35,13 @@
...
+def chaindb_at_genesis(chaindb, genesis_state, genesis_block, fork_choice_scoring):
     chaindb.persist_state(genesis_state)
-    chaindb.persist_block(genesis_block, BeaconBlock)
+    chaindb.persist_block(genesis_block, BeaconBlock, fork_choice_scoring)
     return chaindb
 
 
-@pytest.fixture(params=[0, 10, 999])
+@pytest.fixture(params=[1, 10, 999])
 def block(request, sample_beacon_block_params):
```
ralexstokes: to not conflict w/ genesis block

ralexstokes:
ok! this is a little messy as it tracked my thinking about the best way to get towards a more flexible fork choice :)

after I rebase, it should be ready to merge pending review. the tests all pass locally but circleCI is down at the moment.

ralexstokes:
just to update, this is officially ready for review now :)

eth2/beacon/state_machines/forks/serenity/__init__.py
```diff
@@ -22,15 +26,18 @@
...
-    config = SERENITY_CONFIG
 
     # methods
     @staticmethod
     def create_block_from_parent(parent_block: BaseBeaconBlock,
                                  block_params: FromBlockParams) -> BaseBeaconBlock:
         return create_serenity_block_from_parent(parent_block, block_params)
+
+    def get_fork_choice_scoring(self) -> ForkChoiceScoring:
+        return higher_slot_scoring
```
ChihChengLiang: How would the lmd_ghost looks like in this method?
Does the method produce new functions for every slot and block?

My imagination:
```python
def get_fork_choice_scoring(self) -> ForkChoiceScoring:
    def scoring(block):
       return lmd_ghost_scoring(db=self.chaindb.db, attestation_pool=somewhere.attestation_pool, state=self.state, start_block=somewhere.start_block, target_block=block, config=self.config)
    return scoring

```

ralexstokes: yeah more or less! modulo those parameters changing, e.g. i don't think the `StateMachine` has a reference directly to the `chain_db` and similar for the other dependencies...

i'll think a bit about the best way to keep a unified function interface while still giving us flexibility...

there is also a concern about how often this closure is created -- it would be possible to memoize this function, esp if the attestation pool is a reference to the mutable thing, not just a description of attestations -- this route generally goes against our preference for immutability though so i'd be inclined to find something more, well, immutable :)

ralexstokes: @ChihChengLiang latest thinking here: 

https://github.com/ethereum/trinity/pull/685/commits/08523d3c3a545785c0e457323c8cff1e5cd952ab

https://github.com/ethereum/trinity/pull/685/commits/ec29a7223ed432e365cfbb509d8332860309283b

like i say in the PR, i'm still chewing on a clean way to further modularize the fork choice so it is not so tightly bound to a particular `StateMachine`.

tests/core/p2p-proto/bcc/helpers.py
```diff
@@ -100,10 +102,18 @@ def create_branch(length, root=None, **start_kwargs):
...
+                       fork_choice_scoring=higher_slot_scoring):
     db = AtomicDB()
     chain_db = FakeAsyncBeaconChainDB(db=db, genesis_config=genesis_config)
-    await chain_db.coro_persist_block_chain(blocks, BeaconBlock)
+    await chain_db.coro_persist_block_chain(
+        blocks,
+        BeaconBlock,
+        (
+            higher_slot_scoring for block in blocks
+        ),
```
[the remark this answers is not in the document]
ralexstokes: interesting... looks like the second one is much faster!

eth2/beacon/fork_choice/fork_choice_scoring.py
```diff
@@ -0,0 +1,5 @@
+from typing import Callable
+
+from eth2.beacon.types.blocks import BaseBeaconBlock
+
+ForkChoiceScoring = Callable[[BaseBeaconBlock], int]
```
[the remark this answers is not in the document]
ralexstokes: yeah appending `Fn` seems ok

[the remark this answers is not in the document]
ralexstokes: my thinking was that earlier in this process I wasn't sure this abstraction would just be a `Callable`.... the `ForkChoice` could evolve so that it is more convenient to have more state, in which case we could just go ahead w/ a class and `ForkChoiceScoring` would become an abstract class (for the interface)...

but i think we can get away w/ just a `Callable` so let's do that for now

ralexstokes:
>     * `state_machine.get_fork_choice_scoring()` function refactoring: instead of returning a `Callable`, it seems similar to how eth1 `eth/vm/forks/` overrides the static methods. For example, the `compute_difficulty` [in `Frontier`](https://github.com/ethereum/py-evm/blob/915cec2a475176ad9722869e845dca8bac7a66d8/eth/vm/forks/frontier/__init__.py#L75) and [in `HomesteadVM`](https://github.com/ethereum/py-evm/blob/915cec2a475176ad9722869e845dca8bac7a66d8/eth/vm/forks/homestead/__init__.py#L39). I think we can utilize this pattern in `SerenityStateMachine`. Defer to the snake charmers on which pattern is more appropriate in this case. :)

yeah i went down this route before I landed on the current thing but decided to do this mainly because I ran into issues w/ `mypy`. and so I just looked at the links you added for the `eth1` stuff and guess what... they just `# type: ignore`.

my thinking was to just do the simple thing to keep moving :)

the one caveat is that w/ LMD GHOST there is basically a "setup" step where we need to fetch the latest justified state and it seems much more ergonomic as an instance method... i'll refer you here: https://github.com/ethereum/trinity/pull/685/files#diff-26594f899e3bcc74e868bbab3824d8f9R46

ralexstokes:
unless anyone has further comment or concern, i'll go ahead and merge this soon... i'll address https://github.com/ethereum/trinity/pull/666#discussion_r289296699 in #685 which builds off of this PR

hwwhww:
@ralexstokes 
> issues w/ `mypy`

Right, I recalled it was really weird: https://github.com/ethereum/py-evm/pull/1362#discussion_r223103362 :'(

>  i'll go ahead and merge this soon... i'll address #666 (comment) in #685 which builds off of this PR

Sounds good to me!

PR #667, opened 2019-05-26
Add attestation pool in beacon node `ReceiveServer`

NIC619:
### How was it fixed?
- [x] Add `AttestationPool`
    - `get`/`get_all`/`add`/`remove`
- [x] Add `ReceiveServer.get_ready_attestations`
    - which return attestations that are ready to be included in block
- [x] Expose `ReceiveServer.get_ready_attestations` to `Validator`
    - validator would get ready attestations from receive server when proposing
- [x] Fix: validator should `attest` at second slot tick
- [x] When receiving new not-yet-included attestation, check if it's already in the database
    - pending on #660 
- [x] Remove attestations from attestation pool when identical attestations are included in an imported block

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/tests/eth2/core/beacon/chains/test_beacon_chain.py
+++ b/tests/eth2/core/beacon/chains/test_beacon_chain.py
@@ -3,14 +3,17 @@ import copy
 import pytest
 
 
-from eth2.beacon.chains.base import (
-    BeaconChain,
-)
 from eth2.beacon.exceptions import (
     BlockClassError,
 )
-from eth2.beacon.state_machines.forks.serenity.blocks import (
-    SerenityBeaconBlock,
+from eth2.beacon.chains.base import (
+    BeaconChain,
+)
+from eth2.beacon.db.exceptions import (
+    AttestationRootNotFound,
+)
+from eth2.beacon.types.blocks import (
+    BeaconBlock,
 )
 from eth2.beacon.tools.builder.proposer import (
     create_mock_block,
@@ -19,8 +22,8 @@ from eth2.beacon.tools.builder.proposer import (
 from eth2.beacon.tools.builder.validator import (
     create_mock_signed_attestations_at_slot,
 )
-from eth2.beacon.types.blocks import (
-    BeaconBlock,
+from eth2.beacon.state_machines.forks.serenity.blocks import (
+    SerenityBeaconBlock,
 )
 
 
@@ -189,3 +192,9 @@ def test_get_attestation_root(valid_chain,
     a0 = attestations[0]
     assert valid_chain.get_attestation_by_root(a0.root) == a0
     assert valid_chain.attestation_exists(a0.root)
+    fake_attestation = a0.copy(
+        aggregate_signature=b'\x78' * 96,
+    )
+    with pytest.raises(AttestationRootNotFound):
+        valid_chain.get_attestation_by_root(fake_attestation.root)
+    assert not valid_chain.attestation_exists(fake_attestation.root)
--- a/tests/eth2/core/beacon/chains/test_beacon_chain.py
+++ b/tests/eth2/core/beacon/chains/test_beacon_chain.py
@@ -188,3 +188,4 @@ def test_get_attestation_root(valid_chain,
     # Only one attestation in attestations, so just check that one
     a0 = attestations[0]
     assert valid_chain.get_attestation_by_root(a0.root) == a0
+    assert valid_chain.attestation_exists(a0.root)
--- a/tests/plugins/eth2/beacon/test_receive_server.py
+++ b/tests/plugins/eth2/beacon/test_receive_server.py
@@ -632,3 +632,8 @@ async def test_bcc_receive_server_get_ready_attestations(
             attesting_slot + XIAO_LONG_BAO_CONFIG.MIN_ATTESTATION_INCLUSION_DELAY + 1,
         )
         assert set([a1, a2, a3]) == set(ready_attestations)
+
+        ready_attestations = bob_recv_server.get_ready_attestations(
+            attesting_slot + XIAO_LONG_BAO_CONFIG.SLOTS_PER_EPOCH + 1,
+        )
+        assert set([a3]) == set(ready_attestations)
--- a/tests/plugins/eth2/beacon/test_receive_server.py
+++ b/tests/plugins/eth2/beacon/test_receive_server.py
@@ -20,9 +20,6 @@ from p2p.peer import (
     MsgBuffer,
 )
 
-from eth.constants import (
-    ZERO_HASH32,
-)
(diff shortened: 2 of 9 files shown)


trinity/exceptions.py
```diff
@@ -68,6 +68,13 @@ class BadDatabaseError(BaseTrinityError):
     pass
 
 
+class AttestationNotFound(BaseTrinityError):
```
NIC619: Not sure if this is the right place for this exception. cc @pipermerriam @hwwhww 
This exception is used when [`ReceiveServer`](https://github.com/ethereum/trinity/blob/master/trinity/protocol/bcc/servers.py#L261) can not find the `Attestation` in question in its `AttestationPool`.

pipermerriam:
Easy to move at a later date if it is not.

Sent from ProtonMail mobile

trinity/protocol/bcc/servers.py
```diff
@@ -213,6 +219,49 @@ class BaseReceiveServer(BaseRequestServer):
...
+        raise AttestationNotFound(f"No attestation with root {attestation_root} is found.")
+
+    def get_all(self) -> Tuple[Attestation, ...]:
+        return tuple(self._pool)
+
+    def add(self, attestations: Iterable[Attestation]) -> None:
+        for attestation in attestations:
+            if attestation in self._pool:
+                return
+            self._pool.add(attestation)
```
ChihChengLiang: This method is weird. I suppose the intention is to add all of the attestations to `self._pool` if none of the attestations is in it. But in this method, if called with 10 attestations and the 3rd one is detected already in `self._pool`, then the first two attestations are already added and the rest 8 attestations are discarded.

trinity/protocol/bcc/servers.py
```diff
@@ -352,8 +407,11 @@ def _is_attestation_new(self, attestation: Attestation) -> bool:
         """
         Check if the attestation is already in the database or the attestion pool.
         """
-        # stub
-        return False
+        try:
+            self.attestation_pool.get(attestation.root)
+            return not self.chain.attestation_exists(attestation.root)
+        except (AttestationNotFound, AttestationRootNotFound):
```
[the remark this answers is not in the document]
NIC619: Good catch!

trinity/plugins/eth2/beacon/validator.py
```diff
@@ -98,6 +100,7 @@ def __init__(
             peer_pool: BCCPeerPool,
             validator_privkeys: Dict[ValidatorIndex, int],
             event_bus: TrinityEventBusEndpoint,
+            get_ready_attestations_fn: Callable[[Slot], Sequence[Attestation]],
```
mhchia: Is this function assumed to returns *any* ready attestations right away(i.e. non-blocking)?

trinity/plugins/eth2/beacon/validator.py
```diff
@@ -118,6 +121,7 @@ def __init__(
                 Epoch(-1),
                 CommitteeAssignment((), Shard(-1), Slot(-1), False),
             )
+        self.get_ready_attestations = get_ready_attestations_fn
```
mhchia: To be consistent with the style in this class, could you also annotate this attribute with its type in the class?

trinity/protocol/bcc/servers.py
```diff
@@ -213,6 +217,50 @@ class BaseReceiveServer(BaseRequestServer):
...
+        for attestation in self._pool:
+            if attestation.root == attestation_root:
+                return attestation
+        raise AttestationNotFound(
+            f"No attestation with root {encode_hex(attestation_root)} is found.")
+
+    def get_all(self) -> Tuple[Attestation, ...]:
+        return tuple(self._pool)
+
+    def add(self, attestations: Iterable[Attestation]) -> None:
```
mhchia: For me, it is a little bit confusing that `add` accepts `attestations` as the parameter, instead of `attestation`, since it is an "Attestation" pool. To be consistent with `OrphanBlockPool`, I'd suggest having `add(attestation)`, and add a `batch_add(attestations)`.

tests/eth2/core/beacon/chains/test_beacon_chain.py
```diff
@@ -188,3 +188,4 @@ def test_get_attestation_root(valid_chain,
     # Only one attestation in attestations, so just check that one
     a0 = attestations[0]
     assert valid_chain.get_attestation_by_root(a0.root) == a0
+    assert valid_chain.attestation_exists(a0.root)
```
mhchia: Would you mind adding the `assert not` cases for `get_attestation_by_root` and `attestation_exists`?

tests/plugins/eth2/beacon/test_receive_server.py
```diff
@@ -559,3 +543,97 @@ def _validate_attestations(attestations):
...
+    assert len(pool._pool) == 1
+    # test: add: no side effect for adding twice
+    pool.add([a1])
+    assert len(pool._pool) == 1
+    # test: `__contains__`
+    assert a1.root in pool
+    assert a1 in pool
+    assert a2.root not in pool
+    assert a2 not in pool
+    # test: add: two blocks
```
mhchia:
```suggestion
    # test: add: two attestations
```

trinity/protocol/bcc/servers.py
```diff
@@ -452,6 +513,8 @@ def _try_import_orphan_blocks(self, parent_root: Hash32) -> None:
                 except ValidationError as e:
                     # TODO: Possibly drop all of its descendants in `self.orphan_block_pool`?
                     self.logger.debug("Fail to import invalid block=%s  reason=%s", block, e)
+                    # Remove attestations in block that are also in the attestation pool.
+                    self.attestation_pool.remove(block.body.attestations)
                     pass
```
mhchia:
```suggestion
```

trinity/protocol/bcc/servers.py
NIC619: Added and adjusted `add`, `batch_add`, `remove` and `batch_remove`.

trinity/protocol/bcc/servers.py
```diff
@@ -352,8 +408,11 @@ def _is_attestation_new(self, attestation: Attestation) -> bool:
         """
         Check if the attestation is already in the database or the attestion pool.
         """
-        # stub
-        return False
+        try:
+            self.attestation_pool.get(attestation.root)
```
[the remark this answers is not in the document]
NIC619: Good catch!

trinity/plugins/eth2/beacon/validator.py
NIC619: Yes, currently made as a non-blocking call.

PR #671, opened 2019-05-26
Split isolated plugin base into agnostic async base

pipermerriam:
### What was wrong?

With `trio` coming, we need the `BaseIsolatedPlugin` to not have any `asyncio` specific components.

### How was it fixed?

Made `BaseIsolatedPlugin` a proper abstract base class and implemented `AynscioIsolatedPlugin` which contains all of the `asyncio` specific stuff.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

Cute Baby Hippo (11)

### DIFF
--- a/docs/guides/writing_plugins.rst
+++ b/docs/guides/writing_plugins.rst
@@ -79,7 +79,7 @@ This is the default type of plugin we want to build if:
   runtime (as opposed to e.g. just reading things from the database)
 
 We build this kind of plugin subclassing from
-:class:`~trinity.extensibility.plugin.BaseIsolatedPlugin`.  A detailed example will follow soon.
+:class:`~trinity.extensibility.asyncio.AsyncioIsolatedPlugin`.  A detailed example will follow soon.
 
 
 Plugins that run inside the networking process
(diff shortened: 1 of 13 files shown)


docs/guides/writing_plugins.rst
```diff
@@ -166,9 +166,9 @@ Defining plugins
 
 We define a plugin by deriving from either
 :class:`~trinity.extensibility.plugin.BaseMainProcessPlugin`,
-:class:`~trinity.extensibility.plugin.BaseIsolatedPlugin` or 
+:class:`~trinity.extensibility.asyncio.AsyncioIsolatedPlugin` or
```
cburgdorf: Not to self: this part of the docs is due to an overhaul when the `TrioIsolatedPlugin` lands

trinity/extensibility/asyncio.py
```diff
@@ -0,0 +1,52 @@
...
+        await self.event_bus.announce_endpoint()
+        await self.event_bus.broadcast(
+            PluginStartedEvent(type(self))
+        )
+
+        # Whenever new EventBus Endpoints come up the `main` process broadcasts this event
+        # and we connect to every Endpoint directly
+        asyncio.ensure_future(self.event_bus.auto_connect_new_announced_endpoints())
+
+        self.do_start()
```
cburgdorf: Nothing that we need to fix in this PR but I *think* it would be better if `do_start` was `async` end then we would call `await self.do_start()` here and then inside `spawn_start` instead of using `loop.run_forever` we would use `loop.run_until_complete(self._prepare_start())`. It is then up to the plugin author to decide if `do_start` would ever complete or not.

pipermerriam: Good follow up cleanup but yes, not going to address in this PR

PR #682, opened 2019-05-29
Improve ergonomics, flexibility and consistency of logging

cburgdorf:
### What was wrong?

Configuring the individual loggers is cumbersome (see https://github.com/ethereum/trinity/pull/659#discussion_r287093201)

### How was it fixed?

Let's go though a bunch of examples:

1. Run `trinity` -> stderr prints exactly the same logging output as current master (`INFO` to stderr, `DEBUG` to file as `DEBUG` is the default for file logging) 

2. Run `trinity -l p2p.discovery=DEBUG2` -> stderr + file log print only `INFO` logs except for `p2p.discovery` where `DEBUG2` or higher is shown

3. Run `trinity -l ERROR -l p2p.discovery=DEBUG2` -> stderr + file log print only `ERROR` logs except for `p2p.discovery` where `DEBUG2` or higher is shown

4. Run `trinity --file-log-level DEBUG2 --stderr-log-level INFO` -> stderr + file log print only `INFO` logs but file log prints `DEBUG2` or higher logs

5. Run `trinity --stderr-log-level DEBUG2 --file-log-level INFO` -> stderr prints `DEBUG2` or higher logs but file log prints only `INFO` logs

(The reverse of 4)

Notice that previously the usage of `--log-level` (or `-l`)  without any specified module was a bit inconsistent so that it was taken as default for `--stderr-log-level` but not as default for `--file-log-level`. 

This also meant that previously only the combination of `--log-level` with `--stderr-log-level` was illegal while the combination of `--log-level` with `--file-log-level` was perfectly fine. This was now changed so that both are treated equally and the user showing error message was improved (hopefully!)

One side effect of this is that the behavior for something like `trinity -l ERROR` is now slightly different. Previously this would cause only the stderr logging to reduce to level ERROR but keep file logging on level DEBUG (default). Now, it causes both stderr and file logging to use level ERROR. To get the previous behavior where ERROR is only used for stderr logging while file logging sticks to its default use `trinity --stderr-log-level ERROR` instead.

There were also some minor cleanups and refactoring.

### To-Do

- [x] Clean up commit history
- [x] write tests to proof all the claims this PR makes 

#### Cute Animal Picture

### DIFF
--- a/tests/integration/test_trinity_cli.py
+++ b/tests/integration/test_trinity_cli.py
@@ -17,6 +17,9 @@ from tests.integration.helpers import (
     scan_for_errors,
 )
 
+from trinity.config import (
+    TrinityConfig,
+)
 from trinity._utils.async_iter import (
     contains_all
 )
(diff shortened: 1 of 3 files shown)


pipermerriam:
We should be capable of testing this right?  Any thoughts on taking at least a simple stab at testing the simple cases.

cburgdorf:
Yep, I can add some more test cases to the existing logging tests

https://github.com/ethereum/trinity/blob/afcee339502b6bf422583b8178501000a64ecfe1/tests/integration/test_trinity_cli.py#L167-L181

carver:
> * Run `trinity -l p2p.discovery=DEBUG2` -> stderr + file log print only `INFO` logs except for `p2p.discovery` where `DEBUG2` or higher is shown

I would expect that command to (continue to) print `DEBUG` to the filelog, instead of `INFO`. Is that a typo?

Would `p2p.discovery` `DEBUG2` logs show in both stderr and the file log? I think that's what I would expect based on this command. The only way I can see to avoid this confusion is to have specific flags, something like `-lf` (file) and `-ls` (stderr), so that we could have  `trinity -lf p2p.discovery=DEBUG2` to avoid blasting `DEBUG2` to stderr.

> * Run `trinity --file-log-level DEBUG2 --stderr-log-level INFO` -> stderr + file log print only `INFO` logs but file log prints `DEBUG2` or higher logs

Looks like a typo. Should be "stderr ~~+ file log~~ print**s** only `INFO` logs but file log prints `DEBUG2` or higher logs", right?

pipermerriam:
This has me thinking that we need a table in our documentation that describes what will be printed where in a set of common use cases.  Maybe we can first focus on that table and then ensure that the flags cause the loggers to comply.

One guiding principle that I think is reasonable to maintain.  Absent `--file-log-level` the file logs will be a superset of the stderr logs.

I'll assert this is my expectation of how this behaves.

1. `--log-level DEBUG`
    - stderr: all logs `>=DEBUG`
    - file: all logs `>=DEBUG` (no change)
2. `--log-level DEBUG2`
    - stderr: all logs `>=DEBUG2`
    - file: all logs >=DEBUG2`
3. `--log-level p2p=DEBUG2`
    - stderr: all logs `>=INFO` and anything under `p2p` namespace if `>=DEBUG2`
    - file: all logs `>=DEBUG` and anything under `p2p` namespace if `>=DEBUG2`
4. `--log-level ERROR --log-level p2p=DEBUG
    - stderr: all logs `>=ERROR` and anything under `p2p` namespace if `>=DEBUG`
    - file: all logs `>=DEBUG` and anything under `p2p` namespace if `>=DEBUG`

I'll point out that this is a prime issue for bike shedding so I'm going to refrain from any further discussions on what the *right* behavior is and say that as long as it's documented and remotely resembles this I'm :+1: and fine to have this move forward (I wouldn't mind there being some tests since this area of functionality hasn't exactly always worked as expected).

tests/integration/test_trinity_cli.py
```diff
@@ -165,20 +168,99 @@
...
+            ('trinity', '--stderr-log-level=ERROR', '--file-log-level=DEBUG2',),
+            {},
+            {'Started main process', 'DiscoveryProtocol  >>> ping'},
+            {'Started main process', 'DiscoveryProtocol  >>> ping'},
+            {},
+        ),
+        # This one doesn't work yet but should
+        # (
+        #     ('trinity', '-l=ERROR', '-l', 'p2p.discovery=DEBUG2'),
+        #     {'DiscoveryProtocol  >>> ping'},
```
cburgdorf: @pipermerriam This case doesn't work yet. In other words, whitelisting a specific module to a lower level than the `stderr` or `file` does only work for the stderr logs but not for the file log. So if we run `trinity -l ERROR -l p2p.discovery=DEBUG2` we can observe the discovery logs in stdout but in the file log.

The reason for that (I think) is that with the current approach we just create a dedicated handler for the `p2p.discovery` logger that will be hooked up to `stderr` directly. I've spend some time trying to apply this to the file logging but haven't come up with something that worked yet.

If nobody knows a solution to this, my preferred way of dealing with it would be to turn this commented test case into an issue to come back to it later.

I think the PR is a net positive so far especially as being able to run `trinity -l ERROR -l 'some.module=DEBUG2` greatly improves debugging already.

pipermerriam: I'm :+1: to push this one down the road and track with an issue.  Seems like we're still at a net improvement to the situation.

cburgdorf: Added a tracking issue here: https://github.com/ethereum/trinity/issues/689
Mind giving the PR a review / approval?

PR #684, opened 2019-05-30
[WIP] extract Peer and Protocol concepts.

pipermerriam:
fixes https://github.com/ethereum/trinity/issues/90

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


pipermerriam:
Brain dump.

I've been spinning on this all day and didn't really get much of anywhere, but I think I know why.

I've been trying to model the connection abstraction as something we create and hold onto rather than something ephemeral.  If you look at the picture, the "connection abstraction" is needed to establish the remote capabilities because it's needed for the base `p2p` handshake (which is where we learn about remote capabilities).  We then need the connection abstraction *again* to do all of the sub-protocol handshakes, but this time it will be more aware of the additional protocols it needs to abstract.

Because of this, it looks like we need to first, create one that is only abstracting the base `p2p` protocol and stream the messages **just** for the `p2p` handshake.  Then we discard it and make another that spans all of the protocols we expect to speak.  This *new* one can also be ephemeral depending on the API but the important part is that we can't hold onto the same one across the base `p2p` handshake and all of the sub protocol handshakes without introducing mutation into the API to *add* the sub protocols after the initial `p2p` handshake which is something I'm very :-1: on doing.

We'll see if this "eureka" moment proves to be valid next week when I pick this back up and try to run with it again under this new set of assumptions about how the architecture lines should be drawn.

signal-attachment-2019-05-31-161440

pipermerriam:
@carver I'd be interested in very high level review of the `p2p.connection` module.  I just went in and commented the crap out of it since I had such a hard time figuring out the architecture and don't want to have to figure it all out again later.  Some contextual notes:

-I think that the `establish_protocol_connections` function is probably close to it's final form and is the primary API that the module exposes.
- The `Connection` class is very likely to go through some flux, or maybe be wrapped up by something else that looks more like a service...  I don't know yet.  I do know that there's still code to be written to standardize the interface for how external callers are expected to get at the individual protocols and interact with them.

I *think* that I may be able to package all of this up and keep the current external API of the peer class mostly intact so that we can get this in place and then do the dis-assembly of the `Peer` concept in another step, or even just implement a brand new interface and phase the `Peer/PeerPool` out for something completely new...?  dunno yet.

p2p/connections.py
```diff
@@ -0,0 +1,387 @@
...
+        """
+        Stream the messages for the specified protocol.
+        """
+        lock, msg_queue = self._parking_lot[protocol.name]
+        if lock.locked():
+            raise Exception(f"Streaming lock for {protocol} is not free.")
+        async with lock:
+            while not self._transport.is_closing:
+                # First try to empty the queue
+                yield await msg_queue.get()
```
carver: Looks like this ought to cancel out if `self._token` is triggered.

p2p/connections.py
```diff
@@ -0,0 +1,387 @@
...
+            token = CancelToken('Connection')
+
+        self._transport = transport
+        # the base `p2p` protocol instance.
+        self._base_protocol = base_protocol
+        self._protocols = protocols
+        self._token = token
+        self._stream_lock = asyncio.Lock()
+        self._parking_lot = {
+            protocol.name: (asyncio.Lock(), asyncio.Queue())
```
carver: Maybe could add some maximum size to the queue so we notice if some sub-protocol stops reading messages, rather than just balloon the ram with incoming messages.

p2p/connections.py
```diff
@@ -0,0 +1,387 @@
...
+
+        if cmd_id < base_protocol.cmd_length:
+            msg_proto = base_protocol
+        else:
+            for protocol in protocols:
+                if cmd_id < protocol.cmd_id_offset + protocol.cmd_length:
+                    msg_proto = protocol
+                    break
+            else:
+                raise UnknownProtocolCommand(f"No protocol found for cmd_id {cmd_id}")
```
carver: We'll probably want some more info, like the list of protocols were we choosing from.

p2p/connections.py
```diff
@@ -0,0 +1,387 @@
...
+        cmd = msg_proto.cmd_by_id[cmd_id]
+
+        msg = cast(Dict[str, Any], cmd.decode(raw_msg))
+        # NOTE: This used to be a bottleneck but it doesn't seem to be so anymore. If we notice
+        # too much time is being spent on this again, we need to consider running this in a
+        # ProcessPoolExecutor(). Need to make sure we don't use all CPUs in the machine for that,
+        # though, otherwise asyncio's event loop can't run and we can't keep up with other peers.
+
+        yield msg_proto, cmd, msg
+        await asyncio.sleep(0)
```
carver: Was this just left in for testing, or is it actually necessary? If the latter, a comment would be good.

tests/p2p/test_connection.py
```diff
@@ -0,0 +1,57 @@
...
+
+    alice_result, bob_result = await asyncio.gather(alice_coro, bob_coro)
+
+    alice_connection, alice_protocol_pairs = alice_result
+    bob_connection, bob_protocol_pairs = bob_result
+
+    assert len(alice_protocol_pairs) == 2
+    assert len(bob_protocol_pairs) == 2
+
+    alice_p2p_receipt, alice_paragon_receipt = alice_result
```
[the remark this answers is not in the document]
pipermerriam: This is a typo.  This should be `receipt_1, receipt_2 = alice_receipts`  which are the receipts returned by each of the handshakes.

p2p/connections.py
```diff
@@ -0,0 +1,387 @@
...
+        yield capabilities_by_name[name][0]
+
+
+async def _do_p2p_handshake(transport: Transport,
+                            capabilites: CapabilitiesType,
+                            p2p_handshake_params: DevP2PHandshakeParams,
+                            base_protocol: P2PProtocol,
+                            token: CancelToken) -> P2PProtocol:
+    client_version_string, listen_port = p2p_handshake_params
+    base_protocol.send_handshake(client_version_string, capabilites, listen_port)
```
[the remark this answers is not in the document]
pipermerriam: it could probably know about them, I'll see.

pipermerriam:
Making steady forward progress on this now.  Base concept seems sound at this point and I'm working my way up through the `Peer -> PeerPool -> Server` adjusting how everything works.  I went a little further than I think I had to on this but it's looking a lot like I'll be able to fully drop all of the custom `Peer` and `PeerPool` classes at once.

The concept of `PeerContext` and `PeerFactory` are now gone and the way things are looking, I'm growing convinced that the architecture lines are drawn in much better places.  There may be one more significant thing to be done (which I'll probably delay as a second step) which is splitting the existing `PeerPool` into two components, one which is dumb, not a service and probably deals with `Connection` objects, and another which is *smart*, is a service, and manages adding things like establishing new connections, limiting total connections, etc.

p2p/tools/factories.py
```diff
@@ -95,11 +103,83 @@ def from_seed(cls, seed: bytes, *args: Any, **kwargs: Any) -> discovery.Discover
         return cls(*args, privkey=privkey, **kwargs)
 
 
+async def TransportPairFactory(alice_remote: kademlia.Node = None,
```
pipermerriam: This should be part of #755

p2p/tools/factories.py
```diff
@@ -123,3 +203,140 @@ def MemoryTransportPairFactory(alice_remote: kademlia.Node = None,
...
+
+
+_protocol_name_iter = _protocol_name_enumerator()
+
+
+def ProtocolNameFactory():
+    return next(_protocol_name_iter)
+
+
+def ProtocolFactory(name: str = None,
```
pipermerriam: These factories can be pulled to their own PRs

pipermerriam:
Just a note for self about the status of this.

The concepts for the `Connection` and `Multiplexer` objects and the overall `Handshake` logic are all sound and tested decently at this point.  In progress is updating the codebase to basically gut the `Peer` objects down to their bones and get the tests working again.  I've got a lot of in-progress work that will be merged incrementally that improves `p2p.tools.factories` and creates `trinity.tools.factories` to have lots and lots of tools for generating test data.  This comes in conjunction with full removal of a lot of `helpers.py` stuff that's been bugging me for a long time.

pipermerriam:
I think I've got an implementation plan down for this now.

1. Introduce `Multiplexer`: #835 (not integrated into actual peer code)
2. Modify `BasePeer` to use the `Multiplexer` for splitting up peer messages by protocol.
3. Introduce `p2p.handshaker` API (not integrated into actual peer code)
4. Modify `BasePeer` to use handshakers to perform their actual handshakes.
5. Introduce `p2p.connection.Connection` API (not integrated into actual peer code)
6. Modify `BasePeer` to use actual connection.

At this point the peer classes are effectively stripped down to bones and do very little *extra*.  Probably next steps are.

- collapse to a single `Peer` and/or `PeerPool` class
- remove `PeerSubscriber` in favor of using the `Connection` APIs for registering handlers for different protocols/commands.
- implementation of a `ConnectionStore`: A dumb container that houses active connections.
- implementation of a `ConnectionManager`: An active service that manages dial-out
- implementation of a `ConnectionReceiver`: An active service that manages dial-in
- implementation of a `ConnectionPool`: An active service that abstracts network interactions across multiple connections from the `ConnectionStore`
- update of the Request/Response API to be peer agnostic and to deal only with a `Connection` object.
-

pipermerriam:
This is effectively closed with the merge of #962 

Continuation of this effort will now be tracked in #966

PR #686, opened 2019-05-31
Add arg to take Unix integer as genesis time

ChihChengLiang:
### What was wrong?

Now Eth2 network generator only takes seconds delay as the genesis time.

### How was it fixed?

Add a parser argument to take Unix integer as genesis time

#### Cute Animal Picture

### DIFF
--- a/trinity/plugins/eth2/network_generator/plugin.py
+++ b/trinity/plugins/eth2/network_generator/plugin.py
@@ -139,7 +139,7 @@ class NetworkGeneratorPlugin(BaseMainProcessPlugin):
         logger.info("Generating testnet")
         network_dir = args.network_dir
         if len(os.listdir(network_dir)) > 0:
-            logger.error("This directory is not empty, won't create network files here.")
+            logger.error("This directory is not empty, won't create network files here")
             sys.exit(1)
 
         clients = cls.generate_trinity_root_dirs(network_dir)
@@ -161,7 +161,7 @@ class NetworkGeneratorPlugin(BaseMainProcessPlugin):
                       network_dir: Path,
                       clients: Tuple[Client, ...]) -> Dict[Any, Any]:
         logger = cls.get_logger()
-        logger.info(f"Creating {num} validators' keys")
+        logger.info("Creating %s validators' keys", num)
         keys_dir = network_dir / KEYS_DIR
         keys_dir.mkdir()
 
@@ -206,7 +206,8 @@ class NetworkGeneratorPlugin(BaseMainProcessPlugin):
         genesis_time = get_genesis_time()
         print(genesis_time)
         logger.info(
-            f"Genesis time will be {humanize_seconds(genesis_time-int(time.time()))} from now"
+            "Genesis time will be %s from now",
+            humanize_seconds(genesis_time - int(time.time())),
         )
         state = state.copy(
             genesis_time=genesis_time,
--- a/trinity/plugins/eth2/network_generator/plugin.py
+++ b/trinity/plugins/eth2/network_generator/plugin.py
@@ -122,7 +122,7 @@ class NetworkGeneratorPlugin(BaseMainProcessPlugin):
         )
         genesis_time_group.add_argument(
             "--genesis-delay",
-            help="Seconds before genesis time from now",
+            help="Set seconds delay after the genesis state is created as genesis time",
             type=int,
         )
         genesis_time_group.add_argument(
@@ -204,7 +204,6 @@ class NetworkGeneratorPlugin(BaseMainProcessPlugin):
             genesis_time=dummy_time,
         )
         genesis_time = get_genesis_time()
-        print(genesis_time)
         logger.info(
             "Genesis time will be %s from now",
             humanize_seconds(genesis_time - int(time.time())),
--- a/trinity/plugins/eth2/network_generator/plugin.py
+++ b/trinity/plugins/eth2/network_generator/plugin.py
@@ -11,10 +11,14 @@ import sys
 import time
 from typing import (
     Any,
+    Callable,
     Dict,
     Tuple,
 )
 
+from eth_utils import (
+    humanize_seconds,
+)
 from ruamel.yaml import (
     YAML,
 )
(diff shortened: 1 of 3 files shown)


trinity/plugins/eth2/network_generator/plugin.py
```diff
@@ -100,11 +116,19 @@ def configure_parser(cls, arg_parser: ArgumentParser, subparser: _SubParsersActi
             type=int,
             default=100,
         )
-        testnet_generator_parser.add_argument(
+
+        genesis_time_group = testnet_generator_parser.add_mutually_exclusive_group(
+            required=True,
+        )
+        genesis_time_group.add_argument(
             "--genesis-delay",
```
pipermerriam: You could probably simplify this by changing this to simply compute what the equivalent would be in `--genesis-time` and to have it populate that value onto `genesis_time`.

trinity/plugins/eth2/network_generator/plugin.py
```diff
@@ -161,7 +161,7 @@ def generate_keys(cls,
                       network_dir: Path,
                       clients: Tuple[Client, ...]) -> Dict[Any, Any]:
         logger = cls.get_logger()
-        logger.info(f"Creating {num} validators' keys")
+        logger.info("Creating %s validators' keys", num)
         keys_dir = network_dir / KEYS_DIR
```
ralexstokes: i think we are generally preferring `f-strings`, even if it is inconsistent w/ other uses in this module

trinity/plugins/eth2/network_generator/plugin.py
```diff
@@ -206,7 +206,8 @@ def generate_genesis_state(cls,
         genesis_time = get_genesis_time()
         print(genesis_time)
         logger.info(
```
ralexstokes: this seems unrelated to this PR but since I see it here: we want to avoid having hard-coded `print` stmts because it precludes the user from having control over what IO runs when... annoying to be running a test or something and output just starts spilling to the screen imo

trinity/plugins/eth2/network_generator/plugin.py
ChihChengLiang: Ah, the description "Seconds before genesis time from now" should be changed to "Set seconds delay after the genesis state is created as genesis time".
Since the usage of `--genesis-delay` option is for debugging, and the genesis state takes a long time to create, the "delay" actually starts after the latter is done. So for example when debugging in `eth2/beacon/scripts/run_beacon_nodes.py`, the choice of delay can be independent of the time to create the genesis state.

trinity/plugins/eth2/network_generator/plugin.py
ChihChengLiang: Oops, nice catch. The print shouldn't be there.

trinity/plugins/eth2/network_generator/plugin.py
ChihChengLiang: The logger might be an exception. [The manual](https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#formatting-log-messages) favors the `%` style format in logger so that the log aggregator like Sentry can correctly group the logs.

ralexstokes: yeah i would be interested to know the rationale there....

if we want to refer to the manual, i think it may just be a bit out of date on this :)

https://github.com/ethereum/snake-charmers-tactical-manual/issues/18

trinity/plugins/eth2/network_generator/plugin.py
```diff
@@ -115,12 +139,19 @@ def run_generate_testnet_dir(cls, args: Namespace, trinity_config: TrinityConfig
...
 
         clients = cls.generate_trinity_root_dirs(network_dir)
         keymap = cls.generate_keys(args.num, network_dir, clients)
-        cls.generate_genesis_state(args.genesis_delay, network_dir, keymap, clients)
+
+        get_genesis_time = (
+            get_genesis_time_from_constant(args.genesis_time)
+            if args.genesis_time is not None
+            else get_genesis_time_from_delay(args.genesis_delay)
+        )
```
NIC619: Should we consider the case where neither of `genesis_time` or `genesis_delay` is provided?(throw an exception etc.)

ChihChengLiang: Haha, that's where `testnet_generator_parser.add_mutually_exclusive_group` comes in. It requires exactly one arg must be provided.

pipermerriam: Are you sure about that.  Mutually exclusive doesn't mean the same as require at least one.  I'm pretty sui you can leave both of these out and the parser won't complain.

ChihChengLiang: It will, and I can demonstrate it with a test https://github.com/ethereum/trinity/pull/686/commits/aa85730a54e70c53dfba53980fc572802c6df608

trinity/plugins/eth2/network_generator/plugin.py
carver: Yup, the logger is the only exception I know of to this guideline (as long as the repo is 3.6+, of course)

ralexstokes: huh makes sense, TIL :) scratch what i said

trinity/plugins/eth2/network_generator/plugin.py
```diff
@@ -156,7 +187,7 @@ def generate_keys(cls,
 
     @classmethod
     def generate_genesis_state(cls,
-                               genesis_delay: Second,
+                               get_genesis_time: Callable[[], Timestamp],
```
pipermerriam: Reviewing this a second time I'm a little confused by this API.  Can you explain why we don't/can't just pass in a firm timestamp?

ChihChengLiang: There are two possibilities for `get_genesis_time`: the one produced with `get_genesis_time_from_constant()`and the other by `get_genesis_time_from_delay()`. The reason to have a `get_genesis_time` interface is beacause for the latter case, the genesis time is determined after the create_mock_genesis finished.

An alternative could be 
```python
def generate_genesis_state(..., genesis_time, genesis_delay=None):
    state, _ = create_mock_genesis(
            ...
            genesis_time=genesis_time if genesis_time is not None else 0,
        )
    if genesis_delay is not None:
        state = state.copy(genesis_time=Timestamp(int(time.time()) + genesis_delay))
    ...
```

pipermerriam:
Is this all to account for the time it takes to generate the genesis state?

Sent from ProtonMail mobile

ChihChengLiang:
Exactly

Issue #678, opened 2019-05-28
get_head_state and get_state_by_slot API in BeaconChain object


ChihChengLiang:
### What is wrong?

In beacon chain, not every slot has a block. But the current way to get the state is based on blocks.

`chain.get_state_machine().state` gets the state of `head_block`, but not the state of the head slot.

When a validator proposes a block in a new slot, it requires state at `current_slot - 1` and head_block(not necessary at `current_slot - 1`) as parent_block.

What we need is the `slot_to_state_lookup` and a slot based `chain.get_state_machine(slot)`.

The new API is also an opportunity to refactor the existing code. For example, the following lines, we used to creating a block to initiate the state machine, and the reason is just to get the desired state_root. With the `slot_to_state_lookup` this can be simplified.

https://github.com/ethereum/trinity/blob/6040ed8f3db458a10f61cc8b8f9e8db891e5ed4d/eth2/beacon/chains/base.py#L395-L399

### How can it be fixed

Add the following methods.

- chaindb._add_slot_to_state_lookup()
- chain.get_state_machine(slot)
- chain.get_state(at_slot=slot)

ChihChengLiang:
Fixed in #688

PR #688, opened 2019-06-01
Make `BeaconChain.get_state_machine`/`get_state` by slot

NIC619:
### What was wrong?
#678
Summary: there might be slots during which no block is produced but the state still needs to be updated accordingly(`apply_state_transition_without_block`).

This results in two major changes: 
- 1. `get_state_machine` by slot instead of block
    - add `get_state_by_slot`
- 2. keep track of `head_state_slot` in database and update it whenever state is updated during slot transition
    - if no slot is provided to `get_state_machine()`, `head_state_slot` is assumed
    - open to other naming for `head_state_slot`

### How was it fixed?
- In `BeaconChainDB`
    - [x] add `_add_slot_to_state_root_lookup`
    - [x] add `get_state_by_slot`
        - [x] raise `StateNotFound` if lookup failed
    - [x] add `_add_head_state_slot_lookup`
    - [x] add `get_head_state_slot`
        - [x] raise `HeadStateSlotNotFound` if lookup failed
- In `BeaconChain`
    - [x] add `get_state_by_slot`
    - [x] change `ensure_block` to `ensure_slot`
- In `BeaconStateMachine`
    - [x] `__init__` takes in slot instead of block
    - [x] get `state` by slot instead of block

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

S__15630352


eth2/beacon/chains/base.py
```diff
@@ -403,11 +418,15 @@ def import_block(
...
-        )
-        state, imported_block = self.get_state_machine(base_block_for_import).import_block(block)
+        try:
+            head_state_slot = self.chaindb.get_head_state_slot()
+            if head_state_slot >= block.slot:
+                prev_state_slot = parent_block.slot
+            else:
+                prev_state_slot = head_state_slot
+        except HeadStateSlotNotFound:
+            prev_state_slot = parent_block.slot
```
ChihChengLiang: Not sure how HeadStateSlotNotFound works here. When chaindb is initialized in `BeaconChain.from_genesis`, it first persists a genesis state and later persists a genesis block. So ...
1. The HeadStateSlotNotFound error should only be raised when persisting the genesis state.
2. When HeadStateSlotNotFound is raised, there is no parent_block to look for. Or even why would HeadStateSlotNotFound be raised in import_block in the first place?

eth2/beacon/db/chain.py
```diff
@@ -627,6 +637,63 @@ def _add_block_root_to_slot_lookup(db: BaseDB, block: BaseBeaconBlock) -> None:
...
+        Raises StateNotFound if it is not present in the db.
+        """
+        slot_to_state_root_key = SchemaV1.make_slot_to_state_root_lookup_key(
+            slot
+        )
+        try:
+            state_root_ssz = db[slot_to_state_root_key]
+        except KeyError:
+            raise StateNotFound(
+                "No state root for slot #{0}".format(slot)
```
ChihChengLiang:
```suggestion
                f"No state root for slot #{slot}"
```

tests/plugins/eth2/beacon/test_receive_server.py
```diff
@@ -618,6 +618,10 @@ def test_attestation_pool(mock_attestation):
         )
         bob_recv_server.attestation_pool.batch_add([a1, a2, a3])
 
+        # Workaround: add a fake head state slot
+        # so `get_state_machine` wont's trigger `HeadStateSlotNotFound` exception
```
hwwhww: typo?
```suggestion
        # so `get_state_machine` won't trigger `HeadStateSlotNotFound` exception
```

eth2/beacon/chains/base.py
NIC619: Yeah, this makes sense. Add some comments and remove the `HeadStateSlotNotFound` exception handling.

eth2/beacon/db/exceptions.py
```diff
@@ -5,6 +5,20 @@ class BeaconDBException(Exception):
...
 
 
+class HeadStateSlotNotFound(BeaconDBException):
+    """
+    Exception raised if head state slot does not exist.
+    """
+    pass
+
+
+class StateNotFound(BeaconDBException):
```
hwwhww: Calling it `StateSlotNotFound` may be more accurate since there's also `StateRootNotFound`.

eth2/beacon/db/schema.py
```diff
@@ -39,8 +52,27 @@ def make_finalized_head_root_lookup_key() -> bytes:
     def make_justified_head_root_lookup_key() -> bytes:
         pass
 
+    #
+    # Attestaion
+    #
+    @staticmethod
+    @abstractmethod
+    def make_attestation_root_to_block_lookup_key(attestaton_root: Hash32) -> bytes:
```
ralexstokes: rebase? i feel like this is from another PR, no?

NIC619: Ah, `make_attestation_root_to_block_lookup_key` was missing in `BaseSchema` so I added it in.

PR #690, opened 2019-06-04
Prepare Lahja upgrade

cburgdorf:
### What was wrong?

Trinity is on a git ref for `lahja` and needs to upgrade to `0.14.0` 

### How was it fixed?

1. In previous lahja versions connections where only one-way and we had to manually connect both ways. Lahja now handles the reverse connection on its own which leads to fewer manual connections.

2. It also leads to a new announcement scheme in general where every endpoint only connects to every other endpoint that appears *after* its own position in the centrally planned and maintained endpoint set. This ensures that no duplicate connection attempts can happen.

3. Upgraded to new APIs that let us wait on subscriptions

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/network-db/test_connection_tracker_server.py
+++ b/tests/core/network-db/test_connection_tracker_server.py
@@ -34,7 +34,7 @@ async def test_connection_tracker_server_and_client(event_loop, event_bus):
     bus_tracker = ConnectionTrackerClient(event_bus, config=config)
 
     # Give `bus_tracker` a moment to setup subscriptions
-    await event_bus.wait_until_all_connections_subscribed_to(ShouldConnectToPeerRequest)
+    await event_bus.wait_until_all_endpoints_subscribed_to(ShouldConnectToPeerRequest)
     # ensure we can read from the tracker over the event bus
     assert await bus_tracker.should_connect_to(remote_a) is False
 
--- a/tests/integration/test_trinity_cli.py
+++ b/tests/integration/test_trinity_cli.py
@@ -131,9 +131,9 @@ async def test_web3(command, async_process_runner):
         "Started DB server process",
         "Started networking process",
         "IPC started at",
-        # Ensure we do not start making requests that depend on the networking
-        # process, before it is connected to the JSON-RPC-API
-        "EventBus Endpoint networking connecting to other Endpoint bjson-rpc-api",
+        # Ensure we do not start making requests before Trinity is ready.
+        # Waiting for fast-sync to start gives us a reliable starting point
+        "Starting fast-sync",
     })
 
     attached_trinity = pexpect.spawn('trinity', ['attach'], logfile=sys.stdout, encoding="utf-8")
--- a/trinity/endpoint.py
+++ b/trinity/endpoint.py
@@ -2,6 +2,7 @@ import asyncio
 from typing import (
     Tuple,
 )
+
 from lahja import (
     BroadcastConfig,
     ConnectionConfig,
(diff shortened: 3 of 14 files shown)


cburgdorf:
Ok, looks like this is blocked on us making a decision here https://github.com/ethereum/lahja/pull/105#issuecomment-498624870

cburgdorf:
@pipermerriam This is tracking lahja/master and is :green_heart: I know you have some other PRs that may shift around the plugin logic etc so we may end up with different code. The purpose of this PR is mostly to catch unknown unknowns before we cut a new Lahja version but it is looking :ok_hand: so far.

tests/core/network-db/test_connection_tracker_server.py
```diff
@@ -34,7 +33,7 @@
     bus_tracker = ConnectionTrackerClient(event_bus, config=config)
 
     # Give `bus_tracker` a moment to setup subscriptions
-    await event_bus.wait_until_all_remotes_subscribed_to(ShouldConnectToPeerRequest)
+    await asyncio.sleep(0.01)
```
cburgdorf: Oh the irony...the reason why lahja is on gitref in trinity/master is because we were not settled on how the subscription wait APIs should look like at the when the `RequestServer` PR was finished. :sweat_smile:

pipermerriam:
I think my commit should be dropped, we should get one of https://github.com/ethereum/lahja/pull/121 or https://github.com/ethereum/lahja/pull/122 merged and do this against a new `lahja==0.14.0` release.  I was getting regular timeouts in the places that I tried to work-around the "wait for subscriptions" with sleeps and I think we need the proper wait APIs.

trinity/endpoint.py
```diff
@@ -34,18 +38,18 @@ def request_shutdown(self, reason: str) -> None:
...
-                continue
-            elif self.is_connected_to(connection_config.name):
+        # We only connect to Endpoints that appear after our own Endpoint in the set.
+        # This ensures that we don't try to connect to an Endpoint while that remote
+        # Endpoint also wants to connect to us.
+        endpoints_to_connect_to = reduce(
+            lambda a, b: a + (b,) if b.name == self.name or len(a) > 0 else (),
+            ev.available_endpoints,
+            ()
+        )
```
cburgdorf: @pipermerriam all I want here is to filter the endpoints to only contain those that appear *after* the one that matches our own name. And I wanted to do that without having to rely on a global `after_self` variable but it's all ugly and I feel as if I'm missing some trick. 

I also tried to break it down and first get the index of our element like so:

```python
next(i for i, e in enumerate(ev.available_endpoints) if e.name == self.name)
```

But if we aren't in this set (yet) `next` will throw which makes this solution ugly, too.

I'm ok to go back to the imperative solution (previous commit) if there simply is no elegant way to do this. I feel that my `reduce` scheme is rather ugly and the only reason I show it is to punish myself for not coming up with something more clever :sweat_smile:

tests/integration/test_trinity_cli.py
```diff
@@ -131,9 +131,9 @@
         "Started DB server process",
         "Started networking process",
         "IPC started at",
-        # Ensure we do not start making requests that depend on the networking
-        # process, before it is connected to the JSON-RPC-API
-        "EventBus Endpoint networking connecting to other Endpoint bjson-rpc-api",
```
cburgdorf: Relying on this message isn't safe anymore because who connects to whom is non deterministic  now.

trinity/endpoint.py
```diff
@@ -81,17 +93,6 @@ def track_and_propagate_available_endpoints(self) -> None:
...
-            # In a perfect world, we should only reach this code once for every endpoint.
-            # However, we check `is_connected_to` here as a safe guard because theoretically
-            # it could happen that a (buggy, malicious) plugin raises the `EventBusConnected`
-            # event multiple times which would then raise an exception if we are already connected
-            # to that endpoint.
-            if not self.is_connected_to(ev.connection_config.name):
-                self.logger.info(
-                    "EventBus of main process connecting to EventBus %s", ev.connection_config.name
-                )
-                await self.connect_to_endpoints(ev.connection_config)
```
cburgdorf: Every endpoint connects to `main` and the reverse connection is a thing that lahja does automatically now.

trinity/endpoint.py
cburgdorf: Ok, turns out list comprehension does the trick and I was just being stupid :sweat_smile:

trinity/endpoint.py
```diff
@@ -35,18 +36,24 @@ def request_shutdown(self, reason: str) -> None:
     async def connect_to_other_endpoints(self,
                                          ev: AvailableEndpointsUpdated) -> None:
 
-        for connection_config in ev.available_endpoints:
-            if connection_config.name == self.name:
-                continue
-            elif self.is_connected_to(connection_config.name):
+        # We only connect to Endpoints that appear after our own Endpoint in the set.
+        # This ensures that we don't try to connect to an Endpoint while that remote
+        # Endpoint also wants to connect to us.
```
pipermerriam: I was realizing this is a case that we should figure out how to handle more gracefully.

cburgdorf: Do you mean, Lahja should be less strict about endpoints trying to connect when they are already connected? Or do you mean our announcement ceremony should be different.

Because with this change, the announcement ceremony seems pretty solid to me because duplicates simply can not occur afaik.

pipermerriam: I'm fine with this, I was just pointing out that it would be ideal if we didn't have to think about this type of thing and it just worked.

PR #692, opened 2019-06-04
Update sync tests to latest VM, where possible

carver:
### What was wrong?

Was testing sync against Byzantium.

### How was it fixed?

- Update most of the sync tests to run against Petersburg
- Created a new test chain that will be easier to update to the latest VM

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #694, opened 2019-06-04
Tool to build custom fixture ldb

carver:
### What was wrong?

I needed state that churns in an interesting way on a valid chain, in order to test Beam Sync. This required a lot of manual work, and I didn't want to lose the mechanism that built the data. Also, it's useful to have some examples around for how to go about building more fixture data.

### How was it fixed?

- added the fixture with the chain that churns accounts & storage (to be used for beam sync test)
- added the tools used to generate the above fixture
- moved the integration tool to the `integrations/` dir (was in `core/`)
- added a couple random async method passthroughs that were missing

#### Cute Animal Picture


carver:
Had an error in `py37-eth1-core` that seems unrelated:
```
=================================== FAILURES ===================================
___________________ test_peer_pool_answers_connect_commands ____________________
[gw1] linux -- Python 3.7.3 /home/circleci/repo/.tox/py37-eth1-core/bin/python

event_loop = <uvloop.Loop running=False closed=False debug=False>
event_bus = <trinity.endpoint.TrinityEventBusEndpoint object at 0x7f7ba03d1f28>
server = <test_server.ParagonServer object at 0x7f7ba03df320>

    @pytest.mark.asyncio
    async def test_peer_pool_answers_connect_commands(event_loop, event_bus, server):
        # This is the PeerPool which will accept our message and try to connect to {server}
        initiator_peer_pool = ParagonPeerPool(
            privkey=INITIATOR_PRIVKEY,
            context=ParagonContext(),
            event_bus=event_bus,
        )
        asyncio.ensure_future(initiator_peer_pool.run(), loop=event_loop)
        await initiator_peer_pool.events.started.wait()
        async with run_peer_pool_event_server(
            event_bus,
            initiator_peer_pool,
        ):
    
            assert len(server.peer_pool.connected_nodes) == 0
    
            await event_bus.wait_until_any_connection_subscribed_to(ConnectToNodeCommand)
            await event_bus.broadcast(
                ConnectToNodeCommand(RECEIVER_REMOTE),
                TO_NETWORKING_BROADCAST_CONFIG
            )
    
            await asyncio.sleep(0.1)
    
>           assert len(server.peer_pool.connected_nodes) == 1
E           assert 0 == 1
E            +  where 0 = len({})
E            +    where {} = <p2p.tools.paragon.peer.ParagonPeerPool object at 0x7f7ba03df8d0>.connected_nodes
E            +      where <p2p.tools.paragon.peer.ParagonPeerPool object at 0x7f7ba03df8d0> = <test_server.ParagonServer object at 0x7f7ba03df320>.peer_pool

event_bus  = <trinity.endpoint.TrinityEventBusEndpoint object at 0x7f7ba03d1f28>
event_loop = <uvloop.Loop running=False closed=False debug=False>
initiator_peer_pool = <p2p.tools.paragon.peer.ParagonPeerPool object at 0x7f7ba03cb8d0>
server     = <test_server.ParagonServer object at 0x7f7ba03df320>

tests/core/p2p-proto/test_server.py:187: AssertionError
------------------------------ Captured log setup ------------------------------
WARNING  06-04 22:22:29             server.py  Running with no bootstrap nodes
------------------------------ Captured log call -------------------------------
ERROR  06-04 22:22:29             plugin.py  Task exception was never retrieved
future: <Task finished coro=<<async_generator_asend without __name__>()> exception=StopAsyncIteration()>
StopAsyncIteration
   ERROR  06-04 22:22:29             plugin.py  Task exception was never retrieved
future: <Task finished coro=<<async_generator_asend without __name__>()> exception=StopAsyncIteration()>
StopAsyncIteration
---------------------------- Captured log teardown -----------------------------
ERROR  06-04 22:22:29             plugin.py  Task exception was never retrieved
future: <Task finished coro=<<async_generator_asend without __name__>()> exception=StopAsyncIteration()>
StopAsyncIteration
```

tests/integration/integration_fixture_builders.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def _churn_storage_once(chain, contract_addr, nonce):
+    # invoke function shuffle the storage, with a width of 64
+    # shuffle(uint256)
+    func_id = decode_hex("0xef6537b5")
+    # width is how many different storage slots to fill before wrapping
+    width = b'\x40'
+    # join call with parameters and pad:
+    method_invocation = func_id + width.rjust(32, b'\0')
```
cburgdorf: If we follow my recommendation to put the contracts into `sol` files and generate the interface files, we could use web3 here (after all, Trinity depends on web3 anyway) and make that less cryptic.

That being said, I appreciate the low level code as a learning experience :smile:

trinity/chains/coro.py
```diff
@@ -8,6 +8,7 @@
 class AsyncChainMixin(BaseAsyncChain):
     coro_get_block_by_hash = async_method('get_block_by_hash')
     coro_get_block_by_header = async_method('get_block_by_header')
+    coro_get_block_header_by_hash = async_method('get_block_header_by_hash')
```
cburgdorf: Not for this PR but we could catch these things if we add a test similar to this:

https://github.com/ethereum/trinity/blob/dab778747550c9a8e1ae7ba5339bb4f8685f16ff/tests/core/proxies/test_proxy_implementation.py#L15-L21

tests/core/integration_test_helpers.py
```diff
@@ -165,6 +166,11 @@ class DBFixture(Enum):
     twenty_pow_headers = '20pow_headers.ldb'
     thousand_pow_headers = '1000pow_headers.ldb'
 
+    # this chain updates and churns storage, as well as creating a bunch of
+    # contracts that are later deleted. It was built with:
+    # build_pow_churning_fixture(db, 128)
+    state_churner = 'churn_state.ldb'
```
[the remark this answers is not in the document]
pipermerriam: Seconding the nicety of having some known *metadata* of the expected chain structure.

tests/integration/integration_fixture_builders.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def build_pow_fixture(write_db, num_blocks=20):
+    chain = load_mining_chain(write_db)
+    for i in range(num_blocks):
+        tx = chain.create_unsigned_transaction(
+            nonce=i,
+            gas_price=1234,
+            gas=123400,
+            to=RECEIVER.public_key.to_canonical_address(),
```
pipermerriam: This should probably be computed outside the loop to prevent the crypto and `keccak` operation overhead.

tests/integration/integration_fixture_builders.py
```diff
@@ -0,0 +1,193 @@
...
+                return;
+            }
+        }
+        balances[0] += 1;
+    }
+
+    function delete_index(uint index) public {
+        delete balances[index];
+    }
+}
```
pipermerriam: @njgheorghita could you help explore a way for @carver to easily package these up as packages?  That way we can *preserve* the parity between source and compilation output and not have them get out-of-sync as well as cleaning up this file to not need the sources or bytecode to be inlined.

tests/integration/integration_fixture_builders.py
pipermerriam: This is another place that using ethpm packages would simplify things since you could use the packaged ABI + `eth_utils` to generate the function signatures.  Web3.py is a bit harder to use since it only knows how to interact with the chain over JSON-RPC.

Not for this PR, but a native translation layer for `Web3.py` to interact directly with a `Chain` instance might be cool to create at some point.

cburgdorf: >This is another place that using ethpm packages would simplify things since you could use the packaged ABI + eth_utils

Ah, interesting!

>Web3.py is a bit harder to use since it only knows how to interact with the chain over JSON-RPC.

You *can* use it without JSON-RPC using the `build_transaction` API. We do the same in the benchmarks.

https://github.com/ethereum/py-evm/blob/65e9fd93194336a328fc18157b08e8a2b42f2735/scripts/benchmark/checks/erc20_interact.py#L162

carver: > You _can_ use it without JSON-RPC using the `build_transaction` API. We do the same in the benchmarks.

Hah, good point, it didn't even occur to me for the same reasons Piper said. That seems like a clear improvement.

tests/integration/integration_fixture_builders.py
```diff
@@ -0,0 +1,193 @@
...
+# a contract that exists purely to be deleted later
+"""
+pragma solidity ^0.5.7;
+
+contract Nymph {
+    function poof() public {
+        selfdestruct(msg.sender);
+    }
+}
+"""
```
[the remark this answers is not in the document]
carver: > I read point_up as just being comments to describe the contracts but then later, I have to trust the magic hex data values to actually be that code in the comments, right?

Mostly, although if you compile the code in remix with the same compiler, you should get exactly the same hex code. I :100: agree with the problem of it being difficult to verify, just noting that it's not *impossible* to verify.

> That being said, I realize this is just something in your way to accelerate the beam sync development so if improving this feels like a distraction right now that you'd rather like to avoid how about the following:
> 
> 1. A tracking issue that roughly outlines steps to improve this
> 2. Maybe move everything in this file except `build_pow_fixture` into its own `state_churn`-specific file. (Imho it is not generic enough to live here as it stands)

Yup, you know what is on my mind. :) Especially with the retreat coming up. I'm on board with 99% of this. I think I'd still like to leave one main method for building the fixture in the fixture builders file, and move all the supporting stuff out. Something seems nice about knowing that you can always start in this file to find how all the `DBFixtures` were built.

tests/core/integration_test_helpers.py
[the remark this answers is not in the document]
carver: #698

tests/integration/integration_fixture_builders.py
carver: #700

trinity/chains/coro.py
carver: #697

carver:
> _Maybe_ expose via a simple CLI so that we could test this code using a very small fixture generation size that would be acceptable to run in CI?

I didn't build the CLI part, but added a test that at least makes sure the basic db generation works. (which was helpful during the big `churn_state` refactor, anyway). Good for now?

cburgdorf:
>Good for now?

LGTM. There's one job failing but I guess that is because it needs a rebase after #692 lands

PR #699, opened 2019-06-06
Patch release of alpha.25, to merge in so that it's on master

carver:
bump py-evm, for eth-keys dependency patch

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -130,7 +130,7 @@ with open('./README.md') as readme:
 setup(
     name='trinity',
     # *IMPORTANT*: Don't manually change the version here. Use the 'bumpversion' utility.
-    version='0.1.0-alpha.24',
+    version='0.1.0-alpha.25',
     description='The Trinity client for the Ethereum network',
     long_description=long_description,
     long_description_content_type='text/markdown',
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@
 import os
 from setuptools import setup, find_packages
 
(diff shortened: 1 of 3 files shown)


carver:
Crucially, the docker-image-build-test passed, which is what tipped me off to the dependency issue in the first place.

carver:
So this branch includes the tagged commit that got released. I think the right thing to do is to merge it back into master, which should have pretty minor effects. One bonus is that it bumps the version and folds in the py-evm upgrade. The only "wacky" thing it does is mean that the next version released will not be all based on the previous version. But I think it's better than the alternatives.

Any opposition to merging this in? @pipermerriam @cburgdorf @lithp @ralexstokes

setup.py
```diff
@@ -3,7 +3,7 @@
 import os
 from setuptools import setup, find_packages
 
-PYEVM_DEPENDENCY = "py-evm==0.2.0a43"
+PYEVM_DEPENDENCY = "py-evm==0.3.0a1"
```
cburgdorf: I think we should cherry-pick that into `master` if we don't cut a new release today.

cburgdorf: Oh I just read your comment that you also want it to be merged. Let's do it!

carver: Sweet, I'll wait a few hours to see if anyone has objections.

carver: Actually, since it's holding up #586 I'll probably merge it as soon as I finish clearing my github queue.

PR #704, opened 2019-06-06
Add a maximum batch size for regular sync imports

carver:
Without this maximum batch size, when the first batch takes a long time
to execute, the second batch will be enormous. Perhaps most importantly,
the back-pressure set by `_db_buffer_capacity` is never triggered if the
batch is unlimited size. So the header downloader races ahead unimpeded.

I think this was something that @veox noticed.

Note that beam sync uses the regular syncer updated in this PR, which is why it was originally bundled in #641 .

### To-Do

#### Cute Animal Picture


veox:
> I think this was something that @veox noticed.

Yes, but it seems I never filed an issue.

PR #721, opened 2019-06-14
Split PeerPool and Sync into two isolated processes

cburgdorf:
### What was wrong?

We want `PeerPool` and sync to run in their own separate processes. The theory is that this will help us to maintain more peer connections as it frees the networking code from many CPU intensive tasks that happen during sync. 

### How was it fixed?

Still lots of WIP stuff.

### To-Do

- [ ] Clean up commit history
- [ ] cleanup code
- [ ] ensure this works with light peers
- [x] ensure this works with beam sync
- [ ] fix "gave 0 headers when seeking common meat ancestors from {start_num}"

#### Cute Animal Picture

### DIFF
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -10,6 +10,7 @@ from typing import (
     Callable,
     List,
     Optional,
+    Union,
     cast,
 )
 from weakref import WeakSet
@@ -230,12 +231,18 @@ class BaseService(ABC, CancellableMixin):
 
         self.run_task(_run_daemon_wrapper())
 
-    def call_later(self, delay: float, callback: 'Callable[..., None]', *args: Any) -> None:
+    def call_later(self,
+                   delay: float,
+                   callback: Callable[..., Union[None, Awaitable[None]]],
+                   *args: Any) -> None:
 
         @functools.wraps(callback)
         async def _call_later_wrapped() -> None:
             await self.sleep(delay)
-            callback(*args)
+            if asyncio.iscoroutinefunction(callback):
+                await callback(*args)
+            else:
+                callback(*args)
 
         self.run_task(_call_later_wrapped())
 
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -1,18 +1,19 @@
 import asyncio
 import uuid
-
+from async_generator import asynccontextmanager
 from eth.exceptions import HeaderNotFound
 from lahja import ConnectionConfig, AsyncioEndpoint
 from p2p.service import BaseService
 import pytest
 
-from trinity.protocol.eth.peer import ETHPeerPoolEventServer
+from trinity.protocol.eth.peer import ETHPeer, ETHProxyPeerPool, ETHPeerPoolEventServer
 from trinity.sync.beam.importer import (
     pausing_vm_decorator,
     BlockImportServer,
 )
+from trinity.protocol.eth.servers import ETHRequestServer
 from trinity.protocol.eth.sync import ETHHeaderChainSyncer
-from trinity.protocol.les.peer import LESPeer
+from trinity.protocol.les.peer import LESPeer, LESProxyPeerPool
 from trinity.protocol.les.servers import LightRequestServer
 from trinity.sync.common.chain import (
     SimpleBlockImporter,
@@ -42,6 +43,7 @@ from tests.core.integration_test_helpers import (
     load_mining_chain,
     run_peer_pool_event_server,
     run_request_server,
+    run_proxy_peer_pool,
 )
 from tests.core.peer_helpers import (
     get_directly_linked_peers,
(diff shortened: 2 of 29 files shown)


tests/core/p2p-proto/test_sync.py
```diff
@@ -140,34 +184,32 @@ def small_header_batches(monkeypatch):
 # This test was reduced to a few cases for speed. To run the full suite, use
 # range(1, 130) for beam_to_block. (and optionally follow the instructions at target_head)
 @pytest.mark.asyncio
-@pytest.mark.parametrize('beam_to_block', [1, 66, 68, 129])
+# FIXME
+# @pytest.mark.parametrize('beam_to_block', [1, 66, 68, 129])
+@pytest.mark.parametrize('beam_to_block', [1])
```
cburgdorf: @carver 

I'm still stuck debugging some LES issues but in the mean time I was wondering if you had a tip for me from the top of your head why this beam sync test works for blocks < 7 but not above. I guess there must be some code path that doesn't get triggered for the first few blocks?

cburgdorf:
@carver @pipermerriam This still needs some love (some beam sync tests are commented, some tests are flaky) but in general it works and you might already like to leave some feedback on it.

I feel there is a lot room to further clean things up. Especially all these different events that go in and out of the event server classes. This should be seen as an intermediate step to rip peer pool and sync apart and run in isolation but I think that eventually we want the whole peer pool / protocol stuff to run naively on the event bus (instead of all this mapping) but I guess this might go into the bigger refactoring that Piper is working on.

tests/core/p2p-proto/test_sync.py
carver: Yeah, it's definitely sensitive to which block you're trying to sync to. "Beam" syncing to the first block hardly tests any code path, since you have all the state you need and don't need to request it from your peers. What's happening when you run it on a later block like 66?

carver:
> you might already like to leave some feedback on it.

Are there any areas/files that you want early focus on?

tests/core/p2p-proto/test_sync.py
cburgdorf: Ah ok! So it might not work at all yet. It starts breaking after block 7 so that might be when the first requests to peers are needed. Gotta dig in there.

>What's happening when you run it on a later block like 66?

I was just running into timeouts eventually. It's probably that some APIs on the proxy peer aren't working yet. But all other tests (fast, skeleton, regular, regular with fallback, light) do work.

Could also be that some parts of the test setup itself aren't correct yet. I'll dive in and find out :)

cburgdorf:
>Are there any areas/files that you want early focus on?

I'd say the changes to the syncing components and especially the peer performance metrics and meta data handling.

I'm also not crazy about the whole `PeerPoolEventServer` ping pong but I suspect we'll see big cleanups when we move to an architecture where all these incoming messages are handled natively on the event bus (as opposed to the current mapping-based approach). But I consider that out of the scope of this PR.

tests/core/p2p-proto/test_sync.py
```diff
@@ -406,7 +445,7 @@ def chaindb_churner(leveldb_churner):
     return chain.chaindb
 
 
-async def wait_for_head(headerdb, header, sync_timeout=3):
+async def wait_for_head(headerdb, header, sync_timeout=30):
```
carver: I assume this is just a change during testing. It's a big bummer to wait for 30s to find out something is broken. Also, if it's taking longer than 3s, then performance is so busted that something needs to be fixed.

p2p/service.py
```diff
@@ -230,7 +231,10 @@ def run_daemon(self, service: 'BaseService') -> None:
 
         self.run_task(_run_daemon_wrapper())
 
-    def call_later(self, delay: float, callback: 'Callable[..., None]', *args: Any) -> None:
+    def call_later(self,
+                   delay: float,
+                   callback: Callable[..., Union[None, Awaitable[None]]],
```
carver: I don't think this is sufficient, because the callback needs to be awaited. This could definitely break beam sync. Maybe just add a new `call_later_async()` and send it the `Awaitable`?

tests/core/p2p-proto/test_sync.py
cburgdorf: >I assume this is just a change during testing

Yep, it's just that.

p2p/service.py
cburgdorf: Good point :+1:

tests/core/p2p-proto/test_sync.py
cburgdorf: @carver I've spend two hours trying to get this to work for block `66` (or anything above 10 or so) but I failed to find out what's going on.

From what I see, it's not starting to execute the transactions eventually running into the timeout.

Maybe if you'd check out the code and spend 10 minutes with it you'd quickly see what's wrong and get me on the right path again?

trinity/sync/full/chain.py
```diff
@@ -701,6 +715,8 @@ def register_peer(self, peer: BasePeer) -> None:
     def _mark_complete(self) -> None:
         self.is_complete = True
         self.cancel_nowait()
+        # FIXME: cancel_nowait does not seem to cancel child services?
+        self.proxy_peer_pool.cancel_nowait()
```
carver: Yup, it doesn't *forcibly* close child tasks. A service that cooperatively exits shouldn't need that. Maybe the proxy peer pool isn't exiting cleanly when the cancel token is triggered?

trinity/sync/common/peers.py
```diff
@@ -51,7 +53,8 @@ def _get_peer_rank(self, peer: TChainPeer) -> float:
         # high throughput peers should pop out of the queue first, so ranked as negative
         return -1 * avg_throughput
 
-    def put_nowait(self, peer: TChainPeer) -> None:
+    async def put(self, peer: TChainPeer) -> None:
+        await peer.get_perf_metrics()
```
carver: *tangential nit* The name for this surprised me, since it's not returning the metrics. Maybe something like `reload_` or `refresh_` instead of `get_`?

tests/core/p2p-proto/test_sync.py
carver: I'll take a look. Next time, it would be nice if you leave in the broken tests so I can see the output in CI. :)

carver: Hm, this is odd:
```
    INFO  08-08 15:39:06              state.py  Starting beam state sync
    INFO  08-08 15:39:06              state.py  Beam-Sync: all=0  urgent=0  pred=0  all/sec=0  urgent/sec=0  reqs=0  pred_reqs=0  timeouts=0
   DEBUG  08-08 15:39:06            service.py  <trinity.sync.full.chain.RegularChainBodySyncer object at 0x7fa6fadfe668> finished: Cancellation requested by RegularChainBodySyncer:BeamSyncer token
   DEBUG  08-08 15:39:06            service.py  Waiting for tasks (4): [<Task pending coro=<RegularChainBodySyncer._preview_ready_blocks() running at /home/jcarver/code/trinity/p2p/service.py:151> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7fa6fad146d8>()]>>, <Task pending coro=<RegularChainBodySyncer._display_stats() running at /home/jcarver/code/trinity/p2p/service.py:151> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7fa6fad14888>()]>>, <Task pending coro=<BaseBodyChainSyncer._assign_body_download_to_peers() running at /home/jcarver/code/trinity/p2p/service.py:151> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7fa6fad30e58>()]>>, <Task pending coro=<RegularChainBodySyncer._import_ready_blocks() running at /home/jcarver/code/trinity/p2p/service.py:151> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7fa6fad14528>()]>>]
   DEBUG  08-08 15:39:06            service.py  <trinity.sync.common.headers.HeaderMeatSyncer object at 0x7fa6fadee8d0> finished: Cancellation requested by HeaderMeatSyncer:ETHHeaderChainSyncer:BeamSyncer token
```

Looks like it's being cancelled basically as soon as it starts...

trinity/sync/full/chain.py
carver: In fact, this was causing the problem. A couple `BodySyncer`s are started and stopped during the lifecycle of beam sync. We can't assume that everything is shutting down here. In this case, what happens is:

```
   DEBUG  08-08 15:39:06            service.py  <trinity.protocol.eth.peer.ETHProxyPeerPool object at 0x7fa6fae60c18> finished: Cancellation requested by ETHProxyPeerPool token
```

Then later on:

```
   ERROR  08-08 15:39:06             plugin.py  Task exception was never retrieved
future: <Task finished coro=<async_generator_asend()> exception=StopAsyncIteration()>
StopAsyncIteration
   DEBUG  08-08 15:39:06            service.py  <coroutine object HeaderMeatSyncer._watch_new_peers at 0x7fa6fadb7200> finished while <trinity.sync.common.headers.HeaderMeatSyncer object at 0x7fa6fadee8d0> is still running, terminating as well
   DEBUG  08-08 15:39:06            service.py  Cancelling <trinity.sync.common.headers.HeaderMeatSyncer object at 0x7fa6fadee8d0>
```

Cancelling the peer pool meant that all the other services that were watching for new peers get shut down, since they detected the bad condition that the peer pool is gone.

So then beam sync times out, because it's not trying to sync anymore. :)

cburgdorf: Oh! I understand! 🤯

Thanks for helping me to debug this! I noticed the odd early cancellation but I felt lost finding the root cause of it. :-/

cburgdorf:
I'm going to put this aside for now. Here's why:

1. It is unclear if this would make beam sync faster or slower (My anticipation is that it won't *directly* improve sync speed but improve peer connectivity in general (e.g. can hold more peer connections) which *might* end up improving sync speed)

2. With the former in mind, it would be hard to justify landing this before devcon if it turns out to make sync times worse

3. This code is hard to keep fresh (read: rebase frequently) in a branch as it touches many areas that are currently in flux (beam sync, p2p stuff)

4. This PR currently makes the ergonomics for working on something like beam sync worse. There's a lot of back and forth to bridge the communication between the different processes and we may need better primitives to make that more ergonomic (e.g. a `lahja` version of `ProxyObject` from `multiprocessing` but with a better API)

To be clear, this doesn't mean we don't want to separate sync and peer pool anymore, it just means that this may not be the right code at the right time.

PR #726, opened 2019-06-14
Use dataclass to reduce event boiler plate

cburgdorf:
### What was wrong?

A lot of the boiler plate that comes with creating lahja events can be reduces using [dataclasses](https://docs.python.org/3/library/dataclasses.html). Unfortunately, dataclasses do not exist in Python 3.6 but there is a drop-in backport that we can conditionally install only for Python 3.6 environments.

### How was it fixed?

1. Conditionally add dataclasses backport for Python 3.6
2. Migrate places that can be converted to use dataclass (ongoing)

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/events.py
+++ b/trinity/events.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from typing import (
     Tuple,
 )
@@ -8,12 +9,13 @@ from lahja import (
 )
 
 
+@dataclass
 class ShutdownRequest(BaseEvent):
 
-    def __init__(self, reason: str="") -> None:
-        self.reason = reason
+    reason: str = ""
 
 
+@dataclass
 class EventBusConnected(BaseEvent):
     """
     Broadcasted when a new :class:`~lahja.endpoint.Endpoint` connects to the ``main``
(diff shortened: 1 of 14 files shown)


cburgdorf:
How does everyone feel about this? I think I prefer it over https://github.com/ethereum/lahja/issues/63 mainly because:

- Python stdlib solution (at least for 3.7, 3.6 works with backport)
- dataclasses may have other features that make them superior over a homegrown lahja solution
- even if we ever go with a homegrown lahja solution, this would just mean that we'd have to drop the `@dataclass` decorator from the events

If this is received well, I'd expand the PR to cover all cases where it applies.

trinity/events.py
```diff
@@ -8,12 +9,13 @@
 )
 
 
+@dataclasses.dataclass
```
carver: I'm +0 on switching the import to `from dataclasses import dataclass` and this one to `@dataclass`

cburgdorf: I'm cool with that even though our coding standards discourages that https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#avoid-from-x-import-y-for-most-standard-library-imports

cburgdorf:
@carver went through all events and updated them. Looks so much sweeter imho :)

trinity/protocol/common/events.py
```diff
@@ -21,37 +23,30 @@
...
 class HasRemoteEvent(BaseEvent, ABC):
     """
     Abstract base event for event types that carry a ``Node`` on the ``remote`` property.
     """
 
-    @property
-    @abstractmethod
-    def remote(self) -> Node:
-        pass
+    remote: Node
```
cburgdorf: This isn't abstract anymore but honestly it makes the code so much more concise that I think its a fair trade-off

carver: Maybe drop the `ABC` then?

PR #728, opened 2019-06-19
Use db-shell to connect to a running Trinity

carver:
### What was wrong?

Want to open the db shell on a running trinity client.

### How was it fixed?

Check if trinity is already running by looking for the database IPC file. If it is already created, connect to it using a `DBManager`. Otherwise, just open the database directly (the original case).

### To-Do

#### Cute Animal Picture

### DIFF
--- a/trinity/db/eth1/manager.py
+++ b/trinity/db/eth1/manager.py
@@ -1,3 +1,4 @@
+import multiprocessing
 from multiprocessing.managers import (
     BaseManager,
 )
@@ -19,6 +20,8 @@ from trinity.initialization import (
 )
 from trinity._utils.mp import TracebackRecorder
 
+AUTH_KEY = b"not secure, but only connect over IPC"
+
 
 def create_db_server_manager(trinity_config: TrinityConfig,
                              base_db: BaseAtomicDB) -> BaseManager:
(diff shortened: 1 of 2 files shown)


trinity/db/eth1/manager.py
```diff
@@ -19,6 +20,8 @@
 )
 from trinity._utils.mp import TracebackRecorder
 
+AUTH_KEY = b"not secure, but only connect over IPC"
```
cburgdorf: We *could* think about making this optional so that you have to run trinity with something like `--allow-untrusted-ipc` in order to be able to connect via `trinity db-shell` to a *running* db.

I'm just pointing that out to have it said, I'm not really concerned about it as our security model already assumes that all kind of bad things would happen if someone gets access to this level.

trinity/plugins/builtin/attach/console.py
```diff
@@ -106,6 +116,7 @@ def db_shell(use_ipython: bool, database_dir: Path, trinity_config: TrinityConfi
     Head: #{head.block_number}
     Hash: {head.hex_hash}
     State Root: {encode_hex(head.state_root)}
+    Trinity already running? {trinity_already_running}
```
cburgdorf: Would something like "Live DB connection (Trinity running): true" be overly explicit or add clarity?

carver: Maybe:

- Connected to existing trinity?
- Attached to running process?
- Inspecting active trinity?

Or invert:
- Is DB frozen?
- Direct DB connection?
- Isolated access?

---

I'll go with "Inspecting active trinity?" and we can improve it if we come up with something better.

cburgdorf: *Inspecting active trinity* makes it crystal clear :+1:

PR #725, opened 2019-06-14
Expose PeerJoined/PeerLeft events on eventbus

cburgdorf:
### What was wrong?

In order to be able to build plugins with complex peer pool interactions we need to expose events for when a peer joined or left.

This is a small spin out from my ongoing work on making all peer pool interactions accessible through the event bus. 

### How was it fixed?

Broadcast `PeerJoinedEvent` and `PeerLeftEvent` from within the `PeerPoolEventServer`.

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Clean up commit history

[//]: # (For important changes, please add a new entry to the running release notes PR)
[//]: # (You can find the current one using: https://github.com/ethereum/trinity/pulls?q=is%3Aopen+is%3Apr+label%3A%22Release+Notes%22 )
[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)
- [ ] Add entry to the [release notes PR](https://github.com/ethereum/trinity/pulls?q=is%3Aopen+is%3Apr+label%3A%22Release+Notes%22)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #729, opened 2019-06-19
Make requests work on proxy peers

cburgdorf:
### What was wrong?

This includes #725 but doesn't use these events yet (coming in a follow up PR).
This PR is a spin out from #721 which moves the syncing into isolated processes. I tried to find a clean subset of changes that are ready for review which I think this is. 

### How was it fixed?

1. Introduces a `BaseProxyPeerPool` which can be used from any process that wants to interact with the actual peer pool. In its current version, this doesn't do much apart from registering proxy peers. In a follow up PR it will start monitoring incoming/leaving peers automatically based on the `PeerJoinedEvent` / `PeerLeftEvent`.

2. It proxies all peer requests so that `proxy_peer.get_block_headers(...)` becomes possible. These are core APIs that the syncing code will use when it runs in its own process.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/p2p-proto/test_requests.py
+++ b/tests/core/p2p-proto/test_requests.py
@@ -1,0 +1,167 @@
+import pytest
+
+from p2p.exceptions import PeerConnectionLost
+
+from trinity.protocol.eth.peer import (
+    ETHPeerPoolEventServer,
+)
+from tests.core.integration_test_helpers import (
+    FakeAsyncChainDB,
+    run_peer_pool_event_server,
+    run_proxy_peer_pool,
+    run_request_server,
+)
+from tests.core.peer_helpers import (
+    get_directly_linked_peers,
+    MockPeerPoolWithConnectedPeers,
+)
+
+
+@pytest.mark.asyncio
+async def test_proxy_peer_requests(request,
+                                   event_bus,
+                                   other_event_bus,
+                                   event_loop,
+                                   chaindb_fresh,
+                                   chaindb_20):
+    server_event_bus = event_bus
+    client_event_bus = other_event_bus
+    client_peer, server_peer = await get_directly_linked_peers(
+        request,
+        event_loop,
+        alice_headerdb=FakeAsyncChainDB(chaindb_fresh.db),
+        bob_headerdb=FakeAsyncChainDB(chaindb_20.db),
+    )
+
+    client_peer_pool = MockPeerPoolWithConnectedPeers([client_peer], event_bus=client_event_bus)
+    server_peer_pool = MockPeerPoolWithConnectedPeers([server_peer], event_bus=server_event_bus)
+
+    async with run_peer_pool_event_server(
+        client_event_bus, client_peer_pool, handler_type=ETHPeerPoolEventServer
+    ), run_peer_pool_event_server(
+        server_event_bus, server_peer_pool, handler_type=ETHPeerPoolEventServer
+    ), run_request_server(
+        server_event_bus,
+        FakeAsyncChainDB(chaindb_20.db)
+    ), run_proxy_peer_pool(
+        client_event_bus
+    ) as client_proxy_peer_pool, run_proxy_peer_pool(
+        server_event_bus
+    ):
+
+        proxy_peer = await client_proxy_peer_pool.ensure_proxy_peer(client_peer.remote)
+
+        headers = await proxy_peer.requests.get_block_headers(0, 1, 0, False)
+
+        assert len(headers) == 1
+        block_header = headers[0]
+        assert block_header.block_number == 0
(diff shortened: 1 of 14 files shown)


trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -151,6 +171,43 @@ def get_peer(self, remote: Node) -> TPeer:
...
+            self,
+            event_type: Type[TStreamRequest],
+            event_handler_fn: Callable[[TPeer, TStreamRequest], Any]) -> None:
+
+        async for event in self.wait_iter(self.event_bus.stream(event_type)):
+            try:
+                peer = self.get_peer(event.remote)
+            except PeerConnectionLost as e:
+                await self.event_bus.broadcast(
+                    event.expected_response_type()(None, e),
```
pipermerriam: I'd recommend using this pattern: https://github.com/ethereum/trinity/pull/615/files#diff-bbd04df045409523932bc2c673f748a2R173

That way it preserves the traceback.  You might also look into using [`__context__`](https://stackoverflow.com/questions/11235932/what-is-the-difference-between-cause-and-context) to make it show both exceptions since we might want to see both the original stacktrace **and** the fact that the exception is being raised from here.

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -173,3 +230,47 @@ class DefaultPeerPoolEventServer(PeerPoolEventServer[BasePeer]):
...
+                 event_bus: TrinityEventBusEndpoint,
+                 broadcast_config: BroadcastConfig,
+                 token: CancelToken=None):
+        super().__init__(token)
+        self.event_bus = event_bus
+        self.broadcast_config = broadcast_config
+        self.connected_peers: Dict[Node, TProxyPeer] = dict()
+
+    @abstractmethod
+    def to_proxy_peer(self,
```
pipermerriam: It isn't clear to me what this method is supposed to do by the name.

tests/core/p2p-proto/test_requests.py
```diff
@@ -0,0 +1,167 @@
...
+    ), run_proxy_peer_pool(
+        client_event_bus
+    ) as client_proxy_peer_pool, run_proxy_peer_pool(
+        server_event_bus
+    ):
+
+        proxy_peer = await client_proxy_peer_pool.ensure_proxy_peer(client_peer.remote)
+
+        with pytest.raises(TimeoutError):
+            await proxy_peer.requests.get_block_headers(0, 1, 0, False, timeout=0.1)
```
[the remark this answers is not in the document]
cburgdorf: :+1:  `0.01` works

trinity/protocol/common/peer_pool_event_bus.py
cburgdorf: I tried this out but it appears that tracebacks aren't pickable. Then I looked at this and tried if I could reuse that.

https://github.com/ethereum/trinity/blob/b5057028ab10cd978dec031702165633743500c2/trinity/_utils/mp.py#L69-L99

I played with it but I didn't get to a point where I could prove that it would do what it's supposed to. The tests run in a single process and so the traceback seems to work just fine here and kinda makes it hard for me to figure out what would be the right way to do it.

I propose I open up a ticket that we need to look into the `PeerPoolEventServer` to check what would be the best way to handle this

trinity/protocol/common/peer_pool_event_bus.py
cburgdorf: Alright, changed it to `convert_node_to_proxy_peer`

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -42,15 +48,20 @@
...
     PeerLeftEvent,
 )
+from .peer import (
+    BaseProxyPeer,
+)
 
 
 TPeer = TypeVar('TPeer', bound=BasePeer)
 TStreamEvent = TypeVar('TStreamEvent', bound=HasRemoteEvent)
+TStreamRequest = TypeVar('TStreamRequest', bound=HasRemoteAndTimeoutRequest[Any])
```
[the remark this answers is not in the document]
cburgdorf: Quick sum up of what we discussed offline: 

Unfortunately this doesn't work yet. A `Protocol` can not inherit the properties of `BaseEvent` which means the `stream()` API will scream (rightfully) as we try to pass in such thing.

What we really want to be able is to have something like:

`TStreamEvent = TypeVar('TStreamEvent, bound=(HasRemote, BaseEvent))`

But unfortunately having generics that are bound by multiple things doesn't yet work with mypy.

PR #736, opened 2019-06-21
Implement WHOAREYOU packet encoding/decoding

jannikluhn:
WHOAREYOU discv5 packet en- and decoding

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Clean up commit history

[//]: # (For important changes, please add a new entry to the running release notes PR)
[//]: # (You can find the current one using: https://github.com/ethereum/trinity/pulls?q=is%3Aopen+is%3Apr+label%3A%22Release+Notes%22 )
[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #738, opened 2019-06-21
Discv5 whoareyou preparation

jannikluhn:
A helper function to prepare WHOAREYOU packets. This may seem somewhat unnecessary given that most parameters are just passed to the packet constructor. Preparing message packets however is more involved (mostly because encryption), so for consistency an equivalent function for WHOAREYOU packets are implemented as well.

Builds on top of #736

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


jannikluhn:
Rebased and ready for review.

I only now realized that it would make more sense to transform the packet preparation helpers into classmethods of the packet tuples, similar to the encode functions (but not the `decode` functions, because we don't necessarily know what we're decoding if we just get something from a peer). Unfortunately, the auth header packet preparation function is already merged. But since the rest of the PR is quite small, I just did the move here, to avoid an extra PR (it's in the last commit).

PR #737, opened 2019-06-21
Implement disv5 message packet encoding

jannikluhn:
AUTHHEADER and AUTHTAG message packet encoding and decoding.

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [x] Clean up commit history

[//]: # (For important changes, please add a new entry to the running release notes PR)
[//]: # (You can find the current one using: https://github.com/ethereum/trinity/pulls?q=is%3Aopen+is%3Apr+label%3A%22Release+Notes%22 )
[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #739, opened 2019-06-21
Discv5 Auth Tag Packet Preparation

jannikluhn:
Helper function to create auth tag packets (mostly to automatically encrypt the message).

Builds on top of #737 

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/packets.py
+++ b/p2p/discv5/packets.py
@@ -127,6 +127,39 @@ class AuthTagPacket(NamedTuple):
     auth_tag: Nonce
     encrypted_message: bytes
 
+    @classmethod
+    def prepare(cls,
+                *,
+                tag: Hash32,
+                auth_tag: Nonce,
+                message: BaseMessage,
+                initiator_key: AES128Key,
+                ) -> "AuthTagPacket":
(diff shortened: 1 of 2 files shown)


jannikluhn:
This would be ready for review.

p2p/discv5/packets.py
```diff
@@ -127,6 +127,39 @@ class AuthTagPacket(NamedTuple):
...
+            authenticated_data=tag,
+        )
+        return cls(
+            tag=tag,
+            auth_tag=auth_tag,
+            encrypted_message=encrypted_message,
+        )
+
+    @classmethod
+    def prepare_random(cls,
```
pipermerriam: Is this just for testing purposes or does it have a use in the production API?

jannikluhn: It's used to initiate a handshake. If no session keys have been established or Alice wants to use fresh ones, Alice will send a random packet which Bob will fail to decrypt and respond with `WHOAREYOU`. We could also encrypt a message with a random key that we'll throw away immediately, but having a dedicated method seems a little bit cleaner to me.

PR #741, opened 2019-06-21
Discv5 AuthHeader Packet Preparation

jannikluhn:
Helpers to simplify creation of auth header message packets.

Builds on top of #737 

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/packets.py
+++ b/p2p/discv5/packets.py
@@ -15,7 +15,6 @@ from rlp.exceptions import (
 )
 
 from eth_utils import (
-    int_to_big_endian,
     is_bytes,
     encode_hex,
     is_list_like,
@@ -338,15 +337,10 @@ def compute_encrypted_message(initiator_key: AES128Key,
                               message: BaseMessage,
                               authenticated_data: bytes,
                               ) -> bytes:
-    message_plain_text = b"".join((
-        int_to_big_endian(message_data.message_type),
-        rlp.encode(message_data)
-    ))
     encrypted_message = aesgcm_encrypt(
         key=initiator_key,
         nonce=auth_tag,
-        plain_text=message_plain_text,
+        plain_text=message.to_bytes(),
         authenticated_data=authenticated_data,
     )
     return encrypted_message
->>>>>>> Implement auth header preparation helpers
--- a/p2p/discv5/packets.py
+++ b/p2p/discv5/packets.py
@@ -35,7 +35,7 @@ from p2p.discv5.encryption import (
     validate_nonce,
 )
 from p2p.discv5.messages import (
-    MessageData,
+    BaseMessage,
 )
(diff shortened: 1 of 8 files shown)


p2p/discv5/packets.py
```diff
@@ -170,3 +183,87 @@ def _decode_auth(encoded_packet: bytes) -> Tuple[Union[AuthHeader, Nonce], int]:
...
+        nonce=ZERO_NONCE,
+        plain_text=plain_text_auth_response,
+        authenticated_data=tag,
+    )
+    return encrypted_auth_response
+
+
+def compute_encrypted_message(initiator_key: AES128Key,
+                              auth_tag: Nonce,
+                              message_data: MessageData,
```
pipermerriam: Maybe it would be more clear if this class was called `BaseMessage` and we used a `TMessage = TypeVar("TMessag", bound=BaseMessage)` as the type here.  For a minute I got confused reading this since `MessageData` only defines the `message_type` and I didn't realize on first glance that it was intended to be a base class.

p2p/discv5/packets.py
```diff
@@ -170,3 +183,87 @@ def _decode_auth(encoded_packet: bytes) -> Tuple[Union[AuthHeader, Nonce], int]:
...
+                              authenticated_data: bytes,
+                              ) -> bytes:
+    message_plain_text = b"".join((
+        int_to_big_endian(message_data.message_type),
+        rlp.encode(message_data)
+    ))
```
pipermerriam: This looks like a good candidate for `BaseMessage.serialize_for_wire()` (or some other name).

p2p/discv5/packets.py
jannikluhn: :+1: to `MessageData -> BaseMessage`.

The `TypeVar` trick doesn't seem to work though unfortunately, at least mypy doesn't complain about the following (my guess is that's because `issubclass(A, A)`):

```Py
from typing import TypeVar

class A:
    pass

class B(A):
    pass

TA = TypeVar("TA", bound=A)

def x(a: TA) -> None:
    pass

x(A())
```

But I hope with the `Base` prefix it should be clear enough.

p2p/discv5/packets.py
jannikluhn: :+1: I simply used `to_bytes` as the method name, which I'm not totally happy with, but I think is okay because there's no other binary encoding used anywhere. My problem with `serialize_for_wire` was that it's not directly sent over the wire as it's encrypted first. I also considered `to_plaintext_bytes`, but this might be confusing as `plaintext` only makes sense in the context of encryption of which the message itself doesn't really know about.

jannikluhn:
I've addressed the feedback, put will wait with rebasing until the parent PRs are merged.

PR #745, opened 2019-06-22
Log attestations and fix broadcasting msg to peers

ChihChengLiang:
### What was wrong?

- Attestation included in the state is not logged.
- Trinity broadcasts message back to it's origin.

### How was it fixed?

- Add `__repr__`s and logs current / previous _epoch_attestations from state
- Explicitly compare peer with peer.remote

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/types/blocks.py
+++ b/eth2/beacon/types/blocks.py
@@ -166,7 +166,9 @@ class BaseBeaconBlock(ssz.SignedSerializable, Configurable, ABC):
         return (
             f'<Block #{self.slot} '
             f'signing_root={encode_hex(self.signing_root)[2:10]} '
-            f'root={encode_hex(self.root)[2:10]}>'
+            f'root={encode_hex(self.root)[2:10]} '
+            f'attestations={self.num_attestations} '
+            '>'
         )
 
     @property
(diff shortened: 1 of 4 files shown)


eth2/beacon/types/blocks.py
```diff
@@ -166,7 +166,9 @@ def __repr__(self) -> str:
         return (
             f'<Block #{self.slot} '
             f'signing_root={encode_hex(self.signing_root)[2:10]} '
-            f'root={encode_hex(self.root)[2:10]}>'
+            f'root={encode_hex(self.root)[2:10]} '
+            f'attestations={self.num_attestations} '
```
NIC619: Are we logging attestations or number of attestations here?

eth2/beacon/types/pending_attestations.py
```diff
@@ -38,3 +38,6 @@ def __init__(self,
             custody_bitfield=custody_bitfield,
             inclusion_slot=inclusion_slot,
         )
+
+    def __repr__(self) -> str:
+        return f"<PA inclusion_slot={self.inclusion_slot}>"
```
NIC619: nit: `PA` is kind of not self explanatory IMO.

eth2/beacon/types/blocks.py
ChihChengLiang: "number of attestations" to visually look nicer

PR #743, opened 2019-06-21
Discv5 Packet Decryption

jannikluhn:
Decrypt the message contained in auth header or auth tag packets.

Builds on top of #741 and will require a rebase once #739 is merged to make tests pass.

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Clean up commit history

[//]: # (For important changes, please add a new entry to the running release notes PR)
[//]: # (You can find the current one using: https://github.com/ethereum/trinity/pulls?q=is%3Aopen+is%3Apr+label%3A%22Release+Notes%22 )
[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)
- [ ] Add entry to the [release notes PR](https://github.com/ethereum/trinity/pulls?q=is%3Aopen+is%3Apr+label%3A%22Release+Notes%22)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #747, opened 2019-06-22
Discv5 AuthHeader decryption

jannikluhn:
Function to decrypt the auth header packet during handshake, containing the signed id nonce and, optionally, an updated ENR.

Builds on top of #741 

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/packets.py
+++ b/p2p/discv5/packets.py
@@ -14,11 +14,12 @@ from rlp.sedes import (
 )
 from rlp.exceptions import (
     DecodingError,
+    DeserializationError,
 )
 
 from eth_utils import (
-    is_bytes,
     encode_hex,
+    is_bytes,
     is_list_like,
     ValidationError,
 )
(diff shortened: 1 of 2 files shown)


jannikluhn:
this PR as well as #743 are ready, which are final two of the packeting PRs

p2p/discv5/packets.py
```diff
@@ -112,6 +114,57 @@ def prepare(cls,
...
+        id_nonce_signature, serialized_enr = decoded_rlp
+
+        if is_list_like(id_nonce_signature):
+            raise ValidationError(
+                f"Id nonce signature is a list instead of bytes: {id_nonce_signature}"
+            )
+        elif not is_bytes(id_nonce_signature):
+            raise Exception("Unreachable: RLP is either bytes or list")
+
+        if is_bytes(serialized_enr):
```
pipermerriam: I think that these conditionals that do an `if is_bytes... > elif not is_list_like` could be collapsed to a single `if not is_list_like(...)` since these are required to be list-likes and `is_list_like(some_bytes_type)` returns `False`

PR #749, opened 2019-06-24
Get rid of LightPeerChainBridgePlugin

cburgdorf:
### What was wrong?

We currently have a plugin called the `LightPeerChainBridgePlugin` which was born as a band aid back then when we moved the JSON-RPC server into an isolated plugin. It does two things:

1. Provide a `EventBusLightPeerChain` which is a `LightPeerChain` implementation that runs atop the event bus and is hence usable from all processes
2. Run an `LightPeerChainEventBusHandler` that feeds requests for the `EventBusLightPeerChain`

Beside the sync orchestration this is the last plugin that uses the `BaseAsyncStopPlugin` that we want to get rid off. Also the `LightPeerChainEventBusHandler` looks very much like the `PeerPoolEventServer`.

### How was it fixed?

1. Refactored the `PeerPoolEventServer` to work independently of the `HasRemoteEvent` and `HasRemoteAndTimeoutEvent` abstractions.

2. Move functionality of the `LightPeerChainEventBusHandler` into the `LESPeerPoolEventServer`

3. Move bootstrapping of the event server into the node (was previously in p2p server)

4. Moved `EventBusLightPeerChain` into `trinity.chains`.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


trinity/protocol/bcc/peer.py
```diff
@@ -166,7 +166,10 @@ class BCCPeerPoolEventServer(PeerPoolEventServer[BCCPeer]):
 
         self.run_daemon_event(
             SendBeaconBlocksEvent,
-            lambda peer, ev: peer.sub_proto.send_blocks(ev.blocks, ev.request_id)
+            lambda event: self.try_with_node(
+                event.remote,
+                lambda peer: peer.sub_proto.send_blocks(event.blocks, event.request_id)
```
cburgdorf: @pipermerriam hey, I heard you don't like lambdas so I put a lambda in a lambda :sweat_smile: On a serious note, I got rid of the dependency against `HasRemoteEvent` and `HasRemoteAndTimeoutEvent` which I kinda needed for this PR. But the only way that I could see to make that work was using this construct. The only other way I see would be to duplicate the code of `try_with_node` and `with_node_and_timeout` everywhere. I'm open to other suggestions :)

cburgdorf:
This landed with #763

PR #750, opened 2019-06-24
Fix broken JSON-RPC calls in light mode

cburgdorf:
### What was wrong?

This PR addresses two different things:

1. Making JSON-RPC calls to a light client was broken
2. The `NoConnectedPeers` error had no message which is kinda unfortunate as the text is picked up by `web3` and so it is the only thing that could guide a user towards the problem

### How was it fixed?

1. The `LightPeerChain` needs a `HeaderDB` not a `AsyncHeaderDB`
2. Added a bit of test to the `NoConnectedPeers` error

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/common/peer.py
+++ b/trinity/protocol/common/peer.py
@@ -177,7 +177,7 @@ class BaseChainPeerPool(BasePeerPool):
     def highest_td_peer(self) -> BaseChainPeer:
         peers = tuple(self.connected_nodes.values())
         if not peers:
-            raise NoConnectedPeers()
+            raise NoConnectedPeers("No connected peers")
         peers_by_td = groupby(operator.attrgetter('head_td'), peers)
         max_td = max(peers_by_td.keys())
         return random.choice(peers_by_td[max_td])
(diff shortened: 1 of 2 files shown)


cburgdorf:
@carver @pipermerriam it's kinda sad that we don't have a test that would have caught this. It's probably worth taking the geth testing primitives that we have for the lightchain integration tests and reuse them to get the `test_web3` tests working against the light client.

trinity/protocol/common/peer.py
```diff
@@ -177,7 +177,7 @@ class BaseChainPeerPool(BasePeerPool):
     def highest_td_peer(self) -> BaseChainPeer:
         peers = tuple(self.connected_nodes.values())
         if not peers:
-            raise NoConnectedPeers()
+            raise NoConnectedPeers("No connected peers")
```
carver: Another option is to set a default message in the exception class, so we don't have to explicitly add the default message in every instantiation.

PR #755, opened 2019-06-27
Move responsibility of receiving handshake into Transport

pipermerriam:
### What was wrong?

The process of receiving a handshake wasn't something that could be re-used, involving significant custom code in two locations, one in the `Server` and another *special* one in our tests.

### How was it fixed?

Implement this in a re-usable location on the `Transport` class.

All of the code in the `receive_handshake` method is just copy/pasta and shouldn't contain any functionality changes.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/transport.py
+++ b/p2p/transport.py
@@ -83,9 +83,6 @@ class Transport:
         mac_cipher = Cipher(algorithms.AES(mac_secret), modes.ECB(), default_backend())
         self._mac_enc = mac_cipher.encryptor().update
 
-        self._received_message_count = 0
-        self._sent_message_count = 0
-
     @classmethod
     async def connect(cls,
                       remote: Node,
@@ -131,10 +128,13 @@ class Transport:
                                  writer: asyncio.StreamWriter,
                                  private_key: datatypes.PrivateKey,
                                  token: CancelToken) -> 'Transport':
-        msg = await token.cancellable_wait(
-            reader.readexactly(ENCRYPTED_AUTH_MSG_LEN),
-            timeout=REPLY_TIMEOUT,
-        )
+        try:
+            msg = await token.cancellable_wait(
+                reader.readexactly(ENCRYPTED_AUTH_MSG_LEN),
+                timeout=REPLY_TIMEOUT,
+            )
+        except asyncio.IncompleteReadError as err:
+            raise HandshakeFailure from err
 
         try:
             ephem_pubkey, initiator_nonce, initiator_pubkey = decode_authentication(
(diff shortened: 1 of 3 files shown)


carver:
Taking a look now...

p2p/transport.py
```diff
@@ -73,6 +83,9 @@ def __init__(self,
         mac_cipher = Cipher(algorithms.AES(mac_secret), modes.ECB(), default_backend())
         self._mac_enc = mac_cipher.encryptor().update
 
+        self._received_message_count = 0
+        self._sent_message_count = 0
```
carver: Looks like this isn't read anywhere (just incremented). Maybe the counters could be pulled out here, and added later when they are used.

p2p/transport.py
```diff
@@ -152,15 +251,20 @@ def write(self, data: bytes) -> None:
...
             )
             raise MalformedMessage from err
+
+        # track total messages received
+        self._received_message_count += 1
         return msg
 
     def send(self, header: bytes, body: bytes) -> None:
-        cmd_id = rlp.decode(body[:1], sedes=rlp.sedes.big_endian_int)
+        cmd_id = get_devp2p_cmd_id(body)
```
carver: :heart:

p2p/transport.py
```diff
@@ -112,6 +125,92 @@ def __init__(self,
...
         )
 
+    @classmethod
+    async def receive_connection(cls,
+                                 reader: asyncio.StreamReader,
+                                 writer: asyncio.StreamWriter,
+                                 private_key: datatypes.PrivateKey,
+                                 token: CancelToken) -> 'Transport':
+        msg = await token.cancellable_wait(
+            reader.readexactly(ENCRYPTED_AUTH_MSG_LEN),
```
carver: I noticed this changed from `read()` to `readexactly()`. Maybe catch the new https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.IncompleteReadError and raise a `HandshakeFailure`?

Bonus note for logging:
> Use the `IncompleteReadError.partial` attribute to get the partially read data.

p2p/transport.py
```diff
@@ -112,6 +125,92 @@ def __init__(self,
...
+            ephem_pubkey, initiator_nonce, initiator_pubkey = decode_authentication(
+                msg,
+                private_key,
+            )
+        except DecryptionError as non_eip8_err:
+            # Try to decode as EIP8
+            msg_size = big_endian_to_int(msg[:2])
+            remaining_bytes = msg_size - ENCRYPTED_AUTH_MSG_LEN + 2
+            msg += await token.cancellable_wait(
+                reader.read(remaining_bytes),
```
carver: I suppose we could make this a `readexactly()` too, so that we don't ignore when a message is too short.

trinity/server.py
```diff
@@ -176,90 +165,24 @@ def cleanup_reader_and_writer() -> None:
...
-            socket = writer.get_extra_info("socket")
-            sockname = writer.get_extra_info("sockname")
-            raise HandshakeFailure(
-                "Received incoming connection with no remote information:"
-                f"socket={repr(socket)}  sockname={sockname}"
-            )
-
-        ip, socket, *_ = peername
-        remote_address = Address(ip, socket)
-        self.logger.debug("Receiving handshake from %s", remote_address)
```
carver: We lost this log. If it's too noisy, maybe drop it to debug2 instead? I haven't noticed it being too noisy, so I'm happy to see it added back in, too.

PR #760, opened 2019-07-01
Update the serialization so that the state is interpreted correctly

ralexstokes:
### What was wrong?

Part of our current SSZ typing includes `Vector` lengths that are configured at runtime.

The current strategy to handle this dynamism is to override the `py-ssz` metaclass data once we know the configuration values.

There was an integration test that was not doing this runtime configuration and somehow made its way into master, breaking the build.

### How was it fixed?

Setting the ssz config before the state in the failing test was instantiated.

#### Cute Animal Picture


ralexstokes:
Fixes #759 if this is merged.

PR #761, opened 2019-07-02
Ensure we can only have a single concurrent handshake per remote

cburgdorf:
### What was wrong?

This bug was reported by @NIC619 via discord. Here is what happens:

When you run trinity with `--preferred_nodes` and `--no-discovery` it appears that the `DiscoveryPeerBackend` and the `BootnodesPeerBackend` will both return the same set of peers. This doesn't seem to be the actual problem though because I believe that we could have other circumstances leading to the same scenario.

The actual problem is that **if** (for whatever reason) we make multiple connection attempts to the same peer, we don't recognize and prevent it.

E.g the only check that we do is this one.

https://github.com/ethereum/trinity/blob/891a4254b262542120bef0bd9319ce50c38b2f94/p2p/peer_pool.py#L312-L314

But that does only check if we are already **connected**, not if we are already **trying to connect**.

### How was it fixed?

Added a check if we are already trying to connect to a peer to prevent multiple connections from happening. Notice that while this does incur a *lock*, this *lock* should be extremely swift and does not prevent us from connecting concurrently in general.

I guess that in the future, multiple connections to a single peer should become allowed but for **different protocols** (@pipermerriam this is what you are working on, right). For the current state of affairs, this strikes me as a simple fix.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -10,6 +10,7 @@ from typing import (
     Dict,
     Iterator,
     List,
+    Set,
     Tuple,
     Type,
 )
@@ -122,11 +123,16 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
         self.context = context
 
         self.connected_nodes: Dict[Node, BasePeer] = {}
+
         self._subscribers: List[PeerSubscriber] = []
         self._event_bus = event_bus
 
         # Restricts the number of concurrent connection attempts can be made
         self._connection_attempt_lock = asyncio.BoundedSemaphore(MAX_CONCURRENT_CONNECTION_ATTEMPTS)
+
+        # Ensure we can only have a single concurrent handshake in flight per remote
+        self._handshake_lock = asyncio.Lock()
+        self._inflight_handshakes: Set[Node] = set()
 
         self.peer_backends = self.setup_peer_backends()
         self.connection_tracker = self.setup_connection_tracker()
@@ -313,6 +319,13 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
             self.logger.debug2("Skipping %s; already connected to it", remote)
             raise IneligiblePeer(f"Already connected to {remote}")
 
+        async with self._handshake_lock:
+            if remote in self._inflight_handshakes:
+                self.logger.debug2("Skipping %s; already shaking hands", remote)
+                raise IneligiblePeer(f"Already shaking hands with {remote}")
+            else:
+                self._inflight_handshakes.add(remote)
+
         try:
             should_connect = await self.wait(
                 self.connection_tracker.should_connect_to(remote),
@@ -330,7 +343,6 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
             # We use self.wait() as well as passing our CancelToken to handshake() as a workaround
             # for https://github.com/ethereum/py-evm/issues/670.
             peer = await self.wait(handshake(remote, self.get_peer_factory()))
-
             return peer
         except OperationCancelled:
             # Pass it on to instruct our main loop to stop.
@@ -361,6 +373,10 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
         except Exception:
             self.logger.exception("Unexpected error during auth/p2p handshake with %r", remote)
             raise
+        finally:
+            async with self._handshake_lock:
+                self.logger.debug("Removing %s from connect_lock", remote)
+                self._inflight_handshakes.remove(remote)
 
     async def connect_to_nodes(self, nodes: Iterator[Node]) -> None:
         # create an generator for the nodes


NIC619:
Though on the other note, our [script](https://github.com/ethereum/trinity/blob/master/eth2/beacon/scripts/run_beacon_nodes.py) still fails because both nodes are started at the same time then initiating the connection to each other at the same(only one of them, bob, has `--preferred_nodes` set, but I guess the other node, alice, somehow also discover bob).

Spinning up bob a few seconds later than alice fix the issue. And I guess in practice two nodes initiating connection to each other at the same time is pretty rare so we probably don't need to worry about it too much(other than in the tests/local testnet setup).

cburgdorf:
>still fails because both nodes are started at the same time then initiating the connection to each other at the same
>Spinning up bob a few seconds later than alice fix the issue

Oh interesting, so both connecting at the very same instant will also lead to this issue. Well, to be fair, as you said, this is more unlikely in practice and eventually the connection will timeout and be removed. But still interesting.

>only one of them, bob, has --preferred_nodes set, but I guess the other node, alice, somehow also discover bob

If I'm reading the script correctly, both them each other as preferred node configured.

https://github.com/ethereum/trinity/blob/b35ddb9b90a440ec98d5d141e3fbc6e2f1f35b2b/eth2/beacon/scripts/run_beacon_nodes.py#L231-L233

NIC619:
> If I'm reading the script correctly, both them each other as preferred node configured.

🤦‍♂ My bad. That's why they are both initiating connection to each other...

cburgdorf:
>Looking at this I'm not sure it actually solves the problem at hand since this only keeps us from initiating two outbound connections with the same peer.

Yes, but that is exactly what is happening when you use `--preferred_nodes` and `--no-discovery`  as you run into the scenario that both backends will give you the same peer. This change does fix that issue which is confirmed by me locally testing it and @NIC619 confirming seeing the same.

>In the event that we are dialing out at the same time they are dialing in, I think the same problem will arise.

This is true and is acknowledged in the thread.

>Spinning up bob a few seconds later than alice fix the issue.

So while we can still end up with stale connections when we have an outgoing handshake to a peer that does an inbound handshake  to us at the same time, it does fix the original issue though that arises from *us* trying to call out to the same node twice simultaneously. 

>set will grow unbounded over time anytime either of the early exit conditions occur on lines 336 and 339

Can you elaborate on that one. Why would we not reach the `finally` and clean it up?

>I'm inclined to suggest we roll this back (sorry for not getting this review in before it was merged)

But do we have a different fix for it then?

pipermerriam:
> Yes, but that is exactly what is happening when you use --preferred_nodes and --no-discovery as you run into the scenario that both backends will give you the same peer. 

I missed this part (not realizing that both backends gave the same peer back)

pipermerriam:
> Can you elaborate on that one. Why would we not reach the `finally` and clean it up?

It's this block of code.

https://github.com/cburgdorf/trinity/blob/50f7ef53cf863346916236a0167d7731a5fa6ac0/p2p/peer_pool.py#L329-L339

The `remote` is added above this on line 327.  Note that both lines 336 and 339 cause the function to exit.  The `try/finally` with the removal statement doesn't occur until **after** these lines so it will not be run.

p2p/peer_pool.py
```diff
@@ -122,12 +123,17 @@ def __init__(self,
...
+
         self._subscribers: List[PeerSubscriber] = []
         self._event_bus = event_bus
 
         # Restricts the number of concurrent connection attempts can be made
         self._connection_attempt_lock = asyncio.BoundedSemaphore(MAX_CONCURRENT_CONNECTION_ATTEMPTS)
 
+        # Ensure we can only have a single concurrent handshake in flight per remote
+        self._handshake_lock = asyncio.Lock()
+        self._inflight_handshakes: Set[Node] = set()
```
pipermerriam: FYI, I wrote this little utility for something else that might make this easier.

```python
import asyncio
from collections.abc import Hashable
from typing import AsyncIterator
import weakref

from async_generator import asynccontextmanager


class ResourceLock:
    """
    Manage a set of locks for some set of hashable resources.
    """
    _locks: 'weakref.WeakValueDictionary[Hashable, asyncio.Lock]'

    def __init__(self) -> None:
        self._locks = weakref.WeakValueDictionary()

    @asynccontextmanager
    async def lock(self, resource: Hashable) -> AsyncIterator[None]:
        if resource not in self._locks:
            self._locks[resource] = asyncio.Lock()
        lock = self._locks[resource]
        async with lock:
            yield

    def is_locked(self, resource: Hashable) -> bool:
        if resource not in self._locks:
            return False
        else:
            return self._locks[resource].locked()
```

This would allow you to rewrite the code below to just use a context manager around the connection code so that you don't have to do explicit cleanup.

cburgdorf:
> Note that both lines 336 and 339 cause the function to exit

Damn! I totally missed that these are two different `try` blocks! Yep, looks like I screwed that up.

p2p/peer_pool.py
cburgdorf: This looks handy indeed!

PR #762, opened 2019-07-02
Improve quality of logs for run_beacon_nodes script

cburgdorf:
### What was wrong?

While working on #761 I found these changes useful to reduce the scope of things and make the logs more focused.

### How was it fixed?

- The upnp service isn't that useful yet so I disabled it.
- The networking db *is* useful in general but I think it isn't really useful for this test setup.
- I also enabled DEBUG2 logs to increase output in general.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/scripts/run_beacon_nodes.py
+++ b/eth2/beacon/scripts/run_beacon_nodes.py
@@ -120,7 +120,9 @@ class Node:
             f"--trinity-root-dir={self.root_dir}",
             f"--beacon-nodekey={remove_0x_prefix(self.node_privkey.to_hex())}",
             "--disable-discovery",
-            "-l debug",
+            "--network-tracking-backend=do-not-track",
+            "--disable-upnp",
+            "-l debug2",
         ]
         if len(self.preferred_nodes) != 0:
             preferred_nodes_str = ",".join([node.enode_id for node in self.preferred_nodes])


eth2/beacon/scripts/run_beacon_nodes.py
```diff
@@ -120,7 +120,9 @@ def cmd(self) -> str:
             f"--trinity-root-dir={self.root_dir}",
             f"--beacon-nodekey={remove_0x_prefix(self.node_privkey.to_hex())}",
             "--disable-discovery",
-            "-l debug",
+            "--network-tracking-backend=do-not-track",
+            "--disable-upnp",
+            "-l debug2",
```
NIC619: I would actually prefer to keep `debug` and change to `debug2` locally if needed. 😄  We currently set it to `debug` level because some of the validator logs are `debug` level log.

cburgdorf: Just to make sure we are on the same page. The `debug2` level includes everything from `debug2` or *above* which includes `debug`. So, you won't miss out on any `debug` log but will also gain some additional `debug2` logs. Under normal circumstances, this can be a bit overwhelming but since this test script has discovery and upnp enabled it should be ok to print all logs. Do you still want to reduce this to not show `debug2` logs?

NIC619: > Under normal circumstances, this can be a bit overwhelming but since this test script has discovery and upnp enabled it should be ok to print all logs.

Sounds fair. Let's do `debug2` then.

Issue #370, opened 2019-03-05
Tab-complete CLI options


carver:
### What is wrong?

My fingers expect tab-complete to work, eg~

```sh
$ trinity --ro
<TAB>
# should auto-complete to:
$ trinity --ropsten
```

Also,

```sh
$ trinity fi
<TAB>
# should auto-complete to:
$ trinity fix-unclean-shutdown
```

This should install automatically, when trinity is installed.

### How can it be fixed

It looks like we can get 90% of the way there with an non-customized integration of:
https://argcomplete.readthedocs.io/en/latest/index.html

(We can close this ticket after merging that basic auto-complete and adding another issue for the remaining 10%)

An example of the remaining 10% might be things like:
- auto-completing network ID from the set of recognized IDs
- only auto-completing genesis files to those with a `.json` ending
- etc

s0b0lev:
@carver I mention in PR that autocomplete require to add  in **~/.bashrc**:
```
eval "$(register-python-argcomplete trinity)"
````
Maybe it is better to mention such instruction in README.md? So user will decide if he wanna add additional configuration.

**Comment by carver** [comment]
Hm, can we make this a part of running `pip install`?

I'm not really familiar with best/common practices here.

**Comment by s0b0lev** [comment]
Yes, we can modify setup.py to include post-install instruction.

The post-install script will add `eval "$(register-python-argcomplete trinity)"` to ~/.bashrc.

I will test this scenario and update PR.

**Comment by s0b0lev** [comment]
@carver I update setup.py, it is possible to install trinity with autocomplete now using:
```
$ python setup.py install
```

Also, I add one more shell command. This command provides one-time activation of argcomplete (user can use argcomplete without reloading shell or source ~/.basrhc command):
```python
subprocess.call('eval "$(register-python-argcomplete trinity)"', shell=True)
```
Autocomplete for `python setup.py install` works well.

Did I miss any installations way?

carver:
Now we're getting pretty specific, so I'll move convo to the PR. I think it's okay for that PR to only work for bash completion, but we should not add a bunch of cruft if the user isn't using bash.

s0b0lev:
Thanks for feedback @carver, I think the main issue here that we force argcomplete library installation.

Here for example how it works in other projects.

There are 3 ways to configure it:

1. **Add documentation step:**
- https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/enable-tab-completion.html
[...]

PR #768, opened 2019-07-05
Tab complete CLI options

s0b0lev:
### What was wrong?
Adding tab-complete CLI options (relates to https://github.com/ethereum/trinity/issues/370)

### How was it fixed?
Added new requirements: `argcomplete==1.10.0` and configure `argcomplete.autocomplete` module

Solution require to call `$ activate-global-python-argcomplete` - but it will install autocomplete globally.
We can just specify required module in ~/.bashrc (described in todo list below)

### To-Do
- [x] Update documentation with argcomplete activation steps

### DIFF
--- a/docs/api/api.cli.rst
+++ b/docs/api/api.cli.rst
@@ -106,7 +106,7 @@ The following table shows various combinations of how to use logging in Trinity 
 Enabling tab completion
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-It is possible to configure tab-complete for trinity greater then v0.1.0a25.
+Trinity can be configured to auto complete commands when the <tab> key is pressed.
 
 For permanent, not global tab complete activation, use:
 
--- a/docs/api/api.cli.rst
+++ b/docs/api/api.cli.rst
@@ -108,7 +108,7 @@ Enabling tab completion
 
 Trinity can be configured to auto complete commands when the <tab> key is pressed.
 
-For permanent, not global tab complete activation, use:
+After installing trinity, to activate tab-completion in future bash prompts, use:
 
 .. code:: sh
 
--- a/docs/api/api.cli.rst
+++ b/docs/api/api.cli.rst
@@ -101,3 +101,22 @@ The following table shows various combinations of how to use logging in Trinity 
 
 .. [5] **Increasing** the per-module log level above the general ``--file-log-level`` is not yet supported
        (See `issue 689 <https://github.com/ethereum/trinity/issues/689>`_ )
+
+
+Enabling tab completion
+~~~~~~~~~~~~~~~~~~~~~~~
+
+It is possible to configure tab-complete for trinity greater then v0.1.0a25.
+
+For permanent, not global tab complete activation, use:
+
+.. code:: sh
+
+    register-python-argcomplete trinity >> ~/.bashrc
+
+
+For one-time activation of argcomplete for trinity, use:
+
+.. code:: sh
+
+    eval "$(register-python-argcomplete trinity)"
--- a/setup.py
+++ b/setup.py
@@ -1,20 +1,9 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 import os
-import subprocess
 from setuptools import setup, find_packages
-from setuptools.command.install import install
 
 PYEVM_DEPENDENCY = "py-evm==0.3.0a1"
-
-
-class PostInstallCommand(install):
-    def run(self):
-        """Post-install script which activate argcomplete library."""
-        subprocess.call('register-python-argcomplete trinity >> ~/.bashrc', shell=True)
-        # To start using CLI with autocomplete immediately:
-        subprocess.call('eval "$(register-python-argcomplete trinity)"', shell=True)
-        install.run(self)
 
 
 deps = {
@@ -50,7 +39,7 @@ deps = {
         "mypy_extensions>=0.4.1,<1.0.0",
         "typing_extensions>=3.7.2,<4.0.0",
         "ruamel.yaml>=0.15.87",
-        "argcomplete==1.10.0",
+        "argcomplete>=1.10.0,<2",
     ],
     'test': [
         "hypothesis>=4.24.3,<5",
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -1,11 +1,22 @@
...
 from setuptools import setup, find_packages
+from setuptools.command.install import install
 
 PYEVM_DEPENDENCY = "py-evm==0.3.0a1"
 
 
+class PostInstallCommand(install):
+    def run(self):
+        """Post-install script which activate argcomplete library."""
+        subprocess.call('register-python-argcomplete trinity >> ~/.bashrc', shell=True)
```
carver: I have a couple concerns with this approach:
1. We are creating a `~/.bashrc` file, no matter what OS/shell that the installer uses
2. We append to the bashrc on every install, so if you install it a couple times (including updates?) then you get the eval added a bunch of times to the end.

setup.py
```diff
@@ -1,11 +1,22 @@
...
 import os
+import subprocess
 from setuptools import setup, find_packages
+from setuptools.command.install import install
 
 PYEVM_DEPENDENCY = "py-evm==0.3.0a1"
 
 
+class PostInstallCommand(install):
+    def run(self):
```
carver: I'm not familiar with this `setuptools...install` class. Does `run()` trigger in `pip install`? What about during `pip install -U`?

setup.py
```diff
@@ -1,11 +1,22 @@
...
+        """Post-install script which activate argcomplete library."""
+        subprocess.call('register-python-argcomplete trinity >> ~/.bashrc', shell=True)
+        # To start using CLI with autocomplete immediately:
+        subprocess.call('eval "$(register-python-argcomplete trinity)"', shell=True)
+        install.run(self)
```
carver: Maybe `super().run()`?

setup.py
```diff
@@ -39,6 +50,7 @@
         "mypy_extensions>=0.4.1,<1.0.0",
         "typing_extensions>=3.7.2,<4.0.0",
         "ruamel.yaml>=0.15.87",
+        "argcomplete==1.10.0",
```
carver: As long as they use semantic versioning, then this can be a bit looser, like:
```suggestion
        "argcomplete>=1.10.0,<2",
```

s0b0lev: Thanks, updated.

setup.py
s0b0lev: You are right, also in comments, there was a decision to update the cli documentation page with argcomplete configuration steps.

s0b0lev:
Not sure that this fail https://circleci.com/gh/ethereum/trinity/58068 relates to documentation changes...

docs/api/api.cli.rst
```diff
@@ -100,4 +100,23 @@ The following table shows various combinations of how to use logging in Trinity
...
 .. [5] **Increasing** the per-module log level above the general ``--file-log-level`` is not yet supported
-       (See `issue 689 <https://github.com/ethereum/trinity/issues/689>`_ )
\ No newline at end of file
+       (See `issue 689 <https://github.com/ethereum/trinity/issues/689>`_ )
+
+
+Enabling tab completion
+~~~~~~~~~~~~~~~~~~~~~~~
+
+It is possible to configure tab-complete for trinity greater then v0.1.0a25.
```
cburgdorf: Our docs have a version switcher (even though we do only publish the latest docs at this stage) which also means the text should simply refer to the current version e.g.

>Trinity can be configured to auto complete commands when the `<tab>` key is pressed.

cburgdorf:
>Not sure that this fail https://circleci.com/gh/ethereum/trinity/58068 relates to documentation changes...

It looks like a flaky test unrelated to this change. If you make that one wording change and force push, it will probably go green.

docs/api/api.cli.rst
s0b0lev: Thanks, updated.

docs/api/api.cli.rst
```diff
@@ -100,4 +100,23 @@ The following table shows various combinations of how to use logging in Trinity
...
\ No newline at end of file
+       (See `issue 689 <https://github.com/ethereum/trinity/issues/689>`_ )
+
+
+Enabling tab completion
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Trinity can be configured to auto complete commands when the <tab> key is pressed.
+
+For permanent, not global tab complete activation, use:
```
carver: Out of context of our thread, "not global" doesn't make sense here, I think.

Also, this would have to be re-run if the command arguments change, right? So "permanent" may be confusing.

Maybe something like:

> After installing trinity, to activate tab-completion in future bash prompts, use:

s0b0lev: Thanks! Makes sense :) 

Also, need to re-run command if CLI params were changed.

I test this scenario:
1. Activate argcomplete
2. Install trinity and test cli args tab complete
3. Update trinity cli arg, re-install and it works as expected

s0b0lev:
@cburgdorf I try to re-run build, but it still fails:thinking:

s0b0lev:
@cburgdorf thanks for the advice, didn't notice that `py36-eth2-fixtures` step is disabled in master

cburgdorf:
@s0b0lev yep, me neither :)

Hey, sorry for the hassle but can I ask for one final thing before we can get this in? We've just landed a new release notes process and since this is a really nice feature I'd really like to ensure it is mentioned in the upcoming release notes when we cut the next release.

All you need to do is to rebase this again and then create a file `768.feature.rst` in the `newsfragments` directory with some content such as:

```
Trinity can now autocomplete CLI parameters on ``<tab>``. 
Learn how to activate autocomplete in the :doc:`docs</api/api.cli>`.
```

^-- notice that is rst syntax.

I promise we get this merged right after that :pray:

s0b0lev:
@cburgdorf no problem. 

Added release notes: https://github.com/ethereum/trinity/pull/768/files#diff-89bd2380427aa22f4eb55d9f37fd95e6

cburgdorf:
Meh. Weird unrelated CI fails. I'm triggering re-runs :sweat_smile:

cburgdorf:
Landed! Thank you very much for your contribution! Happy to have you as a contributor :tada:

PR #772, opened 2019-07-09
OrderedTaskPreparation robustness (to support syncing)

carver:
### What was wrong?

When debugging an OTP issue during sync, I wrote a new test, and it kept paying dividends as I kept adding variations. The variation/parameter that caught each issue is combined with the fix in a commit, so it is probably easiest to read this PR's diff commit by commit.

Additionally, I added a feature that will help sync avoid duplicate work: `register_tasks()` now returns the tasks that you registered. If `ignore_duplicates=True`, this is equal to the supplied argument (or an exception would be raised). If `ignore_duplicates=False`, then only the non-duplicates are returned.

The following issues were identified and fixed:
- completing a dependency of a task that's already complete (can happen with tight pruning)
- pruning bug when linking two segments of tasks (like completing C, after A, B, D, & E are registered)
- bug when you try to register duplicates in the same batch, like: `register_tasks((A, A), ignore_duplicates=True)`

### How was it fixed?

In corresponding order:
- Don't mark a dependent task as complete if it was already completed, even if you just completed it's dependency
- Don't try to look up a pruned root if the task is already complete
- If ignoring duplicates, ignore duplicates in the supplied batch, also

### To-Do

#### Cute Animal Picture

### DIFF
--- a/tests/core/task-queue-utils/test_ordered_task_preparation.py
+++ b/tests/core/task-queue-utils/test_ordered_task_preparation.py
@@ -696,7 +696,7 @@ async def test_wait_to_prune_until_yielded():
 @example(task_series=[0, 4, 2, 1, 5], prune_depth=1)
 @pytest.mark.parametrize('ignore_duplicates', (False,))
 @pytest.mark.parametrize('recomplete_idx', (None, 1))
-@pytest.mark.parametrize('batch_size', (1,))
+@pytest.mark.parametrize('batch_size', (1, 3))
 @pytest.mark.asyncio
 async def test_random_pruning(
         ignore_duplicates,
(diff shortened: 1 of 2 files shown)


trinity/_utils/datastructures.py
```diff
@@ -655,7 +675,7 @@ def _mark_one_task_complete(self, task_id: TTaskID) -> Iterable[TTaskID]:
         # resolve tasks that depend on this task
         for depending_task_id in self._roots.get_children(task_id):
             # we already know that this task is ready, so we only need to check completion
```
pipermerriam: This comment seems like it might need to be updated.

tests/core/task-queue-utils/test_ordered_task_preparation.py
```diff
@@ -695,9 +714,19 @@ def fork_prereq(task):
...
+        if idx == recomplete_idx:
+            task_to_mark_finished = task_batch[0] - 1
+            if task_to_mark_finished not in ti._tasks:
+                ti.set_finished_dependency(task_to_mark_finished)
+
         try:
-            ti.register_tasks((task, ))
+            ti.register_tasks(registerable_tasks, ignore_duplicates=ignore_duplicates)
         except DuplicateTasks:
             continue
```
pipermerriam: Should this re-raise in the case that `ignore_duplicates==True`?

PR #732, opened 2019-06-19
State cache

hwwhww:
WIP!!!!

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Clean up commit history

[//]: # (For important changes, please add a new entry to the running release notes PR)
[//]: # (You can find the current one using: https://github.com/ethereum/trinity/pulls?q=is%3Aopen+is%3Apr+label%3A%22Release+Notes%22 )
[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)
- [ ] Add entry to the [release notes PR](https://github.com/ethereum/trinity/pulls?q=is%3Aopen+is%3Apr+label%3A%22Release+Notes%22)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #774, opened 2019-07-09
Fix get state(machine) from non-canonical chain

NIC619:
### What was wrong?

When importing a block, we use head state instead of the state of it's parent block to apply state transition.
- Rationale: Because there could be empty slots between the block and it's parent block, this could save us some work by not having to apply those `per_slot_transition`s again.

However one slot can only map to one state - state of canonical chain. If we are going to process a block from a non-canonical chain, we shouldn't use the head state to apply state transition because the head state is the state of canonical chain.

### How was it fixed?
When importing a block, first check if the block is part of canonical chain.
- If it is, use the head state.
- If not, use the state of it's parent block and apply the `per_slot_transition`s to get the "head state" of the non-canonical chain.

### DIFF
--- a/eth2/beacon/chains/base.py
+++ b/eth2/beacon/chains/base.py
@@ -388,25 +388,27 @@ class BeaconChain(BaseBeaconChain):
                 )
             )
 
-        # Default to use state machine and state at `parent_block.slot`.
-        # Change to the ones at head state slot if `block.slot` is on canonical chain
-        # and newer than head state slot.
-        state_machine = self.get_state_machine(at_slot=parent_block.slot)
-        state = self.get_state_by_slot(parent_block.slot)
         try:
             canonical_root_at_slot = self.get_canonical_block_root(parent_block.slot)
         except BlockNotFound:
             # No corresponding block at this slot which means parent block is not
             # on canonical chain.
-            pass
+            is_new_canonical_block = False
         else:
             is_on_canonical_chain = parent_block.signing_root == canonical_root_at_slot
             is_newer_than_head_state_slot = (
                 self.chaindb.get_head_state_slot() < block.slot
             )
-            if is_on_canonical_chain and is_newer_than_head_state_slot:
-                state_machine = self.get_state_machine()
-                state = self.get_head_state()
+            is_new_canonical_block = (
+                is_on_canonical_chain and is_newer_than_head_state_slot
+            )
+
+        if is_new_canonical_block:
+            state_machine = self.get_state_machine()
+            state = self.get_head_state()
+        else:
+            state_machine = self.get_state_machine(at_slot=parent_block.slot)
+            state = self.get_state_by_slot(parent_block.slot)
 
         state, imported_block = state_machine.import_block(
             block, state, check_proposer_signature=perform_validation
--- a/eth2/beacon/chains/base.py
+++ b/eth2/beacon/chains/base.py
@@ -272,9 +272,7 @@ class BeaconChain(BaseBeaconChain):
             slot = at_slot
         sm_class = self.get_state_machine_class_for_block_slot(slot)
 
-        return sm_class(
-            chaindb=self.chaindb, attestation_pool=self.attestation_pool, slot=slot
-        )
+        return sm_class(chaindb=self.chaindb, attestation_pool=self.attestation_pool)
 
     @classmethod
     def get_genesis_state_machine_class(cls) -> Type["BaseBeaconStateMachine"]:
--- a/eth2/beacon/chains/base.py
+++ b/eth2/beacon/chains/base.py
@@ -388,16 +388,26 @@ class BeaconChain(BaseBeaconChain):
                 )
             )
 
-        head_state_slot = self.chaindb.get_head_state_slot()
-        if head_state_slot >= block.slot:
-            # Importing a block older than the head state. Hence head state can not be used to
-            # perform state transition.
-            prev_state_slot = parent_block.slot
+        # Default to use state machine and state at `parent_block.slot`.
+        # Change to the ones at head state slot if `block.slot` is on canonical chain
+        # and newer than head state slot.
+        state_machine = self.get_state_machine(at_slot=parent_block.slot)
+        state = self.get_state_by_slot(parent_block.slot)
+        try:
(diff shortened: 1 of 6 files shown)


tests/eth2/fixtures-tests/state-fixtures/test_minimal_state.py
```diff
@@ -157,7 +157,7 @@ def execute_state_transtion(test_case, base_db):
 
     post_state = pre_state.copy()
     for block in blocks:
-        sm = sm_class(chaindb, attestation_pool, None, post_state)
+        sm = sm_class(chaindb, attestation_pool, None, state=post_state)
```
mhchia: Would it be more readable to pass all of them in kwargs?
```suggestion
        sm = sm_class(
            chaindb=chaindb,
            attestation_pool=attestation_pool,
            slot=None,
            block=None,
            state=post_state,
        )
```

eth2/README.md
```diff
@@ -114,7 +114,7 @@ The `Chain` represents a single blockchain. One chain might fork by the block nu
 
 ![](https://storage.googleapis.com/ethereum-hackmd/upload_46c3b4a92edeaa18d66f5e2e367f1276.png)
 
-In `BeaconChain` initialization, a `BeaconChainDB` will be set in `BeaconChain`. While processing each block, it will initialize a `BeaconStateMachine` object via `get_state_machine()` function, and use this particular `BeaconStateMachine` to apply state transition of the given block.
+In `BeaconChain` initialization, a `BeaconChainDB` will be set in `BeaconChain`. While processing each block, it will initialize a `BeaconStateMachine` object via `get_state_machine_by_slot()` or `get_state_machine_by_slot` function, and use this particular `BeaconStateMachine` to apply state transition of the given block.
```
hwwhww: I think you meant 
```suggestion
In `BeaconChain` initialization, a `BeaconChainDB` will be set in `BeaconChain`. While processing each block, it will initialize a `BeaconStateMachine` object via `get_state_machine_by_slot()` or `get_state_machine_by_block()` function, and use this particular `BeaconStateMachine` to apply state transition of the given block.
```

eth2/beacon/chains/base.py
```diff
@@ -286,14 +286,27 @@ def get_state_machine_class_for_block_slot(
...
+        return sm_class(
+            chaindb=self.chaindb,
+            attestation_pool=self.attestation_pool,
+            slot=block.slot,
+            block=block,
+        )
+
+    def get_state_machine_by_slot(self, slot: Slot=None) -> 'BaseBeaconStateMachine':
         """
         Return the ``StateMachine`` instance for the given slot number.
```
hwwhww: It's weird that the function is called `get_state_machine_by_slot` while `slot: Slot=None`. Maybe note in the docstring:
```suggestion
        Return the ``StateMachine`` instance for the given `slot`.
        If the given ``slot`` is ``None``, use the head slot number of the canonical chain.
```

eth2/beacon/chains/base.py
```diff
@@ -286,14 +286,27 @@ def get_state_machine_class_for_block_slot(
                 return sm_class
         raise StateMachineNotFound("No StateMachine available for block slot: #{0}".format(slot))
 
-    def get_state_machine(self, at_slot: Slot=None) -> 'BaseBeaconStateMachine':
+    def get_state_machine_by_block(self, block: BaseBeaconBlock=None) -> 'BaseBeaconStateMachine':
```
hwwhww: Need to add `abstractmethod` version.

NIC619:
> But for the `BeaconStateMachine`, you could remove `BeaconStateMachine.state` by replacing it with `chain.get_head_state()` as #732. It will make it easier to keep `state_cache` in `BeaconChain` level.

This makes sense. So we can basically remove `slot`, `block` and `state` from `BeaconStateMachine` and instead passing in `state` to `BeaconStateMachine .import_block()` as in #732 .
Though if we are going to create a `BeaconStateMachine` that does not come with `state`, `get_state_machine_by_slot` seems to do the job just fine, i.e., we don't need `get_state_machine_by_block`

eth2/beacon/state_machines/base.py
```diff
@@ -94,21 +95,30 @@ def __init__(self,
...
+                 block: BaseBeaconBlock=None,
                  state: BeaconState=None) -> None:
         self.chaindb = chaindb
         self.attestation_pool = attestation_pool
         if state is not None:
             self._state = state
+        elif block is not None:
+            self.block = block
         else:
             self.slot = slot
```
ChihChengLiang: Hummm, if we are creating state machine from state or from block, we are passing slot for no reason.

Just thinking should we create these three class methods: `from_state`, `from_block`, `from_slot`.

NIC619:
@hwwhww @ChihChengLiang Rebase after #800 is merged. And thanks to #800 this now becomes a tiny PR.

eth2/beacon/chains/base.py
```diff
@@ -390,16 +388,26 @@ def import_block(
...
+            pass
         else:
-            prev_state_slot = head_state_slot
+            is_on_canonical_chain = parent_block.signing_root == canonical_root_at_slot
+            is_newer_than_head_state_slot = (
+                self.chaindb.get_head_state_slot() < block.slot
+            )
+            if is_on_canonical_chain and is_newer_than_head_state_slot:
+                state_machine = self.get_state_machine()
+                state = self.get_head_state()
```
hwwhww: Refactor idea
```python
        try:
            canonical_root_at_slot = self.get_canonical_block_root(parent_block.slot)
        except BlockNotFound:
            # No corresponding block at this slot which means parent block is not
            # on canonical chain.
            is_new_canonical_block = False
        else:
            is_on_canonical_chain = parent_block.signing_root == canonical_root_at_slot
            is_newer_than_head_state_slot = (
                self.chaindb.get_head_state_slot() < block.slot
            )
            is_new_canonical_block = is_on_canonical_chain and is_newer_than_head_state_slot

        if is_new_canonical_block:
            state_machine = self.get_state_machine()
            state = self.get_head_state()
        else:
            state_machine = self.get_state_machine(at_slot=parent_block.slot)
            state = self.get_state_by_slot(parent_block.slot)
```

PR #775, opened 2019-07-09
Add coverage on config APIs

cburgdorf:
### What was wrong?

We should cover the ``TrinityConfig``, ``Eth1AppConfig`` and ``BeaconAppConfig`` in the docs as these are important pieces that are also exposed to plugins / components.

### How was it fixed?

1. Added / fine tuned API docs directly in source
2. Linked them in our API docs

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/api/api.cli.rst
+++ b/docs/api/api.cli.rst
@@ -1,48 +1,135 @@
 Command Line Interface (CLI)
 ============================
 
+The following is a brief description of all the configurations available on the command line.
+We can also generate an always up-to-date version of them by running ``trinity --help``.
+
 .. code-block:: shell
 
     usage: trinity [-h] [--version] [--trinity-root-dir TRINITY_ROOT_DIR]
-                [-l {debug,info}] [--network-id NETWORK_ID | --ropsten]
-                [--sync-mode {full,light} | --light] [--data-dir DATA_DIR]
-                [--nodekey NODEKEY] [--nodekey-path NODEKEY_PATH]
-                {console,attach} ...
+                [--port PORT] [-l LEVEL] [--stderr-log-level STDERR_LOG_LEVEL]
+                [--file-log-level FILE_LOG_LEVEL]
+                [--network-id NETWORK_ID | --ropsten]
+                [--preferred-node PREFERRED_NODES] [--discv5]
+                [--max-peers MAX_PEERS] [--genesis GENESIS]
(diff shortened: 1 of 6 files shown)


trinity/config.py
```diff
@@ -341,26 +355,7 @@ def log_dir(self) -> Path:
...
-            return self._trinity_root_dir
-        else:
-            return get_xdg_trinity_root()
-
-    @trinity_root_dir.setter
-    def trinity_root_dir(self, value: str) -> None:
-        self._trinity_root_dir = Path(value).resolve()
-
-    @property
-    def trinity_root_dir(self) -> Path:
```
cburgdorf: We had these duplicated :rofl:

docs/api/api.cli.rst
```diff
@@ -1,49 +1,136 @@
...
+                [--enable-experimental-eth1-peer-tracking]
+                [--disable-discovery] [--disable-request-server]
+                [--disable-upnp] [--ethstats]
+                [--ethstats-server-url ETHSTATS_SERVER_URL]
+                [--ethstats-server-secret ETHSTATS_SERVER_SECRET]
+                [--ethstats-node-id ETHSTATS_NODE_ID]
+                [--ethstats-node-contact ETHSTATS_NODE_CONTACT]
+                [--ethstats-interval ETHSTATS_INTERVAL]
+                [--sync-mode {fast,full,beam,light,none}] [--tx-pool]
+                {attach,fix-unclean-shutdown,remove-network-db,db-shell} ...
```
pipermerriam: Can we get a release checklist written up with refreshing this as part of the pre-release part?

trinity/config.py
```diff
@@ -466,6 +465,10 @@ def nodekey(self, value: Union[bytes, PrivateKey]) -> None:
 
     @contextmanager
     def process_id_file(self, process_name: str):  # type: ignore
+        """
+        Context manager API to generate process identification files (pid) in the current
+        :meth:`pid_dir`.
```
pipermerriam: This docstring would be better with a code example showing how it is used.

docs/api/api.cli.rst
cburgdorf: I think we can automate that part so that the doc job simply fails if this doesn't match the output of `trinity --help` anymore.

trinity/config.py
cburgdorf: Added :+1:

PR #777, opened 2019-07-09
Remove visual distraction from docs

cburgdorf:
### What was wrong?

I personally find the theme of the API docs a bit visual distracting (and I say that as someone who usually doesn't give a :hankey: about visuals). Imho, the default Python docs look much cleaner and so I went on to steal some bits.

(I tried stealing the whole theme but it's too heavily customized for a quick win).

Here's how API docs **currently** look.

image

### How was it fixed?

This PR adds a tiny bit of custom css to

- get rid of these gray boxes for methods
- get rid of some margin/padding around method names
- increase size of method/class names
- use mono space font for method parameters

Here is what it would look like if this PR lands

image

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -1,0 +1,20 @@
+tt, code, pre {
+    font-family: monospace, sans-serif;
+    font-size: 96.5%;
+}
+
+.rst-content dl:not(.docutils) tt.descname,
+.rst-content dl:not(.docutils) tt.descclassname,
+.rst-content dl:not(.docutils) tt.descname,
+.rst-content dl:not(.docutils) code.descname,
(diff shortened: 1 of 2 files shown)


cburgdorf:
>Mono-spaced italics font for the parameters looks a bit odd to me,

I took that straight here and liked it (but also didn't think about it much)

https://github.com/python/python-docs-theme/blob/b862f1159388ed3134818e56b065e18b835d652c/python_docs_theme/static/pydoctheme.css#L191-L193

But now I took a second look and noticed that on the live site they actually only do the mono space for the module / class name part. I rolled that one back. The main issue I had was with these gray boxes and big spacing.

PR #784, opened 2019-07-10
Address pr feedback in black demo pr

ralexstokes:
### What was wrong?

Addresses PR feedback on some minor formatting stuff in #581.

### How was it fixed?

Addresses @pipermerriam 's feedback.

#### Cute Animal Picture

### DIFF
--- a/eth2/_utils/ssz.py
+++ b/eth2/_utils/ssz.py
@@ -67,7 +67,7 @@ def validate_ssz_equal(obj_a: BaseBeaconBlock,
     diff = diff_ssz_object(obj_a, obj_b)
     if len(diff) == 0:
         raise TypeError(
-            f"{obj_a_name} ({repr(obj_a)}) != {obj_b_name} ({repr(obj_b)}) but got an empty diff"
+            f"{obj_a_name} ({obj_a!r}) != {obj_b_name} ({obj_b!r}) but got an empty diff"
         )
     longest_field_name = max(len(field_name) for field_name, _, _ in diff)
     diff_error_message = "\n - ".join(
--- a/eth2/_utils/ssz.py
+++ b/eth2/_utils/ssz.py
@@ -30,7 +30,7 @@ def diff_ssz_object(left: BaseBeaconBlock,
                 sub_diff = diff_ssz_object(left_value, right_value)
                 for sub_field_name, sub_left_value, sub_right_value in sub_diff:
                     yield (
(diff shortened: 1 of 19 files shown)


ralexstokes:
pinging @pipermerriam @carver @cburgdorf @lithp as this mostly touches stuff yall are closer to

eth2/_utils/ssz.py
```diff
@@ -67,29 +67,18 @@ def validate_ssz_equal(obj_a: BaseBeaconBlock,
     diff = diff_ssz_object(obj_a, obj_b)
     if len(diff) == 0:
         raise TypeError(
-            "{} ({!r}) != {} ({!r}) but got an empty diff".format(
-                obj_a_name,
-                obj_a,
-                obj_b_name,
-                obj_b,
-            )
+            f"{obj_a_name} ({repr(obj_a)}) != {obj_b_name} ({repr(obj_b)}) but got an empty diff"
```
cburgdorf: This can use the `{expression!r}` syntax to be written as:

>`f"{obj_a_name} ({obj_a!r}) != {obj_b_name} ({obj_b!r}) but got an empty diff"`

trinity/_utils/mp.py
```diff
@@ -87,7 +87,11 @@ def __str__(self) -> str:
 class ChainedExceptionWithTraceback(Exception):
 
     def __init__(self, exc: Exception, tb: TracebackType) -> None:
-        self.tb = '\n"""\n%s"""' % ''.join(traceback.format_exception(type(exc), exc, tb))
+        self.tb = (
+            '\n"""\n'
+            f"{''.join(traceback.format_exception(type(exc), exc, tb))}"
+            '"""'
```
cburgdorf: I wonder if this would be improved with the f-string multi line syntax but I don't understand the odd format in the first place so I don't want to mess with it. Maybe a comment could help :)

ralexstokes: let me look at the f-string multiline syntax

eth2/_utils/ssz.py
ralexstokes: sure

trinity/_utils/mp.py
ralexstokes: it looks kind of confusing w/ the multiline syntax... also not sure why we are formatting this to a string here but i think i'll just leave it

Issue #296, opened 2019-02-20
Wrap `bls.verify` into `validate_bls_signature`


hwwhww:
### What is wrong?
The interface of `bls.verify` is defined in the BLS spec and we'd keep it for test vectors. However, there is a common pattern that we call `bls.verify` and raise `ValidationError` if it returns `False`.
For example:

https://github.com/ethereum/trinity/blob/caf9f3886b3577683dfb5d57aad57146856a755a/eth2/beacon/deposit_helpers.py#L39-L54

and

https://github.com/ethereum/trinity/blob/e93f0e38d124e3a6087dd8ce319dc60a3398a32b/eth2/beacon/state_machines/forks/serenity/block_validation.py#L107-L119

### How can it be fixed

1. We can add a `validate_bls_signature` function in `eth2/beacon/validation.py`, so that we don't need to rewrite the error message format everywhere.
2. Add an extra parameter `validator_index: Validator=None`, so that we can print out the validator index corresponding to the given pubkey in error message. i.e., after #295: `def validate_bls_signature(message_hashes: Hash32, pubkey: BLSPubkey, signature: BLSSignature, domain: int, validator_index: Validator=None) -> None`.

ralexstokes:
seems like we did this, right? if so, let's close it :)

PR #787, opened 2019-07-10
Add Chia BLS

ChihChengLiang:
### What was wrong?

Add Chia BLS

### How was it fixed?

Note that some API change

- Add validate and validate_multiple to distinguish from verify and verify_multiple.

#### Cute Animal Picture


tests/eth2/conftest.py
```diff
@@ -117,7 +120,7 @@ def _get_privkey_for(self, index):
         2. Using ``2**i`` instead of ``i``:
            If using ``i``, the combinations of privkeys would not lead to unique pubkeys.
         """
-        privkey = 2**index
+        privkey = int.from_bytes(hash_eth2(index.to_bytes(32, 'little')), 'little') % 2**32
```
mhchia: Is this used to ensure 1) the generated private key is less than `2**32` and 2) the combinations do not easily collide to the other generated private keys?

tests/eth2/core/beacon/tools/builder/test_validator.py
```diff
@@ -75,4 +75,10 @@ def test_aggregate_votes(votes_count, random, privkeys, pubkeys):
     assert len(voted_index) == len(votes)
 
     aggregated_pubs = bls.aggregate_pubkeys(pubs)
-    assert bls.verify(message_hash, aggregated_pubs, sigs, domain)
+
+    if votes_count != 0:
+        bls.validate(message_hash, aggregated_pubs, sigs, domain)
+    else:
+        # EMPTY_SIGNATURE is consider invalid
```
mhchia: nit: 
```suggestion
        # EMPTY_SIGNATURE is considered invalid
```

eth2/_utils/bls.py
```diff
@@ -0,0 +1,126 @@
...
+    Hash32,
+)
+
+from eth2.beacon.exceptions import (
+    SignatureError,
+)
+
+
+class BaseBLSBackend(ABC):
+
```
ralexstokes: nitpick: remove this newline?

eth2/_utils/bls.py
```diff
@@ -0,0 +1,126 @@
...
```
ralexstokes: nitpick: remove newline?

eth2/_utils/bls_bindings/chia_network/api.py
```diff
@@ -10,13 +10,50 @@
...
+    EMPTY_SIGNATURE,
+)
 from eth_utils import (
     ValidationError,
 )
 
 
-def _privkey_int_to_bytes(privkey: int) -> bytes:
-    return privkey.to_bytes(bls_chia.PrivateKey.PRIVATE_KEY_SIZE, "big")
+def _privkey_from_int(privkey: int) -> 'bls_chia.PrivateKey':
```
ralexstokes: i tried removing the `'` so that we are directly referencing the class and it seems to work

can we update the typing so that we just use the symbol directly?

tests/eth2/conftest.py
ralexstokes: +1 on asking about rationale... if we do keep this change then we need to update the comment about how keys are selected/generated (you can just delete it if its out of date)

also -- i think we can drop the `% 2 ** 32` here right? we never have a number greater or equal to `2**32` here

tests/eth2/fixtures-tests/state-fixtures/test_minimal_state.py
```diff
@@ -59,8 +59,8 @@ def mock_bls(mocker, request):
     if 'noautofixture' in request.keywords:
         return
 
-    mocker.patch('py_ecc.bls.verify', side_effect=mock_bls_verify)
-    mocker.patch('py_ecc.bls.verify_multiple', side_effect=mock_bls_verify_multiple)
+    mocker.patch('eth2._utils.bls.eth2_bls.verify', side_effect=mock_bls_verify)
```
ralexstokes: see other comment about use backends here, rather than mocking

tests/eth2/integration/test_demo.py
```diff
@@ -45,8 +45,8 @@ def mock_bls_verify_multiple(pubkeys,
 
 @pytest.fixture(autouse=True)
 def mock_bls(mocker, request):
-    mocker.patch('py_ecc.bls.verify', side_effect=mock_bls_verify)
-    mocker.patch('py_ecc.bls.verify_multiple', side_effect=mock_bls_verify_multiple)
+    mocker.patch('eth2._utils.bls.eth2_bls.verify', side_effect=mock_bls_verify)
```
ralexstokes: the idea w/ moving to this `BLSBackend` pattern is so that we don't have to mock out bls -- we can just select the BLS backend we want local to this module...

i would think we would rather prefer an import at the top of this file that says something like:

`from eth2._utils.bls import NoOpVerifyBackend as bls`

or something like this

tests/eth2/conftest.py
ChihChengLiang: Yes

ChihChengLiang: Nice catch

ChihChengLiang: Ah, so the purpose here is to generate a valid private key. It should be greater than 0 but less than the curve order, which is a number between `2**254` and `2**255`. 
(It's 32 "bytes" so 2 ** 32 is so wrong 🙈)

Without the modulus, it's really easy to get invalid private key that is greater than the curve order.
```python
>>> from py_ecc.optimized_bls12_381 import curve_order
>>> from eth2._utils.hash import hash_eth2
>>> len([index for index in range(1000) if int.from_bytes(hash_eth2(index.to_bytes(32, 'little')), 'little')> curve_order])
512
```

ChihChengLiang:
@ralexstokes Answer a part of your comments

> 1. do we want to include a BLSBackend for the py_ecc impl? it is a little hard to compare the chia 
backend w/ the py_ecc backend as the py_ecc backend disappears in this PR!

Originally not planing to add them now to speed up the adoption of Chia bls. But happy to add them if that helps the review.

> 2. but i'm wondering if all of the validation code we do in api.py is required?

The bls binding prints only runtime error, which is really hard to debug. Adding the validations helps rule out the known reasons. They are not necessarily to stay in api.py though.

> 3. dropping the bls_chia prefix in api.py.

Done 👍

> 4. what is the rationale for the validate on top of the verify?

The idea is to move toward the pattern described in #296. I like the simplicity of returning bool too. But the `validate` API shows the public key, signature, message, and domain when raising a validation error. In `verify` API those information need to be rewritten again in the catch section of the caller.

ralexstokes:
> @ralexstokes Answer a part of your comments
> 
> > 1. do we want to include a BLSBackend for the py_ecc impl? it is a little hard to compare the chia
> >    backend w/ the py_ecc backend as the py_ecc backend disappears in this PR!
> 
> Originally not planing to add them now to speed up the adoption of Chia bls. But happy to add them if that helps the review.

it isn't urgent but we will almost certainly want it -- i doubt we will have no further use for `py_ecc` ever. we don't have to add it as a backend in this PR but something to keep an eye on

> 
> > 1. but i'm wondering if all of the validation code we do in api.py is required?
> 
> The bls binding prints only runtime error, which is really hard to debug. Adding the validations helps rule out the known reasons. They are not necessarily to stay in api.py though.

cool makes sense to keep it then

> 
> > 1. dropping the bls_chia prefix in api.py.
> 
> Done 👍
> 
> > 1. what is the rationale for the validate on top of the verify?
> 
> The idea is to move toward the pattern described in #296. I like the simplicity of returning bool too. But the `validate` API shows the public key, signature, message, and domain when raising a validation error. In `verify` API those information need to be rewritten again in the catch section of the caller.

gotcha! thanks for explaining

ChihChengLiang:
Hi @ralexstokes @mhchia, this PR is ready to be rereviewed again. Some major changes:

- The directory structure has changed a lot now.
- Added PyECCBackend
- Removed all bls mockers.
- Removed bls-binding test. Replaced it with a test that run through all available backends.

eth2/_utils/bls/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+class Eth2BLS:
+    backend: Type[BaseBLSBackend]
+
+    @classmethod
+    def use(cls, backend: Type[BaseBLSBackend]) -> None:
+        cls.backend = backend
+
+    @classmethod
+    def use_default_backend(cls) -> None:
+        cls.backend = DEFAULT_BACKEND
```
ralexstokes: minor but what do you think about:

```python
def use_default_backend(cls) -> None:
    cls.use(DEFAULT_BACKEND)
```
could be that in the future we have more stuff to do then just set one class var

eth2/_utils/bls/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+
+    @classmethod
+    def use_noop_backend(cls) -> None:
+        cls.backend = NoOpBackend
```
ralexstokes: same about unifying setter logic here

eth2/_utils/bls/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+                f"backend {cls.backend.__name__}\n"
+                f"pubkeys {pubkeys}\n"
+                f"message_hashes {message_hashes}\n"
+                f"signature {signature}\n"
+                f"domain {domain}"
+            )
+
+
+eth2_bls = Eth2BLS
+eth2_bls.use_default_backend()
```
ralexstokes: why not just:

```python
eth2_bls = Eth2BLS() # selects default backend for you
```
```

File: eth2/_utils/bls/__init__.py [hunk 11]
```diff
@@ -0,0 +1,115 @@
...
+            raise SignatureError(
+                f"backend {cls.backend.__name__}\n"
+                f"pubkeys {pubkeys}\n"
+                f"message_hashes {message_hashes}\n"
+                f"signature {signature}\n"
+                f"domain {domain}"
+            )
+
+
+eth2_bls = Eth2BLS
```
ralexstokes: is it wrong to just call this variable `bls`?

ralexstokes: (it would simplify imports for callers)

ChihChengLiang:
Thank you all for the review

Issue #754, opened 2019-06-27
Adopt `newsfragments` pattern currently used in `trio` codebase (and maybe use `towncrier`)


pipermerriam:
### What is wrong?

For each PR we have to also push a commit to the release notes.  I'm not loving the workflow.  I find it easy to forget to do.

### How can it be fixed

Take a look at this: https://github.com/python-trio/trio/tree/master/newsfragments

I believe it uses this tools to aggregate it: https://pypi.org/project/towncrier/

I think this would be a more reliable way to aggregate these changes and it is friendly to 3rd party contributors.

pipermerriam:
cc @lithp @carver @cburgdorf @ralexstokes

carver:
Related: #101

carver:
I think this looks the most like the gitlab-writeup that veox ponted to:

> For this reason, [GitLab generates their change log from contributor-provided snippets](https://docs.gitlab.com/ce/development/changelog.html). The [reasoning section](https://docs.gitlab.com/ce/development/changelog.html#history-and-reasoning) is at the very end, but highlights a few rakes, so not worth skipping.

So probably worth a re-read of that post before changing the process.

carver:
I just did a re-read of the reasoning, and of our conversation in 101. I think maybe some resistance was about the data being structured, which it appears the `towncrier` approach doesn't require (except for `rst` format, which isn't really a new constraint, and the filename format).

So I'm :+1: to switch if others are. It looks quite similar in output to our current release note approach, anyway.

cburgdorf:
I support this and I happily volunteer to set it up if we agree we want it.

pipermerriam:
![make-it-so](https://user-images.githubusercontent.com/824194/60905346-1a9be680-a232-11e9-892f-f967fbcc94cd.jpg)

PR #788, opened 2019-07-10
Setup towncrier to generate release notes

cburgdorf:
### What was wrong?

As explained in #754 we want to automate our release notes process using towncrier.

### How was it fixed?

- Setup towncrier to generate release notes as part of the usual `make release` command. It should just work™
- Release notes can be previewed using `towncrier --draft` at any point
- Updated hints about how to work with this
- Migrated all pending release notes from the `release-notes` branch/pr

### To-Do

- [x] Clean up commit history
- [x] convert pending release notes entries to towncrier fragments

#### Cute Animal Picture


newsfragments/validate_files.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
+
+# Towncrier silently ignores files that do not match the expected ending.
+# We use this script to ensure we catch these as errors in CI.
```
cburgdorf: @carver @pipermerriam I feel that this is something towncrier could probably handle for us but it doesn't and this solution feels ok to me. Without this, it's easy to have things like `4711.docs.rst` go unnoticed (needs to be `4711.doc.rst`)  during a review and then aren't picked up for the release notes generation until someone notices "Oh hey, there is this file sitting there since forever, I guess it never made it into the release notes". This script ensures, this doesn't happen because it will crash CI if you commit unexpected files to the `newsfragments` directory.

pipermerriam: :+1: seems like a good simple way to enforce this in CI.

newsfragments/validate_files.py
```diff
@@ -0,0 +1,31 @@
...
+
+ALLOWED_EXTENSIONS = (
+    '.feature.rst',
+    '.bugfix.rst',
+    '.doc.rst',
+    '.removal.rst',
+    '.misc.rst',
+)
+
+ALLOWED_FILES = (
```
pipermerriam: *nitpick* `set` would be slightly more appropriate in these two collections.

pyproject.toml
```diff
@@ -0,0 +1,12 @@
...
+# - PRs should drop a file like "issuenumber.feature" in newsfragments
+#   (or "bugfix", "doc", "removal", "misc"; misc gets no text, we can
+#    customize this)
+# - At release time after bumping version number, run: towncrier
+#   (or towncrier --draft)
+package = "trinity"
+filename = "docs/release_notes.rst"
+directory = "newsfragments"
+underlines = ["-", "~", "^"]
+issue_format = "`#{issue} <https://github.com/ethereum/trinity/issues/{issue}>`__"
```
carver: *nitpick* add a trailing newline?

Makefile
```diff
@@ -58,7 +60,14 @@ package: clean
...
 	git config commit.gpgSign true
-	bumpversion $(bump)
+	# Let UPCOMING_VERSION be the version that is used for the current bump
+	$(eval UPCOMING_VERSION=$(shell bumpversion $(bump) --dry-run --list | grep new_version= | sed 's/new_version=//g'))
+	# Now generate the release notes to have them included in the release commit
+	towncrier --yes --version $(UPCOMING_VERSION)
+	# We need --allow-dirty because of the generated release_notes file but it is safe because the
+	# previous dry-run runs *without* --allow-dirty which ensures it's really just the release notes
+	# file that we are allowing to sit here dirty, waiting to get included in the release commit.
+	bumpversion --allow-dirty $(bump)
```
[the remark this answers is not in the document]
cburgdorf: The minimalist in me prefers having the release notes in the actual release commit (the one that is also tagged). I did a quick scan of other popular projects and there is evidence for both schemes (single commit, separate commits).

I'd say let's roll with this for now and if we come up with a good reason why we would want to have this in two separate commits, we can change it later.

pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[tool.towncrier]
+# Usage:
+# - PRs should drop a file like "issuenumber.feature" in newsfragments
+#   (or "bugfix", "doc", "removal", "misc"; misc gets no text, we can
+#    customize this)
+# - At release time after bumping version number, run: towncrier
+#   (or towncrier --draft)
```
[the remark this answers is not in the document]
cburgdorf: Good point. It's a leftover from copying what trio had :+1:

docs/index.rst
```diff
@@ -10,7 +10,7 @@ Table of contents
 
    introduction
    quickstart
-   release_notes/index
```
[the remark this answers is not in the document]
cburgdorf: Yep, that's been an leftover from the days where py-evm and trinity where combined and one had to navigate downwards to the release notes of either one.

pipermerriam:
Just stumbled on https://pypi.org/project/rstcheck/ which might work for validating well-formed newsfragment files.  Didn't actually try it.

PR #790, opened 2019-07-10
Add trio based service implementation

pipermerriam:
### What was wrong?

We use the "Service" abstraction heavily in our existing code.  We need something loosely equivalent that can be used with `trio`

### How was it fixed?

Added a re-imagined implementation of the *Service* concept written in `trio`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

pet-rat-photography-diane-ozdamar-7

### DIFF
--- a/p2p/trio_service.py
+++ b/p2p/trio_service.py
@@ -40,15 +40,19 @@ class ServiceAPI(ABC):
 
         .. code-block: python
 
-            # run the service and blocks until service finishes
-            await ManagerAPI.run_service(service)
-
-            # run the service in the background using a context manager
+            # 1. run the service in the background using a context manager
             async with run_service(service) as manager:
                 # service runs inside context block
                 ...
                 # service cancels and stops when context exits
             # service will have fully stopped
+
+            # 2. run the service blocking until completion
+            await Manager.run_service(service)
+
+            # 3. create manager and then run service blocking until completion
+            manager = Manager(service)
+            await manager.run()
         """
         ...
 
@@ -282,7 +286,6 @@ class Manager(ManagerAPI):
     @classmethod
     async def run_service(cls, service: ServiceAPI) -> None:
         manager = cls(service)
-        service.manager = manager
         await manager.run()
 
     async def run(self) -> None:
--- a/p2p/trio_service.py
+++ b/p2p/trio_service.py
@@ -1,8 +1,11 @@
 from abc import ABC, abstractmethod
+import functools
 import logging
 import sys
 from types import TracebackType
 from typing import Any, Callable, Awaitable, Optional, Tuple, Type, AsyncIterator
+
+from mypy_extensions import VarArg, KwArg
 
 from async_generator import asynccontextmanager
 
@@ -11,71 +14,166 @@ import trio
 import trio_typing
 
 
+class ServiceException(Exception):
+    """
+    Base class for Service exceptions
+    """
+    pass
+
+
+class LifecycleError(ServiceException):
+    """
+    Raised when an action would violate the service lifecycle rules.
+    """
+    pass
+
+
+class ServiceAPI(ABC):
+    manager: 'ManagerAPI'
+
+    @abstractmethod
+    async def run(self) -> None:
+        """
+        This method is where all of the Service class logic should be
+        implemented.  It should **not** be invoked by user code but instead run
+        with either:
+
+        .. code-block: python
+
+            # run the service and blocks until service finishes
+            await ManagerAPI.run_service(service)
+
+            # run the service in the background using a context manager
+            async with run_service(service) as manager:
+                # service runs inside context block
+                ...
+                # service cancels and stops when context exits
+            # service will have fully stopped
+        """
+        ...
+
+
 class ManagerAPI(ABC):
     @property
     @abstractmethod
-    def has_started(self) -> bool:
+    def is_started(self) -> bool:
+        """
+        Return boolean indicating if the underlying service has been started.
+        """
         ...
 
     @property
     @abstractmethod
     def is_running(self) -> bool:
+        """
+        Return boolean indicating if the underlying service is actively
+        running.  A service is considered running if it has been started and
+        has not yet been stopped.
+        """
         ...
 
     @property
     @abstractmethod
     def is_cancelled(self) -> bool:
+        """
+        Return boolean indicating if the underlying service has been cancelled.
+        This can occure externally via the `cancel()` method or internally due
+        to a task crash or a crash of the actual :meth:`ServiceAPI.run` method.
+        """
         ...
 
     @property
     @abstractmethod
     def is_stopped(self) -> bool:
+        """
+        Return boolean indicating if the underlying service is stopped.  A
+        stopped service will have completed all of the background tasks.
+        """
         ...
 
     @property
     @abstractmethod
     def did_error(self) -> bool:
+        """
+        Return boolean indicating if the underlying service threw an exception.
+        """
         ...
 
     @abstractmethod
     def cancel(self) -> None:
+        """
+        Trigger cancellation of the service.
+        """
+        ...
+
+    @abstractmethod
+    async def stop(self) -> None:
+        """
+        Trigger cancellation of the service and wait for it to stop.
+        """
         ...
 
     @abstractmethod
     async def wait_started(self) -> None:
+        """
+        Wait until the service is started.
+        """
         ...
 
     @abstractmethod
     async def wait_cancelled(self) -> None:
+        """
+        Wait until the service is cancelled.
+        """
         ...
 
     @abstractmethod
     async def wait_stopped(self) -> None:
+        """
+        Wait until the service is stopped.
+        """
+        ...
+
+    @classmethod
+    @abstractmethod
+    async def run_service(cls, service: ServiceAPI) -> None:
+        """
+        Run a service
+        """
         ...
 
     @abstractmethod
-    async def _run_daemon_task(self,
-                               coro: Callable[..., Awaitable[Any]],
-                               *args: Any,
-                               name: str = None) -> None:
+    async def run(self) -> None:
+        """
+        Run a service
+        """
         ...
 
-
-LogicFnType = Callable[[ManagerAPI], Awaitable[Any]]
-
-
-class Service(ABC):
-    async def start(self, nursery: trio_typing.Nursery) -> 'ServiceManager':
-        manager = ServiceManager(self)
-        nursery.start_soon(manager.run)
-        return manager
-
+    @trio_typing.takes_callable_and_args
     @abstractmethod
-    async def run(self, manager: ManagerAPI) -> None:
+    async def run_task(self,
+                       async_fn: Callable[[VarArg()], Awaitable[Any]],
+                       *args: Any,
+                       daemon: bool = False,
+                       name: str = None) -> None:
+        """
+        Run a task in the background.  If the function throws an exception it
+        will trigger the service to be cancelled and be propogated.
+
+        If `daemon == True` then the the task is expected to run indefinitely
+        and will trigger cancellation if the task finishes.
+        """
         ...
 
 
-def as_service(logic_fn: LogicFnType) -> Type[Service]:
+LogicFnType = Callable[[ManagerAPI, VarArg(), KwArg()], Awaitable[Any]]
+
+
+class Service(ServiceAPI):
+    pass
+
+
+def as_service(service_fn: LogicFnType) -> Type[Service]:
     """
     Create a service out of a simple function
     """
@@ -84,16 +182,18 @@ def as_service(logic_fn: LogicFnType) -> Type[Service]:
             self._args = args
             self._kwargs = kwargs
 
-        async def run(self, manager: ManagerAPI) -> None:
-            await logic_fn(manager, *self._args, **self._kwargs)  # type: ignore
+        async def run(self) -> None:
+            await service_fn(self.manager, *self._args, **self._kwargs)
 
-    _Service.__name__ = logic_fn.__name__
-    _Service.__doc__ = logic_fn.__doc__
+    _Service.__name__ = service_fn.__name__
+    _Service.__doc__ = service_fn.__doc__
     return _Service
 
 
-class ServiceManager(ManagerAPI):
-    logger = logging.getLogger('p2p.trio_service.ServiceManager')
+class Manager(ManagerAPI):
+    logger = logging.getLogger('p2p.trio_service.Manager')
+
+    _service: ServiceAPI
 
     _run_error: Optional[Tuple[
         Optional[Type[BaseException]],
@@ -101,8 +201,21 @@ class ServiceManager(ManagerAPI):
         Optional[TracebackType],
     ]] = None
 
-    def __init__(self, logic: Service) -> None:
-        self.logic = logic
+    # A nursery for system tasks.  This nursery is cancelled in the event that
+    # the service is cancelled or exits.
+    _system_nursery: trio_typing.Nursery
+
+    # A nursery for sub tasks and services.  This nursery is cancelled if the
+    # service is cancelled but allowed to exit normally if the service exits.
+    _task_nursery: trio_typing.Nursery
+
+    def __init__(self, service: ServiceAPI) -> None:
+        if hasattr(service, 'manager'):
+            raise LifecycleError("Service already has a manager.")
+        else:
+            service.manager = self
+
+        self._service = service
 
         # events
         self._started = trio.Event()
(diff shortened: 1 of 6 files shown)


p2p/trio_service.py
```diff
@@ -0,0 +1,256 @@
...
+
+@asynccontextmanager
+async def run_service(service: Service) -> AsyncIterator[ManagerAPI]:
+    async with trio.open_nursery() as nursery:
+        manager = await service.start(nursery)
+        try:
+            yield manager
+        finally:
+            manager.cancel()
+            await manager.wait_stopped()
```
pipermerriam: This should potentially be a single method like:

```python
async def stop(self) -> None:
    self.cancel()
    await self.wait_stopped()
```

p2p/trio_service.py
```diff
@@ -0,0 +1,256 @@
...
+        # If an error occured, re-raise it here
+        if self.did_error:
+            _, exc_value, exc_tb = self._run_error
+            raise exc_value.with_traceback(exc_tb)
+
+    #
+    # Event API mirror
+    #
+    @property
+    def has_started(self) -> bool:
```
pipermerriam: Should maybe be `is_started` for congruence with the other APIs

p2p/trio_service.py
```diff
@@ -0,0 +1,256 @@
...
+
+    @abstractmethod
+    async def _run_daemon_task(self,
+                               coro: Callable[..., Awaitable[Any]],
+                               *args: Any,
+                               name: str = None) -> None:
+        ...
+
+
+LogicFnType = Callable[[ManagerAPI], Awaitable[Any]]
```
pipermerriam: TODO: This type isn't quite right.  What we want is the definition of `def fn(manager: ManagerAPI, *args: Any, **kwargs: Any)` but I don't know how to do that with mypy :(

carver:
Sorry, I've got to wrap early today, but I'll try to pick this up first thing tomorrow.

p2p/trio_service.py
```diff
@@ -0,0 +1,256 @@
...
+    @abstractmethod
+    async def wait_cancelled(self) -> None:
+        ...
+
+    @abstractmethod
+    async def wait_stopped(self) -> None:
+        ...
+
+    @abstractmethod
+    async def _run_daemon_task(self,
```
cburgdorf: I brought this up in the past but we couldn't agree on something. With a fresh service API start, let me try to bring this up again:

I would like us to not use a leading underscore for APIs that aren't private in the traditional sense of how programming languages treat *private* as an access modifier. A private API (in all programming languages that have the concept of *private*) is an API that can only be called by the class itself. Hence an `abstract` API by definition can not be private because it is intended to be implemented by a child class.

I'm not trying to split hairs here but instead fix real problems this creates with docs and IDEs. E.g. in our docs, private members will be ignored (naturally). But an API that is `abstract` is by definition nothing that should be left undocumented because it needs to be implemented by child classes.

I've found that often when we use this pattern of abstract + underscore, what we really mean is *protected* as in: Only be visible to child classes. Python doesn't have a concept of *protected* (even less than that of *private*  which at least exists by convention) but maybe we could just use a prefix such as `protected_` or something else for scenarios such as this?

p2p/trio_service.py
```diff
@@ -0,0 +1,256 @@
...
+        async def run(self, manager: ManagerAPI) -> None:
+            await logic_fn(manager, *self._args, **self._kwargs)  # type: ignore
+
+    _Service.__name__ = logic_fn.__name__
+    _Service.__doc__ = logic_fn.__doc__
+    return _Service
+
+
+class ServiceManager(ManagerAPI):
+    logger = logging.getLogger('p2p.trio_service.ServiceManager')
```
cburgdorf: Not a big deal but I wonder if it would make sense to put this under its own top level module possibly as a launch pad to eventually move it into `eth_utils` when it has stabilized. I know our current service abstraction lives under p2p as well but its more like a *"damn p2p is the lowest component that depends on it"* kind of thing rather than a *"oh this really makes sense to live under p2p"* kind of thing :)

p2p/trio_service.py
```diff
@@ -0,0 +1,256 @@
...
+        """
+        self._started.set()
+        try:
+            await self.logic.run(self)
+        except Exception as err:
+            self.logger.debug(
+                '%s: _handle_run got error, storing exception and setting cancelled',
+                self
+            )
+            self.logger.exception('GOT THIS')
```
cburgdorf: Is this a left over log? The all caps confuse me :)

p2p/trio_service.py
```diff
@@ -0,0 +1,256 @@
...
+            self._run_error = sys.exc_info()
+        finally:
+            self.logger.debug('%s: _handle_run triggering service cancellation', self)
+            self.cancel()
+
+    async def run(self) -> None:
+        if self._run_lock.locked():
+            raise Exception("TODO: run lock already engaged")
```
cburgdorf: Todo logs

p2p/trio_service.py
```diff
@@ -0,0 +1,256 @@
...
+            # Open a nursery
+            async with trio.open_nursery() as nursery:
+                self._nursery = nursery
+
+                nursery.start_soon(self._handle_cancelled, nursery)
+                nursery.start_soon(self._handle_run)
+
+                # This wait is not strictly necessary as this context block
+                # will not exit until the background tasks have completed.
+                await self.wait_cancelled()
```
cburgdorf: Maybe just get rid of this. I feel like what the comment explains is one of the core values of trio so I wonder if this might just leave the code less idiomatic?

p2p/trio_service.py
```diff
@@ -0,0 +1,256 @@
...
+    @property
+    def did_error(self) -> bool:
+        return self._run_error is not None
+
+    #
+    # Control API
+    #
+    def cancel(self) -> None:
+        if not self.has_started:
+            raise Exception("TODO: never started")
```
cburgdorf: Todo comment

p2p/trio_service.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+LogicFnType = Callable[[ManagerAPI], Awaitable[Any]]
+
+
+class Service(ABC):
+    async def start(self, nursery: trio_typing.Nursery) -> 'ServiceManager':
+        manager = ServiceManager(self)
+        nursery.start_soon(manager.run)
+        return manager
```
jannikluhn: What's the reason of separating the service from the manager? Is it to avoid polluting the service namespace with methods from the manager?

I'm worried that it could get annoying to keep track of both the service and the manager, especially if something has multiple sub services. This gets worse if the services are created in an `__init__` method, but one only gets ahold of the managers when they are stated (presumably in `run`).

How about storing the manager in the service instance (`my_service.manager`) instead?

p2p/trio_service.py
```diff
@@ -0,0 +1,256 @@
...
+            )
+        else:
+            self.logger.debug(
+                "Daemon task '%s' exited.",
+                name or coro,
+            )
+        finally:
+            self.cancel()
+
+    def run_daemon_task(self,
```
jannikluhn: In addition to `run_daemon_task`, would it make sense to have something like `run_daemon_service`? With the difference that it not only runs it, but also cancels it when the parent task is cancelled.

p2p/trio_service.py
pipermerriam: This is just a copy/paste mistake.  It's supposed to not have an underscore prefix.

p2p/trio_service.py
pipermerriam: @jannikluhn 

Yes, intent is to keep the service classes free of the manager APIs.

I *think* you're right that the 99% use case is going to be wanting access via `self.manager`.  I'll update so that this is the case.

p2p/trio_service.py
pipermerriam: @cburgdorf I think I'll probably extract this to it's own package.  I guess we could elevate it out of the `p2p` namespace but I'm not inclined to do that just yet.

p2p/trio_service.py
pipermerriam: Yup, will remove.

p2p/trio_service.py
pipermerriam: :+1: will convert to a new service specific exception.

p2p/trio_service.py
pipermerriam: :+1: I'll replace with something like a comment to ensure it is clear that the code blocks at that point.

p2p/trio_service.py
pipermerriam: @jannikluhn yes, just never got around to adding it, are you up for doing that as part of your work (if you run into that use case).

jannikluhn: Sure, happy to implement it!

cburgdorf:
Noting that this should also add a release notes entry :)

tests/p2p-trio/test_service.py
```diff
@@ -0,0 +1,109 @@
...
+    async def ServiceTest(manager):
+        await trigger_error.wait()
+        raise RuntimeError("Service throwing error")
+
+    service = ServiceTest()
+    manager = ServiceManager(service)
+
+    async def do_service_run():
+        with pytest.raises(RuntimeError, match="Service throwing error"):
+            await manager.run()
```
carver: If we didn't capture this exception, where would it be bubbling up in `do_service_lifecycle_check()'? Just curious.

p2p/trio_service.py
jannikluhn: Question after having understood the code and trio a little bit better:

> also cancels it when the parent task is cancelled.

It seems like `run_daemon_task` already handles this out of the box: It will run the task in the service's nursery. The cancel event is handled by the manager and if set will cancel the nursery's cancel scope, which in turn will cancel everything in the nursery. So the only difference between `run_daemon_task` and `run_daemon_service` would be an easier to use interface (which might still be valuable, would have to see it in practice). Is this correct or am I missing something?

p2p/trio_service.py
```diff
@@ -0,0 +1,256 @@
...
+                err,
+                exc_info=True,
+            )
+        else:
+            self.logger.debug(
+                "Daemon task '%s' exited.",
+                name or coro,
+            )
+        finally:
+            self.cancel()
```
jannikluhn: this seems odd. Why would we cancel ourselves if a daemon task finishes?

p2p/trio_service.py
jannikluhn: Unrelated to the `run_daemon_service` discussion: Do we actually want to have `run_daemon_task`? It seems like the main difference to just using `self.manager.nursery.start_soon` would be that an exception doesn't make the parent service fail. But I can't think of a case where this is actually desired -- at least not in this not very explicit fashion.

p2p/trio_service.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+class Service(ABC):
+    async def start(self, nursery: trio_typing.Nursery) -> 'ServiceManager':
+        manager = ServiceManager(self)
+        nursery.start_soon(manager.run)
+        return manager
+
+    @abstractmethod
+    async def run(self, manager: ManagerAPI) -> None:
```
jannikluhn: Is it possible to pass parameters to `run`? For most we can probably just use `__init__`, but it would be nice to support running services with [Nursery.start](https://trio.readthedocs.io/en/latest/reference-core.html#trio.Nursery.start) which requires the parameter `task_status=trio.TASK_STATUS_IGNORED` if I understand correctly.

setup.py
```diff
@@ -47,16 +49,21 @@
...
-        # only needed for p2p
-        "pytest-asyncio-network-simulator==0.1.0a2;python_version>='3.6'",
         # only for eth2
         "ruamel.yaml>=0.15.87,<0.16",
     ],
+    'test-asyncio': [
+        "pytest-asyncio>=0.10.0,<0.11",
+    ],
+    'test-trio': [
+        'pytest-trio==0.5.2',
```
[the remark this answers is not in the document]
pipermerriam: Yes, pytest-asyncio doesn't play well with others.  I'll comment and link to the github issue (that the pytest-asyncio developers seem to be ignoring)

tests/p2p-trio/test_service.py
```diff
@@ -0,0 +1,109 @@
...
+
+        assert manager.has_started is True
+        assert manager.is_running is True
+        assert manager.is_cancelled is False
+        assert manager.is_stopped is False
+
+        # trigger the service to exit
+        trigger_exit_condition_fn()
+
+        with trio.fail_after(10):
```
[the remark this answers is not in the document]
pipermerriam: I don't think it does, will shorten.

tests/p2p-trio/test_service.py
```diff
@@ -0,0 +1,109 @@
...
+            await manager.wait_cancelled()
+
+        assert manager.has_started is True
+        assert manager.is_running is True
+        assert manager.is_cancelled is True
+        assert manager.is_stopped is False
```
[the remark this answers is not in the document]
pipermerriam: `Manager.cancel()` is synchronous, calling `set()` on a `trio.Event`.  The actual cancellation won't occur until an `await` happens to yield control flow, and even then there will be a few more `await/yield`s that happen because the inner nursery will also have to exit it's async context manager.  Only after those things have occured will the stop event get set causing `is_stoppped` to be truthy.

p2p/trio_service.py
pipermerriam: I'm not 100% sure this is right but the idea was that a *daemon* task is one that doesn't finish, but I think I see your point.  cc @carver what do you think about allowing them to finish and only bubbling cancellation if they error?

p2p/trio_service.py
pipermerriam: :+1: I'll extend this API to support the `Nursery.start` pattern.

p2p/trio_service.py
pipermerriam: Good question.  Lemme try to suss out the use cases.

- process that needs to endure for the full lifespan of the service:
    - event bus request/response server
    - queue/channel listener that consumes items and does things
- short lived process:
    - fetch some data from a peer and put it on a queue/channel
    - import a single block in beam sync

I think that for both of the above we want the service to crash if the task fails.  For the first one that needs to always be running I think we also need to crash the service if it terminates.

The theoretical 3rd use case is something like a long lived process that eventually terminates.  I think that rather than have a 3rd type of task, that this type of process should just always be modeled as another service.

This yields the following API:

- daemon:
    - cancel if task throws error or finishes
- task:
    - cancel if task throws error
- service
    - just a special case of *task* 

The API I think I want to change this to:

```
class ManagerAPI:
    async def run_task(coro, *args, daemon=False):
        ...
    async def run_service(service, *, daemon=False):
        ...
```

In either of the above, if `daemon == True` then cancellation is triggered if the thing finishes.

tests/p2p-trio/test_service.py
pipermerriam: It throws at the point that the nursery context exits.

pipermerriam:
Here's the API I'm thinking about:

```python
    # blocking
    await Manager.run_service(service)

    # context
    async with run_service(service) as manager:
        # do things
        pass

    async with trio.open_nursery() as nursery:
        # ensure that service has started before doing next thing
        manager = await nursery.start(service)
```

p2p/trio_service.py
pipermerriam: I decided against this. It complicates the creation of services that don't need this and it's an easy mechanism to implement on your own with a service class simply by exposing an event that you wait for.

p2p/trio_service.py
pipermerriam: You are correct, `run_service` would very likely be built on top of `run_task`, though it make make sense to have some extra niceties to track sub-services so that we can introspect service hierarchy.

pipermerriam:
Test failure can be found on: https://github.com/ethereum/trinity/pull/799

pipermerriam:
lint failures are due to #799

pipermerriam:
@jannikluhn @carver @cburgdorf I've re-worked this a bit.  Wanting another round of *light* review.  I don't think we should futz with the API specifics much more so really just looking for major problems and obvious bugs.

I expect once @jannikluhn starts using this we'll shake out the API improvements incrementally.

p2p/trio_service.py
```diff
@@ -0,0 +1,431 @@
...
+    async def run(self) -> None:
+        """
+        This method is where all of the Service class logic should be
+        implemented.  It should **not** be invoked by user code but instead run
+        with either:
+
+        .. code-block: python
+
+            # run the service and blocks until service finishes
+            await ManagerAPI.run_service(service)
```
jannikluhn:
```suggestion
            await Manager.run_service(service)
```

PR #791, opened 2019-07-10
Updates to v0.8.1

ralexstokes:
Final round of updates to get Trinity up to spec v0.8!

Some things blocked on other work:

~1. Updating BLS so that we can provide a `Domain`, instead of a `int`. Need to update `py_ecc` and `get_domain`, `compute_domain` helpers.~

~NOTE: blocked on releasing https://github.com/ethereum/py_ecc/pull/78~

~2. Update the `justification_bits` pending `py-ssz` updates.~

~3. Use the new `py-ssz` types for `List`, `BitVector` and `BitList`. `py-ssz` updates in https://github.com/ethereum/py-ssz/pull/77.~

~4. Need to fix issue w/ field collision on `root`; see: https://github.com/ethereum/py-ssz/issues/79~

~Blocked on releasing new py-ssz from here: https://github.com/ethereum/py-ssz/commit/4779ce8b2dced0ac575468ff99e69a0073853f98~

#### Cute Animal Picture


ralexstokes:
this is ready for review -- should have `trinity` up to spec `v0.8.1`

eth2/beacon/types/validators.py
```diff
@@ -27,7 +27,7 @@
 )
 
 
-def _round_down_to_previous_multiple(amount: int, increment: int) -> int:
+def round_down_to_previous_multiple(amount: int, increment: int) -> int:
```
hwwhww: 1. What do you mean by "multiple" here?
2. It seems good to use:
```suggestion
def round_down_to_previous_multiple(amount: Gwei, increment: Gwei) -> Gwei:
```
since the names of the variable `amount` and `increment` imply they are in balance units.

tests/eth2/conftest.py
```diff
@@ -172,14 +170,16 @@ def _key_cache(request, _should_persist_bls_keys):
 
     Keys are generated on-demand and cached after creation.
     """
+    cache_key = f"eth2/bls/key-cache/{bls.backend.__name__}"
```
hwwhww: good catch 👍

eth2/beacon/attestation_helpers.py
```diff
@@ -99,9 +99,9 @@ def validate_indexed_attestation_aggregate_signature(state: BeaconState,
     )
 
 
-def validate_indexed_attestation(state: BeaconState,
+def is_valid_indexed_attestation(state: BeaconState,
```
ChihChengLiang: Why are we renaming this one? We kind of accidentally established an implicit convention that a function that starts with `is_valid` returns True; the one starts with `validate` returns None and raises an exception?

eth2/beacon/helpers.py
```diff
@@ -12,6 +13,8 @@
     Hash32,
 )
 
+from py_ecc.bls.typing import Domain
```
ChihChengLiang: What do you think we wrap py_ecc `Domain` type in bls-utils, and import from there?

It looks like the Domain type will eventually belong to a module `eth-bls` but not only to py_ecc.

tests/eth2/conftest.py
ChihChengLiang: Second that! Want the cache_key to take backend into account so badly.

eth2/beacon/operations/pool.py
```diff
@@ -4,27 +4,27 @@
 from eth_typing import Hash32
 
 
-Root = Hash32
+HashTreeRoot = Hash32
```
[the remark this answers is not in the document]
ralexstokes: the thinking was to make it clear to users of the `OperationPool` that it is tracking objects by the `hash_tree_root`, not just some hash... i wouldn't be against deprecating this type but i would hope it adds some additional clarity, esp. to people who may not be that familiar with all of the eth2 stuff

tests/eth2/core/beacon/operations/test_pool.py
```diff
@@ -9,7 +9,7 @@
 
 def mk_attestation(index, sample_attestation_params):
     return Attestation(**sample_attestation_params).copy(
-        custody_bitfield=index.to_bytes(16, byteorder="big"),
+        custody_bits=(True,) * 8 * 16,
```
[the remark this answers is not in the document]
ralexstokes: agreed :)

eth2/beacon/types/validators.py
ralexstokes: 1. it "rounds `amount` down to the previous multiple of `increment`"

e.g. in terms of 1 ETH increment, it goes 1 ETH ~> 1 ETH, 1.1 ETH ~> 1 ETH, 1.9 ETH ~> 1 ETH, etc.

happy to update the name if there is a clearer one

2. given how this function is used across the codebase, i was aiming to skip an additional use of `Gwei` just to satisfy `mypy`. there are two uses at the moment and they both go on to additional math -- if you are concerned about the type safety i would think it is better to just encapsulate the `effective_balance` calculation (which is what this is part of anyway and imo may be a better refactoring here)

eth2/beacon/attestation_helpers.py
ralexstokes: i was following the spec but i'll go in and update it so that it follows the work you did w/ the BLS backends

eth2/beacon/helpers.py
ralexstokes: yeah, i wasn't too sure on this one -- given that it is likely to change, i'm thinking i'll just keep it as is -- if you feel more strongly about this, let me know and i'll happy to update it in a future PR

PR #778, opened 2019-07-09
Group members in docs by classmethods, methods, attributes

cburgdorf:
### What was wrong?

One of the things that annoys me when looking at the API docs is that methods, classmethods and attributes do visually not stand much apart. 

image

By default sphinx orders all members alphabetically and so methods, classmethods and attributes end up mixed and it's really hard to tell them apart.

### How was it fixed?

We can combine two things to improve the situation:

1. We can tell sphinx to order members `groupwise`

2. We can add css pseudo elements to inject headlines for each group

image

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Clean up commit history

[//]: # (For important changes, please add a new entry to the running release notes PR)
[//]: # (You can find the current one using: https://github.com/ethereum/trinity/pulls?q=is%3Aopen+is%3Apr+label%3A%22Release+Notes%22 )
[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)
- [ ] Add entry to the [release notes PR](https://github.com/ethereum/trinity/pulls?q=is%3Aopen+is%3Apr+label%3A%22Release+Notes%22)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #794, opened 2019-07-11
Fix bug in API docs layout

cburgdorf:
### What was wrong?

First off, apologies, I spend way too much time going down this rabbit whole. Been knee deep debugging sphinx for a while :sweat_smile: 

So yesterday #778 introduced a feature for our API docs where members are nicely grouped. Unfortunately after merging #775 I've found out that sphinx groups class methods, static methods and methods all together alphabetically. 

This then leads to this clearly wrong and misleading layout.

image

It looks like a bug on Sphinx side and I've spend a while trying to come up with a monkey patch solution (+ intention to contribute upstream) but eventually I gave up and went for something simpler (for now)

### How was it fixed?

As Sphinx keeps class methods, static methods and methods all alphabetically sorted within one group, we follow the same and only separate methods and attributes.

image

This is a pure CSS fix that simply ensures the different classes for staticmethods, classmethods and methods are all treated equally when it comes to ordering and headlines.

It upsets me to the bone but I had to limit the losses.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -19,23 +19,25 @@ tt, code, pre {
   padding: 0;
 }
 
-/* We instruct sphinx to order members groupwise and use the following css to add headlines */
+/* We instruct sphinx to order members groupwise and use the following css to add headlines.
+Notice that while the css differentiates between methods, classmethods and staticmethods,
+unfortunately, sphinx groups these all alphabetically as methods which means we only
+display them as one "methods" group. */
 
 dl.attribute:before {
   content: "Attributes";
 }
 
-dl.method:before {
+dl.method:before,
+dl.classmethod:before,
+dl.staticmethod:before {
   content: "Methods";
-}
-
-dl.classmethod:before {
-  content: "Classmethods";
 }
 
 dl.method:before,
 dl.attribute:before,
-dl.classmethod:before {
+dl.classmethod:before,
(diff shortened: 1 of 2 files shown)


pipermerriam:
I *think* I'd rather go back to just having them not be grouped....

cburgdorf:
>I *think* I'd rather go back to just having them not be grouped....

Can you elaborate why you think so? I personally think that grouping the members is

1. A big improvement. Without methods and attributes mixed, and with both of them being visually so close together, I feel it is easy for someone reading the docs to constantly get confused about if something is a property or a method.

2. Pretty standard for docs

Here are some popular examples.

**Tensorflow:**

image

**Angular:**

image

**Rust:**

image

carver:
No strong opinion (besides a preference for the thing that you say is too hard to pull off: separating classmethods, etc). :+1:

cburgdorf:
>besides a preference for the thing that you say is too hard to pull off: separating classmethods, etc

Yep, and I probably come back to this some day in the future. But this should probably tackled on the sphinx side where the html is generated. Without properly sorted html there's little we can do on the frontend side (unless we want to resort the html entries with javascript which we really shouldn't be doing).

I'm going to merge this as it is fixing a bug that we have exposed to the live docs (regular methods incorrectly grouped under "classmethods").

PR #795, opened 2019-07-11
Introducing `Node` for py-libp2p

mhchia:
### What was wrong?
- The first step of migrating from the existing p2p code to py-libp2p.
- The local package name `libp2p` conflicts with the dependency [libp2p/py-libp2p](https://github.com/libp2p/py-libp2p/).

### How was it fixed?
- Add `Node` to contain the components in py-libp2p we need.
- Add tests.
- Remove libp2p_daemon_bindings and its tests.

### To-Do
- [x] Add needed API and components in `Node`.
- [x] Test, test, test.

#### Cute Animal Picture


tests/libp2p/bcc/conftest.py
```diff
@@ -1,19 +1,46 @@
...
+def num_nodes():
+    return 3
 
 
 @pytest.fixture
-def peer_id(privkey):
-    return peer_id_from_pubkey(privkey.public_key)
+async def nodes(event_loop, num_nodes, unused_tcp_port_factory):
+    _nodes = tuple(
+        Node(
```
pipermerriam: I'd encourage us to continue using the factory approach found here:

https://github.com/ethereum/trinity/blob/master/p2p/tools/factories.py

When we need to generate test objects.  You'll find two patterns there.

1. use of `factory_boy` when there is an object that needs to be created which can fit into the existing factory API
2. a simple function named `ThingFactory` that generates something.

These have proven useful re-usable tools for generating test objects.

tests/libp2p/bcc/conftest.py
```diff
@@ -1,19 +1,46 @@
...
+            muxer_protocol_ids=[MULTIPLEXING_PROTOCOL_ID],
+            gossipsub_params=None,
+        )
+        for _ in range(num_nodes)
+    )
+    for n in _nodes:
+        asyncio.ensure_future(n.run())
+        await n.events.started.wait()
+    yield _nodes
+    for n in _nodes:
```
pipermerriam: Typically this is done using a `try/finally` to ensure that the last section is always run.

tests/libp2p/bcc/conftest.py
mhchia: Thanks for the pointer! I added `NodeFactory` and used it in the fixture `nodes`. https://github.com/ethereum/trinity/commit/b1b9bd399a95c7c940e356ec7a13000be268a26e#diff-345af0f74433005cdcd3ae9dee64c3acR16.
In this pattern, I still have a `nodes` fixture, calling `NodeFactory` to make the testing nodes. Could you have a look if it is the desired pattern? Or should I get the testing nodes with the factory in every test and clean up them in the end?
```python
def test_node_xxx():
    # begin: make the nodes
    node_0 = NodeFactory()
    node_1 = NodeFactory()
    ...
    # end: clean up
    await node_0.close()
    await node_1.close()
```

tests/libp2p/bcc/conftest.py
mhchia: Sorry, could you elaborate more? I'm not sure if I understand your meaning.

tests/libp2p/bcc/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+
+@pytest.fixture
+def num_nodes():
+    return 3
+
+
+@pytest.fixture
+async def nodes(num_nodes):
+    _nodes = tuple(
+        NodeFactory()
```
pipermerriam: Minor improvement but you can do `nodes = NodeFactory.create_batch(3)` to create multiple in one call.

tests/libp2p/bcc/conftest.py
pipermerriam: This should be:

```python
try:
    yield _nodes
finally:
    for n in _nodes:
        ...
```

Without the `try/finally` if `aclose()` gets called on the generator function, the cleanup code will not be called.

mhchia: Nice catch! Really appreciate your explanation :pray:

mhchia:
Will fix the CI tomorrow and this PR should be ready by the time. 🙏

mhchia:
Should be ready now.

tests/libp2p/bcc/test_node.py
```diff
@@ -0,0 +1,53 @@
...
+async def test_node_dial_peer(nodes):
+    # Test: 0 <-> 1
+    await nodes[0].dial_peer(
+        nodes[1].listen_ip,
+        nodes[1].listen_port,
+        nodes[1].peer_id,
+    )
+    assert nodes[0].peer_id in nodes[1].host.get_network().connections
+    assert nodes[1].peer_id in nodes[0].host.get_network().connections
+    # Test: Reuse the old connection when connecting again
```
NIC619: nit: The wording here sounds like it is reconnecting after _disconnecting_. What about
`# Test: Second dial to a connected peer does not open a new connection`

trinity/protocol/bcc_libp2p/configs.py
```diff
@@ -0,0 +1,48 @@
...
+#   Insecure: when /plaintext/2.0.0 is finalized and implemented in py-libp2p
+#   or
+#   Secure
+SECURITY_PROTOCOL_ID = "/insecure/1.0.0"
+
+MULTISELECT_PROTOCOL_ID = "/multistream/1.0.0"
+
+MULTIPLEXING_PROTOCOL_ID = "/mplex/6.7.0"
+
+PUBSUB_PROTOCOL_ID = "/eth/serenity/gossipsub/1.0.0"
```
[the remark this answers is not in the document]
mhchia: I actually mean "the protocol id of pubsub we use here". Given this context, does `PUBSUB_PROTOCOL_ID` make sense?

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -0,0 +1,153 @@
...
+        self.listen_port = listen_port
+        self.privkey = privkey
+        # TODO: Add key and peer_id to the peerstore
+        network: INetwork = initialize_default_swarm(
+            id_opt=peer_id_from_pubkey(self.privkey.public_key),
+            transport_opt=[self.listen_maddr],
+            muxer_opt=muxer_protocol_ids,
+            sec_opt=security_protocol_ops,
+            peerstore_opt=None,  # let the function initialize it
+            disc_opt=None,  # no routing required here
```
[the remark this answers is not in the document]
mhchia: I think it will be used for a while because we will use discv5 instead of the Kademlia in libp2p for the discovery.

PR #796, opened 2019-07-11
Specify IPC path for attach command

s0b0lev:
### What was wrong?
Add CLI positional argument for attach console plugin to specifying the IPC path.

Related to #23

### How was it fixed?
Added new CLI positional parameter to attach the console plugin.

### To-Do
- [x] Update documentation
- [x] add a newsfragment **(not sure it is required?)**

#### Cute Animal Picture
image

### DIFF
--- a/docs/api/api.cli.rst
+++ b/docs/api/api.cli.rst
@@ -138,11 +138,12 @@ Attach a REPL to a running Trinity instance
 We can attach a REPL to a running Trinity instance to perform RPC request or
 interact with a web3 instance.
 
- .. code-block:: shell
-     usage: trinity attach [-h] [ipc_path]
-     positional arguments:
+.. code-block:: shell
+
+    usage: trinity attach [-h] [ipc_path]
+    positional arguments:
         ipc_path    Specify an IPC path
-     optional arguments:
+    optional arguments:
         -h, --help  show this help message and exit
 
 Check out the :doc:`Quickstart </guides/quickstart>` for a full example.
--- a/docs/api/api.cli.rst
+++ b/docs/api/api.cli.rst
@@ -132,6 +132,23 @@ We can also generate an always up-to-date version of them by running ``trinity -
     --sync-mode {fast,full,beam,light,none}
 
 
+Builtin plugins CLI
+~~~~~~~~~~~~~~~~~~~
+
+**Attach plugin**
+
+Open an REPL attached to a currently running chain.
+
+.. code-block:: shell
+
+    usage: trinity attach [-h] [ipc_path]
+
+    positional arguments:
+        ipc_path    Specify an IPC path
+
+    optional arguments:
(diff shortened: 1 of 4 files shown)


s0b0lev:
@pipermerriam I just noticed that Trinity has `--data-dir` param, which kinda does the same thing:

```
$ trinity --data-dir <path-to-running-nodes-data-dir> attach
```
vs 
```
$ trinity attach <path-to-ipc-path> 
```

Have I misunderstood the idea?

pipermerriam:
No, I think these are sort of two ways to do the same thing.  The `--data-dir` approach allows you to use the same command you used to run trinity to also attach the console.  The other lets you attach directly to the IPC file without having to sort out the actual data-dir.  I think the direct approach that this PR implements is more useful for a situation like logging into a long running machine and connecting directly since the exact path for the data directory might be buried in a configuration file somewhere but you know where the IPC file path is on disk.  I'm :+1: on this approach.

@cburgdorf any thoughts on whether this is a good API?

cburgdorf:
Beside `trinity --data-dir <path-to-running-nodes-data-dir> attach` there's also `trinity --trinity-root-dir <path-to-running-nodes-root-dir> attach` that should also work.

>but you know where the IPC file path is on disk

To play devils advocate, if you know the path of the IPC file, just go one level up and you have the path of the `data_dir`.

So, part of me thinks that we already have two ways of achieving this and we don't need a third but on the other hand it does feel like a nice shorthand and the code footprint isn't big so I'm :+1: to take it.

Unrelated to this specific PR but because you are already in that part of the code base, you might enjoy this other thing to tackle there: https://github.com/ethereum/trinity/issues/797

I'm sure the research team would be thankful to have that!

docs/api/api.cli.rst
```diff
@@ -132,6 +132,23 @@ We can also generate an always up-to-date version of them by running ``trinity -
     --sync-mode {fast,full,beam,light,none}
 
 
+Builtin plugins CLI
```
cburgdorf: This doesn't need to mention plugins. The user doesn't need to care that this feature is built as a plugin.

How about:

```
Attach a REPL to a running Trinity instance
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

We can attach a REPL to a running Trinity instance to perform RPC request or interact with
a web3 instance.

 .. code-block:: shell
     usage: trinity attach [-h] [ipc_path]
     positional arguments:
        ipc_path    Specify an IPC path
     optional arguments:
        -h, --help  show this help message and exit

Check out the `Quickstart </guides/quickstart>` for a full example.
```

s0b0lev: Thanks, updated.

s0b0lev:
Hey @pipermerriam, should I do anything else within this PR?

PR #799, opened 2019-07-12
[hotfix] pin `ruamel.yaml==0.15.98`

hwwhww:
### What was wrong?

Somehow `ruamel.yaml` `0.15.99` is incompatible in type hinting.

### How was it fixed?

Pin `ruamel.yaml==0.15.98` for now.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

🐴

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ deps = {
         "jsonschema==3.0.1",
         "mypy_extensions>=0.4.1,<1.0.0",
         "typing_extensions>=3.7.2,<4.0.0",
-        "ruamel.yaml>=0.15.87",
+        "ruamel.yaml==0.15.98",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -56,7 +56,7 @@
         # only needed for p2p
         "pytest-asyncio-network-simulator==0.1.0a2;python_version>='3.6'",
         # only for eth2
-        "ruamel.yaml>=0.15.87,<0.16",
+        "ruamel.yaml==0.15.98,<0.16",
```
carver: Probably better to remove the upper bound since it is meaningless after this change, and could be confusing:
```suggestion
        "ruamel.yaml==0.15.98",
```

PR #800, opened 2019-07-12
Refactor how we get state from `BeaconChain`

hwwhww:
### What was wrong?
Part of the state cache design (#732) and the precondition of #774.

### How was it fixed?

To prepare for having state cache in `BeaconChain`, we don't query `state` in `StateMachine.state`; instead, we get the state from `BeaconChain`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
480px-Golden_Retriever_Buddy_0311
(Wimting CC BY-SA 4.0)

### DIFF
--- a/eth2/beacon/chains/base.py
+++ b/eth2/beacon/chains/base.py
@@ -18,7 +18,7 @@ from eth2.beacon.types.attestations import Attestation
 from eth2.beacon.types.blocks import BaseBeaconBlock
 from eth2.beacon.types.states import BeaconState
 from eth2.beacon.typing import FromBlockParams, Slot
-from eth2.configs import Eth2GenesisConfig
+from eth2.configs import Eth2Config, Eth2GenesisConfig
 
 if TYPE_CHECKING:
     from eth2.beacon.state_machines.base import BaseBeaconStateMachine  # noqa: F401
@@ -183,6 +183,9 @@ class BeaconChain(BaseBeaconChain):
             raise AttributeError("`chaindb_class` not set")
         return cls.chaindb_class
 
+    def get_config_by_slot(self, slot: Slot) -> Eth2Config:
+        return self.get_state_machine_class_for_block_slot(slot).config
+
     #
     # Chain API
     #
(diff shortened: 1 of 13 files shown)


eth2/beacon/chains/base.py
```diff
@@ -317,7 +324,16 @@ def get_state_by_slot(self, slot: Slot) -> BeaconState:
         """
         sm_class = self.get_state_machine_class_for_block_slot(slot)
         state_class = sm_class.get_state_class()
-        return self.chaindb.get_state_by_slot(slot, state_class)
+        state_root = self.chaindb.get_state_root_by_slot(slot)
+        try:
+            state = self.chaindb.get_state_by_root(state_root, state_class)
+            return state
+        except StateSlotNotFound:
```
NIC619: `chaindb.get_state_by_root()` will throw `StateRootNotFound` while `chaindb.get_state_root_by_slot()` will throw `StateSlotNotFound`.
And I think we need to try/catch both of these two errors?

eth2/beacon/chains/base.py
```diff
@@ -422,15 +442,19 @@ def import_block(
...
+            check_proposer_signature=perform_validation,
+        )
 
         # Validate the imported block.
         if perform_validation:
             validate_imported_block_unchanged(imported_block, block)
 
         # TODO: Now it just persists all state. Should design how to clean up the old state.
-        self.chaindb.persist_state(state)
+        self.chaindb.persist_state(state, is_epoch_boundary=True)
```
NIC619: Is hardcoding `is_epoch_boundary=True` a temporary thing here?

hwwhww: Removed it. :)

eth2/beacon/chains/base.py
hwwhww: Good catch!
I just combine both of them into `StateNotFound` exception and open an issue: https://github.com/ethereum/trinity/issues/843
IMO it should be handled in `trinity`.

eth2/beacon/db/exceptions.py
```diff
@@ -12,9 +12,9 @@ class HeadStateSlotNotFound(BeaconDBException):
     pass
 
 
-class StateSlotNotFound(BeaconDBException):
+class StateNotFound(BeaconDBException):
     """
-    Exception raised if state root with the given slot number does not exist.
+    Exception raised if state with the given state does not exist.
```
NIC619: Maybe
```suggestion
    Exception raised if state with the given slot number or state root does not exist.
```

hwwhww:
### Profiling

`test_demo.py`: 44.51 seconds -> 29.12 seconds 🎉

#### Before (8570b61)
8570b61

#### After (80faf88b)
80faf88b

eth2/beacon/state_machines/base.py
```diff
@@ -65,9 +65,12 @@ def get_fork_choice_scoring(self) -> ForkChoiceScoringFn:
...
     @abstractmethod
     def import_block(
-        self, block: BaseBeaconBlock, check_proposer_signature: bool = True
+        self,
+        block: BaseBeaconBlock,
+        state: BeaconState,
+        check_proposer_signature: bool = True,
     ) -> Tuple[BeaconState, BaseBeaconBlock]:
-        ...
+        pass
```
hwwhww:
```suggestion
        ...
```

Issue #797, opened 2019-07-12
Allow trinity db-shell to inspect the beacon node db.


cburgdorf:
### What is wrong?

The `trinity db-shell` command allows one to inspect/manipulate the database directly which is a nice way for developers to diagnose and mess things up.

Unfortunately the command does only support the Eth1 database but not the beacon node database so far.
### How can it be fixed

Basically implement the `else` branch of this (and turn it into an `elif`)

https://github.com/ethereum/trinity/blob/c35d14ac9f276a9b088363108d3f5c50800a0854/trinity/plugins/builtin/attach/plugin.py#L76-L84

cburgdorf:
This was implemented in #801 but the issue was never closed.

PR #801, opened 2019-07-15
Run db-shell for beacon nodes

s0b0lev:
### What was wrong?
DB shell plugin unavailable for beacon chain
Related to https://github.com/ethereum/trinity/issues/797

### How was it fixed?
Added support for beacon chain.
Added db-shell plugin to beacon registry

:warning:  **Note** I wasn't able to properly run beacon chain (try to run it with run_beacon_nodes.py and just `$ trinity-beacon`). But got this error when attempt to use db-shell plugin: https://gist.github.com/s0b0lev/359ff8601e2445fbd71e0b0f6b63ee7b

### To-Do

- [ ] Test that it works :)


cburgdorf:
The problem is that this code is currently bound to only work with Eth1:

>https://github.com/ethereum/trinity/blob/0ddb45b76a7abc1d93e4a67e58c929e2a64dbbd7/trinity/plugins/builtin/attach/console.py#L103-L115

The beacon node uses entirely different db, chain and chain config primitives.

https://github.com/ethereum/trinity/blob/0ddb45b76a7abc1d93e4a67e58c929e2a64dbbd7/trinity/db/beacon/manager.py#L28-L30

Maybe we can even come up with tests to ensure this works. They *could* be written a lot like these https://github.com/ethereum/trinity/blob/0ddb45b76a7abc1d93e4a67e58c929e2a64dbbd7/tests/integration/test_trinity_cli.py#L128 but instead of spinning up one instance, waiting for it to have booted and then attaching to this live instance we could run `trinity db-shell` and `trinity-beacon db-shell` without a live instance and just test if we can retrieve the genesis block or something like that. That would make these tests less time consuming yet still provide a minimum level of confidence that things don't break.

s0b0lev:
I will wait for merging #791 , and update https://github.com/ethereum/trinity/blob/master/trinity/plugins/builtin/attach/console.py#L103-L121 to run bot eth2 and eth1

s0b0lev:
Running db-shell for Beacon:

1. Run beacon nodes:

```shell
$ python eth2/beacon/scripts/run_beacon_nodes.py
```
It will spawn 2 trinity-beacon nodes which can be accessable:

2. Run db-shell:
```
$ trinity-beacon --data-dir /tmp/aaaa/alice/mainnet/ --trinity-root-dir=/tmp/aaaa/alice  db-shell
```

s0b0lev:
@cburgdorf PR is ready for review

s0b0lev:
I was unable to write tests for `trinity-beacon db-shell` because it requires to have `genesis_state.yml`.
Which mean that it requires additional steps in tests to create testnet and etc (like eth2/beacon/scripts/run_beacon_nodes.py)

trinity/plugins/builtin/attach/console.py
```diff
@@ -98,44 +99,87 @@ def rpc(method: str, params: Dict[str, Any] = None) -> str:
...
+        'trinity_config': config.get("trinity_config"),
+        'chain_config': config.get("chain_config"),
+        'chain': config.get("chain"),
+    }
+    shell(use_ipython, namespace, DB_SHELL_BANNER + greeter)
+
 
-    db_ipc_path = trinity_config.database_ipc_path
-    trinity_already_running = db_ipc_path.exists()
+def eth1_db_shell(database_dir: Path, trinity_config: TrinityConfig) -> Dict[str, Any]:
```
cburgdorf: What do you think about calling this `get_eth1_shell_context` to more precisely capture what it does?

trinity/plugins/builtin/attach/console.py
```diff
@@ -98,44 +99,87 @@ def rpc(method: str, params: Dict[str, Any] = None) -> str:
...
-    Inspecting active Trinity? {trinity_already_running}
 
-    Available Context Variables:
-      - `db`: base database object
-      - `chaindb`: `ChainDB` instance
-      - `trinity_config`: `TrinityConfig` instance
-      - `chain_config`: `ChainConfig` instance
-      - `chain`: `Chain` instance
-    """
+def beacon_db_shell(database_dir: Path, trinity_config: TrinityConfig) -> Dict[str, Any]:
```
cburgdorf: And then this could be  `get_beacon_shell_context`

cburgdorf:
This looks good! Would just suggest two renames.

@ChihChengLiang @NIC619 

This lets you run the `db-shell` command against a beacon node db (cold or while the beacon node is running) and exposes the db primitives to fiddle around with the db.

E.g. 
>`trinity-beacon --data-dir /tmp/aaaa/alice/mainnet/ --trinity-root-dir=/tmp/aaaa/alice  db-shell`

This has been a handy tool for the Eth1 development so I suppose you'll find it handy for beacon node development as well!

trinity/plugins/builtin/attach/console.py
s0b0lev: Yeh, sounds better!

trinity/plugins/builtin/attach/console.py
s0b0lev: Updated, thanks!

NIC619:
Thanks @cburgdorf @s0b0lev ! This would be really helpful!

trinity/db/beacon/manager.py
```diff
@@ -22,6 +23,8 @@
     BeaconBlock,
 )
 
+AUTH_KEY = b"not secure, but only connect over IPC"
```
ChihChengLiang: What do you think renaming it to `AUTH_KEY="DB_AUTHKEY"`?

cburgdorf:
@s0b0lev Thanks for tackling this! I think this is really useful for the beacon node development! Can you make sure to also create a newsfragment so that it gets highlighted as a new feature in the upcoming release notes.

trinity/plugins/builtin/attach/console.py
```diff
@@ -98,44 +99,87 @@ def rpc(method: str, params: Dict[str, Any] = None) -> str:
...
+
+    trinity_already_running = ipc_path.exists()
+    if trinity_already_running:
+        db_manager = beacon.manager.create_db_consumer_manager(ipc_path)  # type: ignore
+        db = db_manager.get_db()
+    else:
+        db = LevelDB(database_dir)
+
+    chain_config = app_config.get_chain_config()
+    chain = chain_config.beacon_chain_class
```
ChihChengLiang: The greeter says `- chain: Chain instance` but the line is only a class of Chain. Need to pass these lines too.

```python
# copy from trinity/plugins/eth2/beacon/plugin.py
from eth2.beacon.operations.attestation_pool import AttestationPool
attestation_pool = AttestationPool()
chain = chain_config.beacon_chain_class(
    db,
    attestation_pool,
    chain_config.genesis_config
)


```

trinity/plugins/builtin/attach/console.py
```diff
@@ -98,44 +99,87 @@ def rpc(method: str, params: Dict[str, Any] = None) -> str:
...
+
+    chaindb = BeaconChainDB(db, chain_config.genesis_config)
+    head = chaindb.get_canonical_head(BeaconBlock)
+    return {
         'db': db,
         'chaindb': chaindb,
         'trinity_config': trinity_config,
         'chain_config': chain_config,
         'chain': chain,
+        'head_block_number': head.slot,
```
ChihChengLiang: This is just `head_slot`.

trinity/plugins/builtin/attach/console.py
```diff
@@ -98,44 +99,87 @@ def rpc(method: str, params: Dict[str, Any] = None) -> str:
...
+        'hex_hash': head.hash_tree_root.hex(),
```
ChihChengLiang: What do you think about replacing `head_slot` and `hex_hash` with `'head': head`?

trinity/plugins/builtin/attach/plugin.py
```diff
@@ -77,16 +81,21 @@ def configure_parser(cls,
             'db-shell',
             help='open a REPL to inspect the db',
         )
-
         attach_parser.set_defaults(func=cls.run_shell)
 
     @classmethod
     def run_shell(cls, args: Namespace, trinity_config: TrinityConfig) -> None:
+        config: Union[Eth1AppConfig, BeaconAppConfig]
```
ChihChengLiang: Is it possible to do `config: BaseAppConfig`?

trinity/db/beacon/manager.py
cburgdorf: I think that's a straight copy from here and that Jason picked this text on purpose to ensure everyone sees a red flag.

>https://github.com/ethereum/trinity/blob/68f8636d5651361aee10084f31ebcb5ea454e293/trinity/db/eth1/manager.py#L26

But maybe instead of duplicating it, we could put this into `trinity.constants`?

trinity/plugins/builtin/attach/console.py
s0b0lev: Yes, it would be better, but there are difference between eth1 and beacon attributes. 

```
# ETH1:
'head_block_number': head.block_number,
'hex_hash': head.hex_hash,

# Beacon:
'head_block_number': head.slot,
'hex_hash': head.hash_tree_root.hex(),
```

If we pass only head, we have to make if / else condition in db_shell for greating message.

But we can return `'head':  head` instead of `'state_root_hex': encode_hex(head.state_root),` it is the same in both cases.

trinity/plugins/builtin/attach/plugin.py
s0b0lev: Updated

trinity/plugins/builtin/attach/console.py
s0b0lev: Related to this comment: https://github.com/ethereum/trinity/pull/801#discussion_r306205258

trinity/plugins/builtin/attach/console.py
s0b0lev: thanks, updated!

trinity/db/beacon/manager.py
s0b0lev: Added new constant `trinity.constants.AUTH_KEY`

trinity/plugins/builtin/attach/console.py
ChihChengLiang: Ah, I see. Adding if / else condition makes the greeter more clear, but it's not too bad if we don't do that. Both approaches work.

cburgdorf:
Landed! :tada: Thanks again for your contribution @s0b0lev

PR #807, opened 2019-07-17
Remove blspy from eth2 extras

cburgdorf:
### What was wrong?

We recently implicitly added a system dependency against `cmake` that effects anyone trying to install trinity.

### How was it fixed?

Just trying to figure out the things that would break and if we can opt-in to install `bls-bindings` for these plalces.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -106,7 +106,6 @@ deps = {
         "rlp>=1.1.0,<2.0.0",
         PYEVM_DEPENDENCY,
         "ssz==0.1.0a10",
-        "blspy>=0.1.8,<1",  # for `bls_chia`
     ],
     'libp2p': [
         "base58>=1.0.3",


ChihChengLiang:
Thinking about letting Trinity try import blspy (Chia BLS) as the default backend, and fallback to py_ecc if blspy is not available.

But I'm not sure what would the CI environments look like. Do we have an environment with blspy and one without?

cburgdorf:
>Thinking about letting Trinity try import blspy (Chia BLS) as the default backend, and fallback to py_ecc if blspy is not available.

That sounds reasonable. Alternatively we could also try import it and raise an error that `blspy` needs to be installed. Then tests that depend on it would also have to depend on the `bls-bindings` extra.

>Do we have an environment with blspy and one without?

We could do that but if we just don't provide a fallback and just make sure tests that depend on it install `bls-bindings` that might be less effort to get around it?

cburgdorf:
Ah ok. I think it might not work without an actual fallback. At least, my two minute research suggest there is no easy way to prevent these imports to get triggered. So maybe the fallback would be the best option.

PR #808, opened 2019-07-17
add bls fallback option

ChihChengLiang:
### What was wrong?

blspy is hard to install

### How was it fixed?

opt out blspy if you want.

#### Cute Animal Picture

🦋

### DIFF
--- a/eth2/_utils/bls/backends/__init__.py
+++ b/eth2/_utils/bls/backends/__init__.py
@@ -1,11 +1,19 @@
-from .chia import ChiaBackend
+
 from .noop import NoOpBackend
 from .py_ecc import PyECCBackend
 
 
-DEFAULT_BACKEND = ChiaBackend
 AVAILABLE_BACKENDS = (
-    ChiaBackend,
     NoOpBackend,
     PyECCBackend,
 )
+
+
+# If blspy not installed, use PyECC as default BLS backend
+
+try:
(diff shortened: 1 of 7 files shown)


cburgdorf:
:+1: on that approach

eth2/_utils/bls/backends/__init__.py
```diff
@@ -1,11 +1,19 @@
...
     PyECCBackend,
 )
+
+
+# If blspy not installed, use PyECC as default BLS backend
+
+try:
+    from .chia import ChiaBackend
+    DEFAULT_BACKEND = ChiaBackend
+    AVAILABLE_BACKENDS += ChiaBackend
```
mhchia: Does this line work? It looks like it should fail, but I didn't see errors in the CI(except for the linter).

cburgdorf: It does work. It returns a new tuple which is then assigned to the old variable. So it is essentially the same as `AVAILABLE_BACKENDS = AVAILABLE_BACKENDS + ChiaBackend`

ChihChengLiang:
@cburgdorf @pipermerriam  The PR is ready to merge if no further objections.

pipermerriam:
No objections, just voicing what I think is the proper *long term* plan.  Short term this is a fine solution.

PR #809, opened 2019-07-17
Properly cancel all of the BaseService tasks by awaiting them after cancellation

pipermerriam:
### What was wrong?

It looks like we've been failing to properly cancel sub-tasks in `BaseService`.  A look at the [documentation for `Task.cancel()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel) shows that after calling `cancel()` we also need to await the task (and then catch the cancellation that arises)

### How was it fixed?

Added an additional `await task` inside of a `try/except` which silently discards the `asyncio.CancellationError`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -313,16 +313,17 @@ class BaseService(ABC, CancellableMixin):
                 self._log_tasks("Pending tasks")
             if self._child_services:
                 self.logger.debug("Pending child services: %s", list(self._child_services))
-            self._forcibly_cancel_all_tasks()
-            # Sleep a bit because the Future.cancel() method just schedules the callbacks, so we
-            # need to give the event loop a chance to actually call them.
-            await asyncio.sleep(0.5)
+            await self._forcibly_cancel_all_tasks()
         else:
             self.logger.debug("%s finished cleanly", self)
 
-    def _forcibly_cancel_all_tasks(self) -> None:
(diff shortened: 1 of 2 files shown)


pipermerriam:
This didn't actually have any effect on the warnings output during test runs (which are currently all from the lahja library). However, in a branch I'm working on this fix cleaned up a lot of dirty warning output coming from the test run.

p2p/service.py
```diff
@@ -313,16 +313,17 @@ def cancel_nowait(self) -> None:
...
+            await self._forcibly_cancel_all_tasks()
         else:
             self.logger.debug("%s finished cleanly", self)
 
-    def _forcibly_cancel_all_tasks(self) -> None:
+    async def _forcibly_cancel_all_tasks(self) -> None:
         for task in self._tasks:
             task.cancel()
+            try:
+                await task
```
cburgdorf: Shouldn't we do that with `await asyncio.gather(...)` ? Because now, it looks like we will tear down subtasks sequentially which may be slower compared to running that concurrently.

pipermerriam: Maybe but I'm inclined to keep it this way for simplicity.

cburgdorf: Fine by me

PR #811, opened 2019-07-18
Fix memory leak in handshake lock routine

cburgdorf:
### What was wrong?

The recently introduced fix that ensures we do not run multiple concurrent
handshakes to the same peer accidentially introduced a (rarely exposed) memory
leak. 

### How was it fixed?

This fix introduces a ``ResourceLock`` and refactores the code to use it
to also fix the previously introduced memory leak.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -10,7 +10,6 @@ from typing import (
     Dict,
     Iterator,
     List,
-    Set,
     Tuple,
     Type,
 )
@@ -69,6 +68,9 @@ from p2p.peer_backend import (
 )
 from p2p.disconnect import (
     DisconnectReason,
+)
+from p2p.resource_lock import (
+    ResourceLock,
 )
 from p2p.service import (
     BaseService,
@@ -129,8 +131,7 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
         self._connection_attempt_lock = asyncio.BoundedSemaphore(MAX_CONCURRENT_CONNECTION_ATTEMPTS)
 
         # Ensure we can only have a single concurrent handshake in flight per remote
-        self._handshake_lock = asyncio.Lock()
-        self._inflight_handshakes: Set[NodeAPI] = set()
+        self._handshake_locks = ResourceLock()
 
         self.peer_backends = self.setup_peer_backends()
         self.connection_tracker = self.setup_connection_tracker()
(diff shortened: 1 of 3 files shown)


p2p/peer_pool.py
```diff
@@ -316,68 +317,61 @@ def _add_peer(self,
...
-            async with self._handshake_lock:
-                self.logger.debug("Removing %s from connect_lock", remote)
-                self._inflight_handshakes.remove(remote)
+        if self._handshake_locks.is_locked(remote):
+            self.logger.debug2("Skipping %s; already shaking hands", remote)
+            raise IneligiblePeer(f"Already shaking hands with {remote}")
+
+        async with self._handshake_locks.lock(remote):
+
+            if remote in self.connected_nodes:
```
cburgdorf: I moved this inside the protected code also I'll admit I'm not entirely sure if it is needed. It just *feels* more secure but if anyone likes to fix my knowledge gap, here's my thought process:

What I'm thinking is that if this check would live *outside* or the protected code, then it might happen that two peers get `False` for `if self._handshake_locks.is_locked(remote)` at the same time and then one of them gets the handshake lock **immediately** while the other one would **wait on** the handshake lock.

This wasn't the case with the previous implementation where the lock was for the entire set. But by moving the `if remote in self.connected_nodes` line into the protected area we ensure that the worst thing that could happen is that someone waits on the lock just to eventually find out we are already connected to the peer by the time that the lock is available.

@pipermerriam  am I doing it wrong?

pipermerriam: This seems appropriate to me.  The fail case with locks that I know of is when you have a deadlock situation but I don't see how this could do that as I'm pretty sure it behaves the way you said with one process just hanging until the lock is released and then aborting early.

:+1:

p2p/peer_pool.py
```diff
@@ -80,6 +79,9 @@
     BaseConnectionTracker,
     NoopConnectionTracker,
 )
+from p2p._utils.async_lock import (
```
pipermerriam: This should go somewhere like `p2p.resource_lock`.  Importing from `p2p._utils`  should be avoided.

p2p/peer_pool.py
```diff
@@ -34,7 +33,7 @@
     BroadcastConfig,
 )
 
-from p2p._utils import clamp
+from p2p._utils.common import clamp
```
pipermerriam: You should be able to fully drop the clamp utility and use the `eth_utils.clamp` version.

p2p/peer_pool.py
cburgdorf: Ah, I think I see what you mean. I actually created the `p2p._utils` directory because I wanted to shield it so that it doesn't become a public api of the `p2p` module that others would rely on. 
Because, I think this should eventually not  live in `p2p` at all. It feels more like an `eth_utils` kinda thing. 

But if we start importing this from `trinity` then I agree that importing from `p2p._utils` would be wrong.

Should I just create a PR against `eth_utils` to put it there so that we really only need it temporary in this code base?

pipermerriam: No, I would see the resource lock going into it's own stand alone.  I don't think it belongs in `eth-utils`.  Currently we've not added much to `eth-utils` that aren't pure function type things and I'm hesitant to start adding data structures like this.

pipermerriam: But for now, I'm fine putting it in `p2p.resource_lock`

PR #814, opened 2019-07-19
enable beacon plugin test

ChihChengLiang:
### What was wrong?

The test has been disabled for a long time

### How was it fixed?

Create a genesis file then run the beacon plugin

#### Cute Animal Picture


ChihChengLiang:
The plan was to experiment on pylibp2p and could break the beacon chain plugin temporarily. So no need the test now

PR #816, opened 2019-07-21
Implement UDP communication services for discv5

jannikluhn:
Add two simple services that read and write data to and from a UDP socket.

### To-Do

- [ ] Fix async fixtures: For some reason they don't work even though they should ("RuntimeError: Cannot be used outside of a run context")
- [ ] Tell CI to run tests with `trio_mode` enabled
- [ ] Fix type hints if possible
- [ ] Clean up commit history

#### Cute Animal Picture

animal-animal-photography-avian-2629372

### DIFF
--- a/p2p/discv5/channel_services.py
+++ b/p2p/discv5/channel_services.py
@@ -1,0 +1,66 @@
+import logging
+from typing import (
+    NamedTuple,
+)
+
+from trio.socket import (
+    SocketType,
+)
+from trio.abc import (
+    ReceiveChannel,
+    SendChannel,
+)
+
+from p2p.trio_service import (
+    as_service,
+    Manager,
+)
+
+from p2p.discv5.constants import (
+    DATAGRAM_BUFFER_SIZE,
+)
+
+
+class Endpoint(NamedTuple):
+    ip_address: bytes
+    port: int
+
+
+class IncomingDatagram(NamedTuple):
+    datagram: bytes
+    sender: Endpoint
+
+
+class OutgoingDatagram(NamedTuple):
+    datagram: bytes
+    receiver: Endpoint
+
+
+@as_service
+async def DatagramReceiver(manager: Manager,
+                           socket: SocketType,
+                           incoming_datagram_send_channel: SendChannel[IncomingDatagram],
+                           ) -> None:
+    """Read datagrams from a socket and send them to a channel."""
+    logger = logging.getLogger('p2p.discv5.channel_services.DatagramReceiver')
+
+    async with incoming_datagram_send_channel:
+        while manager.is_running:
+            datagram, (ip_address, port) = await socket.recvfrom(DATAGRAM_BUFFER_SIZE)
+            logger.debug(f"Received {len(datagram)} bytes from {(ip_address, port)}")
+            incoming_datagram = IncomingDatagram(datagram, Endpoint(ip_address, port))
+            await incoming_datagram_send_channel.send(incoming_datagram)
+
+
+@as_service
+async def DatagramSender(manager: Manager,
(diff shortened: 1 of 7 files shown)


tests/p2p/discv5/test_channel_services.py
```diff
@@ -0,0 +1,62 @@
...
+    send_channel, receive_channel = trio.open_memory_channel(1)
+    async with background_service(DatagramReceiver(receiving_socket, send_channel)):
+        data = b"some packet"
+
+        await sending_socket.sendto(data, receiver_address)
+        with trio.fail_after(0.5):
+            received_datagram = await receive_channel.receive()
+
+        assert received_datagram.datagram == data
+        assert received_datagram.sender == sender_address
```
pipermerriam: I think you should probably assert that you have property access on these by doing `assert received_datagram.sender.address == ...` because I think that this will evaluate truthy if `sender` is just a plain `tuple` (but I'm not 100% sure)

p2p/discv5/channel_services.py
```diff
@@ -0,0 +1,65 @@
...
+@as_service
+async def DatagramReceiver(manager: Manager,
+                           socket: SocketType,
+                           incoming_datagram_send_channel: SendChannel[IncomingDatagram],
+                           ) -> None:
+    """Read datagrams from a socket and send them to a channel."""
+    logger = logging.getLogger('p2p.discv5.channel_services.DatagramReceiver')
+
+    async with incoming_datagram_send_channel:
+        while True:
```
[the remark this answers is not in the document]
jannikluhn: Will change because it looks cleaner than a `while True`. But I don't think it actually changes anything as cancellation would just raise an exception during one of the awaits.

jannikluhn:
> Now, the current model does have this property in that it appears you could just wire up some trio channels and pretend they had networking behind them.

Yes, the intention is that there are separate services for each processing step (socket, decoding, decryption, handshake, ...) that just receive data from some channel and send it to another. This should make it possible to test all components fully independently.

pipermerriam:
> This should make it possible to test all components fully independently.

Cool, sounds like you've got a vision for this so I'll wait to see how it comes together.

p2p/trio_service.py
```diff
@@ -170,7 +170,7 @@ def cancel(self) -> None:
         ...
 
 
-LogicFnType = Callable[[ManagerAPI, VarArg(), KwArg()], Awaitable[Any]]
+LogicFnType = Callable[..., Awaitable[Any]]
```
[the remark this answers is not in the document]
pipermerriam: I kind of suspected this was going to be the case.

p2p/discv5/channel_services.py
```diff
@@ -0,0 +1,66 @@
...
+
+    async with incoming_datagram_send_channel:
+        while manager.is_running:
+            datagram, (ip_address, port) = await socket.recvfrom(DATAGRAM_BUFFER_SIZE)
+            logger.debug(f"Received {len(datagram)} bytes from {(ip_address, port)}")
+            incoming_datagram = IncomingDatagram(datagram, Endpoint(ip_address, port))
+            await incoming_datagram_send_channel.send(incoming_datagram)
+
+
+@as_service
```
pipermerriam: not for this PR but thinking about API.  Maybe we could have this decorator optionally accept some arguments.

```python
@as_service(manager=True)
async def UsesManager(manager: ManagerAPI):
    ...

@as_service
async def DoesNotUseManager():
    ...
```

Might be nice to not always have to include the `manager` argument for services that don't use it.

jannikluhn: :+1: sounds like a good idea. Most services will also do some logging, so the same could be done for a logger instance for convenience (might be confusing though, not sure).

PR #818, opened 2019-07-22
Add abstract base classes for many of the main p2p APIs

pipermerriam:
### What was wrong?

The `p2p` had some warts leftover from original  implementation and organization and it has been proving useful to have true abstract base class API definitions for the main APIs that are exposed in the `p2p` module.

### How was it fixed?

- Added the `p2p.abc` module which contains pure abstract definitions of the APIs for `Command`, `Protocol`, `Transport`, `Node` and `Address`
- Added `p2p.disconnect` to house the `DisconnectReason` enum (this fixes a circular import issue)
- Added `p2p.typing` to house some values that are used for type hints.
- Moved some constants from `p2p.kademlia` to `p2p.constants`
- Removed most places where a module from `p2p` was imported in favor of directly importing the things it uses from that module.
- Condensed many multi-line imports that only import one thing to be a single line.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/abc.py
+++ b/p2p/abc.py
@@ -1,0 +1,245 @@
+from abc import ABC, abstractmethod
+from typing import (
+    Any,
+    ClassVar,
+    Dict,
+    Generic,
+    List,
+    Tuple,
+    Type,
+    Union,
+    TypeVar,
+)
+
+from rlp import sedes
+
+from cancel_token import CancelToken
+
+from eth_keys import datatypes
+
+from p2p.typing import CapabilityType, PayloadType, StructureType
+
+
+TAddress = TypeVar('TAddress', bound='AddressAPI')
+
+
+class AddressAPI(ABC):
+    udp_port: int
+    tcp_port: int
+
+    @abstractmethod
+    def __init__(self, ip: str, udp_port: int, tcp_port: int = 0) -> None:
+        ...
+
+    @property
+    @abstractmethod
+    def is_loopback(self) -> bool:
+        ...
+
+    @property
+    @abstractmethod
+    def is_unspecified(self) -> bool:
+        ...
+
+    @property
+    @abstractmethod
+    def is_reserved(self) -> bool:
+        ...
+
+    @property
+    @abstractmethod
+    def is_private(self) -> bool:
+        ...
+
+    @property
+    @abstractmethod
+    def ip(self) -> str:
+        ...
+
+    @abstractmethod
+    def __eq__(self, other: Any) -> bool:
+        ...
+
+    def __repr__(self) -> str:
+        return 'Address(%s:udp:%s|tcp:%s)' % (self.ip, self.udp_port, self.tcp_port)
+
+    @abstractmethod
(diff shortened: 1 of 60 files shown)


p2p/typing.py
```diff
@@ -0,0 +1,26 @@
...
+]
+
+
+StructureType = Union[
+    Tuple[Tuple[str, Any], ...],
+]
+
+
+CapabilityType = Tuple[str, int]
+CapabilitiesType = Tuple[CapabilityType, ...]
```
cburgdorf: I know these were just moved but allow me to rant one more time that I really dislike suffixing types with `Type`. Similar to how we call a class `Transport` or `Node` and not `TransportType` or `NodeType`, imho all of these were just better named as `Payload`, `Structure`, `Capability`  and `Capabilities`. These are all just aliases for some types (e.g. `Tuple[str, int]]`)  so the `Type` suffix feels redundant plus it always confuses me because when I read `Type` I'm thinking of something like `Kind`, like an enum with different states.

`</rant>`

p2p/abc.py
```diff
@@ -0,0 +1,245 @@
...
+    TypeVar,
+)
+
+from rlp import sedes
+
+from cancel_token import CancelToken
+
+from eth_keys import datatypes
+
+from p2p.typing import CapabilityType, PayloadType, StructureType
```
cburgdorf: Maybe it's time for us to update our style guide. It looks like we've come to accept this style of import more and more while our style guide still forbids that.

>Multiple imports must be split across multiple lines, one per line.

https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#formatting

p2p/abc.py
```diff
@@ -0,0 +1,245 @@
...
+                      ip: str,
+                      udp_port: bytes,
+                      tcp_port: bytes = b'\x00\x00') -> TAddress:
+        ...
+
+
+TNode = TypeVar('TNode', bound='NodeAPI')
+
+
+class NodeAPI(ABC):
```
cburgdorf: I usually decide if we need an abstract type for something or not if we assume it has multiple implementations or not. In essence that boils down to the fact that different implementations can implement abstract methods differently to provider other functionality. 

To me `Address` and `Node` fall more into "data object" kinda bucket and have less room for different implementations so I wouldn't naturally classify these as candidates that need to be implemented against an interface.

I'm not opposed, just curious if you think there will be multiple implementations of `AddressAPI` and `NodeAPI` or if you see other benefits of creating these abstract types?

pipermerriam: I generally agree, however, I ended up having to do `if TYPE_CHECKING` style imports to prevent circular imports in the `p2p.abc` module and it's nice to have the `p2p.abc` module depend on as little of the actual code as possible.

I'm inclined to keep them.

p2p/abc.py
pipermerriam: :+1: 

I didn't even really realize I'd made this switch but I attribute it to `black`.

PR #822, opened 2019-07-23
Expose NewBlockEvent on the event bus

cburgdorf:
### What was wrong?

This is a little spin off from #721 that splits peer pool and sync into two separate processes. This exposes a `NewBlockEvent` on the event bus which is an important event for components to utilize (e.g. to build sync strategies on)

### How was it fixed?

- Created `NewBlockEvent`
- Dispatch the `NewBlockEvent` whenever the native `NewBlock` command pops up

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ develop-eggs
 .installed.cfg
 lib
 lib64
+pip-wheel-metadata
 
 # Installer logs
 pip-log.txt
--- a/newsfragments/822.feature.rst
+++ b/newsfragments/822.feature.rst
@@ -1,0 +1,1 @@
+Expose ``NewBlockEvent`` on the event bus.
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -16,6 +16,7 @@ develop-eggs
 .installed.cfg
 lib
 lib64
+pip-wheel-metadata
```
cburgdorf: This seems to be an issue with recent pip versions https://github.com/pypa/pip/issues/6213

cburgdorf:
Because all of these derive from `PeerPoolMessageEvent` and leave it up to the consumer to unpack things.

https://github.com/ethereum/trinity/blob/06e2c17ae50caf7b81c5386503beb3e94fe67284/trinity/protocol/common/events.py#L89-L98

In the long run, I guess we'll get rid of this mapping anyway and have the peer pool be built natively on Lahja. In the meantime, this is just to keep things simple and don't map properties on lots of places.

PR #825, opened 2019-07-23
Handshake interface

jannikluhn:
### What was wrong?

Identity scheme class only worked with ENR, but it's also responsible for the handshake.

### How was it fixed?

Implement the basic interface and add a placeholder so that we don't have to implement the cryptography as long as it isn't properly specified.

Also some minor stuff:
- rename node address to node id (both are used interchangeably in the spec, but address may collide with endpoint address)
- add method to access public key (we'll need that for the handshake)
- make node id and public key accessible on the ENR via properties

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/identity_schemes.py
+++ b/p2p/discv5/identity_schemes.py
@@ -5,8 +5,9 @@ from abc import (
 from collections import (
     UserDict,
 )
-
+import os
 from typing import (
+    Tuple,
     Type,
     TYPE_CHECKING,
 )
@@ -25,6 +26,14 @@ from eth_utils import (
     encode_hex,
     keccak,
     ValidationError,
+)
+
+from p2p.discv5.typing import (
+    AES128Key,
+    SessionKeys,
+)
+from p2p.discv5.constants import (
+    AES128_KEY_SIZE,
 )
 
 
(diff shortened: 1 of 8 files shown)


p2p/discv5/identity_schemes.py
```diff
@@ -65,51 +77,158 @@ class IdentityScheme(ABC):
...
+    @abstractmethod
+    def validate_enr_signature(cls, enr: "ENR") -> None:
+        """Validate the signature of an ENR."""
+        pass
+
+    @classmethod
+    @abstractmethod
+    def extract_public_key(cls, enr: "BaseENR") -> bytes:
+        """Retrieve the public key from an ENR."""
+        pass
```
pipermerriam: We started moving over to using elipsis for abstract methods.

```python
@abstractmethod
def the_thing(self):
    ...
```

p2p/discv5/identity_schemes.py
```diff
@@ -65,51 +77,158 @@ class IdentityScheme(ABC):
...
 
         if not is_valid:
             raise ValidationError("Invalid signature")
 
     @classmethod
-    def extract_node_address(cls, enr: "ENR") -> bytes:
-        public_key = PublicKey.from_compressed_bytes(enr[cls.public_key_enr_key])
-        return keccak(public_key.to_bytes())
+    def extract_public_key(cls, enr: "BaseENR") -> bytes:
+        return enr[cls.public_key_enr_key]
```
pipermerriam: blind lookups like this always stand out to me as potential places for errors.  How do we know that this key is going to be present.  A quick glance at the `BaseENR` class didn't reveal anything obvious.

p2p/discv5/typing.py
```diff
@@ -1,7 +1,17 @@
...
 )
 
 
 AES128Key = NewType("AES128Key", bytes)
 Nonce = NewType("Nonce", bytes)
+
+RandomBytesFn = Callable[[int], bytes]  # function that returns a number of random bytes
+
+
+class KeyAgreementResult(NamedTuple):
```
pipermerriam: Is the term `KeyAgreement` from the spec?  If not what do you think about something like `SessionKeys`?

tests/p2p/discv5/test_enr.py
```diff
@@ -196,18 +205,25 @@ def test_signature_validation(mock_identity_scheme, identity_scheme_registry):
...
-    assert enr.extract_node_address() == private_key
+    assert enr.node_id == private_key
 
 
 def test_signature_scheme_selection(mock_identity_scheme, identity_scheme_registry):
     mock_enr = ENR(0, {b"id": b"mock"}, b"", identity_scheme_registry)
     assert mock_enr.identity_scheme is mock_identity_scheme
 
-    v4_enr = ENR(0, {b"id": b"v4"}, b"", identity_scheme_registry)
+    v4_enr = ENR(0, {b"id": b"v4", b"secp256k1": b"\x02" * 33}, b"", identity_scheme_registry)
```
pipermerriam: I'd recommend going ahead and making yourself some factory classes for these objects under `p2p.tools.factories`

tests/p2p/discv5/test_identity_schemes.py
```diff
@@ -24,6 +24,16 @@
 )
 
 
+@pytest.fixture
+def get_random_bytes():
+    random = Random(0)
```
pipermerriam: probably better to go ahead and install [`pytest-randomly`](https://github.com/pytest-dev/pytest-randomly) and let it set the random seed.

tests/p2p/discv5/test_identity_schemes.py
```diff
@@ -81,5 +105,27 @@ def test_enr_node_address():
...
+@pytest.mark.parametrize(("public_key", "valid"), (
+    (b"\x02" * 33, True),
+    (b"\x03" * 33, True),
+    (b"", False),
+    (b"\x01" * 33, False),
+    (b"\x02" * 32, False),
+    (b"\x02" * 34, False),
+))
+def test_handshake_public_key_validation(public_key, valid):
+    if valid:
```
pipermerriam: I think we've leaned towards splitting these into two runs, one for the valid ones and one separate test fort he invalid ones.

p2p/discv5/identity_schemes.py
jannikluhn: It's there, but somewhat hidden: `BaseENR.__init__` calls `IdentityScheme.validate_enr_structure` which (if `V4IdentityScheme` is used) checks that the key is present. I was trying to avoid the possibility of an exception here so that we can access the public key and node id via properties that behave nicely.

Do you think a `raise Exception("Unreachable: <comment>")` would be sufficient here to make this a bit more explicit or should I go back to methods instead of properties?

pipermerriam: Probably fine but I'll point out that when things like this are not coupled (the validation and the access) it becomes easy to accidentally break in a refactor such as if the validation step were to move somewhere else.  I don't have a direct suggestion on how to restructure this so fine to leave as-is, or clean up later if you get some inspiration.

jannikluhn:
Sorry for the delay. Changes since last review:

- Ellipsis instead of "pass" in abstract method bodies
- Renamed "KeyAgreementResult" to "SessionKeys" and "perform_key_agreement" to "compute_session_keys"
- In `V4IdentityScheme.extract_public_key(enr)`, I don't assume anymore that the ENR has been validated before (a proper exception is raised). Instead, this assumption is only made in `ENR.public_key`. It's still not ideal, but better than before, as now this issue is only present in one, not two classes. Also, I've added an `Exception("Invariant: ...")` in the property to document this slightly awkward coupling.
- Added id nonce signing and signature verification methods to the identity scheme which are also required for the handshake.
- I got rid of the `get_random_bytes` interface. The only reason why I thought it would be useful is for testing (so that we can replace the non-deterministic `os.urandom` we should use in production with a pseudo random alternative). But now I don't like it anymore as it increases the complexity of the interface for no good reason. Instead, I just mock out `os.urandom` in tests via a fixture.

What I have not done (yet):

- Use factories for creating ENRs. The problem here is that for most of the tests we actually test stuff in the ENRs `__init__` function, so what I'd actually have to do is use some kind of `ENRKwargFactory` which doesn't feel very useful for future tests. So since the existing tests already work fine I'd rather not spend more time on it at the moment to get it right, but revisit this for the next time I need an ENR for tests.
- Cleaned up the commit history. Will do so before I merge

Other than that, if I haven't missed anything, this is ready from my side.

PR #826, opened 2019-07-23
Add P2PProtocol.send_ping

pipermerriam:
extracted from #684 

### What was wrong?

We had no mechanism for sending the `Ping` message and the code for sending the `Hello` message was burried in the handshake code.

### How was it fixed?

Extracted these to be their own methods on the `P2PProtocol` class.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

13-10-12 Summer (1)C650

### DIFF
--- a/newsfragments/828.feature.rst
+++ b/newsfragments/828.feature.rst
@@ -1,0 +1,1 @@
+Add ``p2p.peer.receive_handshake`` to encapsulate the logic for handling incoming connections.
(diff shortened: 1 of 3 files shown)


newsfragments/828.feature.rst
```diff
@@ -0,0 +1 @@
+Add ``p2p.peer.receive_handshake`` to encapsulate the logic for handling incoming connections.
```
carver: Looks like this slipped into the wrong PR, but probably not worth fixing, since they will both get merged.

PR #829, opened 2019-07-23
Remove handshake parameters from p2p protocol class

pipermerriam:
extracted from #684 

### What was wrong?

The `p2p.p2p_proto.P2PProtocol` class diverges from the standard implementation by having some of the handshake parameters get passed in during class instantiation.

### How was it fixed?

Changed these parameters to be provided at the time of sending the handshake.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

13-10-12 Summer (22)C750

### DIFF
--- a/p2p/p2p_proto.py
+++ b/p2p/p2p_proto.py
@@ -12,7 +12,7 @@ from rlp import sedes
 from p2p.abc import TransportAPI
 from p2p.disconnect import DisconnectReason as _DisconnectReason
 from p2p.exceptions import MalformedMessage
-from p2p.typing import PayloadType, CapabilitiesType
+from p2p.typing import CapabilitiesType, PayloadType
 
 from p2p.protocol import (
     Command,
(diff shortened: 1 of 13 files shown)


p2p/p2p_proto.py
```diff
@@ -78,22 +77,19 @@ class P2PProtocol(Protocol):
...
+                 snappy_support: bool) -> None:
         # For the base protocol the cmd_id_offset is always 0.
         # For the base protocol snappy compression should be disabled
         super().__init__(transport, cmd_id_offset=0, snappy_support=snappy_support)
-        self.capabilities = capabilities
-        self.listen_port = listen_port
 
-    def send_handshake(self) -> None:
-        # TODO: move import out once this is in the trinity codebase
-        from trinity._utils.version import construct_trinity_client_identifier
```
pipermerriam: @cburgdorf this is the biggest win, which is really a larger part of why this is a better approach, which is making the class less opinionated on what parameters it uses as a handshake and more of an API for doing the handshake so that the parameters for the handshake are better controlled by whatever code is sending the handshake.

PR #830, opened 2019-07-23
Add factory for transports that are directly connected

pipermerriam:
extracted from #684 

### What was wrong?

We have factory for doing direct in-memory communication using the `TransportAPI` that bypasses all of the encryption components but did not have one for creating real linked transport objects.

### How was it fixed?

Added a new `p2p.tools.factories.TransportPairFactory` which connects two `p2p.transport.Transport` objects together over an in-memory set of `asyncio` streams.  The utilities under `p2p.tools.paragon.helpers` have also been updated to make use of these factories.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

12-05-26 wlh Psych (1)C750

### DIFF
--- a/p2p/tools/asyncio_streams.py
+++ b/p2p/tools/asyncio_streams.py
@@ -1,38 +1,67 @@
 import asyncio
-from typing import cast, Any, Callable, Dict, Tuple
+from typing import Any, Dict, Tuple
 
 
-class MockTransport:
+class MemoryProtocol(asyncio.Protocol):
     def __init__(self) -> None:
-        self._is_closing = False
+        self._closed_event = asyncio.Event()
 
+    async def _drain_helper(self) -> None:
+        pass
+
+    @property
+    async def _closed(self) -> None:
+        await self._closed_event.wait()
+
+
+class MemoryTransport(asyncio.WriteTransport):
+    """
+    A fake version of the ``asyncio.BaseTransport``:
+
+    https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BaseTransport
+    """
+    def __init__(self,
+                 reader: asyncio.StreamReader,
+                 extra: Dict[str, Any] = None) -> None:
+        self._is_closing = False
+        self._reader = reader
+        super().__init__(extra)
+
+    #
+    # BaseTransport methods
+    #
+    # methods we don't overwrite because they already raise NotImplementedError
+    # and we don't need them
+    # - set_protocol
+    # - get_protocol
     def close(self) -> None:
         self._is_closing = True
 
     def is_closing(self) -> bool:
         return self._is_closing
 
+    #
+    # WriteTransport methods
+    #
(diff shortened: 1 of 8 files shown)


p2p/tools/asyncio_streams.py
```diff
@@ -1,5 +1,5 @@
 import asyncio
-from typing import cast, Any, Callable, Tuple
+from typing import cast, Any, Callable, Dict, Tuple
 
 
 class MockTransport:
```
cburgdorf: Shouldn't this inherit from `TransportAPI`?

p2p/tools/factories.py
```diff
@@ -47,26 +53,26 @@ def PublicKeyFactory() -> keys.PublicKey:
 
 class AddressFactory(factory.Factory):
     class Meta:
-        model = kademlia.Address
+        model = Address
```
cburgdorf: I'm not family with the inner workings of these factories but out of curiosity: Does hell break loose if the `model` were set to `AddressAPI`?

p2p/tools/paragon/helpers.py
```diff
@@ -62,79 +58,20 @@
             token=cancel_token,
         )
 
-    alice_private_key = alice_factory.privkey
```
cburgdorf: Wow, all this craziness is gone 🤯 :clap:

tests/p2p/test_transport_factories.py
```diff
@@ -0,0 +1,62 @@
...
+    rlp_messages = tuple(
+        rlp.encode(cmd.cmd_id, sedes.big_endian_int) + rlp.encode((msg,))
+        for msg in messages
+    )
+    asyncio.ensure_future(manage_bob(rlp_messages))
+    for message in messages:
+        header, body = cmd.encode({'data': message})
+        alice_transport.send(header, body)
+
+    await asyncio.wait_for(done.wait(), timeout=0.1)
```
cburgdorf: Did you try using `0.01`  or `0.05` or some other lower number?

p2p/tools/factories.py
pipermerriam: Yes, because it would try to instantiate `AddressAPI` with whatever `kwargs` were provided.

tests/p2p/test_transport_factories.py
pipermerriam: Should work fine at lower values but....  In reality this shouldn't slow down a passing test suite where-as a lower number *might* cause a false negative in a slower CI environment so in cases like this where it isn't a sleep, but rather a wait, I've tended towards larger numbers.  This is only to be sure the test suite doesn't hang.

cburgdorf: Makes perfect sense!

p2p/tools/factories.py
cburgdorf: :+1:

PR #832, opened 2019-07-23
Remove local clamp util in favor of eth-utils version

pipermerriam:
### What was wrong?

`eth-utils==1.5.2` added a numeric `clamp` utility.  We no longer need the one from `p2p._utils`

### How was it fixed?

Removed the one from `p2p._utils` in favor of using the one from `eth-utils`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

11-05-15 babies shear 2nd (11) C700Dude


pipermerriam:
Fixed in https://github.com/ethereum/eth-utils/pull/164

newsfragments/832.feature.rst
```diff
@@ -0,0 +1 @@
+Remove ``p2p._utils.clamp`` in favor of the one from ``eth-utils>=1.5.2``
```
cburgdorf: I think the file name should either be `832.misc.rst` or `832.removal.rst` because otherwise it will be presented as a new feature to the user which it isn't.

With `832.removal.rst` it would be listed under the "Deprecations and Removals" section of the release notes which os ok but not strictly needed because that wasn't a public API in the first place.

`832.misc.rst` might be the best fight as it will just go into a section that links to closed issues without any further text.

pipermerriam: didn't know we had *removal*.  :heart:

PR #835, opened 2019-07-24
Implement multiplexer

pipermerriam:
Extracted from #684

depends on #836

depends on #838 

should be merged after #834 

### What was wrong?

In order to facilitate speaking multiple devp2p protocols, we need mechanisms for dividing the messages up by their individual protocol streams.  

### How was it fixed?

This is part of #684  It brings in a single component, the `p2p.multiplexer.Multiplexer`.  This class is intentionally **not** a service.  This is because it must be used outside of the service context during the handshake process to facilitate individual protocol handshakes streaming only the messages pertinent to that protocol.

This PR only introduces the class and the relevant tests for it in order to keep the PR adequately small.

You can see how this API is used in this commit: https://github.com/ethereum/trinity/pull/847/commits/c6fb7d85f4944b5c6a90362a7e23fa6d6e128281

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

17-05-17 Ramone Sheared (16)C1000

### DIFF
--- a/p2p/multiplexer.py
+++ b/p2p/multiplexer.py
@@ -39,17 +39,17 @@ from p2p.exceptions import (
 )
 from p2p.p2p_proto import P2PProtocol
 from p2p.protocol import (
-    PayloadType,
     Protocol,
 )
 from p2p.resource_lock import ResourceLock
+from p2p.typing import Payload
 
 
 async def stream_transport_messages(transport: TransportAPI,
                                     base_protocol: P2PProtocol,
                                     *protocols: ProtocolAPI,
                                     token: CancelToken = None,
-                                    ) -> AsyncIterator[Tuple[ProtocolAPI, CommandAPI, PayloadType]]:
+                                    ) -> AsyncIterator[Tuple[ProtocolAPI, CommandAPI, Payload]]:
     """
     Streams 3-tuples of (Protocol, Command, Payload) over the provided `Transport`
     """
@@ -100,7 +100,7 @@ class Multiplexer(CancellableMixin, MultiplexerAPI):
     _msg_counts: DefaultDict[Type[CommandAPI], int]
 
     _protocol_locks: ResourceLock
-    _protocol_queues: Dict[Type[ProtocolAPI], 'asyncio.Queue[Tuple[CommandAPI, PayloadType]]']
+    _protocol_queues: Dict[Type[ProtocolAPI], 'asyncio.Queue[Tuple[CommandAPI, Payload]]']
 
     def __init__(self,
                  transport: TransportAPI,
@@ -234,7 +234,7 @@ class Multiplexer(CancellableMixin, MultiplexerAPI):
     #
     async def stream_protocol_messages(self,
                                        protocol_identifier: Union[ProtocolAPI, Type[ProtocolAPI]],
-                                       ) -> AsyncIterable[Tuple[CommandAPI, PayloadType]]:
+                                       ) -> AsyncIterable[Tuple[CommandAPI, Payload]]:
         """
         Stream the messages for the specified protocol.
         """
@@ -253,9 +253,20 @@ class Multiplexer(CancellableMixin, MultiplexerAPI):
 
         async with self._protocol_locks.lock(protocol_class):
             msg_queue = self._protocol_queues[protocol_class]
+            if not hasattr(self, '_multiplex_token'):
+                raise Exception("Multiplexer is not multiplexed")
+            token = self._multiplex_token
 
-            while not self.is_closing:
-                yield await self.wait(msg_queue.get(), token=self._multiplex_token)
+            while not self.is_closing and not token.triggered:
+                try:
+                    # We use an optimistic strategy here of using
+                    # `get_nowait()` to reduce the number of times we yield to
+                    # the event loop.  Since this is an async generator it will
+                    # yield to the loop each time it returns a value so we
+                    # don't have to worry about this blocking other processes.
+                    yield msg_queue.get_nowait()
+                except asyncio.QueueEmpty:
+                    yield await self.wait(msg_queue.get(), token=token)
 
     #
     # Message reading and streaming API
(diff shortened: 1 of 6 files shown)


pipermerriam:
@carver maybe hold on the review until the dependent PR's are merged and I can rebase this.

carver:
> @carver maybe hold on the review until the dependent PR's are merged and I can rebase this.

Ok, sounds good. I was :fearful: looking at the +/-

pipermerriam:
@carver I have one more piece I'll split out tomorrow but if you want to look at `p2p.multiplexer.Multiplexer` and the corresponding tests in `tests/p2p/test_multiplexer.py` those are not going to change.

pipermerriam:
@carver https://github.com/ethereum/trinity/pull/850 and then I can rebase down to my intended review target.

pipermerriam:
Alright @carver .  This is as small as it is getting without doing something like committing the `p2p.abc.MultiplexerAPI` class on it's own or delaying the `MultiplexerPairFactory` to come after this PR, neither of which seem right to me.

carver:
Okay, great. I am on it. Sorry that I didn't get it done today. I promise, first thing tomorrow.

tests/p2p/test_multiplexer.py
```diff
@@ -0,0 +1,220 @@
...
+@pytest.mark.asyncio
+async def test_multiplexer_only_p2p_protocol():
+    alice_multiplexer, bob_multiplexer = MultiplexerPairFactory()
+    async with alice_multiplexer.multiplex():
+        async with bob_multiplexer.multiplex():
+            alice_stream = alice_multiplexer.stream_protocol_messages(P2PProtocol)
+            bob_stream = bob_multiplexer.stream_protocol_messages(P2PProtocol)
+
+            alice_p2p_protocol = alice_multiplexer.get_protocol_by_name('p2p')
+            bob_p2p_protocol = bob_multiplexer.get_protocol_by_name('p2p')
```
carver: Is there a reason to favor `by_name` over `by_type`? The `by_type` is nicer to constrain at type-checking time. 

And if we do prefer `by_type`, how bad would it be to drop the `by_name` option altogether?

tests/p2p/test_multiplexer.py
```diff
@@ -0,0 +1,220 @@
...
+
+            alice_p2p_protocol.send_ping()
+            cmd, _ = await asyncio.wait_for(bob_stream.asend(None), timeout=0.1)
```
carver: `.asend(None)` feels awkward, like it requires you to know something about the internals of how the stream works. Maybe it makes sense to add an alias, so you can do something like`bob_stream.get()`?

tests/p2p/test_multiplexer.py
```diff
@@ -0,0 +1,220 @@
...
+            alice_second_stream = alice_multiplexer.stream_protocol_messages(SecondProtocol)
+            bob_second_stream = bob_multiplexer.stream_protocol_messages(SecondProtocol)
+
+            alice_p2p_protocol = alice_multiplexer.get_protocol_by_name('p2p')
+            alice_second_protocol = alice_multiplexer.get_protocol_by_name('second')
+
+            bob_p2p_protocol = bob_multiplexer.get_protocol_by_name('p2p')
+            bob_second_protocol = bob_multiplexer.get_protocol_by_name('second')
+
+            alice_second_protocol.send_cmd(CommandA)
```
carver: This surprised me: you can send the uninitialized command? I would have expected `send_cmd(CommandA())`

Is it just because `CommandA` happens to have no data? I assume most commands (besides something like `Ping`) would include some kind of data. If it usually requires including data, then supporting the API of sending an class may not be worth being able to drop the parentheses.

tests/p2p/test_multiplexer.py
```diff
@@ -0,0 +1,220 @@
...
+            alice_p2p_protocol.send_ping()
+            alice_second_protocol.send_cmd(CommandB)
+            cmd, _ = await asyncio.wait_for(bob_p2p_stream.asend(None), timeout=0.1)
+            assert isinstance(cmd, Ping)
```
carver: Very cool, this test succinctly identifies the point of the mux, and shows it working. :+1:

tests/p2p/test_multiplexer.py
pipermerriam: These are just vanilla python async generators and I'm calling their internal APIs for getting the next value from them where normally you would use these as `async for thing in bob_stream`.

I can look into another way to structure the tests but this was the most direct way I could think of to test this code.  Any thoughts on how it might be improved?

tests/p2p/test_multiplexer.py
pipermerriam: I'll check the code that's using this in the bigger branch and see if I can remove the `by_name` APIs cleanly.

tests/p2p/test_multiplexer.py
pipermerriam: If you scroll up you can see that `send_cmd` is just a method I implemented on `SecondProtocol` to simplify sending different command types and isn't part of the core `Protocol` API.

p2p/multiplexer.py
```diff
@@ -0,0 +1,337 @@
...
+            self.cancel_token = base_token
+        else:
+            self.cancel_token = base_token.chain(token)
+
+        self._transport = transport
+        # the base `p2p` protocol instance.
+        self._base_protocol = base_protocol
+
+        # the sub-protocol instances
+        self._protocols = protocols
```
carver: If we're going to keep the lookup by name, we should probably validate that they there are no duplicate names here.

p2p/multiplexer.py
```diff
@@ -0,0 +1,337 @@
...
+        """
+        if isinstance(protocol_identifier, Protocol):
+            protocol_class = type(protocol_identifier)
+        elif isinstance(protocol_identifier, type) and issubclass(protocol_identifier, Protocol):
+            protocol_class = protocol_identifier
+        else:
+            raise TypeError("Unknown protocol identifier: {protocol}")
+
+        if not self.has_protocol(protocol_class):
+            raise Exception(f"TODO: Unknown protocol '{protocol_class}'")
```
carver: TODO?

p2p/multiplexer.py
```diff
@@ -0,0 +1,337 @@
...
+    @property
+    def remote(self) -> NodeAPI:
+        return self._transport.remote
+
+    @property
+    def is_closing(self) -> bool:
+        return self._transport.is_closing
+
+    def close(self) -> None:
+        # TODO: maybe graceflly close streams?
```
carver: Maybe move to an issue?

tests/p2p/test_multiplexer.py
carver: No, that's fine. Since the stream is just a vanilla async generator, I don't have any better suggestions. :+1:

tests/p2p/test_multiplexer.py
carver: Ah, right, I was thinking it would be something on every `Protocol` :+1:

p2p/multiplexer.py
pipermerriam: The token is actually closing the streams gracefully so I just removed the comment.

pipermerriam:
Test failures are unrelated.

PR #836, opened 2019-07-24
Add get_cmd_id_offsets helper function

pipermerriam:
extracted from #684 

### What was wrong?

For a given set of protocols we need to compute the offsets for their command ids

### How was it fixed?

Added a helper function for computing these offset values.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

15-01-17 Sport - P1150162 C1000

### DIFF
--- a/p2p/constants.py
+++ b/p2p/constants.py
@@ -166,3 +166,7 @@ KADEMLIA_ID_SIZE = 256
 
 # Maximum node `id` for a kademlia node
 KADEMLIA_MAX_NODE_ID = (2 ** KADEMLIA_ID_SIZE) - 1
+
+
+# Reserved command length for the base `p2p` protocol
+P2P_PROTOCOL_COMMAND_LENGTH = 16
--- a/p2p/constants.py
+++ b/p2p/constants.py
@@ -169,4 +169,5 @@ KADEMLIA_MAX_NODE_ID = (2 ** KADEMLIA_ID_SIZE) - 1
 
 
 # Reserved command length for the base `p2p` protocol
+# - https://github.com/ethereum/devp2p/blob/master/rlpx.md#message-id-based-multiplexing
 P2P_PROTOCOL_COMMAND_LENGTH = 16
--- a/p2p/protocol.py
+++ b/p2p/protocol.py
@@ -13,7 +13,7 @@ from typing import (
 import snappy
 
 from eth_utils import to_tuple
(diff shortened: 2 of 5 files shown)


p2p/constants.py
```diff
@@ -166,3 +166,7 @@
 
 # Maximum node `id` for a kademlia node
 KADEMLIA_MAX_NODE_ID = (2 ** KADEMLIA_ID_SIZE) - 1
+
+
+# Reserved command length for the base `p2p` protocol
+P2P_PROTOCOL_COMMAND_LENGTH = 16
```
lithp: It could be nice to include a link here to https://github.com/ethereum/devp2p/blob/master/rlpx.md#message-id-based-multiplexing, as explanation for why cmd length is 16 even though there are only four commands.

p2p/protocol.py
```diff
@@ -226,3 +227,11 @@ def _pad_to_16_byte_boundary(data: bytes) -> bytes:
...
         data += NULL_BYTE * (16 - remainder)
     return data
+
+
+def get_cmd_offsets(protocol_types: Tuple[Type[ProtocolAPI], ...]) -> Tuple[int, ...]:
+    return tuple(accumulate(
+        lambda prev_offset, protocol_class: prev_offset + protocol_class.cmd_length,
+        protocol_types,
+        P2P_PROTOCOL_COMMAND_LENGTH,
+    ))[:-1]
```
lithp: Now that I know what it does the code as-is makes sense but the easiest way for me to figure out what it was doing was to play with the implementation. Dropping what I came up here with for inspiration, I think a comment explaining why the `[:-1]` exists could be valuable.

```python
def get_cmd_offsets(protocol_types: Tuple[Type[ProtocolAPI], ...]) -> Tuple[int, ...]:
    protocol_lengths = (protocol_type.cmd_length for protocol_type in protocol_types)
    offsets = accumulate(operator.add, protocol_lengths, P2P_PROTOCOL_COMMAND_LENGTH)
    return tuple(take(len(protocol_types), offsets))  # don't compute more offsets than necessary```

lithp: Also, I don't know how you plan on using this method, but it might be slightly cleaner and more general if, instead of hard-coding `P2P_PROTOCOL_COMMAND_LENGTH`, it was always passed `P2PProtocol` as its first argument.

PR #827, opened 2019-07-23
Fix randomness in eth2 tests

ralexstokes:
### What was wrong?

Fix #821.

A misuse of `random.randint` was causing an algorithm to not terminate when it is designed to in normal usage -- a test was exercising it in a valid but non-canonical way...

### How was it fixed?

Fix the test and other uses of `random.randint` in the `eth2` codebase.

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #846, opened 2019-07-25
Fix #842, raise exception in find proposer

### LINKED ISSUE #842: Investigate `_find_proposer_in_committee`
### What is wrong?

The cause of #821 is that `_find_proposer_in_committee` took an incorrect input, which was fixed in #827. The bug lurked in perfect stealth because the `while True` loop in the function body raises no error and hangs forever. 

It might not be a good idea to raise an error if it gets too many rounds or it might block the consensus.

### How can it be fixed

- Calculate the expected number of rounds
- Raise warnings if the rounds go too high

ChihChengLiang:
### What was wrong?

#842

### How was it fixed?

Add docstring and raise an exception.

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/committee_helpers.py
+++ b/eth2/beacon/committee_helpers.py
@@ -97,6 +97,16 @@ def _find_proposer_in_committee(
     A validator with higher balance would be chosen as the proposer more likely.
     It is expected to end in just 1 or 2 rounds.
     More than `MAX_ROUNDS` rounds is rare and could consider as a bug.
+
+    Detail:
+    The committee passed in here should consist 'active' validators.
+    An active validator has a balance of at least 17 Ether and at most 32 Ether.
+    This function choose a number between 0 and 1, which is represented by
+    `random_byte / MAX_RANDOM_BYTE`. The probability of a validator chosen as
+    a proposer is `effective_balance/max_effective_balance`.
+    The worst/easiest possible scenario for the loop to reach more rounds is when every
+    validator has 17 Ether and has the 17/32 probability of being chosen.
+    This requires 1 out of (17/32)^100 chance to reach 100 rounds.
     """
     base = int(epoch)
     committee_len = len(committee)
--- a/eth2/beacon/committee_helpers.py
+++ b/eth2/beacon/committee_helpers.py
@@ -7,6 +7,7 @@ import ssz
 from eth2._utils.hash import hash_eth2
 from eth2._utils.tuple import update_tuple_item
 from eth2.beacon.constants import MAX_INDEX_COUNT, MAX_RANDOM_BYTE
+from eth2.beacon.exceptions import ImprobableToReach
 from eth2.beacon.helpers import get_active_validator_indices, get_seed
 from eth2.beacon.types.compact_committees import CompactCommittee
 from eth2.beacon.types.states import BeaconState
@@ -81,6 +82,9 @@ def get_start_shard(state: BeaconState, epoch: Epoch, config: CommitteeConfig) -
     return shard
 
 
+MAX_ROUNDS = 100
+
+
 def _find_proposer_in_committee(
     validators: Sequence[Validator],
     committee: Sequence[ValidatorIndex],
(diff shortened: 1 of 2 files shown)


eth2/beacon/committee_helpers.py
```diff
@@ -113,16 +113,24 @@ def _find_proposer_in_committee(validators: Sequence[Validator],
...
+    A validator with higher balance would be chosen as the proposer more likely.
+    It is expected to end in just 1 or 2 rounds.
+    More than 10 rounds is rare and could consider as a bug.
+    """
     base = int(epoch)
-    i = 0
     committee_len = len(committee)
-    while True:
+    i = 0
+    while i < range(committee_len):
```
pipermerriam: are you indending this to be `for i in range(committee_len)` or `while i < committee_len`?

ChihChengLiang: Ah nice catch 🙈

tests/eth2/core/beacon/test_committee_helpers.py
```diff
@@ -150,10 +154,18 @@ def test_get_start_shard(genesis_state,
...
 
 
+@given(st.data())
 def test_find_proposer_in_committee(genesis_validators,
-                                    config):
-    epoch = random.randrange(config.GENESIS_EPOCH, 2**64)
-    proposer_index = random.randrange(0, len(genesis_validators))
+                                    config,
+                                    data):
+    epoch = data.draw(st.integers(
```
ralexstokes: after adding the `pytest-randomly` plugin in #827, we get all of the features here with `random.randrange` instead of `data.draw`.

imo there is less fanfare to use the stdlib's `random` than to pull in all of the `hypothesis` machinery if we just need a random element from a basic set. i would suggest using `hypothesis` if we want to do any property-based testing

any particular reason we use `hypothesis` here or do you see it differently?

eth2/beacon/committee_helpers.py
```diff
@@ -113,16 +113,24 @@ def _find_proposer_in_committee(validators: Sequence[Validator],
...
+    A validator with higher balance would be chosen as the proposer more likely.
+    It is expected to end in just 1 or 2 rounds.
+    More than 10 rounds is rare and could consider as a bug.
+    """
     base = int(epoch)
-    i = 0
     committee_len = len(committee)
-    while True:
+    i = 0
+    while i < committee_len:
```
ralexstokes: i'd suggest not using `committee_len` here -- especially for some testing scenarios, the size of the committee could just be 1 or 2, where we may not find a candidate in that number of rounds.

i think a `MAX_ROUNDS` approach will protect us from the bug we want to avoid and be "future-proof" in the sense that we should never have to come back and fiddle w/ this number in order to get some test to pass.

perhaps try `MAX_ROUNDS = 100` and then do the `while/else`?

tests/eth2/core/beacon/test_committee_helpers.py
ChihChengLiang: Yeah, `MAX_ROUNDS` with `random` sounds good. Was shooting myself on the feet with the committee_len approach.

eth2/beacon/committee_helpers.py
```diff
@@ -113,16 +114,24 @@ def _find_proposer_in_committee(validators: Sequence[Validator],
...
+    i = 0
+    while i < MAX_ROUND:
         candidate_index = committee[(base + i) % committee_len]
         random_byte = hash_eth2(seed + (i // 32).to_bytes(8, "little"))[i % 32]
         effective_balance = validators[candidate_index].effective_balance
         if effective_balance * MAX_RANDOM_BYTE >= max_effective_balance * random_byte:
             return candidate_index
         i += 1
+    else:
+        raise ValueError(f"The i={i} is too large")
```
ralexstokes: yes the `i` would be too large :) but this doesn't really tell a client of this code what went wrong...

perhaps something like `ValueError(f"Invariant violated: search for a proposer failed after {MAX_ROUNDS} rounds.")` i could also get nitpicky and ask if the intention of `ValueError` is what we want here...

eth2/beacon/constants.py
```diff
@@ -24,6 +24,8 @@
 
 MAX_RANDOM_BYTE = 2**8 - 1
 
+MAX_ROUND = 100
```
ralexstokes: two options here:

1. change the symbol `MAX_ROUND` to something more specific
2. change the scope of its definition so that `MAX_ROUND` is as specific as possible to `_find_proposer_in_committee`.

i'd think it is clear enough to just define alongside the function (e.g. on line 111 in `committee_helpers.py`) but this may be pending the python style too much.

either way i'd also say `MAX_ROUNDS` (plural), rather than singular, just for the readability of the name.

if we keep this constant in the `constants.py` location, then perhaps `MAX_ROUNDS_OF_PROPOSER_SEARCH` or something...? it would almost be nice to indicate somehow that this is an artifact of the code, rather than a part of the specification of the protocol but i can't think of a nice way to do this that also aligns w/ general python style...

eth2/beacon/committee_helpers.py
```diff
@@ -113,16 +114,24 @@ def _find_proposer_in_committee(validators: Sequence[Validator],
                                 epoch: Epoch,
                                 seed: Hash32,
                                 max_effective_balance: Gwei) -> ValidatorIndex:
+    """
+    Loop through the validators in the committee one by one.
+    A validator with higher balance would be chosen as the proposer more likely.
+    It is expected to end in just 1 or 2 rounds.
+    More than MAX_ROUND rounds is rare and could consider as a bug.
```
ralexstokes: nitpick: s/MAX_ROUND/``MAX_ROUND`` to show the "this is code" affordance

eth2/beacon/committee_helpers.py
ChihChengLiang: Good suggestions. I added `ImprobableToReach` exceptions. I'm also thinking of adding these exceptions in future PR will be more complete and useful:

```python
class InvariantError(PyEVMError):
    pass

class Unreachable(InvariantError):
    pass

class ImprobableToReach(InvariantError):
    pass

```

hwwhww:
@ChihChengLiang could you rebase it? Thanks.

PR #847, opened 2019-07-25
Use multiplexer in peer class

pipermerriam:
Builds from #835 

### What was wrong?

The `p2p.multiplexer.Multiplexer` is not actively being used.

### How was it fixed?

Integrated it into the `BasePeer` class to handling the message stream.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/multiplexer.py
+++ b/p2p/multiplexer.py
@@ -2,7 +2,6 @@ import asyncio
 import collections
 import logging
 from typing import (
-    AsyncIterable,
     AsyncIterator,
     cast,
     Dict,
@@ -15,7 +14,7 @@ from typing import (
 
 from async_generator import asynccontextmanager
 
-from cancel_token import CancelToken, OperationCancelled
+from cancel_token import CancelToken
 
 from eth_utils.toolz import cons
 
@@ -38,9 +37,7 @@ from p2p.exceptions import (
     UnknownProtocolCommand,
 )
 from p2p.p2p_proto import P2PProtocol
-from p2p.protocol import (
-    Protocol,
-)
+from p2p.protocol import Protocol
 from p2p.resource_lock import ResourceLock
 from p2p.typing import Payload
 
@@ -221,7 +218,7 @@ class Multiplexer(CancellableMixin, MultiplexerAPI):
     #
     async def stream_protocol_messages(self,
                                        protocol_identifier: Union[ProtocolAPI, Type[ProtocolAPI]],
-                                       ) -> AsyncIterable[Tuple[CommandAPI, Payload]]:
+                                       ) -> AsyncIterator[Tuple[CommandAPI, Payload]]:
         """
         Stream the messages for the specified protocol.
         """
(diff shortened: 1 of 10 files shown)


p2p/multiplexer.py
```diff
@@ -300,25 +297,22 @@ def get_protocols(self) -> Tuple[ProtocolAPI, ...]:
             *self._protocols,
             token=token,
         ), token=token)
-        try:
```
pipermerriam: This `try/except` doesn't appear to be necessary.

p2p/peer.py
```diff
@@ -416,49 +395,51 @@ def get_protocol_command_for(self, msg: bytes) -> CommandAPI:
...
+    async def handle_p2p_proto_stream(self, multiplexer: MultiplexerAPI) -> None:
         """Handle the base protocol (P2P) messages."""
-        if isinstance(cmd, Disconnect):
-            msg = cast(Dict[str, Any], msg)
-            try:
-                reason = DisconnectReason(msg['reason'])
-            except TypeError:
-                self.logger.info('Unrecognized reason: %s', msg['reason'])
+        async for cmd, msg in self.wait_iter(multiplexer.stream_protocol_messages(self.base_protocol)):  # noqa: E501
+            if isinstance(cmd, Disconnect):
```
carver: If you like it better: leaving `handle_p2p_msg` untouched and calling it from here. That gives the option of later doing something like #857 cleanly. (Plus it makes the diff smaller here)

p2p/peer.py
```diff
@@ -416,49 +395,51 @@ def get_protocol_command_for(self, msg: bytes) -> CommandAPI:
...
+                self.logger.warning("Peer %s has no subscribers, discarding %s msg", self, cmd)
 
-    def process_msg(self, cmd: CommandAPI, msg: Payload) -> None:
-        if cmd.is_base_protocol:
-            self.handle_p2p_msg(cmd, msg)
-        else:
-            self.handle_sub_proto_msg(cmd, msg)
+    def handle_sub_proto_msg(self, cmd: CommandAPI, msg: Payload) -> None:
+        """
+        Hook for peer classes to do something on a given message
```
carver: Why switch the previous mechanism of sub-classing? Just didn't want to have to call `super()`?

I was going to suggest the same thing as with `handle_p2p_msg()`, but the new construction makes that more awkward.

p2p/peer.py
```diff
@@ -361,36 +359,17 @@ def get_protocol_command_for(self, msg: bytes) -> CommandAPI:
...
-                )
-                return
-            except MalformedMessage as err:
-                await self.disconnect(DisconnectReason.bad_protocol)
-                return
-
-            try:
-                self.process_msg(cmd, msg)
-            except RemoteDisconnected as e:
-                if self.uptime < BLACKLIST_SECONDS_QUICK_DISCONNECT:
```
carver: Looks like we lost this check altogether. Just checking if that's intentional, or coming back later.

carver:
Looks like you're still actively working on this, code keeps changing during review. Can you ping again when it's ready?

pipermerriam:
@carver sorry, yeah, I thought it was done and then realized I had some cleanup.  I'll tidy it up and ping you again when it's ready.

carver:
> @carver sorry, yeah, I thought it was done and then realized I had some cleanup. I'll tidy it up and ping you again when it's ready.

Heh, no worries. Sounds good! :+1:

p2p/peer.py
pipermerriam: It was intentional.  I flip-flopped on my stance on the quick-disconnect ban and I no longer think it's right (at least at this stage).  I'm not convinced it adds value at this stage and it could very easily have un-intended negative side effects like making it harder to do development against a local node that you start and stop a lot.

p2p/peer.py
pipermerriam: Your approach of just calling the handle method from the stream method makes sense. I'll update to use that approach.

p2p/service.py
```diff
@@ -318,12 +318,13 @@ def cancel_nowait(self) -> None:
...
-        for task in self._tasks:
+        tasks = tuple(self._tasks)
+        for task in tasks:
             task.cancel()
-            try:
-                await task
-            except asyncio.CancelledError:
-                pass
+        await asyncio.wait_for(
+            asyncio.gather(*tasks, loop=self.get_event_loop(), return_exceptions=True),
```
pipermerriam: @cburgdorf suggested this back when I originally made the change.  I added this to see if it made the shutdown go any smoother but it doesn't appear to help so I'm going to revert it.

pipermerriam:
@carver this is ready for you (for real this time)

p2p/service.py
```diff
@@ -151,7 +151,7 @@ def run_task(self, awaitable: Awaitable[Any]) -> None:
             except OperationCancelled:
                 pass
             except Exception as e:
-                self.logger.warning("Task %s finished unexpectedly: %s", awaitable, e)
+                self.logger.warning("Task %s finished unexpectedly: %r", awaitable, e)
```
carver: Ah yeah, that looks helpful.

tests/p2p/test_peer_collect_sub_proto_msgs.py
```diff
@@ -22,13 +22,11 @@
         remote.sub_proto.send_broadcast_data(b'broadcast-b')
         remote.sub_proto.send_get_sum(7, 8)
         remote.sub_proto.send_broadcast_data(b'broadcast-c')
-        await asyncio.sleep(0.01)
+        # yield to let remote and peer transmit.
+        await asyncio.sleep(0.1)
```
carver: Was this something you tinkered with during implementation, or does it need to stay in? 100ms is on the long side for a test, just wondering if we can drop it back down to `0.01`.

pipermerriam: Ended up needing to be longer but I just pushed a change to see if I can shorten it a little.

PR #848, opened 2019-07-25
[Turbo trinity-eth2] Exploring beacon state wrapper

hwwhww:
### What was wrong?

Running Mainnet test vectors is too painful. :'(

Before this PR, the profiling of `test_demo.py`:
xx

The bottleneck is, for immutability, we do `copy.deepcopy` so frequently in `py-ssz`.

This PR demonstrates the significant performance improvement by sacrificing the immutability of `BeaconState` by adding a `dataclass` wrapper.

### How was it fixed?

* Most of the logic updates are in https://github.com/hwwhww/trinity/blob/56c5343fadce90c7f66df1420f284e4e5c5ecb0b/eth2/beacon/types/states.py
	* Rename original SSZ `BeaconState` to `SSZBeaconState`. (To @ralexstores: I know you suggested renaming the wrapper, but I prefer to not increasing too many git diff in this PR now. :) )
	* Add dataclass `beaconState`. 
		* Set class variable `ssz_class = SSZBeaconState` so we can transform to SSZ object or generate `state: BeaconState` from the SSZ object.
		* Customize optimized `hash_tree_root`!!! <- `hash_tree_root` is the new bottleneck.
* Wherever `state.copy(field=x)`, changing it to simply data assign (`state.field = x` - therefore some functions are impure now.
* When processing state transition (not proposing), the only place we require immutability is `apply_state_transition` function, so I add [`        state = state.copy()` in it](https://github.com/hwwhww/trinity/blob/8bb7801496b44c9ab2f3179e725f708b72ceeb9a/eth2/beacon/state_machines/forks/serenity/state_transitions.py#L31) .
* Add a stub `impure` decorator - just a mark now.

The new profiling:
yy

#### Improvements
- **`247.03 seconds` -> `130.52 seconds`** improvement for `test_demo.py`.
- **`1971.76 seconds` -> `444.05 seconds`** improvement for `test_sanity.py` + Mainnet tests.

#### Pros:
* Performance!
* More custom types friendly! `mypy` found more ambiguous assignments after I use dataclass.
* `state.balances` and `state.validators` is a high-value target to optimize in the next step.
* It's also valuable to hack the `Vector` fields.
* Helpful to optimize `BeaconState` storage and parallelization.

#### Cons:
* Dangerous mutability!
* Too hacky to deal with SSZ inner meta.
* Annoying to switch between `BeaconState` and `SSZBeaconState`.

IMO basically, it's reasonable to add a custom `BeaconState` wrapper. Not sure if `dataclass` is the best choice here, or just define a `class` for higher flexibility.

/cc @pipermerriam 

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


eth2/__init__.py
```diff
@@ -1,2 +1,19 @@
+from functools import wraps
+from typing import Any, Callable
+
 # This is to ensure we call setup_extended_logging() before anything else.
 import eth as _eth_module  # noqa: F401
+
+
+def impure(f: Callable[..., Any]) -> Any:
```
pipermerriam: I'm confused.  Doesn't this introduce a lot of `Any` types to our return values, reducing the effectiveness of `mypy`?

eth2/beacon/types/states.py
```diff
@@ -188,6 +194,204 @@ def __init__(
...
+                    for name in ssz_object._meta.field_names
+                }
+            )
+        )
+
+    @property
+    def ssz_object(self) -> SSZBeaconState:
+        return self.ssz_class(**asdict(self))
+
+    field_hash_tree_root_cache = {}  # type: Dict[Any, Hash32]
```
pipermerriam: mutable object at the class level.... is this intentional?  If so it should be documented why this is "ok"

PR #849, opened 2019-07-25
Simplify the ResourceLock

pipermerriam:
### What was wrong?

The `ResourceLock` introduced in #811 doesn't need to use an `asynccontextmanager`.  This is due to being able to simply return the lock value, resulting in the method being usable as-if it was written as an `asynccontextmanager`

### How was it fixed?

Just removed the async components.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

Spanish_Lace-home

### DIFF
--- a/p2p/resource_lock.py
+++ b/p2p/resource_lock.py
@@ -1,9 +1,6 @@
 import asyncio
 from collections.abc import Hashable
-from typing import AsyncIterator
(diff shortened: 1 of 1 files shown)


pipermerriam:
Doesn't seem like it needs a release notes entry.

PR #850, opened 2019-07-26
Implement factories for generating devp2p protocols

pipermerriam:
pulled from #835 

### What was wrong?

In working on testing the `Multiplexer` and `Connection` objects that are coming in #684 I found I needed protocols for testing purposes.

### How was it fixed?

Created factories for generating random Devp2p protocols and commands in `p2p.tools.factories`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/tools/factories.py
+++ b/p2p/tools/factories.py
@@ -209,15 +209,13 @@ STRUCTURE_SEDES = (
 
 
 @to_tuple
-def StructureFactory(*sedes: Tuple[str, Any],
-                     high_water_mark: int = 4,
+def StructureFactory(high_water_mark: int = 4,
                      ) -> Iterable[Tuple[str, Any]]:
-    yield from sedes
     for idx in range(high_water_mark):
         name = f"field_{idx}"
         sedes = random.choice(STRUCTURE_SEDES)
         yield (name, sedes)
-        if random.randint(idx, high_water_mark + 2) >= high_water_mark:
+        if random.randrange(idx, high_water_mark + 2) >= high_water_mark:
             break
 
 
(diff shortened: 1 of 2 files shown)


p2p/tools/factories.py
```diff
@@ -196,3 +200,140 @@ def MemoryTransportPairFactory(alice_remote: NodeAPI = None,
...
+@to_tuple
+def StructureFactory(*sedes: Tuple[str, Any],
+                     high_water_mark: int = 4,
+                     ) -> Iterable[Tuple[str, Any]]:
+    yield from sedes
+    for idx in range(high_water_mark):
+        name = f"field_{idx}"
+        sedes = random.choice(STRUCTURE_SEDES)
+        yield (name, sedes)
+        if random.randint(idx, high_water_mark + 2) >= high_water_mark:
```
carver: *nit* I agree with @ralexstokes that `randrange` is the nicer API because it matches slices & `range` semantics on the upper boundary. https://docs.python.org/3/library/random.html#random.randrange

What do you think?

p2p/tools/factories.py
```diff
@@ -196,3 +200,140 @@ def MemoryTransportPairFactory(alice_remote: NodeAPI = None,
...
+
+def ProtocolFactory(name: str = None,
+                    version: int = None,
+                    commands: Tuple[Type[Command], ...] = None) -> Type[Protocol]:
+    if name is None:
+        name = ProtocolNameFactory()
+    if version is None:
+        version = 1
+    if commands is None:
+        num_commands = random.randint(1, 6)
```
carver: Another `randrange` opportunity

PR #851, opened 2019-07-26
Add fixture test harness

ralexstokes:
Hooks into pytest for fixture loading, pulls most of the test parsing out into library code under `eth2/beacon/tools/fixtures`

Also adapts the `sanity` and `shuffling` tests to use this format and introduces the `epoch_processing` fixture tests.

#### Cute Animal Picture


eth2/beacon/tools/fixtures/parser.py
```diff
@@ -0,0 +1,85 @@
...
+    return TestSuite(config, test_handler, test_suite_data)
+
+
+def _search_for_dir(target_dir, p):
+    for child in p.iterdir():
+        if not child.is_dir():
+            continue
+        if child.name == target_dir.name:
+            return child
+    return None
```
pipermerriam: It has generally been a good pattern to avoid return-value-checking patterns in favor of exception raising (but I haven't seen how/where this is used yet)

eth2/beacon/tools/fixtures/parser.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def _find_project_root_dir(target: Path) -> Path:
+    """
+    Search the file tree for a path with a child directory equal to ``target``.
+    """
+    p = Path('.').resolve()
+    for _ in range(1000):
+        candidate = _search_for_dir(target, p)
```
pipermerriam: I think this is less prone to errors if `_search_for_dir` is changed to raise an exception and this changes to a `try/except`

eth2/beacon/tools/fixtures/parser.py
```diff
@@ -0,0 +1,85 @@
...
+        if child.name == target_dir.name:
+            return child
+    return None
+
+
+def _find_project_root_dir(target: Path) -> Path:
+    """
+    Search the file tree for a path with a child directory equal to ``target``.
+    """
+    p = Path('.').resolve()
```
pipermerriam: I think you can do this using the standard library https://docs.python.org/3.5/library/glob.html

eth2/beacon/tools/fixtures/test_case.py
```diff
@@ -1,30 +1,46 @@
...
-)
+from eth2.configs import Eth2Config
+
+from eth2._utils.bls import bls
+from eth2._utils.bls.backends import PyECCBackend
+
+from .test_handler import TestHandler
+
+
+def _select_bls_backend(bls_setting: int) -> None:
```
pipermerriam: Maybe use an `enum.Enum` for `bls_setting`?

eth2/beacon/tools/fixtures/test_case.py
```diff
@@ -1,30 +1,46 @@
...
+    if bls_setting == 2:
+        bls.use_noop_backend()
+    elif bls_setting == 1:
+        bls.use(PyECCBackend)
+    else:  # do not verify BLS to save time
```
pipermerriam: This implicit fallback seems like a good way to accidentally be using the noop backend when you intended to use something else.  I think this fits the "avoid implicit else clauses" guideline we've been moving towards.

eth2/beacon/tools/fixtures/test_gen.py
```diff
@@ -0,0 +1,101 @@
...
+            if handler_filter(handler) or handler.name == "core":
+                selected_handler = (test_type, handler)
+                selected_handlers += selected_handler
+    return itertools.product((selected_handlers,), config_types)
+
+
+def _generate_pytest_case_from(test_type, handler_type, config_type, test_case):
+    # special case only one handler "core"
+    if len(test_type.handlers) == 1 or handler_type.name == "core":
+        _id = (
```
pipermerriam: maybe name this `test_id`?

eth2/beacon/tools/fixtures/test_gen.py
```diff
@@ -0,0 +1,101 @@
...
+    function object) and inject them via ``metafunc.parametrize``.
+    """
+    eth2_fixture_request = _read_request_from_metafunc(metafunc)
+    test_suite_descriptors = _generate_test_suite_descriptors_from(eth2_fixture_request)
+    pytest_cases = _generate_pytest_cases_from_test_suite_descriptors(test_suite_descriptors)
+    argvals = tuple()
+    ids = tuple()
+    for test, _id in pytest_cases:
+        argvals += (test,)
+        ids += (_id,)
```
pipermerriam: equivalent to:

```python
if pytest_cases:
    argvals, ids = zip(*pytest_cases)
else:
    argvals, ids = (), ()  # only needed because of zero-length edge case.

File: eth2/beacon/tools/fixtures/test_suite.py [hunk 8]
```diff
@@ -0,0 +1,35 @@
...
+            config,
+        )
+
+
+class TestSuite:
+    def __init__(self,
+                 config: Eth2Config,
+                 test_handler: TestHandler,
+                 test_suite_data: Dict[str, Any]) -> None:
+        self.test_cases = _parse_test_cases(
```
pipermerriam: Since `_parse_test_cases` produces a generator this property will be a generator and will only be able to be consumed once (after which it will appear empty).  Is this what you intended?

eth2/beacon/tools/fixtures/test_suite.py
```diff
@@ -0,0 +1,35 @@
...
+
+from eth2.configs import Eth2Config
+
+from .test_case import TestCase
+from .test_handler import TestHandler
+
+
+def _parse_test_cases(config: Eth2Config,
+                      test_handler: TestHandler,
+                      test_cases: Sequence[Dict[str, Any]]) -> Iterable[TestCase]:
```
pipermerriam: If this is truly intended to be a generator I think it should be specifically typed as such to make it clear.

eth2/beacon/tools/fixtures/test_types/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+    @classmethod
+    def build_path(cls,
+                   tests_root_path: Path,
+                   test_handler: TestHandler,
+                   config_type: ConfigType) -> Path:
+        if len(cls.handlers) == 1:
+            file_name = f"{cls.name}_{config_type.name}.yaml"
+        else:
+            file_name = f"{cls.name}_{test_handler.name}_{config_type.name}.yaml"
+        return tests_root_path / Path(cls.name) / Path(test_handler.name) / Path(file_name)
```
pipermerriam: It'd be good to have a type hint on this class indicating that it has a name property even if it's not actually declared and to use `abc.ABC` since this is sort of an abstract base class definition even if it doesn't have any `abstractmethod`

```python
class TestType:
    name: str
```

eth2/beacon/tools/fixtures/test_types/bls.py
```diff
@@ -0,0 +1,5 @@
+from . import TestType
+
+
+class BLS(TestType):
+    pass
```
pipermerriam: Should this have a `name` property?

eth2/beacon/tools/fixtures/test_types/epoch_processing.py
```diff
@@ -0,0 +1,92 @@
...
+        return from_formatted_dict(
+            test_case_data["post"],
+            BeaconState,
+        )
+
+    def run_with(self,
+                 inputs: BeaconState,
+                 config: Eth2Config) -> BeaconState:
+        state = inputs
+        return self.__class__.processor(state, config)
```
pipermerriam: It looks like you could make this a `classmethod` to avoid the `self.__class__` since this method doesn't use `self`

Also probably good to add a type hint to this class for the `processor` so that it's clear what that property should be.

eth2/beacon/tools/fixtures/test_handler.py
```diff
@@ -0,0 +1,9 @@
+from typing import (
+    Any,
+    Dict,
+)
+
+
+class TestHandler:
+    def valid(self, data: Dict[str, Any]) -> bool:
+        return True
```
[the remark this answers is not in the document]
ralexstokes: i'm open to this.... how do we want to handle disparate return types for the methods though? generally, each test has a different signature

eth2/beacon/tools/fixtures/config_types.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Mainnet(ConfigType):
+    # name is the human-readable name for this configuration.
+    name = "mainnet"
+    # path is the file system path to the config as YAML, relative to the project root.
+    path = "tests/eth2/fixtures/mainnet.yaml"
+
+
+class Full(ConfigType):
```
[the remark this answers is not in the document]
ralexstokes: yep, good call

eth2/beacon/tools/fixtures/test_types/__init__.py
```diff
@@ -0,0 +1,18 @@
+from pathlib import Path
+
+from eth2.beacon.tools.fixtures.config_types import ConfigType
+from eth2.beacon.tools.fixtures.test_handler import TestHandler
+
+
+class TestType:
```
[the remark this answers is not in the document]
ralexstokes: i've seen both names used to refer to this concept... do you think `TestRunner` makes more sense?

eth2/beacon/tools/fixtures/parser.py
ralexstokes: i'd say the exception approach definitely makes sense if we want to signal an error. in this case, i'd say the semantics are less "error condition" and more just "absence of a file path"... we can also skip defining an explicit exception tree and using the `try`/`catch` machinery

eth2/beacon/tools/fixtures/parser.py
ralexstokes: fair enough

eth2/beacon/tools/fixtures/test_case.py
ralexstokes: nice call

eth2/beacon/tools/fixtures/test_case.py
ralexstokes: yes, agreed it is safer

eth2/beacon/tools/fixtures/test_gen.py
ralexstokes: don't like leading underscores, eh?

eth2/beacon/tools/fixtures/test_gen.py
ralexstokes: yeah, this is nicer

eth2/beacon/tools/fixtures/test_suite.py
ralexstokes: should be fine -- i think i'm leaning towards removing this `TestSuite` class as it touches a concept in the fixtures tests but doesn't seem to add much value to this library

eth2/beacon/tools/fixtures/test_types/__init__.py
ralexstokes: good call

eth2/beacon/tools/fixtures/test_types/bls.py
ralexstokes: yeah -- haven't addressed this type of test yet in this PR

ralexstokes:
@hwwhww and @pipermerriam took your suggestions to enhance the typing and/or make it clearer where an interface exists/where certain properties or methods are expected...

going to go ahead and merge this now but happy to address any post-merge PR feedback :)

PR #852, opened 2019-07-26
Update slashing penalty calculation

ralexstokes:
### What was wrong?

We factored out the `EFFECTIVE_BALANCE_INCREMENT` to avoid overflowing a `uint64`-sized type in the spec and I thought it would not introduce a change under Python but it was causing us to fail some epoch processing spec tests.

### How was it fixed?

Use the calculation the spec uses.

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/state_machines/forks/serenity/epoch_processing.py
+++ b/eth2/beacon/state_machines/forks/serenity/epoch_processing.py
@@ -701,13 +701,11 @@ def process_registry_updates(state: BeaconState, config: Eth2Config) -> BeaconSt
 
 def _determine_slashing_penalty(total_penalties: Gwei,
                                 total_balance: Gwei,
-                                balance: Gwei) -> Gwei:
-    return Gwei(
-        balance * min(
-            total_penalties * 3,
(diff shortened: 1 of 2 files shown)


eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -701,13 +701,11 @@ def process_registry_updates(state: BeaconState, config: Eth2Config) -> BeaconSt
...
                                 total_balance: Gwei,
-                                balance: Gwei) -> Gwei:
-    return Gwei(
-        balance * min(
-            total_penalties * 3,
-            total_balance,
-        ) // total_balance
-    )
+                                balance: Gwei,
+                                increment: int) -> Gwei:
```
ChihChengLiang: Should this be type 'Gwei'?

ralexstokes: yep, thanks for pointing that out -- couldn't recall at the time if we were using `int` or `Gwei` for this

PR #855, opened 2019-07-26
Parallel beam sync execution

carver:
### What was wrong?

Beam Sync couldn't keep up

### How was it fixed?

Look ahead to upcoming blocks. Download state required to run those blocks, so that almost all of the state required is available by the time the canonical chain catches up.

Using this strategy, I've been often able to keep up with mainnet when connected to a peer on a local network.

~~I'll peel out pieces of this into smaller chunks until this is a reasonable size to review. It's just a reference point for the other PRs I peel out.~~

In order for this to run locally, you ~~have to patch~~ *need the latest master from* py-evm. ~~I'll add the patch to a comment below.~~

The part I'm least happy with is how the ancestor headers are handled in the preview execution: they are preemptively inserted to the database. This means that other places that assume some meaning in the header being present in the database (like the header syncer) could be confused. The header syncer had to be updated to check if the header was *really* present by also checking for the score.

A close runner up for hackiness is how preview stats are collected. There's some fairly awkward subclass patching. I haven't come up with anything better, yet.

Open to ideas on both!

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/chain.py
+++ b/trinity/sync/beam/chain.py
@@ -580,7 +580,7 @@ class BeamBlockImporter(BaseBlockImporter, BaseService):
         """
         senders = [transaction.sender for transaction in transactions]
         recipients = [transaction.to for transaction in transactions if transaction.to]
-        addresses = set(senders + recipients)
+        addresses = set(senders + recipients + [header.coinbase])
         collected_nodes = await self._state_downloader.download_accounts(
             addresses,
             parent_state_root,
--- a/trinity/sync/beam/chain.py
+++ b/trinity/sync/beam/chain.py
@@ -14,6 +14,7 @@ from eth.exceptions import (
 )
 from eth.rlp.blocks import BaseBlock
 from eth.rlp.headers import BlockHeader
+from eth.rlp.transactions import BaseTransaction
 from eth_typing import (
     Hash32,
     BlockNumber,
@@ -21,6 +22,7 @@ from eth_typing import (
 from eth_utils import (
     ValidationError,
 )
+import rlp
 
 from p2p.service import BaseService
 
@@ -38,6 +40,7 @@ from trinity.sync.common.events import (
     CollectMissingBytecode,
     CollectMissingStorage,
     DoStatelessBlockImport,
+    DoStatelessBlockPreview,
     MissingAccountCollected,
     MissingBytecodeCollected,
     MissingStorageCollected,
@@ -105,7 +108,13 @@ class BeamSyncer(BaseService):
             token=self.cancel_token,
         )
 
-        self._block_importer = BeamBlockImporter(chain, self._state_downloader, event_bus)
+        self._block_importer = BeamBlockImporter(
+            chain,
+            db,
+            self._state_downloader,
+            event_bus,
+            self.cancel_token,
+        )
         self._checkpoint_header_syncer = HeaderCheckpointSyncer(self._header_syncer)
         self._body_syncer = RegularChainBodySyncer(
             chain,
@@ -412,7 +421,7 @@ class HeaderOnlyPersist(BaseService):
             return self._final_headers
 
 
-class BeamBlockImporter(BaseBlockImporter, HasExtendedDebugLogger):
+class BeamBlockImporter(BaseBlockImporter, BaseService):
     """
     Block Importer that emits DoStatelessBlockImport and waits on the event bus for a
     StatelessBlockImportDone to show that the import is complete.
@@ -423,13 +432,22 @@ class BeamBlockImporter(BaseBlockImporter, HasExtendedDebugLogger):
     def __init__(
             self,
             chain: BaseAsyncChain,
+            db: BaseAsyncDB,
             state_getter: BeamDownloader,
-            event_bus: EndpointAPI) -> None:
+            event_bus: EndpointAPI,
+            token: CancelToken=None) -> None:
+        super().__init__(token=token)
+
         self._chain = chain
+        self._db = db
         self._state_downloader = state_getter
 
         self._blocks_imported = 0
         self._preloaded_account_state = 0
+        self._preloaded_previewed_account_state = 0
+        self._preloaded_account_time: float = 0
+        self._preloaded_previewed_account_time: float = 0
+        self._import_time: float = 0
 
         self._event_bus = event_bus
         # TODO: implement speculative execution, but at the txn level instead of block level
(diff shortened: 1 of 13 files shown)


carver:
py-evm patches:

Deal with canonical reorg when transactions were not added yet (in a header-only sync):
```diff
diff --git a/eth/db/chain.py b/eth/db/chain.py
index fa480d18..d41958b3 100644
--- a/eth/db/chain.py
+++ b/eth/db/chain.py
@@ -22,6 +22,9 @@ from eth_utils import (
 )
 
 from eth_hash.auto import keccak
+from trie.exceptions import (
+    MissingTrieNode,
+)
 
 from eth.constants import (
     EMPTY_UNCLE_HASH,
@@ -212,8 +215,13 @@ class ChainDB(HeaderDB, BaseChainDB):
             else:
                 old_header = cls._get_block_header_by_hash(db, old_hash)
                 old_canonical_headers.append(old_header)
-                for transaction_hash in cls._get_block_transaction_hashes(db, old_header):
-                    cls._remove_transaction_from_canonical_chain(db, transaction_hash)
+                try:
+                    for transaction_hash in cls._get_block_transaction_hashes(db, old_header):
+                        cls._remove_transaction_from_canonical_chain(db, transaction_hash)
+                except MissingTrieNode:
+                    # If the transactions were never stored for the (now) non-canonical chain,
+                    #   then you don't need to remove them from the canonical chain lookup.
+                    pass
 
         for h in new_canonical_headers:
             cls._add_block_number_to_hash_lookup(db, h)
```

To solve https://github.com/ethereum/py-evm/issues/1802 : 
```diff
diff --git a/eth/tools/logging.py b/eth/tools/logging.py
index b415a3b5..49de781a 100644
--- a/eth/tools/logging.py
+++ b/eth/tools/logging.py
@@ -10,7 +10,7 @@ class ExtendedDebugLogger(logging.Logger):
 
     @cached_property
     def show_debug2(self) -> bool:
-        return self.isEnabledFor(DEBUG2_LEVEL_NUM)
+        return False
 
     def debug2(self, message: str, *args: Any, **kwargs: Any) -> None:
         if self.show_debug2:
```

This check is now redundant, and makes you watch for a rare and unnecessary exception. Just drop it for now:
```diff
diff --git a/eth/vm/state.py b/eth/vm/state.py
index d3600cd4..12083e6e 100644
--- a/eth/vm/state.py
+++ b/eth/vm/state.py
@@ -313,7 +313,7 @@ class BaseState(Configurable, ABC):
         :param transaction: the transaction to apply
         :return: the computation
         """
-        if self.state_root != BLANK_ROOT_HASH and not self._account_db.has_root(self.state_root):
+        if False:
             raise StateRootNotFound(self.state_root)
         else:
             return self.execute_transaction(transaction)
```

p2p/peer_pool.py
```diff
@@ -471,7 +471,7 @@ def _peer_finished(self, peer: BaseService) -> None:
             if subscribers:
                 longest_queue = max(
                     self._subscribers, key=operator.attrgetter('queue_size'))
-                self.logger.info(
+                self.logger.debug(
```
pipermerriam: :+1:

trinity/sync/full/constants.py
```diff
@@ -22,4 +22,13 @@
...
+# This is also used during Beam sync (maybe we should have a different constant?)
+# The number is derived by:
+#   - number of parallel processes running (currently 4)
+#   - how many block executions can run comfortably in a single process (~2)
+#       About half the time is spent executing, and the other half waiting on nodes
+#       This might change when we start benchmarking against remote nodes
+#   - how many blocks finish early/quickly, ~half, which doubles capacity (~2)
+#   So we multiply all these together to get 16 parallel executions to permit.
+#   The first block in the queue doesn't get previewed, which brings us to 17.
+BLOCK_IMPORT_QUEUE_SIZE_TARGET = 17
```
pipermerriam: It sounds like we can derive this at runtime as something like:

```python
max(0, num_system_cpus - 1) * 2 * 2 + 1
```

trinity/sync/beam/service.py
```diff
@@ -38,7 +38,6 @@ def __init__(
             self.chaindb,
             self.peer_pool,
             self.event_bus,
-            force_beam_block_number=None,
```
carver: The default was already `None`.

trinity/sync/full/constants.py
```diff
@@ -22,4 +22,13 @@
...
-BLOCK_IMPORT_QUEUE_SIZE = 10
+# This is also used during Beam sync (maybe we should have a different constant?)
+# The number is derived by:
+#   - number of parallel processes running (currently 4)
+#   - how many block executions can run comfortably in a single process (~2)
+#       About half the time is spent executing, and the other half waiting on nodes
+#       This might change when we start benchmarking against remote nodes
+#   - how many blocks finish early/quickly, ~half, which doubles capacity (~2)
+#   So we multiply all these together to get 16 parallel executions to permit.
+#   The first block in the queue doesn't get previewed, which brings us to 17.
```
carver: I'm not sure I really believe all this as the reason, it's just what was in my head. I found the number 17 empirically on my local machine.

I don't think we can dynamically adjust the queue size, because all of the inputs seem to be fixed (including the number of preview plugins running).

carver:
Oh, also, I'm using a patched py-evm, because although I merged in https://github.com/ethereum/py-evm/pull/1810 and https://github.com/ethereum/py-evm/pull/1809 , there is another change before them in master that moves around a ton of classes.

So getting trinity working with the latest py-evm master will be critical for this PR to work in practice.

carver:
Ready for a look! I think I mostly isolated this down to a unified change now. I'm happy to strip something else out if you like.

There are some things that I'd like to do follow-up work on, but I didn't want to bloat the PR any more.

trinity/sync/beam/chain.py
```diff
@@ -453,26 +471,113 @@ def __init__(
...
         return import_done.result
 
+    async def preview_transactions(
+            self,
+            header: BlockHeader,
+            transactions: Tuple[BaseTransaction, ...],
+            parent_state_root: Hash32,
+            lagging: bool = True) -> None:
+
+        asyncio.ensure_future(self._preview_address_load(header, parent_state_root, transactions))
```
pipermerriam: Looks like code that will throw errors on shutdown.  Is there a cancellation mechanism we can use for this?

carver: I think wrapping it in `wait()` is enough.

PR #856, opened 2019-07-29
Send preview of upcoming blocks to block importer

carver:
### What was wrong?

Peeled from #855 

Send preview of upcoming blocks to block importer

### How was it fixed?

Note: the diff looks a little larger than it really is, because of a dedent that wasn't properly lined up in the diff. I pulled that into a different commit (89a2a9a859) for a smaller review of the remainder (3bdc6cd5).

- Added a mechanism to "preview" upcoming blocks to the BlockImporter class.
- Added a queue at the end of regular block sync import, to preview some number of blocks ahead of the import
- Added some more logs to help keep track of what's going on

### To-Do

- [x] Clean up commit history

- [x] ~~Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)~~

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/full/chain.py
+++ b/trinity/sync/full/chain.py
@@ -1081,42 +1081,46 @@ class RegularChainBodySyncer(BaseBodyChainSyncer):
 
         for block in unimported_blocks:
             timer = Timer()
-            _, new_canonical_blocks, old_canonical_blocks = await self.wait(
-                self._block_importer.import_block(block)
-            )
+            await self._import_block(block)
 
-            if new_canonical_blocks == (block,):
-                # simple import of a single new block.
+    async def _import_block(self, block: BaseBlock) -> None:
+        timer = Timer()
+        _, new_canonical_blocks, old_canonical_blocks = await self.wait(
+            self._block_importer.import_block(block)
+        )
 
-                # decide whether to log to info or debug, based on log rate
-                if self._import_log_limiter.can_take(1):
-                    log_fn = self.logger.info
-                    self._import_log_limiter.take_nowait(1)
-                else:
-                    log_fn = self.logger.debug
-                log_fn(
-                    "Imported block %d (%d txs) in %.2f seconds, with %s lag",
-                    block.number,
-                    len(block.transactions),
-                    timer.elapsed,
-                    humanize_seconds(time.time() - block.header.timestamp),
-                )
-            elif not new_canonical_blocks:
-                # imported block from a fork.
-                self.logger.info("Imported non-canonical block %d (%d txs) in %.2f seconds",
-                                 block.number, len(block.transactions), timer.elapsed)
-            elif old_canonical_blocks:
-                self.logger.info(
-                    "Chain Reorganization: Imported block %d (%d txs) in %.2f "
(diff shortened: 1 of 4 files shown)


carver:
Note that `preview_transactions` is a no-op just to keep this PR small. It is just an entry point that's used later in #855.

trinity/sync/full/chain.py
```diff
@@ -965,6 +963,9 @@ class RegularChainBodySyncer(BaseBodyChainSyncer):
 
     Here, the run() method will execute the sync loop forever, until our CancelToken is triggered.
     """
+    # track whether there is currently a block being imported
+    _is_importing = False
```
pipermerriam: Thoughts on using an `asyncio.Lock` for this?  You can still expose the boolean using a computed property.

trinity/sync/full/chain.py
```diff
@@ -1061,80 +1070,117 @@ def _mark_body_download_complete(
...
-            yield block_class(header, transactions, uncles)
+    def _header_to_block(self, header: BlockHeader) -> BaseBlock:
+        vm_class = self.chain.get_vm_class(header)
+        block_class = vm_class.get_block_class()
+
+        if _is_body_empty(header):
+            transactions: List[BaseTransaction] = []
+            uncles: List[BlockHeader] = []
+        else:
+            body = self._pending_bodies.pop(header)
```
pipermerriam: How do you know the body is in here?

trinity/sync/full/chain.py
```diff
@@ -1144,11 +1190,12 @@ def _headers_to_blocks(self, headers: Iterable[BlockHeader]) -> Iterable[BaseBlo
         while self.is_operational:
             await self.sleep(5)
             self.logger.debug(
-                "(in progress, queued, max size) of bodies, receipts: %r. Write capacity? %s",
+                "(progress, queued, max) of bodies, receipts: %r. Write capacity? %s Importing? %s",
                 [(q.num_in_progress(), len(q), q._maxsize) for q in (
                     self._block_body_tasks,
                 )],
                 "yes" if self._db_buffer_capacity.is_set() else "no",
+                self._is_importing,
```
pipermerriam: Maybe write a quick `humanize_bool` function to cast to `yes/no` string?

trinity/sync/full/chain.py
carver: Eh, I'm pretty neutral on it, but if you like it better, that's fine. :+1: 

```
- Some perf loss
- We're not really semantically locking a resource
+ Harder for code to change in a broken, unnoticed way, in the future
```

trinity/sync/full/chain.py
carver: I'll write a note about the required state for the method (that it must only be called on a header that has all of it's prerequisites filled, as tracked by `OrderedTaskPrep`).

trinity/sync/full/chain.py
carver: This is a super debugg-y level log. Doesn't really seem worth the effort. I was thinking about the opposite, just reverting the `_db_buffer_capacity` display to `True/False`.

PR #857, opened 2019-07-29
Serve missing beam-sync nodes in parallel

carver:
# What was wrong?

Peeled off of #855 

The previous implementation waited until the requested node data was
returned before accepting more incoming requests of the same type.

Besides being generally slower to only return one node at a time, this can be bad news when a predictive request blocks an urgent request.

### How was it fixed?

When an incoming node request is received, spawn a non-blocking async method that handles the lookup, and then immediately look for the next incoming request.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/chain.py
+++ b/trinity/sync/beam/chain.py
@@ -501,21 +501,33 @@ class MissingDataEventHandler(BaseService):
 
     async def _provide_missing_account_tries(self) -> None:
         async for event in self.wait_iter(self._event_bus.stream(CollectMissingAccount)):
-            await self._state_downloader.ensure_node_present(event.missing_node_hash)
-            await self._state_downloader.download_account(event.address_hash, event.state_root_hash)
-            await self._event_bus.broadcast(MissingAccountCollected(), event.broadcast_config())
+            asyncio.ensure_future(self._serve_account(event))
 
     async def _provide_missing_bytecode(self) -> None:
         async for event in self.wait_iter(self._event_bus.stream(CollectMissingBytecode)):
-            await self._state_downloader.ensure_node_present(event.bytecode_hash)
-            await self._event_bus.broadcast(MissingBytecodeCollected(), event.broadcast_config())
+            asyncio.ensure_future(self._serve_bytecode(event))
 
     async def _provide_missing_storage(self) -> None:
         async for event in self.wait_iter(self._event_bus.stream(CollectMissingStorage)):
-            await self._state_downloader.ensure_node_present(event.missing_node_hash)
-            await self._state_downloader.download_storage(
-                event.storage_key,
-                event.storage_root_hash,
-                event.account_address,
-            )
-            await self._event_bus.broadcast(MissingStorageCollected(), event.broadcast_config())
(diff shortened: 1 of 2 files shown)


carver:
Pretty sure the CI failure is unrelated.

carver:
The cli one also seems flaky. Don't see how it could be related.

trinity/sync/beam/chain.py
```diff
@@ -501,21 +501,33 @@ def __init__(
 
     async def _provide_missing_account_tries(self) -> None:
         async for event in self.wait_iter(self._event_bus.stream(CollectMissingAccount)):
```
pipermerriam: I think that you could replace this function with a single handle registry.

```python
self._event_bus.subscribe(CollectMissingAccount, self._serve_account)
```

I believe this is functionally equivalent and it eliminates a few extra event loop cycle.s

cburgdorf:
Downside of subscribe is that it doesn't implicitly unsubscribe when the
CancelToken is triggered.

carver:
> Downside of subscribe is that it doesn't implicitly unsubscribe when the CancelToken is triggered.

Not sure I get the ramifications of that. Does this mean it would hang on shutdown?

pipermerriam:
> Downside of subscribe is that it doesn't implicitly unsubscribe when the
CancelToken is triggered.

You could register the removal using [`BaseService.add_finished_callback`](https://github.com/ethereum/trinity/blob/2c0a42ce638e1b648d5c1837cc2cc2692b0a5797/p2p/service.py#L136) so that it gets torn down along with the service.

pipermerriam:
> Not sure I get the ramifications of that. Does this mean it would hang on shutdown?

@carver I don't think it would hang, but it would continue to try and serve those requests for the remainder of the lifecycle of the service.

pipermerriam:
Also had this thought: https://github.com/ethereum/lahja/issues/149

carver:
Two issues:

1. It looks like this means that individual incoming events are handled in serial by `subscribe()`:

https://github.com/ethereum/lahja/blob/7dd974cd2c4c1f4c54d2dc6c2f9afb480edf4d14/lahja/asyncio/endpoint.py#L411-L416

Since each `_process_item()` call waits for the previous one to finish:

https://github.com/ethereum/lahja/blob/7dd974cd2c4c1f4c54d2dc6c2f9afb480edf4d14/lahja/asyncio/endpoint.py#L344-L353

2. It looks like the code gets a line longer after adding the unsubscribe, which I think defeats the desired benefit:
```py
    async def _provide_missing_account_tries(self) -> None:
        subscription = self._event_bus.subscribe(CollectMissingAccount, self._serve_account)                                          
        await self.cancellation()
        subscription.unsubscribe()
```

Maybe it's a little prettier, but the important detail about whether this event response blocks the next event response (from 1 above) gets pushed down inside the lahja black box.

pipermerriam:
> It looks like this means that individual incoming events are handled in serial

That sounds right.  Maybe use of `EndpointAPI.subscribe` isn't the right thing.

carver:
:heavy_check_mark: from you then @pipermerriam ?

PR #756, opened 2019-06-28
New database manager

ChihChengLiang:
### What was wrong?

Currently, Trinity client creates an extra process for each database consumer, this cause overheads and performance degrade.

### How was it fixed?

Write a DB manager that handles all the communication to the database and a DB client that talks to the manager. 

Benchmark shows the new DB manager-client has 1.5x to 2x performance comparing to the existing one.

```
$ python scripts/benchmark/db_managers/bench_db_client.py 
Client 1: 6412.4640859929395 get-set per second
Client 0: 6294.6387453949055 get-set per second
Client 2: 5925.177664645094 get-set per second
Client 1: 6466.534301796517 get-set per second
Client 0: 6419.005003848052 get-set per second
Client 2: 6012.5870262597755 get-set per second
Client 1: 6271.011569056251 get-set per second
Client 0: 6263.727016479054 get-set per second
Client 2: 6715.727595295333 get-set per second
$ python scripts/benchmark/db_managers/bench_old_manager.py 
Client 0: 4235.169438244839 get-set per second
Client 1: 4207.37306063274 get-set per second
Client 2: 4158.8369075974715 get-set per second
Client 0: 3686.812208983483 get-set per second
Client 1: 3679.808669849572 get-set per second
Client 2: 3590.928601338237 get-set per second
Client 0: 4002.483401664311 get-set per second
Client 1: 4020.7755111135643 get-set per second
Client 2: 4116.866458079233 get-set per second
```

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #859, opened 2019-07-30
New native database adapter

pipermerriam:
### What was wrong?

Currently, we've been using the `multiprocessing` facilities to access the database and chain from multiple processes.

### How was it fixed?

This implements a custom server that serves the `AtomicDatabaseAPI` over an IPC socket and a client that implements the full `AtomicDatabaseAPI`.

This requires converting all of the various `Chain`, `ChaindDB` and `HeaderDB` operations to now run in the local process using a thread-based executor.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


pipermerriam:
Replaces #756

pipermerriam:
Problems that arose while working on this.

Pretty much everywhere in Trinity where we interact with the chain we use the async wrapper/proxy objects for things like `Chain`, `ChainDB`, `HeaderDB`, and `BaseDB`.  These objects *pretend* like they are the actual object but when you call a method on them what actually occurs is that the execution happens within the database process.

This is how Trinity keeps the networking code from being blocked by expensive CPU-bound operations related to the EVM or Chain.

If we want to replace the way we interact with the database, we will also be required to replace this mechanism.  I.E. we will still need some semblance of a proxy object that represents the actual `Chain/HeaderDB/etc` which actually lives in a separate process and which delegates calls to the object methods in a manner that doesn't block `async` execution.

With the direction that we are currently moving, this probably means architecture that looks something like:

- A service that runs in it's own process with a `lahja` endpoint which listens to and responds to requests.
- A proxy object that delegates calls over the `lahja` event bus to execute specific methods or properties and uses `asyncio.run_in_executor` or `trio.run_sync_in_thread` on the local side to avoid blocking the `async` execution.

Doing this today will involve a lot of boilerplate since we'd need to write events and handlers for all of the methods.  @cburgdorf recently did some similar things for the `ProxyPeerPool`.

What I think we need for this to be less heavy on the boilerplate code and thus on the complexity of working with it is some sort of machinery to automate the concept of serving an object's methods over an event bus.  Something like a metaclass that generates all of the events under the hood as well as the handlers for serving the request/response for those events.

pipermerriam:
@carver I ended up running with this to try an easier way to get it working.  I did a minimal change such that all of the `Chain/ChainDB/HeaderDB` coroutine methods now run in a `ThreadPoolExecutor` which should roughly preserve the previous property of not having Chain operations block the `async` processing, though this is slightly different from the previous model in that we will see *some* performance changes since the database process will be doing less CPU work (maybe good?) and the other processes will be doing more CPU work via threads which is likely to have *some* measurable effect on their performance.

I'm wondering if you can profile beam sync against this branch to see if we actually gain or lose any measurable performance using this mechanism

carver:
> I'm wondering if you can profile beam sync against this branch to see if we actually gain or lose any measurable performance using this mechanism

Hm, maybe a small lift, but nothing obvious. At least, it doesn't seem to hurt! :)

At this point, it can be hard to tell because Beam Sync is usually staying caught up. I have a few profiling improvements I've been thinking about, which might help me give you a definitive answer later today.

pipermerriam:
Cool, I'll work on cleaning this up with intent to merge after some review next week.

pipermerriam:
Test failures are unrelated to this PR

pipermerriam:
local profiling shows these numbers:

```
old-db-manager: ~14,000 get+set ops/sec
new-db-manager: ~20,000 get+set ops/sec
```

Better profiling would be to do something like running a `Chain.import_block` to get a more real usage profile but I don't think we have facilities in place to do that easily.  Either way, I feel good about the direction this takes us.

tests-trio/p2p-trio/test_channel_services.py
```diff
@@ -1,6 +1,7 @@
 import trio
 
 import pytest
+import pytest_trio
```
pipermerriam: extract this re-org to its own PR

trinity/chains/base.py
```diff
@@ -17,36 +17,36 @@ class BaseAsyncChainAPI(ABC):
...
+        ...
 
     @abstractmethod
     async def coro_validate_chain(
             self,
             parent: BlockHeader,
             chain: Tuple[BlockHeader, ...],
             seal_check_random_sample_rate: int = 1) -> None:
-        pass
+        ...
```
pipermerriam: TODO: extract to stand-alone cleanup PR

trinity/db/eth1/header.py
```diff
@@ -1,95 +1,91 @@
...
 
-from trinity._utils.mp import (
-    async_method,
-)
+
+TReturn = TypeVar('TReturn')
 
 
-class BaseAsyncHeaderDB(ABC):
+def async_thread_method(method: Callable[..., Any]) -> Callable[..., Any]:
```
pipermerriam: This probably belongs somewhere under `trinity._utils`

trinity/main.py
```diff
@@ -97,6 +95,15 @@ def run_database_process(trinity_config: TrinityConfig, db_class: Type[BaseDB])
...
+            initialize_database(chain_config, chaindb, base_db)
 
-        manager = create_db_server_manager(trinity_config, base_db)
-        serve_until_sigint(manager)
+        manager = DBManager(base_db)
+        with manager.run(trinity_config.database_ipc_path):
+            try:
+                manager.wait_stopped()
+            except KeyboardInterrupt:
+                pass
```
pipermerriam: TODO: I don't think this is actually running it's cleanup step because the IPC path is consistently left behind.

trinity/protocol/common/peer.py
```diff
@@ -149,7 +149,7 @@ def __init__(self,
         super().__init__(token)
 
     def __str__(self) -> str:
-        return f"{self.__class__.__name__} {self.remote.uri}"
+        return f"{self.__class__.__name__} {self.remote}"
```
pipermerriam: bugfix can be extracted to standalone

trinity/server.py
```diff
@@ -295,87 +293,3 @@ def _make_receive_server(self) -> BCCReceiveServer:
             peer_pool=self.peer_pool,
             token=self.cancel_token,
         )
-
-
-def _test() -> None:
```
pipermerriam: removal of these tests can be extracted to standalone

trinity/sync/full/hexary_trie.py
```diff
@@ -233,7 +234,8 @@ def _schedule(self, node_key: Hash32, parent: SyncRequest, depth: int,
         called.
         """
         self.committed_nodes += 1
-        await self.db.coro_set(request.node_key, request.data)
+        loop = asyncio.get_event_loop()
+        await loop.run_in_executor(None, self.db.set, request.node_key, request.data)
```
pipermerriam: AFAIK this is the only place that used the base async db methods so I just updated it to run the sync versions asynchronously using a thread.

scripts/benchmark/db_managers/bench_db_client.py
```diff
@@ -0,0 +1,73 @@
...
+
+if __name__ == '__main__':
+    with tempfile.TemporaryDirectory() as ipc_base_dir:
+        ipc_path = pathlib.Path(ipc_base_dir) / 'db.ipc'
+
+        server = multiprocessing.Process(target=run_server, args=[ipc_path])
+
+        clients = [
+            multiprocessing.Process(target=run_client, args=[ipc_path, i])
+            for i in range(1)
```
pipermerriam: This benchmark needs to be updated to have a CLI that allows specifying these values like how many clients, how many keys/values and it needs to get added to the CI run.

tests/core/database/test_db_client.py
```diff
@@ -0,0 +1,132 @@
...
+        db_client.get(b'does-not-exists')
+    with pytest.raises(KeyError):
+        db_client[b'does-not-exists']
+
+
+def test_sync_db_client_set_and_get(db_client):
+    db_client.set(b'key-1', b'value-1')
+    db_client.set(b'key-2', b'value-2')
+    assert db_client.get(b'key-1') == b'value-1'
+    assert db_client.get(b'key-2') == b'value-2'
```
pipermerriam: This test suite is only *just* adequate.  I think that I want to pack up the test suite from Py-EVM into something re-usable so that we can run it against this implementation as well.

trinity/db/manager.py
```diff
@@ -0,0 +1,439 @@
...
+
+class Operation(enum.Enum):
+    GET = b'\x00'
+    SET = b'\x01'
+    DELETE = b'\x02'
+    EXISTS = b'\x03'
+    ATOMIC_BATCH = b'\x04'
+
+
+GET = Operation.GET
```
pipermerriam: Need to document the structure of each message payload here somewhere so that it's well defined.

trinity/db/manager.py
```diff
@@ -0,0 +1,439 @@
...
+            # background task to close the socket.
+            threading.Thread(
+                name="_close_socket_on_stop",
+                target=self._close_socket_on_stop,
+                args=(sock,),
+                daemon=False,
+            ).start()
+
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            sock.bind(str(ipc_path))
```
pipermerriam: This section deserves some more comments.

trinity/db/manager.py
```diff
@@ -0,0 +1,439 @@
...
+                self.logger.debug('Server accepted connection from %r', addr)
+                threading.Thread(
+                    name="_serve_conn",
+                    target=self._serve_conn,
+                    args=(conn,),
+                    daemon=False,
+                ).start()
+
+        if ipc_path.exists():
+            ipc_path.unlink()
```
pipermerriam: The location of this makes me think it's going to be un-reliable.  TODO: look for where else this can be placed that would give it higher guarantees of proper cleanup.

trinity/db/manager.py
```diff
@@ -0,0 +1,439 @@
...
+
+        """
+        kv_pair_and_delete_count_data = sock.read_exactly(DOUBLE_LEN_BYTES)
+        kv_pair_count, delete_count = struct.unpack('<II', kv_pair_and_delete_count_data)
+        total_kv_count = 2 * kv_pair_count
+        if kv_pair_count or delete_count:
+            kv_and_delete_sizes_data = sock.read_exactly(
+                DOUBLE_LEN_BYTES * kv_pair_count + LEN_BYTES * delete_count
+            )
+            fmt_str = '<' + 'I' * (total_kv_count + delete_count)
```
pipermerriam: This section deserves more comments.

trinity/db/manager.py
```diff
@@ -0,0 +1,439 @@
...
+            with self.db.atomic_batch() as batch:
+                for key_size, value_size in partition(2, kv_sizes):
+                    combined_size = key_size + value_size
+                    key_and_value_data = sock.read_exactly(combined_size)
+                    key = key_and_value_data[:key_size]
+                    value = key_and_value_data[key_size:]
+                    batch[key] = value
+                for key_size in delete_sizes:
+                    key = sock.read_exactly(key_size)
+                    del batch[key]
```
pipermerriam: This *could* raise a `KeyError` were the key to be deleted by another client with the right timing.  Not sure what the right behavior should be in that case.  This sort of brings up a deeper issue which is what the expected behavior is if we have multiple clients writing to the database....

trinity/db/manager.py
```diff
@@ -0,0 +1,439 @@
...
+        with self._lock:
+            self._socket.sendall(EXISTS.value + len(key).to_bytes(4, 'little') + key)
+            result_byte = self._socket.read_exactly(1)
+
+        if result_byte == SUCCESS_BYTE:
+            return True
+        elif result_byte == FAIL_BYTE:
+            return False
+        else:
+            raise Exception("Unknown result byte: {result_byte.hex}")
```
pipermerriam: missing `f` for f-string.

trinity/db/manager.py
```diff
@@ -0,0 +1,439 @@
...
+        with self._lock:
+            self._socket.sendall(DELETE.value + len(key).to_bytes(4, 'little') + key)
+            result_byte = self._socket.read_exactly(1)
+
+        if result_byte == SUCCESS_BYTE:
+            return
+        elif result_byte == FAIL_BYTE:
+            raise KeyError(key)
+        else:
+            raise Exception("Unknown result byte: {result_byte.hex}")
```
pipermerriam: missing `f` for f-string.

trinity/db/manager.py
```diff
@@ -0,0 +1,439 @@
...
+            result_byte = self._socket.read_exactly(1)
+
+            if result_byte == SUCCESS_BYTE:
+                value_size_data = self._socket.read_exactly(LEN_BYTES)
+                value = self._socket.read_exactly(int.from_bytes(value_size_data, 'little'))
+                return value
+            elif result_byte == FAIL_BYTE:
+                raise KeyError(key)
+            else:
+                raise Exception("Unknown result byte: {result_byte.hex}")
```
pipermerriam: missing `f` for f-string.

trinity/db/manager.py
```diff
@@ -0,0 +1,439 @@
...
+        fmt_str = '<II' + 'I' * (len(kv_sizes) + len(pending_deletes))
+        kv_pair_count_and_size_data = struct.pack(
+            fmt_str,
+            kv_pair_count,
+            delete_count,
+            *kv_sizes,
+            *delete_sizes,
+        )
+        kv_data = b''.join(itertools.chain(*pending_kv_pairs))
+        delete_data = b''.join(pending_deletes)
```
pipermerriam: can combine this with the statement above inside the `chain` call

trinity/db/manager.py
```diff
@@ -0,0 +1,439 @@
...
+        pending_deletes = diff.deleted_keys()
+        pending_kv_pairs = diff.pending_items()
+
+        kv_pair_count = len(pending_kv_pairs)
+        delete_count = len(pending_deletes)
+
+        kv_sizes = tuple(len(item) for item in itertools.chain(*pending_kv_pairs))
+        delete_sizes = tuple(len(key) for key in pending_deletes)
+
+        fmt_str = '<II' + 'I' * (len(kv_sizes) + len(pending_deletes))
```
pipermerriam: This section deserves comments.

trinity/db/manager.py
```diff
@@ -0,0 +1,439 @@
...
+        cls.logger.debug("Opened connection to %s: %s", path, s)
+        s.connect(str(path))
+        return cls(s)
+
+
+class AtomicBatch(BaseDB):
+    """
+    This is returned by a BaseAtomicDB during an atomic_batch, to provide a temporary view
+    of the database, before commit.
+    """
```
pipermerriam: Lots of copy/paste in this class.

trinity/db/manager.py
```diff
@@ -0,0 +1,439 @@
...
+    logger = logging.getLogger("trinity.db.manager.AtomicBatch")
+
+    _write_target_db: BaseDB = None
+    _diff: DBDiffTracker = None
```
pipermerriam: The `DBDiffTracker` isn't necessarily public API in `py-evm` but it probably should be and thus should get an interface defined for it as well as ensuring it is being tested against the standard suite of tests for `DatabaseAPI`

trinity/db/manager.py
```diff
@@ -0,0 +1,511 @@
...
+                    del batch[key]
+
+        sock.sendall(SUCCESS_BYTE)
+
+
+class empty:
+    pass
+
+
+class DBClient(BaseAtomicDB):
```
lithp: Dropping multiprocessing is already a nice improvement and this gives me some ideas for future PRS:

I expect that we'll see an even larger performance improvement if these methods were to become asynchronous! Because these methods are synchronous every call to the database blocks the event loop, if the database ever gets backed up then the other processes will back up along with it when they could be talking over the network or otherwise let their other coros truck along.

And this is a far-future idea, but since so much of our database is immutable data I wonder whether some kind of caching on this side of the process could also improve performance.

trinity/db/manager.py
```diff
@@ -0,0 +1,511 @@
...
+    _write_target_db: BaseDB = None
+    _diff: DBDiffTracker = None
+
+    def __init__(self, db: BaseDB) -> None:
+        self._db = db
+        self._track_diff = DBDiffTracker()
+
+    def __getitem__(self, key: bytes) -> bytes:
+        if self._track_diff is None:
+            raise ValidationError("Cannot get data from a write batch, out of context")
```
lithp: These error messages assume that `AtomicBatch` will only ever be used by `DBClient`, it might be better for `AtomicBatch` to have a name which better indicates that it's private to this file, or for `AtomicBatch` to do it's own context management, or for these messages to say something like "cannot use AtomicBatch after it has been finalized".

trinity/db/manager.py
```diff
@@ -0,0 +1,511 @@
...
+
+            while self.is_running:
+                try:
+                    conn, addr = sock.accept()
+                except (ConnectionAbortedError, OSError) as err:
+                    self.logger.debug("Server stopping: %s", err)
+                    self._stopped.set()
+                    break
+                self.logger.debug('Server accepted connection: %r', addr)
+                threading.Thread(
```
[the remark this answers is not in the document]
pipermerriam: We did both and the synchronous outperformed the `asyncio` implementation by roughly 2-3x.  We didn't do any real measuring of *why* but I suspect event loop overhead is just way more significant than the gains we get from being able to do work while the `i/o` is occuring.

trinity/db/manager.py
```diff
@@ -0,0 +1,511 @@
...
+            # These options help fix an issue with the socket reporting itself
+            # already being used since it accepts many client connection.
+            # https://stackoverflow.com/questions/6380057/python-binding-socket-address-already-in-use
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            sock.bind(str(ipc_path))
+            sock.listen(1)
+
+            # Wait for the IPC socket to appear before marking ourself as
+            # actually having started.
+            wait_for_ipc(ipc_path)
```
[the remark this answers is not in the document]
pipermerriam: I'll try removing it and see if we can get by without it.

trinity/db/manager.py
```diff
@@ -0,0 +1,511 @@
...
+                    self.logger.debug("%s: closing client connection: %s", self, raw_socket)
+                    break
+                except Exception:
+                    self.logger.exception("Error reading operation flag")
+                    break
+
+                try:
+                    operation = Operation(operation_byte)
+                except TypeError:
+                    self.logger.error("Unrecognized database operation: %s", operation_byte.hex())
```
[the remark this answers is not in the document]
pipermerriam: I don't actually think we get any useful information from that which would let us know who is misbehaving...

trinity/db/manager.py
```diff
@@ -0,0 +1,511 @@
...
+    @classmethod
+    def connect(cls, path: pathlib.Path) -> "DBClient":
+        wait_for_ipc(path)
+        s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        cls.logger.debug("Opened connection to %s: %s", path, s)
+        s.connect(str(path))
+        return cls(s)
+
+
+class AtomicBatch(BaseDB):
```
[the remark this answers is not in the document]
pipermerriam: I'll open an issue.

trinity/sync/full/hexary_trie.py
```diff
@@ -163,7 +163,8 @@ def next_batch(self, n: int = 1) -> List[SyncRequest]:
         if node_key in self.nodes_cache:
             self.logger.debug2("Node %s already exists in db", encode_hex(node_key))
             return
-        if await self.db.coro_exists(node_key):
+        loop = asyncio.get_event_loop()
+        if await loop.run_in_executor(None, self.db.exists, node_key):
```
lithp: We're already blocking the loop everywhere we call a method on this database so it probably isn't a big problem to keep the code clean and call `self.db.exists` directly here, and same with `self.db.set` below.

trinity/db/manager.py
pipermerriam: I think that any caching would likely need to happen at a higher level but yes, there are future improvements that could be made and now that we fully own this API we have a lot of freedom to explore.

trinity/sync/full/hexary_trie.py
pipermerriam: Correct, however I don't actually forsee this being an issue because all of our database operations occur behind `HeaderDB` or `ChainDB` and for the most part we use `async_thread_method` to mitigate the blocking nature of these calls.

IIRC, making these calls asynchronous was an intentional choice back when this code was written and it produced a performance gain.  Also, I *think* that this code is going to go away with beam/firehose so I'm not inclined to do much more than maintain status quo.

trinity/db/manager.py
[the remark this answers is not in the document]
pipermerriam: removed and everything seems to work fine.

trinity/db/manager.py
```diff
@@ -0,0 +1,511 @@
...
+                    break
+
+                try:
+                    operation = Operation(operation_byte)
+                except TypeError:
+                    self.logger.error("Unrecognized database operation: %s", operation_byte.hex())
+                    break
+
+                try:
+                    if operation is GET:
```
lithp: I'm not fully on the if-less bandwagon but wanted to drop a comment here, I bet this could be turned into an: `operation.perform(self.db, sock)`. And a world where the different operations were different subclasses of `Operation` is also a world where the method to send an `ATOMIC_BATCH` and the method to receive an `ATOMIC_BATCH` sit right next to each other, which might make future changes easier.

trinity/db/manager.py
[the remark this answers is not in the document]
lithp: You're right, not without probably too much extra work.

trinity/db/manager.py
pipermerriam: @ChihChengLiang originally wrote them that way but I felt they were harder to understand, though that could have just been the code organization or something.  Going to leave that idea for a future refactor.  Also noting that I tried to minimize any overhead since this is somewhat performance sensitive code.

PR #860, opened 2019-07-30
fix loggers not having a level set and always logging debug2

pipermerriam:
### What was wrong?

Setting the root logger to level `0` causes all loggers without an explicit level to have their level be `NOTSET` which results in `debug2` logs always being shown.

### How was it fixed?

Trying just removing the line.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


pipermerriam:
@carver I think test failures here are unrelated and this should fix your `DEBUG2` log message issue.

PR #862, opened 2019-07-31
Add BLS fixtures tests

hwwhww:
### What was wrong?

Add BLS fixtures tests

### How was it fixed?
- Update Trinity-defined test fixtures to `v0.8.1b2`: https://github.com/hwwhww/eth2.0-spec-tests/releases/tag/v0.8.1b2
- Add `tests/eth2/fixtures/bls-fixtures/test_bls.py`

Note that PyECC `msg_hash_compressed` and `msg_hash_uncompressed` APIs are not exposed, skipping these tests now.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

62194279-22235b00-b3ac-11e9-82fa-ff1121fd03af

### DIFF
--- a/tests/eth2/fixtures/bls-fixtures/test_bls.py
+++ b/tests/eth2/fixtures/bls-fixtures/test_bls.py
@@ -1,0 +1,138 @@
+from typing import (
+    Tuple,
+)
+from dataclasses import (
+    dataclass,
+)
+import pytest
+
+from py_ecc.bls.typing import Domain
+
+from eth2._utils.bls import bls
+from eth2._utils.bls.backends import (
+    PyECCBackend,
+)
+from eth2.beacon.tools.fixtures.loading import (
+    get_input_bls_privkey,
+    get_input_bls_pubkeys,
+    get_input_bls_signatures,
+    get_input_sign_message,
+    get_output_bls_pubkey,
+    get_output_bls_signature,
+)
+from eth2.beacon.tools.fixtures.test_case import (
+    BaseTestCase,
+)
+from eth_typing import (
+    BLSPubkey,
+    BLSSignature,
+    Hash32,
+)
+
+
+from tests.eth2.fixtures.helpers import (
+    get_test_cases,
+)
+from tests.eth2.fixtures.path import (
+    BASE_FIXTURE_PATH,
+    ROOT_PROJECT_DIR,
+)
+
+
+# Test files
+RUNNER_FIXTURE_PATH = BASE_FIXTURE_PATH / 'bls'
+HANDLER_FIXTURE_PATHES = (
+    RUNNER_FIXTURE_PATH / 'aggregate_pubkeys',
+    RUNNER_FIXTURE_PATH / 'aggregate_sigs',
+    # RUNNER_FIXTURE_PATH / 'msg_hash_g2_compressed',  # NOTE: No public API in PyEECBackend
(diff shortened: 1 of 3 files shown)


tests/eth2/fixtures/bls-fixtures/test_bls.py
```diff
@@ -0,0 +1,138 @@
...
+
+@dataclass
+class BLSSignaturesAggregationTestCase(BaseTestCase):
+    input: Tuple[BLSPubkey, ...]
+    output: BLSSignature
+
+
+@dataclass
+class BLSPrivToPubTestCase(BaseTestCase):
+    input: bytes
```
ChihChengLiang: should be `int` type?

tests/eth2/fixtures/bls-fixtures/test_bls.py
```diff
@@ -0,0 +1,138 @@
...
+)
+
+
+# Test files
+RUNNER_FIXTURE_PATH = BASE_FIXTURE_PATH / 'bls'
+HANDLER_FIXTURE_PATHES = (
+    RUNNER_FIXTURE_PATH / 'aggregate_pubkeys',
+    RUNNER_FIXTURE_PATH / 'aggregate_sigs',
+    # RUNNER_FIXTURE_PATH / 'msg_hash_g2_compressed',  # NOTE: No public API in PyEECBackend
+    # RUNNER_FIXTURE_PATH / 'msg_hash_g2_uncompressed',  # NOTE: No public API in PyEECBackend
```
ralexstokes: we can still move ahead and just import the function right? having conformance on our `hash_to_g2` impl is important :)

tests/eth2/fixtures/bls-fixtures/test_bls.py
hwwhww:
```suggestion
    input: int
```

tests/eth2/fixtures/bls-fixtures/test_bls.py
hwwhww: The `hash_to_g2` functions is implemented in `py_ecc`.

In eth2.0-specs, `test_generator` implemented the functions for test vectors:
```python
def hash_message_compressed(msg: bytes, domain: bytes) -> Tuple[str, str]:
    """
    Hash message
    Input:
        - Message as bytes32
        - domain as bytes8
    Output:
        - Message hash as a compressed G2 point
    """
    z1, z2 = bls.utils.compress_G2(bls.utils.hash_to_G2(msg, domain))
    return [int_to_hex(z1, G2_COMPRESSED_Z_LEN), int_to_hex(z2, G2_COMPRESSED_Z_LEN)]


def hash_message(msg: bytes,
                 domain: bytes) ->Tuple[Tuple[str, str], Tuple[str, str], Tuple[str, str]]:
    """
    Hash message
    Input:
        - Message as bytes32
        - domain as bytes8
    Output:
        - Message hash as a G2 point
    """
    return [
        [
            int_to_hex(fq2.coeffs[0], F2Q_COEFF_LEN),
            int_to_hex(fq2.coeffs[1], F2Q_COEFF_LEN),
        ]
        for fq2 in bls.utils.hash_to_G2(msg, domain)
    ]
```

But in Trinity, `hash_to_G2` is only called in the `sign`, `verify`, `verify_multiple` APIs; so to test these `msg_hash_g2_compressed` and `msg_hash_g2_uncompressed`test vectors, we will end up just re-implement the *same* `hash_message_compressed` and `hash_message` functions as test generator with the same `py_ecc` module. It seems useless before BLS standardization is done. 😅

PR #863, opened 2019-07-31
Implement packet encoding and decoding services

jannikluhn:
One service that takes incoming datagrams and tries to decode them as packets and another one that does the opposite.

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/channel_services.py
+++ b/p2p/discv5/channel_services.py
@@ -11,16 +11,28 @@ from trio.abc import (
     SendChannel,
 )
 
+from eth_utils import (
+    encode_hex,
+    ValidationError,
+)
+
 from p2p.trio_service import (
     as_service,
     ManagerAPI,
 )
 
+from p2p.discv5.packets import (
+    decode_packet,
+    Packet,
+)
 from p2p.discv5.constants import (
     DATAGRAM_BUFFER_SIZE,
 )
 
 
+#
+# Data structures
+#
 class Endpoint(NamedTuple):
     ip_address: bytes
     port: int
@@ -36,6 +48,19 @@ class OutgoingDatagram(NamedTuple):
     receiver: Endpoint
 
 
+class IncomingPacket(NamedTuple):
+    packet: Packet
+    sender: Endpoint
+
+
+class OutgoingPacket(NamedTuple):
+    packet: Packet
+    receiver: Endpoint
+
+
+#
+# UDP
+#
 @as_service
 async def DatagramReceiver(manager: ManagerAPI,
                            socket: SocketType,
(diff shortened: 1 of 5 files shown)


jannikluhn:
The failing test seems to be unrelated.

p2p/discv5/packets.py
```diff
@@ -335,6 +338,23 @@ def validate_auth_header(auth_header: AuthHeader) -> None:
...
 # Packet decoding
 #
+def decode_packet(encoded_packet: bytes) -> Packet:
+    try:
+        return decode_message_packet(encoded_packet)
+    except ValidationError as error:
+        message_error = error
+
+    try:
+        return decode_who_are_you_packet(encoded_packet)
```
pipermerriam: I'm inlined to push back *just a little* on this approach.  It's probably fine, though I'd have an easier time with a more precise mechanism than iteration over all of the packet types.  It would be nice if we could do a pre-filter that checks general structure against the different packet structures and doesn't try to do a full decode on any packet that doesn't match the general expected *shape* such as number of elements.

If this isn't possible or reasonable to do at this stage (or just gets in your way) feel free to disregard.  If you think we can achieve this at some later stage maybe open up a tracking issue so that we can keep track of the potential improvement to do at some later datae.

p2p/discv5/packets.py
```diff
@@ -335,6 +338,23 @@ def validate_auth_header(auth_header: AuthHeader) -> None:
...
+    except ValidationError as error:
+        who_are_you_error = error
+
+    raise ValidationError(
+        f"Failed to decrypt packet, both as message ({message_error}) and as WhoAreYou packet "
```
pipermerriam: *nit*.  Might be nicer to format this as:

```
Failed to decrypt packet:
  - message: {message_error}
  - who_are_you: {who_are_you_error}
```

tests/p2p-trio/test_channel_services.py
```diff
@@ -29,6 +38,11 @@
     return sending_socket, receiving_socket
 
 
+@pytest.fixture
+def fake_endpoint():
+    return Endpoint(ip_address=b"127.0.0.1", port=1234)
```
pipermerriam: Probably worth making a factory for this class too.

p2p/discv5/channel_services.py
```diff
@@ -64,3 +89,46 @@ class OutgoingDatagram(NamedTuple):
...
+    async with incoming_datagram_receive_channel, incoming_packet_send_channel:
+        async for datagram, sender in incoming_datagram_receive_channel:
+            try:
+                packet = decode_packet(datagram)
+                logger.debug(
+                    f"Successfully decoded {packet.__class__.__name__} with tag "
+                    f"{encode_hex(packet.tag)} from {sender}"
+                )
+            except ValidationError:
+                logger.warn(f"Failed to decode a packet from {sender}", exc_info=True)
```
pipermerriam: I could see this being a problem at some point since it introduces a griefing vector by just sending us garbage discovery packets.  Not a big deal for now but wanted to point it out now.

p2p/discv5/channel_services.py
```diff
@@ -64,3 +89,46 @@ class OutgoingDatagram(NamedTuple):
...
+                        incoming_packet_send_channel: SendChannel[IncomingPacket],
+                        ) -> None:
+    """Decodes incoming datagrams to packet objects."""
+    logger = logging.getLogger('p2p.discv5.channel_services.PacketDecoder')
+
+    async with incoming_datagram_receive_channel, incoming_packet_send_channel:
+        async for datagram, sender in incoming_datagram_receive_channel:
+            try:
+                packet = decode_packet(datagram)
+                logger.debug(
```
pipermerriam: I suspect this will need to be DEBUG2 level eventually as this strikes me as being very noisy.

p2p/discv5/channel_services.py
```diff
@@ -64,3 +89,46 @@ class OutgoingDatagram(NamedTuple):
...
+#
+# Packet encoding/decoding
+#
+@as_service
+async def PacketDecoder(manager: ManagerAPI,
+                        incoming_datagram_receive_channel: ReceiveChannel[IncomingDatagram],
+                        incoming_packet_send_channel: SendChannel[IncomingPacket],
+                        ) -> None:
+    """Decodes incoming datagrams to packet objects."""
+    logger = logging.getLogger('p2p.discv5.channel_services.PacketDecoder')
```
pipermerriam: might be nice if we could reliably use `manage.logger` to get at a properly named/namespaced logger for the service.

p2p/discv5/packets.py
jannikluhn: I see your point and went with checking if the RLP part of the packet ranges until the end of the data (who are you packet) or if it is followed by additional data (the encrypted message, for message packets). The only drawback I see is that I had to import a somewhat internal method from pyrlp (`consume_length_prefix`) but I think that it's the best, "minimally invasive" solution.

p2p/discv5/channel_services.py
jannikluhn: *Relatively* simple way to deal with this would be to blacklist (or rate limit?) IPs that send too many packets per second. But I agree, nothing we should worry about for now.

p2p/discv5/channel_services.py
jannikluhn: My thought process was that we can filter both vertically (by level) and horizontally (by component) and if we've filtered appropriately horizontally (i.e. we only look at this particular service), `DEBUG` is the right level for this kind of message. It would only become too noisy if we'd look at many services at the same time with `DEBUG` which I hope one wouldn't ever have to do. But I'm just guessing and don't really know how people filter logs in practice, so I could be off here.

jannikluhn:
There are failing tests in CI, but they are unrelated to this PR. I've tried rerunning them a couple of times, but that didn't help :(

PR #864, opened 2019-07-31
Implement Tag Utils

jannikluhn:
Each Discv5 packet has a tag that XORs the receiver and the sender node ids together and allows the receiver to recover the sender id. This PR implements the necessary functions to work with those tags.

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/tags.py
+++ b/p2p/discv5/tags.py
@@ -1,0 +1,23 @@
+import hashlib
+
+from eth_typing import (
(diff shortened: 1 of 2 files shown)


jannikluhn:
The failing test is p2p related, but as far as I can see has nothing to do with discovery.

jannikluhn:
Added docstrings and some manually computed test vectors since there isn't anything official yet.

PR #866, opened 2019-07-31
Add black and isort to eth2 code

ralexstokes:
### What was wrong?

Lack of a code style tool.

### How was it fixed?

Introduces `black` and `isort` as a code formatter to the `eth2` code in this repo.

#### Cute Animal Picture

### DIFF
--- a/eth2/_utils/bitfield.py
+++ b/eth2/_utils/bitfield.py
@@ -1,6 +1,4 @@
-from cytoolz import (
-    curry,
-)
+from cytoolz import curry
 from eth2.beacon.typing import Bitfield
 from .tuple import update_tuple_item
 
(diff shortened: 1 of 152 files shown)


eth2/_utils/bitfield.py
```diff
@@ -1,6 +1,4 @@
-from cytoolz import (
-    curry,
-)
+from cytoolz import curry
```
hwwhww: Per https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#import-statements , any chance to allow it?

ralexstokes: which part exactly? i think we do want to add in use of `isort` but the default configuration b/t `isort` and `black` conflicted... in general, the `black` philosophy is that you can't configure it

hwwhww: Sorry, the accurate link is here, formatting part: https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#formatting

> Multiple imports must be split across multiple lines, one per line.

pipermerriam: Yeah, that's one that should be relaxed/removed.  Too strict and better to just let isort/black deal with it.

ralexstokes: i'm in favor of "let isort/black deal with it"

ralexstokes:
i can go back and add `isort` in this PR as the combo is what we want and i had left it out of this PR because I was confused about how they interacted.

ralexstokes:
Closed in favor of #974.

PR #868, opened 2019-08-02
Implement handshake initiator and recipient

jannikluhn:
This adds two trio services that perform a handshake either as initiator or recipient. The idea is that there's some kind of peer class that launches one of them if necessary (i.e. if they receive a message they can't decrypt or they want to send a message to a new peer) and then hands over all responsibility to it. Once the handshake is complete the service exits and the peer class can take over again.

I've tried different architectures and this one seems the cleanest to me, mostly because initiator and recipient logic is separated into different classes, but at the same time fully contained in them. This makes it relatively easy to follow the protocol step by step without having to jump around to much. The disadvantage is that we also have to deal with packets that are not related to the handshake but are received at the same time. Fortunately, according to the spec we can just drop them as peers are expected to resend them once the handshake is complete, which makes the handling quite easy.

I've kept the tests relatively high-level: Just connect an initiator to a recipient and see that it works properly. This is obviously not enough because we're missing all the conditions under which a handshake is supposed to fail. But for now I'd rather not write more detailed tests as it's a pretty open ended task and we (hopefully) will get a standardized official test suite at some point anyway.

It's not ready to merge or for a detailed review yet: I've written it mostly blindly as it builds on top of #825, #864 and also some minor refactoring I'd like to do before (mostly adding some more detailed type hints for stuff like node ids, id nonces, etc.). So this doesn't actually work right now, but the general structure is pretty much finished I would say. But I'm also fine to wait for a review until all lower level work is done. Just wanted to open a PR to get it out.

Unfortunately, the PR is pretty big. The only way I can see to make it split it up into two (one for initiator and one for recipient). But I'm not sure if that actually makes reviewing easier?

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/constants.py
+++ b/p2p/discv5/constants.py
@@ -8,6 +8,9 @@ NONCE_SIZE = 12  # size of an AESGCM nonce
 TAG_SIZE = 32  # size of the tag packet prefix
 MAGIC_SIZE = 32  # size of the magic hash in the who are you packet
 
+RANDOM_ENCRYPTED_DATA_SIZE = 12  # size of the random data we put in packets before handshake
+ID_NONCE_SIZE = 12  # size of the id nonce we generate for the peer during handshake
+
 MAX_PACKET_SIZE = 1280  # maximum allowed size of a packet
 
 ZERO_NONCE = Nonce(b"\x00" * NONCE_SIZE)  # nonce used for the auth header packet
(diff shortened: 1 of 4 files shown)


jannikluhn:
Closing as this approach doesn't work since the latest discv5 spec change.

PR #869, opened 2019-08-02
Implement handshaker api for devp2p protocol

pipermerriam:
extracted from #684 

### What was wrong?

The logic for performing the sub-protocol devp2p handshake currently lives in the peer class.  In order for us to support multiple arbitrary sub-protocols, the peer class needs to be agnostic/unaware of the individual protocol handshake logic.

### How was it fixed?

This implements but does not put into use a new API for encapsulating the p2p handshake logic.  Each sub protocol implements a `Handshaker` which holds the logic for performing the handshake and typically a `HandshakeReceipt` that acts as a data storage object for recording data exchanged during the handshake.

A subsequent PR will modify the actual handshake code to make use of this API.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

15-03-15 Lana wlh (1)C1000face

### DIFF
--- a/p2p/handshake.py
+++ b/p2p/handshake.py
@@ -20,6 +20,7 @@ from eth_utils.toolz import groupby, valmap
 
 from eth.tools.logging import ExtendedDebugLogger
 
+from p2p._utils import duplicates
 from p2p.abc import TransportAPI, MultiplexerAPI
 from p2p.constants import (
     SNAPPY_PROTOCOL_VERSION,
@@ -118,14 +119,10 @@ class DevP2PHandshakeParams(NamedTuple):
 @to_tuple
 def _select_capabilities(remote_capabilities: Capabilities,
                          local_capabilities: Capabilities) -> Iterable[Capability]:
-    """Select the sub-protocol to use when talking to the remote.
+    """
+    Select the appropriate shared capabilities between local and remote.
 
-    Find the highest version of our supported sub-protocols that is also supported by the
-    remote and stores an instance of it (with the appropriate cmd_id offset) in
-    self.protocol.
-
-    Raises NoMatchingPeerCapabilities if none of our supported protocols match one of the
-    remote's protocols.
+    https://github.com/ethereum/devp2p/blob/master/rlpx.md#capability-messaging
     """
     # Determine the remote capabilites that intersect with our own.
     matching_capabilities = tuple(sorted(
@@ -186,6 +183,7 @@ async def _do_p2p_handshake(transport: TransportAPI,
                 # parity sends Ping immediately after handshake
                 protocol = P2PProtocol(
                     transport,
+                    cmd_id_offset=0,
                     snappy_support=True,
                 )
             else:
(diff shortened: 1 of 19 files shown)


p2p/p2p_proto.py
```diff
@@ -70,23 +73,25 @@ class Pong(Command):
...
-        # For the base protocol the cmd_id_offset is always 0.
-        # For the base protocol snappy compression should be disabled
-        super().__init__(transport, cmd_id_offset=0, snappy_support=snappy_support)
-
-    def send_handshake(self,
-                       client_version_string: str,
-                       capabilities: Capabilities,
-                       listen_port: int) -> None:
+    @classmethod
+    def get_cmd_offsets(cls, protocol_types: Tuple[Type[Protocol], ...]) -> Iterable[int]:
```
pipermerriam: This shouldn't be here, remove....

tests/p2p/test_select_capabilities.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+
+from p2p.handshake import _select_capabilities
+
+
+A1 = ('A', 1)
+A2 = ('A', 2)
+A10 = ('A', 10)
```
pipermerriam: TODO: remove, not used.

p2p/handshake.py
```diff
@@ -0,0 +1,317 @@
...
+        # the returned receipts.
+        protocol_receipts = await asyncio.gather(*(
+            handshaker.do_handshake(multiplexer, protocol)
+            for handshaker, protocol
+            in zip(selected_handshakers, selected_protocols)
+        ))
+    # Return the `Multiplexer` object as well as the handshake receipts.  The
+    # `Multiplexer` object acts as a container for the individual protocol
+    # instances.
+    return multiplexer, devp2p_receipt, protocol_receipts
```
lithp: I'll start with a giant caveat: I played around with alternatives but couldn't find any.

This method feels like it's missing some kind of abstraction, though I don't know what it is. This method doesn't actually do very much, it spends around 100 lines unpacking data structures and calling a few other methods. In terms of code clarity it's pretty good so I don't want to block the PR over it, but there just has to be an elegant way to simplify this method.

p2p/handshake.py
```diff
@@ -0,0 +1,317 @@
...
+        transport,
+        local_capabilities,
+        p2p_handshake_params,
+        ephemeral_base_protocol,
+        token,
+    )
+
+    # This data structure is simply for easy retrieval of the proper
+    # `Handshaker` for each selectd protocol.
+    protocol_handshakers_by_capability = dict(zip(local_capabilities, protocol_handshakers))
```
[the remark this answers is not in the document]
pipermerriam: good catch, enforcing uniqueness on capabilities is a good check to add.

trinity/protocol/les/handshaker.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+AnyLESProtocol = Union[LESProtocol, LESProtocolV2]
+AnyLESProtocolClass = Union[Type[LESProtocol], Type[LESProtocolV2]]
+
+
+class LESHandshakeReceipt(HandshakeReceipt):
+    handshake_params: LESHandshakeParams
+
+    def __init__(self,
```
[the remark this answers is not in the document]
pipermerriam: I wanted a firm tie between the receipt and the protocol the receipt was for.  You *can* currently derive protocol from any of the `HandsharkeParams` classes but that's incidental and can't always be guaranteed to be the case and would require custom logic per protocol to do so.  By binding the receipt to the protocol we get a way to programmatically pair them with each other and allow arbitrary retrieval of receipt by protocol which is how we can find out what network_id an ETH peer is part of.

p2p/handshake.py
pipermerriam: The abstraction it feels like it's missing is the `Connection` object which comes in 2-3 steps which wraps up the multiplexer and all the receipts into a single API.

PR #878, opened 2019-08-05
Fix bug with trinity not shutting down cleanly

### LINKED ISSUE #872: Trinity doesn't shut down cleanly
* Trinity version: 6493c94656b907c
* OS: osx/linux

### What is wrong?

Ctrl-C no longer cleanly exits. You need a second Ctrl-C to get out. Occasionally, it leaves dirty IPC files, or even hanging processes that never exit.

### How can it be fixed

:man_shrugging: 

I ran `git bisect` and narrowed it down to the above commit. cc @cburgdorf  @pipermerriam

pipermerriam:
fixes #872 

### What was wrong?

When the `PluginManager` was moved into a service I failed to tie the shutdown process of the plugin manager to the shutdown process of the service and that caused Trinity to hang on shutdown since it didn't propagate the shutdown request to the various plugins.

### How was it fixed?

Made sure to call `PluginManager.shutdown_blocking` in the `BaseService._cleanup`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -201,10 +201,13 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
 
             await self.wait(rate_limiter.take())
 
-            await self.wait(asyncio.gather(*(
-                self._add_peers_from_backend(backend)
-                for backend in self.peer_backends
-            )))
+            try:
+                await self.wait(asyncio.gather(*(
+                    self._add_peers_from_backend(backend)
+                    for backend in self.peer_backends
+                )))
+            except OperationCancelled:
+                break
 
     def __len__(self) -> int:
         return len(self.connected_nodes)
--- a/trinity/_utils/mp.py
+++ b/trinity/_utils/mp.py
@@ -32,12 +32,14 @@ def async_method(method_name: str) -> Callable[..., Any]:
             method_name,
             args,
         )
+    method.__name__ = method_name
     return method
 
 
 def sync_method(method_name: str) -> Callable[..., Any]:
     def method(self: Any, *args: Any, **kwargs: Any) -> Any:
         return self._callmethod(method_name, args, kwargs)
+    method.__name__ = method_name
     return method
 
 
--- a/trinity/event_bus.py
+++ b/trinity/event_bus.py
@@ -54,16 +54,20 @@ class PluginManagerService(BaseService):
             # start the background process that tracks and propagates available
             # endpoints to the other connected endpoints
             self.run_daemon_task(self._track_and_propagate_available_endpoints())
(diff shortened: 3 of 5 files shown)


p2p/peer_pool.py
```diff
@@ -201,10 +201,13 @@ def setup_peer_backends(self) -> Tuple[BasePeerBackend, ...]:
...
-            await self.wait(asyncio.gather(*(
-                self._add_peers_from_backend(backend)
-                for backend in self.peer_backends
-            )))
+            try:
+                await self.wait(asyncio.gather(*(
+                    self._add_peers_from_backend(backend)
+                    for backend in self.peer_backends
+                )))
+            except OperationCancelled:
```
pipermerriam: This catches a shutdown exception I got during experimentation with this.

trinity/_utils/mp.py
```diff
@@ -32,12 +32,14 @@ def async_method(method_name: str) -> Callable[..., Any]:
             method_name,
             args,
         )
+    method.__name__ = method_name
```
pipermerriam: This makes these methods have an identifiable name when they end up in terminal output.

trinity/sync/common/headers.py
```diff
@@ -165,10 +165,10 @@ def __init__(self,
             # validate that parents and children match
             pairs = tuple(zip(parents, children))
             try:
-                validate_pair_coros = [
+                validate_pair_coros = (
```
pipermerriam: lazy-ify this so that the none of the functions actually get called until they are consumed in `asyncio.gather`

cburgdorf: Might be worth putting this into a separate commit for improved clarity

trinity/event_bus.py
```diff
@@ -54,17 +54,21 @@ def __init__(self,
             # start the background process that tracks and propagates available
             # endpoints to the other connected endpoints
             self.run_daemon_task(self._track_and_propagate_available_endpoints())
+            self.run_daemon_task(self._handle_shutdown_request())
```
cburgdorf: Sorry for not spotting this in my initial review.

trinity/sync/common/headers.py
carver: AFAICT, this doesn't have an impact one way or another. It expands the whole collection as an early step of invoking `gather()`. No need to change it back, I just want to make sure we all agree on what's happening here. :)

pipermerriam: :+1: 

This was partially me stabbing at a warning I was getting about coroutines not being awaited and I thought *maybe* it was one of these getting called but not getting awaited due to a shutdown interrupting. (I don't think it actually fixed it)

PR #883, opened 2019-08-05
Beam Sync: annotate event with num nodes collected

carver:
### What was wrong?

We are missing statistics on number of trie nodes collected by block during Beam Sync.

### How was it fixed?

Return the number of nodes requested during a `CollectMissingAccount` request, adding the result to the `MissingAccountCollected` response event. (and the same for storage requests)

Bytecode is always a single node request, so it's irrelevant.

### To-Do

- [ ] Clean up commit history

- [ ] ~~Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)~~

#### Cute Animal Picture

I'm starting to run out of animals on balance beams...

### DIFF
--- a/trinity/sync/beam/importer.py
+++ b/trinity/sync/beam/importer.py
@@ -36,6 +36,9 @@ from trinity.sync.common.events import (
     CollectMissingBytecode,
     CollectMissingStorage,
     DoStatelessBlockImport,
+    MissingAccountCollected,
+    MissingBytecodeCollected,
+    MissingStorageCollected,
     StatelessBlockImportDone,
 )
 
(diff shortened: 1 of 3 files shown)


trinity/sync/beam/importer.py
```diff
@@ -118,7 +121,7 @@ def _pause_on_missing_data(
                 try:
                     return unbound_vm_method(self, *args, **kwargs)  # type: ignore
                 except MissingAccountTrieNode as exc:
-                    future = asyncio.run_coroutine_threadsafe(
+                    account_future = asyncio.run_coroutine_threadsafe(
```
carver: When these three were sharing the same `future` variable, mypy got annoyed that the result value was different. Using three different variables made mypy happy, and is probably a slight improvement.

PR #884, opened 2019-08-06
Add "Performance section" to generated release notes

cburgdorf:
### What was wrong?

The frequently used "Performance improvements" section of our release notes got lost when we migrated to `towncrier`. We want it back.

### How was it fixed?

Manually configure the sections `towncrier` uses and add a `Performance improvements` category.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,3 +5,33 @@ filename = "docs/release_notes.rst"
 directory = "newsfragments"
 underlines = ["-", "~", "^"]
 issue_format = "`#{issue} <https://github.com/ethereum/trinity/issues/{issue}>`__"
+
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -5,3 +5,33 @@ filename = "docs/release_notes.rst"
 directory = "newsfragments"
 underlines = ["-", "~", "^"]
 issue_format = "`#{issue} <https://github.com/ethereum/trinity/issues/{issue}>`__"
+
```
cburgdorf: There is no way to just add a category so :point_down: are all default categories plus a `performance` one.

PR #885, opened 2019-08-07
DiscV5 Packet Update

jannikluhn:
### What was wrong?

The discv5 spec was [updated](https://github.com/ethereum/devp2p/pull/108/files#diff-1c7ef4048f367839576ddbdd164a5432), so our wire protocol code was outdated.

### How was it fixed?

- the id nonce is included in the auth header and has a fixed size now
- there's now a version number in the auth response (which is not expected to change though, so I just hardcoded it)
- auth responses are encrypted without "additional authenticated data"
- who are you packets have lost their tag (unfortunately, this means we can't recover the sender node id which has some implications on the service design, but that's the topic of a different PR)
- ENRs are included in the regtopic message
- who are you packets don't have a tag anymore (which unfortunately

#### Cute Animal Picture


p2p/discv5/packets.py
```diff
@@ -137,11 +142,21 @@ def decrypt_auth_response(self, auth_response_key: AES128Key) -> Tuple[bytes, Op
                 f"Auth response contains bytes instead of list: {encode_hex(decoded_rlp)}"
             )
 
-        if len(decoded_rlp) != 2:
+        if len(decoded_rlp) != 3:
```
pipermerriam: This would return true if `decoded_rlp` was a length-3 byte-string right?  Probably need to verify that `decoded_rlp` is not `bytes`

jannikluhn: Right, this is checked here already: https://github.com/ethereum/trinity/pull/885/files#diff-bb1189a6c7d45537d9701b8331183284L135

probably hidden because it wasn't touched in this PR.

PR #886, opened 2019-08-07
Req/Resp domain: basic utility functions and `Hello`

mhchia:
### What was wrong?
According to https://github.com/ethereum/eth2.0-specs/blob/dev/specs/networking/p2p-interface.md#the-reqresp-domain.

### How was it fixed?
- Add basic functions and configs for Req/Resp domain.
- Add `say_hello` and `handle_hello` for Req/Resp `hello`.
- Add basic test for `hello`.

### To-Do
- [ ] Disconnect from peers that fail to say `hello`
    - Define the timeout for peers to say `hello`.
    - Confirm who says hello. Reference: https://github.com/ethereum/eth2.0-specs/issues/1351

#### Cute Animal Picture


trinity/protocol/bcc_libp2p/node.py
```diff
@@ -214,3 +250,152 @@ def peer_store(self) -> PeerStore:
...
+    #   returns successfully, then close the stream. Otherwise, reset the stream.
+
+    # TODO: Handle the reputation of peers. Deduct their scores and even disconnect when they
+    #   behave.
+
+    # TODO: Register notifee to the `Network` to
+    #   - Record peers' joining time.
+    #   - Disconnect peers when they fail to join in a certain amount of time.
+
+    async def _validate_hello_req(self, hello_other_side: HelloRequest) -> bool:
```
NIC619: Personally I prefer exception based validation scheme(and to be consistent across the codebase).

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -214,3 +250,152 @@ def peer_store(self) -> PeerStore:
...
+            finalized_epoch=finalized_checkpoint.epoch,
+            head_root=state.hash_tree_root,
+            head_slot=state.slot,
+        )
+
+    async def _handle_hello(self, stream: INetStream) -> None:
+        # TODO: Handle `stream.close` and `stream.reset`
+        peer_id = stream.mplex_conn.peer_id
+        if peer_id in self.handshaked_peers:
+            error_msg = f"already handshaked with {peer_id} before"
```
NIC619: nit:
```suggestion
            error_msg = f"already handshaked with peer={peer_id}"
```

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -214,3 +250,152 @@ def peer_store(self) -> PeerStore:
...
+            self.logger.info(f"Handshake failed: {error_msg}.")
+            # FIXME: Use `Stream.reset()` when `NetStream` has this API.
+            # await stream.reset()
+            raise HandshakeFailure(error_msg)
```
NIC619: Question: should we raise exception in the handler? would it cause the `host` to shut down?

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -214,3 +250,152 @@ def peer_store(self) -> PeerStore:
...
+            )
+            # FIXME: Use `Stream.reset()` when `NetStream` has this API.
+            # await stream.reset()
+            # TODO: Disconnect
+            raise HandshakeFailure(error_msg)
+
+        hello_mine = self._make_hello_packet()
+
+        self.logger.debug(f"Sending our hello message {hello_mine}")
+        # TODO: Find out when we should respond the `ResponseCode`
```
NIC619: Maybe move this comment to the top of this function?

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -214,3 +250,152 @@ def peer_store(self) -> PeerStore:
...
+        #   other than `ResponseCode.SUCCESS`.
+        try:
+            await write_resp(stream, hello_mine, ResponseCode.SUCCESS)
+        except WriteMessageFailure as error:
+            raise HandshakeFailure(f"fail to write response={hello_mine}") from error
+
+        self.handshaked_peers.add(peer_id)
+
+        self.logger.debug(f"Handshake from {peer_id} is finished. Added to the `handshake_peers`.")
```
NIC619: nit:
```suggestion
        self.logger.debug(f"Handshake from peer={peer_id} is finished. Added to the `handshake_peers`.")
```

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -214,3 +250,152 @@ def peer_store(self) -> PeerStore:
...
+    async def say_hello(self, peer_id: ID) -> None:
+        # TODO: Handle `stream.close` and `stream.reset`
+        if peer_id in self.handshaked_peers:
+            error_msg = f"already handshaked with {peer_id} before"
+            self.logger.info(f"Handshake failed: {error_msg}.")
+            raise HandshakeFailure(error_msg)
+
+        hello_mine = self._make_hello_packet()
+
+        self.logger.debug(f"Opening new stream to {peer_id} with protocols {[REQ_RESP_HELLO_SSZ]}.")
```
NIC619: nit:
```suggestion
        self.logger.debug(f"Opening new stream to peer={peer_id} with protocols={[REQ_RESP_HELLO_SSZ]}.")
```

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -214,3 +250,152 @@ def peer_store(self) -> PeerStore:
...
+        stream = await self.host.new_stream(peer_id, [REQ_RESP_HELLO_SSZ])
+        self.logger.debug(f"Sending our hello message {hello_mine}.")
+        try:
+            await write_req(stream, hello_mine)
+        except WriteMessageFailure as error:
+            raise HandshakeFailure(f"fail to write request={hello_mine}") from error
```
NIC619: Should we also do `stream.reset()` and disconnect here?

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -214,3 +250,152 @@ def peer_store(self) -> PeerStore:
...
+            error_msg = f"hello message {hello_other_side} is invalid"
+            self.logger.info(f"Handshake failed: {error_msg}. Disconnecting {peer_id}.")
+            # FIXME: Use `Stream.reset()` when `NetStream` has this API.
+            # await stream.reset()
+            # TODO: Disconnect
+            raise HandshakeFailure(error_msg)
+
+        self.handshaked_peers.add(peer_id)
+
+        self.logger.debug(f"Handshake to {peer_id} is finished. Added to the `handshake_peers`.")
```
NIC619: nit:
```suggestion
        self.logger.debug(f"Handshake to peer={peer_id} is finished. Added to the `handshake_peers`.")
```

tests/libp2p/bcc/test_utils.py
```diff
@@ -1,17 +1,101 @@
...
+
+
+@pytest.mark.parametrize(
+    "msg",
+    (
+        hello_req,
+    )
+)
+@pytest.mark.asyncio
+async def test_read_write_req_msg_round_trip(msg):
```
NIC619: nit:
```suggestion
async def test_read_write_req_msg(msg):
```

tests/libp2p/bcc/test_utils.py
```diff
@@ -1,17 +1,101 @@
...
+
+
+@pytest.mark.parametrize(
+    "msg",
+    (
+        hello_req,
+    )
+)
+@pytest.mark.asyncio
+async def test_read_write_resp_msg_round_trip_success_code(msg):
```
NIC619: nit:
```suggestion
async def test_read_write_resp_msg(msg):
```

tests/libp2p/bcc/test_utils.py
```diff
@@ -1,17 +1,101 @@
...
+    s = FakeNetStream()
+    resp_code = ResponseCode.SUCCESS
+    await write_resp(s, msg, resp_code)
+    resp_code_read, msg_read = await read_resp(s, HelloRequest)
+    assert resp_code_read == resp_code
+    assert msg_read == msg
```
NIC619: Should we add more test cases for failure/malformed response?

trinity/protocol/bcc_libp2p/utils.py
```diff
@@ -24,4 +56,125 @@ def make_tcp_ip_maddr(ip: str, port: int) -> Multiaddr:
...
+    stream: INetStream,
+    msg: Union[MsgType, ErrorMsgType],
+    resp_code: ResponseCode,
+) -> None:
+    try:
+        resp_code_byte = resp_code.value.to_bytes(1, "big")
+    except OverflowError as error:
+        raise WriteMessageFailure(f"result={resp_code} is not valid") from error
+    # `ErrorMsgType`
+    if isinstance(msg, str):
```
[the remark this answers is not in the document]
mhchia: Good catch!

tests/libp2p/bcc/test_utils.py
```diff
@@ -1,17 +1,206 @@
...
+    with pytest.raises(ReadMessageFailure):
+        await read_req(s, HelloRequest)
+
+
+@pytest.mark.asyncio
+async def test_write_req_failure(monkeypatch):
+    s = FakeNetStream()
+
+    # Test: Raise `WriteMessageFailure` if `ssz.SerializationError` is thrown.
+    monkeypatch.setattr(ssz, "encode", _fake_encode)
```
pipermerriam: Why do you need to mock this?  Is there no way to actually cause a serialization error?  Can you do something like provide the wrong type for an object's field?

tests/libp2p/bcc/test_utils.py
```diff
@@ -1,17 +1,206 @@
...
+    s = FakeNetStream()
+    # Test: Raise `ReadMessageFailure` if the time is out.
+    with pytest.raises(ReadMessageFailure):
+        await read_resp(s, HelloRequest)
+
+    # Test: Raise `ReadMessageFailure` if `read` returns `b""`.
+
+    async def _fake_read(n):
+        return b""
+    monkeypatch.setattr(s, 'read', _fake_read)
```
pipermerriam: Seems like you should be able to construct a message that would raise the desired error without mocking...

tests/plugins/eth2/beacon/helpers.py
```diff
@@ -39,3 +45,24 @@
...
 )
+
+
+def get_chain_from_genesis(db, indices):
+    # pubkey -> privkey map
+    validator_keymap = {
+        index_to_pubkey[index]: keymap[index_to_pubkey[index]]
+        for index in indices
+    }
+    genesis_data = BeaconGenesisData(
```
pipermerriam: I'd push for not adding anything extra to this file and building these as factories.  I have a branch right now that mostly removes this file, replacing it with factories.  You should be able to pretty easily create granular factories for all of the data types you need and then compose them together to build the test data you need.  When you need something special like `chain_class.from_genesis` you can do something like adding an extra `classmethod` to the factory class or overriding the built-in `_create` method from the factory class.

tests/libp2p/bcc/test_utils.py
mhchia: Thanks for the direction. I tried something but failed with `TypeError` at first. Now it works.

tests/plugins/eth2/beacon/helpers.py
mhchia: @pipermerriam  

I refactored the logic of initializing `BeaconChain` and added a `BeaconChainFactory` in [eth2/beacon/tools/factories](https://github.com/ethereum/trinity/blob/f069abea6889a805b8073a48d0251bdb7e6f0203/eth2/beacon/tools/factories.py#L65-L78). It is currently used to generate `BeaconChain` in the tests. Could you have a review on these functions?

Another thing I'm not sure how to deal with is `keymap`, `genesis_block`, and `genesis_state`. Currently, I leave them in [](https://github.com/ethereum/trinity/blob/f069abea6889a805b8073a48d0251bdb7e6f0203/eth2/beacon/tools/factories.py#L47-L59) with `BeaconChainFactory`. The reason why I didn't add factories for them is that they can be shared among all tests, and the process of generating `keymap` is relatively expensive due to `privtopub`. But I'm not sure whether it is acceptable to leave them as the global vars in `factories.py`. It will be really great to have your comment on this! Thank you :pray:

pipermerriam: I think the global state is fine for now.  There's probably a route that lets us still use factories for those parts but to do some caching either via memoization or something.  For now :+1:

trinity/protocol/bcc_libp2p/messages.py
```diff
@@ -0,0 +1,96 @@
...
+
+from eth_typing import (
+    Hash32,
+)
+
+from eth.constants import (
+    ZERO_HASH32,
+)
+
+from eth2.beacon.typing import (  # noqa: F401
```
ChihChengLiang: this `# noqa: F401` can be removed

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -214,3 +251,170 @@ def peer_store(self) -> PeerStore:
...
+        )
+
+    async def _handle_hello(self, stream: INetStream) -> None:
+        # TODO: Find out when we should respond the `ResponseCode`
+        #   other than `ResponseCode.SUCCESS`.
+
+        # TODO: Handle `stream.close` and `stream.reset`
+        peer_id = stream.mplex_conn.peer_id
+        if peer_id in self.handshaked_peers:
+            self.logger.info(f"Handshake failed: already handshaked with {peer_id} before.")
```
ChihChengLiang: nit: should use `%` style string formatting for log messages. Same comments to other logs in this function

PR #887, opened 2019-08-07
Use p2p handshakers in peer handshake

pipermerriam:
based on https://github.com/ethereum/trinity/pull/955

### What was wrong?

#869 implements a new API for handshaking with peers that supports arbitrary sub-protocols.  This code isn't being used in the `BasePeer`

### How was it fixed?

Update the `BasePeer` and various code that deals with handshaking to use this new API.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


p2p/protocol.py
```diff
@@ -164,6 +162,12 @@ def __init__(self, transport: TransportAPI, cmd_id_offset: int, snappy_support:
         self.cmd_by_type = {type(cmd): cmd for cmd in self.commands}
         self.cmd_by_id = {cmd.cmd_id: cmd for cmd in self.commands}
 
+    def __eq__(self, other: Any) -> bool:
```
pipermerriam: I don't think these are needed anymore.

p2p/service.py
```diff
@@ -98,9 +98,13 @@ def get_event_loop(self) -> asyncio.AbstractEventLoop:
         finished_callback (if one was passed).
         """
         if self.is_running:
-            raise ValidationError("Cannot start the service while it's already running")
+            raise ValidationError("Cannot start the service while it's already running: %s", self)
```
pipermerriam: standalone

p2p/service.py
```diff
@@ -314,6 +318,9 @@ def cancel_nowait(self) -> None:
             if self._child_services:
                 self.logger.debug("Pending child services: %s", list(self._child_services))
             await self._forcibly_cancel_all_tasks()
+            # Sleep a bit because the Future.cancel() method just schedules the callbacks, so we
+            # need to give the event loop a chance to actually call them.
+            await asyncio.sleep(0.01)
```
pipermerriam: standalone

p2p/tools/asyncio_streams.py
```diff
@@ -63,6 +63,15 @@ def can_write_eof(self) -> bool:
     def abort(self) -> None:
         self._is_closing = True
 
+    async def drain(self) -> None:
+        pass
+
+    def set_extra_info(self, name: str, value: Any) -> None:
+        self._extra_info[name] = value
+
+    def get_extra_info(self, name: str) -> Any:
```
pipermerriam: I think these are no longer needed

p2p/transport.py
```diff
@@ -84,6 +84,9 @@ def __init__(self,
         mac_cipher = Cipher(algorithms.AES(mac_secret), modes.ECB(), default_backend())
         self._mac_enc = mac_cipher.encryptor().update
 
+        self._received_message_count = 0
+        self._sent_message_count = 0
```
pipermerriam: can be removed.

tests/core/p2p-proto/test_les_protocol_commands.py
```diff
@@ -1,25 +1,27 @@
...
-        request,
-        event_loop,
-        alice_peer_class=LESPeer,
-    )
+def token(event_loop):
+    token = CancelToken('test-master', loop=event_loop)
+    try:
+        yield token
+    finally:
+        token.trigger()
```
pipermerriam: test if this is actually needed.

tests/core/p2p-proto/test_les_protocol_commands.py
```diff
@@ -43,16 +45,24 @@
...
 
+    # setup message collection
+    messages = []
+    got_message = asyncio.Event()
+
+    async def collect_messages(cmd, msg):
+        messages.append((cmd, msg))
+        got_message.set()
+
+    peer.connection.add_protocol_handler(LESProtocol.name, collect_messages)
```
pipermerriam: rework to use multiplexer.

tests/p2p/test_auth.py
```diff
@@ -141,20 +142,28 @@
...
-        context=ParagonContext(),
-    )
-    responder_peer.base_protocol.send_handshake(
-        client_version_string=responder_peer.context.client_version_string,
-        capabilities=responder_peer.capabilities,
-        listen_port=responder_peer.context.listen_port,
+        base_protocol=responder_p2p_protocol,
+        protocols=(),
     )
+    responder_multiplexer.get_base_protocol().send_handshake('responder', capabilities, 30303)
```
pipermerriam: maybe able to extract now to reduce size of this PR since this uses existing APIs and doesn't involve the handshakers.

trinity/plugins/builtin/request_server/plugin.py
```diff
@@ -73,12 +66,6 @@ def do_start(self) -> None:
                 trinity_config.get_app_config(Eth1AppConfig),
                 db_manager,
             )
-        elif trinity_config.has_app_config(BeaconAppConfig):
```
pipermerriam: TODO: restore.

pipermerriam:
blocked by https://github.com/ethereum/trinity/pull/937

tests/core/p2p-proto/test_server.py
```diff
@@ -78,20 +85,27 @@ def get_server(privkey, address, event_bus):
 @pytest.fixture
 async def server(event_bus):
     server = get_server(RECEIVER_PRIVKEY, SERVER_ADDRESS, event_bus)
```
lithp: I guess you changed this because you also needed `peer_pool` and `receive_server` to be running?

I remember that a couple times now we've wanted to wait until some service actually starts, this seems to be a common problem! Do you think it makes sense to add an event to `ServiceEvents`? If the `PeerPool` called `self.events.running.set()` just before it awaited its cancel token we could replace this loop here with an await.

tests/core/p2p-proto/test_server.py
```diff
@@ -106,18 +120,34 @@ def get_server(privkey, address, event_bus):
...
         ingress_mac=ingress_mac,
     )
-    initiator_peer = ParagonPeer(
-        transport=transport,
+
+    factory = ParagonPeerFactory(
+        initiator.privkey,
         context=ParagonContext(),
         token=token,
     )
```
lithp: This is an aesthetic objection more than anything but it hurts to see this kind of loop when other parts of the program definitely have enough information to call us at the right time.

What do you think of trying to wrap this up into a nicer api?

```python
class NewPeerSubscriber(PeerSubscriber):
    subscription_msg_types = frozenset({})
    msg_queue_maxsize = 0

    def __init__(self):
        self.event = asyncio.Event()
    
    def register_peer(self, peer):
        self.event.set()

    async def new_peer_connected(self):
        await self.event.wait()


async def new_peer_connected(peer_pool):
    """
    Awaitable which returns control to the caller when the peer_pool receives a new peer
    """
    subscriber = NewPeerSubscriber()
    with subscriber.subscribe(peer_pool):
        await subscriber.new_peer_connected()
```

This block then becomes:

```python
await new_peer_connected(server.peer_pool)
```

trinity/protocol/bcc/handshaker.py
```diff
@@ -0,0 +1,84 @@
...
+    @property
+    def genesis_root(self) -> Hash32:
+        return self.handshake_params.genesis_root
+
+    @property
+    def network_id(self) -> int:
+        return self.handshake_params.network_id
+
+
+class BCCHandshaker(Handshaker):
```
lithp: `BBCHandshaker`, `ETHHandshaker`, and `LESHandshaker` are very formulaic, in the future it might be worth figuring out how to factor away some of this duplicated code.

Maybe `do_handshake` can become a method on the Protocol: `Protocol.receive_handshake(local_handshake_params, remote_message) -> HandshakeParams`. The flow could look like:

- The function which does the handshaking (`negotiate_protocol_handshakes`) accepts `HandshakeParams` for each supported protocol.
- Once the mutual protocols have been determined `negotiate_protocol_handshakes` calls `protocol.send_handshake()` for each param.
- Then, probably with a helper, `negotiate_protocol_handshakes` waits for the first message from each protocol, and passes that message to `Protocol.receive_handshake()`.
- `Protocol.receive_handshake()` parses the message and does all the checks that this peer is one we want to be talking to, then returns the remote's `HandshakeParams` that should be passed to the peer.

Something like this would coalesce all the duplicated code into one place, while still leaving somewhere the parts which vary from protocol to protocol can belong.

trinity/protocol/eth/peer.py
```diff
@@ -82,6 +78,22 @@ class ETHPeer(BaseChainPeer):
 
     _requests: ETHExchangeHandler = None
 
+    def process_receipts(self,
```
lithp: I can't quite find it, but there might be a factoring which prevents us from needing to scan through to find the right receipt and duplicating the code which expects specific receipts to exist.

My current idea is for the receipts to be passed as a `Dict[Type[HandshakeReceipt], HandshakeReceipt]`, and maybe `BasePeer` subclasses can declare some kind of `receipt_handlers: Dict[Type[HandshakeReceipt], Callable[[HandshakeReceipt], None]]`, then `BasePeer` can call the correct handler and do the checking that we've received all the expected receipts. That doesn't quite work though, and I know all of this is in flux anyway, maybe we can come back to it in the future.

p2p/peer.py
```diff
@@ -179,39 +180,30 @@ class BasePeer(BaseService):
...
         self.context = context
 
+        # Connection instance
+        self.multiplexer = multiplexer
+
+        self.base_protocol = self.multiplexer.get_base_protocol()
+
+        # TODO: need to remove this property but for now it is here to support
+        # backwards compat
+        self.sub_proto = self.multiplexer.get_protocols()[1]
```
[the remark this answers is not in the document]
pipermerriam: I'll adjust to use `get_protocol_by_type`

p2p/peer.py
```diff
@@ -231,6 +223,16 @@ def __init__(self,
         self.boot_manager = self.get_boot_manager()
         self.connection_tracker = self.setup_connection_tracker()
 
+        self.process_receipts(devp2p_receipt, protocol_receipts)
+
+    def process_receipts(self,
+                         devp2p_receipt: DevP2PReceipt,
```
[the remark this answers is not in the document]
pipermerriam: This is basically a temporary API so I'm inclined to keep it as-is for now since it should be removed in a few PRs.

p2p/peer.py
```diff
@@ -454,8 +337,8 @@ def handle_p2p_msg(self, cmd: CommandAPI, msg: Payload) -> None:
         else:
             raise UnexpectedMessage(f"Unexpected msg: {cmd} ({msg})")
 
-    async def handle_sub_proto_stream(self, multiplexer: MultiplexerAPI) -> None:
-        async for cmd, msg in self.wait_iter(multiplexer.stream_protocol_messages(self.sub_proto)):
+    async def handle_sub_proto_stream(self) -> None:
```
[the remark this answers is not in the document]
pipermerriam: All of this is just glue to hold things together during the transition.  I believe our concept of a `Peer` is going to either shift dramatically or go away entirely sometime soon.

tests/core/p2p-proto/test_server.py
pipermerriam: We've avoided a a running event because it would require every service to mark itself as `running` which would be easy to forget.

Also, I tried that approach just using a custom event inside the class but the race condition was still present, hence having these loops in place (for now).

I'm planning to address this separately later as part of the `ConnectionPool` which would replace the `PeerPool` and would expose APIs for both dial out and listening for dial in rather than our current state which is that we require an additional service to do the listening.

trinity/protocol/bcc/handshaker.py
pipermerriam: :+1: to this type of improvement.  We'll get some good opportunities to experiment with this stuff over as this API evolves.

trinity/server.py
```diff
@@ -168,7 +175,23 @@ def _cleanup_reader_and_writer() -> None:
     async def _receive_handshake(
             self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter) -> None:
         factory = self.peer_pool.get_peer_factory()
-        peer = await receive_handshake(reader, writer, factory)
+        handshakers = await factory.get_handshakers()
+        multiplexer, devp2p_receipt, protocol_receipts = await receive_handshake(
```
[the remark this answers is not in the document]
pipermerriam: The factory class should get eliminated soon so I'm going to leave as-is.

PR #889, opened 2019-08-07
added from field to transactions requested through JSON-RPC

voith:
### What was wrong?
`eth_getTransactionByBlockHashAndIndex` and `eth_getTransactionByBlockNumberAndIndex` did not include `from` address. The output looked like this:
```
In [9]: w3.eth.getTransactionByBlock('0x45bb3a355c28c2f4c966eaad1e54b59b1f663c09264c5ab5c12218aac4cb4c0d', 1)
Out[9]:
AttributeDict({'hash': HexBytes('0x251112243025884f168152933401d96f9579a0ebcb433e305ab56347a2731a23'),
 'nonce': 0,
 'gas': 2000000,
 'gasPrice': 4000000000,
 'to': '0xF6DB16a918C5486458d90025204Fd6fA46c5273C',
 'value': 12100000000000000,
 'input': '0x',
 'r': HexBytes('0x44fa73dadee897539109a12a38a67ad51a58f66f5db2544564fc9d71c4fab243'),
 's': HexBytes('0x138a8452f0b880ce2ae3cbd27b4072279a7a8877eda74971a13352f8f383d997'),
 'v': 41})
```

### How was it fixed?
`from` field was added for `eth_getTransactionByBlockHashAndIndex` and `eth_getTransactionByBlockNumberAndIndex`. The new response looks like this:
```
In [14]: w3.eth.getTransactionByBlock('0x45bb3a355c28c2f4c966eaad1e54b59b1f663c09264c5ab5c12218aac4cb4c0d', 1)
Out[14]:
AttributeDict({'hash': HexBytes('0x251112243025884f168152933401d96f9579a0ebcb433e305ab56347a2731a23'),
 'nonce': 0,
 'gas': 2000000,
 'gasPrice': 4000000000,
 'to': '0xF6DB16a918C5486458d90025204Fd6fA46c5273C',
 'value': 12100000000000000,
 'input': '0x',
 'r': HexBytes('0x44fa73dadee897539109a12a38a67ad51a58f66f5db2544564fc9d71c4fab243'),
 's': HexBytes('0x138a8452f0b880ce2ae3cbd27b4072279a7a8877eda74971a13352f8f383d997'),
 'v': 41,
 'from': '0x340AEdf6a7309456045740Bd014EE8bE30D568a6'})
```

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
+++ b/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
@@ -6,6 +6,7 @@ from pathlib import Path
 import pytest
 
 from eth_utils.toolz import (
+    compose,
     dissoc,
     identity,
     get_in,
@@ -13,10 +14,15 @@ from eth_utils.toolz import (
 
 from eth_utils import (
     add_0x_prefix,
+    is_address,
     is_hex,
     is_integer,
     is_string,
+    to_checksum_address,
     to_tuple,
+)
+from eth_utils.curried import (
+    apply_formatter_if,
 )
 
 from eth.chains.mainnet import (
@@ -164,7 +170,10 @@ RPC_TRANSACTION_NORMALIZERS = {
     'gasPrice': remove_leading_zeros,
     'value': remove_leading_zeros,
     'data': empty_to_0x,
-    'to': add_0x_prefix,
+    'to': compose(
+        apply_formatter_if(is_address, to_checksum_address),
+        add_0x_prefix
+    ),
     'r': remove_leading_zeros,
     's': remove_leading_zeros,
     'v': remove_leading_zeros,
@@ -335,6 +344,7 @@ def validate_rpc_transaction_vs_fixture(transaction, fixture):
     actual_transaction = dissoc(
         transaction,
         'hash',
+        'from',
     )
     assert actual_transaction == expected
 
--- a/trinity/rpc/format.py
+++ b/trinity/rpc/format.py
@@ -12,10 +12,13 @@ from eth_utils.toolz import (
 )
 
 from eth_utils import (
+    apply_formatter_if,
     apply_formatters_to_dict,
     decode_hex,
     encode_hex,
     int_to_big_endian,
+    is_address,
+    to_checksum_address,
 )
 
 import rlp
@@ -37,18 +40,23 @@ from trinity.chains.base import BaseAsyncChain
 
 
 def transaction_to_dict(transaction: BaseTransaction) -> Dict[str, str]:
-    return dict(
-        hash=encode_hex(transaction.hash),
-        nonce=hex(transaction.nonce),
-        gas=hex(transaction.gas),
-        gasPrice=hex(transaction.gas_price),
-        to=encode_hex(transaction.to),
-        value=hex(transaction.value),
-        input=encode_hex(transaction.data),
-        r=hex(transaction.r),
-        s=hex(transaction.s),
-        v=hex(transaction.v),
-    )
+    return {
+        'hash': encode_hex(transaction.hash),
+        'nonce': hex(transaction.nonce),
+        'gas': hex(transaction.gas),
+        'gasPrice': hex(transaction.gas_price),
+        'from': to_checksum_address(transaction.sender),
+        'to': apply_formatter_if(
+            is_address,
+            to_checksum_address,
+            encode_hex(transaction.to)
+        ),
+        'value': hex(transaction.value),
+        'input': encode_hex(transaction.data),
+        'r': hex(transaction.r),
+        's': hex(transaction.s),
+        'v': hex(transaction.v),
+    }
 
 
 hexstr_to_int = functools.partial(int, base=16)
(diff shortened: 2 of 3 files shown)


voith:
will add a test tomorrow. Marked it as WIP.

voith:
cc @carver, ready for review.

trinity/rpc/format.py
```diff
@@ -48,6 +48,7 @@ def transaction_to_dict(transaction: BaseTransaction) -> Dict[str, str]:
         r=hex(transaction.r),
         s=hex(transaction.s),
         v=hex(transaction.v),
+        **{'from': encode_hex(transaction.sender)}
```
pipermerriam: this should probably get checksummed.  Maybe use `eth_utils.to_checksum_address`?

trinity/rpc/format.py
```diff
@@ -48,6 +48,7 @@ def transaction_to_dict(transaction: BaseTransaction) -> Dict[str, str]:
         r=hex(transaction.r),
         s=hex(transaction.s),
```
pipermerriam: might be nicer to just convert this whole `dict` to simply be done as a dictionary like:

```python
{
   'key': value,
}
```

That way we don't have to have the awkward `**{'from': ...}` at the end.

trinity/rpc/format.py
voith: @pipermerriam AFAIK the code, sender is always extracted from `r`, `s`, `v`. And since this is done using a standard library, the address will always be checksummed:
https://github.com/ethereum/py-evm/blob/d2ccc1e8a9be1cb92cb15a8bb117bb6a1972636e/eth/_utils/transactions.py#L95-L112

trinity/rpc/format.py
voith: done

trinity/rpc/format.py
carver: huh? @pipermerriam is right, it should produce an EIP-55 checksum by using `eth_utils.to_checksum_address()` instead of `encode_hex()`.

trinity/rpc/format.py
```diff
@@ -37,18 +37,19 @@
...
-        s=hex(transaction.s),
-        v=hex(transaction.v),
-    )
+    return {
+        'hash': encode_hex(transaction.hash),
+        'nonce': hex(transaction.nonce),
+        'gas': hex(transaction.gas),
+        'gasPrice': hex(transaction.gas_price),
+        'from': encode_hex(transaction.sender),
+        'to': encode_hex(transaction.to),
```
carver: FWIW, the checksum got missed in the `to` field that was already there. Let's fix that now, also! :)

trinity/rpc/format.py
voith: oops my bad, I used web3 to fetch the results. I got fooled by the result because web3 does  the conversion in any case. Tested it with `eth-keys`, it is not in fact checksummned.

voith: done

voith:
oops got to fix tests now

carver:
Ah yeah, maybe that's why it was in the non-EIP-55 form. Probably the right thing to do is to convert the rpc result back to canonical before comparing against the fixture.

voith:
@carver Hopefully teats are fixed now

trinity/rpc/format.py
```diff
@@ -37,18 +42,22 @@
...
-        s=hex(transaction.s),
-        v=hex(transaction.v),
-    )
+    return {
+        'hash': encode_hex(transaction.hash),
+        'nonce': hex(transaction.nonce),
+        'gas': hex(transaction.gas),
+        'gasPrice': hex(transaction.gas_price),
+        'from': to_checksum_address(transaction.sender),
+        'to': compose(
```
voith: since to address can be empty during contract creation.

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
```diff
@@ -164,7 +170,10 @@
     'gasPrice': remove_leading_zeros,
     'value': remove_leading_zeros,
     'data': empty_to_0x,
-    'to': add_0x_prefix,
+    'to': compose(
+        apply_formatter_if(is_address, to_checksum_address),
```
voith: mypy is giving the following error:
```
trinity/rpc/format.py:52: error: Too few arguments for "apply_formatter_if"
```
I don't know how to handle this for curried function. Any hints? 
or `# type: ignore`?

trinity/rpc/format.py
```diff
@@ -37,18 +42,22 @@
...
+            apply_formatter_if(is_address, to_checksum_address),
```
voith: mypy is giving the following error:
```
trinity/rpc/format.py:52: error: Too few arguments for "apply_formatter_if"
```
I don't know how to handle this for curried function. Any hints? 
or `# type: ignore`?

voith: ok, this was easy to fix as curried `apply_formatter_if ` is not really needed here.

voith:
Ready for review: Tests are green now and I have addressed all comments.

PR #890, opened 2019-08-08
Cleanup Quickstart and start cookbook

cburgdorf:
### What was wrong?

1. The Quickstart is meant to teach how to *install and run* Trinity and then stop there, giving pointers into deeper parts of the documentation. It became cluttered with other things also burying valuable documentation in a place where people would not expect it.

2. We haven't had a good place to put small recipes which is basically the reason why we just stuffed more and more things into the Quickstart.

3. Some of the grammar used "You" which we typically try to avoid

### How was it fixed?

- I introduced a "Cookbook" section so that the Trinity docs are now structured as follows:

API = Focused on explaining the APIs, mostly generated from source
Cookbook = Short recipes/snippets a la "How do I ..."
Guides = Tutorial style articles

-  Ended the Quickstart right after the point where Trinity is running and syncing and introduced a "Whats next?" section with pointers to guides and the cookbook.

- Rewrote some sentences to avoid "You"

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/cookbook.rst
+++ b/docs/cookbook.rst
@@ -1,0 +1,101 @@
+Cookbook
+========
+
+The Cookbook is a collection of simple recipes that demonstrate good practices to accomplish
+common tasks. The examples are usually short answers to simple "How do I..." questions that go
+beyond simple API descriptions but also don't need a full guide to become clear.
+
+
+.. _cookbook_recipe_running_as_a_light_client:
+
+Running as a light client
+-------------------------
+
+.. warning::
+
+    It may take a **very** long time for Trinity to find an LES node with open
+    slots.  This is not a bug with trinity, but rather a shortage of nodes
+    serving LES. Everyone should consider running their own LES server to improve
+    the health of the network.
+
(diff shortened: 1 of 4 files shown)


docs/cookbook.rst
```diff
@@ -0,0 +1,100 @@
...
+.. code:: sh
+
+  trinity attach
+
+Now that Trinity runs in an interactive shell mode, let's try to get some information about the
+latest block by calling ``w3.eth.getBlock('latest')``.
+
+.. code:: sh
+
+  In [9]: w3.eth.getBlock('latest')
```
pipermerriam: This might be better using `getBlock(0)` since that would be easier to actually doctest.

cburgdorf: That code was just moved but I changed it nonetheless. I'm not sure if there's an easy way to doctest this (since it involves a `trinity attach` to a running instance) but at least `getBlock(0)` gives us a deterministic result.

Issue #893, opened 2019-08-08
The eth_call RPC works during Beam Sync


lithp:
Some components:
- `eth_call` checks whether we're in beam-sync mode. If we are:
- `eth_call` returns a useful error if you specify a block which is before when we started syncing and also more than [128 blocks](https://github.com/ethereum/go-ethereum/blob/c9cdf144d55f6bbd75314e812eeefc18d7e8c40e/core/blockchain.go#L81) behind the tip.
- `eth_call` does what it used to, but it catches `MissingAccountTrieNode` exceptions and fires off `CollectMissingAccount` events, once the syncer responds the RPC server retries execution.

What's a good way to test this?

lithp:
Solved by #975

PR #894, opened 2019-08-09
getBalance leans on the beam syncer if data is missing

lithp:
### What was wrong?

Beginning of solution to #893, `eth_getBalance` works during beam sync.

### Future work:

- This also needs to intercept `MissinBytecode` and `MissingStorateTrieNode`.
- This needs to support and be tested with more methods than just `eth_getBalance`
- This needs to generate a friendly error if you ask for a state so old the network won't have it (that's also from before we started beam syncing)
- This needs to return a nice error if the beam syncer can't fetch the state (it shouldn't hang forever)
- It'd be nice to include an integration test that actually runs `RPCServer` and beam sync and verifies that they do the intended thing.

#### Cute Animal Picture

marmot

### DIFF
--- a/trinity/rpc/main.py
+++ b/trinity/rpc/main.py
@@ -17,6 +17,9 @@ from eth_utils import (
 from trinity.rpc.modules import (
     BaseRPCModule,
 )
+from trinity.rpc.retry import (
+    execute_with_retries,
+)
 
 REQUIRED_REQUEST_KEYS = {
     'id',
@@ -53,7 +56,7 @@ class RPCServer:
     This "server" accepts json strings requests and returns the appropriate json string response,
     meeting the protocol for JSON-RPC defined here: https://github.com/ethereum/wiki/wiki/JSON-RPC
 
-    The key entry point for all requests is :meth:`RPCServer.request`, which
+    The key entry point for all requests is :meth:`RPCServer.execute`, which
     then proxies to the appropriate method. For example, see
     :meth:`RPCServer.eth_getBlockByHash`.
     """
@@ -62,6 +65,7 @@ class RPCServer:
     def __init__(self,
                  modules: Sequence[BaseRPCModule],
                  event_bus: EndpointAPI=None) -> None:
+        self.event_bus = event_bus
         self.modules: Dict[str, BaseRPCModule] = {}
 
         for module in modules:
@@ -108,7 +112,9 @@ class RPCServer:
 
             method = self._lookup_method(request['method'])
             params = request.get('params', [])
-            result = await method(*params)
+            result = await execute_with_retries(
+                self.event_bus, method, params
+            )
 
             if request['method'] == 'evm_resetToGenesisFixture':
                 result = True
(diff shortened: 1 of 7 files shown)


trinity/rpc/main.py
```diff
@@ -53,7 +57,7 @@ class RPCServer:
     This "server" accepts json strings requests and returns the appropriate json string response,
     meeting the protocol for JSON-RPC defined here: https://github.com/ethereum/wiki/wiki/JSON-RPC
 
-    The key entry point for all requests is :meth:`RPCServer.request`, which
+    The key entry point for all requests is :meth:`RPCServer.execute`, which
```
carver: :heart:

trinity/rpc/retry.py
```diff
@@ -0,0 +1,52 @@
...
+    CollectMissingAccount,
+)
+
+
+Func = Callable[..., Any]
+Meth = TypeVar('Meth', bound=Func)
+
+
+def retryable(func: Meth) -> Meth:
+    setattr(func, 'retryable', True)
```
carver: Seems like it would be better for this attribute to be more unique, and implied private, like: `_rpc_retryable`.  (Also, maybe moved to a little local `RETRYABLE_ATTRIBUTE_NAME`).

trinity/rpc/retry.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def is_retryable(func: Func) -> bool:
+    return getattr(func, 'retryable', False)
+
+
+def should_execute_with_retries(event_bus: EndpointAPI, func: Func) -> bool:
+    return all((
+        event_bus.is_any_endpoint_subscribed_to(CollectMissingAccount),
+        is_retryable(func),
```
carver: Maybe flip this order, since I expect `is_retryable()` to be faster to resolve.

trinity/rpc/retry.py
```diff
@@ -0,0 +1,52 @@
...
+    ))
+
+
+async def execute_with_retries(event_bus: EndpointAPI, func: Func, params: Any) -> None:
+    assert should_execute_with_retries(event_bus, func)
```
carver: Hm, if we're paranoid enough that this will be launched without checking first, maybe we should change this to be `maybe_execute_with_retries`, that checks `should_execute_with_retries()` first. Like:

```py
async def maybe_execute_with_retries(event_bus: EndpointAPI, func: Func, params: Any) -> None:
    should_retry = should_execute_with_retries(event_bus, func)

    while True:
        try:
            return await func(*params)
        except MissingAccountTrieNode as exc:
            if not should_retry:
                raise

            await event_bus.request(CollectMissingAccount(
                exc.missing_node_hash,
                exc.address_hash,
                exc.state_root_hash,
                urgent=True,
            ))
```

Then you could drop the `if` check in trinity/rpc/main

trinity/rpc/retry.py
```diff
@@ -0,0 +1,52 @@
...
+    while True:
```
carver: A thing that's saved us a few times is to avoid the infinite loop:

```py
TOO_MANY_RETRIES = 100_000  # or whatever makes sense

for attempt_number in range(TOO_MANY_RETRIES):
    ...
else:
    raise Exception(f"Failed to get a result after {TOO_MANY_RETRIES} attempts")
```

Issue #871, opened 2019-08-02
IncompleteReadError comes out of Transport/Multiplexer


pipermerriam:
### What is wrong?

```
   ERROR  08-02 16:59:44               ETHPeer  Unexpected error in ETHPeer <Node(0xebbc1d05fec4e7e859fdea4f7ba8a4fe408eb289f74c4e7a8c2d35eaab60677f1bacd434cc2f151f9028c516d70b568d4d875af4f98399e4e6ac9f836b494a1b@178.62.196.234:30303)>, exiting
Traceback (most recent call last):
  File "/home/piper/projects/trinity/p2p/multiplexer.py", line 276, in multiplex
    yield
  File "/home/piper/projects/trinity/p2p/peer.py", line 371, in _run
    await self.cancellation()
  File "/home/piper/projects/trinity/p2p/service.py", line 344, in cancellation
    await self.wait(self.events.cancelled.wait())
  File "/home/piper/projects/trinity/p2p/cancellable.py", line 20, in wait
    return await self.wait_first(awaitable, token=token, timeout=timeout)
  File "/home/piper/projects/trinity/p2p/cancellable.py", line 42, in wait_first
    return await token_chain.cancellable_wait(*awaitables, timeout=timeout)
  File "/home/piper/python-environments/trinity/lib/python3.6/site-packages/cancel_token/token.py", line 150, in cancellable_wait
    "Cancellation requested by {} token".format(self.triggered_token)
cancel_token.exceptions.OperationCancelled: Cancellation requested by ETHPeer:ETHPeerPool:FullServer:FullNode token

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/piper/projects/trinity/p2p/service.py", line 112, in run
    await self._run()
  File "/home/piper/projects/trinity/p2p/peer.py", line 371, in _run
    await self.cancellation()
  File "/home/piper/python-environments/trinity/lib/python3.6/site-packages/async_generator/_util.py", line 53, in __aexit__
    await self._agen.athrow(type, value, traceback)
  File "/home/piper/projects/trinity/p2p/multiplexer.py", line 281, in multiplex
    fut.result()
  File "/home/piper/projects/trinity/p2p/multiplexer.py", line 300, in _do_multiplexing
    async for protocol, cmd, msg in msg_stream:
  File "/home/piper/projects/trinity/p2p/cancellable.py", line 66, in wait_iter
    timeout=timeout,
[...]

PR #895, opened 2019-08-09
Handling for PeerConnectionLost during multiplexing

pipermerriam:
fixes https://github.com/ethereum/trinity/issues/871

### What was wrong?

There are `PeerConnectionLost` exceptions coming out of the multiplexer that aren't being handled.

### How was it fixed?

Add error handling for them.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -38,6 +38,7 @@ from p2p.disconnect import DisconnectReason
 from p2p.exceptions import (
     HandshakeFailure,
     MalformedMessage,
+    PeerConnectionLost,
     TooManyPeersFailure,
     UnexpectedMessage,
     UnknownProtocolCommand,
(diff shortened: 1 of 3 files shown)


p2p/peer.py
```diff
@@ -366,10 +367,15 @@ def get_protocol_command_for(self, msg: bytes) -> CommandAPI:
...
         self.run_child_service(self.boot_manager)
-        async with multiplexer.multiplex():
-            self.run_daemon_task(self.handle_p2p_proto_stream(multiplexer))
-            self.run_daemon_task(self.handle_sub_proto_stream(multiplexer))
-            await self.cancellation()
+        try:
+            async with multiplexer.multiplex():
+                self.run_daemon_task(self.handle_p2p_proto_stream(multiplexer))
+                self.run_daemon_task(self.handle_sub_proto_stream(multiplexer))
+                #
```
lithp: Was adding this line intentional?

p2p/multiplexer.py
```diff
@@ -296,6 +297,9 @@ def stream_protocol_messages(self,
             # wait for the multiplexing to actually start
             try:
                 yield
+            except PeerConnectionLost as err:
+                self.close()
```
lithp: This is a little scary for me, I can understand wanting to trigger the cancel token but `Multiplexer.close` also closes the transport and I'm not sure that's something every user of `Multiplexer` will be expecting. Will every transport be okay with being closed twice? `BasePeer._cleanup()` also attempts to close the transport, it seems like those two classes disagree over who has ownership over it.

pipermerriam: Good catch.  I think I'm gonna re-work this to move the handling of this outside of the multiplexer and allow whatever code is operating the multiplexer to be in charge of it's lifecycle.

PR #898, opened 2019-08-09
No need to use unbound methods during Beam Sync

### LINKED ISSUE #716: Document why PausingVMState uses unbound method
### What is wrong?

`trinity.sync.beam.importer.PausingVMState._pause_on_missing_data()` uses an unusual unbound approach. I originally used a bound-method approach, then switched  to fix something, but it's possible it was a red-herring fix. Now I've lost track of the rationale for why I did it that way.

> Why use the unbound super-class method? I remember it fixing a bug, but
>  it doesn't make sense why we can't just use:
>    - pass super().method as bound_vm_method
>    - call bound_vm_method(*args, **kwargs)
>  I have a vague memory of it somehow not catching exceptions, but that doesn't make
>  sense either. I think it con be switched to use the above approach, but let's get
>  the tests green again first and then come back to it.

### How can it be fixed

Figure out if/why it's necessary. Switch back to bound method if it's unnecessary (be sure to run the full 130-block suite, and some mainnet smoke tests). If it is necessary, document why.

carver:
### What was wrong?

Fixes #716 

### How was it fixed?

Used the more natural bound method, when wrapping in the beam importer.

Ran a full-sweep test on the `beam_syncer` test (the whole `range(1, 130)`). I also ran a smoke test on mainnet for ~15 min.

### To-Do

- [x] Clean up commit history

- [x] ~~Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)~~

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/importer.py
+++ b/trinity/sync/beam/importer.py
@@ -180,7 +180,7 @@ def pausing_vm_decorator(
 
         def _pause_on_missing_data(
                 self,
-                unbound_vm_method: Callable[['PausingVMState', Any], TVMFuncReturn],
+                vm_method: Callable[[Any], TVMFuncReturn],
                 *args: Any,
                 **kwargs: Any) -> TVMFuncReturn:
             """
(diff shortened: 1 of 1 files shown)


carver:
This one is a trivial change. I'll fix it with a follow-up if some feedback comes in after the merge.

PR #899, opened 2019-08-09
Beam Sync: Speculative Execution

carver:
### What was wrong?

Beam Sync is still too slow, when connected to peers with real-world. There is another angle for execution.

### How was it fixed?

Peek ahead at individual transactions within a block. Execute them as if they were the only transaction in the block. Often, transactions don't depend much on the others bundled before them. So it gives us a chance to look for many more nodes in parallel, without waiting to lag so far behind the chain tip.

I started running into I/O contention, so I had to limit the parallel look-ahead cap. When I tried to run all N transactions at the same time (say 150), it really choked the DB, so I limited it to a total of 40 concurrently running.

I also turned on parallel/preview execution for *all* blocks, and started ignoring the lag. This means we get to look ahead at transactions even for the current block that we are executing.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/importer.py
+++ b/trinity/sync/beam/importer.py
@@ -1,4 +1,6 @@
 import asyncio
+from concurrent.futures import ThreadPoolExecutor
+from operator import attrgetter
 from typing import (
     Any,
     Callable,
@@ -27,6 +29,9 @@ from eth_typing import (
 from eth_utils import (
     ValidationError,
 )
+from eth_utils.toolz import (
+    groupby,
+)
 
 from lahja import EndpointAPI
 from lahja.common import BroadcastConfig
@@ -36,7 +41,10 @@ from p2p.service import BaseService
 from trinity._utils.timer import Timer
 from trinity.chains.base import BaseAsyncChain
 from trinity.chains.full import FullChain
-from trinity.sync.beam.constants import NUM_PREVIEW_SHARDS
+from trinity.sync.beam.constants import (
+    MAX_SPECULATIVE_EXECUTIONS_PER_PROCESS,
+    NUM_PREVIEW_SHARDS,
+)
 from trinity.sync.common.events import (
     CollectMissingAccount,
     CollectMissingBytecode,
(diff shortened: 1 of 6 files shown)


trinity/sync/beam/importer.py
```diff
@@ -467,5 +523,27 @@ def __init__(
...
+            # Speculative Execution:
+            # Execute each transaction individually, effectively assuming that it was not
+            #   affected by any transactions before it in the block. This is often true,
+            #   so it helps speed up the search for data.
+            for index, transaction in enumerate(event.transactions):
+                if index == 0:
+                    # There is no benefit to sepeculatively executing the first transaction
+                    continue
+                elif transaction.gas <= transaction.get_intrinsic_gas():
+                    # There is not enough gas for any EVM code to run, so skip the speculation
```
carver: TODO: this doesn't seem to be working as expected

trinity/sync/full/constants.py
```diff
@@ -22,13 +22,9 @@
...
-#   - how many blocks finish early/quickly, ~half, which doubles capacity (~2)
-#   So we multiply all these together to get 16 parallel executions to permit.
-#   The first block in the queue doesn't get previewed, which brings us to 17.
-BLOCK_IMPORT_QUEUE_SIZE = 17
+# This is also used during Beam sync, to limit how many previews are emitted at once
+# If you increase the number too high, then your I/O latency can skyrocket,
+#   causing a massive slowdown.
+# Every block gets previewed, and a block only enters the queue if another block import
+#   is active. So a queue size of 3 means that up to 4 previews are happening at once.
+BLOCK_IMPORT_QUEUE_SIZE = 3
```
carver: I am experimenting with increasing this to 7, when connected to remote peers, which seems to be going okay.

carver:
Still not having much luck with remote peers. Part of the problem is that because we're only talking to one peer, we don't get much/any data on which one is the fastest right now. That problem will be fixed by adding a background thread that backfills data. Then we'll gave stats on all the peers.

PR #902, opened 2019-08-10
Cleanup logging output

pipermerriam:
### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -282,7 +282,7 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
         Appart from adding it to our list of connected nodes and adding each of our subscriber's
         to the peer, we also add the given messages to our subscriber's queues.
         """
-        self.logger.info('Adding %s to pool', peer)
+        self.logger.debug('Adding %s to pool', peer)
         self.connected_nodes[peer.remote] = peer
         peer.add_finished_callback(self._peer_finished)
         for subscriber in self._subscribers:
@@ -444,7 +444,7 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
         """
         peer = cast(BasePeer, peer)
         if peer.remote in self.connected_nodes:
-            self.logger.info("%s finished[%s], removing from pool", peer, peer.disconnect_reason)
+            self.logger.debug("%s finished[%s], removing from pool", peer, peer.disconnect_reason)
             self.connected_nodes.pop(peer.remote)
         else:
             self.logger.warning(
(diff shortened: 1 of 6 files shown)


p2p/peer_pool.py
```diff
@@ -468,8 +468,8 @@ def _peer_finished(self, peer: BaseService) -> None:
         while self.is_operational:
             inbound_peers = len(
                 [peer for peer in self.connected_nodes.values() if peer.inbound])
-            self.logger.info("Connected peers: %d inbound, %d outbound",
-                             inbound_peers, (len(self.connected_nodes) - inbound_peers))
+            self.logger.debug("Connected peers: %d inbound, %d outbound",
```
carver: I kind of like this one in `info`.

p2p/peer_pool.py
```diff
@@ -282,7 +282,7 @@ def _add_peer(self,
         Appart from adding it to our list of connected nodes and adding each of our subscriber's
         to the peer, we also add the given messages to our subscriber's queues.
         """
-        self.logger.info('Adding %s to pool', peer)
+        self.logger.debug('Adding %s to pool', peer)
```
carver: :+1: Although I have a local patch that I like even better, because it's still nice to track peers:

```py
        self.run_task(self._log_peer_added(peer))

    async def _log_peer_added(self, peer):
        await self.sleep(3)
        if peer.is_running:
            self.logger.info('Added %s to pool', peer)
```

This locally gets rid of the flutter but still shows me the important steps.

trinity/sync/beam/chain.py
```diff
@@ -463,7 +463,7 @@ def __init__(
     async def import_block(
             self,
             block: BaseBlock) -> Tuple[BaseBlock, Tuple[BaseBlock, ...], Tuple[BaseBlock, ...]]:
-        self.logger.info("Beam importing %s (%d txns) ...", block.header, len(block.transactions))
+        self.logger.debug("Beam importing %s (%d txns) ...", block.header, len(block.transactions))
```
carver: I'll patch this back to `info` locally, for sure. It's super helpful to immediately see whether you're stuck waiting for the next block to arrive or are in the middle of processing one.

trinity/sync/beam/state.py
```diff
@@ -524,5 +524,5 @@ def register_peer(self, peer: BasePeer) -> None:
             msg += "reqs=%d  " % (self._total_requests)
             msg += "pred_reqs=%d  " % (self._predictive_only_requests)
             msg += "timeouts=%d" % self._total_timeouts
-            self.logger.info("Beam-Sync: %s", msg)
+            self.logger.debug("Beam-Sync: %s", msg)
```
carver: Yeah, I'll definitely patch this back locally, too.

p2p/peer_pool.py
cburgdorf: Yep, I also think this would be nice to keep in `info`

trinity/sync/beam/chain.py
cburgdorf: Just noting that you can run trinity with `-l trinity.sync=DEBUG2` to just whitelist all of the `trinity.sync` `>=DEBUG2` logs.

carver: Yeah, but I don't want to scan through *all* the debug (or debug2) stuff in all of sync. I like having just a few things show up that are important. FWIW, I agree that this should go back to debug when beam sync is regularly working. :)

pipermerriam:
I'm going to hold off on merging since it seems there are some differing thoughts on changing these levels.  I'm not in a hurry for this but it's something that I'll want to merge sometime between MVP and beta.

PR #905, opened 2019-08-12
Minor discv5 typing and naming changes

jannikluhn:
### Notes

contains commits of #863, will rebase once that one is merged.

### What was wrong?

- in the discv5 code we use `bytes` where more specific types would be appropriate
- the parameter `initiator_key` for preparing or decrypting auth tag is incorrectly named most of the time (it's only true for auth header packets which are sent only by the handshake initiator -- in all other cases it could also be the recipient)
- `sender` and `receiver` in in incoming/outgoing packets/datagrams refer to the peer's endpoint, even though usually we'll want to refer to peer's by their node id

### How was it fixed?

- Add `IDNonce`, `NodeID`, and `Tag` type
- rename `initiator_key` to `key` for all auth tag encryption and decryption functions
- rename `sender/receiver` to `sender/receiver_endpoint`

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


p2p/discv5/channel_services.py
```diff
@@ -33,28 +33,28 @@
 # Data structures
 #
 class Endpoint(NamedTuple):
-    ip_address: bytes
+    ip_address: str
```
pipermerriam: Curious if you looked into using `p2p.kademlia.Address` for this and whether we could re-use that instead of using `Endpoint`?

p2p/discv5/channel_services.py
```diff
@@ -70,9 +70,9 @@ class OutgoingPacket(NamedTuple):
 
     async with incoming_datagram_send_channel:
         while manager.is_running:
-            datagram, (ip_address, port) = await socket.recvfrom(DATAGRAM_BUFFER_SIZE)
-            logger.debug(f"Received {len(datagram)} bytes from {(ip_address, port)}")
-            incoming_datagram = IncomingDatagram(datagram, Endpoint(ip_address, port))
+            datagram, endpoint = await socket.recvfrom(DATAGRAM_BUFFER_SIZE)
```
pipermerriam: looks like `endpoint` here will just be a regular `tuple` instead of `Endpoint`.  Is this intentional?

p2p/discv5/channel_services.py
jannikluhn: I haven't until now, but I'd like to keep `Endpoint` for two reasons:

- We don't need the TCP port in `Address`
- I like the immutability of `Endpoint`, so we don't have to think if we reuse it (e.g. when creating the response to an incoming message)

PR #914, opened 2019-08-13
Fix eth2-fixtures CI cache

hwwhww:
# What was wrong?

* Per @protolambda 's reminder on eth2 discord, checking if our CI setting is correct...
* Realized that we didn't cache the fixture files properly!!!!
* Since we are uploading `tar.gz` (120 MB) in CI, most overheads were for uncompressing the file.

### How was it fixed?
* Include `.circleci/config.yml` to the cache key
* Check if `.eth2-fixtures/tests` exsits; if not, download the `tar.gz` file
* Add `.eth2-fixtures` to the cache pathes

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

cute-animal-pictures-animals-eating-watermelon-013

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ geth_steps: &geth_steps
     - checkout
     - restore_cache:
         keys:
-          - cache-v2-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}
+          - cache-v2-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}-{{ checksum "./.circleci/build_geth.sh" }}
     - run:
         name: install libsnappy-dev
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -121,8 +121,10 @@ eth2_fixtures: &eth2_fixtures
     - run:
         name: download the required yaml files
         command: |
-          wget -c https://github.com/hwwhww/eth2.0-spec-tests/releases/download/v0.8.1b2/archive.tar.gz
-          tar zxvf archive.tar.gz -C ./eth2-fixtures
+          if [ ! -d "./eth2-fixtures/tests" ]; then
```
pipermerriam: maybe place this in a separate file and run it as a bash script so that we can compute the cache based on that file rather than the entire circle CI config file.

hwwhww: Good point! I also cleaned up `geth_steps` in https://github.com/ethereum/trinity/pull/914/commits/ce56f74094f164adc59eeaa10d1f9fb0cddf8381.

PR #918, opened 2019-08-13
[WIP] Prefer low latency in beam sync

carver:
### What was wrong?

During Beam Sync, we care about the constant cost of a request much more than the total bandwidth of nodes that we can power through a node. But we're currently aiming at the peer with the highest throughput. (and we don't even know what the constant cost of a request is)

### How was it fixed?

Just a hack job for now:

1. Added in some stats to track and infer the constant cost of a request (called latency in the printout)
2. Also tracking the incremental cost of adding more items in a request (often fairly close to 0, though)
3. Prefer the lowest latency peers
4. Presume the latency is 0 at first, so that we always try sending requests to new peers, and help determine quickly if they have better latency stats
5. At first I was just basing this on messages per second, which is probably a useful metric also, but the latency metric felt more tangible and accessible, and ultimately correct.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/_utils/regression.py
+++ b/trinity/_utils/regression.py
@@ -1,0 +1,23 @@
+from typing import Tuple
+
+
+def linear_regression(points: Tuple[Tuple[float, float], ...]) -> Tuple[float, float]:
+    """
+    :return: the slope and intercept of the linear regression
+
+    Not convinced? Check:
+    https://www.khanacademy.org/math/statistics-probability/describing-relationships-quantitative-data/more-on-regression/v/regression-line-example  # noqa: E501
+    """
+    xpoints = tuple(point[0] for point in points)
+    ypoints = tuple(point[1] for point in points)
+    num_points = float(len(points))
+
+    xmean = sum(xpoints) / num_points
+    ymean = sum(ypoints) / num_points
+    product_mean = sum(x * y for x, y in points) / num_points
+    x_square_mean = sum(x * x for x in xpoints) / num_points
+
+    slope = (xmean * ymean - product_mean) / ((xmean * xmean) - x_square_mean)
+    intercept = ymean - slope * xmean
+
+    return slope, intercept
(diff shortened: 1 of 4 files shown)


trinity/sync/common/peers.py
```diff
@@ -28,28 +28,40 @@ class WaitingPeers(Generic[TChainPeer]):
     """
     _waiting_peers: 'PriorityQueue[SortableTask[TChainPeer]]'
 
-    def __init__(self, response_command_type: Type[Command]) -> None:
+    def __init__(self, response_command_type: Type[Command], latency_focused: bool=False) -> None:
```
carver: Maybe instead of adding `latency_focused`, there could be a new `Callable[[Tracker], float]`, and the default is:
```py
lambda tracker: -1 * tracker.items_per_second_ema.value
```

trinity/_utils/regression.py
```diff
@@ -0,0 +1,23 @@
...
+
+def linear_regression(points: Tuple[Tuple[float, float], ...]) -> Tuple[float, float]:
+    """
+    :return: the slope and intercept of the linear regression
+
+    Not convinced? Check:
+    https://www.khanacademy.org/math/statistics-probability/describing-relationships-quantitative-data/more-on-regression/v/regression-line-example  # noqa: E501
+    """
+    xpoints = tuple(point[0] for point in points)
+    ypoints = tuple(point[1] for point in points)
```
carver: Ah, maybe:
```py
xpoints, ypoints = zip(*points)

carver:
Eh, I'll probably do some different version of this. RTT seems basically just as effective, since the marginal cost of serving each item is negligible. Often, it shows up as negative, in fact, which just shows you that it's too close to zero to be measurable.

So RTT should be sufficient. We could play some games by optimizing to `(mean - std_dev * 2)`, which kind of gives you a multi-armed-bandit approach of explore vs exploit, but I don't think that's necessary if we can spread the backfill requests around enough to explore all peers naturally.

PR #920, opened 2019-08-14
Discv5 Handshake

jannikluhn:
This adds two classes that encapsulate the steps necessary to do a handshake. They are fully synchronous and need a service to use them which will come in a follow up PR. The general flow will look like this:

1) Initialize `HandshakeInitiator` (with the initial message we want to send to the peer) or `HandshakeReceiver` (with the `auth_tag` in the `AuthTagPacket` we've received.
2) Send `handshake_participant.first_packet_to_send` to the peer
3) Check all packets received from this point on with `handshake_participant.is_response_packet`
4) If the check succeeds, call `handshake_participant.complete_handshake` to get a `HandshakeResult`
5) The result contains the `SessionKeys`. Additionally, for the handshake initiator, it contains a `AuthHeaderPacket` that we have to send to the peer so that they can complete the handshake on their side. For the handshake recipient, it contains the initial message from the initiator. Optionally, it may also contain an `ENR` of the initiator (if they sent one).

For now I've kept the tests relatively light (only the successful cases). I plan to add more in the future, but would like to wait a bit with that as I hope we'll get some official test vectors at some point.

I've also renamed `pubkey` to `public_key` in the whole discv5 code for consistency with `private_key`, `initiator_key`, etc.

Supersedes #868.

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/handshake.py
+++ b/p2p/discv5/handshake.py
@@ -19,6 +19,7 @@ from p2p.exceptions import (
     HandshakeFailure,
 )
 
+from p2p.discv5.abc import HandshakeParticipantAPI
 from p2p.discv5.enr import ENR
 from p2p.discv5.identity_schemes import IdentityScheme
 from p2p.discv5.messages import BaseMessage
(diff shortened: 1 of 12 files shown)


p2p/discv5/handshake.py
```diff
@@ -0,0 +1,351 @@
...
+    enr: Optional[ENR]
+    message: Optional[BaseMessage]
+    auth_header_packet: Optional[AuthHeaderPacket]
+
+
+class BaseHandshakeParticipant(ABC):
+    def __init__(self,
+                 is_initiator: bool,
+                 our_private_key: bytes,
+                 our_enr: ENR,
```
pipermerriam: I'd advocate for using `local` / `remote` terminology.

p2p/discv5/handshake.py
```diff
@@ -0,0 +1,351 @@
...
+                packet.tag,
+                self.our_node_id,
+            ) == self.their_node_id
+        )
+
+    def complete_handshake(self, response_packet: Packet) -> HandshakeResult:
+        if not self.is_response_packet(response_packet):
+            raise ValueError("Packet is not the expected response packet")
+        if not isinstance(response_packet, AuthHeaderPacket):
+            raise Exception("Invariant: Only AuthHeader packets are valid responses")
```
pipermerriam: I think this is more appropriate as a `TypeError`

p2p/discv5/handshake.py
```diff
@@ -0,0 +1,351 @@
...
+        try:
+            id_nonce_signature, enr = auth_header_packet.decrypt_auth_response(auth_response_key)
+        except DecryptionError as error:
+            raise HandshakeFailure("Unable to decrypt auth response") from error
+        except ValidationError as error:
+            raise HandshakeFailure("Invalid auth response content") from error
+
+        # validate ENR if present
+        if enr is None:
+            if self.their_enr is None:
```
pipermerriam: This is standing out to me as *maybe* a sign that something is off in the architecture but only because we have to do this value checking and I see multiple APIs that access `self.their_enr.some_property` which don't do a check for whether the ENR is present meaning all of those APIs can throw `AttributeError` if accessed from the wrong location.

I don't have a specific suggestion at this stage but I would be looking for a way that we could get some stronger guarantees in place about this value.

p2p/discv5/packets.py
```diff
@@ -559,3 +562,18 @@ def _decrypt_message(key: AES128Key,
         raise ValidationError("Encrypted message does not contain valid RLP") from error
 
     return message
+
+
+#
+# Random packet data
+#
+def get_random_encrypted_data() -> bytes:
+    return os.urandom(RANDOM_ENCRYPTED_DATA_SIZE)
```
pipermerriam: suggest we use `secrets.token_bytes`

pipermerriam:
Can you do a scan for places that you compare values that should use constant time comparisons.  The `secrets` module has a constant time comparison function that we can use for this.

p2p/discv5/handshake.py
jannikluhn: Yeah, I agree, that part is confusing, both in the code and in the spec. But if correctly implemented, there should be no issue.

The situation is that as an initator we always know the peer's ENR as that's where we get their ip address from. As a handshake recipient we may or may not know the peer's ENR. So in `HandshakeInitiator` we can assume that we have an `ENR`, only in `HandshakeRecipient` we cannot.

I'll go through it one more time, but think it should be ok right now (i.e. at every place where we use it we do a `if their_enr is not None` check prior).

p2p/discv5/handshake.py
jannikluhn: :+1: sounds much more professional

jannikluhn:
As suggested in the ENR DB PR, I've removed `enr_db` from `BaseHandshakeParticipant`. It's now defined as `ENR` in `HandshakeInitiator` and `Optional[ENR]` in `HandshakeRecipient which is more clear than before.

I've also added a `HandshakeParticipantAPI` in `p2p.discv5.api` that defines the shared interface.

> Can you do a scan for places that you compare values that should use constant time comparisons. The secrets module has a constant time comparison function that we can use for this.

I did not find any such places except for the one you pointed out (in the handshake, there's one other istance where we compare with `==`, but it's for node ids which are not private).

So this should be ready for merge now. However, some eth1 tests are failing right now in CI. Pretty sure it's unrelated.

Issue #892, opened 2019-08-08
Allow beam sync to sync from a checkpoint header provided via CLI flag


cburgdorf:
### What is wrong?

Before beam sync can start of doing its "beam magic", it still needs to download the chain of headers. While that's not the end of the world, it would be much nicer if the user could provide a checkpoint for a header that beam sync would trust (to start with) and build up on top.

### How can it be fixed

1. Add support for a CLI flag `--beam-sync-checkpoint <block-hash> <td>`

2. Make sure beam sync can build on top of this checkpoint

3. It should still fill up the headers in the background but same goes for blocks so not sure if this needs to be part of this mechanism.

**Nice to have:**

- Add support for a CLI flag that delegates reading a recent block / td from a website and use that as a checkpoint (e.g. `--beam-sync-checkpoint-source https://etherscan.io/`)

pipermerriam:
We've used URI formats for things like this nicely.  dunno how you or others feel about this but.

- `block://<block-hash>/<total-difficulty>`
- `etherscan://<block-number-or-hash-or-latest>`
- `infura://<block-number-or-hash-or-latest>`

For the `block-number-or-hash-or-latest` my thought was that we could accept it as either the hash, the integert number, or the string `latest` which would do something like auto-computing the checkpoint at something like `HEAD - Nblocks`

cburgdorf:
@carver @pipermerriam am I correct to assume that we need to do something about persist_header_chain in py-evm?

https://github.com/ethereum/py-evm/blob/d2ccc1e8a9be1cb92cb15a8bb117bb6a1972636e/eth/db/header.py#L194-L196

Because, on the evm level, persisting headers with missing parents is currently not possible. We could relax this but I'm wondering if you had already thought about this and have alternative ideas?

cburgdorf:
But I guess we can not just relax it, we would probably need a different API to persist such "orphaned" headers which would also take the score into account.

Something like:

```python
    def persist_orphan_header(self, header: BlockHeaderAPI, score: int) -> None:
```

That's because in the regular case we would calculate the score based on the parent which we can't in this case. So, I'm thinking we just have to accept whatever is passed as `score` in another parameter. Does that make sense?

cburgdorf:
This is turning into a public brain dump for myself. I figured out that even with a new API `persist_orphan_header` we'll still be in trouble because as soon as we start persisting blocks on top of the orphaned block we'll run into an exception here:

[...]

PR #921, opened 2019-08-14
Allow beam sync to start off from a trusted checkpoint

cburgdorf:
### What was wrong?

Relates to #892 

Before beam sync can start of doing its "beam magic", it still needs to download the chain of headers. While that's not the end of the world, it would be much nicer if the user could provide a checkpoint for a header that beam sync would trust (to start with) and build up on top.

This allows beam sync to get to work *immediately* and fill up missing headers (and eventually also blocks) in the background.

### How was it fixed?

1. Add a `SyncLaunchStrategyAPI` with two implementations:

-  created a new class `Checkpoint` that represents a blockhash, a blocknumber as well as a score
- `FromGenesisLaunchStrategy`: Download the full header chain first (same as today)
- `FromCheckpointLaunchStrategy`: Launch from a given checkpoint

2. Added a new parameter `--beam-from-checkpoint` that takes an URI as a form of micro DSL.

This is following what @pipermerriam proposed in #892 but it's been a bit controversial and I'm open to other ideas. However, here's my attempt to defend it:

For the manual checkpoints (e.g. not retrieving one through the etherscan API) there are three parts to a checkpoint, the blockhash, the blocknumber and the score. Creating three different command line parameters for this:

- A. Is hard to remember (I think it's easier to get used to some form of micro DSL)
- B. Makes Trinitys invocation command much longer
- C. Rips apart three things that essentially need to be supplied together

### Seeing it in action:

1. `rm -rf /home/cburgdorf/.local/share/trinity/mainnet/chain-eth1/`

2. `trinity --sync-mode beam  --beam-from-checkpoint="eth://block/byhash/0x113f05289c685eb5b87d433c3e09ec2bfa51d6472cc37108d03b0113b11e3080?number=8360824&score=11,487,662,456,884,849,810,705"` 

Then observe:

```
INFO  08-16 16:54:26       BeamSyncService  Starting beam-sync; using checkpoint: <Checkpoint block hash=#113f..3080 block number=8360824 score=11487662456884849810705>
...
INFO  08-16 16:54:59     HeaderOnlyPersist  Imported 192 headers in 0.21 seconds, new head: <BlockHeader #8361016 b7e740c2>
    INFO  08-16 16:55:01     HeaderOnlyPersist  Imported 388 headers in 0.06 seconds, new head: <BlockHeader #8361404 44c602c9>
    INFO  08-16 16:55:02     HeaderOnlyPersist  Caught up to skeleton peer. Switching to beam mode at <BlockHeader #8362170 e4e1ea7c>
    INFO  08-16 16:55:02            BeamSyncer  Downloading 8 block bodies for uncle validation, before <BlockHeader #8362170 e4e1ea7c>
    INFO  08-16 16:55:02            BeamSyncer  Getting recent block data for uncle validation
    INFO  08-16 16:55:06            BeamSyncer  Have all data needed for Beam validation, continuing...
    INFO  08-16 16:55:06  HeaderCheckpointSyncer  Choosing (<eth.rlp.headers.BlockHeader object at 0x7f6c8effca90>,) as checkpoint headers to sync from
    INFO  08-16 16:55:06        BeamDownloader  Starting beam state sync
    INFO  08-16 16:55:06        BeamDownloader  Beam-Sync: all=0  urgent=0  pred=0  all/sec=0  urgent/sec=0  reqs=0  pred_reqs=0  timeouts=0
    INFO  08-16 16:55:06           ETHPeerPool  Adding ETHPeer <Node(0x1fdd@91.195.52.166)> to pool
    INFO  08-16 16:55:06  RegularChainBodySyncer  Waited 0.6s for <BlockHeader #8362170 e4e1ea7c> body
    INFO  08-16 16:55:06     BeamBlockImporter  Beam importing <BlockHeader #8362170 e4e1ea7c> (69 txns) ...
    INFO  08-16 16:55:16        BeamDownloader  Beam-Sync: all=959  urgent=758  pred=201  all/sec=95  urgent/sec=75  reqs=44  pred_reqs=0  timeouts=1
    INFO  08-16 16:55:23           ETHPeerPool  Adding ETHPeer <Node(0x58dc@23.111.173.234)> to pool
    INFO  08-16 16:55:26        BeamDownloader  Beam-Sync: all=2943  urgent=800  pred=2143  all/sec=145  urgent/sec=39  reqs=86  pred_reqs=0  timeouts=1
    INFO  08-16 16:55:26           ETHPeerPool  Connected peers: 0 inbound, 7 outbound
    INFO  08-16 16:55:36        BeamDownloader  Beam-Sync: all=4693  urgent=838  pred=3855  all/sec=155  urgent/sec=27  reqs=124  pred_reqs=0  timeouts=1
    INFO  08-16 16:55:46        BeamDownloader  Beam-Sync: all=5761  urgent=890  pred=4871  all/sec=143  urgent/sec=22  reqs=176  pred_reqs=0  timeouts=1
    INFO  08-16 16:55:56        BeamDownloader  Beam-Sync: all=8190  urgent=926  pred=7264  all/sec=163  urgent/sec=18  reqs=216  pred_reqs=4  timeouts=1
    INFO  08-16 16:56:06        BeamDownloader  Beam-Sync: all=10329  urgent=967  pred=9362  all/sec=171  urgent/sec=16  reqs=257  pred_reqs=4  timeouts=1
    INFO  08-16 16:56:16        BeamDownloader  Beam-Sync: all=11746  urgent=1010  pred=10736  all/sec=167  urgent/sec=14  reqs=300  pred_reqs=4  timeouts=1
    INFO  08-16 16:56:26        BeamDownloader  Beam-Sync: all=12933  urgent=1051  pred=11882  all/sec=161  urgent/sec=13  reqs=341  pred_reqs=4  timeouts=1
    INFO  08-16 16:56:26           ETHPeerPool  Connected peers: 0 inbound, 7 outbound
    INFO  08-16 16:56:36        BeamDownloader  Beam-Sync: all=14168  urgent=1098  pred=13070  all/sec=156  urgent/sec=12  reqs=388  pred_reqs=4  timeouts=1
    INFO  08-16 16:56:46        BeamDownloader  Beam-Sync: all=15599  urgent=1137  pred=14462  all/sec=155  urgent/sec=11  reqs=427  pred_reqs=4  timeouts=1
    INFO  08-16 16:56:53           ETHPeerPool  Adding ETHPeer <Node(0xf9ae@203.238.15.78)> to pool
    INFO  08-16 16:56:56           ETHPeerPool  Adding ETHPeer <Node(0x0122@47.254.45.126)> to pool
    INFO  08-16 16:56:56           ETHPeerPool  ETHPeer <Node(0x0122@47.254.45.126)> finished[DisconnectReason.subprotocol_error], removing from pool
    INFO  08-16 16:56:56        BeamDownloader  Beam-Sync: all=17160  urgent=1179  pred=15981  all/sec=155  urgent/sec=10  reqs=469  pred_reqs=4  timeouts=1
    INFO  08-16 16:57:01  RegularChainBodySyncer  Imported block 8362170 (69 txs) in 114.37 seconds, with 2m7s lag


```

### To-Do

- [x] Hook it up to the command line (currently uses a hardcoded checkpoint)
- [x] polish local py-evm changes and open up Py-EVM PR
- [x] print out used checkpoint
- [x] switch to proper Py-EVM release
- [x] parse out seperators from block number / score
- [x] Get rid of FIXMEs
- [x] add a test that launches sync from a checkpoint
- [x] Clean up commit history
- [x] Add entry to the [release notes]

#### Cute Animal Picture

### DIFF
--- a/newsfragments/921.feature.rst
+++ b/newsfragments/921.feature.rst
@@ -1,0 +1,8 @@
+Allow beam sync to start from a trusted checkpoint.
+Specify a checkpoint via CLI parameter such as:
+
+``--beam-from-checkpoint="eth://block/byhash/<hash>?score=<score>"``
+
+When given, beam sync will use this as a checkpoint
+to avoid having to download the entire chain of headers
+first.
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@
 import os
 from setuptools import setup, find_packages
 
-PYEVM_DEPENDENCY = "py-evm==0.3.0a3"
+PYEVM_DEPENDENCY = "py-evm==0.3.0a5"  # noqa: E501
 
 
 deps = {
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -4,6 +4,7 @@ import uuid
 from eth.db.atomic import AtomicDB
 from eth.exceptions import HeaderNotFound
 from eth.vm.forks.petersburg import PetersburgVM
+from eth_utils import decode_hex
 from lahja import ConnectionConfig, AsyncioEndpoint
 from p2p.service import BaseService
 import pytest
@@ -16,6 +17,7 @@ from trinity.sync.beam.importer import (
 )
 from trinity.protocol.eth.sync import ETHHeaderChainSyncer
 from trinity.protocol.les.servers import LightRequestServer
+from trinity.sync.common.checkpoint import Checkpoint
 from trinity.sync.common.chain import (
     SimpleBlockImporter,
 )
@@ -146,6 +148,30 @@ async def test_skeleton_syncer(request, event_loop, event_bus, chaindb_fresh, ch
             assert head.state_root in chaindb_fresh.db
 
 
+@pytest.mark.asyncio
+async def test_beam_syncer_with_checkpoint(
+        request,
+        event_loop,
+        event_bus,
+        chaindb_fresh,
+        chaindb_churner):
+
+    checkpoint = Checkpoint(
+        block_hash=decode_hex('0x5b8d32e4aebda3da7bdf2f0588cb42256e2ed0c268efec71b38278df8488a263'),
+        score=55,
+    )
+
+    await test_beam_syncer(
+        request,
+        event_loop,
+        event_bus,
+        chaindb_fresh,
+        chaindb_churner,
+        beam_to_block=66,
+        checkpoint=checkpoint,
+    )
+
+
 # Identified tricky scenarios:
 # - 66: Missing an account trie node required for account deletion trie fixups,
 #       when "resuming" execution after completing all transactions
@@ -163,7 +189,8 @@ async def test_beam_syncer(
         event_bus,
         chaindb_fresh,
         chaindb_churner,
-        beam_to_block):
+        beam_to_block,
+        checkpoint=None):
 
     client_context = ChainContextFactory(headerdb__db=chaindb_fresh.db)
     server_context = ChainContextFactory(headerdb__db=chaindb_churner.db)
@@ -209,7 +236,8 @@ async def test_beam_syncer(
                 AsyncChainDB(chaindb_fresh.db),
                 client_peer_pool,
                 gatherer_endpoint,
-                beam_to_block,
+                force_beam_block_number=beam_to_block,
+                checkpoint=checkpoint,
             )
 
             client_peer.logger.info("%s is serving churner blocks", client_peer)
@@ -311,7 +339,7 @@ class FallbackTesting_RegularChainSyncer(BaseService):
     def __init__(self, chain, db, peer_pool, token=None) -> None:
         super().__init__(token=token)
         self._chain = chain
-        self._header_syncer = ETHHeaderChainSyncer(chain, db, peer_pool, self.cancel_token)
+        self._header_syncer = ETHHeaderChainSyncer(chain, db, peer_pool, token=self.cancel_token)
         self._single_header_syncer = self.HeaderSyncer_OnlyOne(self._header_syncer)
         self._paused_header_syncer = self.HeaderSyncer_PauseThenRest(self._header_syncer)
         self._draining_syncer = RegularChainBodySyncer(
--- a/trinity/plugins/builtin/syncer/cli.py
+++ b/trinity/plugins/builtin/syncer/cli.py
@@ -1,0 +1,81 @@
+import argparse
+from pathlib import PurePosixPath
+import re
+import urllib
+from typing import (
+    Any,
+)
+
+from eth_typing import (
+    Hash32,
+)
+
+from eth_utils import (
+    is_hex,
+    remove_0x_prefix,
+    decode_hex,
+    ValidationError,
+)
+
+from trinity.sync.common.checkpoint import Checkpoint
+
+
+def is_block_hash(value: str) -> bool:
+    return is_hex(value) and len(remove_0x_prefix(value)) == 64
+
+
+def remove_non_digits(value: str) -> str:
+    return re.sub("\D", "", value)
+
+
+def parse_checkpoint_uri(uri: str) -> Checkpoint:
+    try:
+        parsed = urllib.parse.urlparse(uri)
+    except ValueError as e:
+        raise ValidationError(str(e))
+
+    scheme, netloc, path, query = parsed.scheme, parsed.netloc, parsed.path.lower(), parsed.query
+
+    try:
+        parsed_query = urllib.parse.parse_qsl(query)
+    except ValueError as e:
+        raise ValidationError(str(e))
+
+    query_dict = dict(parsed_query)
+
+    # we allow any kind of separator for a nicer UX. e.g. instead of "11487662456884849810705"
+    # one can use "114 876 624 568 848 498 107 05" or "1,487,662,456,884,849,810,705". This also
+    # allows copying out a value from e.g etherscan.
+    score = remove_non_digits(query_dict.get('score', ''))
+
+    is_by_hash = path.startswith('/byhash')
+    parts = PurePosixPath(parsed.path).parts
+
+    if len(parts) != 3 or scheme != 'eth' or netloc != 'block' or not is_by_hash or not score:
+        raise ValidationError(
+            'checkpoint string must be of the form'
+            '"eth://block/byhash/<hash>?score=<score>"'
+        )
+
+    block_hash = parts[2]
+
+    if not is_block_hash(block_hash):
+        raise ValidationError(f'Block hash must be valid hex string, got: {block_hash}')
+
+    if not score.isdigit():
+        raise ValidationError(f'Score (total difficulty) must be an integer, got: {score}')
+
+    return Checkpoint(Hash32(decode_hex(block_hash)), int(score))
+
+
+class NormalizeCheckpointURI(argparse.Action):
+    """
+    Normalize the URI describing a sync checkpoint.
+    """
+    def __call__(self,
+                 parser: argparse.ArgumentParser,
+                 namespace: argparse.Namespace,
+                 value: Any,
+                 option_string: str=None) -> None:
+        parsed = parse_checkpoint_uri(value)
+        setattr(namespace, self.dest, parsed)
--- a/trinity/plugins/builtin/syncer/plugin.py
+++ b/trinity/plugins/builtin/syncer/plugin.py
@@ -75,6 +75,7 @@ from trinity.sync.light.chain import (
 from trinity._utils.shutdown import (
     exit_with_services,
 )
+from .cli import NormalizeCheckpointURI
 
 
 class BaseSyncStrategy(ABC):
@@ -200,6 +201,16 @@ class BeamSyncStrategy(BaseSyncStrategy):
             default=None,
         )
 
+        arg_parser.add_argument(
+            '--beam-from-checkpoint',
+            action=NormalizeCheckpointURI,
+            help=(
+                "Start beam sync from a trusted checkpoint specified using URI syntax:"
+                "eth://block/byhash/<hash>?score=<score>"
+            ),
+            default=None,
+        )
+
     async def sync(self,
                    args: Namespace,
                    logger: Logger,
@@ -215,6 +226,7 @@ class BeamSyncStrategy(BaseSyncStrategy):
             base_db,
             cast(ETHPeerPool, peer_pool),
             event_bus,
+            args.beam_from_checkpoint,
             args.force_beam_block_number,
             cancel_token,
         )
--- a/trinity/sync/beam/service.py
+++ b/trinity/sync/beam/service.py
@@ -1,5 +1,4 @@
 from lahja import EndpointAPI
-
 from cancel_token import CancelToken
 
 from eth.abc import AtomicDatabaseAPI
@@ -9,6 +8,7 @@ from p2p.service import BaseService
 from trinity.chains.base import AsyncChainAPI
 from trinity.db.eth1.chain import BaseAsyncChainDB
 from trinity.protocol.eth.peer import ETHPeerPool
+from trinity.sync.common.checkpoint import Checkpoint
 
 from .chain import BeamSyncer
 
@@ -22,6 +22,7 @@ class BeamSyncService(BaseService):
             base_db: AtomicDatabaseAPI,
             peer_pool: ETHPeerPool,
             event_bus: EndpointAPI,
+            checkpoint: Checkpoint = None,
             force_beam_block_number: int = None,
             token: CancelToken = None) -> None:
         super().__init__(token)
@@ -30,17 +31,26 @@ class BeamSyncService(BaseService):
         self.base_db = base_db
         self.peer_pool = peer_pool
         self.event_bus = event_bus
+        self.checkpoint = checkpoint
         self.force_beam_block_number = force_beam_block_number
 
     async def _run(self) -> None:
         head = await self.wait(self.chaindb.coro_get_canonical_head())
-        self.logger.info("Starting beam-sync; current head: %s", head)
+
+        if self.checkpoint is not None:
+            self.logger.info(
+                "Starting beam-sync; current head: %s, using checkpoint: %s", head, self.checkpoint
+            )
+        else:
+            self.logger.info("Starting beam-sync; current head: %s", head)
+
         beam_syncer = BeamSyncer(
             self.chain,
             self.base_db,
             self.chaindb,
             self.peer_pool,
             self.event_bus,
+            self.checkpoint,
             self.force_beam_block_number,
             token=self.cancel_token,
         )
--- a/trinity/sync/common/headers.py
+++ b/trinity/sync/common/headers.py
@@ -34,7 +34,6 @@ from eth_utils.toolz import (
     take,
 )
 
-from eth.constants import GENESIS_BLOCK_NUMBER
 from eth.exceptions import (
     HeaderNotFound,
 )
@@ -62,6 +61,10 @@ from trinity.sync.common.constants import (
     MAX_SKELETON_REORG_DEPTH,
 )
 from trinity.sync.common.peers import TChainPeer, WaitingPeers
+from trinity.sync.common.strategies import (
+    FromGenesisLaunchStrategy,
+    SyncLaunchStrategyAPI,
+)
 from trinity._utils.datastructures import (
     DuplicateTasks,
     OrderedTaskPreparation,
@@ -87,10 +90,15 @@ class SkeletonSyncer(BaseService, Generic[TChainPeer]):
                  chain: AsyncChainAPI,
                  db: BaseAsyncHeaderDB,
                  peer: TChainPeer,
-                 token: CancelToken) -> None:
+                 token: CancelToken,
+                 launch_strategy: SyncLaunchStrategyAPI = None) -> None:
         super().__init__(token=token)
         self._chain = chain
         self._db = db
+        if launch_strategy is None:
+            launch_strategy = FromGenesisLaunchStrategy(db, chain)
+
+        self._launch_strategy = launch_strategy
         self.peer = peer
         max_pending_headers = peer.max_headers_fetch * 8
         self._fetched_headers = asyncio.Queue(max_pending_headers)
@@ -229,7 +237,8 @@ class SkeletonSyncer(BaseService, Generic[TChainPeer]):
             previous_tail_header = final_headers[-1]
 
     async def _find_newest_matching_skeleton_header(self, peer: TChainPeer) -> BlockHeader:
-        start_num = await self._get_starting_block_num()
+        start_num = await self.wait(self._launch_strategy.get_starting_block_number())
+
         # after returning this header, we request the next gap, and prefer that one header
         # is new to us, which may be the next header in this mini-skeleton. (hence the -1 below)
         skip = MAX_HEADERS_FETCH - 1
@@ -336,7 +345,6 @@ class SkeletonSyncer(BaseService, Generic[TChainPeer]):
                 raise ValidationError(
                     f"First header {new_headers[0]} did not have parent in DB"
                 ) from exc
-
             # validate new headers against the parent in the database
             await self.wait(self._chain.coro_validate_chain(
                 launch_parent,
@@ -418,15 +426,6 @@ class SkeletonSyncer(BaseService, Generic[TChainPeer]):
                 # skip the pair following the gap, include all the following pairs
                 pairs[gap_index + 2:],
             ))
-
-    async def _get_starting_block_num(self) -> BlockNumber:
-        head = await self.wait(self._db.coro_get_canonical_head())
-
-        # When we start the sync with a peer, we always request up to MAX_REORG_DEPTH extra
-        # headers before our current head's number, in case there were chain reorgs since the last
-        # time _sync() was called. All of the extra headers that are already present in our DB
-        # will be discarded so we don't unnecessarily process them again.
-        return max(GENESIS_BLOCK_NUMBER, head.block_number - self.max_reorg_depth)
 
     async def _fetch_headers_from(
             self,
@@ -783,6 +782,7 @@ class BaseHeaderChainSyncer(BaseService, HeaderSyncerAPI, Generic[TChainPeer]):
                  chain: AsyncChainAPI,
                  db: BaseAsyncHeaderDB,
                  peer_pool: BaseChainPeerPool,
+                 launch_strategy: SyncLaunchStrategyAPI = None,
                  token: CancelToken = None) -> None:
         super().__init__(token)
         self._db = db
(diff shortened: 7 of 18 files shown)


setup.py
```diff
@@ -3,7 +3,7 @@
 import os
 from setuptools import setup, find_packages
 
-PYEVM_DEPENDENCY = "py-evm==0.3.0a3"
+PYEVM_DEPENDENCY = "py-evm@git+https://github.com/cburgdorf/py-evm.git@christoph/spike/persist-orphan-header"  # noqa: E501
```
cburgdorf: Can't merge this before we cut a new py-evm release

cburgdorf:
@pipermerriam @carver This isn't ready to be merged yet because we need to cut a new Py-EVM release first. It also has no tests yet (That said, at least the current tests prove that it's not breaking existing functionality)

But I'm curious about your feedback on the taken approach and maybe things that are missing.

trinity/sync/common/headers.py
```diff
@@ -75,6 +81,93 @@
...
+        self._checkpoint = checkpoint
+
+    async def can_start_with_header(self, header: BlockHeader) -> bool:
+        if header.hash == self._checkpoint.block_hash:
+            # The side effect is unfortunate but it's a special case of this strategy that we
+            # have to persist this header at this point. After this, the sync will continue
+            # in the usual way.
+
+            # TODO: Do we need to do any further validation before we persist this?
+            await self._db.coro_persist_checkpoint_header(header, self._checkpoint.score)
```
cburgdorf: @pipermerriam @carver 

I wonder if we would need to perform any special kind of validation here first. We essentially only the hash and then save it in our db with our given score. What if a peer serves us a header with that hash that is actually pure garbage (only the hash field is set to match our expectations but the rest doesn't match).

cburgdorf: Also, as the comment hints, I'm not happy about the side effect of this method but I'm also unsure about better alternatives.

carver: > What if a peer serves us a header with that hash that is actually pure garbage (only the hash field is set to match our expectations but the rest doesn't match).

The header hash isn't a field in the header RLP returned by the client, we calculate it locally. Also, we verify that the first header received over the network has the hash that we requested in `BaseBlockHeadersValidator`. So I think we're doing okay on validation that the header has the specified hash.

carver:
Looks/sounds like this is a WIP, so just adding it to the title.

trinity/sync/common/headers.py
cburgdorf: >The header hash isn't a field in the header RLP returned by the client, we calculate it locally

:see_no_evil: SURE!

See, I'm a real blockchain professional. Been in this space for 30 years :joy:

cburgdorf:
@pipermerriam @carver This is ready for a full review. I also added a test that syncs from a checkpoint (and by which I also fixed a bug that existed in a previous version of this PR).

The only things missing from this PR (afaik) is support for directly fetching checkpoints through etherscan or infura which I intend to add with another PR.

tests/core/p2p-proto/test_sync.py
```diff
@@ -206,7 +235,7 @@ def small_header_batches(monkeypatch):
         #   to the chaindb_churner canonical head, and increase the timeout significantly
         target_block_number = min(beam_to_block + 10, 129)
         target_head = chaindb_churner.get_canonical_block_header_by_number(target_block_number)
-        await wait_for_head(chaindb_fresh, target_head, sync_timeout=4)
+        await wait_for_head(chaindb_fresh, target_head, sync_timeout=5)
```
cburgdorf: I needed this locally to work with these tests but I *think* I've also seen this timeout in CI from time to time so maybe increasing this will make this test less flaky under CI, too.

trinity/plugins/builtin/syncer/cli.py
```diff
@@ -0,0 +1,73 @@
...
+    if not block_number.isdigit():
+        raise ValidationError(f'Block number must be an integer, got: {block_number}')
+
+    if not score.isdigit():
+        raise ValidationError(f'Score (total difficulty) must be an integer, got: {score}')
+
+    return Checkpoint(Hash32(decode_hex(block_hash)), BlockNumber(int(block_number)), int(score))
+
+
+class NormalizeCheckpointUri(argparse.Action):
```
pipermerriam: *naming nitpick*.  I'd all-caps `URI`

trinity/plugins/builtin/syncer/plugin.py
```diff
@@ -200,6 +201,13 @@ def configure_parser(cls, arg_parser: ArgumentParser) -> None:
             default=None,
         )
 
+        arg_parser.add_argument(
+            '--beam-from-checkpoint',
+            action=NormalizeCheckpointUri,
+            help="Start beam sync from a trusted checkpoint",
```
pipermerriam: Help text here could include the template for how a checkpoint URI should be formatted.

trinity/sync/common/headers.py
```diff
@@ -75,6 +81,92 @@
...
+    async def get_expected_score(self) -> int:
+        head_td = await self._genesis_strategy.get_expected_score()
+        return head_td if head_td > self._checkpoint.score else self._checkpoint.score
+
+    def get_genesis_parent_hash(self) -> Hash32:
+        return self._checkpoint.block_hash
+
+    async def get_starting_block_number(self) -> BlockNumber:
+        default_starting_block_number = await self._genesis_strategy.get_starting_block_number()
+        if default_starting_block_number < self._checkpoint.block_number:
```
carver: Okay, I see how the block number is used... I still think it's worth adding a pre-sync stage to download just the one checkpoint header that was specified. That will give a nice clear error message if the checkpoint was mis-specified somehow (rather than a surprising/confusing skeleton sync one).

So the rough idea is:
- loop over peers, asking for the header, until you get it back
- block any other syncing until ^ is complete
- maybe timeout after 60s (given that you are actually connected to a peer), and print a clear log like f"your connected peers did not respond with the checkpoint header specified in the `--whatever` argument: {checkpoint}"

Maybe this belongs in something like a new `await SyncLaunchStrategyAPI.fulfill_prerequisites()` (probably with a better name). The genesis one could be a noop, and the other one could explicitly download the header. An awkward part is that it would need the peer pool or something as an argument. But maybe that's reasonable. :man_shrugging: 

---

I just noticed that maybe it removes the side-effect mentioned here, too:

```py
            # The side effect is unfortunate but it's a special case of this strategy that we
            # have to persist this header at this point. After this, the sync will continue
            # in the usual way.
```

trinity/sync/common/headers.py
```diff
@@ -75,6 +81,92 @@
...
 
 
+class SyncLaunchStrategyAPI(ABC):
+
+    @abstractmethod
+    async def can_start_with_header(self, header: BlockHeader) -> bool:
+        ...
+
+    @abstractmethod
+    async def get_expected_score(self) -> int:
```
carver: The word `expected` in here was a surprise to me. Isn't this just the current head's local score? `get_current_score()` or `get_head_score()`, maybe?

Either way, it sounds like there is some subtlety in here that is probably worth adding a docstring for.

trinity/sync/common/headers.py
```diff
@@ -75,6 +81,92 @@
 )
 
 
+class SyncLaunchStrategyAPI(ABC):
```
carver: This API neatly peels out a contained piece of functionality. Nice! :+1:

trinity/sync/common/headers.py
```diff
@@ -75,6 +81,92 @@
...
+    @abstractmethod
+    def get_genesis_parent_hash(self) -> Hash32:
+        ...
+
+    @abstractmethod
+    async def get_starting_block_number(self) -> BlockNumber:
+        ...
+
+
+class FromGenesisLaunchStrategy(SyncLaunchStrategyAPI):
```
carver: Maybe these belong in a new `sync/common/strategies.py`, since this module is already pretty big.

trinity/sync/common/headers.py
```diff
@@ -243,17 +340,19 @@ def __init__(self,
         # check the first returned value
         first = skeleton_launch_headers[0]
 
-        first_is_present = await self.wait(self._db.coro_header_exists(first.hash))
+        can_start_with_header = await self.wait(
+            self._launch_strategy.can_start_with_header(first)
+        )
 
-        if not first_is_present:
+        if not can_start_with_header:
```
carver: Nice generalization of the existing code. :+1:

trinity/sync/common/headers.py
```diff
@@ -243,17 +340,19 @@ def __init__(self,
...
             await self._log_ancester_failure(peer, first)
             raise ValidationError(f"No common ancestor with {peer}, who started with {first}")
         elif len(skeleton_launch_headers) == 1:
             return skeleton_launch_headers[0]
         else:
             for parent, child in sliding_window(2, skeleton_launch_headers):
-                is_present = await self.wait(self._db.coro_header_exists(child.hash))
-                if not is_present:
+                usable_header = await self.wait(self._launch_strategy.can_start_with_header(child))
```
carver: Maybe `is_usable_header`? Although, maybe this whole line change reverts, if we do the pre-downloading of the checkpoint header.

newsfragments/921.feature.rst
```diff
@@ -0,0 +1,8 @@
+Allow beam sync to start from a trusted checkpoint.
+Specify a checkpoint via CLI parameter such as:
+
+--beam-from-checkpoint="block://blockhash/blocknumber/score"
```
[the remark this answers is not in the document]
cburgdorf: I like your proposed URI format. Let's do that :+1:

tests/core/p2p-proto/test_sync.py
```diff
@@ -140,14 +167,15 @@ def small_header_batches(monkeypatch):
 # This test was reduced to a few cases for speed. To run the full suite, use
 # range(1, 130) for beam_to_block. (and optionally follow the instructions at target_head)
 @pytest.mark.asyncio
-@pytest.mark.parametrize('beam_to_block', [1, 66, 68, 129])
+@pytest.mark.parametrize('beam_to_block', [66, 1, 66, 68, 129])
```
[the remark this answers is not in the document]
cburgdorf: Oops, it is!

trinity/sync/beam/service.py
```diff
@@ -30,17 +31,22 @@ def __init__(
...
         self.peer_pool = peer_pool
         self.event_bus = event_bus
+        self.checkpoint = checkpoint
         self.force_beam_block_number = force_beam_block_number
 
     async def _run(self) -> None:
-        head = await self.wait(self.chaindb.coro_get_canonical_head())
-        self.logger.info("Starting beam-sync; current head: %s", head)
+        if self.checkpoint is not None:
+            self.logger.info("Starting beam-sync; using checkpoint: %s", self.checkpoint)
```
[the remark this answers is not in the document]
cburgdorf: Good point :+1:

trinity/sync/common/headers.py
cburgdorf: It's not necessary the score of the head because it may also be the score of the defined checkpoint. What about `get_launch_score` or `get_start_score`?

trinity/sync/common/headers.py
cburgdorf: Yep, agreed!

trinity/sync/common/headers.py
cburgdorf: How about we delay that to the next PR that will add the functionality to defer to *etherscan | infura* to figure out the specific checkpoint?

With these, we'll first need to resolve to a checkpoint anyway before the sync starts so it is somewhat related.

I'm leaning towards the idea that this would happen outside/before the `SyncLaunchStrategyAPI` comes into play but that's just a feeling at this point. But I do think that postponing this until we tackle the *etherscan | infura* resolving makes sense as there may be case for a streamlined API.

In fact, instead of asking our peers for the block (and therefore depending on the peer pool) we *could* also default to the etherscan API to resolve information that wasn't specified by the user (in fact, that would then also work for `score`).

Well, these are just loose thoughts, my main point is, maybe we first tackle the *etherscan | infura* resolving to see how that turns out / forms the API.

This would leave us with the options to:

- merge this already (addressed all other poiints)
- keep it open and add *etherscan | infura* resolving to this PR

cburgdorf:
Thank you both for the reviews! I think I addressed all points except one which I explain in this comment.

https://github.com/ethereum/trinity/pull/921#discussion_r315621873

Basically, my thought is to first add the *etherscan | infura* resolving to see how that might form the API.

I can start working on that right away and we could do one of these things:

- keep this PR open until *etherscan | infura* resolving is finished
- merge it already

My preference would be to merge this already but if you prefer not to then that's ok, too :)

trinity/plugins/builtin/syncer/plugin.py
```diff
@@ -215,6 +226,7 @@ def configure_parser(cls, arg_parser: ArgumentParser) -> None:
             base_db,
             cast(ETHPeerPool, peer_pool),
             event_bus,
+            args.beam_from_checkpoint,
```
cburgdorf: @carver I *think* that we could handle the resolving of checkpoints nicely on top of this:



```python

if args.beam_from_checkpoint:
    checkpoint = await checkpoint_resolver.resolve(args.beam_from_checkpoint)

syncer = BeamSyncService(
            chain,
            AsyncChainDB(base_db),
            checkpoint
            ...
)
``` 

This would mean that:

1. `args.beam_from_checkpoint` is just the URI string that was validated by the parser, not the checkpoint itself. Actually, I'd thought this could just be a `Future[Checkpoint]` but that doesn't work because it needs to be pickable. Not big deal.

2. There's no extra resolving complexity *within* the syncing primitives

3. Resolving could theoretically call out to various APIs (peers, infura, etherscan) and just use whoever responds fastest

4. Naturally, the sync won't until the checkpoint is resolved and hence there's nothing special to do *within* `BeamSyncService` to wait on anything.

What do you think?

newsfragments/921.feature.rst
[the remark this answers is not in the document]
pipermerriam: I want to push back and propose we stick with the existing standard from [BIP-122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki)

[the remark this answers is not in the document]
cburgdorf: >I want to push back and propose we stick with the existing standard from BIP-122

If I understand that standard correctly it doesn't cover how to retrieve a block hash **and** a score together (or even block hash + score + number if we want it self contained, independent from requesting out to get the block number from somewhere else).

I think @carver brought up good reasons to adjust the theme (and the PR implements it already).

Basically the only thing I personally care about is to have a scheme with a command line parameter because I feel multiple hurt the UX.

How do we move on here?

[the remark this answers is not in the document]
pipermerriam: Ah, I realize this is missing `score`.... hrm,  I don't have a good answer and don't want to hold this up.  carry on with whatever you were doing before and we can adjust the scheme later if we find something we really like.

trinity/sync/common/headers.py
carver: I am thinking of setting the checkpoint as like the user pre-defining a starting head, like setting a genesis file is defining a starting head. Are you shy to call it the head because we haven't verified it back to the genesis yet?

If we called `chain.get_score(chain.get_canonical_head())`, then I believe it will return the same thing an `get_expected_score()`, whether or not we used a checkpoint. So `get_head_score()` still makes a lot of sense to me.

`launch` and `start` scores both sound like things that wouldn't/shouldn't change over time as you sync, but this returned value will.

trinity/sync/common/headers.py
carver: Hm, I'm not wild about using etherscan/etc for anything other than giving us a checkpoint hash & score. Open to the conversation, though.

I don't necessarily think that how we resolve the block number requires us to bring in 3rd-party services.

The main reason to tackle this now is: we can entirely remove the block number CLI specification, and logic handling, from this PR now. Otherwise, we are adding it as something that we expect to become cruft soon.

trinity/plugins/builtin/syncer/plugin.py
carver: Yeah, totally, handling it before sync even starts sounds great. I assume this would not only resolve, but also *persist* the header.

I think it's worth building the peer-based resolver now, and removing the block number from the checkpoint. Then some of the follow-on logic gets more unified, when we can assume the starting header (genesis or checkpoint) is already persisted and considered canonical.

trinity/sync/common/headers.py
cburgdorf: My initial thought was that it would probably be faster to resolve the block number through 3rd party services because we don't need to wait for connected peers.

But now that I give it more thought, maybe that doesn't really matter because having peers is a pre-condition to syncing anyway so by the time that we'd be ready to start syncing, we'll also be ready to resolve the given hash with a peer.

I'm gonna give it a try.

trinity/sync/common/headers.py
cburgdorf: Yes, I think you are right. Storing the header before sync even starts seems to eliminate some of the different code paths needed within the syncing logic. That's a good thing. I'm gonna give it a try.

trinity/sync/common/strategies.py
```diff
@@ -0,0 +1,136 @@
...
+        self._db = db
+        self._chain = chain
+        # We wrap the `FromGenesisLaunchStrategy` because we delegate to it at times and
+        # reaching for inheritance seems wrong in this case.
+        self._genesis_strategy = FromGenesisLaunchStrategy(self._db, self._chain)
+        self._checkpoint = checkpoint
+        self._peer_pool = peer_pool
+
+    async def fulfill_prerequisites(self) -> None:
+        max_attempts = 1000
```
cburgdorf: @carver instead of a fixed timeout I used a limit of attempts here. The actual time it waits is therefore a bit more dynamic depending on the fact whether we are connected to peers or not. This feels more natural to me. E.g. if I'm in general trouble to find peers it seems reasonable that this task would factor that in and also wait a little longer. Does that make sense?

cburgdorf:
@carver I refactored this again and (to me) it feels indeed much cleaner now :)

It does depend on a new upcoming py-evm version though https://github.com/ethereum/py-evm/pull/1830

(I have to rush to my daughters birthday so skipping further explanations for now)

trinity/sync/common/strategies.py
carver: Yup, sounds good!

trinity/sync/beam/service.py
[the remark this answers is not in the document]
carver: Seems a little verbose to show `Starting beam-sync` twice, if there is a checkpoint.

trinity/sync/common/strategies.py
```diff
@@ -0,0 +1,136 @@
...
+                    reverse=False,
+                )
+            except (TimeoutError, PeerConnectionLost, ValidationError):
+                # Nothing to do here. The ExchangeManager will disconnect if appropriate
+                # and eventually lead us to a better peer.
+                pass
+
+            if not headers:
+                await peer.disconnect(DisconnectReason.useless_peer)
+            elif headers[0].hash != self._checkpoint.block_hash:
```
carver: This is already validated internally to the `get_block_headers()` call, but I guess no objection to re-validating it here, since it's cheap and only happens once.

trinity/sync/common/strategies.py
```diff
@@ -0,0 +1,136 @@
...
+                await peer.disconnect(DisconnectReason.useless_peer)
+            else:
+                await self._db.coro_persist_checkpoint_header(headers[0], self._checkpoint.score)
+                return
+
+            await asyncio.sleep(0.5)
```
carver: Not sure we even need to wait this long, though I'm also not sure it matters much. We can only ask 2 peers per second for the data, but in practice, it should only take a couple tries at most, I hope/expect.

trinity/sync/common/strategies.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def get_genesis_parent_hash(self) -> Hash32:
+        return self._checkpoint.block_hash
+
+    async def get_starting_block_number(self) -> BlockNumber:
+        head = await self._db.coro_get_canonical_head()
+        if head.hash == self._checkpoint.block_hash:
+            return head.block_number
+        else:
+            return await self._genesis_strategy.get_starting_block_number()
```
carver: I notice this is making two `coro_get_canonical_head` calls on every check that isn't the checkpoint. Not a big deal, but maybe an easy optimization.

This whole method has exactly the same effect as the genesis launch but with `GENESIS_BLOCK_NUMBER` set to the checkpoint block number. Maybe the sync strategy can collect and cache that value, and then reuse the logic from the genesis version, with a custom minimum block number.

setup.py
```diff
@@ -3,7 +3,7 @@
 import os
 from setuptools import setup, find_packages
 
-PYEVM_DEPENDENCY = "py-evm==0.3.0a3"
+PYEVM_DEPENDENCY = "py-evm@git+https://github.com/cburgdorf/py-evm.git@ce0470801426166d28b826bbc71a57598ef395d1"  # noqa: E501
```
carver: Just a reminder to get this updated, as you already mentioned.

trinity/sync/common/strategies.py
cburgdorf: :+1:  Changed it to `0.05`

trinity/sync/common/strategies.py
cburgdorf: Ah wait, if I just define `min_block_number = BlockNumber(0)` in the strategy and then implement 

```python
    async def get_starting_block_number(self) -> BlockNumber:
        block_number = await self._genesis_strategy.get_starting_block_number()
        return block_number if block_number > self.min_block_number else self.min_block_number
```

There there is no exception to check, no duplicate checking of `coro_get_canonical_head` and no way we could return a block number that is lower than the one coming from our checkpoint.

Sweet!

trinity/sync/beam/service.py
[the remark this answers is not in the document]
cburgdorf: :+1: fixed

cburgdorf:
Landed :tada: Thanks for guiding me through it @carver !

setup.py
```diff
@@ -3,7 +3,7 @@
 import os
 from setuptools import setup, find_packages
 
-PYEVM_DEPENDENCY = "py-evm==0.3.0a3"
+PYEVM_DEPENDENCY = "py-evm==0.3.0a5"  # noqa: E501
```
carver: Hm, I'm tempted to add `yesqa` as a lint check: https://github.com/asottile/yesqa

trinity/sync/common/strategies.py
```diff
@@ -0,0 +1,136 @@
...
+            await asyncio.sleep(0.05)
+
+        raise TimeoutError(f"Failed to get checkpoint header within {max_attempts} attempts")
+
+    def get_genesis_parent_hash(self) -> Hash32:
+        return self._checkpoint.block_hash
+
+    async def get_starting_block_number(self) -> BlockNumber:
+        block_number = await self._genesis_strategy.get_starting_block_number()
+        return block_number if block_number > self.min_block_number else self.min_block_number
```
carver: IIUC, I think it might be a *good* idea to fail here if you incorrectly call this before finishing the fulfill call. It might help catch a bug earlier.

setup.py
cburgdorf: Oops, I send a follow up to fix this! And yep, linting for unnecessary `# noqa` comments would be really good!

PR #922, opened 2019-08-14
ENR DB

jannikluhn:
### What was wrong?

We need to store ENRs somehow.

### How was it fixed?

Define an interface for ENR DBs and provide a simple in-memory implementation.

This PR builds on top of #920 because it uses the `ENRFactory` in the tests added there. Other than that it's independent.

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/enr_db.py
+++ b/p2p/discv5/enr_db.py
@@ -1,0 +1,93 @@
+from abc import ABC
+from typing import (
+    Dict,
+)
+
+import trio
+
+from p2p.discv5.enr import ENR
+from p2p.discv5.identity_schemes import IdentitySchemeRegistry
+from p2p.discv5.typing import NodeID
+
+
+class EnrDbApi(ABC):
+
+    def __init__(self, identity_scheme_registry: IdentitySchemeRegistry):
+        self.identity_scheme_registry = identity_scheme_registry
+
+    def validate_identity_scheme(self, enr: ENR) -> None:
+        """Check that we know the identity scheme of the ENR.
+
+        This check should be performed whenever an ENR is inserted or updated in serialized form to
+        make sure retrieving it at a later time will succeed (deserializing the ENR would fail if
+        we don't know the identity scheme).
+        """
+        if enr.identity_scheme.id not in self.identity_scheme_registry:
+            raise ValueError(
+                f"ENRs identity scheme with id {enr.identity_scheme.id} unknown to ENR DBs "
+                f"identity scheme registry"
+            )
+
+    async def insert(self, enr: ENR) -> None:
+        """Insert an ENR into the database."""
+        ...
+
+    async def update(self, enr: ENR) -> None:
+        """Update an existing ENR if the sequence number is greater."""
+        ...
+
+    async def remove(self, node_id: NodeID) -> None:
+        """Remove an ENR from the db."""
+        ...
+
+    async def insert_or_update(self, enr: ENR) -> None:
+        """Insert or update an ENR depending if it is already present already or not."""
+        ...
+
+    async def get(self, node_id: NodeID) -> ENR:
+        """Get an ENR by its node id."""
+        ...
+
+    async def contains(self, node_id: NodeID) -> bool:
+        """Check if the db contains an ENR with the given node id."""
+        ...
+
+
+class MemoryEnrDb(EnrDbApi):
+
+    def __init__(self, identity_scheme_registry: IdentitySchemeRegistry):
+        self.identity_scheme_registry = identity_scheme_registry
+        self.key_value_storage: Dict[NodeID, ENR] = {}
+
+    async def insert(self, enr: ENR) -> None:
+        self.validate_identity_scheme(enr)
+
+        if await self.contains(enr.node_id):
+            raise ValueError(f"ENR with nodeid {enr.node_id} already exists.")
+        else:
+            self.key_value_storage[enr.node_id] = enr
+
+    async def update(self, enr: ENR) -> None:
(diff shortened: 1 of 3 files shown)


p2p/discv5/handshake.py
```diff
@@ -0,0 +1,351 @@
...
+    enr: Optional[ENR]
+    message: Optional[BaseMessage]
+    auth_header_packet: Optional[AuthHeaderPacket]
+
+
+class BaseHandshakeParticipant(ABC):
+    def __init__(self,
+                 is_initiator: bool,
+                 our_private_key: bytes,
+                 our_enr: ENR,
```
pipermerriam: maybe use `local` / `remote` to differentiate the two records.

p2p/discv5/handshake.py
```diff
@@ -0,0 +1,351 @@
...
+                 their_node_id: Optional[NodeID] = None,
+                 their_enr: Optional[ENR] = None,
+                 ) -> None:
+        self.is_initiator = is_initiator
```
pipermerriam: Maybe use `inbound` / `outbound` or `dial_in` / `dial_out` terminology to stay consistent with the existing peer terminology.

p2p/discv5/handshake.py
```diff
@@ -0,0 +1,351 @@
...
+
+        self.initiating_packet = AuthTagPacket.prepare_random(
+            tag=self.tag,
+            auth_tag=get_random_auth_tag(),
+            random_data=get_random_encrypted_data(),
+        )
+
+    @property
+    def identity_scheme(self) -> Type[IdentityScheme]:
+        return self.their_enr.identity_scheme
```
pipermerriam: It appears that `self.their_enr` can be `None` which would cause this to raise an `AttributeError`.

p2p/discv5/handshake.py
```diff
@@ -0,0 +1,351 @@
...
+
+    @property
+    def first_packet_to_send(self) -> Packet:
+        return self.initiating_packet
+
+    def is_response_packet(self, packet: Packet) -> bool:
+        return (
+            isinstance(packet, WhoAreYouPacket) and
+            packet.token == self.initiating_packet.auth_tag
```
pipermerriam: should this use a constant time comparison?

p2p/discv5/handshake.py
```diff
@@ -0,0 +1,351 @@
...
+
+        # compute session keys
+        (
+            ephemeral_private_key,
+            ephemeral_public_key,
+        ) = self.identity_scheme.create_handshake_key_pair()
+
+        session_keys = self.identity_scheme.compute_session_keys(
+            local_private_key=ephemeral_private_key,
+            peer_public_key=self.their_enr.public_key,
```
pipermerriam: Another place where it appears an `AttributeError` would occur of `their_enr` is `None`

p2p/discv5/packets.py
```diff
@@ -559,3 +562,18 @@ def _decrypt_message(key: AES128Key,
...
+def get_random_encrypted_data() -> bytes:
+    return os.urandom(RANDOM_ENCRYPTED_DATA_SIZE)
+
+
+def get_random_id_nonce() -> IDNonce:
+    return IDNonce(os.urandom(ID_NONCE_SIZE))
+
+
+def get_random_auth_tag() -> Nonce:
+    return Nonce(os.urandom(NONCE_SIZE))
```
pipermerriam: We should use the [`secrets.token_bytes`](https://docs.python.org/3/library/secrets.html#secrets.token_bytes) API for this.  (and probably audit our existing code to migrate to use `secrets` as well)

p2p/discv5/handshake.py
jannikluhn: I'd like to keep initiator/recipient because that's the terminology the spec uses.

p2p/discv5/handshake.py
jannikluhn: :+1: very good point, didn't think about this attack vector at all

p2p/discv5/handshake.py
jannikluhn: should be fine as this is in the initiator where we have a guarantee that we know the peer's ENR.

p2p/discv5/handshake.py
jannikluhn: same here, it's in the initiator, so the remote ENR is known.

p2p/discv5/handshake.py
pipermerriam: Maybe it would be better then to remove the common base implentation so that for the initiator class we have a typing-level guarantee that the value is not `None` because as it stands, our types allow this to be `None` *I think*

p2p/discv5/handshake.py
cburgdorf: You fired up my curiosity. Wow, I had no idea! For anyone else wondering here's a good read on the issue: https://codahale.com/a-lesson-in-timing-attacks/

p2p/discv5/enr_db.py
```diff
@@ -0,0 +1,74 @@
...
+        if await self.contains(enr.node_id):
+            raise ValueError(f"ENR with nodeid {enr.node_id} already exists.")
+        else:
+            self.key_value_storage[enr.node_id] = enr
+
+    async def update(self, enr: ENR) -> None:
+        self.validate_identity_scheme(enr)
+        existing_enr = await self.get(enr.node_id)
+        if existing_enr.sequence_number < enr.sequence_number:
+            self.key_value_storage[enr.node_id] = enr
```
pipermerriam: This might be a good place for a logging message.  Same with inside `insert` and `remove`

PR #923, opened 2019-08-14
Expose force-beam-block-number parameter on command line

cburgdorf:
### What was wrong?

The `force_beam_block_number` config is very useful for testing. I think it's worth exposing it as a CLI argument.

### How was it fixed?

1. In general, allow sync strategies to make modifications to the parser
2. Have the beam sync strategy add this specific config.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/plugins/builtin/syncer/plugin.py
+++ b/trinity/plugins/builtin/syncer/plugin.py
@@ -4,6 +4,7 @@ from abc import (
 )
 from argparse import (
     ArgumentParser,
+    Namespace,
     _SubParsersAction,
 )
 import asyncio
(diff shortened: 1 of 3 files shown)


trinity/plugins/builtin/syncer/plugin.py
```diff
@@ -176,7 +188,17 @@ class BeamSyncStrategy(BaseSyncStrategy):
...
+    def configure_parser(cls, arg_parser: ArgumentParser) -> None:
+        arg_parser.add_argument(
+            '--force-beam-block-number',
+            type=int,
+            help="Force beam sync to activate on a specific block number (for testing)",
+            default=None,
+        )
+
     async def sync(self,
+                   args: Namespace,
```
pipermerriam: I'm (minorly) :-1: on having the CLI args be exposed this far down.  It *might* be more appropriate to do something like a `SyncConfig` object that aggregates all of the configuration parameters from the CLI args and then for us to pass that through here.

This doesn't necessarily have to happen at this stage but I think it would be relatively simple to implement..

cburgdorf: The thing with a `SyncConfig` object is that this takes away some of the flexibility that I think we like to embrace. E.g. Today, someone can implement a new sync strategy as a plugin and just register the strategy so that it naturally becomes available via `--sync-mode whatever`. But this new sync mode might have options such as `--whatever-unknown-unknown` so having access to the `args`, while being *very* dynamic, might still be the best option to keep things very modular.

PR #925, opened 2019-08-14
Protocol.__eq__ and Protocol.__hash__

pipermerriam:
### What was wrong?

Work in #684 has me needing the `Protocol` class to be hashable and to be able to compare protocol classes.

### How was it fixed?

Added `__eq__` and `__hash__` implementations.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

names-for-big-dogs

### DIFF
--- a/p2p/protocol.py
+++ b/p2p/protocol.py
@@ -2,6 +2,7 @@ import logging
 import operator
 import struct
 from typing import (
+    Any,
     ClassVar,
     Iterable,
     Sequence,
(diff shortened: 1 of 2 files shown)


p2p/protocol.py
```diff
@@ -164,6 +165,12 @@ def __init__(self, transport: TransportAPI, cmd_id_offset: int, snappy_support:
         self.cmd_by_type = {type(cmd): cmd for cmd in self.commands}
         self.cmd_by_id = {cmd.cmd_id: cmd for cmd in self.commands}
 
+    def __eq__(self, other: Any) -> bool:
+        return type(self) is type(other)
```
lithp: Huh, I'm not sure where this will be used so I'm not sure how much of a problem it is that two different instances with two different values for `snappy_support` and `cmd_id_offset` will be equal, but it feels dangerous to assume that no uses of `Protocol` will ever want to distinguish between different instances of a Protocol. 

Anyone who doesn't want to distinguish can call `type(left) is type(right)` themselves, and I like the idea of them opting into this weird behavior more than I like the idea of this weird behavior being on by default.

What's your use-case for this?

pipermerriam: > What's your use-case for this?

I may not have one now that I'm looking for it.  I'm gonna close this an re-open later if I find it's necessary.

PR #927, opened 2019-08-14
Add MultiplexerAPI.cancel_token

pipermerriam:
### What was wrong?

In later work for #684 we need to use the multiplexer's cancel token but the formal API doesn't have a cancel token property.

### How was it fixed?

Added the property to the `MultiplexerAPI`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

top-10-large-breed-dogs

### DIFF
--- a/newsfragments/927.misc.rst
+++ b/newsfragments/927.misc.rst
@@ -1,0 +1,1 @@
+Add ``MultiplexerAPI.cancel_token`` property.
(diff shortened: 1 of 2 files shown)


newsfragments/927.feature.rst
```diff
@@ -0,0 +1 @@
+Add ``MultiplexerAPI.cancel_token`` property.
```
cburgdorf: I think this would be better placed as a `927.misc.rst` file. It's not really a user facing feature.

PR #931, opened 2019-08-14
Add BasePeerContext.p2p_version

pipermerriam:
### What was wrong?

The base protocol version should come from the handshake parameters (this is used in later work under #684 to cleanup how the base protocol is tested).

### How was it fixed?

Changed where the version comes from to allow it to be passed in as part of the handshake parameters.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

bernese-mountain-dog-ap-0t0bvi-590

### DIFF
--- a/trinity/server.py
+++ b/trinity/server.py
@@ -17,12 +17,13 @@ from eth_typing import BlockNumber
 from eth.abc import VirtualMachineAPI
 
 from p2p.abc import NodeAPI
-from p2p.constants import DEFAULT_MAX_PEERS
+from p2p.constants import DEFAULT_MAX_PEERS, DEVP2P_V5
 from p2p.disconnect import DisconnectReason
 from p2p.exceptions import (
     HandshakeFailure,
     PeerConnectionLost,
 )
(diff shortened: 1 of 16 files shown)


trinity/server.py
```diff
@@ -193,8 +201,9 @@ def _make_peer_pool(self) -> ETHPeerPool:
             headerdb=self.headerdb,
             network_id=self.network_id,
             vm_configuration=self.chain.vm_configuration,
-            client_version_string=construct_trinity_client_identifier(),
-            listen_port=self.port,
+            client_version_string=self.p2p_handshake_params.client_version_string,
+            listen_port=self.p2p_handshake_params.listen_port,
+            p2p_version=self.p2p_handshake_params.version,
```
carver: Maybe we could update the `ChainContext` to accept the handshake params directly here? Not sure that's a good idea.

pipermerriam: I'm planning on getting rid of the `ChainContext` entirely at a later stage so I'm inclined to keep it as-is and things get automatically cleaned up later.

PR #932, opened 2019-08-14
Beam sync fix hang

carver:
### What was wrong?

If a peer cancelled at just the right moment, your data request could hang indefinitely. If this is a `GetNodeData` request in Beam Sync, the whole sync service can get stuck.

### How was it fixed?

When waiting for the response stream to launch, wrap it in a call that gets cancelled if the peer is shut down.

Also, some other random Beam Sync fixups.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/common/managers.py
+++ b/trinity/protocol/common/managers.py
@@ -273,7 +273,7 @@ class ExchangeManager(Generic[TRequestPayload, TResponsePayload, TResult]):
             self._cancel_token,
         )
         self._peer.run_daemon(self._response_stream)
-        await self._response_stream.events.started.wait()
+        await self._peer.wait(self._response_stream.events.started.wait())
 
     @property
     def is_operational(self) -> bool:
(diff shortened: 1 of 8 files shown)


trinity/protocol/common/managers.py
```diff
@@ -273,7 +273,7 @@ def __init__(
             self._cancel_token,
         )
         self._peer.run_daemon(self._response_stream)
-        await self._response_stream.events.started.wait()
+        await self._peer.wait(self._response_stream.events.started.wait())
```
carver: This is the core fix

trinity/sync/beam/state.py
```diff
@@ -397,9 +397,7 @@ def _append_unique_hashes(
             self._node_tasks.complete(urgent_batch_id, tuple(urgent_nodes.keys()))
 
         if predictive_batch_id is not None:
-            # retire all predictions, if the responding node doesn't have them, then we don't
-            # want to keep asking
-            self._maybe_useful_nodes.complete(predictive_batch_id, predictive_node_hashes)
+            self._maybe_useful_nodes.complete(predictive_batch_id, tuple(predictive_nodes.keys()))
```
carver: This is another important one, we don't want to give up on predictive nodes, we want to keep retrying them. (This logic was leftover from the old concept of what predictive meant)

PR #935, opened 2019-08-15
LESHandshakeParams.as_payload_dict no longer takes version

pipermerriam:
### What was wrong?

The `LESHandshakeParams` class exposes an `as_payload_dict` which took a parameter to specify the version of the `LES` protocol that should be used.

It was more appropriate to just have this be part of the class parameters.

### How was it fixed?

Added it to the main class parameters.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

landscape-gettyimages-530330473-1


pipermerriam:
will get this green tomorrow.

PR #944, opened 2019-08-17
eth_getCode fetches data from beam syncer

lithp:
### What was wrong?

Continuing to address #893, this is the next step after #894.

### Future work:

-  This also needs to intercept MissingStorageTrieNode
- This needs to support and be tested with more RPC methods
- This needs to generate a friendly error if you ask for a state so old the network won't have it (that's also from before we started beam syncing)
- This needs to return a nice error if the beam syncer can't fetch the state (it shouldn't hang forever)
- It'd be nice to include an integration test that actually runs RPCServer and beam sync and verifies that they do the intended thing.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/json-rpc/test_ipc.py
+++ b/tests/core/json-rpc/test_ipc.py
@@ -8,6 +8,7 @@ import uuid
 from eth._utils.address import (
     force_bytes_to_address,
 )
+from eth_hash.auto import keccak
 from eth_utils.toolz import (
     assoc,
 )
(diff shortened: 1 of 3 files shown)


tests/core/json-rpc/test_ipc.py
```diff
@@ -656,3 +659,60 @@ def missing_node(chain_with_block_validation):
     response = await ipc_request('eth_getBalance', [funded_address.hex(), 'latest'])
     assert 'error' not in response
     assert response['result'] == hex(funded_address_initial_balance)
+
+
+# The same as above, but for eth_getCode
```
carver: "Same as above" works in this diff, but not if you see this comment in isolation, I think. :) Maybe just repeat the text.

Another option: we could move these to `test_ipc_on_beam.py`, have a general comment at the top about the category of tests, and then all these individual comments could be removed.

lithp: Yeah, good point :+1: 

Going to merge this as-is for now but about to open a PR which moves them into their own file.

carver:
Ooh, don't forget to add release notes!

lithp:
Nice catch! Adding it to #957

PR #900, opened 2019-08-10
fix `eth_getBalance` to give correct result when `at_block` of type integer is provided 

voith:
### What was wrong?
`w3.eth.getBalance('some_address', some_block_number)` would return the same result for 
some_block_number=0. 

### How was it fixed?

fixed `eth_getBalance` to use `at_block` when `at_block` is of type integer.
### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [x] Clean up commit history

[//]: # (For important changes that should go into the release notes please add a newsfragment file as explained here: https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)

[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)
- [x] Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #945, opened 2019-08-19
Add test for eth_getBalance at a specific block_number

voith:
### What was wrong?
There was a test missing for the bug fixed in #900

### How was it fixed?
 - Made `chain_without_block_validation` fixture use `MiningChain` in order to send eth from one address to another. I've searched for the usage of this fixture in code and it seems like it was stale and not used anywhere, so was safe to modify it. 
- Added test for` eth_getBalance` at a specific block_number by sending eth to an account from `funded_address` and then by comparing the balance at `block_no_before_transfer` and at `block_no_after_transfer `.
### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -19,7 +19,10 @@ from eth_utils import (
 from eth_keys import keys
 
 from eth import constants as eth_constants
-from eth.chains.base import Chain
+from eth.chains.base import (
+    Chain,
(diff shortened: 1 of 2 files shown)


voith:
Test failures seem unrelated to this change. The same tests are failing for me locally on master.

PR #937, opened 2019-08-15
Create PeerPairFactory and remove many *helper* functions from testing.

pipermerriam:
### What was wrong?

We had a small smattering of tools for generating connected peers.  In working on the handshake process, these utilities became a pain point as they were often heavily tied into the implementation details, they are not subject to `mypy` checks since they live under `./tests/` and they end up being scattered around the tests.

### How was it fixed?

Convert these to use the factory pattern using an `asynccontextmanager` to manage the lifecycle of the peers

This also converts the `FakeAsyncXXXDB` to live under `trinity.tools` as well.  Once #859 is merged these extra `FakeAsync` can actually go away entirely.

The diff here is large but there should be very minimal actual functionality changes to the test logic and this should have near zero effect to production code.

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Clean up commit history

[//]: # (For important changes that should go into the release notes please add a newsfragment file as explained here: https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)

[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)
- [x] Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #949, opened 2019-08-19
Remove "Fake" chain and database objects in favor of real ones.

pipermerriam:
This should be merged before #937 

### What was wrong?

With #859 merged we no longer need "fake" versions of our async chain and database objects in our tests.

### How was it fixed?

This removes the various `FakeAsync...` objects that are used in testing in favor of the real deal.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

Big-Dog-Sulley-Mastiff

### DIFF
--- a/tests/core/p2p-proto/bcc/helpers.py
+++ b/tests/core/p2p-proto/bcc/helpers.py
@@ -14,17 +14,12 @@ from eth.constants import (
 )
 from eth.db.atomic import AtomicDB
 
-from eth2.beacon.db.chain import BeaconChainDB
 from eth2.beacon.types.blocks import (
     BeaconBlock,
     BeaconBlockBody,
 )
 
-from tests.core.integration_test_helpers import (
-    async_passthrough,
-)
-
-from trinity.db.beacon.chain import BaseAsyncBeaconChainDB
+from trinity.db.beacon.chain import AsyncBeaconChainDB
 from trinity.protocol.bcc.context import BeaconContext
 from trinity.protocol.bcc.peer import (
     BCCPeerFactory,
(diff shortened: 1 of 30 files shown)


tests/core/p2p-proto/bcc/helpers.py
```diff
@@ -49,27 +44,6 @@
 SERENITY_GENESIS_CONFIG = Eth2GenesisConfig(SERENITY_CONFIG)
 
 
-class FakeAsyncBeaconChainDB(BaseAsyncBeaconChainDB, BeaconChainDB):
-
-    coro_persist_block = async_passthrough('persist_block')
```
pipermerriam: It is my thought that now that the real objects can work we should use them in tests. The previous async_passthrough approach is going to be more performant but it seems better to test against the real-deal.

tests/core/p2p-proto/test_sync.py
```diff
@@ -85,15 +89,15 @@ def small_header_batches(monkeypatch):
...
+        await asyncio.wait_for(client.run(), timeout=5)
 
         head = chaindb_fresh.get_canonical_head()
         assert head == chaindb_20.get_canonical_head()
 
         # Now download the state for the chain's head.
         state_downloader = StateDownloader(
             chaindb_fresh, chaindb_fresh.db, head.state_root, client_peer_pool)
-        await asyncio.wait_for(state_downloader.run(), timeout=2)
+        await asyncio.wait_for(state_downloader.run(), timeout=5)
```
pipermerriam: I increased all of these timeouts.  It seems better to have to wait a little longer for a failure than to have flaky CI tests.

trinity/_utils/async_dispatch.py
```diff
@@ -1,18 +1,43 @@
...
+TReturn = TypeVar('TReturn')
 
-def async_method(method_name: str) -> Callable[..., Any]:
-    async def method(self: Any, *args: Any, **kwargs: Any) -> Awaitable[Any]:
+
+def async_method(method: Callable[..., TReturn],
+                 ) -> Callable[..., Coroutine[Any, Any, TReturn]]:
+    @functools.wraps(method)
+    async def wrapper(cls_or_self: Any, *args: Any, **kwargs: Any) -> TReturn:
+        cls_method = getattr(cls_or_self, method.__name__)
```
pipermerriam: feedback appreciated.  This approach seems to work nicely for maintaining the right types for these *wrapped* methods and still *just working* for both normal instance methods and class methods.

pipermerriam:
reviewed as part of #937

PR #950, opened 2019-08-19
fix flakey test using msg buffer API

pipermerriam:
### What was wrong?

One of the LES command tests was flaky due to underlying changes to how subscriptions work.

### How was it fixed?

Changed to use a more precise waiting condition.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

s-l1000

### DIFF
--- a/tests/core/p2p-proto/test_les_protocol_commands.py
+++ b/tests/core/p2p-proto/test_les_protocol_commands.py
@@ -48,7 +48,14 @@ async def test_les_protocol_methods_request_id(
         generated_request_id = peer.sub_proto.send_get_block_headers(
             b'1234', 1, 0, False, request_id=request_id
         )
-        await asyncio.sleep(0.1)
+
+        # yield to let remote and peer transmit messages.  This can take a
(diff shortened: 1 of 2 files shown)


tests/core/p2p-proto/test_les_protocol_commands.py
```diff
@@ -48,7 +48,14 @@
...
         )
-        await asyncio.sleep(0.1)
+
+        # yield to let remote and peer transmit messages.  This can take a
+        # small amount of time so we give it a few rounds of the event loop to
+        # finish transmitting.
+        for _ in range(10):
+            await asyncio.sleep(0.01)
+            if buffer.msg_queue.qsize() >= 1:
+                break
```
lithp: Ideally here we would `await` (w/ a timeout) an event becoming available, I took a quick look but I guess `asyncio.Queue` doesn't expose a way to `await` new events without also consuming those events. Looks like we could relatively easily write one as a subclass of `asyncio.Queue` but it would depend on `asyncio.Queue` implementation details :cry:

PR #954, opened 2019-08-20
Adding Milagro BLS

ChihChengLiang:
### What was wrong?

To interop with other clients, which are all using Milagro BLS, we need a compatible library too. The PR attempts to use [a binding](https://github.com/ChihChengLiang/milagro_bls_binding) of [Sigma Prime's implementaion](https://github.com/sigp/milagro_bls/). 

Also, I discovered that bls-utils haven't been covered by the CI for a long time 😱, so I attempt to fix it in this PR too.

### How was it fixed?

- Add Milagro BLS backend
- see if it pass YAML tests

#### Cute Animal Picture

### DIFF
--- a/eth2/_utils/bls/__init__.py
+++ b/eth2/_utils/bls/__init__.py
@@ -97,6 +97,7 @@ class Eth2BLS:
         if cls.backend != NoOpBackend:
             validate_signature(signature)
             validate_public_key(pubkey)
+
         if not cls.verify(message_hash, pubkey, signature, domain):
             raise SignatureError(
                 f"backend {cls.backend.__name__}\n"
@@ -115,6 +116,7 @@ class Eth2BLS:
         if cls.backend != NoOpBackend:
             validate_signature(signature)
             validate_many_public_keys(pubkeys)
+
         if not cls.verify_multiple(pubkeys, message_hashes, signature, domain):
             raise SignatureError(
                 f"backend {cls.backend.__name__}\n"
--- a/eth2/_utils/bls/__init__.py
+++ b/eth2/_utils/bls/__init__.py
@@ -25,7 +25,8 @@ from .backends.base import (
 from .validation import (
     validate_private_key,
     validate_signature,
-    validate_empty_public_key,
+    validate_public_key,
+    validate_many_public_keys,
 )
 
 
@@ -95,7 +96,7 @@ class Eth2BLS:
                  domain: Domain) -> None:
         if cls.backend != NoOpBackend:
             validate_signature(signature)
-            validate_empty_public_key(pubkey)
+            validate_public_key(pubkey)
         if not cls.verify(message_hash, pubkey, signature, domain):
             raise SignatureError(
                 f"backend {cls.backend.__name__}\n"
@@ -113,6 +114,7 @@ class Eth2BLS:
                           domain: Domain) -> None:
         if cls.backend != NoOpBackend:
             validate_signature(signature)
+            validate_many_public_keys(pubkeys)
         if not cls.verify_multiple(pubkeys, message_hashes, signature, domain):
             raise SignatureError(
                 f"backend {cls.backend.__name__}\n"
--- a/eth2/_utils/bls/__init__.py
+++ b/eth2/_utils/bls/__init__.py
@@ -8,9 +8,6 @@ from eth_typing import (
     BLSSignature,
     Hash32,
 )
-from eth_utils import (
-    ValidationError,
-)
 
 from py_ecc.bls.typing import Domain
 
@@ -28,6 +25,7 @@ from .backends.base import (
 from .validation import (
     validate_private_key,
     validate_signature,
+    validate_empty_public_key,
 )
 
 
@@ -79,12 +77,6 @@ class Eth2BLS:
                pubkey: BLSPubkey,
                signature: BLSSignature,
                domain: Domain) -> bool:
-        if cls.backend != NoOpBackend:
-            try:
-                validate_signature(signature)
-            except ValidationError:
-                return False
-
         return cls.backend.verify(message_hash, pubkey, signature, domain)
 
     @classmethod
@@ -93,12 +85,6 @@ class Eth2BLS:
                         message_hashes: Sequence[Hash32],
                         signature: BLSSignature,
                         domain: Domain) -> bool:
-        if cls.backend != NoOpBackend:
-            try:
-                validate_signature(signature)
-            except ValidationError:
-                return False
-
         return cls.backend.verify_multiple(pubkeys, message_hashes, signature, domain)
 
     @classmethod
@@ -107,6 +93,9 @@ class Eth2BLS:
                  pubkey: BLSPubkey,
                  signature: BLSSignature,
                  domain: Domain) -> None:
+        if cls.backend != NoOpBackend:
+            validate_signature(signature)
+            validate_empty_public_key(pubkey)
         if not cls.verify(message_hash, pubkey, signature, domain):
             raise SignatureError(
                 f"backend {cls.backend.__name__}\n"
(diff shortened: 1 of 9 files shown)


eth2/_utils/bls/__init__.py
```diff
@@ -79,12 +77,6 @@ def verify(cls,
                pubkey: BLSPubkey,
                signature: BLSSignature,
                domain: Domain) -> bool:
-        if cls.backend != NoOpBackend:
-            try:
-                validate_signature(signature)
```
ChihChengLiang: validate_signature goes with validate* APIs

eth2/_utils/bls/backends/__init__.py
```diff
@@ -17,9 +18,15 @@
 
 DEFAULT_BACKEND = None  # type: Type[BaseBLSBackend]
 
+try:
+    from .milagro import MilagroBackend
+    DEFAULT_BACKEND = MilagroBackend
```
ChihChengLiang: Set `MilagroBackend` as default

eth2/_utils/bls/backends/milagro.py
```diff
@@ -0,0 +1,105 @@
...
+from eth2._utils.bls.backends.base import (
+    BaseBLSBackend,
+)
+from eth2.beacon.constants import (
+    EMPTY_PUBKEY,
+    EMPTY_SIGNATURE,
+)
+
+
+def to_int(domain: Domain) -> int:
```
ChihChengLiang: sigp/milagro_bls use int as domain

eth2/_utils/bls/backends/milagro.py
```diff
@@ -0,0 +1,105 @@
...
+                          message_hashes: Sequence[Hash32]) -> Iterator[Tuple[BLSPubkey, Hash32]]:
+    for i, pubkey in enumerate(pubkeys):
+        if pubkey != EMPTY_PUBKEY:
+            yield pubkey, message_hashes[i]
+
+
+class MilagroBackend(BaseBLSBackend):
+    @staticmethod
+    def privtopub(k: int) -> BLSPubkey:
+        return privtopub(k.to_bytes(48, 'big'))
```
ChihChengLiang: sigp/milagro_bls use 48 bytes privkey

tests/eth2/utils-tests/bls-utils/test_backends.py
```diff
@@ -26,13 +29,17 @@ def assert_signature(obj):
     assert isinstance(obj, bytes) and len(obj) == 96
 
 
+@pytest.fixture
+def domain():
+    return (123).to_bytes(8, 'big')
```
ChihChengLiang: use domain as bytes

ChihChengLiang:
I think it's ready for review now. Probably need to review relevant rust code. So far I'm panic about if ill-formed input making the methods panic. In the verification rust codebase, exceptions of bad public keys and signatures are caught, so it should return false and not panic when an input is not on the curve.

Source: https://github.com/ChihChengLiang/milagro_bls_binding/blob/master/src/lib.rs
Tests: https://github.com/ChihChengLiang/milagro_bls_binding/blob/master/test.py

ChihChengLiang:
This CI failure was caused by an empty `keymap`. So potentially it is a key cache bug. Not sure if it worth investigating though.
https://circleci.com/gh/ethereum/trinity/89150

eth2/_utils/bls/validation.py
```diff
@@ -3,21 +3,46 @@
...
+def validate_public_key(pubkey: BLSPubkey, allow_empty: bool =False) -> None:
+    if len(pubkey) != 48:
+        raise PublicKeyError(
+            f"Invalid public key length, expect 48 got {len(pubkey)}. pubkey: {pubkey}"
+        )
+    if not allow_empty and pubkey == EMPTY_PUBKEY:
+        raise PublicKeyError(f"Empty public key is invalid  pubkey={pubkey}")
+
+
+def validate_many_public_keys(pubkeys: Sequence[BLSPubkey]) -> None:
```
hwwhww: Maybe remove `many`:
```suggestion
def validate_public_keys(pubkeys: Sequence[BLSPubkey]) -> None:
```

eth2/_utils/bls/__init__.py
```diff
@@ -107,6 +94,9 @@ def validate(cls,
                  pubkey: BLSPubkey,
                  signature: BLSSignature,
                  domain: Domain) -> None:
+        if cls.backend != NoOpBackend:
+            validate_signature(signature)
+            validate_public_key(pubkey)
         if not cls.verify(message_hash, pubkey, signature, domain):
```
hwwhww: nitpick: add a newline
```suggestion

        if not cls.verify(message_hash, pubkey, signature, domain):
```

eth2/_utils/bls/__init__.py
```diff
@@ -122,6 +112,9 @@ def validate_multiple(cls,
                           message_hashes: Sequence[Hash32],
                           signature: BLSSignature,
                           domain: Domain) -> None:
+        if cls.backend != NoOpBackend:
+            validate_signature(signature)
+            validate_many_public_keys(pubkeys)
         if not cls.verify_multiple(pubkeys, message_hashes, signature, domain):
```
hwwhww: ditto
```suggestion

        if not cls.verify_multiple(pubkeys, message_hashes, signature, domain):
```

eth2/_utils/bls/backends/__init__.py
```diff
@@ -17,9 +17,15 @@
...
+    AVAILABLE_BACKENDS += (MilagroBackend,)
+except ImportError:
+    DEFAULT_BACKEND = PyECCBackend
```
hwwhww: Maybe assign `DEFAULT_BACKEND` to  `PyECCBackend` when declaring it.
```python
DEFAULT_BACKEND = PyECCBackend  # type: Type[BaseBLSBackend]
```

then

```python
try:
    from .milagro import MilagroBackend
    DEFAULT_BACKEND = MilagroBackend  # set default to MilagroBackend
    AVAILABLE_BACKENDS += (MilagroBackend,)
except ImportError:
    pass

try:
    from .chia import ChiaBackend
    AVAILABLE_BACKENDS += (ChiaBackend,)
except ImportError:
    pass
```

eth2/_utils/bls/validation.py
```diff
@@ -3,21 +3,46 @@
...
+    for pubkey in pubkeys:
+        validate_public_key(pubkey, allow_empty=True)
```
hwwhww: What's the reason for only setting `allow_empty=True` here?

eth2/_utils/bls/validation.py
ChihChengLiang: `validate_public_keys` would look very similar to `validate_public_key`, not sure if we want that. That's why making it `validate_many_public_keys`.

eth2/_utils/bls/validation.py
ChihChengLiang: In the codebase, it is not likely to produce an empty public key for the single verification.
But empty it is possible for `verify_multiple`, [`validate_indexed_attestation_aggregate_signature`](https://github.com/ethereum/trinity/blob/f7a4ec6259e2ea31e89abe37d6fecac3e28a1e0e/eth2/beacon/attestation_helpers.py#L62) is an example.

ChihChengLiang:
Merging now. Post review welcomed.

eth2/_utils/bls/validation.py
ralexstokes: `EMPTY_PUBKEY` is `b'\x00' * 48` right? can you elaborate where we are dealing w/ `EMPTY_PUBKEY`s in actual cryptographic operations?

ChihChengLiang: Depending on what spec we are following, `EMPTY_PUBKEY` (`EMPTY_SIGNATURE` has a similar issue) could be `b'\x00' * 48`, `b'\xc0' + b'\x00' * 47`, or sometimes not even allowed in different backends.

In most of the cases, `EMPTY_PUBKEY` and `EMPTY_SIGNATURE` should be rejected when seen. Otherwise, the client might verify this dangerous case as True:
- any message
- empty aggregation bits (producing `EMPTY_PUBKEY` in Trinity) 
- `EMPTY_SIGNATURE`.

But in the case of `verify_multiple`, it is improbable but possible that 128 validators all signed the same custody bit, producing an empty public key for another custody bit.

PR #955, opened 2019-08-20
Add run_service helper for legacy BaseService

pipermerriam:
### What was wrong?

In some cases, we want to run a service for a more constrained period with control over the lifecycle.

### How was it fixed?

Add a `run_service` that works for the legacy `p2p.service.BaseService` and use it in the various places where it can cleanup the code.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

mastiff-AP-VZGHOK-590sm12913

### DIFF
--- a/p2p/tools/factories/peer.py
+++ b/p2p/tools/factories/peer.py
@@ -11,6 +11,7 @@ from eth_keys import keys
 
 from p2p.abc import NodeAPI
 from p2p.peer import BasePeer, BasePeerContext, BasePeerFactory
+from p2p.service import run_service
 
 from p2p.tools.paragon import ParagonPeer, ParagonContext, ParagonPeerFactory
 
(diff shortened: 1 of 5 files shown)


p2p/tools/factories/peer.py
```diff
@@ -74,31 +75,8 @@
...
-    alice_task = asyncio.ensure_future(alice.run())
-    bob_task = asyncio.ensure_future(bob.run())
-
-    await asyncio.wait_for(asyncio.gather(
-        alice.events.started.wait(),
-        bob.events.started.wait(),
-    ), timeout=1)
-
-    try:
+    async with run_service(alice), run_service(bob):
```
carver: Thoughts on making this a member method? Something like:
```suggestion
    async with alice.spawn(), bob.spawn():
```

pipermerriam: I'm gonna hold off for now on this.

https://github.com/ethereum/trinity/blob/b912a5ede37d6e9efc1137fc599561e4c7daed85/p2p/trio_service.py#L251

It keeps it congruent with the `trio` based service API which I like because it means less code changes when we switch from one to the other.

PR #956, opened 2019-08-20
Add connection abstraction to devp2p networking

pipermerriam:
Extracted from #684 

Builds on #887

### What was wrong?

As a continuation of #684 we previously implemented the `Multiplexer` and new facilities to handle the P2P handshake under `p2p.handshake`.  Now we need something *active* (aka a Service) to manage the multiplexed connection and allow multiple consumers for each protocol or command.

### How was it fixed?

The implements a `Connection` service (which is implemented but not yet used by the `p2p.BasePeer`.  This service houses a `Multiplexer` and exposes the following APIs.

1. Mirrors access to the base `p2p` protocol from the `Multiplexer`.
2. Mirrors access to the `remote` from the `Multiplexer`
3. Meta-data about the remote connection like their public key, client version string, listen port, etc.
4. Meta-data about the handshake via the handshake receipts.
5. APIs for registering handlers for individual Commands or full Protocols.  This can be thought of as similar to the existing peer subscription API.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

extra-large-dog-breeds-great-dane

### DIFF
--- a/p2p/abc.py
+++ b/p2p/abc.py
@@ -3,6 +3,8 @@ from typing import (
     Any,
     AsyncContextManager,
     AsyncIterator,
+    Awaitable,
+    Callable,
     ClassVar,
     Dict,
     Generic,
@@ -21,7 +23,8 @@ from cancel_token import CancelToken
 
 from eth_keys import datatypes
 
-from p2p.typing import Capability, Payload, Structure
+from p2p.disconnect import DisconnectReason
+from p2p.typing import Capability, Capabilities, Payload, Structure
 
 if TYPE_CHECKING:
     from p2p.p2p_proto import (  # noqa: F401
@@ -320,4 +323,78 @@ class MultiplexerAPI(ABC):
     # Message reading and streaming API
     #
     def multiplex(self) -> AsyncContextManager[None]:
+        ...
+
+
+class HandlerSubscriptionAPI:
+    @abstractmethod
+    def cancel(self) -> None:
+        ...
+
+
+class ConnectionAPI(ABC):
+    disconnect_reason: DisconnectReason = None
+
+    #
+    # Primary properties of the connection
+    #
+    @property
+    @abstractmethod
+    def is_dial_in(self) -> bool:
+        ...
+
+    @property
+    @abstractmethod
+    def remote(self) -> NodeAPI:
+        ...
+
+    #
+    # Subscriptions/Handler API
+    #
+    @abstractmethod
+    def add_protocol_handler(self,
+                             protocol_type: Type[ProtocolAPI],
+                             handler_fn: Callable[[CommandAPI, Payload], Awaitable[Any]],
+                             ) -> HandlerSubscriptionAPI:
+        ...
+
+    @abstractmethod
+    def add_command_handler(self,
+                            command_type: Type[CommandAPI],
+                            handler_fn: Callable[[Payload], Awaitable[Any]],
+                            ) -> HandlerSubscriptionAPI:
+        ...
+
+    #
+    # Access to underlying Multiplexer
+    #
+    @abstractmethod
+    def get_multiplexer(self) -> MultiplexerAPI:
+        ...
+
+    #
+    # Base Protocol shortcuts
+    #
+    @abstractmethod
+    def get_base_protocol(self) -> 'BaseP2PProtocol':
+        ...
+
+    @property
+    @abstractmethod
+    def remote_capabilities(self) -> Capabilities:
+        ...
+
+    @property
+    @abstractmethod
+    def remote_p2p_version(self) -> int:
+        ...
+
+    @property
+    @abstractmethod
+    def client_version_string(self) -> str:
+        ...
+
+    @property
+    @abstractmethod
+    def safe_client_version_string(self) -> str:
         ...
--- a/p2p/abc.py
+++ b/p2p/abc.py
@@ -23,7 +23,6 @@ from cancel_token import CancelToken
 
 from eth_keys import datatypes
 
-from p2p.disconnect import DisconnectReason
 from p2p.typing import Capability, Capabilities, Payload, Structure
 
 if TYPE_CHECKING:
@@ -333,8 +332,6 @@ class HandlerSubscriptionAPI:
 
 
 class ConnectionAPI(ABC):
-    disconnect_reason: DisconnectReason = None
-
     #
     # Primary properties of the connection
     #
--- a/p2p/connection.py
+++ b/p2p/connection.py
@@ -1,7 +1,7 @@
 import asyncio
 import collections
 import functools
-from typing import Any, Awaitable, Callable, DefaultDict, Set, Tuple, TypeVar, Type
+from typing import Any, Awaitable, Callable, DefaultDict, Set, Tuple, Type
 
 from eth_keys import keys
 
@@ -13,7 +13,6 @@ from p2p.abc import (
     ProtocolAPI,
     ConnectionAPI,
 )
-from p2p.disconnect import DisconnectReason
 from p2p.exceptions import (
     UnknownProtocol,
     UnknownProtocolCommand,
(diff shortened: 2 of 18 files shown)


pipermerriam:
Blocked by #887

p2p/connection.py
```diff
@@ -0,0 +1,215 @@
...
+    #
+    # Subscriptions/Handler API
+    #
+    async def _feed_protocol_handlers(self, protocol: ProtocolAPI) -> None:
+        # do not start consuming from the protocol stream until
+        # `start_protocol_streams` has been called and the multiplexer is
+        # active.
+        try:
+            await asyncio.wait_for(asyncio.gather(self._handlers_ready.wait()), timeout=10)
+        except TimeoutError:
```
pipermerriam: I think this actually has to be `asyncio.TimeoutError` because it isn't the same as the built-ins `TimeoutError`

p2p/connection.py
```diff
@@ -0,0 +1,215 @@
...
+            self.logger.info('Timedout waiting for handler ready signal')
+            raise TimeoutError(
+                "The handlers ready event was never set.  Ensure that "
+                "`Connection.start_protocol_streams()` is being called"
+            )
+
+        # we don't need to use wait_iter here because the multiplexer does it
+        # for us.
+        async for cmd, msg in self._multiplexer.stream_protocol_messages(protocol):
+            self.logger.debug('Handling command: %s', type(cmd))
```
pipermerriam: probably too noisy.

p2p/tools/factories/multiplexer.py
```diff
@@ -46,15 +60,22 @@ def MultiplexerPairFactory(*,
         in zip(protocol_types, cmd_id_offsets)
     )
 
-    alice_p2p_protocol = P2PProtocol(alice_transport, 0, snappy_support)
+    p2p_protocol_class: Type[BaseP2PProtocol]
+
+    if snappy_support:
+        p2p_protocol_class = P2PProtocol
+    else:
+        p2p_protocol_class = P2PProtocolV4
```
pipermerriam: These changes can get extracted into a separate PR.

pipermerriam:
Requested review from both @lithp and @carver just to get both of you to look at it.  I think Jason has seen this code before when I was presenting it in a preliminary fashion, but Brian has reviewed some of the recent pull requests in this feature line.

p2p/abc.py
```diff
@@ -321,3 +324,77 @@ def stream_protocol_messages(self,
     #
     def multiplex(self) -> AsyncContextManager[None]:
         ...
+
+
+class HandlerSubscriptionAPI:
```
lithp: Some alternatives:
- Making this more abstract: Calling it `Cancellable` might make it more clear what you can do with these objects.
- Making it more concrete: mypy kind of gets in the way but this function also could return the function directly, first class functions are pretty dope! `CancelFn = Callable[[], None]`
- It doesn't look like these subscriptions are ever used, I bet we could get away with not providing a way to cancel subscriptions for quite a while.

p2p/connection.py
```diff
@@ -0,0 +1,214 @@
...
+class HandlerSubscription(HandlerSubscriptionAPI):
+    def __init__(self, remove_fn: Callable[[], Any]) -> None:
+        self._remove_fn = remove_fn
+
+    def cancel(self) -> None:
+        self._remove_fn()
+
+
+class Connection(ConnectionAPI, BaseService):
+    disconnect_reason: DisconnectReason = None
```
lithp: I don't think this attribute is ever set or read from

p2p/connection.py
```diff
@@ -0,0 +1,214 @@
...
+)
+from p2p.service import BaseService
+from p2p.p2p_proto import (
+    BaseP2PProtocol,
+)
+from p2p.typing import Capabilities, Payload
+
+
+TProtocol = TypeVar('TProtocol', bound=ProtocolAPI)
+TReceipt = TypeVar('TReceipt', bound=HandshakeReceipt)
```
lithp: Neither of these are used anywhere

p2p/connection.py
```diff
@@ -0,0 +1,214 @@
...
+        self._command_handlers = collections.defaultdict(set)
+
+        # An event that controls when the connection will start reading from
+        # the individual multiplexed protocol streams and feeding handlers.
+        # This ensures that the connection does not start consuming messages
+        # before all necessary handlers have been added
+        self._handlers_ready = asyncio.Event()
+
+    def start_protocol_streams(self) -> None:
+        self._handlers_ready.set()
```
lithp: Why not hold off on calling `_run` until all the correct handlers have been added?

p2p/connection.py
```diff
@@ -0,0 +1,214 @@
...
+        return not self.is_dial_out
+
+    @property
+    def remote(self) -> NodeAPI:
+        return self._multiplexer.remote
+
+    async def _run(self) -> None:
+        async with self._multiplexer.multiplex():
+            for protocol in self._multiplexer.get_protocols():
+                self.run_daemon_task(self._feed_protocol_handlers(protocol))
```
lithp: The multiplexer goes through a lot of effort to disaggregate these streams, and then it spends most of its life here in `Connection` where those streams are brought back together again! It might simplify things for the handshaking code could use something less general than the full `Multiplexer`, and for `Connection` to take a transport directly.

p2p/connection.py
```diff
@@ -0,0 +1,214 @@
...
+    @property
+    def remote_public_key(self) -> keys.PublicKey:
+        return keys.PublicKey(self._devp2p_receipt.remote_public_key)
+
+    @property
+    def client_version_string(self) -> str:
+        return self._devp2p_receipt.client_version_string
+
+    @property
+    def safe_client_version_string(self) -> str:
```
lithp: I'm not sure where this will be used so maybe this is bad feedback, but you probably still want to make the string printable, even if it's under 256 characters!

There's also no test for this behavior! If you drop the lines which checks for isprintable and always return the string (never calling `repr`) the tests still pass.

p2p/tools/factories/connection.py
```diff
@@ -0,0 +1,131 @@
...
+from p2p.service import run_service
+
+from .cancel_token import CancelTokenFactory
+from .multiplexer import MultiplexerPairFactory
+from .p2p_proto import DevP2PHandshakeParamsFactory
+from .transport import MemoryTransportPairFactory
+
+
+@asynccontextmanager
+async def ConnectionPairFactory(*,
```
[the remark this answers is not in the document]
pipermerriam: Yes but that happens in the next pull request #962 .

p2p/abc.py
pipermerriam: In a subsequent PR I'm expanding this to be a context manager so that it cancels the handler on exit so I'm going to keep it as-is since that context management API proves useful for things that have a shorter lifecycle than the connection and thus need to have their handlers reliably de-registered.

p2p/connection.py
pipermerriam: will remove.  This is a carryover from when I had a `Connection.disconnect` method but I removed it late in the process because I wanted to keep the `Connection` *pure* and devoid of any actual business logic.

p2p/connection.py
pipermerriam: In theory that would work, except that it ends up working out that the places where the handlers get registered isn't at the right level to also be in charge of running the connection (the `Connection` running part happens at a level higher than the `Peer` and so the peer can't be responsible for running it).

This mechanism is pretty simple and it's working well but I'm happy to re-visit it if something better shows up later.

p2p/connection.py
pipermerriam: I think you might be right that we could adjust the internal `Connection` to just directly use `p2p.multiplexer.stream_transport_messages` and only use the multiplexer during the handshake.  I'm gonna hold off on making this change for now but good suggestion (I'm gonna open an issue to track)

p2p/connection.py
pipermerriam: :+1: will get this updated.  It's used for logging messages so that if someone sends us a spammy long client string with a bunch of newlines and ascii stuff to advertise their ICO we don't end up with it clogging up our logging messages.

PR #958, opened 2019-08-20
Beam backfill

### LINKED ISSUE #951: Simple Beam Sync backfill
### What is wrong?

Related to #854 

We need some mechanism for backfill. Also, we need some way to make sure we're regularly inspecting peers to determine their `GetNodeData` RTT. These two things line up somewhat.

### How can it be fixed

Do a *very* simple backfill approach, iterating through the peers, asking them for nodes. (maybe not asking for too many nodes, since that might cause them to spend more time on disk. This requires more investigation)

This will also help establish the RTT of all the peers.

We might intentionally skip the lowest RTT peer when we loop through, since that one should be dedicated to the most time-sensitive Beam Sync requests.

carver:
### What was wrong?

Fix #951 

### How was it fixed?

Pretty much as planned.

Some stuff came up. Seems like the lock approach to make sure we don't ask the same peer for data twice is pretty finicky. It was triggering timeouts, even validation failures. So there's a hack to prevent asking for backfill nodes from a peer that already has an outstanding `GetNodeData` request.

### To-Do

- [x] Stop interfering with the primary beam sync, maybe maintaining a blacklist of peers that have recently been used by the beam sync importer. Currently, backfill is causing beam sync to have growing lag
- [x] Still getting some timeout errors on the lock itself, which assumes all requesters use the same timeout.
- [x] ~~Maybe a more robust locking solution that doesn't break so easily~~ :football: to later
- [x] It's not finding out data as quickly as I'd like, track down why. (First guess is ~~the sleep when there are outstanding requests to all the peers~~ the code was asking each peer in series and waiting for the previous one to finish :man_facepalming: )
- [x] ~~A better story for what happens when we finish backfill. Right now it just keeps scanning the DB over and over~~ :football: to later

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/backfill.py
+++ b/trinity/sync/beam/backfill.py
@@ -166,12 +166,12 @@ class BeamStateBackfill(BaseService, PeerSubscriber, QueenTrackerAPI):
             old_queen = self._queen_peer
             self._update_queen(peer)
             if peer == self._queen_peer:
-                self.logger.warning("Switching queen peer from %s to %s", old_queen, peer)
+                self.logger.debug("Switching queen peer from %s to %s", old_queen, peer)
                 continue
 
             if peer.requests.get_node_data.is_requesting:
                 # skip the peer if there's an active request
-                self.logger.warning("Skipping active peer %s", peer)
+                self.logger.debug("Backfill is skipping active peer %s", peer)
                 self.call_later(10, self._waiting_peers.put_nowait, peer)
                 continue
 
@@ -183,7 +183,7 @@ class BeamStateBackfill(BaseService, PeerSubscriber, QueenTrackerAPI):
             if len(on_deck) == 0:
                 # Nothing left to request, break and wait for new data to come in
                 self._waiting_peers.put_nowait(peer)
-                self.logger.warning("Backfill is waiting for more hashes to arrive")
+                self.logger.debug("Backfill is waiting for more hashes to arrive")
                 await self.sleep(2)
                 continue
 
@@ -317,14 +317,14 @@ class BeamStateBackfill(BaseService, PeerSubscriber, QueenTrackerAPI):
             msg += "  new=%d" % self._num_added
             msg += "  missed=%d" % self._num_missed
             msg += "  queen=%s" % self._queen_peer
-            self.logger.info("Beam-Backfill: %s", msg)
+            self.logger.debug("Beam-Backfill: %s", msg)
 
             self._num_added = 0
             self._num_missed = 0
 
             # log peer counts
             show_top_n_peers = 3
-            self.logger.info(
+            self.logger.debug(
                 "Beam-Backfill-Peer-Usage-Top-%d: %s",
                 show_top_n_peers,
                 self._num_requests_by_peer.most_common(show_top_n_peers),
--- a/trinity/sync/beam/backfill.py
+++ b/trinity/sync/beam/backfill.py
@@ -1,0 +1,284 @@
+import asyncio
+from collections import Counter
+import typing
+from typing import (
+    FrozenSet,
+    Iterable,
+    List,
+    Set,
+    Tuple,
+    Type,
+)
+
+from cancel_token import CancelToken, OperationCancelled
+from eth.abc import AtomicDatabaseAPI
+from eth_typing import Hash32
+import rlp
+
+from p2p.abc import CommandAPI
+from p2p.exceptions import BaseP2PError, PeerConnectionLost
+from p2p.peer import BasePeer, PeerSubscriber
+from p2p.service import BaseService
+
+from trinity.protocol.eth.commands import (
+    NodeData,
+)
+from trinity.protocol.eth.peer import ETHPeer, ETHPeerPool
+from trinity.protocol.common.abc import (
+    PerformanceAPI,
+)
+from trinity.sync.beam.constants import (
+    GAP_BETWEEN_TESTS,
+)
+from trinity.sync.common.peers import WaitingPeers
+
+REQUEST_SIZE = 16
+
+
+def _get_items_per_second(tracker: PerformanceAPI) -> float:
+    return -1 * tracker.items_per_second_ema.value
+
+
+def _peer_sort_key(peer: ETHPeer) -> float:
+    return _get_items_per_second(peer.requests.get_node_data.tracker)
+
+
+class BeamStateBackfill(BaseService, PeerSubscriber):
+    """
+    Use a very simple strategy to fill in state in the background.
+
+    Ask each peer in sequence for some nodes, ignoring the lowest RTT node.
+    Reduce memory pressure by using a depth-first strategy.
+
+    An intended side-effect is to build & maintain an accurate measurement of
+    the round-trip-time that peers take to respond to GetNodeData commands.
+    """
+    # We are only interested in peers entering or leaving the pool
+    subscription_msg_types: FrozenSet[Type[CommandAPI]] = frozenset()
+
+    # This is a rather arbitrary value, but when the sync is operating normally we never see
+    # the msg queue grow past a few hundred items, so this should be a reasonable limit for
+    # now.
+    msg_queue_maxsize: int = 2000
+
+    _total_processed_nodes = 0
+    _num_added = 0
+    _num_missed = 0
+    _report_interval = 10
+
+    _num_requests_by_peer: typing.Counter[ETHPeer]
+
+    def __init__(
+            self,
+            db: AtomicDatabaseAPI,
+            peer_pool: ETHPeerPool,
+            token: CancelToken = None) -> None:
+
+        # in case there is no token set, make sure this gets cancelled when the peer pool does
+        if token is None:
+            token = peer_pool.cancel_token
+
+        super().__init__(token=token)
+        self._db = db
+
+        # Pending nodes to download
+        self._node_hashes: List[Hash32] = []
+
+        self._peer_pool = peer_pool
+        self._available_peers = asyncio.Event()
+        # The best peer gets skipped for backfill, because we prefer to use it for
+        #   urgent beam sync nodes
+        self._queen_peer: ETHPeer = None
+        self._waiting_peers = WaitingPeers[ETHPeer](NodeData, sort_key=_get_items_per_second)
+
+        self._is_missing: Set[Hash32] = set()
+
+        self._num_requests_by_peer = Counter()
+
+    def _update_queen(self, peer: ETHPeer) -> None:
+        if self._queen_peer is None:
+            self._queen_peer = peer
+        elif _peer_sort_key(peer) < _peer_sort_key(self._queen_peer):
+            self._waiting_peers.put_nowait(self._queen_peer)
+            self._queen_peer = peer
+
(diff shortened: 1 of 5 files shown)


trinity/sync/beam/backfill.py
```diff
@@ -0,0 +1,226 @@
...
+
+            # collect node hashes that might be missing
+            await self._walk()
+
+            peer = await self._waiting_peers.get_fastest()
+            await self._update_queen(peer)
+            if peer == self._queen_peer:
+                continue
+
+            if peer.requests.get_node_data.is_requesting:
```
pipermerriam: I'm not sure it's the right tool for this but just in case you missed it.

https://github.com/ethereum/trinity/blob/7f0350d55760f34ef03124651187e6b9ebf3cc33/p2p/resource_lock.py

Might be a more controlled way to limit your concurrency level.

carver:
I'm in the process of pulling things out into their own PRs:
- #1034  
- #1035 
- #1043

trinity/sync/beam/backfill.py
```diff
@@ -0,0 +1,152 @@
...
+            # branch node
+            return set(node_hash for node_hash in decoded_node[:16] if len(node_hash) == 32)
+        elif len(decoded_node) == 2 and len(decoded_node[1]) == 32:
+            # leaf or extension node
+            return {decoded_node[1]}
+        else:
+            # final value, ignore
+            return set()
+
+    def set_root_hash(self, root_hash: Hash32) -> None:
```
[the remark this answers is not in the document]
carver: I think the ideal would be to maintain a limit of accounts that have been explored. In other words, to say we have locally walked all addresses above `0xFFFFFF03784098472347`.

The neat thing is that if beam sync can keep up, then we only have to make a single sweep of the state DB.

Then, every N blocks we pivot (just like fast sync pivot), and reset the pending queue. We re-walk from the new root down to the limit account, and keep walking down toward account 0 from there.

(This wording pretends that the accounts aren't hashed before putting them in the state trie, but that's irrelevant)

carver:
#1045

trinity/sync/beam/backfill.py
```diff
@@ -0,0 +1,226 @@
...
+from trinity.sync.common.peers import WaitingPeers
+
+REQUEST_SIZE = 8
+
+
+def _get_rtt(tracker: BasePerformance) -> float:
+    """
+    Extract the round trip time from the peer's tracker.
+    """
+    return -1 * tracker.items_per_second_ema.value
```
carver: Reminder to self to remove this, and use the default items-per-second.

trinity/sync/beam/backfill.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+def _get_peer_rtt(peer: ETHPeer) -> float:
+    return _get_rtt(peer.requests.get_node_data.tracker)
+
+
+GAP_BETWEEN_TESTS = 2
```
carver: How much shorter can I get this without impacting beam sync? Maybe 0.25s?

trinity/sync/beam/backfill.py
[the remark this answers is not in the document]
pipermerriam: I like this.  Nice and simple with minimalistic tracking.

I'm wondering if we can tweak this in some way to give it the same swarming behavior that exists with the way warm trie nodes are retrieved.  It'd be nice if all of the nodes could somehow converge so that two nodes that start syncing a few hours apart end up *mostly* requesting the same data.

Maybe instead of doing a sequential walk of the trie we could track the ranges that have been walked and when we pivot, we randomly choose the next range using something like the block hash at the pivot point.  Then clients who at least are syncing against the same block would likely end up requesting the same data.  

I think the behavior improves if we can improve the likelihood of clients pivoting to the same block.  My first thought is that we choose pivot points probabalistically with some sort of weighted choice which favors blocks which are evenly divisible by some fixed-ish value.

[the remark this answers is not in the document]
carver: > I'm wondering if we can tweak this in some way to give it the same swarming behavior that exists with the way warm trie nodes are retrieved. It'd be nice if all of the nodes could somehow converge so that two nodes that start syncing a few hours apart end up _mostly_ requesting the same data.

That's a really cool idea.

> Maybe instead of doing a sequential walk of the trie we could track the ranges that have been walked and when we pivot, we randomly choose the next range using something like the block hash at the pivot point. Then clients who at least are syncing against the same block would likely end up requesting the same data.

Yup, I like this approach. The state root is something that the backfill needs access to anyway, so I like potentially using that as a pseudo-random source. Of course, some peers will get lucky and race ahead, but at least at the beginning of the pivot, all peers are requesting roughly the same nodes.

> I think the behavior improves if we can improve the likelihood of clients pivoting to the same block. My first thought is that we choose pivot points probabalistically with some sort of weighted choice which favors blocks which are evenly divisible by some fixed-ish value.

Yeah, definitely. I *think* we can even get simpler and just pivot on block `block.number % 75 == 0`. Since the predictability of data being requested (or not requested) can only be used to attack peers who are mid-sync, and probably requires spending significant assets, I don't think it's a critical attack vector.

carver:
Some more that I will rebase on when they are closed:
- #1073 
- #1074 
- #1075

carver:
Still more ideas of things to work on, but I think this is a good chunk for review, and it seems to help identify the peers that can provide the node data quickly. If we can get this merged early enough tomorrow, then I can start a release to go out tomorrow. Then I'll start writing up Beam Sync, which might require a series of posts to adequately explain.

pipermerriam:
> If we can get this merged early enough tomorrow, then I can start a release to go out tomorrow

I'll review this evening after kids are down to bed.

trinity/sync/beam/backfill.py
```diff
@@ -0,0 +1,319 @@
...
+
+def _get_items_per_second(tracker: BasePerformance) -> float:
+    return -1 * tracker.items_per_second_ema.value
+
+
+def _peer_sort_key(peer: ETHPeer) -> float:
+    return _get_items_per_second(peer.requests.get_node_data.tracker)
+
+
+GAP_BETWEEN_TESTS = 0.25
```
pipermerriam: Probably deserves a comment.

trinity/sync/beam/backfill.py
```diff
@@ -0,0 +1,319 @@
...
+    # the msg queue grow past a few hundred items, so this should be a reasonable limit for
+    # now.
+    msg_queue_maxsize: int = 2000
+
+    _total_processed_nodes = 0
+    _num_added = 0
+    _num_missed = 0
+    _report_interval = 10
+
+    _num_requests_by_peer: typing.Counter[ETHPeer]
```
pipermerriam: Maybe use a `weakref.WeakKeyDictionary` so that these values get cleaned up when we disconnect from a peer?

Not a big deal since the actual memory footprint of this will still be very small over a long period of time.

trinity/sync/beam/backfill.py
```diff
@@ -0,0 +1,319 @@
...
+
+    def __init__(
+            self,
+            db: AtomicDatabaseAPI,
+            peer_pool: ETHPeerPool,
+            token: CancelToken = None) -> None:
+        super().__init__(token=token)
```
pipermerriam: Maybe add some logic so that `if token is None: token = peer_pool.cancel_token` so that it gets cancelled when the peer pool cancels.  Otherwise, the omission of a properly linked token would cause this service to hang on shutdown.

trinity/sync/beam/backfill.py
```diff
@@ -0,0 +1,319 @@
...
+        self._waiting_peers = WaitingPeers[ETHPeer](NodeData, sort_key=_get_items_per_second)
+
+        self._is_missing: Set[Hash32] = set()
+
+        self._num_requests_by_peer = Counter()
+
+    def _update_queen(self, peer: ETHPeer) -> None:
+        if self._queen_peer is None:
+            self._queen_peer = peer
+        elif _peer_sort_key(peer) < _peer_sort_key(self._queen_peer):
```
pipermerriam: Without this line, if this function was called with the current queen (`peer == self._queen_peer`) then we'd end up in a situation where the queen is also in the backfill rotation.  I think this comparison *implicitly* keeps this situation from happening but it'd be an easy thing to inadvertently change in a refactor.

Maybe add a check that `peer != self._queen_peer` or an early exit if that is true?

trinity/sync/beam/backfill.py
```diff
@@ -0,0 +1,319 @@
...
+            self.run_task(self._make_request(peer, on_deck))
+
+    async def _make_request(self, peer: ETHPeer, request_hashes: Tuple[Hash32, ...]) -> None:
+        self._num_requests_by_peer[peer] += 1
+        try:
+            nodes = await peer.requests.get_node_data(request_hashes)
+        except asyncio.TimeoutError:
+            self._node_hashes.extend(request_hashes)
+            self.call_later(GAP_BETWEEN_TESTS * 2, self._waiting_peers.put_nowait, peer)
+        except (PeerConnectionLost, OperationCancelled, CancelledError):
```
pipermerriam: supressing `CancelledError` is typicall wrong.  Is this intentional?

trinity/sync/beam/backfill.py
```diff
@@ -0,0 +1,319 @@
...
+        except (BaseP2PError, Exception) as exc:
+            self.logger.info("Unexpected err while getting background nodes from %s: %s", peer, exc)
+            self.logger.debug("Problem downloading background nodes from peer...", exc_info=True)
+            self._node_hashes.extend(request_hashes)
+            self.call_later(GAP_BETWEEN_TESTS * 2, self._waiting_peers.put_nowait, peer)
+        else:
+            self.call_later(GAP_BETWEEN_TESTS, self._waiting_peers.put_nowait, peer)
+            self._insert_results(request_hashes, nodes)
+        finally:
+            pass
```
pipermerriam: remove empty `finaly`?

trinity/sync/beam/backfill.py
```diff
@@ -0,0 +1,319 @@
...
+
+        self._peer_pool = peer_pool
+        self._available_peers = asyncio.Event()
+        # The best peer gets skipped for backfill, because we prefer to use it for
+        #   urgent beam sync nodes
+        self._queen_peer: ETHPeer = None
+
+        self._waiting_peers = WaitingPeers[ETHPeer](NodeData, sort_key=_get_items_per_second)
+
+        self._is_missing: Set[Hash32] = set()
```
pipermerriam: It *might* be viable to use a `weakref.WeakSet` and have this data structure automatically clean itself up.... but I'm actually kind of skeptical because I'm not sure about the lifecycle of the variables we're using to track keys...

trinity/sync/beam/backfill.py
carver: Actually:

- the `Counter` gets `clear()`d every `_report_interval`
- the backfiller makes meaningful use of the `Counter` API (eg~ `most_common()`)
- I don't see a builtin way to do a "weak key reference Counter"

So I think I'll leave it as-is for now.

PR #960, opened 2019-08-21
Do some logging in ENR DB

jannikluhn:
### What was wrong?

The ENR DB doesn't log anything.

### How was it fixed?

Log something when inserting, removing, or updating ENRs.

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/enr_db.py
+++ b/p2p/discv5/enr_db.py
@@ -51,11 +51,12 @@ class MemoryEnrDb(BaseEnrDb):
         self.validate_identity_scheme(enr)
 
         if await self.contains(enr.node_id):
-            raise ValueError(f"ENR with nodeid {enr.node_id} already exists.")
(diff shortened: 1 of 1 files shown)


p2p/discv5/enr_db.py
```diff
@@ -46,16 +53,30 @@ def __init__(self, identity_scheme_registry: IdentitySchemeRegistry):
         if await self.contains(enr.node_id):
             raise ValueError(f"ENR with nodeid {enr.node_id} already exists.")
         else:
+            self.logger.debug(
+                f"Inserting new ENR of {encode_hex(enr.node_id)} with sequence number "
```
pipermerriam: See https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#formatting-log-messages for reasoning behind always using `%` formatting style and positional arguments for logging messages.

PR #961, opened 2019-08-21
Peer Packer

jannikluhn:
This PR adds the `PeerPacker`, the class responsible for packing messages into packets and unpacking packets to messages for a single peer. Unfortunately, it's rather big, even after a few iterations of stripping it down. I thought about making it even smaller by separating handshake related stuff from normal operation, but this wouldn't help much (because normal operation is pretty trivial) and increase complexity (because you have now two components that are strongly linked and have to communicate with each other). Fortunately, it's not that complex because most of what it has to do is feed the `HandshakeInitiator` and `HandshakeRecipient` class with the packets it received on some channel.

### External interface

The `PeerPacker` is supposed to be created and owned by a `Packer` class that manages one `PeerPacker` for every peer. The `Packer` sends incoming packets to the peer's `incoming_packet_receive_channel` if they come from the right sender, or if `PeerPacker.is_expecting_handshake_packet` returns `True` (we need this because `WhoAreYou` packets don't have an explicit sender). The `PeerPacker` also receives messages it should pack and send to the peer in the `outgoing_messages_receive_channel`. Encrypted packets that should be sent over the wire go to the `outgoing_packet_send_channel`, decrypted messages go to the `incoming_message_send_channel`.

The handshake can fail in a couple of ways with a `HandshakeFailure` exception which the packer should catch.

### Internals

There's some internal state:
- `handshake_participant` which is either `None` (before or after the handshake), a `HandshakeInitiator`, or a `HandshakeRecipient`
- `session_keys` which is `None` before or during the handshake, or set after the handshake
- `outgoing_message_backlog`: a list of messages we have to send to the peer, but we haven't yet because the handshake is still ongoing

At all times, the peer packer is either `is_pre_handshake`, `is_during_handshake`, or `is_post_handshake`, depending on which of `handshake_participant` and `session_keys` is set. Incoming packets and outgoing messages are dispatched to three different methods each based on this state. Incoming packets/outgoing messages generally change the state.

We don't store our or the peer ENR directly, but get it from the db instead, because it might change over time (e.g. if we get a new IP).

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/packer.py
+++ b/p2p/discv5/packer.py
@@ -1,0 +1,445 @@
+import logging
+from typing import (
+    List,
+    Optional,
+)
+
+from eth_utils import (
+    encode_hex,
+    ValidationError,
+)
+
+import trio
+from trio.abc import (
+    ReceiveChannel,
+    SendChannel,
+)
+
+from p2p.trio_service import Service
+
+from p2p.discv5.abc import (
+    EnrDbApi,
+    HandshakeParticipantAPI,
+)
+from p2p.discv5.channel_services import (
+    Endpoint,
+    IncomingMessage,
+    IncomingPacket,
+    OutgoingMessage,
+    OutgoingPacket,
+)
+from p2p.discv5.enr import (
+    ENR,
+)
+from p2p.discv5.handshake import (
+    HandshakeInitiator,
+    HandshakeRecipient,
+)
+from p2p.discv5.messages import (
+    BaseMessage,
+    MessageTypeRegistry,
+)
+from p2p.discv5.packets import (
+    AuthTagPacket,
+    get_random_auth_tag,
+)
+from p2p.discv5.tags import (
+    compute_tag,
+)
+from p2p.discv5.typing import (
+    NodeID,
+    Nonce,
+    SessionKeys,
+)
+
+from p2p.exceptions import (
+    DecryptionError,
+    HandshakeFailure,
+)
+
+
+class PeerPacker(Service):
+    handshake_participant: Optional[HandshakeParticipantAPI] = None
+    session_keys: Optional[SessionKeys] = None
+
+    def __init__(self,
+                 local_private_key: bytes,
+                 local_node_id: NodeID,
+                 remote_node_id: NodeID,
+                 enr_db: EnrDbApi,
+                 message_type_registry: MessageTypeRegistry,
+                 incoming_packet_receive_channel: ReceiveChannel[IncomingPacket],
+                 incoming_message_send_channel: SendChannel[IncomingMessage],
+                 outgoing_message_receive_channel: ReceiveChannel[OutgoingMessage],
+                 outgoing_packet_send_channel: SendChannel[OutgoingPacket],
+                 ) -> None:
+        self.local_private_key = local_private_key
+        self.local_node_id = local_node_id
+        self.remote_node_id = remote_node_id
+        self.enr_db = enr_db
+        self.message_type_registry = message_type_registry
+
+        self.incoming_packet_receive_channel = incoming_packet_receive_channel
+        self.incoming_message_send_channel = incoming_message_send_channel
+        self.outgoing_message_receive_channel = outgoing_message_receive_channel
+        self.outgoing_packet_send_channel = outgoing_packet_send_channel
+
+        self.logger = logging.getLogger(
+            f"p2p.discv5.packer.PeerPacker[{encode_hex(remote_node_id)[2:10]}]"
+        )
+
+        self.outgoing_message_backlog: List[OutgoingMessage] = []
+
+    async def run(self) -> None:
+        async with self.incoming_packet_receive_channel, self.incoming_message_send_channel,  \
+                self.outgoing_message_receive_channel, self.outgoing_packet_send_channel:
+            self.manager.run_task(self.handle_incoming_packets, daemon=True)
+            self.manager.run_task(self.handle_outgoing_messages, daemon=True)
+            await self.manager.wait_stopped()
+
+    async def handle_incoming_packets(self) -> None:
+        async for incoming_packet in self.incoming_packet_receive_channel:
+            # Handle packets sequentially, so that the rest of the code doesn't have to deal
+            # with multiple packets being processed at the same time.
+            await self.handle_incoming_packet(incoming_packet)
+
+    async def handle_incoming_packet(self, incoming_packet: IncomingPacket) -> None:
+        if self.is_pre_handshake:
+            await self.handle_incoming_packet_pre_handshake(incoming_packet)
+        elif self.is_during_handshake:
+            await self.handle_incoming_packet_during_handshake(incoming_packet)
+        elif self.is_post_handshake:
+            await self.handle_incoming_packet_post_handshake(incoming_packet)
+        else:
+            raise Exception("Invariant: All states handled")
+
+    async def handle_outgoing_messages(self) -> None:
+        async for outgoing_message in self.outgoing_message_receive_channel:
+            # Similar to the incoming packets outgoing messages are processed in sequence, even
+            # though it's not that critical here
+            await self.handle_outgoing_message(outgoing_message)
+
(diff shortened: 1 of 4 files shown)


p2p/discv5/packer.py
```diff
@@ -0,0 +1,435 @@
...
+
+from p2p.exceptions import (
+    DecryptionError,
+    HandshakeFailure,
+)
+
+
+class PeerPacker(Service):
+    def __init__(self,
+                 local_private_key: bytes,
```
pipermerriam: What type of key is this?  Can we use `eth-keys` data structures for this type?

p2p/discv5/packer.py
```diff
@@ -0,0 +1,435 @@
...
+        self.enr_db = enr_db
+        self.message_type_registry = message_type_registry
+
+        self.incoming_packet_receive_channel = incoming_packet_receive_channel
+        self.incoming_message_send_channel = incoming_message_send_channel
+        self.outgoing_message_receive_channel = outgoing_message_receive_channel
+        self.outgoing_packet_send_channel = outgoing_packet_send_channel
+
+        self.logger = logging.getLogger(
+            f"p2p.discv5.packer.PeerPacker[{encode_hex(remote_node_id)[:10]}]"
```
pipermerriam: You might strip the `0x` from the front of this to conserve a little more space.

p2p/discv5/packer.py
```diff
@@ -0,0 +1,435 @@
...
+        )
+
+        self.handshake_participant: Optional[HandshakeParticipantAPI] = None
```
pipermerriam: This could be set at the class level rather than needing it to be set in the constructor.  Same with `session_keys` below.

p2p/discv5/packer.py
```diff
@@ -0,0 +1,435 @@
...
+        self.session_keys: Optional[SessionKeys] = None
+        self.outgoing_message_backlog: List[OutgoingMessage] = []
+
+    async def run(self) -> None:
+        self.manager.run_task(self.handle_incoming_packets)
```
pipermerriam: These should be `daemon=True` right?  Otherwise they would be allowed to finish and the service would continue on.

p2p/discv5/packer.py
```diff
@@ -0,0 +1,435 @@
...
+        self.manager.run_task(self.handle_outgoing_messages)
+        await trio.sleep_forever()
```
pipermerriam: There are a few ways we could do this and it seems like it would be good to settle on a best practice for "wait until the service exits".  I'm inclined to have it be an API on the service manager.  We could use the existing `self.manager.wait_cancelled()`

p2p/discv5/packer.py
```diff
@@ -0,0 +1,435 @@
...
+
+    async def handle_incoming_packets(self) -> None:
+        async with self.incoming_packet_receive_channel, self.incoming_message_send_channel:
```
pipermerriam: curious why this also includes the `self.incoming_message_send_channel`?

p2p/discv5/packer.py
```diff
@@ -0,0 +1,435 @@
...
+            await self.handle_outgoing_message_post_handshake(outgoing_message)
+        else:
+            raise Exception("Invariant: All states handled")
+
+    #
+    # Incoming packet handlers
+    #
+    async def handle_incoming_packet_pre_handshake(self, incoming_packet: IncomingPacket) -> None:
+        if not self.is_pre_handshake:
+            raise ValueError("Can only handle packets pre handshake")
```
pipermerriam: It seems like there may be a race condition that needs to be handled where `is_pre_handshake` was `True` when this handler was called but then the state changes before this coroutine gets executed causing this to now be `False`?

Maybe not since the packages are handled sequentially from the channel.  This is probably something worth commenting since that property of messages being handled sequentially (in order) could change were we to do something like run them in a nursery asynchronously.

p2p/discv5/packer.py
```diff
@@ -0,0 +1,435 @@
...
+                    "Dropping AuthTag packet previously thought to initiate handshake as we have "
+                    "initiated handshake ourselves in the meantime"
+                )
+
+            self.logger.debug("Responding with WhoAreYou packet")
+            await self.send_first_handshake_packet(incoming_packet.sender_endpoint)
+        else:
+            self.logger.debug(
+                "Dropping %s as handshake has not been started yet",
+                incoming_packet.packet.__class__.__name__,
```
pipermerriam: Maybe just define a nice `__str__` method on the `IncomingPacket` class?

p2p/discv5/packer.py
```diff
@@ -0,0 +1,435 @@
...
+            self.logger.debug(
+                "Dropping %s unexpectedly received during handshake",
+                packet.__class__.__name__,
+            )
+            return
+
+        try:
+            handshake_result = self.handshake_participant.complete_handshake(packet)
+        except HandshakeFailure:
+            self.logger.warning("Handshake has failed")
```
pipermerriam: Probably want some more information about which parties the failure was between.

p2p/discv5/packer.py
```diff
@@ -0,0 +1,435 @@
...
+            raise  # let the service fail
+        else:
+            self.logger.info("Handshake was successful")
```
pipermerriam: Same.  Info about who the handshake is with would be a good addition to this message.

p2p/discv5/packer.py
```diff
@@ -0,0 +1,435 @@
...
+            if handshake_result.enr is not None:
+                self.logger.debug("Updating ENR in DB with %r", handshake_result.enr)
+                await self.enr_db.insert_or_update(handshake_result.enr)
+
+            if handshake_result.auth_header_packet is not None:
+                outgoing_packet = OutgoingPacket(
+                    handshake_result.auth_header_packet,
+                    incoming_packet.sender_endpoint,
+                )
+                self.logger.debug("Sending AuthHeader packet to let peer complete handshake")
```
pipermerriam: Maybe include who the remote peer is here.

p2p/discv5/packer.py
jannikluhn: > It seems like there may be a race condition that needs to be handled where is_pre_handshake was True when this handler was called but then the state changes before this coroutine gets executed causing this to now be False?

My understanding is that there's no schedule point between an `await coroutine` and the first line of the coroutine, so there shouldn't be a race condition even if we handle packets concurrently. I'm not entirely certain, but I did the following test which seems to confirm this:

```Py
import trio
import trio.testing

async def main():
    with trio.testing.assert_no_checkpoints():
        await f()

async def f():
    print()

trio.run(main)  # doesn't fail
```

I'll add a comment anyways, because the rest of the code still assumes that packets are handled sequentially.

p2p/discv5/packer.py
jannikluhn: :+1: `self.manager.wait_cancelled()` seems like the right way to go

p2p/discv5/packer.py
jannikluhn: In general, I used to think that private key, public key, and signature formats are totally up to the identity scheme being used, so that's why I used `bytes` for everything in `identity_scheme.py`. We have to use the same types here (and in `handshake.py` for that matter).

But you're right, by now I realized that at least the `local_private_key` needs to be in a common base format that different identity schemes can interpret, so we can type annotate it more strictly. I think it should just be `NewType(bytes)` though (because there can be a non-ECDSA identity scheme for which `eth-keys` isn't applicable).

But this would be a bigger change as it affects not only the peer packer, but also the identity schemes and the handshakers. So I'd rather do this in a follow up PR.

p2p/discv5/packer.py
jannikluhn: Just noticed that the daemon tasks also makes the `sleep_forever`/`wait_cancelled` unnecessary.

pipermerriam: I think that behavior is implicit.  If we're going to depend on it i'd like to be sure we have a test for the trio-based service that tests that behavior.

PR #963, opened 2019-08-22
Add support for etherscan resolved checkpoints

### LINKED ISSUE #892: Allow beam sync to sync from a checkpoint header provided via CLI flag

cburgdorf:
### What was wrong?

This adds support for a checkpoint URI that automatically resolves a trusted checkpoint via the etherscan API.

Relates to #892 

### How was it fixed?

I considered a different design for this initially, one that would have consulted the etherscan API in an async fashion at a later stage of the boot. But here's why I went for this instead:

1. Trinity already depends against the `request` library and that library is entirely synchronous

2. Making synchronous requests at this super early stage (it's pretty much the very first thing that happens) will only delay the boot a bit but doesn't cause trouble anywhere else so that might be acceptable

3. It's very simple. The whole change only happens on the argument parsing stage and is otherwise transparent (uses the existing checkpoint processing pipeline)

Usage: `-beam-from-checkpoint="eth://block/byetherscan/latest"`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/integration/test_etherscan_checkpoint_resolver.py
+++ b/tests/integration/test_etherscan_checkpoint_resolver.py
@@ -1,0 +1,23 @@
+import pytest
+
+from eth_utils import encode_hex
+from trinity.plugins.builtin.syncer.cli import (
+    parse_checkpoint_uri,
+    is_block_hash,
+)
+
+# This is just the score at the tip as it was at some point on August 26th 2019
+# It serves as anchor so that we have *some* minimal expected score to test against.
+MIN_EXPECTED_SCORE = 11631608640717612820968
+
+
+@pytest.mark.parametrize(
+    'uri',
+    (
+        'eth://block/byetherscan/latest',
+    )
+)
+def test_parse_checkpoint(uri):
+    checkpoint = parse_checkpoint_uri(uri)
+    assert checkpoint.score >= MIN_EXPECTED_SCORE
+    assert is_block_hash(encode_hex(checkpoint.block_hash))
--- a/trinity/plugins/builtin/syncer/cli.py
+++ b/trinity/plugins/builtin/syncer/cli.py
@@ -11,13 +11,19 @@ from eth_typing import (
 )
 
 from eth_utils import (
-    is_hex,
-    remove_0x_prefix,
     decode_hex,
+    is_hex,
+    to_int,
+    remove_0x_prefix,
     ValidationError,
 )
 
 from trinity.sync.common.checkpoint import Checkpoint
+
+from .etherscan_api import (
+    get_block_by_number,
+    get_latest_block,
+)
 
 
 def is_block_hash(value: str) -> bool:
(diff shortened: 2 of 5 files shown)


pipermerriam:
FYI, I'm :+1: with the current approach, but when we do want async requests: https://asks.readthedocs.io/en/latest/

trinity/plugins/builtin/syncer/cli.py
```diff
@@ -34,7 +40,31 @@ def parse_checkpoint_uri(uri: str) -> Checkpoint:
     except ValueError as e:
         raise ValidationError(str(e))
 
-    scheme, netloc, path, query = parsed.scheme, parsed.netloc, parsed.path.lower(), parsed.query
+    path = parsed.path.lower()
+    if path.startswith('/byhash'):
+        return parse_byhash_uri(parsed)
+    elif path.startswith('/byetherscan/latest'):
```
carver: I guess this could be equals rather than `startswith` to be more precise.

trinity/plugins/builtin/syncer/cli.py
```diff
@@ -34,7 +40,31 @@ def parse_checkpoint_uri(uri: str) -> Checkpoint:
...
+        raise ValidationError("Not a valid checkpoint URI")
+
+
+BLOCKS_FROM_TIP = 50
+
+
+def parse_byetherscan_uri(parsed: urllib.parse.ParseResult) -> Checkpoint:
+
+    latest_block_number = get_latest_block()
+    checkpoint_block_number = latest_block_number - BLOCKS_FROM_TIP
```
carver: It wasn't obvious to me from reading the URL that it goes 50 blocks back. (Maybe add a new example to the argparse help text?)

Another option is to use a flexible URI that lets you pick how far back you want to go, like:
`eth://block/byetherscan/-50` to go 50 blocks. I kind of like this better since it's more likely to be self-evident from the URL, but it's not a blocker.

cburgdorf: >Maybe add a new example to the argparse help text

I added a new example that makes it clear we are auto selecting a block *near* the tip. I like your suggested syntax for a more specific block selection. I'd leave this open for a future PR though.

tests/integration/test_etherscan_checkpoint_resolver.py
```diff
@@ -0,0 +1,23 @@
...
+MIN_EXPECTED_SCORE = 11631608640717612820968
+
+
+@pytest.mark.parametrize(
+    'uri',
+    (
+        'eth://block/byetherscan/latest',
+    )
+)
+def test_parse_checkpoint(uri):
```
cburgdorf: @pipermerriam @carver does this count? I'm not sure what exactly to test for as we'll have a moving target. I also considered using the etherscan API to call out and fetch the block that is behind `checkpoint.block_hash` and then compare the score of that with `checkpoint.score` but tbh, that is exactly what is happening behind the scenes so I'm not sure we'd actually gain much by doing that.

PR #964, opened 2019-08-22
Fix base protocol selection in MultiplexerPairFactory

pipermerriam:
### What was wrong?

The `MultiplexerPairFactory` did not properly negotiate the proper version of the base protocol.

### How was it fixed?

Added code to negotiate the base protocol version based on the `p2p_version` supplied for each peer.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

Newfoundland_dog_Smoky

### DIFF
--- a/tests/p2p/test_multiplexer_pair_factory.py
+++ b/tests/p2p/test_multiplexer_pair_factory.py
@@ -19,36 +19,29 @@ def test_multiplexer_pair_factory():
 
 
 @pytest.mark.parametrize(
-    'alice_p2p_version,bob_p2p_version',
+    'alice_p2p_version,bob_p2p_version,expected_base_protocol_class',
     (
-        (DEVP2P_V4, DEVP2P_V4),
-        (DEVP2P_V4, DEVP2P_V5),
-        (DEVP2P_V5, DEVP2P_V4),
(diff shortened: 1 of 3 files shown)


pipermerriam:
extracted from https://github.com/ethereum/trinity/pull/956

tests/p2p/test_multiplexer_pair_factory.py
```diff
@@ -0,0 +1,54 @@
...
+@pytest.mark.asyncio
+async def test_multiplexer_pair_factory_with_different_p2p_versions(
+    alice_p2p_version,
+    bob_p2p_version,
+):
+    alice_multiplexer, bob_multiplexer = MultiplexerPairFactory(
+        alice_p2p_version=alice_p2p_version,
+        bob_p2p_version=bob_p2p_version,
+    )
+    expected_base_protocol_version = min(alice_p2p_version, bob_p2p_version)
```
cburgdorf: Instead of the following block that picks the `expected_base_protocol_class`, it might be clearer to just add another parameter to the test `expected_base_protocol_class` with it's value being set to either `P2PProtocolV4` or `P2PProtocol` for each case.

PR #969, opened 2019-08-23
Discv5 Packer

jannikluhn:
The packer's job is it to mange peer packers and dispatch messages/packets to/from them. Whenever it receives a packet from the outside or a message to send from the inside, it passes it on to the responsible peer packer. If it doesn't exist yet, it launches a new one and manages its lifecycle.

Most of the time, the dispatching is based on the remote node id that can be extracted from the packet's tag. However, during handshake this doesn't necessarily work and we have to explicitly ask all peer packers if they are waiting for the packet. I thought about doing this in a smarter way by having a way of peer packers registering handlers at the main packer, but this turned out to be rather complicated and the performance impact of the brute force solution is probably negligible.

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/packer.py
+++ b/p2p/discv5/packer.py
@@ -1,7 +1,10 @@
 import logging
 from typing import (
+    Dict,
     List,
+    NamedTuple,
     Optional,
+    Tuple,
 )
 
 from eth_utils import (
@@ -15,7 +18,11 @@ from trio.abc import (
     SendChannel,
 )
 
-from p2p.trio_service import Service
+from p2p.trio_service import (
+    LifecycleError,
+    Service,
+    Manager,
+)
 
 from p2p.discv5.abc import (
     EnrDbApi,
@@ -45,6 +52,7 @@ from p2p.discv5.packets import (
 )
 from p2p.discv5.tags import (
     compute_tag,
+    recover_source_id_from_tag,
 )
 from p2p.discv5.typing import (
     NodeID,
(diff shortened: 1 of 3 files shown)


p2p/discv5/packer.py
```diff
@@ -432,3 +442,185 @@ def is_expecting_handshake_packet(self, incoming_packet: IncomingPacket) -> bool
...
+                f"Peer packer for {encode_hex(remote_node_id)} is still running"
+            )
+
+        self.managed_peer_packers.pop(remote_node_id)
+
+    async def run_peer_packer(self, remote_node_id: NodeID) -> None:
+        if not self.is_peer_packer_registered(remote_node_id):
+            raise ValueError("Peer packer for {encode_hex(remote_node_id)} is not registered")
+        managed_peer_packer = self.managed_peer_packers[remote_node_id]
+        if managed_peer_packer.manager.is_started:
```
pipermerriam: Instead of doing this pre-check you could just rely on catching the `LifecycleError` when you try to start it.

tests-trio/p2p-trio/test_packer.py
```diff
@@ -333,3 +382,95 @@ def remote_outgoing_message_channels():
...
+                                    outgoing_packet_channels):
+    assert not packer.is_peer_packer_registered(remote_enr.node_id)
+
+    # send message
+    outgoing_message = OutgoingMessage(
+        message=PingMessageFactory(),
+        receiver_endpoint=remote_endpoint,
+        receiver_node_id=remote_enr.node_id,
+    )
+    nursery.start_soon(outgoing_message_channels[0].send, outgoing_message)
```
pipermerriam: FYI. in case you weren't aware, there is also a `send_nowait`

p2p/discv5/packer.py
```diff
@@ -432,3 +442,185 @@ def is_expecting_handshake_packet(self, incoming_packet: IncomingPacket) -> bool
...
+        self.outgoing_packet_send_channel = outgoing_packet_send_channel
+
+        self.logger = logging.getLogger(f"p2p.discv5.packer.Packer")
+
+        self.managed_peer_packers: Dict[NodeID, ManagedPeerPacker] = {}
+
+    async def run(self) -> None:
+        self.manager.run_task(self.handle_incoming_packets, daemon=True)
+        self.manager.run_task(self.handle_outgoing_messages, daemon=True)
+        await self.manager.wait_stopped()
```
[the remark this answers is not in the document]
jannikluhn: Agree, it seems wrong to wait for something if the thing can only occur once we're done waiting.

tests-trio/p2p-trio/test_packer.py
jannikluhn: I was aware of it, but didn't think it would solve my problem here. But looks like it does, thanks!

jannikluhn:
> One thing I'm not finding but maybe I missed it. There doesn't seem to be a mechanism that cleans-up old peer packers that aren't in use anymore. It seems like this will continually grow its memory footprint the longer it runs and the more peers it handshakes with.

That's correct, I don't clean up old connections yet (unless a handshake with them failed). The solution I had in mind is to just raise some kind of timeout exception in the peer packer if it doesn't receive a packet in a while. We could also catch this exception in the main packer and remember the session keys for future connections. If the overhead of having many peer packers turns out to be too large, we could even go one step further and use the peer packer only for handshake and store all session keys in the peer packer (which I hope won't be necessary because I like the fact that we have one service responsible for every peer).

But all of that is something I'd like to tackle once there is a working version that can talk to other clients.

PR #970, opened 2019-08-24
Message Dispatcher

jannikluhn:
The message dispatcher is the interface that the business logic will use to communicate with peers. They can register handlers at the dispatcher, which will then pass the corresponding messages to the handler via channels. For convenience, there's also a `request` method that one can await to get the response, which will automatically register and deregister handlers and look up ip address and udp port from the ENR db.

I also unified how we handle IP addresses: In ENRs they are `bytes`, in the `Endpoint` tuple they used to be .-delimited `str`s (now everything is `bytes`).

Some request-responses will consist of multiple messages (in particular, `Nodes` messages). Those cannot be nicely handled with `requests` at the moment. I'll probably add a `request_nodes` method for that when I need it.

It might be necessary in the future to move some of the functionality down to the packer in order to handle timeouts and failed handshakes properly. But the interface of the message dispatcher hopefully shouldn't be affected by this, so I just wrote the simple version for now.

#### Cute Animal Picture

animal-animal-world-bird-portrait-54103

### DIFF
--- a/p2p/discv5/message_dispatcher.py
+++ b/p2p/discv5/message_dispatcher.py
@@ -1,0 +1,319 @@
+import logging
+import random
+from types import (
+    TracebackType,
+)
+from typing import (
+    AsyncIterator,
+    Callable,
+    Dict,
+    Optional,
+    Tuple,
+    Type,
+    TypeVar,
+)
+
+import trio
+from trio.abc import (
+    ReceiveChannel,
+    SendChannel,
+)
+from trio.hazmat import (
+    checkpoint,
+)
+
+from eth_utils import (
+    encode_hex,
+)
+
+from p2p.trio_service import Service
+
+from p2p.discv5.channel_services import (
+    Endpoint,
+    IncomingMessage,
+    OutgoingMessage,
+)
+from p2p.discv5.abc import (
+    ChannelHandlerSubscriptionAPI,
+    EnrDbApi,
+    MessageDispatcherAPI,
+)
+from p2p.discv5.constants import (
+    IP_V4_ADDRESS_ENR_KEY,
+    MAX_REQUEST_ID,
+    MAX_REQUEST_ID_ATTEMPTS,
+    UDP_PORT_ENR_KEY,
+)
+from p2p.discv5.messages import (
+    BaseMessage,
+)
+from p2p.discv5.typing import (
+    NodeID,
+)
+
+
+def get_random_request_id() -> int:
+    return random.randint(0, MAX_REQUEST_ID)
+
+
+ChannelContentType = TypeVar("ChannelContentType")
+
+
+class ChannelHandlerSubscription(ChannelHandlerSubscriptionAPI[ChannelContentType]):
+    def __init__(self,
+                 send_channel: SendChannel[ChannelContentType],
+                 receive_channel: ReceiveChannel[ChannelContentType],
+                 remove_fn: Callable[[], None],
+                 ) -> None:
+        self._send_channel = send_channel
+        self.receive_channel = receive_channel
+        self.remove_fn = remove_fn
+
+    def cancel(self) -> None:
+        self.remove_fn()
+
+    async def __aenter__(self) -> "ChannelHandlerSubscription[ChannelContentType]":
+        await self._send_channel.__aenter__()
+        await self.receive_channel.__aenter__()
+        return self
+
+    async def __aexit__(self,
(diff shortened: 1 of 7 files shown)


p2p/discv5/message_dispatcher.py
```diff
@@ -0,0 +1,252 @@
...
+        try:
+            async with response_receive_channel:
+                outgoing_message = await self.prepare_outgoing_message(receiver_node_id, message)
+                self.logger.debug(
+                    "Sending %s to %s with request id %d",
+                    outgoing_message,
+                    encode_hex(receiver_node_id),
+                    message.request_id,
+                )
+                await self.outgoing_message_send_channel.send(outgoing_message)
```
pipermerriam: Using channels for the response seems heavy.  Thoughts on using the [`Future`](https://github.com/ethereum/lahja/blob/master/lahja/trio/future.py) concept from the `lahja` codebase for this?

tests-trio/p2p-trio/test_message_dispatcher.py
```diff
@@ -0,0 +1,186 @@
...
+    await message_dispatcher.register_request_handler(PingMessage.message_type, ping_send_channel)
+
+    incoming_message = IncomingMessage(
+        message=PingMessageFactory(),
+        sender_endpoint=remote_endpoint,
+        sender_node_id=remote_enr.node_id,
+    )
+    await incoming_message_channels[0].send(incoming_message)
+
+    with trio.fail_after(0.1):
```
pipermerriam: You *might* bump this up to `1` since it'll produce fewer false negatives but shouldn't slow down the happy path at all.

p2p/discv5/message_dispatcher.py
jannikluhn: The problem is that a response can consist of multiple messages (in particular, a list of nodes can be split up to keep individual messages short). So I think we need channels at least at the lowest level.

I also don't like this pattern too much though, but that's exactly the reason why I added the `request` method. So that one can just use `pong = await dispatcher.request(ping)` and ignore channels completely. Maybe we can add some similar convenience method in the `HandlerSubscription` context manager you proposed:

```Py
with dispatcher.add_request_handler(PingMessage.message_type) as request_handler:
    # either
    await request_handler.receive_single_response()
    # or
    async for response in request_handler.response_receive_channel:
        ....
```

Regarding futures: I've tried to avoid them because the [trio docs](https://trio.readthedocs.io/en/stable/design.html#user-level-api-principles) say:

> Task spawning is always explicit. No callbacks, no implicit concurrency, no futures/deferreds/promises/other APIs that involve callbacks.

I personally don't feel very strongly about this, but I suspected that they know better than me and as long as a normal async function is good enough I thought avoiding the issue is best.

jannikluhn:
As suggested I've implemented the subscription interface for request/response handlers which I think is a nice improvement. In addition to just adding/removing the handler, I've also pulled in the send/receive channels. This has some advantages:

- We can automatically enter and exit the channels when entering exiting the context manager. This saves one indentation level for basically all use case.
- Adding the handler becomes a synchronous function (because we don't have to `__aenter__` the send channel which means we can write

```py
async with message_dispatcher.add_request_handler(PingMessage.message_type) as channel_handler_subscription:
    ...
```

instead having to write 

```Py
channel_handler_subscription = await message_dispatcher.add_request_handler(PingMessage.message_type)
async with channel_handler_subscription:
    ...
```

- We don't have to create the send/receive channels manually anymore
- We can mirror the receive channel functionality (receive method and iteration) on the subscription object, so we don't have to touch the channels itself:

```Py
async with message_dispatcher.add_request_handler(PingMessage.message_type) as channel_handler_subscription:
    incoming_message = await channel_handler_subscription.receive()
    # or
    async for incoming_message in channel_handler_subscription:
        ...
```

But there are disadvantages:

- Coming from the `p2p.connection` handler subscription, combining handling and subscription management feels wrong
- It somewhat breaks with the convention I've followed so far of always creating channel pairs externally and passing the relevant ones where needed (only "somewhat" because while it is the messgae dispatcher which opens the channels, it's only temporary -- eventually the subscription object eventually becomes the owner of the channels)
- It's not really possible to use it without a `with` block (one would have to call `__aenter__` and `__aexit__` manually)

All in all I think I'm happy with the `ChannelSubscriptionHandler` (maybe the name isn't ideal though). It's conceptually not as clean, but easier to use.

PR #971, opened 2019-08-25
Finish `Hello`, `BeaconBlocks`, `RecentBeaconBlocks` req/resp

NIC619:
### What was wrong?
Finish `Hello`, `BeaconBlocks`, `RecentBeaconBlocks` request/response
- spec for `Hello`: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/networking/p2p-interface.md#hello
- spec for `BeaconBlocks`: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/networking/p2p-interface.md#beaconblocks
- spec for `RecentBeaconBlocks`: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/networking/p2p-interface.md#recentbeaconblocks

### How was it fixed?
Add handler for these requests and functions to make these requests

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trinity/protocol/bcc_libp2p/node.py
+++ b/trinity/protocol/bcc_libp2p/node.py
@@ -333,8 +333,8 @@ class Node(BaseService):
             hello_other_side.finalized_epoch,
             config.SLOTS_PER_EPOCH,
         )
-        block = self.chain.get_canonical_block_by_slot(finalized_epoch_start_slot)
-        finalized_root = block.signing_root
+        finalized_root = self.chain.get_canonical_block_by_slot(
+            finalized_epoch_start_slot).block.signing_root
 
         if hello_other_side.finalized_root != finalized_root:
             raise ValidationError(
@@ -645,7 +645,26 @@ class Node(BaseService):
             return
         self.logger.debug("Received the beacon blocks request message %s", beacon_blocks_request)
 
-        # TODO: Validate `start_slot`(>0)?
+        # Validate `beacon_blocks_request.start_slot`
+        state_machine = self.chain.get_state_machine()
+        finalized_epoch_start_slot = compute_start_slot_of_epoch(
+            epoch=state_machine.state.finalized_checkpoint.epoch,
+            slots_per_epoch=state_machine.config.SLOTS_PER_EPOCH,
+        )
+        if beacon_blocks_request.start_slot < max(0, finalized_epoch_start_slot):
+            reason = (
+                f"Invalid request: `start_slot`({beacon_blocks_request.start_slot})"
+                f" lower than our latest finalized slot({finalized_epoch_start_slot})"
+            )
+            try:
+                await write_resp(stream, reason, ResponseCode.INVALID_REQUEST)
+            except WriteMessageFailure as error:
+                self.logger.info(
+                    "Processing beacon blocks request failed: failed to write message %s",
+                    reason,
+                )
+            # await stream.reset()
+            return
 
         try:
             peer_head_block = self.chain.get_block_by_root(beacon_blocks_request.head_block_root)
--- a/trinity/protocol/bcc_libp2p/node.py
+++ b/trinity/protocol/bcc_libp2p/node.py
@@ -11,6 +11,11 @@ from cancel_token import (
     CancelToken,
 )
 
+from eth.constants import ZERO_HASH32
+
+from eth2.beacon.helpers import (
+    get_block_root,
+)
 from eth2.beacon.chains.base import (
     BaseBeaconChain,
 )
@@ -278,15 +283,39 @@ class Node(BaseService):
     #   - Disconnect peers when they fail to join in a certain amount of time.
 
     async def _validate_hello_req(self, hello_other_side: HelloRequest) -> None:
-        state = self.chain.get_state_machine().state
+        state_machine = self.chain.get_state_machine()
+        state = state_machine.state
+        config = state_machine.config
         if hello_other_side.fork_version != state.fork.current_version:
             raise ValidationError(
                 "`fork_version` mismatches: "
                 f"hello_other_side.fork_version={hello_other_side.fork_version}, "
                 f"state.fork.current_version={state.fork.current_version}"
             )
-        # TODO: Reject if the (finalized_root, finalized_epoch) shared by the peer
-        #   is not in the client's chain at the expected epoch.
+
+        # Can not validate the checkpoint with `finalized_epoch` higher than ours
+        if hello_other_side.finalized_epoch > state.finalized_checkpoint.epoch:
+            return
+
+        # Get the finalized root at `hello_other_side.finalized_epoch`
+        # FIXME: This check is added here because `get_block_root` would fail when
+        # `state.slot == 0`.
+        if hello_other_side.finalized_epoch == 0 and state.slot == 0:
+            finalized_root = ZERO_HASH32
+        else:
+            finalized_root = get_block_root(
+                state,
+                hello_other_side.finalized_epoch,
+                config.SLOTS_PER_EPOCH,
+                config.SLOTS_PER_HISTORICAL_ROOT,
+            )
+        if hello_other_side.finalized_root != finalized_root:
+            raise ValidationError(
+                "`finalized_root` mismatches: "
+                f"hello_other_side.finalized_root={hello_other_side.finalized_root}, "
+                f"hello_other_side.finalized_epoch={hello_other_side.finalized_epoch}, "
+                f"our `finalized_root` at the same `finalized_epoch`={finalized_root}"
+            )
 
     async def _request_beacon_blocks(self) -> None:
         """
@@ -328,17 +357,21 @@ class Node(BaseService):
         except ReadMessageFailure as error:
             # FIXME: Use `Stream.reset()` when `NetStream` has this API.
             # await stream.reset()
-            # TODO: Disconnect
+            # TODO: send `Goodbye` req then disconnect
             return
         self.logger.debug("Received the hello message %s", hello_other_side)
 
         try:
             await self._validate_hello_req(hello_other_side)
-        except ValidationError:
-            self.logger.info("Handshake failed: hello message %s is invalid", hello_other_side)
+        except ValidationError as error:
+            self.logger.info(
+                "Handshake failed: hello message %s is invalid: %s",
+                hello_other_side,
+                str(error)
+            )
             # FIXME: Use `Stream.reset()` when `NetStream` has this API.
             # await stream.reset()
-            # TODO: Disconnect
+            # TODO: send `Goodbye` req then disconnect
             return
 
         hello_mine = self._make_hello_packet()
@@ -361,7 +394,19 @@ class Node(BaseService):
             "Handshake from %s is finished. Added to the `handshake_peers`",
             peer_id,
         )
-        # TODO: If we have lower `finalized_epoch` or `head_slot`, request the later beacon blocks.
+
+        # Check if we are behind the peer
+        checkpoint = self.chain.get_state_machine().state.finalized_checkpoint
+        head_block = self.chain.get_canonical_head()
+        peer_has_higher_finalized_epoch = hello_other_side.finalized_epoch > checkpoint.epoch
+        peer_has_equal_finalized_epoch = hello_other_side.finalized_epoch == checkpoint.epoch
+        peer_has_higher_head_slot = hello_other_side.head_slot > head_block.slot
+        if (
+            peer_has_higher_finalized_epoch or
+            (peer_has_equal_finalized_epoch and peer_has_higher_head_slot)
+        ):
+            # TODO: kickoff syncing process with this peer
+            pass
 
         await stream.close()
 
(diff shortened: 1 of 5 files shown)


NIC619:
Would appreciate @hwwhww and @ralexstokes 's feedbacks on two functions: `_validate_hello_req()` and `_handle_beacon_blocks()` since they involve verifying peer's checkpoints data and retrieving history blocks for peer.

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -328,17 +379,21 @@ def _make_hello_packet(self) -> HelloRequest:
         except ReadMessageFailure as error:
             # FIXME: Use `Stream.reset()` when `NetStream` has this API.
             # await stream.reset()
-            # TODO: Disconnect
+            # TODO: send `Goodbye` req then disconnect
```
mhchia: Should we just `reset` and `disconnect` here, instead of sending `Goodbye`? IMO the [spec](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/networking/p2p-interface.md#requesting-side) only mentions `reset`.

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -328,17 +379,21 @@ def _make_hello_packet(self) -> HelloRequest:
...
+        except ValidationError as error:
+            self.logger.info(
+                "Handshake failed: hello message %s is invalid: %s",
+                hello_other_side,
+                str(error)
+            )
             # FIXME: Use `Stream.reset()` when `NetStream` has this API.
             # await stream.reset()
-            # TODO: Disconnect
+            # TODO: send `Goodbye` req then disconnect
```
mhchia: ditto

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -361,7 +416,19 @@ def _make_hello_packet(self) -> HelloRequest:
...
+        checkpoint = self.chain.get_state_machine().state.finalized_checkpoint
+        head_block = self.chain.get_canonical_head()
+        peer_has_higher_finalized_epoch = hello_other_side.finalized_epoch > checkpoint.epoch
+        peer_has_equal_finalized_epoch = hello_other_side.finalized_epoch == checkpoint.epoch
+        peer_has_higher_head_slot = hello_other_side.head_slot > head_block.slot
+        if (
+            peer_has_higher_finalized_epoch or
+            (peer_has_equal_finalized_epoch and peer_has_higher_head_slot)
+        ):
+            # TODO: kickoff syncing process with this peer
```
mhchia: Could you add a log here, indicating this event occurs?

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -442,6 +510,339 @@ def _make_hello_packet(self) -> HelloRequest:
...
+            pass
```
mhchia: Is it the same code snippet as the one in `_handle_hello`? Could we move it to another method?

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -442,6 +510,339 @@ def _make_hello_packet(self) -> HelloRequest:
...
+                "resp_code != ResponseCode.SUCCESS, "
+                f"resp_code={resp_code}, error_msg={beacon_blocks_response}"
+            )
+            self.logger.info("Request beacon blocks failed: %s", error_msg)
+            # FIXME: Use `Stream.reset()` when `NetStream` has this API.
+            # await stream.reset()
+            raise RequestFailure(error_msg)
+
+        beacon_blocks_response = cast(BeaconBlocksResponse, beacon_blocks_response)
+        asyncio.ensure_future(stream.close())
```
mhchia: Is there a specific reason that `asyncio.ensure_future` is used?

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -442,6 +510,339 @@ def _make_hello_packet(self) -> HelloRequest:
...
+        self.logger.debug("Sending recent beacon blocks response %s", recent_beacon_blocks_response)
+        try:
+            await write_resp(stream, recent_beacon_blocks_response, ResponseCode.SUCCESS)
+        except WriteMessageFailure as error:
+            self.logger.info(
+                "Processing recent beacon blocks request failed: failed to write message %s",
+                recent_beacon_blocks_response,
+            )
+            # await stream.reset()
+            # TODO: Disconnect
```
mhchia: Probably don't need to disconnect here?

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -442,6 +510,339 @@ def _make_hello_packet(self) -> HelloRequest:
...
+            self.logger.info("Request recent beacon blocks failed: %s", error_msg)
+            # FIXME: Use `Stream.reset()` when `NetStream` has this API.
+            # await stream.reset()
+            raise RequestFailure(error_msg)
+
+        recent_beacon_blocks_response = cast(
+            RecentBeaconBlocksResponse,
+            recent_beacon_blocks_response,
+        )
+        asyncio.ensure_future(stream.close())
```
mhchia: ditto, could you elaborate the reason why `asyncio.ensure_future` is used?

trinity/protocol/bcc_libp2p/node.py
NIC619: It's mentioned in `Goodbye` req (https://github.com/ethereum/eth2.0-specs/blob/dev/specs/networking/p2p-interface.md#goodbye):
```
Client MAY send goodbye messages upon disconnection. The reason field MAY be one of the following values:

…
2: Irrelevant network.
...
```

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -278,15 +307,42 @@ def _register_rpc_handlers(self) -> None:
...
+            hello_other_side.finalized_root == ZERO_HASH32
+        ):
+            return
+
+        finalized_epoch_start_slot = compute_start_slot_of_epoch(
+            hello_other_side.finalized_epoch,
+            config.SLOTS_PER_EPOCH,
+        )
+        block = self.chain.get_canonical_block_by_slot(finalized_epoch_start_slot)
+        finalized_root = block.signing_root
```
hwwhww: Could combine into:
```python
        finalized_root = self.chain.get_canonical_block_root(finalized_epoch_start_slot)
```

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -442,6 +523,351 @@ def _make_hello_packet(self) -> HelloRequest:
...
+            hello_other_side.finalized_epoch,
+            hello_other_side.head_slot,
+        )
 
         await stream.close()
+
+    @to_tuple
+    def _get_blocks_from_canonical_chain_by_slot(
+        self,
+        slot_of_requested_blocks: Sequence[BaseBeaconBlock],
```
hwwhww:
```suggestion
        slot_of_requested_blocks: Sequence[Slot],
```

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -442,6 +523,351 @@ def _make_hello_packet(self) -> HelloRequest:
...
+                pass
+            else:
+                yield block
+
+    @to_tuple
+    def _get_blocks_from_fork_chain_by_root(
+        self,
+        start_slot: Slot,
+        peer_head_block: BaseBeaconBlock,
+        slot_of_requested_blocks: Sequence[BaseBeaconBlock],
```
hwwhww:
```suggestion
        slot_of_requested_blocks: Sequence[Slot],
```

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -442,6 +523,351 @@ def _make_hello_packet(self) -> HelloRequest:
...
+
+        # We have the peer's head block in our database,
+        # next check if the head block is on our canonical chain.
+        try:
+            canonical_block_at_slot = self.chain.get_canonical_block_by_slot(
+                peer_head_block.slot
+            )
+        except BlockNotFound:
+            # Peer's head block is not on our canonical chain
+            return self._get_blocks_from_fork_chain_by_root(
```
hwwhww: The spec says:
> Clients MUST support head_block_root values since the latest finalized epoch.

Maybe check if `beacon_blocks_request.start_slot` >= our latest finalized slot before calling it.

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -442,6 +523,351 @@ def _make_hello_packet(self) -> HelloRequest:
...
+                beacon_blocks_request.start_slot,
+                peer_head_block,
+                slot_of_requested_blocks,
+            )
+        else:
+            if canonical_block_at_slot != peer_head_block:
+                # Peer's head block is not on our canonical chain
+                return self._get_blocks_from_fork_chain_by_root(
```
hwwhww: ditto

trinity/protocol/bcc_libp2p/node.py
NIC619: Added a check on `beacon_blocks_request.start_slot` after `beacon_blocks_request` is received.

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -278,15 +307,42 @@ def _register_rpc_handlers(self) -> None:
...
+            hello_other_side.finalized_root == ZERO_HASH32
+        ):
+            return
+
+        finalized_epoch_start_slot = compute_start_slot_of_epoch(
+            hello_other_side.finalized_epoch,
+            config.SLOTS_PER_EPOCH,
+        )
+        finalized_root = self.chain.get_canonical_block_by_slot(
+            finalized_epoch_start_slot).block.signing_root
```
hwwhww: `get_canonical_block_root` should be enough; no need to get the block.

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -645,7 +645,26 @@ def _get_requested_beacon_blocks(
...
         self.logger.debug("Received the beacon blocks request message %s", beacon_blocks_request)
 
-        # TODO: Validate `start_slot`(>0)?
+        # Validate `beacon_blocks_request.start_slot`
+        state_machine = self.chain.get_state_machine()
+        finalized_epoch_start_slot = compute_start_slot_of_epoch(
+            epoch=state_machine.state.finalized_checkpoint.epoch,
+            slots_per_epoch=state_machine.config.SLOTS_PER_EPOCH,
+        )
+        if beacon_blocks_request.start_slot < max(0, finalized_epoch_start_slot):
```
hwwhww: I think it's off here; it's only invalid in the "Peer's head block is not on our canonical chain" case; otherwise, it's just a normal sync historical blocks sync.

trinity/protocol/bcc_libp2p/node.py
hwwhww: see https://github.com/ethereum/trinity/pull/971/commits/0690bc6ca85f1b930d88e9bca26c748cc5e4c519#r318717133

trinity/protocol/bcc_libp2p/node.py
NIC619: Yeah you are right. Checking `start_slot < finalized_epoch` here is definitely a mistake.

> it's only invalid in the "Peer's head block is not on our canonical chain" case; otherwise

Then if that's the case I suppose it shouldn't be considered invalid? Just don't return blocks older than our finalized epoch to requesting peer.
Otherwise requesting peer would need to make another request beforehand to know requested peer's finalized epoch in case he requests with `start_slot` older than the finalized epoch which results in an invalid request(getting no blocks at all)?

NIC619: On a second thought,
I interpret
> Clients MUST support `head_block_root` values since the latest finalized epoch.

as saying `head_block_root` must be block root whose `block.slot >= latest finalized epoch` and
> Clients MUST support requesting blocks since the start of the weak subjectivity period and up to the given head_block_root.

as saying we must provide blocks since weak subjectivity period.
Hence IMO as long as `head_block_root` is valid we should provide the blocks even if they are not on canonical chain and `start_slot > latest finalized epoch`.

hwwhww: Right. I interpret it as:
```python
if head_block_root is on our canonical chain:
    MUST support
elif head_block_root is not on our canonical chain and start_slot > latest finalized slot:
    MUST support
else:  # head_block_root is not on our canonical chain and start_slot <= latest finalized slot
   Don't have to support
```

NIC619: Added the check on `start_slot` when being requested blocks not on our canonical chain.

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -436,12 +506,383 @@ def _make_hello_packet(self) -> HelloRequest:
...
+                # Validate `start_slot` is greater than our latest finalized slot
+                self._validate_start_slot(beacon_blocks_request.start_slot)
+                return self._get_blocks_from_fork_chain_by_root(
+                    beacon_blocks_request.start_slot,
+                    peer_head_block,
+                    slot_of_requested_blocks,
+                )
+            else:
+                # Peer's head block is on our canonical chain
+                return self._get_blocks_from_canonical_chain_by_slot(slot_of_requested_blocks)
```
hwwhww: Refactoring idea:
```python
        try:
            canonical_block_at_slot = self.chain.get_canonical_block_by_slot(
                peer_head_block.slot
            )
            block_match = canonical_block_at_slot == peer_head_block
        except BlockNotFound:
            block_match = False
        finally:
            if block_match:
                # Peer's head block is on our canonical chain
                return self._get_blocks_from_canonical_chain_by_slot(slot_of_requested_blocks)
            else:
                # Peer's head block is not on our canonical chain
                # Validate `start_slot` is greater than our latest finalized slot
                self._validate_start_slot(beacon_blocks_request.start_slot)
                return self._get_blocks_from_fork_chain_by_root(
                    beacon_blocks_request.start_slot,
                    peer_head_block,
                    slot_of_requested_blocks,
                )
```

PR #990, opened 2019-08-28
Connection handlers take connection arg

pipermerriam:
### What was wrong?

For the handler function API on the `Connection` object to be useful, the handlers need to have access to the connection.

### How was it fixed?

Now handler functions expect the connection as the first argument.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

Halloween-Pet-Costumes22__880


carver:
ICYMI:
> collect_messages() takes 2 positional arguments but 3 were given

PR #994, opened 2019-08-28
Bugfix for P2PProtocol.send_disconnect

pipermerriam:
### What was wrong?

`P2PProtocol.send_disconnect` treated the reason argument as the raw value when really we want to use the enum value

### How was it fixed?

Adjusted it to properly accept an enum value.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

costume 1

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -370,7 +370,7 @@ class BasePeer(BaseService):
             )
 
         self.logger.debug("Disconnecting from remote peer %s; reason: %s", self.remote, reason.name)
-        self.base_protocol.send_disconnect(reason.value)
+        self.base_protocol.send_disconnect(reason)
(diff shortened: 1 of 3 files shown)


p2p/peer.py
```diff
@@ -384,7 +384,7 @@ def _disconnect(self, reason: DisconnectReason) -> None:
             )
 
         self.logger.debug("Disconnecting from remote peer %s; reason: %s", self.remote, reason.name)
-        self.base_protocol.send_disconnect(reason.value)
```
cburgdorf: These things show how much we are still in `Any`-land. Ideally, `mypy` should have yelled at us all the time because we were sending a `int` where a `_DisconnectReason` was expected.

Issue #786, opened 2019-07-10
Can we improve dynamic bounds in ssz types?


ralexstokes:
### What is wrong?

We currently set the size of some `ssz` types at runtime as we do not know their length until we have generated a configuration for the beacon chain.

The current way to handle this dynanism is in `eth2/beacon/tools/misc/ssz_vector.py` and requires the user to set it manually in each place we use `ssz` types.

### How can it be fixed

There are a few things to address:

1. Move the configuration inside the configuration type (this would avoid a bug that broke master last week) so that users of the `Eth2Config` do not have to remember to call the helper function each time they make a new config.

2. If we can't find another way to provide the configuration data at runtime then we should carve out a safer interface to do so in`py-ssz`. Right now we are working against `py-ssz` internals which in general we should avoid (so they can change...)

ralexstokes:
Here is an example of how we currently do it, re: the `py-ssz` internals:

https://github.com/ethereum/trinity/pull/785/commits/581ababede180096021b71d525217cb8b514a3e6

jannikluhn:
As pointed out by @hwwhww the switch to hashables might be a good opportunity to tackle this.

As far as I can see, currently we configure the (max) lengths of lists, bitlists, and vectors. Is this all for the foreseeable future, or can we imagine that at some point, for instance, we want to remove certain fields for containers, or change the size of uints?

> Move the configuration inside the configuration type (this would avoid a bug that broke master last week) so that users of the Eth2Config do not have to remember to call the helper function each time they make a new config.

Can you give an example how this would look like? Do you mean something like this:

```
attestation = config.create(Attestation, *args, **kwargs)
```

> If we can't find another way to provide the configuration data at runtime then we should carve out a safer interface to do so inpy-ssz. Right now we are working against py-ssz internals which in general we should avoid (so they can change...)

One option would be to not have the configurable parameters in the sedes object at all and keep them dynamic by default. Then in the low level interfaces, we'd have to pass the config explicitly:

```Py
ssz.encode([1, 2, 3], List(UInt), {"max_length": 4, "element": {"size": 5}})
ssz.encode({"field1": 1, "field2": b"test"}, Container, {"field2": {"length": 4}})

# or going all the way and putting everything in the config
ssz.encode([1, 2, 3], {
    "sedes": List,
    "config": {
        "max_length": 5,
        "element": {
            "sedes": UInt,
            "config": {
                "size": 8
            }
        }
    }
})
[...]

PR #998, opened 2019-08-28
Update fixtures to v0.8.3

ralexstokes:
### What was wrong?

Updates the spec test infrastructure to reflect the changes made in spec v0.8.2 and v0.8.3.

Includes #973.

### How was it fixed?

Updating how we handle test loading

#### Cute Animal Picture


eth2/beacon/attestation_helpers.py
```diff
@@ -134,13 +134,8 @@ def is_slashable_attestation_data(
...
-        or
-        # Surround vote
-        (
-            data_1.source.epoch < data_2.source.epoch
-            and data_2.target.epoch < data_1.target.epoch
-        )
+    # NOTE: checking 'double vote' OR 'surround vote'
+    return (data_1 != data_2 and data_1.target.epoch == data_2.target.epoch) or (
+        data_1.source.epoch < data_2.source.epoch
+        and data_2.target.epoch < data_1.target.epoch
```
ralexstokes: this avoids a `flake8` issue

eth2/beacon/tools/builder/state.py
```diff
@@ -25,7 +25,7 @@ def _check_no_missing_balances(
 def _check_sufficient_balance(balances: Sequence[Gwei], threshold: Gwei) -> None:
     for balance in balances:
         if balance < threshold:
-            assert False
+            raise AssertionError()
```
ralexstokes: new linter rule in updated version

ralexstokes: apparently `assert False` can be compiled away in certain situations

hwwhww:
Post-#851

Each case is set with `"config_types": (Minimal,),`. e.g., https://github.com/ethereum/trinity/blob/9edd694d6fbb5b147cce85adc0e1f1f562429744/tests/eth2/fixtures/test_epoch_processing.py#L17

I think we'd unlikely to "test `registry_updates` with `Minimal` and test `slashings` with `Mainnet`".
How about creating a parameter for it? i.e., `"config_types": config_types`. Or even [adding pytest options](https://docs.pytest.org/en/latest/example/simple.html#pass-different-values-to-a-test-function-depending-on-command-line-options) so that we can set it by command line.

setup.py
```diff
@@ -115,6 +115,10 @@
     'eth2-lint': [
         "black==19.3b0",
         "isort==4.3.21",
+        "flake8==3.7.8",
+        "flake8-bugbear==19.8.0",
+        "mypy==0.720",
+        "sqlalchemy-stubs==0.2",
```
hwwhww: Where do we use `sqlalchemy` in eth2?

hwwhww: Note that these upgrades made `pip install -e .[dev]` install the `eth2-lint` versions instead of `lint`. (`eth2-lint` overwrites `lint`). It's okay for CI, but I wonder if it causes any problem for eth1 development on the local environment.

One simple solution is adding an `eth2-dev` setting. So it will be:
```python
deps['dev'] = (
    deps['dev'] +
    deps['p2p'] +
    deps['trinity'] +
    deps['test'] +
    deps['doc'] +
    deps['lint'] +
    deps['eth2']
)


deps['eth2-dev'] = (
    deps['dev'] +
    deps['eth2-lint']
)
```

eth2/beacon/tools/fixtures/parser.py
```diff
@@ -14,41 +20,106 @@
...
     test_handler: TestHandler[Input, Output],
-    config_type: Optional[ConfigType],
-    config: Optional[Eth2Config],
-) -> TestSuite:
-    test_suite_path = _build_test_suite_path(
-        tests_root_path, test_type, test_handler, config_type
+    config_type: ConfigType,
+    fork_type: ForkType,
+    format_type: FormatType,
+) -> Iterator[TestSuite]:
```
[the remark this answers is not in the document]
ralexstokes: yep, good catch

setup.py
ralexstokes: we may not -- i'll check and clean this up after figuring out what to do w/ the `setup.py` changes

ralexstokes:
> Post-#851: Should it be marked by `pytest.skip`?
> 
> https://github.com/ethereum/trinity/blob/9edd694d6fbb5b147cce85adc0e1f1f562429744/tests/eth2/fixtures/test_shuffling.py#L20-L22

the issue here is that the code will currently fail if we try to load multiple configuration types as we overwrite global instances of the SSZ bounds for our various vectors, lists, etc.

i think even w/ `pytest.skip` it will still generate the test, causing the conflicting update and then proceed to not run the test, even though the side-effect of updating the SSZ bounds has occurred.

one solution here is to never run the fixture tests in parallel and then just change the bounds before each run based on the config... do you like this approach better? @hwwhww 

ultimately, we need to move away from the globally-scope SSZ bounds (see #786) but given priority for interop stuff soon i'd think we handle that at the earliest in the few weeks. in the meantime, we only check-in one set of fixtures tests by configuration and can manually manage it locally as we run different versions of the tests

ralexstokes:
> Post-#851
> 
> Each case is set with `"config_types": (Minimal,),`. e.g.,
> 
> https://github.com/ethereum/trinity/blob/9edd694d6fbb5b147cce85adc0e1f1f562429744/tests/eth2/fixtures/test_epoch_processing.py#L17
> 
> I think we'd unlikely to "test `registry_updates` with `Minimal` and test `slashings` with `Mainnet`".
> How about creating a parameter for it? i.e., `"config_types": config_types`. Or even [adding pytest options](https://docs.pytest.org/en/latest/example/simple.html#pass-different-values-to-a-test-function-depending-on-command-line-options) so that we can set it by command line.

We can change which `ConfigType` is in that field or can provide a `Tuple[ConfigType]` to handle multiple types of configurations in a given testrun... the constraint here is laid out above w/ conflicting configuration types.... ideally we run all tests w/ both configuration types

eth2/beacon/tools/fixtures/test_generation.py
```diff
@@ -87,65 +163,64 @@ def _generate_test_suite_descriptors_from(
...
     test_type: TestType[HandlerType],
     handler_type: TestHandler[Input, Output],
+    suite_name: str,
     config_type: ConfigType,
+    fork_type: ForkType,
     test_case: TestCase,
 ) -> Tuple[TestCase, str]:
+    """
+    id format:
+      f"{TEST_TYPE_NAME}_{HANDLER_TYPE_NAME}_{TEST_SUITE_NAME}_{TEST_CASE_NAME}:{CONFIG_TYPE_NAME}_{FORK_TYPE_NAME}"  # noqa: E501
```
hwwhww: Why not use the directory format? https://github.com/ethereum/eth2.0-specs/tree/v0.8.3/specs/test_formats#test-structure
Except for
- We don't have the optional `runner`.
- For alignment, move `TEST_TYPE_NAME` to the first.

```python
f"{TEST_TYPE_NAME}_{CONFIG_TYPE_NAME}_{FORK_TYPE_NAME}_{HANDLER_TYPE_NAME}_{TEST_SUITE_NAME}_{TEST_CASE_NAME}"
```

eth2/beacon/tools/fixtures/test_generation.py
```diff
@@ -87,65 +163,64 @@ def _generate_test_suite_descriptors_from(
...
-    else:
-        config_name = ""
+    handler_name = handler_type.name
 
     test_id_prefix = thread_last(
-        (test_name, handler_name, config_name),
+        (test_name, handler_name, suite_name, test_case.name),
         (filter, lambda component: component != ""),
         lambda components: "_".join(components),
     )
```
hwwhww: ❤️

eth2/beacon/tools/fixtures/test_types/ssz_static.py
```diff
@@ -1,10 +1,311 @@
...
+        Eth1DataHandler,
+        ForkHandler,
+        HistoricalBatchHandler,
+        IndexedAttestationHandler,
+        PendingAttestationHandler,
+        ProposerSlashingHandler,
+        TransferHandler,
+        ValidatorHandler,
+        VoluntaryExitHandler,
+    )
```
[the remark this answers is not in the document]
ralexstokes: it wasn't that bad haha, could do a lot of automation once the core names are written once (which can be grabbed from output of `ls`, etc.)

eth2/beacon/tools/fixtures/test_generation.py
```diff
@@ -44,9 +62,64 @@ def decorator(func: DecoratorTarget) -> DecoratorTarget:
...
+    NOTE: currently only takes one option but could be extended to respect a
+    comma-separated list of several configuration options.
+    """
+    config_str = metafunc.config.option.config
+    if config_str:
+        return (_config_from_str(config_str),)
+    return None
+
+
+def _keep_first_some(values: Sequence[Any]) -> Any:
```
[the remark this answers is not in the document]
ralexstokes: i mean the first element of the sequence that is _not_ `None`.

didn't put a ton of thought into this name so happy to field alternatives

eth2/beacon/tools/fixtures/test_generation.py
```diff
@@ -44,9 +62,64 @@ def decorator(func: DecoratorTarget) -> DecoratorTarget:
...
+    going in reverse to respect precedence.
+    """
+    if len(values) == 1:
+        return values[0]
+
+    for value in reversed(values):
+        if value:
+            return value
+    raise AssertionError(
+        "``_keep_some`` should find at least one valid option; check configuration."
```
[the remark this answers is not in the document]
ralexstokes: thank you :)

eth2/beacon/tools/fixtures/test_generation.py
ralexstokes: sure

setup.py
ralexstokes: yep this is great

ralexstokes:
going to go ahead and merge this (one failure is from RTD build that we know is failing)

Issue #897, opened 2019-08-09
Trinity leaves behind locked-up processes


carver:
* Trinity version: 96b1213ce1f4 (or before)

### What is wrong?

Trinity *appears* to shut down cleanly, but actually leaves a bunch of processes running. This is a quick and easy way to run out of RAM.

Example processes:
```sh
$ ps aux | grep 'trinity/venv'
jcarver  11964  0.0  0.0  33484 11132 pts/8    S    15:29   0:00 ./trinity/venv/bin/python3 -Wdefault -c from multiprocessing.semaphore_tracker import main;main(19)
jcarver  11965  0.1  0.2 262108 82552 pts/8    S    15:29   0:01 ./trinity/venv/bin/python3 -Wdefault -c from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=20, pipe_handle=24) --multiprocessing-fork
jcarver  11966  0.1  0.2 265108 85828 pts/8    S    15:29   0:02 ./trinity/venv/bin/python3 -Wdefault -c from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=20, pipe_handle=25) --multiprocessing-fork
...
```

### How can it be fixed

I suspect it's related to Beam Sync plugin processes that are locked up. Maybe we need some kind of return event that indicates a shutdown? (The processes might be stuck in a `lahja` `CollectMissingAccount` request that is waiting for a response)

cburgdorf:
Uh, that sounds horrible. I'll dive in.

>(The processes might be stuck in a lahja CollectMissingAccount request that is waiting for a response)

Trinity should eventually `sigkill` every process on shutdown. Not sure how that happens. Anyway, I'll see what I can figure out about that.

PR #1004, opened 2019-08-29
Remove global ProcessPoolExecutor

cburgdorf:
### What was wrong?

Trinity does currently leave processes behind. Read #897 for a more detailed description.

I used the following steps to reproduce the issue:

1. Check you have no dangling processes `ps aux |grep trinity`
2. If you do, kill them. (`pkill python` if you are brave)
3. Run beam sync (*should* work with other sync, too but I used that) until it is fully involved syncing
4. ☝ This is important, killing Trinity when it isn't yet syncing does not leave these processes behind.
5. Kill trinity with `CTRL + C`

Another important hint: One can check the process ids with these of the plugin processes. It makes it clear that those process ids are not ours. They really are from the `ProcessPoolExecutor`

### How was it fixed?

I've spent almost the entire day on this and it got me so frustrated that the outcome now is this removal of the entire `ProcessPoolExecutor`.

Here's the backstory:

1. The problem of leftover processes was originally discovered by @veox and back then it was even more severe as every isolated plugin would create a ton of idle processes.

2. I fixed this back then in https://github.com/ethereum/trinity/pull/227

3. Back then, @pipermerriam already [raised concerns](https://github.com/ethereum/trinity/pull/227#pullrequestreview-197099468) with the solution, noting that someone may accidentally remove it at some point. He also proposed using a `weakref.finalizer` for a more robust solution.

4. Ironically it is @pipermerriam 's PR under my review that then happens to do exactly that: [accidentally remove the fix](https://github.com/ethereum/trinity/commit/6493c94656b907ca9f73ce2fe677dfd9f03a2974#diff-31de14ce34c9eb0bd0c3a7fed15bda1eL141).

Now, I've spend a fair amount of time trying to come up with something more robust including the proposed `weakref.finalizer` solution. Unfortunately, it has proven to be really hard (or maybe that's just me) to come up with something that works reliably.

This got me questioning if we really need the `ProcessPoolExecutor` here or if a simple `ThreadPoolExecutor` might do it as well.

I haven't done any profiling about the performance impact of this.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1004.bugfix.rst
+++ b/newsfragments/1004.bugfix.rst
@@ -1,0 +1,7 @@
+Instead of the ``ProcessPoolExecutor`` use a ``ThreadPoolExecutor`` to normalize
+expensive messages. This fixes a bug where Trinity would leave idle processes
+from the ``ProcessPoolExecutor`` behind every time it shuts down after a sync.
+
+Performance wise, both methods should be roughly compareable and since many
+task have already been moved to their own managed processes over time, using
+a ``ThreadPoolExecutor`` strikes as a simple solution to fix that bug.
--- a/p2p/_utils.py
+++ b/p2p/_utils.py
@@ -1,8 +1,4 @@
 import collections
-from concurrent.futures import Executor, ProcessPoolExecutor
-import logging
-import os
-import signal
 from typing import Hashable, Sequence, Tuple, TypeVar
 
 import rlp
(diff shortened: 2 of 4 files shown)


pipermerriam:
I'm :+1: on trying this approach if the performance comparison is comprable.  I'll leave it to @carver to evaluate as he's the defacto "is this ok for beam sync" person.

carver:
:x: for CAP ;)

cburgdorf:
>For Beam Sync, every method that needs a separate process is in its own isolated process plugin, AFAICT.

Correct. And it's also the case that when the decision for the `ProcessPoolExecutor` was made, there were many things that were running in the same process that are running in their own processes today anyway. (request server, discovery, upnp, tx pool etc)

 
>:x: for CAP X)

Fixed that and added a release notes entry :+1:

carver:
Confirmed, fixes #897

PR #1005, opened 2019-08-29
Fix bug where variable might not be set

cburgdorf:
### What was wrong?

The `headers` var might end up not being initialized leaving us with an error when we try to start beam sync from a checkpoint.

### How was it fixed?

Set `headers` to `None` to start with.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/common/strategies.py
+++ b/trinity/sync/common/strategies.py
@@ -3,6 +3,7 @@ from abc import (
     abstractmethod
 )
 import asyncio
+import logging
 
 from eth_typing import (
     BlockNumber,
@@ -77,6 +78,8 @@ class FromCheckpointLaunchStrategy(SyncLaunchStrategyAPI):
 
     min_block_number = BlockNumber(0)
 
+    logger = logging.getLogger('trinity.sync.common.strategies.FromCheckpointLaunchStrategy')
+
     def __init__(self,
                  db: BaseAsyncHeaderDB,
                  chain: AsyncChainAPI,
@@ -113,11 +116,13 @@ class FromCheckpointLaunchStrategy(SyncLaunchStrategyAPI):
             except (TimeoutError, PeerConnectionLost, ValidationError):
                 # Nothing to do here. The ExchangeManager will disconnect if appropriate
                 # and eventually lead us to a better peer.
-                pass
+                continue
 
             if not headers:
-                await peer.disconnect(DisconnectReason.useless_peer)
-            elif headers[0].hash != self._checkpoint.block_hash:
+                self.logger.debug(
+                    "Disconnecting from %s. Returned no header while resolving checkpoint",
+                    peer
+                )
                 await peer.disconnect(DisconnectReason.useless_peer)
             else:
                 self.min_block_number = headers[0].block_number
--- a/newsfragments/1005.bugfix.rst
+++ b/newsfragments/1005.bugfix.rst
@@ -1,0 +1,3 @@
+Fix a bug where trying to start beam sync from a checkpoint would throw an error
+due to an uninitialized var if a request to a peer would raise an error while
+we are trying to resolve a header from it.


trinity/sync/common/strategies.py
```diff
@@ -103,6 +103,7 @@ def __init__(self,
                 await asyncio.sleep(2)
                 continue
 
+            headers = None
```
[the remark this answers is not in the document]
cburgdorf: Ah good catch handling it in the `except`. But I think we just want to turn the `pass` into a `continue` and leave it up to the `ExchangeManager` to decide when to disconnect. As I understand, the `ExchangeManager` uses better heuristics such as "How often did this peer time out". So one `Timeout` may be fine and the peer could return the correct header on our next try. It feels better to just let the `ExchangeManager` decide when to disconnect here.

The only cases where we disconnect here are:

1. The hash doesn't match what we request. I think @carver mentioned that the `ExchangeManager` already does that, too so this might be redundant (but might still live here for extra clarity?)

2. The peer does not return the header so it doesn't seem to be a useful peer for us.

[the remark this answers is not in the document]
cburgdorf: I responded to that here https://github.com/ethereum/trinity/pull/1005#discussion_r319385704

trinity/sync/common/strategies.py
```diff
@@ -77,6 +78,8 @@ class FromCheckpointLaunchStrategy(SyncLaunchStrategyAPI):
 
     min_block_number = BlockNumber(0)
 
+    logger = logging.getLogger('trinity.sync.common.strategies.FromGenesisLaunchStrategy')
```
carver: Wrong logger name?

trinity/sync/common/strategies.py
```diff
@@ -113,11 +116,19 @@ def __init__(self,
             except (TimeoutError, PeerConnectionLost, ValidationError):
                 # Nothing to do here. The ExchangeManager will disconnect if appropriate
                 # and eventually lead us to a better peer.
-                pass
+                continue
```
carver: > It feels better to just let the ExchangeManager decide when to disconnect here.

Yeah, I agree with that. The more real-world testing I do, the less aggressive I am about disconnecting. :+1: to `continue`

... but note that peers can lie about their TD, which means they can DoS this code by supplying a high TD and ~~returning no headers~~ timing out. We would just keep asking the same peer over and over.

This is a reason to go back to the simple model of iterating through all the peers, since we aren't actively disconnecting them.

trinity/sync/common/strategies.py
```diff
@@ -113,11 +116,19 @@ def __init__(self,
...
 
             if not headers:
+                self.logger.debug(
+                    "Disconnecting from %s. Returned no header while resolving checkpoint",
+                    peer
+                )
```
carver: > The peer does not return the header so it doesn't seem to be a useful peer for us.

In this case, I think we can just debug log and `continue`, instead of disconnecting.

trinity/sync/common/strategies.py
```diff
@@ -113,11 +116,19 @@ def __init__(self,
...
                 await peer.disconnect(DisconnectReason.useless_peer)
             elif headers[0].hash != self._checkpoint.block_hash:
+                self.logger.debug(
+                    "Disconnecting from %s. Returned wrong header while resolving checkpoint",
+                    peer
+                )
```
carver: > The hash doesn't match what we request. I think @carver mentioned that the ExchangeManager already does that, too so this might be redundant (but might still live here for extra clarity?)

Yeah, well, what I said was: 
> This is already validated internally to the get_block_headers() call, but I guess no objection to re-validating it here, since it's cheap and only happens once.

So I'm actually slightly happier to just remove this entire `elif` block.

trinity/sync/common/strategies.py
cburgdorf: >iterating through all the peers,

The reason I went for the current model is because it is *slightly* simpler and I haven't found it to be a problem in practice. The code runs only on startup and if it fails within 1000 tries it would eventually fail the boot and we would notice. But that has never happened during testing so I'm not sure if we really need it.

trinity/sync/common/strategies.py
cburgdorf: Sure, removed it!

trinity/sync/common/strategies.py
carver: Yeah, I suppose in practice it's difficult for a naughty peer to connect to you before you ask for this header. But if they succeed in connecting, the ability to so easily DoS you makes me uncomfortable.

PR #1006, opened 2019-08-29
Fix transport corruption

### LINKED ISSUE #1001: A MalformedMessage escaping during sync
on master

### What is wrong?

```
   ERROR  08-28 16:13:14            Connection  Unexpected error in <p2p.connection.Connection object at 0x7f3a00b1f198>, exiting
  File "/home/jcarver/code/trinity/p2p/transport.py", line 241, in recv
    header = self._decrypt_header(header_data)
  File "/home/jcarver/code/trinity/p2p/transport.py", line 322, in _decrypt_header
    f'Invalid header mac: expected {expected_header_mac}, got {header_mac}'
p2p.exceptions.DecryptionError: Invalid header mac: expected b'\xbe\xe5\x7fmM\xe9\xc6\x05\xbc\xb2\x93P:2b\xdf', got b"\xa1\xd1o\x06'`\xfd\x8f\xcd~<\xb82\x96\xa8\x95"

The above exception was the direct cause of the following exception:

  ...
  File "/home/jcarver/code/trinity/p2p/multiplexer.py", line 322, in _do_multiplexing
    async for protocol, cmd, msg in msg_stream:
  File "/home/jcarver/code/trinity/p2p/cancellable.py", line 66, in wait_iter
    timeout=timeout,
  File "/home/jcarver/code/trinity/p2p/cancellable.py", line 20, in wait
  File "/home/jcarver/code/trinity/p2p/cancellable.py", line 42, in wait_first
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/cancel_token/token.py", line 153, in cancellable_wait
    return done.pop().result()
  File "/home/jcarver/code/trinity/p2p/multiplexer.py", line 58, in stream_transport_messages
    raw_msg = await transport.recv(token)
  File "/home/jcarver/code/trinity/p2p/transport.py", line 247, in recv
    raise MalformedMessage from err
```

### How can it be fixed

Related to #901 ?

cc @pipermerriam

pipermerriam:
fixes #1001 

### What was wrong?

When exiting the `multiplexer.multiplex()` context blocks, we were issueing a hard cancellation to the `Transport.recv`.  If this happened to occur in-between reading the header and body, then the transport would become "corrupt".  On the next `Transport.recv` call it would *think* that it was reading the message header, but would actually be reading into the body.

### How was it fixed?

The `Transport` object now exposes a `Transport.read_state` which can be in any of the three `IDLE/HEADER/BODY` states.  When we exit the multiplexing context we issue a soft cancellation first, allowing the message stream loop to exit naturally.

After that we issue a *harder* cancellation via `fut.cancel()`.  If the transport is simply waiting for a header, this cancellation is handled and the state is reset to `IDLE`.  This should work fine since a cancellation at this stage should not result in us having read any bytes from the `StreamReader`.  If I'm wrong about this then we'll start getting `MalformedMessage` exceptions again.

After the `fut.cancel()` we issue a cancellation via the `CancelToken` mechanism (this is not handled and will corrupt the transport if it isn't `IDLE`).

Then, the multiplexer checks whether the transport has been corrupted, logs some info loudly and closes itself.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

Halloween Dog 1

### DIFF
--- a/p2p/multiplexer.py
+++ b/p2p/multiplexer.py
@@ -39,6 +39,7 @@ from p2p.exceptions import (
 from p2p.p2p_proto import BaseP2PProtocol
 from p2p.protocol import Protocol
 from p2p.resource_lock import ResourceLock
+from p2p.transport_state import TransportState
 from p2p.typing import Payload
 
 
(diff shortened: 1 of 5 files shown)


p2p/multiplexer.py
```diff
@@ -338,3 +389,6 @@ def stream_protocol_messages(self,
                     protocol,
                     cmd,
                 )
+
+            if stop.is_set():
```
pipermerriam: If anyone has any better ideas on how I should *signal* the *stop* in a better way let me know.  This use of `asyncio.Event()` seems odd since I'm not using the `Event.wait()` API at all...

p2p/multiplexer.py
```diff
@@ -291,14 +292,53 @@ def stream_protocol_messages(self,
...
+                #
+                # In the HEADER case we can issue a hard stop either via
+                # cancellation or the cancel token.  The read *should* be
+                # interrupted without consuming any bytes from the
+                # `StreamReader`.
+                #
+                # In the BODY case we want to give the `Transport.recv` call a
+                # moment to finish reading the body after which it will be IDLE
+                # and will exit via the IDLE exit mechanism.
+                stop.set()
```
carver: I don't understand why this works. It triggers the stop event after leaving the multiplexer. Doesn't that only happen after you've already decided to exit the peer?

If a message that was working before (that wouldn't have caused a peer exit) started failing, how does it work to add a fix that triggers after the peer exits?

I think I'm missing something simple and fundamental.

pipermerriam: This is the part you are missing:

https://github.com/ethereum/trinity/blob/d322e7975283f4bb2f89180a881b0a17c3f66d3e/p2p/handshake.py#L303-L310

The multiplexer exiting doesn't mean that the peer is exiting.  The multiplexer actually activates twice, once during handshake, and then again inside of the `Connection`.

So the `stop.set()` here allows the multiplexer to exit cleanly between message boundaries.

p2p/multiplexer.py
carver: The only other idea I had is a `self._is_closing: bool` that gets set to `True` when the context manager exits. Probably a wash:
- object state instead of argument :-1: 
- simpler value, rather than "over-using" `asyncio.Event()` :+1:

newsfragments/1006.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixes issue where exiting ``Multiplexer.multiplex()`` would cause the underlying ``Transport`` to become corrupt.
```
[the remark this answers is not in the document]
cburgdorf: Agreed, bugfixes in release notes (to me) only make sense if the bug was present in the previous release.

Issue #805, opened 2019-07-16
docs build failing


carver:
### What is wrong?

readthedocs build is failing, eg~
https://readthedocs.org/projects/trinity-client/builds/9381206/

Because:
```
Building wheels for collected packages: blspy
  Building wheel for blspy (setup.py): started
  Building wheel for blspy (setup.py): finished with status 'error'
  ERROR: Complete output from command /home/docs/checkouts/readthedocs.org/user_builds/trinity-client/envs/latest/bin/python -u -c 'import setuptools, tokenize;__file__='"'"'/tmp/pip-install-57700gbe/blspy/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /tmp/pip-wheel-4ppoyr_f --python-tag cp36:
  ERROR: running bdist_wheel
  running build
  running build_ext
  Traceback (most recent call last):
    File "/tmp/pip-install-57700gbe/blspy/setup.py", line 22, in run
      out = subprocess.check_output(['cmake', '--version'])
    File "/home/docs/.pyenv/versions/3.6.8/lib/python3.6/subprocess.py", line 356, in check_output
      **kwargs).stdout
    File "/home/docs/.pyenv/versions/3.6.8/lib/python3.6/subprocess.py", line 423, in run
      with Popen(*popenargs, **kwargs) as process:
    File "/home/docs/.pyenv/versions/3.6.8/lib/python3.6/subprocess.py", line 729, in __init__
      restore_signals, start_new_session)
    File "/home/docs/.pyenv/versions/3.6.8/lib/python3.6/subprocess.py", line 1364, in _execute_child
      raise child_exception_type(errno_num, err_msg, err_filename)
  FileNotFoundError: [Errno 2] No such file or directory: 'cmake': 'cmake'
  
  During handling of the above exception, another exception occurred:
  
  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "/tmp/pip-install-57700gbe/blspy/setup.py", line 78, in <module>
      zip_safe=False,
    File "/home/docs/checkouts/readthedocs.org/user_builds/trinity-client/envs/latest/lib/python3.6/site-packages/setuptools/__init__.py", line 145, in setup
      return distutils.core.setup(**attrs)
[...]

PR #1020, opened 2019-08-30
Fix RTD build by excluding blspy dependency for RTD

cburgdorf:
### What was wrong?

Turns out the RTD build has been failing for 4 weeks (we should have a system in place that informs us about failed RTD builds!).

It turns out that `blspy` regained the hard dependency against `cmake`

See #805

### How was it fixed?

Excluded `blspy` from RTD builds. I also turned the the code a little bit around so that we *remove* the dependencies for the RTD environments instead of conditionally adding them for non-RTD environments. Even though that mutates a list, I think it is clearer to have all the dependencies visually grouped together. This also means that `python-snappy` moved back into the `p2p` section where it belongs.

I tested this with a branch directly in this repo and it indeed fixes the build.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 import os
+import re
 from setuptools import setup, find_packages
 
 PYEVM_DEPENDENCY = "py-evm==0.3.0a5"
@@ -18,6 +19,7 @@ deps = {
         "eth-keys>=0.2.4,<0.3.0",
         "netifaces>=0.10.7<1",
         "pysha3>=1.0.0,<2.0.0",
+        "python-snappy>=0.5.3",
         "SQLAlchemy>=1.3.3,<2",
         'trio==0.11.0,<0.12',
         'trio-typing>=0.2.0,<0.3',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -118,11 +120,24 @@
     ],
 }
 
-# NOTE: Snappy breaks RTD builds. Until we have a more mature solution
-# we conditionally add python-snappy based on the presence of an env var
-rtd_build_env = os.environ.get('READTHEDOCS', False)
-if not rtd_build_env:
-    deps['p2p'].append("python-snappy>=0.5.3")
```
cburgdorf: Previously we conditionally *added* dependencies if we were NOT in an RTD environment. Considering that this is the exceptional case and that it is clearer to read if all dependencies are visually together in one group, I moved this around so that we are now *remove* dependencies when we are in an RTD environment.

Sure, I'm not a fan of the list mutation either but considering pro and cons and this is "just" the `setup.py` file I think that's ok :)

cburgdorf: Alternatively we *could* do something like this to avoid the mutation:

```python
deps['eth2'= filter_dependencies(deps['eth2'], 'blspy')
deps['p2p'= filter_dependencies(deps['p2p'], 'python-snappy')
```

carver: Yeah, I guess a slight preference for this ^ approach. I am worried that the regex will be brittle, but it's not critical for it to be a general solution, so it's probably not worth putting more time into.

cburgdorf: >Yeah, I guess a slight preference for this ^ approach

Updated!

>I am worried that the regex will be brittle

Naah, I've put all my Regex fu into it! :joy:

PR #1026, opened 2019-09-01
Discv5 ping handler

jannikluhn:
Builds on top of #1025. Only the last two commits are new.

This PR implements the basic architecture of the routing table manager and the first of its components, the ping handler.

The routing table manager is quite simple: All it does is start a bunch of subservices that do the real work. The components derive from a common base class which contains some shared functionality. Originally, I've implemented all functionality in a single class, but it got quite big for no good reason, so I split it up.

For every incoming ping, the ping handler's job is to

- add the sender to the routing table or update its position in the routing table
- respond with an appropriate pong message
- if the ENR we have of the peer is outdated based on the sequence number in the ping message, ask for an update with a FindNode request

#### Cute Animal Picture

animal-animal-photography-avian-2679504

### DIFF
--- a/p2p/discv5/constants.py
+++ b/p2p/discv5/constants.py
@@ -36,3 +36,5 @@ MAX_REQUEST_ID_ATTEMPTS = 100  # number of attempts we take to guess a available
 # ENR keys for endpoint information
 IP_V4_ADDRESS_ENR_KEY = b"ip"
 UDP_PORT_ENR_KEY = b"udp"
+
+REQUEST_RESPONSE_TIMEOUT = 0.5  # timeout for waiting for response after request was sent
--- a/p2p/discv5/routing_table_manager.py
+++ b/p2p/discv5/routing_table_manager.py
@@ -1,5 +1,4 @@
 import logging
-import time
 
 from eth_utils import (
     encode_hex,
--- a/p2p/discv5/routing_table_manager.py
+++ b/p2p/discv5/routing_table_manager.py
@@ -1,0 +1,270 @@
+import logging
+import time
+
+from eth_utils import (
+    encode_hex,
+)
+
+import trio
+from trio.abc import (
+    SendChannel,
+)
+
+from p2p.trio_service import (
+    Service,
+)
+
+from p2p.discv5.abc import (
+    EnrDbApi,
+    MessageDispatcherAPI,
+)
+from p2p.discv5.channel_services import (
+    IncomingMessage,
+    OutgoingMessage,
+)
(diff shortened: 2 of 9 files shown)


p2p/discv5/constants.py
```diff
@@ -36,3 +36,5 @@
 # ENR keys for endpoint information
 IP_V4_ADDRESS_ENR_KEY = b"ip"
 UDP_PORT_ENR_KEY = b"udp"
+
+REQUEST_RESPONSE_TIMEOUT = 0.5  # timeout for waiting for response after request was sent
```
pipermerriam: This feels very short....  Not sure of exact context but it seems reasonable to encounter network conditions that make round trips take longer than half a second.

p2p/discv5/routing_table_manager.py
```diff
@@ -0,0 +1,270 @@
...
+                 enr_db: EnrDbApi,
+                 outgoing_message_send_channel: SendChannel[OutgoingMessage]
+                 ) -> None:
+        super().__init__(local_node_id, routing_table, message_dispatcher, enr_db)
+        self.outgoing_message_send_channel = outgoing_message_send_channel
+
+    async def run(self) -> None:
+        channel_handler_subscription = self.message_dispatcher.add_request_handler(PingMessage)
+        async with channel_handler_subscription:
+            async for incoming_message in channel_handler_subscription:
```
pipermerriam: Nice API! :smile:

p2p/discv5/routing_table_manager.py
```diff
@@ -0,0 +1,270 @@
...
+            outgoing_message_send_channel=outgoing_message_send_channel,
+            **shared_component_kwargs,
+        )
+
+    async def run(self) -> None:
+        components = (
+            self.ping_handler,
+        )
+        for component in components:
+            self.manager.run_daemon_task(component.run)
```
pipermerriam: I think this is worth implementing `Manager.run_child_service`.  This allows us to generate a graph of the service hierarchy assuming each service keeps track of which child services it is running...

p2p/discv5/constants.py
jannikluhn: It's what the spec suggests. I think the timeout has to be short to keep lookup times short. If a request times out we just try another node. This should not be a big deal and to some extent expected for some fraction of all requests. But I guess we have to see how it turns out in a live network.

p2p/discv5/routing_table_manager.py
jannikluhn: I'm going to open an issue for this.

PR #1028, opened 2019-09-01
Discv5 Ping Sender

jannikluhn:
Builds on top of #1027, please only review the last commit.

This implements the last component for the most basic version of the routing table manager. It regularly pings the oldest node in the routing table. If they respond in time, they are pushed to the top.

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/routing_table_manager.py
+++ b/p2p/discv5/routing_table_manager.py
@@ -18,11 +18,13 @@ from p2p.discv5.abc import (
     MessageDispatcherAPI,
 )
 from p2p.discv5.channel_services import (
+    Endpoint,
     IncomingMessage,
     OutgoingMessage,
 )
 from p2p.discv5.constants import (
     REQUEST_RESPONSE_TIMEOUT,
+    ROUTING_TABLE_PING_INTERVAL,
 )
 from p2p.discv5.endpoint_tracker import (
     EndpointVote,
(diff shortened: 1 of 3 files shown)


p2p/discv5/routing_table_manager.py
```diff
@@ -289,6 +296,79 @@ def __init__(self,
...
+    async def run(self) -> None:
+        while True:
+            deadline = trio.current_time() + ROUTING_TABLE_PING_INTERVAL
+
+            if len(self.routing_table) > 0:
+                node_id = self.routing_table.get_oldest_entry()
+                self.logger.debug("Pinging %s", encode_hex(node_id))
+                await self.ping(node_id)
+            else:
+                self.logger.warning("Routing table is empty, no one to ping")
```
pipermerriam: This looks like the type of message that will cause a warning most of the time during startup.  Thoughts on how we can delay this a little to be sure that it isn't throwing erronious warnings before the routing table has really had any time to initialize (not urgent for this PR but worth addressing at some point)

p2p/discv5/routing_table_manager.py
```diff
@@ -289,6 +296,79 @@ def __init__(self,
...
+                 routing_table: FlatRoutingTable,
+                 message_dispatcher: MessageDispatcherAPI,
+                 enr_db: EnrDbApi,
+                 endpoint_vote_send_channel: SendChannel[EndpointVote]
+                 ) -> None:
+        super().__init__(local_node_id, routing_table, message_dispatcher, enr_db)
+        self.endpoint_vote_send_channel = endpoint_vote_send_channel
+
+    async def run(self) -> None:
+        while True:
```
pipermerriam: *idle thoughts not necessarily for this PR*

This might be a neat async iterable.

```python
async for moment in every_N_seconds(ROUTING_TABLE_PING_INTERVAL):
    ...
```

An async iterable that did the interval sleeps internally.

jannikluhn: Ooh, that would be very nice indeed :+1:

p2p/discv5/routing_table_manager.py
jannikluhn: I was thinking we could just store the routing table on disk or fill it with boot nodes on first start up.

pipermerriam: :+1: just pointing it out.  We had a similar issue with discv4 and fixed it by not raising warnings until the service uptime exceeded 60 seconds or something like that. (I don't think we've implemented uptime for the trio service...)

PR #1030, opened 2019-09-02
libp2p connection peer factory

ChihChengLiang:
### What was wrong?

Need factories to create connected Libp2p peers for testing.

### How was it fixed?

Fix some `get_head_state` bug btw

#### Cute Animal Picture

### DIFF
--- a/trinity/tools/bcc_factories.py
+++ b/trinity/tools/bcc_factories.py
@@ -102,6 +102,7 @@ class NodeFactory(factory.Factory):
 @asynccontextmanager
 async def ConnectionPairFactory(
     cancel_token: CancelToken = None,
+    say_hello: bool = True,
 ) -> AsyncIterator[Tuple[Node, Node]]:
     if cancel_token is None:
         cancel_token = CancelTokenFactory()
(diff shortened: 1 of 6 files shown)


trinity/tools/bcc_factories.py
```diff
@@ -87,7 +90,7 @@ class Meta:
     cancel_token = None
     bootstrap_nodes = None
     preferred_nodes = None
-    chain = None
+    chain = factory.SubFactory(BeaconChainFactory)
```
NIC619: With this change we also remove and replace the `nodes_with_chain` fixture with `nodes` fixture. (We set `chain`  to `None` because some of the tests do not need a functioning `chain`. We used `nodes_with_chain` for the tests that need one.)

trinity/tools/bcc_factories.py
```diff
@@ -96,6 +99,21 @@ def create_batch(cls, number: int) -> Tuple[Node, ...]:
         )
 
 
+@asynccontextmanager
+async def ConnectionPairFactory(
+    cancel_token: CancelToken = None,
+) -> AsyncIterator[Tuple[Node, Node]]:
+    if cancel_token is None:
+        cancel_token = CancelTokenFactory()
```
NIC619: Curious about when will the token be triggered(if it's generated here but is not accessible outside the factory)?

trinity/tools/bcc_factories.py
ChihChengLiang: Will remove the fixture and update the related tests with the new factory.

trinity/tools/bcc_factories.py
ChihChengLiang: It's triggered in `run_service`

PR #1037, opened 2019-09-03
Document how to run custom networks

### LINKED ISSUE #83: Expand docs to demonstrate how to use custom genesis files
### What is wrong?

We recently added support for custom genesis files. Still need docs.

https://github.com/ethereum/py-evm/pull/1299

### How can it be fixed

:writing_hand: :wine_glass: :writing_hand: :wine_glass:

cburgdorf:
### What was wrong?

We are lacking docs on how to run private networks 

Fixes #83

### How was it fixed?

1. Wrote a guide that demonstrates how to setup a devnet using a custom genesis file
2. Created a cross link from the Cookbook (for added visibility)

peek1TKE7Z

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/cookbook.rst
+++ b/docs/cookbook.rst
@@ -26,13 +26,29 @@ Use the ``--sync-mode=light`` flag to instruct Trinity to run as a light node.
 Ropsten vs Mainnet
 ------------------
 
-Trinity currently only supports running against either the Ethereum Mainnet or
-Ropsten testnet.  Use ``--ropsten`` to run against Ropsten.
+Trinity supports running against either the Ethereum Mainnet or
+Ropsten testnet out of the box.  Use ``--ropsten`` to run against Ropsten.
 
 
 .. code:: sh
 
   trinity --ropsten
+
+
+.. _cookbook_recipe_running_a_custom_network:
+
+
+Running a custom network
+------------------------
+
+Trinity can run custom networks via configuration file per
+`EIP-1085 <https://github.com/ethereum/EIPs/issues/1085>`_.
+
+.. code:: sh
+
+  trinity --genesis <PATH-TO-EIP-1085-CONFIG> --data-dir <PATH-TO-DATA-DIR> --network-id <NETWORK-ID>
+
+For a more detailed writeup, checkout the guide about :doc:`Creating a custom testnet</guides/creating_a_custom_testnet>`
 
 
 .. _cookbook_recipe_connecting_to_preferred_nodes:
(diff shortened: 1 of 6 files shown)


cburgdorf:
@pipermerriam Here's the full guide on our custom genesis config feature that I promised. Let me try to justify the `db-shell` change:

1. It makes the ergonomics of the example much simpler. Without this, you either have to setup a script, parse the genesis file to a chain config etc or (preferrable) write a plugin that would mine a few blocks. It can be done but it blows up the example

2. The `db-shell` is a developer tool anyway, a swiss army knife if you will. I don't think it hurts to have a mining chain exposed there.

3. This gives us a good target to optimize against. If we can *avoid* creating blocks manually because Trinity can mine them autonomously (preferable using PoA) then we can simplify this guide further. So far, this is the simplest I can think of that actually builds a "net" between two nodes.

docs/cookbook.rst
```diff
@@ -26,15 +26,30 @@ Use the ``--sync-mode=light`` flag to instruct Trinity to run as a light node.
...
   trinity --ropsten
 
 
+.. _cookbook_recipe_running_a_custom_network:
+
+
+Running a custom network
+------------------------
+
+Trinity can run custom networks via configuration file per EIP1085.
```
pipermerriam: Would be good to link `EIP1085` to the actual EIP

docs/guides/creating_a_custom_testnet.rst
```diff
@@ -0,0 +1,202 @@
...
+        - `trinity_config`: `TrinityConfig` instance
+        - `chain_config`: `ChainConfig` instance
+        - `chain`: `Chain` instance
+        - `mining_chain: `MiningChain` instance. (use a REPL to create blocks)
+
+  In [1]:
+
+Let's just copy the following lines and paste them in and hit enter.
+
+.. code:: sh
```
pipermerriam: why not `code:: python`?

docs/cookbook.rst
cburgdorf: :+1:

docs/guides/creating_a_custom_testnet.rst
cburgdorf: Copy and pasting like a pro!

cburgdorf:
Btw, it's super cool that with this new RTD build we can link to a preview of the docs for the PR prior to merging! :tada: 

https://external-builds.readthedocs.io/html/trinity-client/1037/guides/creating_a_custom_testnet.html

PR #1040, opened 2019-09-03
Fix how trio services manage tasks

### LINKED ISSUE #1032: Trio service task exception handling
### What is wrong?

When a coroutine run inside a trio service with `run_task` raise an exception, the manager will catch and swallow it. Neither the service itself or anyone else will know that the task has failed. For daemon tasks, the service will be cancelled, but the reason for this is still hidden.

For both normal and daemon tasks, I think this is surprising. As an example, see [this snippet](https://github.com/ethereum/trinity/blob/6963342e2baaaad82c13b6457acbd1f8a3a12a57/p2p/discv5/packer.py#L612-L627) which looks quite natural, but is broken (because it's not the service that raises the `HandshakeFailure`, but one of its subtasks).

### How can it be fixed

I'm not sure. At first I thought we can just reraise the exception in the manager. This might work for the example above, but it still doesn't allow the service to handle the exception.

I haven't thought this through, but maybe we can create something similar to a cancel scope for the service's subtasks that can be used to `await` the exceptions raised in tasks. But this seems quite complicated and confusing.

The easiest and currently my favorite solution is to get rid of `run_task` completely and instead create task nurseries manually. This also removes one layer of abstraction, so the service design would become a bit simpler. The only thing we lose seems to be a bit of logging.

pipermerriam:
fixes #1032 

### What was wrong?

This addresses a few things.

- The `TrioService` incorrectly allowed background tasks to error out without causing the whole service to cancel itself.
- A service run using `await Manager.run()` didn't end up having the `stopped` event set when an **exception** occurs. 

### How was it fixed?

Added two new tests.

1. a task that errors to show it causes the whole service to exit with the exception.
2. a daemon task that finishes early to show it causes the whole service to exit.

Functionality to ensure this behavior is followed by the service.

Also added a `try/finally` block to the main `Manager.run()` body to ensure that `stopped` gets properly set.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

a15187aefb02e2fee1e99e68197d1fe9--cats-in-costumes-animals-in-costumes

### DIFF
--- a/p2p/trio_service.py
+++ b/p2p/trio_service.py
@@ -400,6 +400,8 @@ class Manager(ManagerAPI):
                 err,
                 exc_info=True,
             )
+            self._run_error = sys.exc_info()
+            self.cancel()
         else:
             self.logger.debug(
                 "task '%s[daemon=%s]' finished.",
@@ -408,7 +410,12 @@ class Manager(ManagerAPI):
             )
         finally:
             if daemon:
-                self.cancel()
+                self.logger.debug(
+                    "daemon task '%s' exited unexpectedly.  Cancelling service: %s",
+                    name,
+                    self,
+                )
+                raise LifecycleError(f"Daemon task {name} exited")
 
     def run_task(self,
                  async_fn: Callable[..., Awaitable[Any]],
--- a/p2p/trio_service.py
+++ b/p2p/trio_service.py
@@ -3,7 +3,7 @@ import functools
 import logging
 import sys
 from types import TracebackType
-from typing import Any, Callable, Awaitable, Optional, Tuple, Type, AsyncIterator
+from typing import Any, Callable, Awaitable, List, Optional, Tuple, Type, AsyncIterator
 
 from async_generator import asynccontextmanager
 
@@ -20,6 +20,13 @@ class ServiceException(Exception):
 
 
 class LifecycleError(ServiceException):
+    """
+    Raised when an action would violate the service lifecycle rules.
+    """
+    pass
+
+
+class DaemonTaskExit(ServiceException):
     """
     Raised when an action would violate the service lifecycle rules.
     """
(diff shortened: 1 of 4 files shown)


pipermerriam:
@cburgdorf RTD CI build failure?

cburgdorf:
>RTD CI build failure?

Sweet. They have activated the feature for this repo (and apparently I had nothing to do, it just magically appeared).

The failure here seems that in this RTD build it looks like the environment variable `READTHEDOCS` wasn't set and so it tried to install blspy and failed.

Looking at the past builds, it seems as if someone at RTD has noticed this (maybe this was a bug with this new beta feature) and fixed it.

somone_fixed_it

The one thing that I don't know is how to re-trigger the RTD build for this PR. But I'll find out. I think you should feel free to ignore the failed build for this exceptional case in this PR.

p2p/trio_service.py
```diff
@@ -309,34 +319,38 @@ def __init__(self, service: ServiceAPI) -> None:
...
+                finally:
+                    # Mark as having stopped
+                    self._stopped.set()
         self.logger.debug('%s stopped', self)
 
         # If an error occured, re-raise it here
         if self.did_error:
-            _, exc_value, exc_tb = self._run_error
-            raise exc_value.with_traceback(exc_tb)
+            raise trio.MultiError(tuple(
```
jannikluhn: At the moment, trio collapses multi errors to a plain exception if there's only a single one. According to [this discussion](https://github.com/python-trio/trio/issues/611), they are aware that this is problematic and want to change it, but maybe it's better for now to mirror trio's behavior?

pipermerriam: what would mirroring that behavior look like?  I was aware of the collapsing behavior.  I wasn't aware they were changing it, but I think this code behaves appropriately under either paradigm.

Issue #1046, opened 2019-09-04
Beam sync following the ETH Classic chain


cburgdorf:
### What is wrong?

I ran a Trinity node with beam sync and it apparently synced the ETH Classic chain.

![eth_classic](https://user-images.githubusercontent.com/521109/64231163-208b0c80-ceef-11e9-8048-d09440fb5712.png)

https://gastracker.io/block/8717984

Notice that I synced this node from genesis (**not** using `--beam-from-checkpoint`).

I thought we have a protection in place to disconnect when we notice the other node is on the wrong network but apparently there is something not working as it should.

Here are the full debug logs that have lead to this state:

[beam_ending_on_etc.zip](https://github.com/ethereum/trinity/files/3573673/beam_ending_on_etc.zip)

And here is my investigation:

Ok, here's what I found:

Our Skeleton peer is on the right ETH chain and races past the DAO point:

```shell
DEBUG  09-02 14:24:29        SkeletonSyncer  Requsting chain of headers from ETHPeer <Node(0xb1ac@150.109.100.120)> starting at #2210376
```

At some point we persist an **ETC** header and this is the last time that we see `HeaderOnlyPersist` doing its job.

```shell
    INFO  09-02 14:25:04     HeaderOnlyPersist  Imported 388 headers in 0.36 seconds, new head: <BlockHeader #1919771 8fc507d3>
[1m[33m WARNING  09-02 14:25:04      HeaderMeatSyncer  ETHPeer <Node(0x1525@176.99.6.205)> returned segment starting <BlockHeader #1920354 82f4f14a> & parent 347f..81f2, doesn't match <BlockHeader #1920353 286700a1>, ignoring result...[0m
   DEBUG  09-02 14:25:04      HeaderMeatSyncer  Pausing ETHPeer <Node(0x1525@176.99.6.205)> for 15.0s, for sending 0 headers
   DEBUG  09-02 14:25:04      HeaderMeatSyncer  Requesting 192 headers from ETHPeer <Node(0xb1ac@150.109.100.120)>
    INFO  09-02 14:25:04     HeaderOnlyPersist  Imported 386 headers in 0.32 seconds, new head: <BlockHeader #1920157 f71b68ab>
```

This is the **ETC header** `<BlockHeader #1920157 f71b68ab>`

Our skeleton peer, continues to be a valid ETH peer and is continue to serve us but we do not persist anything anymore.

```shell
[...]

PR #1050, opened 2019-09-04
Fix issue where Trinity accidentially ends up as ETC client

cburgdorf:
### What was wrong?

As demonstrated in #1046 Trinity might end up syncing the ETC chain by accident.

### How was it fixed?

Turns out, we were not checking the extradata in the post-DAO headers.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/eip1085-utils/test_preconfigured_genesis_files.py
+++ b/tests/core/eip1085-utils/test_preconfigured_genesis_files.py
@@ -2,10 +2,17 @@ import json
 
 import pytest
 
+from eth_utils import ValidationError
+
 from eth.chains.base import Chain
-from eth.chains.mainnet import MainnetChain, MAINNET_GENESIS_HEADER
+from eth.chains.mainnet import (
+    HOMESTEAD_MAINNET_BLOCK,
+    MainnetChain,
+    MAINNET_GENESIS_HEADER,
+)
 from eth.chains.ropsten import RopstenChain, ROPSTEN_GENESIS_HEADER
 from eth.db.atomic import AtomicDB
+from eth.rlp.headers import BlockHeader
 from eth.vm.forks.homestead import HomesteadVM
 
 from trinity.config import (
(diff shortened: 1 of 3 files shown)


trinity/_utils/eip1085.py
```diff
@@ -129,8 +132,7 @@ def _extract_vm_config(vm_config: Dict[str, str]) -> Iterable[VMFork]:
         if 'DAOForkBlock' in vm_config:
             dao_fork_block_number = to_int(hexstr=vm_config['DAOForkBlock'])
 
-            HomesteadVM = BaseHomesteadVM.configure(
```
cburgdorf: Notice that `support_dao_fork` defaults to `True` hence the removal

pipermerriam:
Thoughts on adding a test for this so that we don't accidentally regress at some point?

cburgdorf:
>Thoughts on adding a test for this so that we don't accidentally regress at some point?

yeah, that's probably a good idea. I'll come up with something tomorrow (End of day + kiddo around)

trinity/_utils/eip1085.py
```diff
@@ -129,8 +132,7 @@ def _extract_vm_config(vm_config: Dict[str, str]) -> Iterable[VMFork]:
...
-                support_dao_fork=True,
+            HomesteadVM = MainnetDAOValidatorVM.configure(
```
carver: Nice find! Something still strikes me as problematic, though, because testnets with the DAO fork might be validated differently. (Or maybe this `VM` is badly named?)

cburgdorf: Maybe `HomesteadWithDAOValidatorVM` would be clearer?

carver: Yeah, I think this docstring is not super helpful, since there isn't any meaningful example or definition of running the dao fork on a non-mainnet (since it affects specific accounts that were in the mainnet state):
https://github.com/ethereum/py-evm/blob/f23355050e975b559b2140fc0a4ca717165e06d7/eth/chains/mainnet/__init__.py#L42-L43
So I guess I'm in favor of just importing that header-checking logic into the base homestead vm in py-evm, and getting rid of the concept of the `MainnetDAOValidatorVM`

Also, I notice that the default homestead code seems to support the dao fork, but the eip1085 code near the PR seems to think it defaults false:
https://github.com/ethereum/py-evm/blob/f23355050e975b559b2140fc0a4ca717165e06d7/eth/vm/forks/homestead/__init__.py#L20-L21

cburgdorf: >So I guess I'm in favor of just importing that header-checking logic into the base homestead vm in py-evm, and getting rid of the concept of the MainnetDAOValidatorVM

Now I'm confused. My understanding is that the general `HomesteadVM` shouldn't know about this DAO check because it has nothing to do with the VM rules per se. E.g. a `HomesteadVM` on Ropsten or Görli wouldn't bother about this DAO check.

Doesn't this leave us with?

- `HomesteadVM`
  - does not validate the `extra_data`
  - implements `get_dao_fork_block_number` as a clutch so that we can generalize

- `MainnetHomesteadVM`
  - like `HomesteadVM` but does the `extra_data` check

So, if we want to consolidate the classes then I'd understand to get rid of `MetaHomesteadVM` and move that functionality directly into `HomesteadVM`

And I guess we could also rename `MainnetDAOValidatorVM` to `MainnetHomesteadVM`so that we set the `DAO_FORK_MAINNET_BLOCK` there directly instead of inheriting `MainnetHomesteadVM` from `MainnetDAOValidatorVM` (because `mainnet` is the only one with that extra check anyway)

carver: > And I guess we could also rename `MainnetDAOValidatorVM` to `MainnetHomesteadVM`so that we set the `DAO_FORK_MAINNET_BLOCK` there directly instead of inheriting `MainnetHomesteadVM` from `MainnetDAOValidatorVM` (because `mainnet` is the only one with that extra check anyway)

Agree with all this ^ :+1: 

---

Yeah, on second thought, I'm basically fine with the code as is (though I think the rename you suggested in that quote is correct). My original comment came from the confusion that there would be a dao fork on any of the testnets (of course not). So I guess `MainnetHomesteadVM` is good. We might have to rename it if there's a new eth-classic-lite, which follows the dao fork but doesn't follow some other more recent fork.

tests/core/eip1085-utils/test_preconfigured_genesis_files.py
```diff
@@ -58,6 +65,72 @@ def test_mainnet_eip1085_matches_mainnet_genesis_header(mainnet_genesis_config):
     assert actual_homestead_vm.get_dao_fork_block_number() == expected_homestead_vm.get_dao_fork_block_number()  # noqa: E501
 
 
+def test_mainnet_eip1085_rejects_etc_homestead_header(mainnet_genesis_config):
+    PRE_FORK_HEADER = BlockHeader(
```
cburgdorf: We have these headers in the tests of `py-evm` but I guess I can't (and probably shouldn't!) import from there. It got me thinking if something like a `eth-constants` library would be useful. A library purely dedicated to useful constants for the Python Ethereum ecosystem. Nothing for this PR but I thought I'd mention it here.

trinity/_utils/eip1085.py
cburgdorf: I'm inclined to just leave it as is. The EIP1085 has `DAOForkBlock` as a valid breaking point so *theoretically* one is free to use it to define custom networks that define the `DAOForkBlock` at block `666` if they like. Not sure if that's of any practical use (maybe for simulating specific things regarding the fork). With that in mind, I guess we could leave the classes as they are and just rename `MainnetDAOValidatorVM` to `HomesteadDAOValidatorVM` to get rid of the direct `Mainnet` relation.

Probably not worth holding back this PR though.

I also added a test that ensures the VM we derive from `mainnet.json` doesn't let ETC headers pass. I also double checked that this test fails on current master.

cburgdorf:
Looks like we might cut a new Trinity release today. I'm going to merge this. Happy to address any issues in a follow up PR.

tests/core/eip1085-utils/test_preconfigured_genesis_files.py
carver: Agreed! I actually nabbed the name over a year ago but haven't had time to work on it yet:
https://pypi.org/project/eth-constants/#history

trinity/_utils/eip1085.py
carver: > I also double checked that this test does indeed fail on current master as it should.

Awesome, it makes me so happy when folks do this :)

tests/core/eip1085-utils/test_preconfigured_genesis_files.py
cburgdorf: Ha! Great!

PR #1052, opened 2019-09-04
ABC base class for HandshakerAPI

pipermerriam:
extracted from #966 

### What was wrong?

Found need to formalize the `Handshaker` with an ABC base class

### How was it fixed?

Created `ABC` base class for it.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

cowboy

### DIFF
--- a/p2p/handshake.py
+++ b/p2p/handshake.py
@@ -1,4 +1,3 @@
-from abc import ABC, abstractmethod
 import asyncio
 import functools
 import logging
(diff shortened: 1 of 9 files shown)


p2p/handshake.py
```diff
@@ -57,25 +57,25 @@
 )
 
 
-class Handshaker(ABC):
+class HandshakeReceipt(HandshakeReceiptAPI):
```
cburgdorf: I guess `HandshakeReceipt` shouldn't be in here because it already lives in `p2p/receipt.py`.

pipermerriam: good catch.  bad rebase on my part. fixing

Issue #373, opened 2019-03-07
Unclean shutdowns


lithp:
* py-evm Version: 0e389db67b671d51d56c35837f496e03bd088e75
* Python Version (python --version): 3.7.2

### What is wrong?

I think everyone is already aware of this issue, just opening it to have a record somewhere and have somewhere to track progress.

Every time I ctrl-c out of trinity I'm faced with a LOT of output:

```
^C    INFO  03-06 17:13:36               trinity  Shutting down Trinity (CTRL+C / Keyboard Interrupt)
    INFO  03-06 17:13:36         PluginManager  Shutting down PluginManager with scope <class 'trinity.extensibility.plugin_manager.MainAndIsolatedProcessScope'>
    INFO  03-06 17:13:36         PluginManager  Stopping plugin: JSON-RPC API
    INFO  03-06 17:13:36         PluginManager  Stopping plugin: Discovery
Process SpawnProcess-3:
    INFO  03-06 17:13:36         PluginManager  Sent SIGINT to process 21158, waiting 10 seconds for it to terminate
    INFO  03-06 17:13:36         PluginManager  Sent SIGINT to process 21207, waiting 10 seconds for it to terminate
   ERROR  03-06 17:13:36               asyncio  Task exception was never retrieved
future: <Task finished coro=<<async_generator_asend without __name__>()> exception=StopAsyncIteration()>
StopAsyncIteration
StopAsyncIteration
   ERROR  03-06 17:13:36               asyncio  Task exception was never retrieved
future: <Task finished coro=<<async_generator_asend without __name__>()> exception=StopAsyncIteration()>
StopAsyncIteration
StopAsyncIteration
   ERROR  03-06 17:13:36               asyncio  Task exception was never retrieved
future: <Task finished coro=<<async_generator_asend without __name__>()> exception=StopAsyncIteration()>
StopAsyncIteration
StopAsyncIteration
    INFO  03-06 17:13:36     DiscoveryProtocol  stopping discovery
Traceback (most recent call last):
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/process.py", line 300, in _bootstrap
    util._exit_function()
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/util.py", line 299, in _exit_function
    _run_finalizers(0)
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/util.py", line 265, in _run_finalizers
[...]

PR #1061, opened 2019-09-05
Fix _GatheringFuture exceptions during shutdown

lithp:
### What was wrong?

During shutdown triinty would emit two sets of:

```
    ERROR  09-04 19:40:12               asyncio  _GatheringFuture exception was never retrieved
future: <_GatheringFuture finished exception=CancelledError()>
concurrent.futures._base.CancelledError
```

[Long explanation that could probably be skipped]

Before this PR, `BaseService.wait()` wrapped the calls to `asyncio.gather()`. `BaseService.wait()` eventually turns into a call to `CancelToken.cancellable_wait()`. When the token is cancelled (which is what happens during trinity shutdown) `cancellable_wait()`, tries to cancel all the futures. This is usually fine, however, `asyncio.gather()` returns a special kind of future. When a `_GatheringFuture` is cancelled it sets an exception on itself! This is why we're seeing `exception was never retrieved`, the cancel token is cancelling the future and assumes that it can then forge about the future, but instead the future, when cancelled, raises an exception.

There are a couple ways forward, and I don't know which is best:

- I think `asyncio.gather()` was only ever intended to be used as `await asyncio.gather(...)`, we could replace our calls to `gather()` with a call to something else that's better behaved.
- I don't think `asyncio.gather(..., return_exceptions=True)` would have this behavior, we could add that flag and be careful to raise exceptions which look important.
- `CancelToken` could hard-code knowledge of how `_GatheringFuture` works. It could cancel the future and then let the event loop spin a few times and then retrieve the cancelled exception.

[/long explanation]

### How was it fixed?

In this PR I opted to remove the call to `cancellable_wait()`. It's fine in this situation, because each of the coros which are being gathered are awaiting on the same token, they'll be cancelled and the cancellation will trickle up through the call to `gather()`. I think that in general calling `gather()` without `self.wait()` is okay, as long as the coros being gathered call `wait`. (Just like how it's fine to call `await self.xxx()` methods, as long as those methods eventually block on `self.wait()`)

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

black-tailed-prairie-dog

### DIFF
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -203,10 +203,10 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
             await self.wait(rate_limiter.take())
 
             try:
-                await self.wait(asyncio.gather(*(
+                await asyncio.gather(*(
                     self._add_peers_from_backend(backend)
                     for backend in self.peer_backends
-                )))
+                ))
             except OperationCancelled:
                 break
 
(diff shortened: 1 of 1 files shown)


lithp:
Part of the #373 train.

PR #1064, opened 2019-09-05
Make beacon node runnable

NIC619:
### How was it fixed?
- [x] Add back `ReceiveServer`
- [x] Deactivae `BCCRequestServer`
- [x] Various fixed to make beacon node runnable again

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/eth2/beacon/tools/builder/validator.py
+++ b/eth2/beacon/tools/builder/validator.py
@@ -505,7 +505,7 @@ def _create_mock_signed_attestation(
         )
     else:
         privkeys = tuple(keymap.values())
-        attesting_indices = _attesting_indices
+        attesting_indices = _attesting_indices  # type: ignore
 
     # Use privkeys to sign the attestation
     signatures = [
--- a/eth2/beacon/tools/builder/validator.py
+++ b/eth2/beacon/tools/builder/validator.py
@@ -359,9 +359,8 @@ def create_mock_slashable_attestation(
         crosslink=previous_crosslink,
     )
 
-    message_hash, attesting_indices = _get_mock_message_and_attesting_indices(
-        attestation_data, committee, num_voted_attesters=1
-    )
+    message_hash = _get_mock_message(attestation_data)
+    attesting_indices = _get_mock_attesting_indices(committee, num_voted_attesters=1)
 
     signature = sign_transaction(
         message_hash=message_hash,
@@ -457,18 +456,23 @@ def _get_target_root(
         )
 
 
-def _get_mock_message_and_attesting_indices(
-    attestation_data: AttestationData,
-    committee: Sequence[ValidatorIndex],
-    num_voted_attesters: int,
-) -> Tuple[Hash32, Tuple[CommitteeIndex, ...]]:
+def _get_mock_message(attestation_data: AttestationData) -> Hash32:
     """
-    Get ``message_hash`` and voting indices of the given ``committee``.
+    Get ``message_hash`` of ``attestation_data``.
     """
     message_hash = AttestationDataAndCustodyBit(
         data=attestation_data, custody_bit=False
     ).hash_tree_root
 
+    return message_hash
+
+
+def _get_mock_attesting_indices(
+    committee: Sequence[ValidatorIndex], num_voted_attesters: int
+) -> Tuple[CommitteeIndex, ...]:
+    """
+    Get voting indices of the given ``committee``.
+    """
     committee_size = len(committee)
     assert num_voted_attesters <= committee_size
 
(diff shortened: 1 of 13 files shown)


eth2/beacon/tools/builder/validator.py
```diff
@@ -488,6 +488,8 @@ def _create_mock_signed_attestation(
     num_voted_attesters: int,
     keymap: Dict[BLSPubkey, int],
     slots_per_epoch: int,
+    is_for_simulation: bool = True,
+    _attesting_indices: Sequence[ValidatorIndex] = None,
```
hwwhww: It's weird variable naming here. Perhaps (i) use `attesting_indices` for this input and (ii) rename the old `attesting_indices` to `simulation_attesting_indices`?

eth2/beacon/tools/builder/validator.py
```diff
@@ -496,17 +498,26 @@ def _create_mock_signed_attestation(
         attestation_data, committee, num_voted_attesters
     )
```
hwwhww: Since `attesting_indices` is only used for the simulation cases, I'd say refactor 
https://github.com/ethereum/trinity/blob/20f20b790b09f3b36c124b58302a5454340e187b/eth2/beacon/tools/builder/validator.py#L475-L480 to be `if is_for_simulation: ...` case.

trinity/protocol/bcc_libp2p/topic_validators.py
```diff
@@ -40,21 +41,8 @@ def beacon_block_validator(msg_forwarder: ID, msg: rpc_pb2.Message) -> bool:
...
             return False
 
-        state_machine = chain.get_state_machine()
-        config = state_machine.config
-        slots_per_epoch = config.SLOTS_PER_EPOCH
-        state = chain.get_head_state()
-        head_slot = state.slot
-        current_epoch = compute_epoch_of_slot(head_slot, slots_per_epoch)
-        block_epoch = compute_epoch_of_slot(block.slot, slots_per_epoch)
-        if block_epoch > current_epoch:
```
NIC619: @hwwhww I remove this check as at epoch boundary this would be a problem(some peers are one slot before epoch boundary and hence reject the block). Or perhaps relaxed the condition to block_epoch > current_epoch + 1? But I suppose one or two epochs doesn't matter much as validators' balance are going to be updated at epoch boundary anyway.

cburgdorf:
It looks like the `fastecdsa` dependency of `libp2p` breaks the RTD build (hooray, that we catch these now!).

Maybe we can exclude `libp2p` here:

https://github.com/ethereum/trinity/blob/248efac0cc6daf43e49e5739a80f11737dd06161/setup.py#L141

cburgdorf:
Ah damn it, I think the git dependency broke my bullet proof regex 

https://github.com/ethereum/trinity/blob/248efac0cc6daf43e49e5739a80f11737dd06161/setup.py#L131

Reading this will put a smile on @carver 's face in the morning :see_no_evil:

cburgdorf:
Something like `[!=<>@ ](.|)+` *might* serve us better :)

eth2/beacon/tools/builder/validator.py
```diff
@@ -488,33 +492,46 @@ def _create_mock_signed_attestation(
     num_voted_attesters: int,
     keymap: Dict[BLSPubkey, int],
     slots_per_epoch: int,
+    is_for_simulation: bool = True,
```
mhchia: Could you elaborate more why we need this flag?

carver:
> Ah damn it, I think the git dependency broke my bullet proof regex
> 
> https://github.com/ethereum/trinity/blob/248efac0cc6daf43e49e5739a80f11737dd06161/setup.py#L131
> 
> Reading this will put a smile on @carver 's face in the morning

Lol, that didn't take long.

now you have another problem

If this ever breaks again, let's nuke it from orbit and find another solution.

eth2/beacon/tools/builder/validator.py
NIC619: `_create_mock_signed_attestation` is used by both tests and beacon node validator and it was configured to randomly sample validators to attest but this does not suit beacon node validator's need so this flag is added to indicate if `_create_mock_signed_attestation` is used by tests(i.e., simulation) or beacon node validator.

trinity/protocol/bcc_libp2p/servers.py
```diff
@@ -183,11 +183,17 @@ def __init__(
         await self.cancellation()
 
     async def _handle_beacon_attestation_loop(self) -> None:
+        while PUBSUB_TOPIC_BEACON_ATTESTATION not in self.topic_msg_queues:
+            await self.sleep(1.0)
+
```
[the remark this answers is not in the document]
NIC619: I have `Node.start()` to be scheduled during instantiation of `Node`(instead of being scheduled after `asyncio.ensure_future(libp2p_node.run()`) but still no 100% guarantee so I added `Node.is_started` for `ReceiveServer` to check before starting the loop.

NIC619:
Merging this one. Feel free to leave comments/reviews here and I will fix them in subsequent PR.

PR #1072, opened 2019-09-06
Better mechanism for closing incoming handshake streams

pipermerriam:
### What was wrong?

It was possible for us to leave the streams open for incoming peer connections because the closing of them was manually done by ensuring that we call an ugly `_cleanup...` function in each of the exception blocks.

### How was it fixed?

Made an inner exception block that is purely responsible for closing the connections that re-raises.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

326560714_ulkYzOvf_c

### DIFF
--- a/trinity/server.py
+++ b/trinity/server.py
@@ -150,21 +150,23 @@ class BaseServer(BaseService, Generic[TPeerPool]):
     async def receive_handshake(
             self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter) -> None:
 
-        def _cleanup_reader_and_writer() -> None:
-            if not reader.at_eof():
-                reader.feed_eof()
-            writer.close()
-
         try:
-            await self._receive_handshake(reader, writer)
+            try:
+                await self._receive_handshake(reader, writer)
+            except Exception:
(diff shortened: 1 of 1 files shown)


trinity/server.py
```diff
@@ -150,21 +150,24 @@ def _make_peer_pool(self) -> TPeerPool:
...
         try:
-            await self._receive_handshake(reader, writer)
+            try:
+                await self._receive_handshake(reader, writer)
+            except Exception:
+                if not reader.at_eof():
+                    reader.feed_eof()
+                writer.close()
+                await asyncio.sleep(0)
+                raise
```
cburgdorf: Oh that's a clever pattern that I had never considered. The one thing I'm wondering is why this introduces a `await asyncio.sleep(0)` that wasn't needed before. If it's needed, maybe a comment would help. I think in general, placing `await asyncio.sleep(0)` in non-test code should usually be accompanies by a comment.

pipermerriam: Not sure it's strictly necessary, removing and we can re-add another time if it seems like it fixes something.

PR #1074, opened 2019-09-07
Catch all PeerConnectionLost in try_with_node

carver:
Failures here would stop the peer pool events from working at all, if
a send action were to get a PeerConnectionLost, for example.

### How was it fixed?
When handling requests via the server, catch and drop `PeerConnectionLost` exceptions.

Hm, as I'm writing this, I realize it may be too aggressive, and drop PCLs related to other peers during handling.

Anyway, we need to do something here, because unhandled exceptions cause the PeerPoolEventServer loop to break, and stops handling events.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -387,10 +387,16 @@ workflows:
   version: 2
   test:
     jobs:
+      # These tests are long, so should be started first to optimize for total suite run time
+      - py36-integration
+      - py36-wheel-cli
+      - py37-wheel-cli
+      - py36-long_run_integration
+      - py37-rpc-state-sstore
+
       - py36-docs
 
       - py37-eth1-core
-      - py37-wheel-cli
       - py37-p2p
       - py36-p2p-trio
       - py37-eth2-core
@@ -402,7 +408,6 @@ workflows:
       - py37-eth1-plugins
 
       - py37-rpc-state-quadratic
-      - py37-rpc-state-sstore
       - py37-rpc-state-zero_knowledge
 
       - py36-rpc-state-byzantium
@@ -415,7 +420,6 @@ workflows:
       - py36-rpc-blockchain
 
       - py36-eth1-core
-      - py36-wheel-cli
       - py36-p2p
       - py36-p2p-trio
       - py36-eth2-core
@@ -425,9 +429,7 @@ workflows:
       - py36-eth2-plugins
       - py36-eth1-plugins
 
-      - py36-integration
       - py36-lightchain_integration
-      - py36-long_run_integration
 
       - py36-lint
       - py36-lint-eth2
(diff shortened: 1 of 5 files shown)


carver:
So maybe a better place to do this is here: 
https://github.com/ethereum/trinity/blob/67ffacca5d022684f4ee8c42bc46605a898a2792/trinity/protocol/common/peer_pool_event_bus.py#L163-L164

We could catch *all* exceptions and continue, since almost all handler exceptions should not cause the handler loop to break. Even things like `CancelledError` and `OperationCancelled` aren't obvious reasons to break the loop. Of course, the exceptions should be logged to `logging.exception`, but then, we could continue the loop.

carver:
> We could catch _all_ exceptions and continue, since almost all handler exceptions should not cause the handler loop to break. Even things like `CancelledError` and `OperationCancelled` aren't obvious reasons to break the loop. Of course, the exceptions should be logged to `logging.exception`, but then, we could continue the loop.

Okay, it changed fairly significantly with this, but luckily the PR is still relatively small.

trinity/_utils/decorators.py
```diff
@@ -1,8 +1,41 @@
...
+    Type,
 )
 
 
 class classproperty(property):
     def __get__(self, obj: Any, objtype: Any = None) -> Any:
         return super().__get__(objtype)
+
+
+def suppress_exceptions(*exception_types: Type[BaseException]) -> Callable[..., Any]:
```
pipermerriam: So many `Any` types.  Can't you use a `TypeVar['TReturn']` on these to maintain the type information about the return value?



```python
TReturn = TypeVar("TReturn")
TFn = TypeVar("TFn", bound=Callable[..., TReturn])


def suppress_exceptions(*exception_types: Type[BaseException]) -> Callable[[TFn], TFn]:
    def _suppress_exceptions(fn: TFn) -> TFn:
        @functools.wraps(fn)
        def wrapper(*args: Any, **kwargs: Any) -> TReturn:
            try:
                return fn(*args, **kwargs)
            except exception_types:
                pass
        return wrapper
    return _suppress_exceptions
```

Tried this out and it does retain the function type correctly but looks like it also needs some `# type: ignore` comments to make `mypy` happy.
def supress

carver:
Hm, this caused an XFAIL'd test to pass, which means CI turns read. I *think* it might be a fluke. Re-running it.

.circleci/config.yml
```diff
@@ -387,10 +387,16 @@ workflows:
   version: 2
   test:
     jobs:
+      # These tests are long, so should be started first to optimize for total suite run time
+      - py36-integration
+      - py36-wheel-cli
+      - py37-wheel-cli
+      - py36-long_run_integration
+      - py37-rpc-state-sstore
+
```
cburgdorf: Nice! :+1:

PR #1075, opened 2019-09-07
Several logging improvements

carver:
### What was wrong?

Needed more info on screen

### How was it fixed?

- Show the lag, even if the imported block was non-canonical
- Log the round trip time needed to get each node during beam import (perhaps the most critical measure of health)
- Add debug log showing which peers are serving most of the beam sync data (this is almost always one peer: the fastest one)

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


pipermerriam:
Shipit

Sent from ProtonMail mobile

PR #1078, opened 2019-09-08
Add HTTPServer

hwwhww:
### What was wrong?

* For eth2 validator APIs (https://github.com/ethereum/eth2.0-APIs), we need a base HTTP Server.
* The interface could be either JSON-RPC or HTTP RESTful. Now it's set to JSON-RPC. 
    * For JSON-RPC, @cburgdorf has made some example for us: https://github.com/ethereum/trinity/blob/master/trinity/rpc/modules/beacon.py

@pipermerriam @ralexstokes 
While I think this PR should be shared with eth1, the interop is more time-sensitive... which means I'm included to *not* test it with eth1 APIs in this PR. What do you think about it?

### How was it fixed?
* Add `HTTPServer`. `--enable-http` is `False` by default.
* Default port: `8545`

### To-Do

- [x] Clean up commit history

- [x] Add some basic tests
- [ ] `aiohttp.StreamReader`?

#### Cute Animal Picture

dolphin-203875_640

### DIFF
--- a/trinity/rpc/main.py
+++ b/trinity/rpc/main.py
@@ -110,7 +110,6 @@ class RPCServer:
         """
         try:
             validate_request(request)
-            logging.debug(f'request: {request}')
 
             if request.get('jsonrpc', None) != '2.0':
                 raise NotImplementedError("Only the 2.0 jsonrpc protocol is supported")
@@ -118,11 +117,9 @@ class RPCServer:
             method = self._lookup_method(request['method'])
             params = request.get('params', [])
 
-            # TODO: Beacon chain doesn't support retry now
-            if not isinstance(self.chain, BaseAsyncBeaconChainDB):
-                result = await execute_with_retries(
-                    self.event_bus, method, params, self.chain,
-                )
+            result = await execute_with_retries(
+                self.event_bus, method, params, self.chain,
+            )
 
             if request['method'] == 'evm_resetToGenesisFixture':
                 result = True
--- a/trinity/rpc/main.py
+++ b/trinity/rpc/main.py
@@ -15,6 +15,7 @@ from eth_utils import (
 )
 
 from trinity.chains.base import AsyncChainAPI
+from trinity.db.beacon.chain import BaseAsyncBeaconChainDB
 from trinity.rpc.modules import (
     BaseRPCModule,
 )
@@ -65,7 +66,7 @@ class RPCServer:
 
     def __init__(self,
                  modules: Sequence[BaseRPCModule],
-                 chain: AsyncChainAPI=None,
+                 chain: Union[AsyncChainAPI, BaseAsyncBeaconChainDB],
                  event_bus: EndpointAPI=None) -> None:
         self.event_bus = event_bus
         self.modules: Dict[str, BaseRPCModule] = {}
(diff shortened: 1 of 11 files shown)


trinity/rpc/main.py
```diff
@@ -109,6 +109,7 @@ def _lookup_method(self, rpc_method: str) -> Any:
         """
         try:
             validate_request(request)
+            logging.debug(f'request: {request}')
```
hwwhww: will remove this line later.

pipermerriam:
Help me understand why we *need* HTTP (as opposed to our current status quo using JSON-RPC over an IPC socket?

(I'm fine with doing whatever is fast to make this work, but we plan to use something like dopple to achieve HTTP based JSON-RPC so any HTTP server that gets added now very likely ends up getting removed later...)

ralexstokes:
@pipermerriam @hwwhww and i were chatting about this and were aiming to do whatever was fast just so we can align w/ other clients sooner rather than later...

its conceivable that we, e.g. just refactor this next week to run over dopple

hwwhww:
@pipermerriam 
* The current *interoperable* validator APIs are defined [in HTTP RESTful format](https://github.com/ethereum/eth2.0-APIs/blob/master/apis/validator/beacon-node-validator-api.md). And the HTTP based APIs seems to be favored in the long term. (more info: https://github.com/ethereum/eth2.0-APIs/pull/3).
    * While the validator API standard has not been finalized yet, I'd bet HTTP RESTful APIs will be the endgame.
    * For interop, we can accelerate our implementation with JSON-RPC scheme that we already have now.
    * IMO HTTP is also more flexible for us to integrate with https://github.com/ethereum/eth2.0-metrics or other monitor service (like ethstats) for interop.
* Aww I forgot that dopple exists. Thank you for the reminder!
    * Do you mean you plan to have an "HTTP -> IPC -> RPC" proxy? (like dopple)? If so, isn't this "HTTP -> RPC" workflow more efficient than "HTTP -> IPC -> RPC"? Or anything I missed?

pipermerriam:
> Do you mean you plan to have an "HTTP -> IPC -> RPC" proxy like dopple?

Yes, that's the loose plan.  The idea is that a well maintained version of a tool like `dopple` would allow layering an HTTP API on top of any arbitrary IPC based JSON-RPC API making it work for any client.  This lets every client avoid having to *also* ship an HTTP server as well as avoids all the little nuanced differences that can arise from that (for example parity and geth required different HTTP headers at some point)

hwwhww:
@pipermerriam oh I see. Thank you for answering. :)
My intention was having `HTTPServer` to support both JSON-RPC and HTTP RESTful formats for us. I can see how dopple-like proxy will help eth1. But once eth1 uses the dopple-like proxy, the proxy may not be able to convert RESTful to IPC and follow the eth2 standard. That is, if dopple-like proxy gets implemented, we might still need another HTTP server for supporting RESTful format.

I'd say keeping this HTTPServer for now and we’ll see where things go. 😄

cburgdorf:
Coincidentally, I made a [comment](https://github.com/ethereum/dopple/issues/1#issuecomment-528833434) in the dopple repo last Friday that I'd be looking into packaging it up so that we can re use it in Trinity for a generic solution to expose our IPC based APIs through HTTP.

I can't tell yet if this ends up satisfying all your eth2 needs but we can revisit this then and decide.

PR #1081, opened 2019-09-09
Add Goodbye req and enable/handle stream close/reset

NIC619:
### How was it fixed?
pending on https://github.com/libp2p/py-libp2p/pull/287

- [x] Add `Node.disconnect_peer()`
- [x] Add `say_goodbye` and `_handle_goodbye`
    - will disconnect from the peer regardless how the request ends
- [x] Handle possible `stream.close()`/`stream.reset()` cases during sending/handling request/response

### To-Do

- [ ] Clean up commit history


trinity/protocol/bcc_libp2p/configs.py
```diff
@@ -82,6 +82,12 @@ class ResponseCode(Enum):
     SERVER_ERROR = 2
 
 
+class GoodbyeReasonCode(IntEnum):
```
ralexstokes: minor suggestion but it is worthwhile to use `enum.unique` decorator?

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -305,6 +316,7 @@ def peer_store(self) -> PeerStore:
 
     def _register_rpc_handlers(self) -> None:
         self.host.set_stream_handler(REQ_RESP_HELLO_SSZ, self._handle_hello)
+        self.host.set_stream_handler(REQ_RESP_GOODBYE_SSZ, self._handle_goodbye)
```
ralexstokes: nice :)

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -860,24 +977,36 @@ def _get_requested_beacon_blocks(
...
+        except (ReadMessageFailure, MplexStreamEOF, MplexStreamReset) as error:
+            has_error = True
+            if isinstance(error, ReadMessageFailure):
+                await stream.reset()
+            elif isinstance(error, MplexStreamEOF):
+                await stream.close()
+        finally:
+            if has_error:
+                self.logger.info("Request recent beacon blocks failed: fail to read the response")
+                raise RequestFailure("fail to read the response")
```
ralexstokes: seems worthwhile to abstract out this pattern into some kind of rpc wrapper -- not enough to block this merge but something to keep in mind in the future

PR #1084, opened 2019-09-09
Remove fast sync and make beam sync default

### LINKED ISSUE #1042: Make beam sync the default
### What is wrong?

Default sync mode is `fast`.  We have no intention of making "fast" sync mode work anytime soon.

### How can it be fixed

Remove "fast" from the options and make "beam" the default?

cburgdorf:
### What was wrong?

Fast sync is still the default sync strategy while we actually have no plan to continue working on it.

Closes #1042 

### How was it fixed?

Removed it and made beam sync the default.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


tests/integration/test_trinity_cli.py
```diff
@@ -195,7 +195,7 @@
     )
 )
 @pytest.mark.asyncio
-async def test_does_not_throw(async_process_runner, command):
+async def test_no_errors(async_process_runner, command):
```
cburgdorf: Gotta love some `OSError: AF_UNIX path too long` for no obvious reason (nothing changed here :man_shrugging: )

pipermerriam: I want...

```
from eth_utils import temporary_ipc_path

with temporary_ipc_path as path:
    ...
```

Something simple which is guaranteed to be short enough and to be cleaned up when the context exits...

carver:
`trinity/sync/full/hexary_trie.py` is the main thing I was thinking of

tests/integration/test_trinity_cli.py
cburgdorf: The thing here is that I don't understand how this PR changes anything about the `path`..anyway, there must be *something*.

However, on the issue, my understanding was that we want to preserve meaningful subdirectories for the test runs to aid debugging.

Currently, these look like:

`/tmp/pyt--circleci/0/popen-gw0/test_does_not_throw_command0_0/t/ropsten/ipcs-eth1/jsonrpc.ipc`

If we don't care about these then, I don't think we need another tool as the `tempfile.TemporaryDirectory` already does seem to maintain the properties we need.

E.g. 

```
with tempfile.TemporaryDirectory() as tmpdir:
        print(tmpdir)
```

...gives me `/tmp/tmpsydzmq74` and is cleaned up after the test run. I personally don't care about these long semantic paths for the trinity root directory so I'd actually be happy to just do that.

cburgdorf: If we don't care about the `TRINITY_XDG_ROOT` being under the directory hierarchy that pytest goes with, it could be that simple: https://github.com/ethereum/trinity/pull/1084/commits/45fbd8da51278415cf35011b9a9fc8dacbb21daf

cburgdorf: Oh, now I know *why* this error came up. Because we are now running beam sync and that causes different processes!

cburgdorf:
@carver yep, this wasn't yet ready for review yesterday. I think it now is. 

Some notes:

1. We had a pre-check in our full sync that would kick off state sync if we notice that we do not have the state root in the db. I removed that so that sync simply exits and tells the user to run beam sync instead.

2. Since we are using beam sync now for all tests, the unix path too long for IPC paths came up again. I added a commit that simply proposes to put the trinity root directory into a temporary directory without the whole bloated directory hierarchy (`/tmp/user/<some-incremental-num>/<process-something>/<test-name-with-params>/`). Without this, we have lots of space and don't need to juggle with the test names.

3. Our shutdowns are getting cleaner which causes our shutdown error scan to occasionally fail the xfail. I tried removing the `xfail` entirely and hunt the remaining issues but they were to hard to crack for me in this PR. Instead, I added `ResourceWarning` as another thing to scan for so that the `xfail` should not become flaky.

tests/integration/helpers.py
```diff
@@ -15,6 +15,7 @@
 
     error_trigger = (
         "exception was never retrieved",
+        "ResourceWarning: unclosed resource",
```
pipermerriam: :+1:

PR #1085, opened 2019-09-09
Use hash tree root for Node

ChihChengLiang:
This builds on top of #1077

### What was wrong?

Node should request and handle blocks with hash_tree_roots 

### How was it fixed?

- add chain.get_block_by_hash_tree_root
- fix node

#### Cute Animal Picture

### DIFF
--- a/.flake8-eth2
+++ b/.flake8-eth2
@@ -1,4 +1,4 @@
 [flake8]
-max-line-length= 100
+max-line-length= 120
 exclude=
 ignore=W503,E203
--- a/.flake8-eth2
+++ b/.flake8-eth2
@@ -1,4 +1,4 @@
 [flake8]
-max-line-length= 120
+max-line-length= 100
 exclude=
 ignore=W503,E203
--- a/eth2/beacon/db/chain.py
+++ b/eth2/beacon/db/chain.py
@@ -93,6 +93,7 @@ class BaseBeaconChainDB(ABC):
     def get_slot_by_root(self, block_root: SigningRoot) -> Slot:
         pass
 
+    @abstractmethod
     def get_block_signing_root_by_hash_tree_root(
         self, block_root: HashTreeRoot
     ) -> SigningRoot:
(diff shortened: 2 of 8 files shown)


eth2/beacon/db/chain.py
```diff
@@ -612,6 +640,18 @@ def _add_block_root_to_slot_lookup(db: DatabaseAPI, block: BaseBeaconBlock) -> N
...
+    @staticmethod
+    def _add_block_hash_tree_root_to_signing_root_lookup(
+        db: DatabaseAPI, block: BaseBeaconBlock
+    ) -> None:
+        hash_tree_root_to_signing_root_key = SchemaV1.make_block_hash_tree_root_to_signing_root_lookup_key(
+            block.hash_tree_root
+        )
+        db.set(
+            hash_tree_root_to_signing_root_key,
+            ssz.encode(block.signing_root, sedes=ssz.sedes.bytes32),
```
NIC619: I wonder do we need to ssz encode block root? (Or we simply ssz encode every value?)
If that's the case, after a quick lookup, we seem to be storing plain block root for keys `make_canonical_head_root_lookup_key`, `make_finalized_head_root_lookup_key` and `make_justified_head_root_lookup_key`. Do you mind doing a quick fix on that along in this PR?

ChihChengLiang: We don't really have to `ssz.encode(block.signing_root, sedes=ssz.sedes.bytes32)` the result doesn't change. Should get rid of the unnecessary encodes.

eth2/beacon/db/chain.py
```diff
@@ -93,6 +93,11 @@ def get_block_by_root(
     def get_slot_by_root(self, block_root: SigningRoot) -> Slot:
         pass
 
+    def get_block_signing_root_by_hash_tree_root(
```
mhchia: Missing `abstractmethod`

.flake8-eth2
```diff
@@ -1,4 +1,4 @@
 [flake8]
-max-line-length= 100
+max-line-length= 120
```
mhchia: Is there specific reason to have this change?

eth2/beacon/db/chain.py
ChihChengLiang: Nice catch!

.flake8-eth2
ChihChengLiang: The file still has lines exceed 100 characters after running black. I'll remove this change and do future investigation.

PR #1087, opened 2019-09-09
Move ResponseCandidateStream to own module

pipermerriam:
### What was wrong?

I've got some work that's moving things around and refactoring.  The diff it generates is ugly because it contains both code changes and moves.

### How was it fixed?

Just do the move in this PR which makes the upcoming diff way easier to read.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

f2fe8dfe58e07f739ee8927b0db3cf46--cute-pets-cute-puppies

### DIFF
--- a/trinity/protocol/common/candidate_stream.py
+++ b/trinity/protocol/common/candidate_stream.py
@@ -1,0 +1,193 @@
+import asyncio
+import time
+from typing import (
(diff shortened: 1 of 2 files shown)


pipermerriam:
This should be zero functionality change.

PR #1088, opened 2019-09-09
Move stats utils to `p2p` module

pipermerriam:
### What was wrong?

The various statistics utils that we use for tracking peer performance are located in `trinity._utils`.

The round-trip APIs are about to move into the `p2p` module and they need access to these APIs, however, the `p2p` module shouldn't import from the `trinity._utils` module.

### How was it fixed?

Move the utilities into `p2p.stats`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

tumblr_nd5ir7cgu41s8mgkyo2_500


pipermerriam:
This will be zero functionality change.

PR #1089, opened 2019-09-10
Suppress exception during header sync

### LINKED ISSUE #1083: Beam sync went stale after error
### What is wrong?

This error came up during beam sync

```python
   ERROR  09-09 11:44:41  ETHHeaderChainSyncer  Unexpected error in <trinity.protocol.eth.sync.ETHHeaderChainSyncer object at 0x7fc3dfd1beb8>, exiting
  File "/home/ubuntu/trinity/p2p/service.py", line 118, in run
  File "/home/ubuntu/trinity/trinity/sync/common/headers.py", line 866, in _run
  File "/home/ubuntu/trinity/p2p/cancellable.py", line 20, in wait
  File "/home/ubuntu/trinity/p2p/cancellable.py", line 43, in wait_first
  File "/home/ubuntu/trinity/venv/lib/python3.6/site-packages/cancel_token/token.py", line 178, in cancellable_wait
    return done.pop().result()
  File "/home/ubuntu/trinity/trinity/sync/common/headers.py", line 880, in _build_skeleton
  File "/home/ubuntu/trinity/trinity/sync/common/headers.py", line 961, in _full_skeleton_sync
  File "/home/ubuntu/trinity/p2p/cancellable.py", line 20, in wait
  File "/home/ubuntu/trinity/p2p/cancellable.py", line 43, in wait_first
  File "/home/ubuntu/trinity/venv/lib/python3.6/site-packages/cancel_token/token.py", line 178, in cancellable_wait
    return done.pop().result()
  File "/home/ubuntu/trinity/trinity/sync/common/headers.py", line 596, in schedule_segment
  File "/home/ubuntu/trinity/p2p/cancellable.py", line 20, in wait
  File "/home/ubuntu/trinity/p2p/cancellable.py", line 43, in wait_first
  File "/home/ubuntu/trinity/venv/lib/python3.6/site-packages/cancel_token/token.py", line 178, in cancellable_wait
    return done.pop().result()
  File "/home/ubuntu/trinity/trinity/_utils/datastructures.py", line 160, in add
eth_utils.exceptions.ValidationError: Duplicate tasks detected: {(<eth.rlp.headers.BlockHeader object at 0x7fc39a1aee80>, 192, ETHPeer <Node(0xd64684e7a34c4f0aaaa07e1f6143ec91610b99b120b441387356def326890b40867a4aa4384af7fe4088858c161a05efb61499c35d194b4f46c4a0cac82f2ee2@35.193.221.102:40780)>)} are already present in the queue

```

After that, beam sync went stale but the node kept running.

### How can it be fixed

:man_shrugging:

carver:
Catch, log, and suppress duplicate ValidationError that happens from
trying to add a duplicate task. It shouldn't be happening, but it's
recoverable, so we don't need to crash the node for it.

### What was wrong?

(kind of) fixes #1083 

### How was it fixed?

Well, the error is recoverable, so we basically just log it and continue, so the node doesn't crash. It would still be good to figure out the root cause.

### To-Do

- [x] Clean up commit history
- [x] add task to find/resolve the root cause

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/common/headers.py
+++ b/trinity/sync/common/headers.py
@@ -592,9 +592,20 @@ class HeaderMeatSyncer(BaseService, PeerSubscriber, Generic[TChainPeer]):
         :param gap_length: how long is the header gap
         :param skeleton_peer: the peer that provided the parent_header - will not use to fill gaps
         """
-        await self.wait(self._filler_header_tasks.add((
-            (parent_header, gap_length, skeleton_peer),
-        )))
+        try:
+            await self.wait(self._filler_header_tasks.add((
+                (parent_header, gap_length, skeleton_peer),
+            )))
+        except ValidationError as exc:
+            self.logger.debug(
+                "Tried to re-add a duplicate list of headers to the download queue: %s",
+                exc,
+            )
+            # Since the task is already queued up, there is no value in
+            # re-adding it, so it is safe to drop the exception after logging
+            # it. One example of a time it happens is when the skeleton sync
+            # restarts, and happens to choose the same skeleton structure. It
+            # tries to reinsert the same duplicate meat filler tasks.
 
     async def _run(self) -> None:
         self.run_daemon_task(self._display_stats())
--- a/newsfragments/1083.bugfix.rst
+++ b/newsfragments/1083.bugfix.rst
@@ -1,0 +1,2 @@
+An occasional warning "ValidationError: Duplicate tasks detected" was crashing the node. It's
+recoverable, so log it, but don't crash.


trinity/sync/common/headers.py
```diff
@@ -592,9 +592,19 @@ def register_peer(self, peer: BasePeer) -> None:
...
+            )))
+        except ValidationError as exc:
+            self.logger.warning(
+                "Tried to re-add a duplicate list of headers to the download queue: %s",
+                exc,
+            )
+            # Since the task is already queued up, it will just be handled later. It's
+            # safe to drop the exception after catchi/logging it. This
+            # shouldn't ever happen, so it would be good to track down why it
+            # does, occasionally.
```
pipermerriam: Can you open an issue to track this since it seems like the kind of thing that will just get lost (and maybe add a link to the issue to this comment)

carver: After [looking at the issue longer](https://github.com/ethereum/trinity/issues/1083#issuecomment-530098078), I didn't see a clean way to always avoid duplicates, so I think I'll just downgrade this to a debug log, and rewrite the comment to give an example of when it happens.

PR #1090, opened 2019-09-10
Refactor p2p request/response API to have ABC bases.

pipermerriam:
extracted from #966 

### What was wrong?

The Request/Response API needs to be part of the p2p module. 

### How was it fixed?

This is a minor refactor that should contain only superficial functionality changes such as the removal of `classproperty` and which implements ABC base classes for the full API.  This will be followed up by moving the non-trinity parts of this API to live under `p2p.exchange`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

c405ee4ee935a7cd61e0e3b64dd49617

### DIFF
--- a/trinity/protocol/common/abc.py
+++ b/trinity/protocol/common/abc.py
@@ -38,7 +38,7 @@ if TYPE_CHECKING:
 class NormalizerAPI(ABC, Generic[TResponsePayload, TResult]):
     # This variable indicates how slow normalization is. If normalization requires
     # any non-trivial computation, consider it slow. Then, the Manager will run it in
-    # a different process.
+    # a thread to ensure it doesn't block the main loop.
     is_normalization_slow: bool
 
     @staticmethod
@@ -56,10 +56,7 @@ class ValidatorAPI(ABC, Generic[TResponse]):
         ...
 
 
-class PerformanceTrackerAPI(ABC, Generic[TRequest, TResult]):
-    """
-    The statistics of how a command is performing.
-    """
+class PerformanceAPI(ABC):
     total_msgs: int
     total_items: int
     total_timeouts: int
@@ -77,6 +74,12 @@ class PerformanceTrackerAPI(ABC, Generic[TRequest, TResult]):
         Return a human readable string representing the stats for this tracker.
         """
         ...
+
+
+class PerformanceTrackerAPI(PerformanceAPI, Generic[TRequest, TResult]):
+    """
+    The statistics of how a command is performing.
+    """
 
     @abstractmethod
     def record_timeout(self, timeout: float) -> None:
--- a/trinity/protocol/common/abc.py
+++ b/trinity/protocol/common/abc.py
@@ -1,0 +1,255 @@
+from abc import ABC, abstractmethod
+from typing import (
+    Any,
+    AsyncIterator,
+    Callable,
+    Dict,
+    Generic,
+    Iterable,
+    Optional,
+    Tuple,
+    Type,
+    TYPE_CHECKING,
+)
+
+
+from p2p.abc import (
+    AsyncioServiceAPI,
+    CommandAPI,
+    ConnectionAPI,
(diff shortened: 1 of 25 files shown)


trinity/protocol/common/abc.py
```diff
@@ -0,0 +1,255 @@
...
+from .typing import TResult, TRequest, TResponse
+
+if TYPE_CHECKING:
+    import asyncio  # noqa: F401
+
+
+class NormalizerAPI(ABC, Generic[TResponsePayload, TResult]):
+    # This variable indicates how slow normalization is. If normalization requires
+    # any non-trivial computation, consider it slow. Then, the Manager will run it in
+    # a different process.
```
carver: I think this comment became stale, and it's now a different thread, instead of process.

trinity/protocol/common/candidate_stream.py
```diff
@@ -188,6 +188,3 @@ def __del__(self) -> None:
             _, future = self.pending_request
             if future.cancel():
                 self.logger.debug("Forcefully cancelled a pending response in %s", self)
-
-    def __repr__(self) -> str:
-        return f'<ResponseCandidateStream({self._connection!s}, {self.response_msg_type!r})>'
```
carver: This repr has helped in the logs before. Do we need to drop it?

trinity/sync/common/peers.py
```diff
@@ -46,7 +45,7 @@ class WaitingPeers(Generic[TChainPeer]):
     def __init__(
             self,
             response_command_type: Union[Type[CommandAPI], Sequence[Type[CommandAPI]]],
-            sort_key: Callable[[BasePerformance], float]=_items_per_second) -> None:
+            sort_key: Callable[[PerformanceTrackerAPI[Any, Any]], float]=_items_per_second) -> None:
```
carver: This was the nice thing about having `BasePerformance` and `BasePerformanceTracker` split: in this case, the caller doesn't care about the request/response that created the stats, so we didn't have to bother with the generics.

If you're on the fence about joining them back together, I slightly like them split. But not strongly enough to hold this up...

trinity/protocol/common/exchanges.py
```diff
@@ -79,21 +46,14 @@ def __init__(self, mgr: ExchangeManager[TRequestPayload, TResponsePayload, TResu
             timeout,
         )
 
-    @classproperty
-    def response_cmd_type(cls) -> Type[CommandAPI]:
+    @classmethod
+    def get_response_cmd_type(cls) -> Type[CommandAPI]:
```
carver: The only reason I can think to like the property style is that it can have less overhead with a `cached_property` if we find that to be in a hotspot.

What was the rationale again for dropping the class property?

pipermerriam: No good way to have `classproperty` without the resulting type ending up as `Any`

trinity/protocol/common/candidate_stream.py
pipermerriam: Didn't remove, it just got moved up higher in the class definitionjust after `__init__`

trinity/sync/common/peers.py
pipermerriam: Splitting them into `PerformanceAPI` and `PerformanceTrackerAPI` to preserve this nicety.

trinity/protocol/common/exchanges.py
carver: Which type? The existing class property had a return type already. (just want to understand, in case I run into a similar issue)

pipermerriam: It's a `mypy` thing

```python
from trinity._utils.decorators import classproperty


class Foo:
    @classproperty
    def bar(cls) -> int:
        return 3


reveal_type(Foo.bar)
```

Produces...

```
arst.py:10: error: Revealed type is 'Any'
```

PR #1092, opened 2019-09-10
[interop] Make `run_beacon_nodes.py` work again

hwwhww:
### What was wrong?
For easy debugging.

### How was it fixed?
1. Fix `run_beacon_nodes.py`
2. Fix one missing `from eth2.beacon.tools.fixtures.loading import load_config_at_path` 

#### Cute Animal Picture
dolphin-3416519_640

### DIFF
--- a/eth2/beacon/scripts/run_beacon_nodes.py
+++ b/eth2/beacon/scripts/run_beacon_nodes.py
@@ -110,7 +110,7 @@ class Node:
     @property
     def cmd(self) -> str:
         keys_path = (
-            Path().absolute() / 'eth2/beacon/scripts' /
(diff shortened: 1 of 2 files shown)


eth2/beacon/scripts/run_beacon_nodes.py
```diff
@@ -106,6 +109,10 @@ def maddr(self) -> Multiaddr:
 
     @property
     def cmd(self) -> str:
+        keys_path = (
+            Path().absolute() / 'eth2/beacon/scripts' /
```
ChihChengLiang:
```suggestion
            Path().absolute() / 'eth2' / 'beacon' / 'scripts' /
```

PR #1094, opened 2019-09-10
Report correct ethstats client API version

cburgdorf:
### What was wrong?

The Trinity client is shown with a red mark on ethstats, explaining that the client needs to be updated.

image

### How was it fixed?

We were incorrectly reporting Trinitys version number where the ethstats API version number was expected.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/plugins/builtin/ethstats/ethstats_service.py
+++ b/trinity/plugins/builtin/ethstats/ethstats_service.py
@@ -11,9 +11,6 @@ from p2p.service import (
     BaseService,
 )
 
-from trinity import (
-    __version__,
-)
 from trinity.config import (
     Eth1AppConfig,
 )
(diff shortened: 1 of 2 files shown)


trinity/plugins/builtin/ethstats/ethstats_service.py
```diff
@@ -125,7 +125,7 @@ def get_node_info(self) -> EthstatsData:
             'port': self.boot_info.trinity_config.port,
             'os': platform.system(),
             'os_v': platform.release(),
-            'client': __version__,
+            'client': '0.1.1',
```
cburgdorf: Go has this hardcoded, too

https://github.com/ethereum/go-ethereum/blob/46b437f39c66d228294acb76878fea890af77949/ethstats/ethstats.go#L394

pipermerriam: why do we need to hard code it?  I don't understand.

cburgdorf: I blame it on ethstats. They expect us to tell them which version of *their* API we are using.

PR #1108, opened 2019-09-11
updated dockerfile.eth2 and whiteblock_start.sh to match requirements…

renaynay:
* changed way peers are added (Alex was right)
* changed dockerfile to comply with whiteblock scripts

### DIFF
--- a/eth2/beacon/scripts/whiteblock_start.sh
+++ b/eth2/beacon/scripts/whiteblock_start.sh
@@ -50,7 +50,8 @@ do
             IDENTITY=$VALUE
             ;;
         --peers)
-            PEERS+=",$VALUE"
+            [ ! -z "$PEERS" ] && PEERS+=","
+	    PEERS+="$VALUE"
             ;;
         --validator-keys)
             VALIDATOR_KEYS=$VALUE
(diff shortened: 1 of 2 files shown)


eth2/beacon/scripts/whiteblock_start.sh
```diff
@@ -78,6 +79,8 @@ export PYTHONWARNINGS=ignore::DeprecationWarning
 
 mkdir -p /tmp/aa
 
+cp $GEN_STATE /tmp/aa/genesis_state.ssz
```
ralexstokes: this line should be unnecessary given the command line argument on line 92... i'm happy to come help debug, ping me in dm

eth2/beacon/scripts/whiteblock_start.sh
```diff
@@ -50,7 +50,8 @@ do
             IDENTITY=$VALUE
             ;;
         --peers)
-            PEERS+=",$VALUE"
+            [ ! -z "$PEERS" ] && PEERS+=","
```
ralexstokes: yay! although this should actually be unnecessary given the change i made in https://github.com/ethereum/trinity/commit/87f2b61675d0f71949d1971cb64ddd91d02d9e21 (sorry i didn't push until now!)

renaynay: We had to change the script for the other clients because they were all bugging out trying to ingest "," as a peer. Can we impl the changes here for consistency?

ralexstokes: sure sure

PR #1116, opened 2019-09-11
Catch OperationCancelled during checkpoint fulfillment

### LINKED ISSUE #1115: Beam Sync checkpoint fetcher fails if peer quits
### What is wrong?

`OperationCancelled` was not being caught in the checkpoint sync strategy, which was causing the whole sync to exit if the peer was cancelled out while asking it for headers.

### How can it be fixed

Opening a PR shortly.

carver:
### What was wrong?

Builds on #1112 

Fixes #1115 

### How was it fixed?

Catch and continue if peer is cancelled while we ask it for the checkpoint header.

Added some logs that helped identify the problem.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


carver:
@cburgdorf only the last two commits are relevant for review, the others are in another PR

trinity/sync/beam/chain.py
```diff
@@ -403,14 +403,6 @@ def __init__(self,
         # run sync until cancelled
         await self.cancellation()
 
-    async def _persist_header_chain(self, headers: Tuple[BlockHeader, ...]) -> None:
-        await self.wait(
-            self._db.coro_persist_header_chain(
-                headers,
-                self._launch_strategy.get_genesis_parent_hash(),
```
carver: Any idea what would be a problem with removing this line setting the genesis parent hash to the checkpoint hash? It's definitely problematic from py-evm's point of view, because if the parent hashes match, then py-evm forces the starting score back to 0. (which is why the headers were non-canonical)

carver: Nevermind, presumably something to do with this:

```
   DEBUG  09-11 16:47:03     HeaderOnlyPersist  Task failure traceback
  File "/home/jcarver/code/trinity/trinity/db/manager.py", line 488, in __getitem__
    value = self._track_diff[key]
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/eth/db/diff.py", line 80, in __getitem__
    raise DiffMissingError(key, result)  # type: ignore # ignore over cast for perf reasons
eth.db.diff.DiffMissingError: (b'block-hash-to-score:-62\x9f\x1b\x14\xa7F\x06\xc0\x8d\x94\x8e\x14_\x87\xe3a\xc4_[p\xd7\xf4n\x94\xea\xc3v\xe1m\x03', 'Key is missing because it was never inserted')

  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/eth/db/header.py", line 134, in _get_score
    encoded_score = db[SchemaV1.make_block_hash_to_score_lookup_key(block_hash)]
  File "/home/jcarver/code/trinity/trinity/db/manager.py", line 493, in __getitem__
    return self._db[key]
  File "/home/jcarver/code/trinity/trinity/db/manager.py", line 384, in __getitem__
    raise KeyError(key)
KeyError: b'block-hash-to-score:-62\x9f\x1b\x14\xa7F\x06\xc0\x8d\x94\x8e\x14_\x87\xe3a\xc4_[p\xd7\xf4n\x94\xea\xc3v\xe1m\x03'

  File "/home/jcarver/code/trinity/p2p/service.py", line 159, in _run_task_wrapper
    await awaitable
  File "/home/jcarver/code/trinity/p2p/service.py", line 178, in _run_daemon_task_wrapper
    await awaitable
  File "/home/jcarver/code/trinity/trinity/sync/beam/chain.py", line 410, in _persist_headers
    exited = await self._exit_if_checkpoint(headers)
  File "/home/jcarver/code/trinity/trinity/sync/beam/chain.py", line 484, in _exit_if_checkpoint
    self._db.coro_persist_header_chain(persist_headers)
  File "/home/jcarver/code/trinity/p2p/cancellable.py", line 20, in wait
  File "/home/jcarver/code/trinity/p2p/cancellable.py", line 43, in wait_first
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/cancel_token/token.py", line 178, in cancellable_wait
    return done.pop().result()
  File "/home/jcarver/code/trinity/trinity/_utils/async_dispatch.py", line 23, in wrapper
    *args
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/eth/db/header.py", line 167, in persist_header_chain
    return self._persist_header_chain(db, headers, genesis_parent_hash)
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/eth/db/header.py", line 231, in _persist_header_chain
    score = cls._get_score(db, first_header.parent_hash)
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/eth/db/header.py", line 137, in _get_score
    encode_hex(block_hash)))
eth.exceptions.HeaderNotFound: No header with hash 0x2d36329f1b14a74606c08d948e145f87e361c45f5b70d7f46e94eac376e16d03 found
```

My day is about to wrap up, so I'm not sure I'll be able to follow up until tomorrow morning.

carver: Actually, I think this is an unrelated issue. First guess from scanning it:
- run beam sync
- preview blocks
  - store header rlp
  - preview execution might get ancestor block hashes, which are retrievable using ^
- close out of beam sync before the previewed headers are imported
- ...
- start beam sync again
- header syncer skips over the headers that are persisted, thinking they've already been imported
- the next header import tries to get the score of the parent
- that score isn't available, because the header was never persisted
- `HeaderNotFound` crash

I can't come up with a good reason why I haven't seen this before, though. It seems like the kind of thing that would happen often, if I understand the series of steps above correctly.

carver: Ah, perhaps because occasionally, the header syncer can intentionally return a header that is known to already exist:

It might return the checkpoint header this way, and cause problems when trying to look up the parent. To me, a reasonable option is to log and drop when the header syncer produces a header that's older than or equal to the checkpoint.

carver:
I'll take a look at the test failure in the morning.

cburgdorf:
>I'll take a look at the test failure in the morning.

Apparently, `beam_to_block=1` fails, all blocks coming after the first one are working fine.

PR #1122, opened 2019-09-12
Add sync service to plugin

ChihChengLiang:
### What was wrong?

BeaconChainSyncer is not in Beacon plugin 

### How was it fixed?

Add BeaconChainSyncer to Beacon plugin 

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/bcc_libp2p/node.py
+++ b/trinity/protocol/bcc_libp2p/node.py
@@ -339,6 +339,7 @@ class Node(BaseService):
                 addrs=[make_tcp_ip_maddr(ip, port)],
             )
         )
+        await self.say_hello(peer_id)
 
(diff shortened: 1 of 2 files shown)


trinity/protocol/bcc_libp2p/node.py
```diff
@@ -603,6 +604,7 @@ def _compare_chain_tip_and_finalized_epoch(self,
         await stream.close()
 
     async def say_hello(self, peer_id: ID) -> None:
+        self.logger.info("Say hello")
```
hwwhww:
```suggestion
        self.logger.info("Say hello to %s", str(peer_id))
```

ChihChengLiang:
Opened another PR #1125 that changes the base branch to master and addresses the PR feedbacks. I also removed the line `await self.say_hello(peer_id)`, the part has already been addressed in #1120 well.

PR #1125, opened 2019-09-13
add sync service to plugin

ChihChengLiang:
### What was wrong?

BeaconChainSyncer is not in Beacon plugin 

### How was it fixed?

Add BeaconChainSyncer to Beacon plugin 

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/db/chain.py
+++ b/eth2/beacon/db/chain.py
@@ -201,7 +201,7 @@ class BeaconChainDB(BaseBeaconChainDB):
         """
         with self.db.atomic_batch() as db:
             if block.is_genesis:
-                self._handle_exceptional_justification_and_finality(db, block)
+                self._handle_exceptional_justification_and_finality(block)
 
             return self._persist_block(db, block, block_class, fork_choice_scoring)
 
@@ -805,7 +805,7 @@ class BeaconChainDB(BaseBeaconChainDB):
             self._update_justified_head(*result)
 
     def _handle_exceptional_justification_and_finality(
-        self, db: DatabaseAPI, genesis_block: BaseBeaconBlock
+        self, genesis_block: BaseBeaconBlock
     ) -> None:
         """
         The genesis ``BeaconState`` lacks the correct justification and finality
--- a/eth2/beacon/scripts/run_beacon_nodes.py
+++ b/eth2/beacon/scripts/run_beacon_nodes.py
@@ -188,10 +188,10 @@ class Node:
         async for line_bytes in stream_reader:
             line = line_bytes.decode("utf-8").replace("\n", "")
             # TODO: Preprocessing
-            self._record_happenning_logs(from_stream, line)
+            self._record_happening_logs(from_stream, line)
             print(f"{self.logging_name}\t{line}")
 
-    def _record_happenning_logs(self, from_stream: str, line: str) -> None:
+    def _record_happening_logs(self, from_stream: str, line: str) -> None:
         for log in self.logs_expected[from_stream]:
             if log.pattern in line:
                 self.logger.debug('log "log.name" occurred in %s', from_stream)
--- a/eth2/beacon/scripts/run_beacon_nodes.py
+++ b/eth2/beacon/scripts/run_beacon_nodes.py
@@ -189,7 +189,7 @@ class Node:
             line = line_bytes.decode("utf-8").replace("\n", "")
             # TODO: Preprocessing
             self._record_happenning_logs(from_stream, line)
-            print(f"{self.logging_name}.{from_stream}\t: {line}")
+            print(f"{self.logging_name}\t{line}")
 
     def _record_happenning_logs(self, from_stream: str, line: str) -> None:
         for log in self.logs_expected[from_stream]:
@@ -233,6 +233,7 @@ async def main():
     )
 
     asyncio.ensure_future(node_alice.run())
+    await asyncio.sleep(30)
     asyncio.ensure_future(node_bob.run())
 
     await asyncio.sleep(1000000)
--- a/trinity/protocol/bcc_libp2p/servers.py
+++ b/trinity/protocol/bcc_libp2p/servers.py
@@ -281,7 +281,7 @@ class BCCReceiveServer(BaseService):
             # Successfully imported the block. See if any blocks in `self.orphan_block_pool`
             # depend on it. If there are, try to import them.
             # TODO: should be done asynchronously?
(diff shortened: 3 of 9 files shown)


trinity/sync/beacon/chain.py
```diff
@@ -208,11 +208,18 @@ def is_sync_peer_selected(self) -> bool:
                 "genesis block"
```
hwwhww: suggestion: remove `parent_slot = batch[0].slot - 1`
and change it to:
```
        if batch[0].slot <= 0:
            raise Exception(
                "Invariant: Syncing starts with the child of a finalized block, so never with the "
                "genesis block"
```

ralexstokes:
@ChihChengLiang what is the status here? is this just blocked on review before we can merge it in?

ChihChengLiang:
@ralexstokes Feel free to give this a review. I currently test this with 
```
python eth2/beacon/scripts/run_beacon_nodes.py
```
Peers generally sync well, but the delayed peer has trouble in proposing a block.

ralexstokes:
@ChihChengLiang what kind of trouble? until i have synced the head, i'm not going to know if i need to propose or not so it would make sense to hold off on this

ralexstokes:
@ChihChengLiang running it now -- i like how easy it is to get two nodes up to get one syncing :)

i see each node boot and run for some time -- bob seems to be running while syncing alice and it is throwing a bunch of errors -- we will need to dive into specifics but essentially bob should ignore any requests to participate as a gossiping node while syncing in my opinion

then we get some errors w/ alice and eventually both nodes stop w/ validation errors

some of this has been fixed in the skeleton lake branch so i'm inclined to rebase this once we merge in the other stuff to test it out

ChihChengLiang:
Rebased, but the `run_beacon_nodes` seemed broken. Have to investigate this tomorrow.

ralexstokes:
i'll take a look later today

ChihChengLiang:
This now depends on #1200. Bob syncs, imports blocks, but fails to validate blocks and attestations, and forks after that.

trinity/protocol/bcc_libp2p/servers.py
```diff
@@ -281,7 +281,7 @@ def _process_received_block(self, block: BaseBeaconBlock) -> None:
             # Successfully imported the block. See if any blocks in `self.orphan_block_pool`
             # depend on it. If there are, try to import them.
             # TODO: should be done asynchronously?
-            self._try_import_orphan_blocks(block.signing_root)
+            self._try_import_orphan_blocks(block.parent_root)
```
ChihChengLiang: This is the critical one that fixes the PR. The orphan block was not correctly imported before this fix.

trinity/protocol/bcc_libp2p/topic_validators.py
```diff
@@ -51,8 +55,8 @@ def beacon_block_validator(msg_forwarder: ID, msg: rpc_pb2.Message) -> bool:
             future_slot=block.slot,
         )
         try:
-            process_block_header(state, block, state_machine.config, True)
-        except (ValidationError, SignatureError) as error:
+            validate_proposer_signature(state, block, CommitteeConfig(state_machine.config))
```
ChihChengLiang: When Trinity receives a block, validate the proposer signature but not the state transition.
Consider this case:

1. Alice operates normally since genesis but Bob starts at slot 3 and want to sync from Alice.
2. From the handshake, Bob learned the slot 3 head block of Alice.
3. But before Bob starts to sync, Alice already broadcasted the slot 4 block `0xabcd` to Bob.
4. Bob has no parent of `0xabcd` but Bob still needs to put this block at orphan pool.
5. So that when Alice broadcast block at slot 5, Bob has its parent `0xabcd` and syncs.

ralexstokes: i think we do want this validation to just be "is the signature valid"... i know there is some discussion among the eth2 client devs as to what is the proper check here...

i do want to better understand the scenario you gave here though -- the issue is that the state transition fails because the state doesn't have the correct parent data in it?

trinity/sync/beacon/chain.py
```diff
@@ -197,22 +197,30 @@ def is_sync_peer_selected(self) -> bool:
             yield batch
 
             slot = batch[-1].slot + 1
+            if slot > self.sync_peer.head_slot:
+                break
 
     async def validate_first_batch(self, batch: Tuple[BaseBeaconBlock, ...]) -> None:
-        parent_root = batch[0].parent_root
-        parent_slot = batch[0].slot - 1
+        first_block = batch[0]
```
ralexstokes: will `batch` always have a first element here?

trinity/sync/beacon/chain.py
```diff
@@ -197,22 +197,30 @@ def is_sync_peer_selected(self) -> bool:
...
 
-        if parent_slot < 0:
+        if first_block.slot <= 0:
```
ralexstokes: given that we don't have signed types in `ssz`, seems like it would be a bug in our code if the `slot` is negative, right?

so we can just check that we aren't syncing the genesis block, unless there is something else i'm missing

trinity/protocol/bcc_libp2p/topic_validators.py
ChihChengLiang: > because the state doesn't have the correct parent data in it

Exactly. You can pay attention to the block#4. If we are validating parent root and Bob did not keep block#4 in the orphan pool, Bob will also drop the block#5 that Alice just produced.

eth2/beacon/db/chain.py
```diff
@@ -201,7 +201,7 @@ def persist_block(
         """
         with self.db.atomic_batch() as db:
             if block.is_genesis:
-                self._handle_exceptional_justification_and_finality(db, block)
+                self._handle_exceptional_justification_and_finality(block)
```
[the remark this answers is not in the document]
ChihChengLiang: Ah, I get your point. I'll keep this as is, let's fix this in the next PR.

eth2/beacon/scripts/run_beacon_nodes.py
```diff
@@ -233,6 +233,7 @@ def sigint_handler(sig, frame):
     )
 
     asyncio.ensure_future(node_alice.run())
+    await asyncio.sleep(30)
```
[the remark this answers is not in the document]
ChihChengLiang: It would be totally fine. I'm blocking 30 seconds here to make Bob behind Alice for 3 blocks. So we can observe the sync behavior.

PR #1129, opened 2019-09-13
Fix is_valid_connection_candidate

pipermerriam:
### What was wrong?

The #1054 broke the `PeerPool.is_valid_connection_candidate` function since the `connected_peers` keys are no longer `NodeAPI` objects.

### How was it fixed?

Added the right path to get to the address.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

0d6a2236f3d8fec39bc041095e1985be--animals-in-costumes-cats-in-costumes


carver:
This one deserves a release note, since bug went out with the last release and it's pretty visible and appears to prevent inbound connections from succeeding.

pipermerriam:
https://github.com/ethereum/trinity/pull/1179

Issue #1127, opened 2019-09-13
Rolling bloom filters for tx pool


pipermerriam:
### What is wrong?

The transaction pool uses a bloom filter to ensure we don't re-send transactions to peers that they've already sent us.  This bloom filter will eventually fill up, returning more and more false positives.

### How can it be fixed

I think it should be simple to implement a rolling bloom filter data structure.  It houses a single *active* bloom filter and tracks number of entries that the bloom filter contains.  The active bloom filter is deemed *full* when the number of entries exceeds the point where false positive rates will go above some threshold (this is a statistical property of bloom filters).

Once the active bloom filter is full, it is put into a backlog queue.  If the backlog queue is full, the oldest entry is dropped.

- Inclusion checks are performed across both the active and the old bloom filters.
- All entries are added to the active bloom filter (even if they are found present in one of the old bloom filters).

PR #1131, opened 2019-09-13
Use a rolling bloom filter for transaction pool.

pipermerriam:
### What was wrong?

The transaction pool uses a single large bloom filter to determine whether we've already broadcast a given transaction to a given peer.  Eventually this bloom filter will fill up, returning more and more false positives.

### How was it fixed?

Use a *rolling* bloom filter.  We use many smaller bloom filters.  Each time one "fills up" we rotate a new one into place, holding onto the recent history of the old ones.  This should allow the transaction pool to operate indefinitely while retaining reasonably accurate information about what transactions have been sent to which peers.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

181838-harry-potter-costume

### DIFF
--- a/tests/core/utils/test_rolling_bloom_filter.py
+++ b/tests/core/utils/test_rolling_bloom_filter.py
@@ -1,0 +1,31 @@
+from trinity._utils.bloom import RollingBloom
+
+
+def test_rolling_bloom_maintains_history():
+    # only one historical filter
+    bloom = RollingBloom(generation_size=10, max_generations=2)
+
+    # fill up the main filter and the history
+    for i in range(20):
+        value = bytes((i,))
+        bloom.add(value)
+        assert value in bloom
+
+    for i in range(20):
+        value = bytes((i,))
+        assert value in bloom
+
+    # this should eject all of the 0-9 values
+    bloom.add(b'\xff')
+
+    # this must be done probabalistically since bloom filters have false
+    # positives.  At least one of the 0-9 values should be gone.
+    assert any(
+        bytes((value,)) not in bloom
+        for value in range(10)
+    )
+
+    for i in range(10, 20):
+        value = bytes((i,))
+        assert value in bloom
+    assert b'\xff' in bloom
--- a/tests/core/utils/test_rolling_bloom_filter.py
+++ b/tests/core/utils/test_rolling_bloom_filter.py
@@ -5,27 +5,28 @@ def test_rolling_bloom_maintains_history():
     # only one historical filter
     bloom = RollingBloom(generation_size=10, max_generations=2)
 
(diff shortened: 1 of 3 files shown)


pipermerriam:
related to #1127

tests/core/utils/test_rolling_bloom_filter.py
```diff
@@ -0,0 +1,31 @@
...
+    # only one historical filter
+    bloom = RollingBloom(generation_size=10, max_generations=2)
+
+    # fill up the main filter and the history
+    for i in range(20):
+        value = bytes((i,))
+        bloom.add(value)
+        assert value in bloom
+
+    for i in range(20):
```
cburgdorf: I don't understand why we are going over the 20 entries again. We already asserted that they exist right after we inserted them (line 12). There's nothing that could have changed in between.

trinity/_utils/bloom.py
```diff
@@ -0,0 +1,40 @@
...
+        if max_generations < 2:
+            raise ValueError(f"max_generations must be 2 or more: got {max_generations}")
+        self._max_bloom_elements = generation_size
+        self._max_generations = max_generations
+        self._history: Deque[BloomFilter] = collections.deque()
+        self._active = BloomFilter(max_elements=self._max_bloom_elements)
+        self._items_in_active = 0
+
+    def add(self, key: bytes) -> None:
+        # before adding check to see if we need to cycle the push the active
```
cburgdorf: >cycle ~~the~~ **to** push

trinity/components/builtin/tx_pool/pool.py
```diff
@@ -54,10 +51,27 @@ def __init__(self,
...
+        #
+        # We want our BF to remain effective for at least 24 hours -> 1440 min -> 144 generations
+        #
+        # Memory size can be computed as:
+        #
+        # bits_per_bloom = (-1 * generation_size * log(0.1)) / (log(2) ** 2) -> 479252
+        # kbytes_per_bloom = bits_per_bloom / 8 / 1024 -> 58
+        # kbytes_total = max_generations * kbytes_per_bloom -> 8424
+        #
+        # We can expect the maximum memory footprint to be about 8.5mb for the bloom filters.
```
cburgdorf: I'm jealous of your fancy math :sweat_smile:

tests/core/utils/test_rolling_bloom_filter.py
pipermerriam: Previously we assert as we insert them.  However, because the filter is "rolling" we now need to check over them again to ensure that none have been ejected.

cburgdorf: Ah, right :+1:

PR #1135, opened 2019-09-13
Enable the transaction pool by default

### LINKED ISSUE #1132: Turn on transaction pool by default
### What is wrong?

Transaction pool isn't on by default.  Part of being a good network citizen is relaying transactions...

### How can it be fixed

Turn it on by default.

pipermerriam:
fixes #1132 

### What was wrong?

The transaction pool is off by default.  A good network citizen should relay transactions.

### How was it fixed?

Turned it on by default.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/api/api.cli.rst
+++ b/docs/api/api.cli.rst
@@ -7,129 +7,145 @@ We can also generate an always up-to-date version of them by running ``trinity -
 .. code-block:: shell
 
     usage: trinity [-h] [--version] [--trinity-root-dir TRINITY_ROOT_DIR]
-                [--port PORT] [-l LEVEL] [--stderr-log-level STDERR_LOG_LEVEL]
-                [--file-log-level FILE_LOG_LEVEL]
-                [--network-id NETWORK_ID | --ropsten]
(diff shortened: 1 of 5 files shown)


trinity/plugins/builtin/tx_pool/plugin.py
```diff
@@ -47,26 +46,26 @@ def name(self) -> str:
...
-
-        if is_enabled and not unsupported:
+        is_disable = self.boot_info.args.disable_tx_pool
+        is_supported = not light_mode
+        is_enabled = not is_disable and is_supported
+
+        if is_disable:
+            self.logger.debug("Transaction pool disabled")
+        elif not is_supported:
+            self.logger.debug("Transaction pool does not support light mode")
```
cburgdorf: Previously, we would shut down Trinity on this invalid configuration whereas now we just log a `debug` line. If we don't want to shut down Trinity maybe we should at least make this a warning so that the user really notices that it is running without tx pool.

PR #1140, opened 2019-09-16
Rename plugins to components

cburgdorf:
### What was wrong?

We plan to make various functional changes to our plugin API as well as renaming them to components.

Since we've already started to call them components, we often find ourselves writing "Plugins / Components" when we talk about them.

I think we can already rename plugins to components to avoid confusion.

### How was it fixed?

Renamed everything from `plugin` to `component`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

put a cute animal picture link inside the parentheses/cdn.vox-cdn.com/uploads/chorus_image/image/45750688/479508329.0.jpg)


cburgdorf:
@pipermerriam I know this touches a lot of code but it's all just renames (zero change in functionality) and it respects casing as well as singular / plural.

PR #1142, opened 2019-09-16
Add behavior api

pipermerriam:
extracted from #966 and #1149

builds on #1148 #1145 #1147 #1144 

### What was wrong?

In order to get rid of the Peer classes, we need a way to define units of functionality that are "conditionally" applied to a `ConnectionAPI`.

### How was it fixed?

This introduces the "Behavior" API.  A `BehaviorAPI` is a unit of functionality that applies to a `ConnectionAPI`.

One of the simplest behaviors is the `p2p.behaviors.CommandHandler`.  This behavior installs a handler function via `ConnectionAPI.add_command_handler`.  Here's an example that uses this to implement a "send a pong anytime I get a ping"

```python
class PongWhenPinged(CommandHandler):
    cmd_type = Ping
    async def handle(self, connection, msg):
        connection.get_base_protocol().send_pong()

# or if you like decorators and functions
@command_handler(Ping)
async def send_pong_when_pinged(connection, msg):
    connection.get_base_protocol().send_pong()
```

Here's one that triggers the `ConnectionAPI.cancel_token` when it receives a `Disconnect` msg.

```python
class TriggerCancelOnDisconnect(CommandHandler):
    cmd_type = Disconnect
    disconnect_reason = None
    async def handle(self, connection, msg):
        self.disconnect_reason = DisconnectReason(msg['reason'])
        await connection.cancel()
```

Another built-in helper is the `p2p.behaviors.Application`.  These are designed to be *containers* of multiple behaviors, typically exposing a public API of some sort.

```python
class P2PAPI(Application):
    def get_behaviors():
        return (PongWhenPonged(), TriggerCancelOnDisconnect)
    def disconnect(self, reason):
        ...
    def send_ping(self):
        ...
```

The class above combines usage of both some custom local functions as well as wrapping up some other behaviors.

For a look in how these APIs are actually used, see https://github.com/ethereum/trinity/pull/1149

Here's a more polished version with only the `p2p` APIs: https://github.com/ethereum/trinity/pull/1150

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

f8cb3d88be23b45e54e646d7b17e9d79

### DIFF
--- a/p2p/abc.py
+++ b/p2p/abc.py
@@ -440,6 +440,32 @@ ProtocolHandlerFn = Callable[['ConnectionAPI', CommandAPI, Payload], Awaitable[A
 CommandHandlerFn = Callable[['ConnectionAPI', Payload], Awaitable[Any]]
 
 
+class BehaviorAPI(ABC):
+    """
+    A behavior is some unit of logic that applies to a `ConnectionAPI`.
+    Typical behaviors may perform an action when a certain command is received
+    such as responding to a `Ping` with a `Pong` message.
+    """
+    @abstractmethod
+    def applies_to(self, connection: 'ConnectionAPI') -> bool:
+        """
+        Return `True` if this Behavior should be applied to the `connection``
+        """
+        ...
+
+    @abstractmethod
+    def apply(self, connection: 'ConnectionAPI') -> AsyncContextManager[None]:
+        """
+        Context manager API used programatically by the `ContextManager` to
+        apply the behavior to the connection during the lifecycle of the
+        connection.
+        """
+        ...
+
+
+TBehavior = TypeVar('TBehavior', bound=BehaviorAPI)
+
+
 class ConnectionAPI(AsyncioServiceAPI):
     protocol_receipts: Tuple[HandshakeReceiptAPI, ...]
 
@@ -487,6 +513,21 @@ class ConnectionAPI(AsyncioServiceAPI):
                             command_type: Type[CommandAPI],
                             handler_fn: CommandHandlerFn,
                             ) -> SubscriptionAPI:
+        ...
+
+    #
+    # Behavior API
+    #
+    def add_api(self, name: str, behavior: BehaviorAPI) -> SubscriptionAPI:
+        ...
+
+    def remove_api(self, name: str) -> None:
+        ...
+
+    def has_api(self, name: str) -> bool:
+        ...
+
+    def get_api(self, name: str, behavior_type: Type[TBehavior]) -> TBehavior:
         ...
 
     #
--- a/p2p/abc.py
+++ b/p2p/abc.py
@@ -518,16 +518,16 @@ class ConnectionAPI(AsyncioServiceAPI):
     #
     # Behavior API
     #
-    def add_api(self, name: str, behavior: BehaviorAPI) -> SubscriptionAPI:
+    def add_behavior(self, name: str, behavior: BehaviorAPI) -> SubscriptionAPI:
         ...
 
-    def remove_api(self, name: str) -> None:
+    def remove_behavior(self, name: str) -> None:
         ...
 
-    def has_api(self, name: str) -> bool:
+    def has_behavior(self, name: str) -> bool:
         ...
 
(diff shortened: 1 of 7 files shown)


pipermerriam:
@lithp this builds on multiple other PRs so feel free to either only review the latest commit or wait for the other PRs to be merged before looking this over.

pipermerriam:
One thing that I've considered changing...

Right now you have to have some long-running code somewhere that does this.

```python
my_behavior = MyBehavior
async with my_behavior.apply(connection):
    ...  # wait for exit condition
```

The "wait for exit condition" looks like it will almost always be "wait for cancellation of the connection".  Under this paradigm, it might make sense to have a simpler API like:

```python
my_behavior = MyBehavior()
connection.add_behavior(my_behavior)
```

I don't think this needs to be addressed in this PR but it's something I've considered.  A case for keeping it as-is would be that the current API is going to be friendly towards `trio` where-as the `Connection.add_behavior` won't since it will have to run something in the background (which doens't make it impossible, just against the basic ethos of `trio`).

p2p/abc.py
```diff
@@ -489,6 +515,21 @@ def add_command_handler(self,
                             ) -> SubscriptionAPI:
         ...
 
+    #
+    # Behavior API
+    #
+    def add_api(self, name: str, behavior: BehaviorAPI) -> SubscriptionAPI:
```
lithp: There are enough `API`s in `trinity.abc` that this might become confusing. Methods like `add_behavior`, `remove_behavior`, might be more clear.

pipermerriam: If you look in #966 you can see that the `PoolManager` has methods like that (at least it has some kind of `add_behavior`).  I was kind of thinking about these as being conceptually different from those and so I went with a different name.  The idea was that an "api" unit is always one or more behaviors but each behavior may not in-of-itself justify being thought of an an "api".  This distinction may be un-necessary (I'm making this all up as I go...).

I'll convert these to be `add_behavior` for now.

p2p/exceptions.py
```diff
@@ -197,3 +197,17 @@ class ReceiptNotFound(BaseP2PError):
...
+class DuplicateAPI(BaseP2PError):
+    """
+    Raised when trying to add an API to a connection under an existing key
+    """
+    pass
+
+
+class UnknownAPI(BaseP2PError):
+    """
+    Raised when trying to retrieve an API from a connection that.
```
lithp: Maybe `Raised when trying to retrieve an unknown API from a connection`

p2p/connection.py
```diff
@@ -188,6 +196,36 @@ def add_command_handler(self,
                 f"protocols: {self._multiplexer.get_protocols()}"
             )
 
+    #
+    # API extension
+    #
+    def add_api(self, name: str, behavior: BehaviorAPI) -> SubscriptionAPI:
```
lithp: Do you expect that we'll have duplicates, the same `BehaviorAPI` used twice? A fair amount of the complication in `get_api` comes from the fact that behaviors have a composite primary key, this interface would be a lot simpler if behaviors could be referenced solely by their type!

pipermerriam:
> might be worth a quick call to talk about what your eventual plan is

:+1: sounds like a good plan.  I'm available anytime after 9am (my time) tomorrow.

p2p/behaviors.py
```diff
@@ -0,0 +1,152 @@
...
+
+        class Disconnect(ConnectionBehavior):
+            def applies_to(self, connection):
+                return True
+
+            def __call__(self, reason: DisconnectReason):
+                self._connection.get_base_protocol().send_disconnect(reason)
+    """
+    @asynccontextmanager
+    async def apply(self, connection: ConnectionAPI) -> AsyncIterator[None]:
```
lithp: It might be confusing for `_connection` never to be cleared. I can definitely imagine myself writing something like:

```python
while True:
  fastest_node = pool.get_fastest()
  with beam_syncer.apply(fastest_node):
    ...
```

lithp: (Same thing with `ConnectionBehavior`)

PR #1144, opened 2019-09-16
Fix typing for cached_property and expand usage.

pipermerriam:
### What was wrong?

The `cached_property` decorator was causing functions to return `Any` types and it wasn't being used as broadly as it could be.

### How was it fixed?

Added a stub-file to help mypy retain type information when `cached_property` is used and apply it to a few more `@property` methods.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

ss-141008-animal-costumes-shrek today-ss-slide-desktop

### DIFF
--- a/p2p/connection.py
+++ b/p2p/connection.py
@@ -73,11 +73,11 @@ class Connection(ConnectionAPI, BaseService):
     #
     # Primary properties of the connection
     #
-    @property
+    @cached_property
     def is_dial_in(self) -> bool:
         return not self.is_dial_out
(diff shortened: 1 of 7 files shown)


pipermerriam:
I agree that the caching is relatively small gains.  Looking back over everything that this PR adds it to I'm not concerned with the cache invalidation issue.  None of this stuff **should** change and our somewhat hard stance on mutation should act as an extra layer of protection to the bugs that could arise.

PR #1145, opened 2019-09-16
add get_protocol_for_command_type

### LINKED ISSUE #1041: Extract logic to new Connection API
### What is wrong?

https://github.com/ethereum/trinity/blob/ce1f5ae3bfc28a45dc872c8e6f5fa068ee75094d/trinity/protocol/common/handlers.py#L48-L67

This logic from the `BaseExchangeHandler` probably belongs in the `Connection` or `Multiplexer` object.

### How can it be fixed

I'm inclined to put it on the `Connection` as `get_protocol_for_command_type`

This API should be able to be reasonably easily tested against the various code paths.

pipermerriam:
fixes #1041

### What was wrong?

Given a command type it was non-trivial to get the appropriate protocol from the `ConnectionAPI` or `MultiplexerAPI`.  Similarly, there are a few APIs on the multiplexer that would be useful if mirrored on the `Connection`.

### How was it fixed?

Added `MultiplexerAPI.get_protocol_for_command_type`.

Mirrored: `get_protocols` and `get_protocol_by_type`, and `get_protocol_for_command_type` on the `ConnectionAPI`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

facf6d4122c8617a7c698ab5f04fbdcd--dog-costumes-costume-ideas

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -123,7 +123,6 @@ class BasePeer(BaseService):
 
         # Connection instance
         self.connection = connection
-        self.multiplexer = connection.get_multiplexer()
 
         self.base_protocol = self.connection.get_base_protocol()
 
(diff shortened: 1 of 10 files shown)


pipermerriam:
@carver I'll ping you when this settles down, still fixing tests.

pipermerriam:
@carver this is green and ready for review.

p2p/peer.py
```diff
@@ -222,7 +221,7 @@ def get_boot_manager(self) -> BasePeerBootManager:
 
     @property
     def received_msgs_count(self) -> int:
-        return self.multiplexer.get_total_msg_count()
+        return self.connection.get_multiplexer().get_total_msg_count()
```
carver: Why is this the one thing that doesn't make sense to pass through like the others? I don't really have a reason that it should be a pass-through. It just stands out now as the only one.

pipermerriam: Didn't create the passthrough for this because I see this API going away and being replaced by a "behavior".

PR #1148, opened 2019-09-16
Add connection.get receipt by type

pipermerriam:
extracted from #966 

builds on https://github.com/ethereum/trinity/pull/1145

### What was wrong?

Need a generic API for retrieving handshake receipts from the `ConnectionAPI`

### How was it fixed?

Added `Connection.get_receipt_by_type`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

18d18485f548f5d3b750daac8e7ae09a

### DIFF
--- a/trinity/protocol/bcc/peer.py
+++ b/trinity/protocol/bcc/peer.py
@@ -1,4 +1,4 @@
-from typing import Sequence, Tuple
+from typing import Tuple
 
 from lahja import EndpointAPI
 
@@ -10,8 +10,7 @@ from lahja import (
     BroadcastConfig,
 )
 
-from p2p.abc import CommandAPI, HandshakeReceiptAPI, SessionAPI
-from p2p.handshake import DevP2PReceipt
+from p2p.abc import CommandAPI, SessionAPI
 from p2p.peer import (
     BasePeer,
     BasePeerFactory,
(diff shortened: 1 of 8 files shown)


trinity/protocol/bcc/peer.py
```diff
@@ -76,20 +75,11 @@ class BCCPeer(BasePeer):
...
-            if isinstance(receipt, BCCHandshakeReceipt):
-                self.head_slot = receipt.handshake_params.head_slot
-                self.genesis_root = receipt.handshake_params.genesis_root
-                self.network_id = receipt.handshake_params.network_id
-                break
-        else:
-            raise Exception(
-                "Did not find an `BCCHandshakeReceipt` in {protocol_receipts}"
-            )
+    def process_handshake_receipts(self) -> None:
```
lithp: Now that there's a clean `self.connection.get_receipt_by_type()` it might be even cleaner for `process_handshake_receipts()` to no longer exist, and for subclasses to set all these fields on themselves from inside `__init__`, after they've called `super().__init__`.

pipermerriam: Agreed, however, I'm going to get rid of it in a subsequent PR.  Trying to avoid shuffling functionality around inside the `Peer` classes and focus on pure extraction when possible.  If you look at #1150  and #1156 I think they leave the peer in a state where this can be deleted because everything it's used for is now under one of the new mechanisms that is fully detached from the `Peer` class architecture.

It's like surgery at this point.  I'm at the stage where I can carve off little pieces easier and easier!!!!

PR #1152, opened 2019-09-17
Add more RPC commands for debugging

hwwhww:
### What was wrong?

Add some non-standardized command for debugging.

### How was it fixed?

Add
- beacon_getFinalizedHead()
- beacon_getCanonicalBlockBySlot(slot: Slot)
- beacon_getBlockByRoot(root: SigningRoot)
- beacon_getGenesisBlockRoot()

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

dolphin

### DIFF
--- a/trinity/rpc/modules/beacon.py
+++ b/trinity/rpc/modules/beacon.py
@@ -1,10 +1,19 @@
 from typing import Any, Dict
 
 from eth_utils import (
+    decode_hex,
     encode_hex,
+)
+from ssz.tools import (
+    to_formatted_dict,
 )
 
 from eth2.beacon.types.blocks import BeaconBlock
+from eth2.beacon.typing import SigningRoot, Slot
+from trinity.rpc.format import (
+    format_params,
+    to_int_if_hex,
+)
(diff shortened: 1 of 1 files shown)


trinity/rpc/modules/beacon.py
```diff
@@ -20,3 +29,36 @@ class Beacon(BeaconChainRPCModule):
...
         )
+
+    #
+    # Debug
+    #
+    async def getFinalizedHead(self) -> Dict[Any, Any]:
+        """
+        Return finalized head block.
+        """
+        block = await self.chain.coro_get_finalized_head(BeaconBlock)
```
ChihChengLiang: The variable `self.chain` is actually a `chain_db` object here, we'll need to add `AsyncBeaconChainAPI` sooner or later.

hwwhww: As we discussed on discord and offline yesterday, before `AsyncChainAPI` pattern is solid, the workaround is reimplementing `BeaconChainRPCModule` so that it doesn't need to share the same interface with eth1.

PR #1154, opened 2019-09-17
Fix the way to get `state` in beacon block topic validator

NIC619:
### What was wrong?
L45: `state = chain.get_state_by_slot(block.slot - 1)` will raise error if the state of that slot is not persisted or if the slot is in the future.

### How was it fixed?
We need to get the state with `state.slot < block.slot` where `block` is the block to be validated. So we first check if head state slot is less than `block.slot`, if not then we check head block slot.
If both slots are greater than `block.slot` then we check the ancestor blocks of head block.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trinity/protocol/bcc_libp2p/topic_validators.py
+++ b/trinity/protocol/bcc_libp2p/topic_validators.py
@@ -19,9 +19,13 @@ from eth2.beacon.chains.base import BaseBeaconChain
 from eth2.beacon.helpers import compute_start_slot_of_epoch
 from eth2.beacon.types.blocks import BeaconBlock
 from eth2.beacon.types.states import BeaconState
-from eth2.beacon.state_machines.forks.serenity.block_processing import process_block_header
-from eth2.beacon.state_machines.forks.serenity.block_validation import validate_attestation
+from eth2.beacon.state_machines.forks.serenity.block_validation import (
+    validate_attestation,
+    validate_proposer_signature,
+    validate_proposer_is_not_slashed,
+)
 from eth2.beacon.typing import Slot
+from eth2.configs import CommitteeConfig
 
 from libp2p.peer.id import ID
 from libp2p.pubsub.pb import rpc_pb2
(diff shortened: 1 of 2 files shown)


trinity/protocol/bcc_libp2p/topic_validators.py
```diff
@@ -29,6 +35,59 @@
...
+    latest_finalized_slot = compute_start_slot_of_epoch(
+        state.finalized_checkpoint.epoch,
+        state_machine.config.SLOTS_PER_EPOCH,
+    )
+    if block.slot < latest_finalized_slot:
+        raise ValidationError("block is older than latest finalized slot")
+
+    # Find the proper state to verify the proposer's signature of the block
+    # head state slot >= target block slot, try head block
+    if state.slot >= block.slot:
```
ralexstokes: why would we not just get the state at the `block`'s `slot`? i don't see why we need to pull any beacon blocks out of our local state here... i'll move my thinking to the top-level comment on this review

NIC619:
> 1. keeping the `process_block_header` interface

One problem is that the `validate_block_parent_root` check in `process_block_header`:
https://github.com/ethereum/trinity/blob/96cb38c41a8d22b8d0d7f18132b8bbe0f38cd085/eth2/beacon/state_machines/forks/serenity/block_validation.py#L80-L87

This would require us to pass in the correct state, i.e., pass in the fork chain state when validating a fork block, so we need to fetch state by `block.state_root` instead of by `slot`. (I realize this part: `if state.slot >= block.slot:` is wrong in this case)

> why would we not just get the state at the `block`'s `slot`?

See the reason above and also the stale/incorrect state problem with `get_state_by_slot` mentioned in #1165 .

> there is another concern here that we may have a synchronization issue b/t the Chain and this code, i.e. the various calls we make while interacting w/ the internals of the beacon chain could change out from under us, invalidating the remainder of this computation. 

Good point! This is definitely a problem and need to be handled carefully...

NIC619:
This problem is way trickier than I thought. @ralexstokes @hwwhww @ChihChengLiang What do you think about changing it to 
1. first get the state of the block's parent block, apply state transition and use it to verify the block
2. if no parent block is found, skip validation(IIUC this is how lighthouse handles it **_for now_**)

NIC619:
Fixed by #1427

PR #1155, opened 2019-09-17
fix for mismatch genisis hash

pipermerriam:
### What was wrong?

The previous fix for the `ETHPeerPairFactory` genesis hash mismatch was adequate, but it turns out a *better* fix was pretty easy.

### How was it fixed?

In the factory we now just use the computed timestamp from the other peer if explicit contexts are not provided.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trinity/tools/factories.py
+++ b/trinity/tools/factories.py
@@ -1,4 +1,3 @@
-import time
 from typing import (
     cast,
     Any,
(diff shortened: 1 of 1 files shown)


pipermerriam:
These peer-factories should go away soon.... so probably just short term debt.

PR #1157, opened 2019-09-17
Batch the transactions we relay to peers to reduce number of outbound messages

### LINKED ISSUE #1136: Improve transaction pool efficiency
### What is wrong?

We get lots of messages from `ETH` peers with a single transaction.

Consider the case where 1 peer sends us 10x messages each with a single transaction.

If we're connected to 10 other peers, we'll end up sending 100 messages, 10 to each peer when we could have combined them and sent only 10 messages.

### How can it be fixed

The transaction pool should pipeline the incoming transactions so that it can pull them off in batches to reduce the number of outbound peer messages we end up having to send.

pipermerriam:
fixes #1136

### What was wrong?

There were some efficiency gains to be had in the transaction pool by pipelining and batching the outgoing messages.

### How was it fixed?

Buffered the transactions as they come in and send them back out in batches.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

6ff484792bbacbbc1bc681bd7269cfee

### DIFF
--- a/trinity/components/builtin/tx_pool/pool.py
+++ b/trinity/components/builtin/tx_pool/pool.py
@@ -28,8 +28,17 @@ from trinity.protocol.eth.events import (
 )
 
 
-BATCH_LOW_WATER = 200
-BATCH_HIGH_WATER = 225
+# The 'LOW_WATER` mark determines the minimum size at which we'll choose to
+# broadcast a chunk of transactions to our peers (even if we have more than
+# this locally available and ready).
+BATCH_LOW_WATER = 100
+
+# The `HIGH_WATER` mark determines the maximum number of transactions we'll
+# send in a batch to any given peer.  This is purely in place to ensure that we
+# have a strict upper bound on the total size of a `Transactions` message for
+# abnormal cases where we suddenly get a very large batch of transactions all
+# at once.
+BATCH_HIGH_WATER = 200
 
 
 class TxPool(BaseService):
--- a/trinity/components/builtin/tx_pool/pool.py
+++ b/trinity/components/builtin/tx_pool/pool.py
@@ -1,3 +1,4 @@
+import asyncio
 from typing import (
     cast,
     Callable,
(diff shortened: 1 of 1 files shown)


trinity/components/builtin/tx_pool/pool.py
```diff
@@ -67,31 +74,47 @@ def __init__(self,
...
+            # queue is either empty or we hit a sufficient size to justify
+            # sending to our peers.
+            while not self._internal_queue.empty():
+                if len(buffer) > BATCH_LOW_WATER:
+                    break
+                buffer.extend(self._internal_queue.get_nowait())
+
+            # Now that the queue is either empty or we have an adequate number
+            # to send to our peers, broadcast them to the appropriate peers.
+            for batch in partition_all(BATCH_HIGH_WATER, buffer):
```
cburgdorf: I don't understand where the difference between `BATCH_LOW_WATER` and `BATCH_HIGH_WATER` is coming from. It seem that we are saying "get me a max of 201 transactions" and then when we partition them on the different peers we say "take these maximum of 201 transactions and partition them in batches of a max size of 225". It seems as if this is a safety measure but I don't understand how it could ever be more than 201 entries (we break reading on `if len(buffer) > BATCH_LOW_WATER`)

pipermerriam: I'll add some comments.  The idea is that we might have 199 in the batch and then next chunk of transactions that comes through has 200 more.  Now our `batch` has 399 transactions in them.  I *arbitrarily* decided that 225 is a good max batch size so now we'll send them out in two chunks to avoid having a case where we could end up sending too many at the same time.  

The numbers aren't very meaningful here.  I looked briefly in the go-ethereum codebase but couldn't find any limits to number of transactions in a given `Transactions` message so I just made some up.

cburgdorf: Oh, now I see that the `_internal_queue` holds is of type `asyncio.Queue[Sequence[SignedTransactionAPI]]` not `asyncio.Queue[SignedTransactionAPI]` :see_no_evil:

PR #1159, opened 2019-09-17
Disconnect from peers that send too many duplicate transactions

pipermerriam:
### What was wrong?

A "good" peer should not send us the same transaction twice.  

### How was it fixed?

Now the tx-pool maintains two bloom filters.  One for the transactions sent to us by a peer, and one for the transactions we send to a peer.  This distinction is important since a common scenario would be for Alice and Bob to both learn about new transactions at the same time from another peer, and then to send them to each other.  If Alice is a little faster then when she receives the same transactions from Bob they will register as him sending duplicates which is not accurate.

Next, since bloom filters yield false positives, we need to account for that.  Each peer now has a "token bucket" with a refil rate of 2-per-minute and a capacity of 10.  We deduct a token each time we encounter a duplicate transaction sent from a peer.  The peer is disconnected if the bucket is empty.  This allows for the occasional false-positive to not cause un-necessary disconnections.

Experimental results suggest this is reasonable.  TODO: letting this run longer...

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/tx-pool/test_tx_pool.py
+++ b/tests/core/tx-pool/test_tx_pool.py
@@ -16,6 +16,7 @@ from trinity.components.builtin.tx_pool.validators import (
     DefaultTransactionValidator
 )
 from trinity.protocol.common.events import (
+    DisconnectPeerEvent,
     GetConnectedPeersRequest,
     GetConnectedPeersResponse,
 )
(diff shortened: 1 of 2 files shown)


carver:
:-1: on disconnecting for sending duplicate data. Some peers might retry if they think the message was lost. The more I experiment with real peers, the less I think we should disconnect (and the more I agree with the geth approach of *very* rarely disconnecting)

pipermerriam:
I now agree with @carver .  Lots and lots and lots of nodes send duplicates.

PR #1166, opened 2019-09-18
Implement proper handshake crypto

jannikluhn:
### What was wrong?

Handshake cryptography was stubbed.

### How was it fixed?

Implement proper handshake crypto. The spec isn't very detailed for this part yet, so I had to guess a little: For the key agreement step, I just reused `ecdh_agree` from the old discovery implementation which I think hasn't been changed. For the key expansion step, I guessed the hash function (SHA256, which is used everywhere else).

### DIFF
--- a/p2p/discv5/constants.py
+++ b/p2p/discv5/constants.py
@@ -26,6 +26,7 @@ IP_V4_ADDRESS_ENR_KEY = b"ip"
 UDP_PORT_ENR_KEY = b"udp"
 
 WHO_ARE_YOU_MAGIC_SUFFIX = b"WHOAREYOU"
+HKDF_INFO = b"discovery v5 key agreement"
 
 # buffer size used for incoming UDP datagrams (should be larger than MAX_PACKET_SIZE)
 DATAGRAM_BUFFER_SIZE = 2048
--- a/p2p/discv5/identity_schemes.py
+++ b/p2p/discv5/identity_schemes.py
@@ -12,6 +12,10 @@ from typing import (
     TYPE_CHECKING,
 )
 
+from cryptography.hazmat.primitives.hashes import SHA256
+from cryptography.hazmat.primitives.kdf.hkdf import HKDF
+from cryptography.hazmat.backends import default_backend as cryptography_default_backend
+
 from eth_keys.datatypes import (
     PrivateKey,
     PublicKey,
@@ -28,6 +32,10 @@ from eth_utils import (
     ValidationError,
 )
 
+from p2p.ecies import (
+    ecdh_agree,
+)
+
 from p2p.discv5.typing import (
     AES128Key,
     IDNonce,
@@ -36,6 +44,7 @@ from p2p.discv5.typing import (
 )
 from p2p.discv5.constants import (
     AES128_KEY_SIZE,
+    HKDF_INFO,
 )
 
 
@@ -169,6 +178,7 @@ class V4IdentityScheme(IdentityScheme):
     public_key_enr_key = b"secp256k1"
 
     private_key_size = 32
+    cryptography_backend = cryptography_default_backend()
 
     #
     # ENR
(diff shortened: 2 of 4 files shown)


p2p/discv5/identity_schemes.py
```diff
@@ -12,6 +12,10 @@
     TYPE_CHECKING,
 )
 
+from cryptography.hazmat.primitives.hashes import SHA256
```
pipermerriam: thoughts on using the builtin version of this from [`hashlib`](https://docs.python.org/3/library/hashlib.html)

p2p/discv5/identity_schemes.py
```diff
@@ -232,19 +242,43 @@ def compute_session_keys(cls,
...
+            initiator_node_id, recipient_node_id = remote_node_id, local_node_id
+
+        info = b"".join((
+            HKDF_INFO,
+            initiator_node_id,
+            recipient_node_id,
+        ))
+
+        hkdf = HKDF(
+            algorithm=SHA256(),
```
pipermerriam: Ah, I see maybe the `hashlib` version isn't compatible here?

p2p/discv5/identity_schemes.py
```diff
@@ -232,19 +242,43 @@ def compute_session_keys(cls,
...
+            length=3 * AES128_KEY_SIZE,
```
pipermerriam: maybe make this it's own constant?

p2p/discv5/identity_schemes.py
```diff
@@ -232,19 +242,43 @@ def compute_session_keys(cls,
...
+            salt=id_nonce,
+            info=info,
+            backend=cls.cryptography_backend,
+        )
+        expanded_key = hkdf.derive(secret)
+
+        initiator_key = expanded_key[:AES128_KEY_SIZE]
+        recipient_key = expanded_key[AES128_KEY_SIZE:2 * AES128_KEY_SIZE]
+        auth_response_key = expanded_key[2 * AES128_KEY_SIZE:3 * AES128_KEY_SIZE]
```
pipermerriam: if we want to be really paranoid, validating that each of these values is indeed `AES128_KEY_SIZE` would be good.  It shouldn't be possible but *if* `len(expanded_key) < 3 x AES_KEY_SIZE` then some of these values would be short a few bytes.

p2p/discv5/identity_schemes.py
jannikluhn: I think it's clearer as it is right now. The 3 comes from the fact that we want to split the expanded key up into three AES128 keys. If the length is a constant imported from somewhere else this relationship is harder to see.

p2p/discv5/identity_schemes.py
jannikluhn: This shouldn't be possible because the length of the expanded key is guaranteed to be `3 * AES128_KEY_SIZE` (by the `length` argument in the HKDF initialization). I can add a `raise Exception("invariant")` here though to make this more clear.

p2p/discv5/constants.py
```diff
@@ -26,6 +26,7 @@
 UDP_PORT_ENR_KEY = b"udp"
 
 WHO_ARE_YOU_MAGIC_SUFFIX = b"WHOAREYOU"
+HKDF_INFO = b"discovery v5 key agreement"
```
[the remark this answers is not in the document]
jannikluhn: Yes, this comes from the spec, nothing we can change here

p2p/discv5/identity_schemes.py
jannikluhn: Exactly, it seems to be some special object that contains some metadata, so we can't just use a plain function here.

p2p/discv5/identity_schemes.py
pipermerriam: Up to you.  The `Exception("Invariant")` approach was what I was thinking here.  It's a code path that we should never reach but it provides us with a guarantee that is otherwise *kind-of* implicit.

PR #1168, opened 2019-09-18
Slightly cleaner shutdown, CancelledError's are expected

ralexstokes:
Part of the eth2 `interop` cleanup.

From @lithp 

Catch `CancelledError`

#### Cute Animal Picture

### DIFF
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -159,6 +159,8 @@ class BaseService(CancellableMixin, AsyncioServiceAPI):
                 await awaitable
             except OperationCancelled:
                 pass
+            except asyncio.CancelledError:
+                pass
             except Exception as e:
                 self.logger.warning("Task %s finished unexpectedly: %r", awaitable, e)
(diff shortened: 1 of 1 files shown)


p2p/service.py
```diff
@@ -159,6 +159,8 @@ def run_task(self, awaitable: Awaitable[Any]) -> None:
                 await awaitable
             except OperationCancelled:
                 pass
+            except asyncio.CancelledError:
+                pass
```
ralexstokes: @lithp without looking at more context here, is this a spot where we are expected to `raise`? the `asyncio` docs suggest re-raising `CancelledError` iirc until you get to the very top-level of your app

lithp: Yeah, good question, I don't know! I think you're right that re-raising is preferable, depending on what we do with `self._tasks` it's probably better for the Future to be given the correct exception. It's worth trying that and making sure no new problems emerge!

PR #1169, opened 2019-09-18
Add ms-resolution to log messages

ralexstokes:

From @lithp .

More data!

#### Cute Animal Picture

### DIFF
--- a/trinity/_utils/logging.py
+++ b/trinity/_utils/logging.py
@@ -37,8 +37,8 @@ LOG_MAX_MB = 5
 
 class TrinityLogFormatter(logging.Formatter):
 
-    def __init__(self, fmt: str, datefmt: str) -> None:
-        super().__init__(fmt, datefmt)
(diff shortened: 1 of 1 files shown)


lithp:
In particular, this was added to try to debug a race condition where Trinity might have been creating and gossipping a block before lighthouse had noticed that a new slot had started. Because Trinity was only giving us second-resolution logs it was impossible to tell when exactly the block was gossiped, and I think this additional data is generally useful!

I feels weird to approve my own PR though, so roping @carver in :)

ralexstokes:
going to go ahead and merge this in so we can keep moving on the `interop` front...

@carver feel free to revisit if you have more input to add here :)

carver:
I think my ideal would be something like: show ms in the debug log file, show the truncated `%S` in the console output. I don't have any advice on implementing that, would just be digging into the `logging` docs to figure it out.

PR #1170, opened 2019-09-18
Fix bug in crosslink processing

ralexstokes:

Fixes a bug due to not respecting the mutability found in the spec definition of `process_crosslinks`.

Crosslinks that can be accepted earlier should be favored over crosslinks that can be accepted later during a single epoch processing.

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/state_machines/forks/serenity/epoch_processing.py
+++ b/eth2/beacon/state_machines/forks/serenity/epoch_processing.py
@@ -266,6 +266,8 @@ def process_crosslinks(state: BeaconState, config: Eth2Config) -> BeaconState:
     current_epoch = state.current_epoch(config.SLOTS_PER_EPOCH)
     previous_epoch = state.previous_epoch(config.SLOTS_PER_EPOCH, config.GENESIS_EPOCH)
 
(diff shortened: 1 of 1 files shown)


eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -296,14 +298,14 @@ def process_crosslinks(state: BeaconState, config: Eth2Config) -> BeaconState:
                 state, attesting_indices, crosslink_committee
             )
             if threshold_met:
-                new_current_crosslinks = update_tuple_item(
-                    new_current_crosslinks, shard, winning_crosslink
+                state = state.copy(
+                    current_crosslinks=update_tuple_item(
+                        new_current_crosslinks, shard, winning_crosslink
```
ChihChengLiang: Is it possible to use only `state.current_crosslinks`, and remove all other `new_current_crosslinks`?
```suggestion
                        state.current_crosslinks, shard, winning_crosslink
```

ChihChengLiang:
Note for other potential reviewers: The function `get_winning_crosslink_and_attesting_indices`, which is hidden in this PR, consumes the intermediate status of `state.current_crosslinks` in the loop.

PR #1171, opened 2019-09-18
Standardize how types are printed

ralexstokes:

### What was wrong?

This PR attempts to standardize how types are printed.

I'd prefer to start from a uniform base and then deviate from this format when/if it is necessary.

### How was it fixed?

Fixing how types print.

Adds a `__str__` for a `Bitfield` type, although it seems to only apply through one application of `__str__`, not recursively.

### TODO

- [x] Remove unnecessary `__str__` (thanks @carver !)
- [ ] Calls to `str(TYPE)` via the `logging` package do not seem to apply `str` recursively, only once... consult the `logging` docs to see if we can fix this behavior to improve the logs

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/types/attestation_data.py
+++ b/eth2/beacon/types/attestation_data.py
@@ -36,10 +36,10 @@ class AttestationData(ssz.Serializable):
 
     def __str__(self) -> str:
         return (
-            f"beacon_block_root={humanize_hash(self.beacon_block_root)[2:10]}"
-            f" source={self.source}"
-            f" target={self.target}"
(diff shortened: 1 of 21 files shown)


ralexstokes:
wip: want to take another look at the stuff under _TODO_ before merging this in

ralexstokes:
update: we get `logging.debug` calls in a class that derives `p2p.BaseService` that prints the `repr` instead of the `str`. tried a few things to replicate in an isolated example but could not recreate the behavior. going to move ahead for now and move this last part to an issue

PR #1172, opened 2019-09-18
Add dockerfile for trinity build

ralexstokes:

Adds dockerfile specifically for the beacon node.

#### Cute Animal Picture

### DIFF
--- a/Dockerfile.eth2
+++ b/Dockerfile.eth2
@@ -1,0 +1,18 @@
+FROM python:3.7
+# Set up code directory
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+COPY . /usr/src/app
+
+# Install deps
+RUN apt-get update
+RUN apt-get -y install libsnappy-dev gcc g++ cmake
+
+RUN pip install -e .[dev]  --no-cache-dir
+RUN pip install -U trinity --no-cache-dir
+
+RUN echo "Type \`trinity-beacon\` to boot or \`trinity-beacon --help\` for an overview of commands"
+
+EXPOSE 30303 30303/udp
+ENTRYPOINT ["trinity-beacon"]
--- a/Dockerfile.eth2
+++ b/Dockerfile.eth2
@@ -1,18 +1,18 @@
 FROM python:3.7
-# Set up code directory
+
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
-
-COPY . /usr/src/app
 
 # Install deps
 RUN apt-get update
 RUN apt-get -y install libsnappy-dev gcc g++ cmake
 
-RUN pip install -e .[dev]  --no-cache-dir
+ARG GITREF=interop
+
+RUN git clone https://github.com/ethereum/trinity.git .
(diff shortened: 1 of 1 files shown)


ralexstokes:
@cburgdorf aiui you were involved w/ the existing dockerfile -- is there a preferred way to have one dockerfile for both binaries?

e.g. `trinity` and `trinity-beacon`?

Dockerfile.eth2
```diff
@@ -0,0 +1,18 @@
+FROM python:3.7
+# Set up code directory
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+COPY . /usr/src/app
```
cburgdorf: I know this is also what our eth1 Dockerfile does but I'm unsure if it's the preferred approach. It means that when the image is created, the files will be copied from the host into the image. An alternative approach is to simply clone the repo from within the image.

https://github.com/ethereum/trinity/blob/6707dbf1ea2c3c50ca0b10c432e070948efe6afc/dappnode/build/Dockerfile#L11-L14

(Ironically this Dockerfile still carries a `COPY` instruction at the top that I *think* is just a leftover)

Dockerfile.eth2
```diff
@@ -0,0 +1,18 @@
...
+
+# Install deps
+RUN apt-get update
+RUN apt-get -y install libsnappy-dev gcc g++ cmake
+
+RUN pip install -e .[dev]  --no-cache-dir
+RUN pip install -U trinity --no-cache-dir
+
+RUN echo "Type \`trinity-beacon\` to boot or \`trinity-beacon --help\` for an overview of commands"
```
cburgdorf: This `RUN` :point_up: can be removed. It's not doing what it was supposed to do. The user that runs `docker run -it ethereum/trinity-beacon` won't see it and the node will immediately start because of the cmd in `ENTRYPOINT`.

Dockerfile.eth2
```diff
@@ -0,0 +1,18 @@
...
+
+EXPOSE 30303 30303/udp
+ENTRYPOINT ["trinity-beacon"]
```
cburgdorf: You might want to pass an environment variable here like `$EXTRA_OPTS` as we did in the Dappnode image.

https://github.com/ethereum/trinity/blob/6707dbf1ea2c3c50ca0b10c432e070948efe6afc/dappnode/build/Dockerfile#L18

In the case of DappNode it allows the user to set CLI parameters via the DappNode UI but even if you just run the docker image without DappNode it still allows you to pass CLI params via the docker `-e` flag.

Of course, someone using this docker container can also always do something like `docker run -it ethereum/trinity-beacon /bin/sh` hence bypassing the `ENTRYPOINT` and then once being on the shell, they can run whatever they want.

ralexstokes:
@cburgdorf thanks! i'm leaning towards just keeping two separate files; i will address the other feedback before merging this in -- looks like we have a fresher iteration of the dockerfile in the dappnode config...

Dockerfile.eth2
ralexstokes: `COPY . $DIR` seems to be a norm in the Docker community but I do like pulling from source

PR #1179, opened 2019-09-19
Add newsfragment for 1129

pipermerriam:
Just a follow up adding a newsfragment for #1129


newsfragments/1129.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix bug when negotiating inbound handshake.
```
carver: If you like: It may help people identify what this bug is fixing to include a part of the error message, and be a little more explicit about all inbound connections not working. Something like:

> Fixed handshake bug that caused all inbound connections to fail with: "AttributeError: 'Session' object has no attribute 'address'"

PR #1182, opened 2019-09-19
Use secrets.choice over random.choice

pipermerriam:
### What was wrong?

Use of `random.choice` when we now have the better option of `secrets.choice`

### How was it fixed?

Changed to use `secrets.choice`

#### Cute Animal Picture

dog-costume-with-guitar-2


pipermerriam:
This code is going away soon anyways, closing.

PR #1187, opened 2019-09-19
Add Istanbul fork block to Ropsten definition

### LINKED ISSUE #907: Include the new Istanbul fork in the Ropsten genesis json
This depends on: https://github.com/ethereum/py-evm/issues/1838

carver:
### What was wrong?

Fixes #907 

### How was it fixed?

- Added `istanbulForkBlock` to genesis.json

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


trinity/assets/eip1085/ropsten.json
```diff
@@ -1 +1 @@
...
...000000000000000000000000000000000000f6":{"balance":"0x0"},"0x00000000000000000000000000000000000000f7":{"balance":"0x0"},"0x00000000000000000000000000000000000000f8":{"balance":"0x0"},"0x00000000000000000000000000000000000000f9":{"balance":"0x0"},"0x00000000000000000000000000000000000000fa":{"balance":"0x0"},"0x00000000000000000000000000000000000000fb":{"balance":"0x0"},"0x00000000000000000000000000000000000000fc":{"balance":"0x0"},"0x00000000000000000000000000000000000000fd":{"balance":"0x0"},"0x00000000000000000000000000000000000000fe":{"balance":"0x0"},"0x00000000000000000000000000000000000000ff":{"balance":"0x0"},"0x874b54a8bd152966d63f706bae1ffeb0411921e5":{"balance":"0xc9f2c9cd04674edea40000000"}},"genesis":{"author":"0x0000000000000000000000000000000000000000","difficulty":"0x100000","extraData":"0x3535353535353535353535353535353535353535353535353535353535353535","gasLimit":"0x1000000","nonce":"0x0000000000000042","timestamp":"0x0"},"params":{"EIP150ForkBlock":"0x0","EIP158ForkBlock":"0xa","byzantiumForkBlock":"0x19f0a0","chainId":"0x3","constantinopleForkBlock":"0x408b70","petersburgForkBlock":"0x4b5e82","istanbulForkBlock":"0x62f756","miningMethod":"ethash"},"version":"1"}
```
carver: When I see diffs like this, it makes me want to switch to using whitespace, especially newlines. Any objections?

The example on https://github.com/ethereum/EIPs/issues/1085 has whitespace, and didn't seem to mention any prohibition.

trinity/protocol/common/peer.py
```diff
@@ -56,7 +56,8 @@
 
 
 class BaseChainPeer(BasePeer):
-    boot_manager_class = DAOCheckBootManager
+    # mypy doesn't understand that cached_property in the base is equivalent to assigning directly
+    boot_manager_class = DAOCheckBootManager  # type: ignore
```
carver: I don't understand what changed in this PR that made this line start failing.

carver: Woah, no idea why this is showing up for me locally, but not in CI. Removed this change.

I have:
```sh
$ pip freeze | grep mypy
mypy==0.701
mypy-extensions==0.4.1
```

which seems to be how the setup.py file defines it.

trinity/tools/chain.py
```diff
@@ -26,10 +27,10 @@ class LatestTestChain(FullChain):
     A test chain that uses the most recent mainnet VM from block 0.
     That means the VM will explicitly change when a new network upgrade is locked in.
     """
-    vm_configuration = ((0, PetersburgVM),)
+    vm_configuration = ((BlockNumber(0), PetersburgVM),)
```
pipermerriam: maybe better to just import and use `eth.constants.GENESIS_BLOCK_NUMBER`

trinity/protocol/common/peer.py
pipermerriam: Happened to me too, no clue why

trinity/assets/eip1085/ropsten.json
pipermerriam: The `mainnet` one has a *ton* of accounts and I think the file size blows up a bit when you add whitespace.  I'm not opposed to it but I think that's why I did them without whitespace initially.

carver: I added whitespace/newlines around everything but accounts. It added about 90 bytes, and gave all the benefit I was looking for. Since changes to accounts would fork the network from the start, it seems like they are always written in stone once launched, anyway. So no need to worry about big diffs there.

PR #1191, opened 2019-09-24
Handshake timeout

jannikluhn:
### What was wrong?

During handshake, the peer packer was waiting indefinitely for a response.

### How was it fixed?

Raise a `HandshakeFailure` if the handshake takes longer than `1s` (as suggested by the spec).

I also fixed some other small issues:

- use a lock in the peer packer to ensure that we don't send and receive messages simultaneously. This makes the code a little bit simpler.
- `clone` the channels that are shared between multiple peer packers
- use child services instead of tasks in the routing table manager

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/routing_table_manager.py
+++ b/p2p/discv5/routing_table_manager.py
@@ -407,12 +407,12 @@ class RoutingTableManager(Service):
         )
 
     async def run(self) -> None:
-        components = (
+        child_services = (
             self.ping_handler,
             self.find_node_handler,
             self.ping_sender,
         )
-        for component in components:
-            self.manager.run_daemon_task(component.run)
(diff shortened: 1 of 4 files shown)


p2p/discv5/routing_table_manager.py
```diff
@@ -407,12 +407,12 @@ def __init__(self,
         )
 
     async def run(self) -> None:
-        components = (
+        child_services = (
             self.ping_handler,
```
pipermerriam: *nitpick*.  Since the term `handler` could be conflated with "function" I'm inclined to suggest naming these `thing_handler_service` to make it clear that they are `ServiceAPI` instances.

PR #1196, opened 2019-09-25
Add support for Görli

cburgdorf:
### What was wrong?

Trinity should support Clique based networks such as Görli and Rinkeby

### How was it fixed?

1. This uses the recently exposed Clique consensus support in Py-EVM
2. It adds `validate_extension_chain` calls at two places to perform chain validation that is dependent on previous headers (which it does under CliqueConsensus)
### Further notes

If someone sends us a bad header this will just kill our node in its current form.

#### Cute Animal Picture


trinity/sync/common/headers.py
```diff
@@ -851,6 +851,13 @@ def _reset_buffer(self) -> None:
                 if headers:
                     # Note lack of capacity if the headers are non-consecutive
                     self._buffer_capacity.clear()
+                
+                self._buffer_capacity.clear()
+                head = await self._chain.coro_get_canonical_head()
+                self.logger.error("Current head %s, validating %s header", head, len(consecutive_batch))
+                await self._chain.coro_validate_chain(head, consecutive_batch)
```
cburgdorf: @pipermerriam @carver may I have your help with something. Please ignore the rest of the PR. Mostly a make-it-work-somehow scheme at this point.

Here's the thing that I'm not sure how to do:

As I already mentioned in Clique consensus we can not validate the seal of a header until we've validated all previous headers. That's because if you validate header `n`then it might be that the signer of that header `n` was only signed in at header `n-1`.

In our POW based syncing we call `validate_seal` on lots of headers that we haven't processed the parents for (yet). That' because we assign downloads to different peers and since POW based seals can be validated out of order it makes sense to validate them early to detect errors.

So, when I started implementing Clique support I focused on the Py-EVM side of things and for now implemented a hack to move validation to the very end of the sync. This is what we are looking at in the line above :point_up:.

Now that the Py-EVM side of things is maturing (With this hack, I can already follow the Görli chain with all its in and out votes of members :tada: ) I'm turning to the Trinity side to find out how to get that past the proof of concept phase.

1. If I validate the chain at this point and run into a validation error I have no idea which peer has caused us the bad header

One idea to solve this would be to subclass `BlockHeader` and attach the peer to it (no meaning in the RLP context, just a helper primitive within `trinity.sync`) so that we could somehow preserve which peer that header was coming from until the very last moment when we need that info. I'm personally :-1: on that but I haven't had other bright ideas yet :sweat_smile: 

2. However, even if I *knew*  which peer caused us the trouble, I'm not sure what to actually do. Just disconnecting from that peer won't cut it I guess because we've already let the machine run long past the point where the broken header appeared. I mean, we haven't *persisted* anything yet but all the headers that we hold on to in memory may be wrong. I have no idea what the right thing is at that point?

Do you have any suggestions on what to do here? I was also considering writing different syncing classes more tailored to this deferred validation approach but before I start anything crazy I'd rather hear your input first :smile:

carver: > 1. If I validate the chain at this point and run into a validation error I have no idea which peer has caused us the bad header

I'm more and more convinced that we shouldn't "punish" the peer beyond validating the data and ignoring invalid responses. So maybe it's not that big a deal that we can't trace it back to the peer. We just drop headers/blocks that are bad and never import them. As long as *some* peer is sending us good headers/blocks then we will continue down the right chain eventually.

cburgdorf: >I'm more and more convinced that we shouldn't "punish" the peer beyond validating the data and ignoring invalid responses.

Agreed! I guess tracing it back to the peer that caused the faulty header isn't the main priority.

Let me make an updated brain dump. Both to give you a better impression of what I'm struggling with and to have an extension to my 640 kb in-head memory :sweat_smile: 

>We just drop headers/blocks that are bad and never import them. As long as some peer is sending us good headers/blocks then we will continue down the right chain eventually.

That's exactly the center of my current problem!

In the regular flow today, when we fetch a segment and do the validation *on arrival* (line `729` down here :point_down:), two important things are guaranteed.

https://github.com/ethereum/trinity/blob/79c21f8ae90bc765a78cb8052af0e4271e4c25e1/trinity/sync/common/headers.py#L692-L746

1. Only validated headers go into the stitcher (line `745` up there :point_up: )
2. Invalid headers will cause us to return an empty tuple which in turn will cause the segment to become rescheduled to be downloaded by another peer.

These two properties are what keeps the machine running and be fault tolerant.

So, my current hack just validates everything that comes out of the stitcher. But that's too late because then when there's an error the beast 👹 just carries on as if the header were persisted. It won't get rescheduled to be downloaded by another peer.

But if validation can only happen after stitching then maybe the solution is to learn how to recover from a validation error *post-stitching*.

I'm thinking of doing these two things (It's all just brain gibberish so far):

1. Reset the stitcher (maybe migrate headers over that we *think* might be non affected. This is where it could be useful to know which headers came from which peer)

2. Reschedule downloads for the invalid headers

But I also think that there's a difference between an invalid header that came through *some* peer and one that came through the *skeleton* peer. And that may be another reason why we might want to carry that info further along. Because if we the faulty header comes from the skeleton peer but we have no clue won't we get stuck in an infinite loop of getting junk and throwing it away again?

carver: > So, my current hack just validates everything that comes out of the stitcher. But that's too late because then when there's an error the beast japanese_ogre just carries on as if the header were persisted. It won't get rescheduled to be downloaded by another peer.

Got it :+1: 

Maybe another option is to have another validation step just as the header syncer is emitting the headers, so that we can be guaranteed that the parent headers have already been downloaded. At that point, the headers could be rescheduled for download.

Also it's potentially worth noting that Clique is not securing any value-containing network. It's less bad if Goerli goes belly-up than if mainnet does. (There are also fewer incentives to try to actively break it)

trinity/config.py
```diff
@@ -155,22 +158,35 @@ def chain_name(self) -> str:
...
 
     @property
-    def full_chain_class(self) -> Type['FullChain']:
+    def mining_method(self) -> MiningMethod:
+        return self.genesis_data.mining_method
+
+    def apply_consensus_engine_to_vms(self, base_db: AtomicDatabaseAPI) -> VMConfiguration:
+        if self.mining_method == MiningMethod.Clique:
+            clique = CliqueConsensus(base_db)
+            return clique.amend_vm_configuration(self.vm_configuration)
```
carver: Hm, I find this approach pretty problematic: it makes the VM classes database-linked. Before they didn't know anything about the database until they were instantiated.

It looks to me like a better approach here would be to have the `vm_configuration` property amend the VMs internally before returning. Then we can drop all of this *_`chain_class` churn in the PR.

trinity/config.py
```diff
@@ -155,22 +158,35 @@ def chain_name(self) -> str:
...
+            return self.vm_configuration
+
+    def get_full_chain_class(self, base_db: AtomicDatabaseAPI=None) -> Type['FullChain']:
         from trinity.chains.full import FullChain  # noqa: F811
 
+        vms = self.apply_consensus_engine_to_vms(base_db)
         return FullChain.configure(
             __name__=self.chain_name,
-            vm_configuration=self.vm_configuration,
+            vm_configuration=vms,
```
carver: So for example, none of this has to change if `self.vm_configuration` just returns the right thing, already configured for the mining method.

trinity/sync/common/headers.py
```diff
@@ -176,7 +180,9 @@ def __init__(self,
             pairs = tuple(zip(parents, children))
             try:
                 validate_pair_coros = (
-                    self.wait(self._chain.coro_validate_chain(parent, (child, )))
+                    self.wait(self._chain.coro_validate_chain(
+                        parent, (child, ), self._validation_config.full_rate)
```
carver: Yup, this approach seems clean :+1:

trinity/sync/common/headers.py
```diff
@@ -883,6 +893,22 @@ def _reset_buffer(self) -> None:
                 if headers:
                     # Note lack of capacity if the headers are non-consecutive
                     self._buffer_capacity.clear()
+
+                if self._validation_config is VALIDATE_AFTER_STITCHING:
```
carver: Hm, this seems brittle. It seems like it would break down if we want to have two different validation configurations that validate the headers after stitching them together. Maybe add a boolean property?

(Maybe it could even be a derived property, where it's True any time the `full_rate < 1`.)

cburgdorf: >It seems like it would break down if we want to have two different validation configurations that validate the headers after stitching them together

Good point. I didn't see a case for this beyond the cases that this PR adds but maybe that's just my lack of imagination and I agree that it's easy enough to derive a property to become independent of this hardcoded condition. :+1:

trinity/config.py
cburgdorf: >Hm, I find this approach pretty problematic

I'm with you. It was hunting me in my dream already and I basically kicked it down the road knowing that it would come back up at a later point. Hopefully, we can find a way to get rid of it.

>it makes the VM classes database-linked. Before they didn't know anything about the database until they were instantiated.

That exactly is the crux! The VM classes need to be database-inked to make Clique work...unless we find a way to get around it :) Here's why: `validate_seal` is a class method and it is used from within other classmethods e.g `validate_header`.

If we want `validate_seal` to validate according to the Clique rules then it needs to be database aware because validation in clique is database dependent. We could try to lift `validate_seal` and `validate_header` to become instance methods but it will probably cause a cascade of other issues at places that expect these to be classmethods.

But that's not the only problem. Let's look at `from_genesis`

```
    @classmethod
    def from_genesis(cls,
                     base_db: AtomicDatabaseAPI,
                     genesis_params: Dict[str, HeaderParams],
                     genesis_state: AccountState=None) -> 'BaseChain':
        """
        Initializes the Chain from a genesis state.
        """
        genesis_vm_class = cls.get_vm_class_for_block_number(BlockNumber(0))

        pre_genesis_header = BlockHeader(difficulty=0, block_number=-1, gas_limit=0)
        chain_context = ChainContext(cls.chain_id)
        state = genesis_vm_class.build_state(base_db, pre_genesis_header, chain_context)
        ...
```

It looks up the vm class which means the vm class needs to be Clique-ready at this point which again means Clique needs to be setup with a database connection at this point.

So, turning back to the Trinity side, whereever we call `full_chain_class` or `light_chain_class` we need to ensure that these chains operate under Clique consensus which means we **have to** make it database aware.

I vaguely see two possible strategies:

1. Lift up a lot of things in Py-EVM to become instance methods. E.g. no more accessing vms from classmethods. I believe that would be quite a bit undertaking plus I actually like the fact that a lot of these things are classmethods today.

2. Make it so that `validate_seal` takes in an additional parameter that would be something like `ConsensusEngineAPI` and which may default to something like `PowConsensus`. And then this would likely bubble up into all the other classmethods that use `validate_seal` so that e.g. `validate_header` would also take a parameter `ConsensusEngineAPI` etc. I don't know exactly where the instance of `CliqueConsensus` would live but it would have to be passed into at places where classmethods are called that depend on `validate_seal`.

I'd be more in favor of the second option and I could explore this further unless you have other ideas (or want to take a stab at it).

Or..we could live with the annoyances of the current approach :man_shrugging:

carver: > the vm class needs to be Clique-ready at this point which again means Clique needs to be setup with a database connection at this point.

I don't follow. What goes wrong if the VM class is Clique-aware, but not database-linked, until the `build_state()` call?

> 1. Lift up a lot of things in Py-EVM to become instance methods. E.g. no more accessing vms from classmethods. I believe that would be quite a bit undertaking plus I actually like the fact that a lot of these things are classmethods today.

I *think* there will still be other classmethods that can stay in place, but we'll see.

I like that they are classmethods too! But a big part of the reason I like it is that they can't accidentally access the database from the classmethod, currently. (That is used to provide a guarantee when running batch writes, for example) Clique just forced our hand: validating the seal is no longer something you can do using a VM without state. Now we are deciding whether to change the concept that a classmethod can access state, or that validating the seal is a classmethod.

I don't think the fallout of switching to an instance method will be terrible actually. `Chain.validate_seal` is already an instance method. `VM.validate_header` is the only sacrifice, but... it's already got a weird `check_seal` boolean. It looks like it might even be a general improvement to split into `VM.validate_header_logic()` and `vm.validate_seal()`. Only the latter has to become an instance method. This affects `Chain.validate_chain()` which is currently a class-method, but isn't actually used that way anywhere, in practice.

If performance becomes a problem down the line, and we want to avoid database-linking the VMs to validate a chain in PoW chains, we could do something fancy-ish, like:
- add a `VM.seal_check_requires_db` boolean property
- `VM.stateless_validate_seal()` would throw `NotImplementedError` on the Clique VMs
- `vm.validate_seal()` would always work, but sometimes be unnecessary.

So you could validate this way:
```py
if VM.seal_check_requires_db:
  vm = get_vm()  # let's imagine this is slow
  vm.validate_seal()
else:
  VM.stateless_validate_seal()
```

The DAO check that uses `validate_header` can always use the stateless one, it doesn't even have to check first.

cburgdorf: >I don't follow. What goes wrong if the VM class is Clique-aware, but not database-linked, until the `build_state()` call?

Well, I guess I'd like to avoid a situation where we'd create the clique engine without the db first only to then set the db at a later point in time.

E.g.
```python
clique = CliqueConsensus():
# somehow attach it to VM
```

And then later have to do something like

```python
clique.set_db(db)
```

I actually had something like that as an intermediate step during development but it felt brittle since then you have parts within the `CliqueConsensus` that need to be prepared for not having the `db` set. I guess if the DB is a fundamental part of the class then it should be passed in at creation time.

>Now we are deciding whether to change the concept that a classmethod can access state, or that validating the seal is a classmethod.

Yes, good point.

You are making good points and recommendations, let me dive back into py-evm to see if I can come up with something and then we can continue to iron it out on a PR on the Py-EVM side (hopefully).

carver: > Well, I guess I'd like to avoid a situation where we'd create the clique engine without the db first only to then set the db at a later point in time.
> 
> E.g.
> 
> ```python
> clique = CliqueConsensus():
> # somehow attach it to VM
> ```
> 
> And then later have to do something like
> 
> ```python
> clique.set_db(db)
> ```

What about matching the VM model, like:

```py
# attach to VM
consensus_class = CliqueConsensus
```

And then later initialize the consensus engine when initializing the VM
```py
consensus_engine = consensus_class(db)
```

Or, we can delay creating the consensus class until it's used, doing something like putting this property on the VM:

```py
@cached_property
def consensus_engine(self):
  return self.consensus_class(self._db)
```

pipermerriam: Maybe orthogonal but I see `from_genesis` as potentially being a problematic method that could be removed and replaced by a better mechanism if it is part of the problem.  This might be tangentially related to https://github.com/ethereum/py-evm/issues/1807

setup.py
```diff
@@ -4,7 +4,7 @@
 import re
 from setuptools import setup, find_packages
 
-PYEVM_DEPENDENCY = "py-evm==0.3.0a10"
+PYEVM_DEPENDENCY = "py-evm@git+https://github.com/cburgdorf/py-evm.git@6e8484ba6475433280119167b884c76de8b594fd"
```
veox: You must've rebased the relevant `py-evm` branch or something. I wanted to check the branch out locally, but can't `pip`-install, the commit seems missing:

```
Collecting py-evm@ git+https://github.com/cburgdorf/py-evm.git@6e8484ba6475433280119167b884c76de8b594fd
  Cloning https://github.com/cburgdorf/py-evm.git (to revision 6e8484ba6475433280119167b884c76de8b594fd) to /tmp/pip-install-rnn1bh1y/py-evm
  Running command git clone -q https://github.com/cburgdorf/py-evm.git /tmp/pip-install-rnn1bh1y/py-evm
  Running command git checkout -q 6e8484ba6475433280119167b884c76de8b594fd
  fatal: reference is not a tree: 6e8484ba6475433280119167b884c76de8b594fd
ERROR: Command errored out with exit status 128: git checkout -q 6e8484ba6475433280119167b884c76de8b594fd Check the logs for full command output.
```

cburgdorf: Yeah, the latest commit would be https://github.com/ethereum/py-evm/pull/1899/commits/b9b0908cdc1e769ae7861a77ece96b359a914413

I'm optimistic that https://github.com/ethereum/py-evm/pull/1899 should be ready to land pretty soon and than Görli support should finally land in Trinity soon after.

veox: Ah, it's from a not-yet-merged PR, gotcha.

cburgdorf: It [just landed](https://github.com/ethereum/py-evm/pull/1899) so I'm going to update this PR now.

setup.py
```diff
@@ -5,7 +5,7 @@
 import re
 from setuptools import setup, find_packages
 
-PYEVM_DEPENDENCY = "py-evm==0.3.0a12"
+PYEVM_DEPENDENCY = "py-evm@git+https://github.com/cburgdorf/py-evm.git@4d0b7c0082d29f3842fbc4170b4d2d1f190e9527"  # noqa: E501
```
cburgdorf: This is waiting on https://github.com/ethereum/py-evm/pull/1908

trinity/protocol/common/boot.py
```diff
@@ -82,7 +83,11 @@ class DAOCheckBootManager(BasePeerBootManager):
                 parent, header = headers
 
             try:
-                vm_class.validate_header(header, parent, check_seal=True)
+                validate_header_is_on_intended_dao_fork(
```
cburgdorf: This is depending on https://github.com/ethereum/py-evm/pull/1908
I'm trying to avoid instantiating a VM at this point because we can not easily fulfill the constructor args. Notice that `validate_header_is_on_intended_dao_fork` does only perform the DAO check, **not** the seal check. I'm not 100% sure if it is a problem because the seal check will happen anyway as part of the sync.

trinity/sync/full/chain.py
```diff
@@ -696,6 +696,8 @@ def register_peer(self, peer: BasePeer) -> None:
             get_completed_coro = self._block_persist_tracker.ready_tasks(BLOCK_QUEUE_SIZE_TARGET)
             completed_headers = await self.wait(get_completed_coro)
 
+            self.chain.validate_chain_extension(completed_headers)
```
cburgdorf: @carver now that we have the extension check at this later stage, I'm not exactly sure how we want to handle the exception case? We would need to reschedule the bad headers to be downloaded from other peers somehow. In it's current form I guess it will simply fail the daemon task and consequentially stop the sync/node. Maybe that is enough for a Görli MVP (considering the lower reason to attack), maybe not :man_shrugging:

cburgdorf:
@carver Pinging you for a preliminary review here. It still depends on a Py-EVM release and has some rough corners (like being an easy DOS target)

trinity/protocol/common/boot.py
carver: Hm, this seems like more evidence to me that we should take the seal check out of `validate_header()` and make it a classmethod again, instead of trying to combine the seal check into the `validate_header(check_seal=True)` and forcing it into an instance method.

It comes with a pretty big downside, of course, that someone might *forget* to run the two separate validations. But being able to run a validation as a `classmethod` without initializing the VM is useful, as you discovered here.

trinity/sync/full/chain.py
carver: Hm, I think the MVP would be to log the bad block (maybe even the full RLP to debug logs, for inspection) and continue. Making it quite that easy for someone else to knock all trinity nodes offline is a bad idea, even for Goerli. Sure, if a bug happens or someone mines a long line of bad blocks, then there will be a lot of wasted, weird logs, but that's better than crashing.

cburgdorf: >Hm, I think the MVP would be to log the bad block (maybe even the full RLP to debug logs, for inspection) and continue.

If I'm not mistaken this will just mean that our sync stops from there. Because if header/block `n` is not persisted then clearly `n+1`, `n+2`, `n+x` will also fail the `validate_chain_extension` call.

I guess I need to figure out how to reschedule the bad header to be downloaded from another peer (Also I don't know the peer that gave us the bad header, so I might get stuck in an infinite loop here).

trinity/protocol/common/boot.py
cburgdorf: Yes, I agree. It also makes it symmetric with `validate_seal_extension` which needs to be called as a standalone, as well.

Here's the PR for that change.
https://github.com/ethereum/py-evm/pull/1909

carver:
I didn't have a time block for a 1 kiloline PR today, but I'll aim for Monday

trinity/sync/full/chain.py
carver: Are you talking about a block that's actually good, but some bug causes trinity to think it's bad? Or some actually bad block? Because an actually bad block shouldn't typically have a ton of children.

cburgdorf: >Because an actually bad block shouldn't typically have a ton of children.

Unless it might be some kind of malicious attack. Also keep in mind that there's no PoW overhead here to create as many child blocks as we want.

Let's say we are syncing headers with a peer that for whatever reasons turns out to be sending us invalid headers. At the very first header `n` where we notice that the validation fails, we might already have thousands of subsequent headers downloaded from that peer (I'm assuming bare header sync now).

But even if it is just a single header, the validation step is now at a late stage of the processing pipeline, so there's currently nothing that would reschedule the download. And even if we reschedule the download, we likely fetch it from the same (attacker) peer again.

I think on the PoW chain, our `ExchangeManager` will at some point disconnect from a client that is sending us headers that fail to validate but that doesn't work here because we notice the failure too late in the processing pipeline.

So it seems to me that all it takes for us to be wiped from the network is to be connected to an attacker that is serving us a bad chain forever.

Maybe, because this is just Görli and an attack is unlikely, it would be reasonable just disconnect from all peers (or all but the bootnode peers?) and then reschedule the downloads? Or am I looking at this all wrong?

tests/conftest.py
```diff
@@ -241,19 +243,14 @@ def chain_without_block_validation(
     which can be found in the funded_address and private_keys variables in the
     chain itself.
     """
-    # Disable block validation so that we don't need to construct finalized blocks.
-    overrides = {
-        'import_block': import_block_without_validation,
-        'validate_block': lambda self, block: None,
-    }
-    SpuriousDragonVMForTesting = SpuriousDragonVM.configure(validate_seal=lambda block: None)
```
carver: Getting rid of all this is great.

tests/core/configuration/test_chain_config_object.py
```diff
@@ -93,7 +92,10 @@ def test_chain_config_from_eip1085_genesis_config():
     chain_config = Eth1ChainConfig.from_eip1085_genesis_config(EIP1085_GENESIS_CONFIG)
 
     assert chain_config.chain_id == 1234
-    assert chain_config.vm_configuration == ((0, ConstantinopleVM),)
+    assert len(chain_config.vm_configuration) == 1
+    fork_block, fork_vm = chain_config.vm_configuration[0]
+    assert fork_block == 0
+    assert fork_vm.fork == 'constantinople'
```
carver: :+1: for shipping this way, but it feels like we ought to expose some kind of identifier on the VM besides a string to test for the fork version, like an enum.

tests/core/eip1085-utils/test_preconfigured_genesis_files.py
```diff
@@ -123,11 +127,32 @@ def test_mainnet_eip1085_rejects_etc_homestead_header(mainnet_genesis_config):
     homestead_vm = dict(genesis_data.vm_configuration)[HOMESTEAD_MAINNET_BLOCK]
 
     # The VM we derive from mainnet.json should validate the ETH header at the fork
-    homestead_vm.validate_header(ETH_HEADER_AT_FORK, PRE_FORK_HEADER, check_seal=True)
+    homestead_vm.validate_header(ETH_HEADER_AT_FORK, PRE_FORK_HEADER)
 
     # But it should reject the ETC header
     with pytest.raises(ValidationError, match='must have extra data 0x64616f2d686172642d666f726b'):
-        homestead_vm.validate_header(ETC_HEADER_AT_FORK, PRE_FORK_HEADER, check_seal=True)
+        homestead_vm.validate_header(ETC_HEADER_AT_FORK, PRE_FORK_HEADER)
```
carver: Maybe worth adding a new test case that shows that `validate_header` passes even if POW fails, but `validate_seal` raises.

trinity/_utils/eip1085.py
```diff
@@ -267,13 +291,16 @@ def extract_genesis_data(raw_genesis_config: RawEIP1085Dict) -> GenesisData:
...
+
+    miningMethod = extract_mining_method(raw_genesis_config)
     state = extract_genesis_state(raw_genesis_config)
     params = extract_genesis_params(raw_genesis_config)
     vm_configuration = extract_vm_configuration(raw_genesis_config)
     chain_id = extract_chain_id(raw_genesis_config)
 
     return GenesisData(
         chain_id=chain_id,
+        mining_method=miningMethod,
```
carver: camel case?

trinity/sync/full/chain.py
carver: Ah, so it seems like the biggest challenge we face here is that by the time we do this validation, we've dropped any information about who the header came from, and don't have any way to feed the bad header back to the downloader. So a peer could serve millions of bad (PoA) headers to us and we can't act on it.

I'm tempted to try to add a new way to feed "bad header" info back to the downloader, with a new API. But probably not in this PR. As you mention, this will only affect PoA for now, because PoW is checked earlier in the process. So I'm okay with leaving a little funky Goerli behavior for now.

Probably worth adding an issue to remember to handle the case better, and double-check that PoW is being validated earlier in the flow.

tests/core/configuration/test_chain_config_object.py
cburgdorf: Yeah, I agree we should address this. Another thought would be to change our `ConfigureableAPI` to expose a property such as `ancestry` which would have two subproperties `parent_cls`and `root_cls`. The first giving you the immediate parent (allowing to go up `.ancestry.parent_cls.ancestry.parent_cls.ancestry.parent_cls` until we've reached the root) and the second being a shortcut to get to the very root directly.

If we had this, we could always compare `vm_class.ancestry.root_cls` against something like `ConstantinopleVM` (assuming that `vm_class` is rooted in a `ConstantinopleVM`.

I'm not sure if it's worth it (considering that we [might want to migrate away](https://github.com/ethereum/py-evm/issues/622) from the `ConfigureableAPI` in the longer term anyway)

One possible downside of this is that each configured class carries a bit of extra bloat in memory. It would also be a foot gun if we had code (I'm not aware that we have) that continuously would configure a class to derive things dynamically. In that case, the memory growth would be unbounded.

trinity/_utils/eip1085.py
cburgdorf: Ooops...fixed

tests/core/eip1085-utils/test_preconfigured_genesis_files.py
cburgdorf: I agree that this is a useful test but I think we should add it on the Py-EVM side. After all it has nothing to do with the genesis file parsing or ETH/ETC fork and is a core assumption of the Py-EVM library, not Trinity. I'll make sure to open a PR on that end.

trinity/sync/full/chain.py
cburgdorf: >Ah, so it seems like the biggest challenge we face here is that by the time we do this validation, we've dropped any information about who the header came from, and don't have any way to feed the bad header back to the downloader. So a peer could serve millions of bad (PoA) headers to us and we can't act on it.

Exactly!

>I'm tempted to try to add a new way to feed "bad header" info back to the downloader, with a new API

Yeah, that sounds right. I wonder if we might only be able to catch the bad peer on the second attempt. Like, we signal that header `x` was bad via a new API but then it might be that it remains unclear which peer served us that header, however the bad header is put on a block list and rescheduled for download, then if we come by that header again (maybe by the same peer that served us that header before) we can disconnect from the peer this time because are earlier in the processing pipeline now.

>But probably not in this PR. As you mention, this will only affect PoA for now, because PoW is checked earlier in the process. So I'm okay with leaving a little funky Goerli behavior for now.

Cool :pray: 

>Probably worth adding an issue

I'll do that!

tests/core/eip1085-utils/test_preconfigured_genesis_files.py
cburgdorf: Here it is https://github.com/ethereum/py-evm/pull/1911

trinity/_utils/eip1085.py
```diff
@@ -234,6 +246,18 @@ def extract_chain_id(genesis_config: RawEIP1085Dict) -> int:
     return to_int(hexstr=genesis_config['params']['chainId'])
 
 
+def extract_mining_method(genesis_config: RawEIP1085Dict) -> MiningMethod:
+    miningMethod = genesis_config['params']['miningMethod']
+    if miningMethod.lower() == "clique":
```
pipermerriam: If you use these string values in the `MiningMethod` enum then you could just do `return MiningMethod(miningMethod.lower())` which will raise a `TypeError` if the string isn't part of the enum values.

trinity/config.py
```diff
@@ -154,13 +165,35 @@ def chain_name(self) -> str:
...
 
+    @property
+    def mining_method(self) -> MiningMethod:
+        return self.genesis_data.mining_method
+
+    def apply_consensus_engine(self, vms: VMConfiguration) -> VMConfiguration:
+        if self.mining_method == MiningMethod.Clique:
+            return CliqueApplier().amend_vm_configuration(vms)
+        elif self.mining_method == MiningMethod.NoProof:
+            return ConsensusApplier(NoProofConsensus).amend_vm_configuration(vms)
```
pipermerriam: Do we support applying different mining methods to different VMs?  If not then we need to start at least thinking about how we'll do that since at some point the mainnet will go POW -> POS and we'll need the ability to have different forks use different mining methods.

trinity/sync/full/chain.py
carver: > I wonder if we might only be able to catch the bad peer on the second attempt. Like, we signal that header `x` was bad via a new API but then it might be that it remains unclear which peer served us that header, however the bad header is put on a block list and rescheduled for download, then if we come by that header again (maybe by the same peer that served us that header before) we can disconnect from the peer this time because are earlier in the processing pipeline now.

Hm, that approach seems like it might get gamed. Another option is to keep around a mapping of who sent us which headers. Then we have to decide how long to keep them. We could also have some kind of feedback to the downloader about which headers were successful. Like `header_downloader.report_header_validity(header, valid=True)`, at which point we can drop the mapping. Or on `valid=False`, we can blacklist the peer, etc.

trinity/config.py
carver: Yeah, it's more work of course, but I don't see anything stopping us. Something like:

```py
pow_vms = ConsensusApplier(PowConsensus).amend_vm_configuration(vms[:FIRST_POS_VM_INDEX))
pos_vms = ConsensusApplier(PosConsensus).amend_vm_configuration(vms[FIRST_POS_VM_INDEX:]))
return pow_vms + pos_vms
```

trinity/_utils/eip1085.py
cburgdorf: Uh good point. I'll send a PR to clean that up :+1:

trinity/config.py
cburgdorf: Correct, or just a specialized Applier such as `PowThenPosAfterForkApplier().amend_vm_configuration(vms, SomeCrazyCityNameVM)`

trinity/sync/full/chain.py
cburgdorf: >The biggest downside is that we have to decide how long to keep the mapping.

If we not only remember the hash of the bad header but also the number then we can clean up the mapping of all headers were the number is lower than the highest persisted number. If I'm not entirely off then not even reorgs affect this because even if we reorg to another chain, the *number* would not get down again.

PR #1197, opened 2019-09-25
Kademlia Routing Table

jannikluhn:
### What was wrong?

The routing table used in the discv5 module was just an unstructured list.

### How was it fixed?

The routing table still is an unstructured list, but I've added an implementation for a Kademlia based alternative. The routing table manager will be updated accordingly in a future PR.

#### Cute Animal Picture


p2p/discv5/routing_table.py
```diff
@@ -1,19 +1,31 @@
 from collections import (
     deque,
 )
+from functools import (
+    partial,
+)
+from itertools import (
+    chain,
```
pipermerriam: *my nitpicky code style but....*

I'm generally :-1: on importing like this.

https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#dont-use-from-datetime-import-datetime

It's not exactly the same but lower down when someone sees `chain` they have to jump up here to the import statements to know whether it is `itertools.chain` or some other `chain` function...

p2p/discv5/routing_table.py
```diff
@@ -67,3 +79,185 @@ def get_random_entry(self) -> NodeID:
...
+        if node_id == self.center_node_id:
+            raise ValueError("Cannot insert center node into routing table")
+
+        bucket_index = self.get_bucket_index(node_id)
+        bucket = self.buckets[bucket_index]
+        replacement_cache = self.replacement_caches[bucket_index]
+
+        is_bucket_full = len(bucket) >= self.bucket_size
+
+        if node_id not in bucket and not is_bucket_full:
```
pipermerriam: probably also worth caching this value as `is_node_in_bucket = node in bucket`

p2p/discv5/routing_table.py
```diff
@@ -67,3 +79,185 @@ def get_random_entry(self) -> NodeID:
...
+        If the bucket is already full, the node id will be added to the replacement cache and
+        the oldest node is returned as an eviction candidate. Otherwise, the return value is
+        `None`.
+        """
+        if node_id == self.center_node_id:
+            raise ValueError("Cannot insert center node into routing table")
+
+        bucket_index = self.get_bucket_index(node_id)
+        bucket = self.buckets[bucket_index]
+        replacement_cache = self.replacement_caches[bucket_index]
```
pipermerriam: These three lines are repeated verbatim.  Maybe add a helper: `index, bucket, replacement_cache = self.get_index_bucket_and_replacement_cache(node_id)`

p2p/discv5/routing_table.py
```diff
@@ -67,3 +79,185 @@ def get_random_entry(self) -> NodeID:
...
+        except IndexError:
+            raise ValueError("Routing table is empty")
+        else:
+            return bucket_index
+
+    def iter_nodes_around(self, reference_node_id: NodeID) -> Generator[NodeID, None, None]:
+        """Iterate over all nodes in the routing table ordered by distance to a given reference."""
+        all_node_ids = chain(*self.buckets)
+        distance_to_reference = partial(compute_distance, reference_node_id)
+        sorted_node_ids = sorted(all_node_ids, key=distance_to_reference)
```
pipermerriam: Worth benchmarking how expensive this is when all the buckets are full because it *appears* that there are some more complex methods to achieving this which don't require sorting then entire data set.

p2p/tools/factories/discovery.py
```diff
@@ -145,6 +157,69 @@ class Meta:
     node_id = factory.Faker("binary", length=32)
 
 
+class NodeIDAtDistanceFactory(factory.Factory):
```
pipermerriam: Maybe instead of another factory this could be implemented as a `classmethod` on the `NodeIDFactory`?

```python
node = NodeIDFactory.at_distance(...)
```

p2p/discv5/routing_table.py
jannikluhn: with a completely full routing table it takes roughly 3-4 ms on my machine. With 16 buckets (which is probably still unrealistic) it's 0.5 ms. I think that's acceptable.

Not sure where I should put the benchmark function, just as a test that checks the time is below a threshold? Or would that be too flaky?

pipermerriam: I don't think it needs a test, I just wanted to validate that it wasn't going to be expensive.

PR #1198, opened 2019-09-25
Implement gather function with trio

jannikluhn:
### What was wrong?

Running a bunch of tasks in parallel is easy with trio using nurseries. Getting their return values however is surprisingly cumbersome.

One use case where this is needed in discv5 is during the handling of `FINDNODE` requests: We get up to 16 node ids from the routing table and now have to get the corresponding ENRs from the ENR db which we'd like to do in parallel instead of sequentially.

### How was it fixed?

Implement a function similar to [`asyncio.gather`](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather) that runs async functions in parallel and returns the results as a tuple.

The `FINDNODE` example would be a simple one liner:

```Py
enrs = gather((enr_db.get, node_id) for node_id in node_ids)
```

I've deviated a little from the asyncio API:

- Accept tuples of the form `(async_fn, *args)` as arguments instead of awaitable objects. This mirrors the `nursery.start_soon` syntax.
- Don't implement the `return_exceptions` argument. I'm not sure if it is needed and I didn't want to think about how this would look like with trio's multi errors at the moment.

Unfortunately, I have not managed to come up with an appropriate type hint.

#### Cute Animal Picture

### DIFF
--- a/p2p/trio_utils.py
+++ b/p2p/trio_utils.py
@@ -1,0 +1,49 @@
+import asyncio
+import operator
+from typing import (
+    Any,
+    Awaitable,
+    Callable,
+    Iterable,
+    Tuple,
+    Union,
+)
+
+import trio
+
+
+AsyncFnsAndArgsType = Union[
+    Callable[..., Awaitable[Any]],
+    Tuple[Any, ...],
+]
+
+
+async def gather(*async_fns_and_args: AsyncFnsAndArgsType) -> Tuple[Any, ...]:
+    """Run a collection of async functions in parallel and collect their results.
+
+    The results will be in the same order as the corresponding async functions.
+    """
+    indices_and_results = []
+
+    async def get_result(index: int) -> None:
+        async_fn_and_args = async_fns_and_args[index]
+        if isinstance(async_fn_and_args, Iterable):
+            async_fn, *args = async_fn_and_args
+        elif asyncio.iscoroutinefunction(async_fn_and_args):
+            async_fn = async_fn_and_args
+            args = []
+        else:
+            raise TypeError(
+                "Each argument must be either an async function or a tuple consisting of an "
+                "async function followed by its arguments"
+            )
+
+        result = await async_fn(*args)
+        indices_and_results.append((index, result))
+
+    async with trio.open_nursery() as nursery:
+        for index in range(len(async_fns_and_args)):
+            nursery.start_soon(get_result, index)
+
+    indices_and_results_sorted = sorted(indices_and_results, key=operator.itemgetter(0))
+    return tuple(result for _, result in indices_and_results_sorted)
--- a/tests-trio/p2p-trio/test_trio_utils.py
+++ b/tests-trio/p2p-trio/test_trio_utils.py
@@ -1,0 +1,44 @@
+import pytest
+
+import trio
+
+from p2p.trio_utils import (
+    gather,
+)
+
+
+@pytest.mark.trio
+async def test_empty_gather():
+    result = await gather()
+    assert result == ()
+
+    with trio.testing.assert_checkpoints():
+        await gather()
+
+
+@pytest.mark.trio
+async def test_gather_sorted(autojump_clock):
+    async def f(return_value, sleep_time):
+        await trio.sleep(sleep_time)
+        return return_value
+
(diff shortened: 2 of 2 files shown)


p2p/trio_utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+    The results will be in the same order as the corresponding async functions.
+    """
+    indices_and_results = []
+
+    async def get_result(index: int) -> None:
+        async_fn_and_args = async_fns_and_args[index]
+        if isinstance(async_fn_and_args, Iterable):
+            async_fn, *args = async_fn_and_args
+        else:
```
pipermerriam: rather than this implicit approach, what about this being:

```
elif asyncio.iscoroutinefunction(async_fn_and_args):
    async_fn = async_fn_and_args
```

Then we can just have a final else clase that raises an exception.

p2p/trio_utils.py
```diff
@@ -0,0 +1,45 @@
...
+            args = []
+
+        result = await async_fn(*args)
+        indices_and_results.append((index, result))
+
+    async with trio.open_nursery() as nursery:
+        for index in range(len(async_fns_and_args)):
+            nursery.start_soon(get_result, index)
+
+    indices_and_results_sorted = sorted(indices_and_results, key=first)
```
pipermerriam: *nitpick*

since this *might* be something we extract to a stand-alone library, maybe use `operator.itemgetter(0)` to keep the cytoolz dependency out of this?

tests-trio/p2p-trio/test_trio_utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+@pytest.mark.trio
+async def test_gather_sorted():
+    async def f(return_value, sleep_time):
+        await trio.sleep(sleep_time)
+        return return_value
+
+    results = await gather(
+        (f, 0, 0.1),
+        (f, 1, 0.2),
```
pipermerriam: Instead of using sleep times in these, how about using `trio.Event()` like this:

```python
ev_a = trio.Event()
ev_b = trio.Event()

async def f(wait_event, set_event=None):
    if wait_event:
        await wait_event.wait()
    if set_event:
        set_event.set()

results = await gather(
    (f, ev_a, ev_b),
    (f, ev_b, None),
    (f, None, ev_a)
)
```

This should have the same effect of having the coroutines finishe out-of-order but without the artificial wait times.

p2p/trio_utils.py
```diff
@@ -0,0 +1,45 @@
...
+        if isinstance(async_fn_and_args, Iterable):
+            async_fn, *args = async_fn_and_args
+        else:
+            async_fn = async_fn_and_args
+            args = []
+
+        result = await async_fn(*args)
+        indices_and_results.append((index, result))
+
+    async with trio.open_nursery() as nursery:
```
pipermerriam: might be nice to allow the caller to pass in their own nursery but this is probably a premature optimization.

tests-trio/p2p-trio/test_trio_utils.py
jannikluhn: I like the sleeps because they make the test easier to write and understand than the events. I've added the (pretty amazing) `autojump_clock` fixture that jumps over the idle times to get the best of both approaches.

jannikluhn:
> I'm in no means sold on this as the right API. It would naturally limit us to only homogeneous coroutines which I think covers the vast majority of use cases, but I also know we do use asyncio.gather on heterogeneous sets of awaitables....

Yeah, I think we should allow this. It's not very clean from a typing perspective, but for simple cases I'd say it's justified (e.g. `a, b = await gather(get_a, get_b)`).

pipermerriam:
I was curious so I went and looked how `asyncio.gather` is typed with `mypy`.

https://github.com/python/typeshed/blob/04bcb89aea991e45cffab05cb9520577dba9ad20/stdlib/3/asyncio/tasks.pyi#L30-L51

Looks like it will only preserve type data with up to 5 arguments, after which it falls back to `Any`.

cburgdorf:
>It's kind of ugly but maybe practical

I think that's the most practical solution here. I like it :+1:

PR #1199, opened 2019-09-25
Use behavior api for exchange api

pipermerriam:
### What was wrong?

The request/response API needs to migrate onto the new Behavior/Logic based APIs

### How was it fixed?

Made minimal adjustments to move the API over onto the `ETHAPI` and `LESAPI`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

happy-dog-running-by-500px


pipermerriam:
FYI, my current development philosophy for some of this stuff is "do the fastest thing that works" to get functionality off of the peer class.

p2p/exchange/abc.py
```diff
@@ -191,7 +183,7 @@ def is_requesting(self) -> bool:
         ...
 
 
-class ExchangeAPI(ABC, Generic[TRequestPayload, TResponsePayload, TResult]):
+class ExchangeAPI(Generic[TRequestPayload, TResponsePayload, TResult]):
     """
     The exchange object handles a few things, in rough order:
```
carver: Worth a quick review to see if anything in this list changed...

p2p/exchange/exchange.py
```diff
@@ -31,9 +60,6 @@ def __init__(self, mgr: ExchangeManager[TRequestPayload, TResponsePayload, TResu
         - the manager service is running
         - the payload validator is primed with the request payload
         """
-        if not self._manager.is_operational:
```
carver: I imagine the error message later on will be confusing, if you forgot to use the `run_exchange()` context manager.

Maybe something like:
```py
try:
  manager = self._manager
except AttributeError as exc:
  raise ValidationError("Cannot get a result outside of an exchange.run_exchange() context") from exc
```

Then `s/self._manager/manager/` later in the method.

p2p/exchange/manager.py
```diff
@@ -68,16 +44,6 @@ def is_operational(self) -> bool:
...
             timeout: float = None) -> TResult:
 
-        if not self.is_operational:
-            if self.service is None or not self.service.is_cancelled:
-                raise ValidationError(
-                    f"Must call `launch_service` before sending request to {self._connection}"
-                )
-            else:
-                raise PeerConnectionLost(
-                    f"Response stream closed before sending request to {self._connection}"
```
carver: Hm, I wonder if this is still handled somewhere: if we make a request to a peer who has gone away.

p2p/tools/factories/peer.py
```diff
@@ -74,6 +75,8 @@
         bob = bob_factory.create_peer(connection=bob_connection)
 
         async with run_service(alice), run_service(bob):
+            await asyncio.wait_for(alice.ready.wait(), timeout=1)
+            await asyncio.wait_for(bob.ready.wait(), timeout=1)
```
carver: I wonder if we're failing to test some code paths here of what happens if we try to talk to the peer too soon.

Maybe this should be somewhere in the `get_result()` code path so that we don't need this here, and people don't get bitten by forgetting to check `.ready` before making a request.

tests/core/p2p-proto/test_peer_block_body_validator_api.py
```diff
@@ -90,7 +90,7 @@ def mk_headers(*counts):
         remote.sub_proto.send_block_bodies([])
         await asyncio.sleep(0)
 
-    get_bodies_task = asyncio.ensure_future(peer.requests.get_block_bodies(headers))
+    get_bodies_task = asyncio.ensure_future(peer.eth_api.get_block_bodies(headers))
```
carver: :tada:

trinity/sync/common/peers.py
```diff
@@ -63,7 +63,7 @@ def __init__(
     def _get_peer_rank(self, peer: TChainPeer) -> float:
         scores = [
             self._sort_key(exchange.tracker)
-            for exchange in peer.requests
+            for exchange in peer.chain_api.exchanges
             if issubclass(exchange.get_response_cmd_type(), self._response_command_type)
```
carver: `_get_peer_rank()` really wants some kind of `peer.get_trackers_by_command_type(self._response_command_type)` here. (or maybe `get_exchanges'?)

It seems like something similar might be available now, but I'm a little lost in all the new APIs.

pipermerriam: I'm invisioning a generic stats API being added that could expose APIs like this but since this seems to work as-is I'm gonna punt.

p2p/tools/factories/peer.py
pipermerriam: I'm adding the same `await peer.ready.wait()` to the `PeerPool.start_peer()` which will keep the peers out of the pool until the APIs are available.

p2p/exchange/manager.py
pipermerriam: I've re-added this check and exception raising.

PR #1200, opened 2019-09-26
Fix run beacon nodes

ChihChengLiang:
### What was wrong?

`python eth2/beacon/scripts/run_beacon_nodes.py` breaks after interops merged.

### How was it fixed?

- Use the new interops command in `run_beacon_nodes`
- Also fix a jsonrpc component bug

#### Cute Animal Picture

### DIFF
--- a/trinity/components/builtin/json_rpc/component.py
+++ b/trinity/components/builtin/json_rpc/component.py
@@ -3,7 +3,7 @@ from argparse import (
     _SubParsersAction,
 )
 import asyncio
-from typing import Union
+from typing import Union, Tuple
 
 from lahja import EndpointAPI
 
@@ -42,6 +42,7 @@ from trinity.rpc.http import (
 from trinity._utils.shutdown import (
     exit_with_services,
 )
+from p2p.service import BaseService
 
 
 class JsonRpcServerComponent(AsyncioIsolatedComponent):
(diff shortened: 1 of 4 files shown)


trinity/components/builtin/json_rpc/component.py
```diff
@@ -124,17 +125,16 @@ def do_start(self) -> None:
...
         if self.boot_info.args.enable_http:
             http_server = HTTPServer(rpc, port=self.boot_info.args.rpcport)
-            asyncio.ensure_future(exit_with_services(
-                http_server,
-                self._event_bus_service,
-            ))
             asyncio.ensure_future(http_server.run())
+            services_to_exit += (http_server,)
+
+        asyncio.ensure_future(exit_with_services(*services_to_exit))
```
ChihChengLiang: It looks like `exit_with_services` can't be used twice in a component, otherwise `asyncio` will complain about the first one is already waited. So I fix it like this. /cc @hwwhww

trinity/components/eth2/interop/component.py
```diff
@@ -143,8 +143,7 @@ def munge_all_args(cls, args: Namespace, trinity_config: TrinityConfig) -> None:
         now = int(time.time())
         if args.start_time:
             if args.start_time <= now:
-                logger.info(f"--start-time must be a time in the future. Current time is {now}")
-                sys.exit(1)
+                logger.warning(f"--start-time must be a time in the future. Current time is {now}")
```
ChihChengLiang: Allowing a start time before now here, so we can simulate bob wakes up 20 seconds after the start_time and syncs. /cc @ralexstokes

ChihChengLiang:
Merging this one to unblock #1201 . Post PR feedback welcomed

PR #1202, opened 2019-09-26
Upgrade to mypy 0.730

cburgdorf:
### What was wrong?

We need to upgrade to mypy 0.730 ([blog announcement](https://mypy-lang.blogspot.com/2019/09/mypy-730-released.html))

### How was it fixed?

Pinned new version and fixed everything that was reported.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


p2p/p2p_api.py
```diff
@@ -51,9 +51,6 @@ class P2PAPI(Application):
 
     logger = logging.getLogger('p2p.p2p_api.P2PAPI')
 
-    client_version_string: str
-    safe_client_version_string: str
-
```
cburgdorf: Duplicate definition (they are defined further down)

p2p/service.py
```diff
@@ -37,7 +37,6 @@ def __init__(self) -> None:
 
 
 class BaseService(CancellableMixin, AsyncioServiceAPI):
-    logger: ExtendedDebugLogger = None
```
cburgdorf: Duplicated definiton (is defined further down)

trinity/components/builtin/network_db/eth1_peer_db/tracker.py
```diff
@@ -261,21 +261,21 @@ def _get_candidate_filter_query(self) -> Iterable[Any]:
 
         if self.protocols is not None:
             if len(self.protocols) == 1:
-                yield Remote.protocol == self.protocols[0]  # type: ignore
```
cburgdorf: Gotta love it when mypy just gets smarter

trinity/server.py
```diff
@@ -264,11 +264,11 @@ def __init__(self,
                  ) -> None:
         # mypy does not like `BaseAsyncBeaconChainDB` in this super call since
         # the server defines it as a ChainDB.
-        super().__init__(  # type: ignore
+        super().__init__(
             privkey,
             port,
             chain,
-            chaindb,
+            chaindb,  # type: ignore
```
cburgdorf: yay for `# type: ignore` finally being requested on the affected line

PR #1204, opened 2019-09-27
Use Kademlia Routing Table

jannikluhn:
### What was wrong?

The routing table manager used the flat routing table.

### How was it fixed?

Migrate it to use the Kademlia table.

In addition, I had to add a `get_random_node` method which I forgot earlier so that the ping sender can find someone to ping. The old implementation actually pinged the oldest node instead of a random one, but this was a bad idea as we would just end up repeatedly pinging the peer that's offline the longest. We can't rely on them being replaced quickly as they might by chance be in a mostly empty bucket.

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/routing_table.py
+++ b/p2p/discv5/routing_table.py
@@ -2,6 +2,7 @@ import collections
 import functools
 import itertools
 import logging
+import random
 import secrets
 from typing import (
     Any,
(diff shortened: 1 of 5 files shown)


p2p/discv5/routing_table.py
```diff
@@ -260,3 +261,15 @@ def iter_nodes_around(self, reference_node_id: NodeID) -> Generator[NodeID, None
...
         for node_id in sorted_node_ids:
             yield node_id
+
+    def get_random_node(self) -> NodeID:
+        if self.is_empty:
+            raise ValueError("The routing table is empty")
+
+        bucket = random.choices(
+            self.buckets,
+            weights=tuple(len(bucket) for bucket in self.buckets),
```
pipermerriam: I hadn't seen this API yet, awesome!

Issue #1158, opened 2019-09-17
Single-step Trinity Installation


lithp:
I couldn't find an existing issue on the topic of packaging so I'm opening this one. During the ETH2 interop retreat a couple people had a hard time installing Trinity, though most people were eventually able to. Opening a ticket to start a discussion about what we can do to make installing Trinity a seamless process.

I think there are two sticking points:
- Trinity relies on a modern version of python, version 3.6, which is not the default for I think any distribution. For people unfamiliar with python installing 3.6 and then virtualenv are unfortunate barriers.
- Trinity depends on some packages which depend on the system leveldb and snappy libraries.
  - The Arch snappy package appears to be broken (or at least it was last week) so we were unable to get Trinity running on the machine of someone at the interop retreat who was using arch.
  - The brew packages (or the python packages) for leveldb and snappy are entirely broken on OSX. Installing them requires stackoverflow and cargo-culting some command line flags.

There are a couple things we could do to make installing Trinity easier (which will be useful both for end-users and also for other eth2 teams who are trying to make their clients work well with Trinity)

- We could look into the `python-snappy` and `plyvel` packages and try to upstream changes which make installing them on OSX easier.
- We could push our own versions of `python-snappy` and `plyvel` which build their own snappy and leveldb packages so that we're no longer relying on the system libraries.
- We could make our own trinity homebrew package which includes snappy and leveldb libraries, so we're no longer relying on the system libraries
- We could create a [snap](https://snapcraft.io/) package so that people can install trinity with a single command.
- We could build an executable by [freezing](https://docs.python-guide.org/shipping/freezing/#freezing-your-code-ref) trinity and distribute that built executable.
- Release as a combined AppImage
- Host a repository to be used with debian-like installs via apt

ralexstokes:
i'm hoping `pyoxidizer` solves all of our problems here...

https://github.com/ethereum/trinity/issues/753

carver:
Some more options to consider for *nix-y distros:
- Flatpak
- AppImage

[Comparison of snap to flatpack to appimage](https://askubuntu.com/a/1009061/725986).

lithp:
AppImage actually looks super cool, I don't think it requires users installing anything, it should let us distribute a `./trinity` binary which just runs! #1160 experiments with getting `snap` to work, you can now download and run trinity in about 20 seconds if you already have snap installed. I think I'll play with building an AppImage next.

jannikluhn:
[...]

PR #1205, opened 2019-09-27
Add a Github action for building app images upon every push

lithp:
### What was wrong?

It was previously relatively difficult to start using Trinity, as someone unfamiliar with how to install Python packages.

Part of #1158, the thread on Trinity packaging.

### How was it fixed?

This PR adds a github action which, for every push, builds two AppImages (trinity, and trinity-beacon) which bundle up Trinity as of that push and uploads the result as an artifact: https://github.com/lithp/trinity/commit/cb1216f70be40185da317f0a1bafd20dbe2d83cd/checks

### Questions and Future Work

- [x] Is it okay to include binaries in the repo like this? Maybe the script should instead download the binaries every time? I could probably put them into IPFS somehow, so the script which builds knows it's always using the correct version.
- [ ] Maybe adding a Github Action isn't what we want, and instead we'd prefer for the script which builds appimages to be something that's manually run during the release process? As is, I think the release process would involve taking a recent artifact and uplaoding it to the release.
- [ ] I've tested that these work on fresh ubuntu (18.04) and fedora (30) virtual machines, but I haven't tested on older versions of either OS, or on other OS's such as Arch Linux.
- [ ] This is grossly inefficient, it downloads and builds python from scratch every time. There has to be a way to cache some of the intermediate results, making each build go a lot faster.
- [x] Github build artifacts are actually zip files, so there's an opportunity to add a README to the artifact.

- [ ] Clean up commit history

#### Cute Animal Picture

tiny-turtle


pipermerriam:
Just tried the one you linked.  I'm running Pop-os which is System76's version of Ubuntu that they ship.

```shell
   ERROR  2019-09-27 16:59:19,199  ComponentManagerService  Unexpected error in <trinity.event_bus.ComponentManagerService object at 0x7f609f9090b8>, exiting
  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/site-packages/sqlalchemy/dialects/sqlite/pysqlite.py", line 338, in dbapi
    from pysqlite2 import dbapi2 as sqlite
ModuleNotFoundError: No module named 'pysqlite2'

  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/site-packages/p2p/service.py", line 117, in run
    await self._run()
  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/site-packages/trinity/event_bus.py", line 61, in _run
    self.component_manager.prepare(self._boot_info)
  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/site-packages/trinity/extensibility/component_manager.py", line 66, in prepare
    component.ready(self.event_bus_endpoint)
  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/site-packages/trinity/extensibility/component.py", line 123, in ready
    self.on_ready(manager_eventbus)
  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/site-packages/trinity/components/builtin/network_db/component.py", line 132, in on_ready
    get_tracking_database(get_networkdb_path(self.boot_info.trinity_config))
  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/site-packages/trinity/db/orm.py", line 100, in get_tracking_database
    session = _get_session(db_path)
  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/site-packages/trinity/db/orm.py", line 52, in _get_session
    engine = create_engine(database_uri)
  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/site-packages/sqlalchemy/engine/__init__.py", line 456, in create_engine
    return strategy.create(*args, **kwargs)
  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/site-packages/sqlalchemy/engine/strategies.py", line 87, in create
    dbapi = dialect_cls.dbapi(**dbapi_args)
  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/site-packages/sqlalchemy/dialects/sqlite/pysqlite.py", line 343, in dbapi
    raise e
  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/site-packages/sqlalchemy/dialects/sqlite/pysqlite.py", line 341, in dbapi
    from sqlite3 import dbapi2 as sqlite  # try 2.5+ stdlib name.
  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/sqlite3/__init__.py", line 23, in <module>
    from sqlite3.dbapi2 import *
  File "/tmp/.mount_TrinitfDTfxv/usr/python/lib/python3.7/sqlite3/dbapi2.py", line 27, in <module>
    from _sqlite3 import *
ModuleNotFoundError: No module named '_sqlite3'
```

appimage/trinity.desktop
```diff
@@ -0,0 +1,6 @@
+[Desktop Entry]
+Type=Application
+Name=Trinity
+Exec=env PYTHONWARNINGS=ignore::DeprecationWarning usr/bin/python usr/bin/trinity
```
carver: Another place that squashes warnings

lithp:
Ahhh, that's frustrating. Okay, thanks for telling me, I'll look into it!

appimage/AppRun.trinity
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+"export" "PYTHONWARNINGS=ignore::DeprecationWarning"
```
carver: I know the warnings are ugly, but I think it's better not to silence them because:
1. Users are implicitly reminded that Trinity is still in alpha
2. It serves as a TODO to actually fix the underlying problems
3. We will notice faster when new problems show up

.github/workflows/build-appimage.yaml
```diff
@@ -0,0 +1,32 @@
...
+          fetch-depth: 1
+      - run: |
+          cd $GITHUB_WORKSPACE/appimage
+          ROOT=$(dirname $(pwd))
+          PIP_REQUIREMENTS="$ROOT" ./linuxdeploy-x86_64.AppImage --plugin python --appdir AppDir -i triceratops.png -d trinity.desktop --custom-apprun=AppRun.trinity
+          ./appimagetool-x86_64.AppImage AppDir Trinity-x86_64.AppImage
+          mkdir $GITHUB_WORKSPACE/trinity-appimage
+          mv Trinity-x86_64.AppImage $GITHUB_WORKSPACE/trinity-appimage
+
+          cp AppRun.trinity-beacon AppDir/AppRun
```
carver: I was expecting a little more commonality between the two builds (with and without beacon). Why do we have to move the AppRun file here? Maybe worth a comment or two explaining what's going on and why, during this `run`. Especially where/why the two builds aren't the same.

appimage/AppRun.trinity
```diff
@@ -0,0 +1,18 @@
...
+"exec" "$(dirname $(readlink -f ${0}))/usr/bin/python3.7" "$0" "$@"
```
carver: This whole file is a super weird construction, and could definitely use a comment explaining what's going on. ie~ Why does this file look like both a shell file and a python file?

carver (commented):
> * Is it okay to include binaries in the repo like this? Maybe the script should instead download the binaries every time? I could probably put them into IPFS somehow, so the script which builds knows it's always using the correct version.

Yeah, I'd rather avoid adding several megabytes worth of binaries if we can. Download (with checksum) or IPFS sounds promising. :+1:

appimage/AppRun.trinity
lithp: Yeah, good point, I should definitely add some comments here. It's actually both a shell file *and* a python file!

lithp:
@pipermerriam okay, that was a stupid mistake. I've fixed the problem and tested again that it works in a fresh ubuntu vm. Try this new version: https://github.com/lithp/trinity/commit/5cd7ba2a7529a48bd47d55fedf628ef57e2b48d2/checks

carver:
According to https://github.com/ethereum/trinity/graphs/commit-activity we are doing 50-100 commits a week. So ~10GB per week of new AppImage files built (not counting if this builds on every PR push). Yeah, someone is going to come knocking sooner or later to tell us to tone it down. X)

pipermerriam:
@lithp seems to be working :sparkles: 

Very cool to have a binary we can distribute.  How would someone install this and have it available globally on the command line?  Are there established conventions for this?  Along the same line, is there any established convention for streamlining the installation to a single command, after which you have a globally runnable CLI `trinity` command?

pipermerriam:
As for distribution, it should be reasonably trivial to push this data into IPFS/Swarm

For IPFS we should coordinate with @njgheorghita on running a server to persist this data (I think the berlin office has a dedicated IPFS node as well).

My preference would be to use Swarm but I'm not sure what the state of persisting assets is (last I checked they didn't have asset pinning).

appimage/AppRun.trinity
carver: Just noting our in-person convo. I can get on-board with silencing these warnings in the binary, if we add a big, tall, colored banner at the top of the stdout logs, that says something like: "This is in alpha. There are bugs. Do not use for any production use cases."

.github/workflows/build-appimage.yaml
```diff
@@ -12,14 +12,27 @@ jobs:
...
-          PIP_REQUIREMENTS="$ROOT" ./linuxdeploy-x86_64.AppImage --plugin python --appdir AppDir -i triceratops.png -d trinity.desktop --custom-apprun=AppRun.trinity
-          ./appimagetool-x86_64.AppImage AppDir Trinity-x86_64.AppImage
+          PIP_REQUIREMENTS="$ROOT" ./bin/linuxdeploy-x86_64.AppImage --plugin python --appdir AppDir -i triceratops.png -d trinity.desktop --custom-apprun=AppRun.trinity
+
+          # Build the first appimage
+          ./bin/appimagetool-x86_64.AppImage AppDir Trinity-x86_64.AppImage
           mkdir $GITHUB_WORKSPACE/trinity-appimage
           mv Trinity-x86_64.AppImage $GITHUB_WORKSPACE/trinity-appimage
 
+          # Copy over trinity-beacon's launcher, and build the second appimage
```
carver: Ah, this is replacing the `AppRun` file that was created during the last build? I had thought `AppRun` was a folder. So the idea is that the first build creates some artifacts, and you replace one of those artifacts, before rebuilding?

.github/workflows/build-appimage.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
           tar -xvf ipget_v0.5.0_linux-amd64.tar.gz
```
pipermerriam: should we maybe include a checksum here to verify that we got the file we think we got?

.github/workflows/build-appimage.yaml
lithp: Yeah! It's a little confusing, The `AppDir` is a directory which is compressed and turned into an AppImage, it contains an `AppRun`. When an AppImage is run the image is decompressed back into an `AppDir`, and the `AppRun` invoked.

Here what I'm doing is building an `AppDir`. At first it has an `AppRun` which launches Trinity. After the AppImage is built, I'm then replacing the `AppDir`s `AppRun` with one which launches trinity-beacon, and building a second AppImage. The two appimages are nearly identical, I'm just changing which module is imported and then run.

.github/workflows/build-appimage.yaml
lithp: Yep, pushed a new version! `ipget` kept timing out on me, it can take up to 10 minutes to find enough peers to finally find the file, but I've now got a version working which downloads from Cloudflare's gateway and checks the checksum.

appimage/AppRun.trinity
lithp: Okay, I've added a nice fat banner:

Selection_029

.github/workflows/build-appimage.yaml
lithp: I've added a larger comment here, let me know if you think it should say even more!

appimage/AppRun.trinity
lithp: Comments added

lithp:
> Very cool to have a binary we can distribute. How would someone install this and have it available globally on the command line? Are there established conventions for this? Along the same line, is there any established convention for streamlining the installation to a single command, after which you have a globally runnable CLI trinity command?

I've added a small README which gets copied into the build artifact: https://github.com/lithp/trinity/blob/lithp/appimage/appimage/README

I *think* the convention is to copy it into `/usr/local/bin`, but I'm only 60% on that. I'm not sure who I expect the median user of these images to be, but I expect they're technical enough that we can figure out installation later.

lithp:
Test failure is spurious:https://github.com/ethereum/trinity/pull/1212

cburgdorf:
This is really cool! Skimming through the changes I can definitely see how this was a tough battle to fight! Thanks for working through this!

.github/workflows/build-appimage.yaml
carver: > After the AppImage is built, I'm then replacing the `AppDir`'s `AppRun` with one which launches trinity-beacon, and building a second AppImage.

:+1: Something like that ^ is what I'd like to see in the comment.

lithp: What do you think of the comment which is there now?

https://github.com/ethereum/trinity/pull/1205/files#diff-5f993e76238179028392d85559e6fe50R54

cburgdorf:
We might want to reactivate this for when we get into beta.

PR #1207, opened 2019-09-30
Handle find node requests for remote ENRs

jannikluhn:
Note: this includes two commits from #1204 

### What was wrong?

We only handled find node requests for our own ENR, but the majority will of course be for ENRs of other nodes.

### How was it fixed?

Update the existing handler for this case. As a temporary shortcut, it only sends one message per ENR instead of bunching them.

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/routing_table_manager.py
+++ b/p2p/discv5/routing_table_manager.py
@@ -1,5 +1,8 @@
 import logging
 import time
+from typing import (
+    Optional,
+)
 
 from eth_utils import (
     encode_hex,
@@ -16,6 +19,9 @@ from mypy_extensions import (
 
 from p2p.trio_service import (
     Service,
+)
+from p2p.trio_utils import (
+    gather,
 )
 
 from p2p.discv5.abc import (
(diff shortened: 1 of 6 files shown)


p2p/trio_utils.py
```diff
@@ -29,14 +28,9 @@
         async_fn_and_args = async_fns_and_args[index]
         if isinstance(async_fn_and_args, Iterable):
             async_fn, *args = async_fn_and_args
-        elif asyncio.iscoroutinefunction(async_fn_and_args):
```
jannikluhn: I had to remove this check as it doesn't work for all cases (e.g., `ReceiveChannel.receive`, see also https://github.com/python-trio/trio/issues/635). Fortunately I don't think it matters much, we'll still get a type error if we `await` something we shouldn't await.

p2p/discv5/routing_table_manager.py
```diff
@@ -315,8 +354,8 @@ def __init__(self,
         while True:
             deadline = trio.current_time() + ROUTING_TABLE_PING_INTERVAL
 
-            if len(self.routing_table) > 0:
-                node_id = self.routing_table.get_oldest_entry()
+            if not self.routing_table.is_empty:
```
pipermerriam: Maybe add a secondary api `RoutingTable.has_capacity` or `RoutingTable.is_not_empty` to avoid having to `not` prefix things like this?

PR #1208, opened 2019-09-30
"Every" async generator

jannikluhn:
This implements another trio utility function: The `every` generator as suggested [here](https://github.com/ethereum/trinity/pull/1028#discussion_r320814270). I'll probably use it for implementing the lookup service.

Basic usage example:

```Py
async for time in every(2, offset=2):
    print(time, trio.current_time())
    await trio.sleep(random.randint(0, 2))
```

outputting

```
200762.82529603032 200762.82812670834
200764.82529603032 200764.82748336735
200766.82529603032 200766.82880680333
200768.82529603032 200768.8274904043
```

One can also send a value back to the generator to delay the next timestamp (as well as all future ones). I think this might be useful in some cases, e.g. if one loop iteration took too long and we don't want to continue immediately but sleep a little, even though that's not allowed according to the original schedule.

#### Cute Animal Picture

### DIFF
--- a/p2p/trio_utils.py
+++ b/p2p/trio_utils.py
@@ -1,10 +1,13 @@
 import asyncio
+import itertools
 import operator
 from typing import (
     Any,
+    AsyncGenerator,
     Awaitable,
     Callable,
     Iterable,
+    Optional,
     Tuple,
     Union,
 )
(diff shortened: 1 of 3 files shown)


p2p/trio_utils.py
```diff
@@ -47,3 +50,31 @@
 
     indices_and_results_sorted = sorted(indices_and_results, key=operator.itemgetter(0))
     return tuple(result for _, result in indices_and_results_sorted)
+
+
+async def every(interval: float, offset: float = 0) -> AsyncGenerator[float, Optional[float]]:
```
pipermerriam: `offset` might be better named `initial_delay`

p2p/trio_utils.py
```diff
@@ -47,3 +50,31 @@
...
+    The generator will attempt to yield at a sequence of target times, defined as
+    `start_time + offset + N * interval` seconds where `start_time` is trio's current time at
+    instantiation of the generator and `N` starts at `0`. The target time is also the value that
+    is yielded.
+
+    If at a certain iteration the target time has already passed, the generator will yield
+    immediately (with a checkpoint in between). The yield value is still the target time.
+
+    The generator accepts an optional send value which acts as an additional offset that is added
+    for all future timestamps.
```
pipermerriam: IIUC would the following wording be accurate?

> The generator accepts an optional send value which will delay the next iteration of the generator by that amount.

I don't know if that wording is any clearer, but I think the *delay* terminology might be easier to understand than *offset*

p2p/trio_utils.py
```diff
@@ -47,3 +50,31 @@
...
+    """
+    start_time = trio.current_time()
+    yield_times_without_offset = (
+        start_time + interval * iteration for iteration in itertools.count()
```
pipermerriam: Nice use of a generator

tests-trio/p2p-trio/test_trio_utils.py
```diff
@@ -42,3 +43,57 @@
         (return_args, 1, 2, 3)
     )
     assert results == ((), (), (1, 2, 3))
+
+
+@pytest.mark.trio
+async def test_every(autojump_clock):
```
pipermerriam: `autojump_clock` is pretty cool.

tests-trio/p2p-trio/test_trio_utils.py
```diff
@@ -42,3 +43,57 @@
...
+    assert first_time <= trio.current_time()
+    assert first_time == pytest.approx(start_time + 1)
+
+    second_time = await every_generator.__anext__()
+    assert second_time == pytest.approx(trio.current_time())
+    assert second_time == pytest.approx(first_time + 2)
+
+    third_time = await every_generator.__anext__()
+    assert third_time == pytest.approx(trio.current_time())
+    assert third_time == pytest.approx(first_time + 4)
```
pipermerriam: I did not know about the `pytest.approx` API.  nice!

p2p/trio_utils.py
pipermerriam: I just realized you should be able to do this with `itertools.count(start_time, interval)` as the [`itertools.count`](https://docs.python.org/3/library/itertools.html#itertools.count) API takes both a `start` and `step` value.

PR #1212, opened 2019-10-02
Make flakey test more permissive

lithp:
### What was wrong?

`tests/p2p/test_token_bucket.py::test_token_bucket_initial_tokens` was a little flakey, I ran it 10k times (while compiling rust to simulate a bursty load on the CI box) and it failed 8 times. As a quick solution I've made the test more permissive and it hasn't failed on me again.

A better solution might be to change the test. I think it's making sure that we never sleep when pulling tokens from a full bucket, maybe this test could mock out the time api, and then check that `sleep()` is never called.

#### Cute Animal Picture

dog-play-snow

### DIFF
--- a/tests/p2p/test_token_bucket.py
+++ b/tests/p2p/test_token_bucket.py
@@ -35,9 +35,9 @@ async def test_token_bucket_initial_tokens():
     # since the bucket starts out full the loop
     # should take near zero time
     expected = await measure_zero(10)
-    # drift is allowed to be up to 200% since we're working with very small
+    # drift is allowed to be up to 1000% since we're working with very small
     # numbers.
-    assert_fuzzy_equal(delta, expected, allowed_drift=2)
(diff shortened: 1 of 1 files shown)


lithp:
Hmm, even with this change it fails: 

Selection_031

lithp:
`TokenBucket` eventually has to make the migration to trio, so maybe `asyncio.sleep()` could be injected through the constructor. This would make testing it much easier, especially under trio: https://trio.readthedocs.io/en/stable/reference-testing.html#time-and-timeouts

lithp:
Example test failure: https://circleci.com/gh/ethereum/trinity/129012

lithp:
A test flaked on a PR meant to fix a different flaking test :roll_eyes: 

Opened a ticket: https://github.com/ethereum/trinity/issues/1213

PR #1214, opened 2019-10-02
Use minimal config and fix finality check

NIC619:
### What was wrong?
finality is not reached in `test_demo.py`

### How was it fixed?
Use minimal config and fix finality check in `test_demo.py`

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/tests/eth2/integration/test_demo.py
+++ b/tests/eth2/integration/test_demo.py
@@ -7,7 +7,7 @@ from eth2.beacon.helpers import compute_epoch_of_slot
 from eth2.beacon.operations.attestation_pool import AttestationPool
 from eth2.beacon.state_machines.forks.serenity import SerenityStateMachine
 from eth2.beacon.state_machines.forks.serenity.blocks import SerenityBeaconBlock
-from eth2.beacon.state_machines.forks.serenity.configs import SERENITY_CONFIG
+from eth2.beacon.state_machines.forks.skeleton_lake import MINIMAL_SERENITY_CONFIG
 from eth2.beacon.tools.builder.initializer import create_mock_genesis
 from eth2.beacon.tools.builder.proposer import create_mock_block
 from eth2.beacon.tools.builder.validator import create_mock_signed_attestations_at_slot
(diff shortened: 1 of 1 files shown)


tests/eth2/integration/test_demo.py
```diff
@@ -106,5 +97,5 @@ def test_demo(base_db, validator_count, keymap, pubkeys, fork_choice_scoring):
     assert state.slot == chain_length + genesis_slot
 
     # Justification assertions
-    assert state.current_justified_checkpoint.epoch == genesis_epoch
-    assert state.finalized_checkpoint.epoch == genesis_epoch
+    assert state.current_justified_checkpoint.epoch == 3 + genesis_epoch
+    assert state.finalized_checkpoint.epoch == 2 + genesis_epoch
```
NIC619: I increase them to (`3 + genesis_epoch` and `2 + genesis_epoch`) instead of (`2 + genesis_epoch` and `1 + genesis_epoch`) is because in epoch 4, `state.current_justified_checkpoint.epoch == 2` while `state.finalized_checkpoint.epoch == 0`. Not sure if this is just one of the edge cases around genesis epoch or something is off.

hwwhww: Could you also add `assert` for `state.previous_justified_checkpoint`? Thanks.

ralexstokes: i think you mean in epoch 2, rather than in epoch 4 in this comment

tests/eth2/integration/test_demo.py
```diff
@@ -22,16 +21,7 @@ def fork_choice_scoring():
...
-    config = SERENITY_CONFIG._replace(
-        SLOTS_PER_EPOCH=slots_per_epoch,
-        GENESIS_EPOCH=compute_epoch_of_slot(
-            SERENITY_CONFIG.GENESIS_SLOT, slots_per_epoch
-        ),
-        TARGET_COMMITTEE_SIZE=3,
-        SHARD_COUNT=2,
-        MIN_ATTESTATION_INCLUSION_DELAY=2,
-    )
+    config = MINIMAL_SERENITY_CONFIG
```
ralexstokes: so simple!

ralexstokes:
for the record, we do not check any justification and finality during epochs (written as offsets from `GENESIS_EPOCH`) 0 or 1. we do check for justification and finality from epoch 2 onward.

in epoch 2, we justify the current epoch. this epoch is in fact finalized but we do not recognize it in the protocol due to an artifact of the construction of the genesis state (using the "zero" value for `Checkpoint` type).

in epoch 3, we have the previous justified checkpoint at the prior current justified checkpoint (so `GENESIS_EPOCH + 2`) and we justify this current epoch. we check finality here and see that we finalize the prior justified checkpoint at epoch 2.

given the way we handle epoch processing (i.e. process a given epoch at the start of the next epoch), we need to transition through `4 * SLOTS_PER_EPOCH` worth of slots to include the processing of epoch 3.

tests/eth2/integration/test_demo.py
pipermerriam: I think these deserve comments explaining these offsets, maybe even with something like an ASCII diagram?  Not that I understand exactly what is going on here but it seems like it's the type of code that will take some brain space to understand why these numbers are used.

ralexstokes: good call, i worked thru this the other day to make sure these numbers followed the spec; you can see the rationale here:

https://github.com/ethereum/trinity/pull/1214#issuecomment-546184080

ralexstokes: agreed that a content-rich comment (graphics etc) would be nice :)

PR #1215, opened 2019-10-02
Method to request ENRs

jannikluhn:
When requesting ENRs (e.g. with `FindNode` or in the future `TOPICQUERY`), the response will consist not of a single, but of a sequence of `Nodes` messages (number given by the `total` attribute). This PR adds a convenience method of waiting for all individual messages and just return the list of all ENRs.

I've added the `aiostream` library as a dependency, mostly so that I can easily `take` the first N response messages (I've noticed I never merged #11 for some reason and `aiostream` was mentioned as possible alternative). It has a single dependency which is just a backport of some Python 3.7 method.

I've also added a custom exception class `InvalidResponse`. Most likely I should have started with this sooner, but here I didn't find a reason not to.

#### Cute Animal Picture

### DIFF
--- a/tests-trio/p2p-trio/test_message_dispatcher.py
+++ b/tests-trio/p2p-trio/test_message_dispatcher.py
@@ -4,6 +4,10 @@ import pytest_trio
 
 import trio
 
+from eth_utils.toolz import (
+    take,
+)
+
 from p2p.trio_service import (
     background_service,
 )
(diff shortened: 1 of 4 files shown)


tests-trio/p2p-trio/test_message_dispatcher.py
```diff
@@ -184,3 +193,101 @@ def outgoing_message_channels():
...
+                            remote_enr,
+                            remote_endpoint,
+                            incoming_message_channels,
+                            outgoing_message_channels,
+                            nursery,
+                            ):
+    request_id = message_dispatcher.get_free_request_id(remote_enr.node_id)
+    request = FindNodeMessageFactory(request_id=request_id)
+
+    enrs = tuple(ENRFactory() for _ in range(3))
```
pipermerriam: You can use `ENRFactory.create_batch(3)` to do this.

Issue #1213, opened 2019-10-02
Test failure: test_web3_commands_via_attached_console


lithp:
This one happens all the time. Example failure: https://circleci.com/gh/ethereum/trinity/129067

Another failure: https://circleci.com/gh/ethereum/trinity/128660

lithp:
https://circleci.com/gh/ethereum/trinity/129067 has an interesting cause. I think `trinity` starts downloading headers too quickly! The test assumes that `w3.eth.getBlock('latest')` will return the genesis. However, if the trinity node is too quick then it will have already downloaded some headers, and the test will fail because it waits forever, looking for the genesis hash.

cburgdorf:
Ha, that's an easy fix! We may just ask for `w3.eth.getBlock(0)` (or is it `0x0`?)  instead.

lithp:
After I fixed that problem I ran into a few more problems. My current theory is that:
- Something is happening which takes up a LOT of cpu, so our 40 second timeout isn't long enough. Bumping the timeout to 60 seconds made the test pass more often.
- `AsyncProcessRunner` attaches `stderr` and `stdout` to pipes, and the test stops reading from the pipes as soon as `contains_all` returns. My theory is that the pipes fill up, then the trinity process hangs and stops sending messages to the `trinity attach` process. I added a coro which continually reads from stdout/stderr but `attached_trinity.expect()` blocks the event loop.

cburgdorf:
I think this issue has been addressed (at least, the test job appears to be stable)

PR #1216, opened 2019-10-02
Reduce spurious web3 attach failures

lithp:
### What was wrong?

[test_web3_commands_via_attached_console fails a lot](https://github.com/ethereum/trinity/issues/1213)

### How was it fixed?

1. If the node has already synced a few headers before we get a chance to send it rpc requests, we accept the answers it sends us.
2. Use a random port for each run, not just 30303. If Trinity failed to shutdown from a previous test run later tests don't fail.

#### Cute Animal Picture

sad-cat-luhu

### DIFF
--- a/tests/integration/test_trinity_cli.py
+++ b/tests/integration/test_trinity_cli.py
@@ -136,16 +136,20 @@ async def test_expected_logs_for_light_mode(command):
 async def test_web3_commands_via_attached_console(command,
                                                   expected_network_id,
                                                   expected_genesis_hash,
-                                                  xdg_trinity_root):
+                                                  xdg_trinity_root,
+                                                  unused_tcp_port):
 
     command = tuple(
         fragment.replace('{trinity_root_path}', str(xdg_trinity_root))
         for fragment
         in command
     )
+    # use a random port each time, in case a previous run went awry and left behind a
+    # trinity instance
+    command += (f'--port={str(unused_tcp_port)}',)
     attach_cmd = list(command[1:] + ('attach',))
 
-    async with AsyncProcessRunner.run(command, timeout_sec=40) as runner:
+    async with AsyncProcessRunner.run(command, timeout_sec=120) as runner:
         assert await contains_all(runner.stderr, {
             "Started DB server process",
             "Component started: Sync / PeerPool",
@@ -165,9 +169,9 @@ async def test_web3_commands_via_attached_console(command,
             attached_trinity.expect(f"'{expected_network_id}'")
             attached_trinity.sendline("w3")
             attached_trinity.expect("web3.main.Web3")
-            attached_trinity.sendline("w3.eth.getBlock('latest').number")
+            attached_trinity.sendline("w3.eth.getBlock(0).number")
             attached_trinity.expect(str(GENESIS_BLOCK_NUMBER))
-            attached_trinity.sendline("w3.eth.getBlock('latest').hash")
+            attached_trinity.sendline("w3.eth.getBlock(0).hash")
             attached_trinity.expect(expected_genesis_hash)
         except pexpect.TIMEOUT:
             raise Exception("Trinity attach timeout")


lithp:
Welp, that fixed some problems but there are more: https://circleci.com/gh/ethereum/trinity/129422

lithp:
Okay, that didn't fix it, and I spent a few more hours trying to get this test passing but then I started getting segfaults and I gave up :see_no_evil: Marking the test as skip for now, we're wasting a lot of time re-running the tests.

tests/integration/test_trinity_cli.py
```diff
@@ -166,9 +169,9 @@
             attached_trinity.sendline("w3")
             attached_trinity.expect("web3.main.Web3")
             attached_trinity.sendline("w3.eth.getBlock('latest').number")
-            attached_trinity.expect(str(GENESIS_BLOCK_NUMBER))
+            attached_trinity.expect("[0-9]+")
             attached_trinity.sendline("w3.eth.getBlock('latest').hash")
```
cburgdorf: What do you think about asking for block `0` instead of `latest` so that we have a fixed target and don't have to relax our expectations?

cburgdorf:
>Marking the test as skip for now, we're wasting a lot of time re-running the tests.

I'm really torn about this. I get the frustration with the failing tests and I don't have a solution either. It's just that disabling it reminds me of the [broken window theory](https://en.wikipedia.org/wiki/Broken_windows_theory) and that it might be even more frustrating to find out we've shipped a Trinity version where the web3 console might be broken.

I personally would favor if we take the quick wins such as increasing the timeout, using a fixed target for the `getBlock` and maybe the random port selection and enjoy a few less CI failures.

I also think it's still under-explored what ways we might have to automatically re-run failed tests once to further reduce manual re triggering of these tests.

tests/integration/test_trinity_cli.py
lithp: Sure, that makes sense too!

lithp:
> It's just that disabling it reminds me of the broken window theory and that it might be even more frustrating to find out we've shipped a Trinity version where the web3 console might be broken.

I agree that disabling this test is not ideal! And I think someone would come back to this test and really fix is soon. But, I don't think that person is me, when there's so little time left before Osaka. I see a different kind of broken window here, we have a page of PRs and most of them are red so each of us spends a lot of our time re-running tests. The web3 console is unlikely to break and these spurious failures are definitely wasting our time, the cost/benefit of running this test is not in alignment.

My hope is that skipping the test lets us win back a smoother development process, and after Osaka one of us can try again to make the test work.

> I personally would favor if we take the quick wins such as increasing the timeout, using a fixed target for the getBlock and maybe the random port selection and enjoy a few less CI failures.

All of these are nice but they take the test from failing all the time to failing a lot, there's still a core problem that we're papering over by doing them, and I don't think they win enough successful builds that we'll have a page of PRs with a bunch of green check marks.

> I also think it's still under-explored what ways we might have to automatically re-run failed tests once to further reduce manual re triggering of these tests.

I don't think that saves us much time here. If the tests have to be re-run then this test suite takes a very long time to run, if the tests are run twice or three times we'll still have to wait 15 or 30 minutes for them to finish! I think re-running flakey tests is covering up a real problem that should be solved.

pipermerriam:
A few quick brain dumps.

- I lean towards maintaining status quo of re-running failed builds.  It's unfortunate but it keeps this on our radar.
- I think the *real* fix for these integration test issues is going to require us to branch beyond CircleCI.  Expecting the client to run well on a VPS node with limited CPU and poor disk performance isn't reasonable.  I think that we need an environment that allows us to run on a beefier VPS (Jenkins?).

My experience has been that most integration failures tend to pass on the second run and that while annoying, it isn't overly disruptive to getting things done.  My preferred way forward is to improve things in their current state and to prioritize putting in the work to get a better environment available for real integration testing.

lithp:
Okay, I'm okay with being outvoted on this. I spent some more time on this today but still wasn't able to find a nice solution. I've unskipped the test and bumped the timeout and it doesn't fail on me quite as often now, ready for review!

tests/integration/test_trinity_cli.py
cburgdorf: Mind updating this as well before it gets merged?

cburgdorf:
I just came across another CI failure and was reminded of this issue. I was so free to amend the final tweak (changing from `getBlock('latest')` to `getBlock(0)` to avoid a moving target) and will merge it in once it goes green.

Issue #1138, opened 2019-09-16
Notes about Validator/BeaconNode Separation


hwwhww:
### TODO:
1. Implement the [Validator APIs](https://github.com/ethereum/eth2.0-APIs/blob/master/apis/validator/beacon-node-oapi.yaml) on JSON-RPC Server
	* Some notes:
		* For the query requests: use the given `chain: AsyncBeaconChainDB` to query.
		* For the broadcast request: use `EventBus` to send request to `Node` to broadcast.
2. Implement `Validator`  **Plugin**
	* Each validator has its own `SlotTicker` services
3. Final goal: Split trinity-beacon (the node) and the validator

ralexstokes:
this would be a great opportunity to look at a "pure" `trio` application -- i'd be quite curious to see the difference b/t a more 'structured' approach to concurrency and our useage of `asyncio` in the existing stuff

ralexstokes:
i'd like to look into writing the `Validator` component as a pure `trio` component -- if someone else has started on this, then let's coordinate as we will want to move in that direction eventually either way

PR #1217, opened 2019-10-04
High-level pass at validator client

ralexstokes:
This PR is intended to serve as basis for discussion for some ideas around how to integrate trio moving forward; the particulars of the validator client (the "business logic") are missing as this PR just wants to showcase the high-level design.

Some points to note:

- Uses `trio` for the concurrency backend.

- Attempts to abstract over the particulars of `trio` so that different
implementations can be provided. Unclear how useful this is.

- Structures high-level components as single-purpose "tasks" that communicate
by sharing over a channel abstraction. Actions taken by the client are driven
by events/data moving through the channel construction.

Discussion in https://github.com/ethereum/trinity/issues/1138 is relevant.

#### Cute Animal Picture

### DIFF
--- a/validator_client/client.py
+++ b/validator_client/client.py
@@ -1,0 +1,51 @@
+import logging
+
+from async_service.base import Service
+
+from validator_client.beacon_node import MockBeaconNode as BeaconNode
+from validator_client.clock import Clock
+from validator_client.context import Context
+from validator_client.duty_executor import duty_executor
+from validator_client.duty_scheduler import DutyScheduler
+from validator_client.duty_store import DutyStore
+from validator_client.publisher import publisher
+from validator_client.signatory import Signatory
+from validator_client.streams.tee import Tee
+
+logger = logging.getLogger("validator_client.client")
+logger.setLevel(logging.DEBUG)
+
+
+class Client(Service):
+    def __init__(self, context: Context) -> None:
+        self._context = context
+
+    async def run(self) -> None:
+        logger.info("booting client from the provided context...")
+
+        clock = Clock(self._context)
+        clock_tee = Tee(clock.stream_ticks())
+        duty_store = DutyStore(self._context)
+        beacon_node = BeaconNode(self._context)
+
+        duty_scheduler = DutyScheduler(
+            self._context, await clock_tee.clone(), duty_store, beacon_node
+        )
+        duty_provider = duty_executor(
+            self._context, await clock_tee.clone(), duty_store, beacon_node
+        )
+        signatory = Signatory(self._context, duty_provider)
+
+        try:
+            self.manager.run_daemon_task(clock_tee.run)
+            # TODO: ``publisher`` should run as a daemon task
+            # see: https://github.com/ethereum/async-service/issues/51
+            self.manager.run_task(publisher, self._context, signatory, beacon_node)
+            ds = self.manager.run_child_service(duty_scheduler)
+            ss = self.manager.run_child_service(signatory)
+
+            await ds.wait_finished()
+            await ss.wait_finished()
+            await self.manager.wait_stopping()
+        finally:
+            logger.info("shutting down client...")
--- a/validator_client/clock.py
+++ b/validator_client/clock.py
@@ -1,0 +1,97 @@
+from dataclasses import dataclass, field
+import logging
+import time
+from typing import Callable, Tuple
+
+from cached_property import cached_property
+import trio
+
+from eth2.beacon.typing import Epoch, Slot
+from validator_client.context import Context
+
+logger = logging.getLogger("validator_client.clock")
+logger.setLevel(logging.DEBUG)
+
+TICKS_PER_SLOT = 2
+
+
+@dataclass(eq=True, frozen=True)
+class Tick:
+    t: float = field(repr=False)
+    slot: int
+    epoch: int
+    count: int  # non-negative counter for the tick w/in a slot
+
+    @cached_property
+    def first_in_slot(self) -> bool:
+        return self.tick == 0
+
+
+def _get_unix_time():
+    return time.time()
+
+
+TimeProvider = Callable[[], float]
+
+
+class Clock:
+    def __init__(
+        self, context: Context, time_provider: TimeProvider = _get_unix_time
+    ) -> None:
+        self._context = context
+        self._time_provider = time_provider
+        self._tick_interval = self._compute_tick_interval()
+
+    def _compute_tick_interval(self) -> int:
+        """
+        The validator client needs to take action with sub-slot granularity.
+        Each period on this finer resolution is called a "tick".
+        """
+        return self._context.seconds_per_slot // TICKS_PER_SLOT
+
+    def _compute_slot_and_alignment(self, t: int) -> Tuple[int, bool]:
+        """
+        Compute the slot corresponding to a unix time ``t``.
+        Indicate if ``t`` corresponds to the first tick in the relevant slot
+        with a boolean return value.
+        """
+        time_since_genesis = t - self._context.genesis_time
+        slot, sub_slot = divmod(time_since_genesis, self._context.seconds_per_slot)
+        return slot, sub_slot == 0
+
+    def _compute_epoch(self, slot: Slot) -> Epoch:
+        return Epoch(slot // self._context.slots_per_epoch)
+
+    def _compute_tick(self) -> Tick:
+        t = self._time_provider()
+        slot, aligned = self._compute_slot_and_alignment(int(t))
+        epoch = self._compute_epoch(slot)
+        count = int(not aligned)
+        return Tick(t, slot, epoch, count)
+
+    async def _wait_for_genesis(self) -> None:
+        t = self._time_provider()
+        genesis_time = self._context.genesis_time
+        if t >= genesis_time:
+            return
+
+        logger.info(
+            "genesis time has not elapsed; blocking client until %s", genesis_time
+        )
+        duration = genesis_time - t
+        await trio.sleep(duration)
+
+    async def _wait_until_next_tick(self, tick: Tick) -> None:
+        next_tick_at = tick.t + self._tick_interval
+        t = self._time_provider()
+        duration = next_tick_at - t
+        if duration >= 0:
+            await trio.sleep(duration)
+
+    async def stream_ticks(self) -> None:
+        await self._wait_for_genesis()
+
+        while True:
+            tick = self._compute_tick()
+            yield tick
+            await self._wait_until_next_tick(tick)
--- a/validator_client/context.py
+++ b/validator_client/context.py
@@ -1,0 +1,53 @@
+from cached_property import cached_property
+from eth_typing import BLSPubkey
+
+from validator_client.key_store import MockKeyStore as KeyStore
+
+
+class Context:
+    """
+    Represents data specific to a particular instance of a ``Client``.
+    """
+
+    def __init__(self, config):
+        self._config = config
+        self._key_store = KeyStore(config)
+
+    @classmethod
+    def from_config(cls, config):
+        return cls(config)
+
+    @cached_property
+    def config(self):
+        return self._config
+
+    @cached_property
+    def genesis_time(self):
+        return self._config["genesis_time"]
+
+    @cached_property
+    def seconds_per_slot(self):
+        return self._config["seconds_per_slot"]
+
+    @cached_property
+    def slots_per_epoch(self):
+        return self._config["slots_per_epoch"]
+
+    @cached_property
+    def beacon_node_endpoint(self):
+        return self._config["beacon_node_endpoint"]
+
+    @cached_property
+    def signatory_db_handle(self):
+        return self._config["signatory_db_handle"]
+
+    @cached_property
+    def validator_public_keys(self):
+        return self._key_store.public_keys
+
+    @cached_property
+    def key_store_location(self):
+        return self._key_store.location
+
+    def private_key_for(self, public_key: BLSPubkey) -> int:
+        return self._key_store.private_key_for(public_key)
--- a/validator_client/duty.py
+++ b/validator_client/duty.py
@@ -1,0 +1,54 @@
+from dataclasses import dataclass, field
+from enum import Enum, auto, unique
+
+from eth_typing import BLSPubkey
+from eth_utils import humanize_hash
+
+from eth2.beacon.signature_domain import SignatureDomain
+from eth2.beacon.typing import CommitteeIndex, DomainType, Slot
+
+
+@unique
+class DutyType(Enum):
+    Attestation = auto()
+    BlockProposal = auto()
+
+
+@dataclass(eq=True, frozen=True)
+class Duty:
+    """
+    A ``Duty`` represents some work that needs to be performed on behalf of some
+    validator, like signing an ``Attestation``.
+    """
+
+    validator_public_key: BLSPubkey
+    slot: Slot
+
+
+@dataclass(eq=True, frozen=True)
+class AttestationDuty(Duty):
+    duty_type: DutyType = field(init=False, default=DutyType.Attestation)
+    domain_type: DomainType = field(
+        init=False, default=SignatureDomain.DOMAIN_BEACON_ATTESTER
+    )
+    committee_index: CommitteeIndex
+
+    def __repr__(self) -> str:
+        return (
+            f"AttestationDuty(validator_public_key={humanize_hash(self.validator_public_key)},"
+            f" slot={self.slot}, committee_index={self.committee_index})"
+        )
+
+
+@dataclass(eq=True, frozen=True)
+class BlockProposalDuty(Duty):
+    duty_type: DutyType = field(init=False, default=DutyType.BlockProposal)
+    domain_type: DomainType = field(
+        init=False, default=SignatureDomain.DOMAIN_BEACON_PROPOSER
+    )
+
+    def __repr__(self) -> str:
+        return (
+            f"BlockProposalDuty(validator_public_key={humanize_hash(self.validator_public_key)},"
+            f" slot={self.slot})"
+        )
--- a/validator_client/duty_scheduler.py
+++ b/validator_client/duty_scheduler.py
@@ -1,0 +1,63 @@
+import logging
+from typing import AsyncIterable
+
+from async_service.base import Service
+
+from validator_client.beacon_node import MockBeaconNode as BeaconNode
+from validator_client.clock import Tick
+from validator_client.context import Context
+from validator_client.duty_store import DutyStore
+
+logger = logging.getLogger("validator_client.duty_scheduler")
+logger.setLevel(logging.DEBUG)
+
+
+def _filter_new_duties(old, new):
+    return new
+
+
+def _get_duties_for(tick, duties):
+    return duties
+
+
+async def _select_distinct_epochs(clock):
+    current_epoch = None
+    async for tick in clock:
+        if tick.epoch != current_epoch:
+            current_epoch = tick.epoch
+            yield tick
+
+
+class DutyScheduler(Service):
+    """
+    ``DutyScheduler`` manages the set of duties for the validator public keys in the key store.
+
+    This task involves polling the beacon node periodically to get the latest set of assignments
+    based on the beacon chain state.
+    """
+
+    def __init__(
+        self,
+        context: Context,
+        clock: AsyncIterable,
+        duty_store: DutyStore,
+        beacon_node: BeaconNode,
+    ) -> None:
+        self._context = context
+        self._epoch_clock = _select_distinct_epochs(clock)
+        self._validator_public_keys = context.validator_public_keys
+        self._beacon_node = beacon_node
+        self._store = duty_store
+
+    async def _process_tick(self, tick: Tick) -> None:
+        logger.info("checking for new duties in epoch %s", tick.epoch)
+        duties = await self._beacon_node.fetch_duties(
+            self._validator_public_keys, tick.epoch
+        )
+        logger.info("found duties %s in epoch %s", duties, tick.epoch)
+        # TODO manage duties correctly, accounting for re-orgs, etc.
+        await self._store.add_duties_at_tick(duties, tick)
+
+    async def run(self) -> None:
+        async for tick in self._epoch_clock:
+            await self._process_tick(tick)
--- a/validator_client/main.py
+++ b/validator_client/main.py
@@ -1,0 +1,85 @@
+import argparse
+import logging
+import signal
+from typing import Optional
+
+from async_service.trio import background_trio_service
+from eth_utils import humanize_hash
+import trio
+
+from trinity._utils.logging import LOG_FORMATTER
+from validator_client.cli_parser import parse_cli_args
+from validator_client.client import Client
+from validator_client.config import Config, config
+from validator_client.context import Context
+from validator_client.key_store import KeyStore
+
+
+def _setup_logging() -> logging.Logger:
+    logger = logging.getLogger()
+    logger.setLevel(logging.INFO)
+
+    stream_handler = logging.StreamHandler()
+    stream_handler.setLevel(logging.DEBUG)
+    stream_handler.setFormatter(LOG_FORMATTER)
+    logger.addHandler(stream_handler)
+
+    return logger
+
+
+async def _wait_for_interrupts(logger: logging.Logger) -> None:
+    with trio.open_signal_receiver(signal.SIGINT) as stream:
+        async for _ in stream:
+            logger.info("received interrupt; shutting down...")
+            return
+
+
+async def _main(
+    logger: logging.Logger, config: Config, arguments: argparse.Namespace
+) -> None:
+
+    # TODO sanity check the validator key pairs we find are valid
+    # perhaps by fetching their indices in the state and adding them to the following
+    # logging statement.
+    context = Context.from_config(config)
+    validator_public_keys = context.validator_public_keys
+    logger.info(
+        "found %d validator key pair(s) at %s for public key(s) %s",
+        len(validator_public_keys),
+        context.key_store_location,
+        tuple(map(humanize_hash, validator_public_keys)),
+    )
+    client = Client(context)
+
+    async with background_trio_service(client):
+        await _wait_for_interrupts(logger)
+
+
+async def _import_key(
+    logger: logging.Logger, config: Config, arguments: argparse.Namespace
+) -> None:
+    logger.info("importing private key...")
+    try:
+        KeyStore.import_private_key(config, arguments.private_key)
+    except Exception as e:
+        logger.warn(e)
+
+
+def _fn_for_command(command: Optional[str]):
+    return {None: _main, "import-key": _import_key}[command]
+
+
+def _set_demo_mode_in_config(config: Config, demo_mode: bool) -> None:
+    config["demo_mode"] = demo_mode
+
+
+def main() -> None:
+    # TODO:
+    # load config from CLI parser
+    # load config from file
+    # ... including logging config which can then go to `_setup_logging`
+    logger = _setup_logging()
+    arguments = parse_cli_args()
+    fn = _fn_for_command(arguments.subparser_name)
+    _set_demo_mode_in_config(config, arguments.demo_mode)
+    trio.run(fn, logger, config, arguments)
--- a/validator_client/publisher.py
+++ b/validator_client/publisher.py
@@ -1,0 +1,12 @@
+import logging
+
+from eth_utils import humanize_hash
+
+logger = logging.getLogger("validator_client.publisher")
+logger.setLevel(logging.DEBUG)
+
+
+async def publisher(context, signatory, beacon_node):
+    async for duty, signature in signatory.signatures():
+        logger.info("got signature %s for duty %s", humanize_hash(signature), duty)
+        await beacon_node.publish(duty, signature)
--- a/validator_client/signatory.py
+++ b/validator_client/signatory.py
@@ -1,0 +1,87 @@
+import logging
+from typing import AsyncIterable, Tuple
+
+from async_service.base import Service
+from eth_typing import BLSSignature
+from eth_utils import ValidationError, humanize_hash
+import trio
+
+from eth2._utils.bls import bls
+from eth2.beacon.helpers import signature_domain_to_domain_type
+from validator_client.context import Context
+from validator_client.duty import Duty
+from validator_client.signatory_db import BaseSignatoryDB, InMemorySignatoryDB
+
+logger = logging.getLogger("validator_client.signatory")
+logger.setLevel(logging.DEBUG)
+
+
+async def _validate_duty(duty: Duty, signable, db: BaseSignatoryDB) -> None:
+    """
+    ``db`` contains a persistent record of all signatures with
+    enough information to prevent the triggering of any slashing conditions.
+    """
+    if await db.is_slashable(duty, signable):
+        raise ValidationError(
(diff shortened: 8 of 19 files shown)


validator_client/client.py
```diff
@@ -0,0 +1,91 @@
...
+    def _build_tasks(self, shell):
+        context = self._context
+        component_dependencies = get_component_dependencies(context)
+
+        send_channels = {}
+        receive_channels = {}
+        for component_type in component_dependencies:
+            send_channel, receive_channel = shell.open_channel()
+            send_channels[component_type] = send_channel
+            receive_channels[component_type] = receive_channel
```
pipermerriam: This might cause problems.  All of your channels are created with the same buffer size which means that all of the channels have the same "shape", not allowing a producer to push more than that constant size of things into the channel before hitting back pressure.

My intuition says that there will be different needs for different channels and this constraint is likely to be problematic.

This *could* be addressed in the dependency configuration, but I don't think that this complexity is justified.  I'd be much happier to see all the dependencies and their channels be handled in a more explicit way (see my other comment)

validator_client/client.py
```diff
@@ -0,0 +1,91 @@
...
+
+def get_component_dependencies(_context):
+    return OrderedDict(
+        (
+            (Clock, ()),
+            (DutyScheduler, (Clock,)),
+            (Signatory, (DutyScheduler,)),
+            (Publisher, (Signatory,)),
+        )
+    )
```
pipermerriam: I can see us moving this direction eventually but I'm inclined to push back on such a design this early in the process.  I like the idea of having declarative dependencies between the various components that need to communicate, however, this introduces level of complexity and indirection that I don't think is justified at this stage.

The primary thing that this seems to complicate is understanding how all of these get wired together.  This logic is all buried in the *framework*.  I'd have a much easier time if this was all done with less framework architecture and more focus on minimal application code to make this work, after which we can move in the direction of *framework-style* abstraction as we see where the architecture is justified to solve objective problems.

validator_client/shells/trio.py
```diff
@@ -0,0 +1,121 @@
...
+logger = logging.getLogger("validator_client.shells.shell[trio]")
+logger.setLevel(logging.DEBUG)
+
+shell = sys.modules[__name__]
+
+cancelled = trio.Cancelled
+
+current_time = trio.current_time
+sleep = trio.sleep
+
```
pipermerriam: If we're going to abstract away the async backend (which I'm :-1: on) we should use `anyio` rather than rolling it ourselves.  I'm a reasonably strong -1 on this but I'm always open to someone changing my mind.

validator_client/clock.py
```diff
@@ -0,0 +1,84 @@
...
+            await shell.sleep(duration)
+
+    async def run(self, shell):
+        await self._wait_for_genesis(shell)
+
+        while True:
+            tick = self._compute_tick()
+            logger.debug("sending %s", tick)
+            await self._send(shell, tick)
+            await self._wait_until_next_tick(shell, tick)
```
pipermerriam: This seems like a good case for an async generator.

```python
async for tick in self.slot_clock():
    self.logger.debug(...)
    self._send(tick)
```

drops two lines from the loop.

validator_client/clock.py
```diff
@@ -0,0 +1,84 @@
...
+            return
+
+        logger.info(
+            "genesis time has not elapsed; blocking client until %s", genesis_time
+        )
+        duration = genesis_time - t
+        await shell.sleep(duration)
+
+    async def _wait_until_next_tick(self, shell, tick) -> None:
+        next_tick_seconds = tick.t + self._tick_interval
```
pipermerriam: *naming nitpick*

Maybe `next_tick_at`.  The name `next_tick_seconds` makes me think *number of seconds until next tick*.  `next_tick_timestamp` could also work, but I personally like `_at` better.

validator_client/clock.py
```diff
@@ -0,0 +1,84 @@
...
+    epoch: int
+    first_in_slot: bool
+
+
+def _get_unix_time():
+    return time.time()
+
+
+class Clock(Task):
+    def __init__(self, context, send_channel, time_provider=_get_unix_time) -> None:
```
pipermerriam: Might be good to use the `trio` clock for the time provider to make testing easier via the built-in auto-jump-clock

validator_client/clock.py
```diff
@@ -0,0 +1,84 @@
...
+
+TICKS_PER_SLOT = 2
+
+
+@dataclass
+class Tick:
+    t: float = field(repr=False)
+    slot: int
+    epoch: int
+    first_in_slot: bool
```
pipermerriam: *if* the `Tick` was provided with the context, I think most of these fields could be computed properties which would cleanup a lot of logic from the `Clock`.

validator_client/clock.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def _get_unix_time():
+    return time.time()
+
+
+class Clock(Task):
```
pipermerriam: Not sure about this one but unless we currently have multiple consumers of the `Tick` (which it appears we don't but I'm not 100% sure about that), maybe it makes better sense to just have a `SlotClock` class which implements the `AsyncGenerator` interface and embed that directly in place that consumes the `Tick` messages.

validator_client/context.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def __init__(self, config):
+        self.config = config
+        self.key_store = KeyStore(config)
+
+    @classmethod
+    def from_config(cls, config):
+        return cls(config)
+
+    @property
```
pipermerriam: minor performance improvement if you use `cached_property` for the ones that are cacheable.

validator_client/signatory.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    ``db`` contains a persistent record of all signatures with
+    enough information to prevent the triggering of any slashing conditions.
+    """
+    # verify no slashing conditions would be violated
+    # by signing the data pertaining to this duty
+    pass
+
+
+class Signatory(Task):
```
pipermerriam: Maybe `Notary`?  Not sure if that name captures the concept any better.

validator_client/task.py
```diff
@@ -0,0 +1,19 @@
+from abc import ABC, abstractmethod
+
+
+class BaseTask(ABC):
+    @abstractmethod
+    async def run(self, shell) -> None:
+        ...
+
+
+class Task(BaseTask):
```
pipermerriam: Regardless of whether you agree with dropping the *framework* complexity, I'd advocate to use the `TrioService`.  It has a concept of child service running (and thus a *graph* of parent/child services) as well as strong lifecycle management that is well tested.  That should mean that you're *interrupt* logic should only be required at the very outermost layer. 

It also has a nice feature for converting functions to services.  Your `Publisher` service would benefit from this simplicity since it's is just an `__init__` and `run` method.

validator_client/publisher.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, context, send_channel, *dependencies):
+        super().__init__(context, send_channel)
+        signatory, = dependencies
+        self._signatory = signatory
+        self._beacon_client = BeaconClient(context.beacon_node_endpoint)
+
+    async def run(self, shell):
+        async for duty, signature in shell.iter_channel(self._signatory):
+            logger.info("got signature %s for duty %s", humanize_hash(signature), duty)
+            await self._beacon_client.publish(duty, signature)
```
pipermerriam: *If* you used the `TrioService` then this class could be written as:

```python
@as_service
def publisher(manager, context, send_channel, signatory):
    async for duty, signature in send_channel:
        ...
```

validator_client/beacon_client.py
```diff
@@ -0,0 +1,25 @@
...
+
+    @property
+    def message(self):
+        return self._duty
+
+    def __repr__(self):
+        return self.message.hex()
+
+
+class BeaconClient:
```
pipermerriam: cc @kclowes I would assume this gets replaced by Web3.py `w3.beacon.do_thing(...)`

validator_client/client.py
ralexstokes: that's reasonable and possibly underscores that i didn't sufficiently motivate this PR.

the idea is that you just get to declare the components you have and you have the guarantee that you can talk to your children / you can get messages from your dependencies...

an earlier version of this PR had the duties split up so that the `Clock` injected timing data into two other components which I think shows off this approach in a more favorable light and also motivates some of the other code around setting up channels

i could very well just not have spent as much time w/ the existing way of doing things but to me this PR represents something far more explicit (by using indirection to hide the "wiring bits") so that a user can just focus on the important details -- what can talk to what.

the point of comparison in my perspective is that we have a bunch of components and they can all just publish to an event bus but you don't have a clear idea who or what can respond to a given message on the event bus

validator_client/shells/trio.py
ralexstokes: yeah -- i took a glance at `anyio` a few weeks ago and it is more or less what would be happening here. after putting this PR together i agree we don't gain enough by abstracting the async library and think we should not proceed :)

validator_client/clock.py
ralexstokes: agreed

validator_client/clock.py
ralexstokes: works for me :)

validator_client/clock.py
ralexstokes: i believe this was the intention w/ allowing a caller to provide a `time_provider` -- i can take a closer look at the `trio.Clock` interface and make sure it aligns w/ what i'm doing here

validator_client/clock.py
ralexstokes: currently no but earlier versions of this PR did... i realized i could simplify by structuring it the current way.

even still, i think what you are suggesting here breaks the uniformity of "everything is a `Task` and tasks can talk to each other over channels"

validator_client/clock.py
ralexstokes: yeah -- it would essentially be moving the code under `Clock` into this `dataclass`... the thinking here was that instances of this object are just "plain data containers" in the sense that they just shuffle data around b/t tasks

no fishy logic to get in the way of testing or downstream interpretation...

validator_client/context.py
ralexstokes: good to know

validator_client/signatory.py
ralexstokes: another option is just "Signer" but I see that symbol used to refer to actual software/hardware implementations of signature producers so don't want to conflate ideas...

validator_client/publisher.py
ralexstokes: just noting a similar decorator could be written here as well ;)

validator_client/beacon_client.py
ralexstokes: yeah i'd expect so! i'm not sure anyone has thought this far ahead tbh... i'm gonna ask around for more info (i think the chainsafe team has been working on some of this...) but generally i'd assume this is the case...

in the meantime, we have a very vanilla HTTP API that needs calling ;)

validator_client/client.py
ralexstokes: yeah, moving further down this path implies allowing a user to configure the channel size.

my intuition says that the cases where a non-blocking buffer is *required* are fewer than we may expect apriori -- but we would still want an escape hatch for those rare cases...

ralexstokes:
thanks for the feedback @pipermerriam :)

i think overall my main goal is to have very explicit paths of communication b/t components that moreover are isolated and/or having a minimal interface. the goal is to make it easy to test/reason about behavior while admitting as few concepts as possible (e.g. don't pull in the idea of lifecycle mgmt or hierarchies of components/tasks if you can avoid it).

having explored some of what this could look like in this PR, i do think its prudent to look at our `trio.Service` abstraction so now i think i'll port what I can in that direction incorporating the feedback here as is relevant...

validator_client/main.py
```diff
@@ -0,0 +1,46 @@
...
+import trio
+
+from p2p.trio_service import background_service
+from trinity._utils.logging import LOG_FORMATTER
+from validator_client.client import Client
+from validator_client.config import config
+from validator_client.context import Context
+
+
+def _setup_logging():
```
jannikluhn: IMO the best way to allow users to configure logging is to load a JSON/YAML/TOML file and pass it to [logging.dictConfig](https://docs.python.org/3.7/library/logging.config.html#logging.config.dictConfig). Easy to do and maximally powerful (but probably unnecessary at this stage).

validator_client/client.py
```diff
@@ -0,0 +1,31 @@
...
+
+    async def run(self) -> None:
+        logger.info("booting client...")
+        clock = Clock(self._context)
+        duty_scheduler = DutyScheduler(self._context, clock)
+        signatory = Signatory(self._context, duty_scheduler)
+        sink = publisher(self._context, signatory)
+        try:
+            self.manager.run_daemon_child_service(sink)
+            await self.manager.wait_stopped()
```
jannikluhn: IIRC, `wait_cancelled` is slightly preferable here because `wait_stopped` would only be finished after the `run` method has exited (which cannot happen here as it appears within `run`).

validator_client/clock.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def _compute_slot_and_alignment(self, t: int) -> Tuple[int, bool]:
+        """
+        Compute the slot corresponding to a unix time ``t``.
+        Indicate if ``t`` corresponds to the first tick in the relevant slot
+        with a boolean return value.
+        """
+        time_since_genesis = t - self._context.genesis_time
+        slot, sub_slot = divmod(time_since_genesis, self._context.seconds_per_slot)
+        return slot, sub_slot == 0
```
jannikluhn: It seems that because of leap seconds there might be cases in which two subsequent ticks would be regarded as `first_in_slot` for the same slot. I have no idea if this is an issue or not and if we should care about it or not, but wanted to point it out.

validator_client/client.py
pipermerriam: This is a foot gun that I've been pondering over for a while.  Here's why.

`ManagerAPI.wait_cancelled()` is not guaranteed to ever return.

This doesn't have to be the behavior but I would like to have a discussion about it because I'm not sure which is better.

- if services **always** end in the cancelled state:
    - `wait_cancelled()` and `wait_stopped()` are mostly equivalent and only in very rare cases would using one over the other make any difference....
    - You can't easily differentiate between a service that *finished* naturally and one that was cancelled...
- if cancellation is **only** triggered by an explicit call to `Manager.cancel()` or an unhandled task error:
    - `wait_cancelled()` is generally safe to use *internally* 
    - `wait_cancelled()` called from an *external* perspective will likely not behave as the user expected because it won't ever return in the case that the service exits naturally.
    - You can easily differentiate between a service that was cancelled and a service that exited naturally.

I suspect there is a nice 3rd option that gets us the best of both worlds.  Lately I've been thinking about making a clearer differentation between the *external* manager API and the *internal* manager API (I think the *internal* manager api is a superset of the *external* API)

pipermerriam: I went ahead and made an issue for this: https://github.com/ethereum/trinity/issues/1284

ralexstokes: in this case, i want to have a single thing that handles the orchestration of its internal components w/ an easy way to have an "entry" and "exit" point around the execution of these internal components. i think this "structural pattern" is common enough to be able to easily support in whatever API we have.

i think the thing to do here is to do `wait_stopped` for each of the `child_managers` that is created. waiting on the parent manager is "overkill" in this case, and as @jannikluhn points out actually not very helpful in terms of control flow

validator_client/duty_scheduler.py
```diff
@@ -70,10 +68,12 @@ def __init__(self, context, clock):
...
+    async def duties(self):
         async for tick in self._clock.ticks():
-            await shell.run_concurrently(
-                (self._process_tick, tick), (self._dispatch_duties_for, tick)
-            )
+            # NOTE: may want to *concurrently* process the current tick
+            # and also dispatch the duties for the current tick
+            # NOTE: for now, perform the tasks serially
+            await self._process_tick(tick)
+            duties = await self._resolve_duties_for(tick)
```
carver: > a regression: loss of concurrency w/in a task -- ideally the duty scheduler can launch concurrent tasks from within itself -- i think the solution here is to move from DutyScheduler running as an asynchronous generator into a full-fledged Service (with access to a manager to launch more tasks as desired)

I think most of the awkwardness here comes from us trying to abstract away the async infrastructure into a manager. (with only partially success, anyway, since we have `trio.sleep()` calls, etc). If we don't have to worry about hiding `trio` away, then we can do something simple-ish like this, right?

```py
    async with trio.open_nursery() as nursery:
        nursery.start_soon(self._process_tick(tick))
        duties = await self._resolve_duties_for(tick)
        ...
```

validator_client/publisher.py
```diff
@@ -0,0 +1,17 @@
...
+logger = logging.getLogger("validator_client.publisher")
+logger.setLevel(logging.DEBUG)
+
+
+@as_service
+async def publisher(manager, context, signatory):
+    beacon_client = BeaconClient(context.beacon_node_endpoint)
+    async for duty, signature in signatory.signatures():
+        logger.info("got signature %s for duty %s", humanize_hash(signature), duty)
+        await beacon_client.publish(duty, signature)
```
carver: You mentioned in chat that something in `publisher` was weird now. I'm not seeing it yet. What are you seeing?

> there is a bit of weird asymmetry w/ the publisher where before each piece was more of an independent concurrent part

validator_client/publisher.py
```diff
@@ -0,0 +1,17 @@
...
+from eth_utils import humanize_hash
+
+from p2p.trio_service import as_service
+from validator_client.beacon_client import BeaconClient
+
+logger = logging.getLogger("validator_client.publisher")
+logger.setLevel(logging.DEBUG)
+
+
+@as_service
```
carver: > before using the @as_service decorator, I just had a run method on the Publisher class that ran as a "daemon task" -- this may be splitting hairs but i think its nice to use the least heavy machinery possible when we can (task vs child service)...

Yeah, I agree! I don't *think* being a service is necessary here. In `p2p.BaseService` you would do something like use `self.run_daemon_task(publisher(...))`. Assuming the trio service has something similar, I think it would be fine to leave off the decorator here.

---

*meta note: I'm finding it awkward to copy/paste your comments from chat in here, so that there is some continuity to the conversation. I think it would be easier to have the whole convo in-PR, just having you comment on your own code in-line.*

validator_client/signatory.py
```diff
@@ -0,0 +1,57 @@
...
+                    continue
+
+                await self._record_signature_for(duty)
+                signature = await self._sign(duty.message)
+
+                logger.info("Signed %s", duty)
+
+                yield (duty, signature)
+        except trio.Cancelled as cancellation:
+            await self._stop()
```
carver: > how to execute service/task specific stop logic where i think its clear to catch trio.Cancelled as appropriate w/in each generator body

If you're looking for more alternatives, this is the kind of thing that a service can do in its `BaseService._cleanup()` (or whatever the trio service equivalent is). But I think this is fine, too.

validator_client/duty_scheduler.py
pipermerriam: Yes, using nurseries from within a `Service` is just fine.

validator_client/clock.py
```diff
@@ -0,0 +1,84 @@
...
+            "genesis time has not elapsed; blocking client until %s", genesis_time
+        )
+        duration = genesis_time - t
+        await trio.sleep(duration)
+
+    async def _wait_until_next_tick(self, tick) -> None:
+        next_tick_at = tick.t + self._tick_interval
+        t = self._time_provider()
+        duration = next_tick_at - t
+        if duration >= 0:
```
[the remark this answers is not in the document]
ralexstokes: `trio.sleep` raises `ValueError` if its argument is negative so we only want to call it here if the `duration` is non-negative. am i missing something else?

validator_client/clock.py
```diff
@@ -0,0 +1,84 @@
...
+    slot: int
+    epoch: int
+    first_in_slot: bool
+
+
+def _get_unix_time():
+    return time.time()
+
+
+class Clock:
```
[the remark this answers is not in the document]
ralexstokes: reading thru the docs, they seem pretty set on not providing a unix clock that satisfies the `trio.abc.Clock` interface so that clients of the library handle time in a generic way.

for the moment i'm thinking we can just keep the interface of `time_provider` (aka provides the current time as a `float` when you call it) and then inject whichever clock we need. note that this can be the `trio.testing.MockClock` just as easily as a call to `time.time`

validator_client/clock.py
ralexstokes: i'll think about this some more... it is a good thing to keep in mind!

validator_client/clock.py
```diff
@@ -65,20 +70,19 @@ def _compute_tick(self):
...
-        await shell.sleep(duration)
+        await trio.sleep(duration)
 
     async def _wait_until_next_tick(self, shell, tick) -> None:
         next_tick_seconds = tick.t + self._tick_interval
         t = self._time_provider()
         duration = next_tick_seconds - t
         if duration >= 0:
-            await shell.sleep(duration)
+            await trio.sleep(duration)
```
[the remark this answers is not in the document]
ralexstokes: `sleep_until` would be nice -- however the way trio implements time, we would have to compute the `duration` in any case and then add to the `trio.current_time`. given that we can stop once we have the `duration` w/ `sleep`, i don't see a reason to do the extra computation to use `sleep_until`.

validator_client/duty_scheduler.py
```diff
@@ -0,0 +1,79 @@
...
+        duties = _get_duties_for(tick, self._current_duties)
+        resolved_duties = ()
+        for duty in duties:
+            resolved_duty = await self._resolve_duty(duty)
+            resolved_duties += (resolved_duty,)
+        return resolved_duties
+
+    async def _process_tick(self, tick):
+        if tick.slot == self._current_slot:
+            return
```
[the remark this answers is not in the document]
ralexstokes: could you elaborate? there are multiple ticks per slot and we only need to update the duty state once per slot so we just return early in the case that we have already addressed this slot.

validator_client/beacon_client.py
```diff
@@ -0,0 +1,25 @@
+import random
+
+
+class Duty:
```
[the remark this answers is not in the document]
ralexstokes: good question! this is a specific unit of work. essentially, something like "i need to sign an attestation in some upcoming period of time" or "i need to sign a block in some upcoming period of time"

validator_client/client.py
```diff
@@ -0,0 +1,31 @@
...
+from validator_client.duty_scheduler import DutyScheduler
+from validator_client.publisher import publisher
+from validator_client.signatory import Signatory
+
+logger = logging.getLogger("validator_client.client")
+logger.setLevel(logging.DEBUG)
+
+
+class Client(Service):
+    def __init__(self, context: Context) -> None:
```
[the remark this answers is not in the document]
ralexstokes: yeah makes sense -- i'll keep this in mind once i get to the configuration. in general, the validator client config and the beacon node config will share some fields but then others i'd think are specific to each type of client.

validator_client/beacon_client.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, duty):
+        self._duty = duty
+
+    @property
+    def message(self):
+        return self._duty
+
+    def __repr__(self):
+        return self.message.hex()
```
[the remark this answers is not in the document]
ralexstokes: nice call! this `self.message.hex()` is here just to inject some dummy data while sketching out the higher-level bits... i'll keep this in mind once i get to writing the real `Duty`.

validator_client/signatory.py
ralexstokes: i may be missing it but it looks like `BaseService._cleanup` didn't make it into the `async_service` library that hosts our new fancy `TrioService`.

i asked for some more recent info about the way to catch/handle lifecycle events in `async_service` in the trinity gitter

validator_client/publisher.py
ralexstokes: i meant more here: https://github.com/ethereum/trinity/pull/1217/files#diff-80b05a9a97bdd8c90a6bde01a927018aR21-R33

before, each "thing"/service was an independent unit of execution that (hopefully) just had one job -- any downstream/upstream communication was handled semi-transparently and they were all run the same way w/ the expectation that they would all have the same lifecycle semantics.

it seems that collectively we prefer to break this uniformity in the conceptual model for (possibly) simpler code, i.e. turning some things in just async generators (cf. clock and signatory) and some things into full-fledged services (cf. duty_scheduler and the publisher_service).

the asymmetry i was referring to was about this break in the uniformity but at this point it seems like what is happening :)

validator_client/publisher.py
ralexstokes: ah nice point! thanks!!

validator_client/client.py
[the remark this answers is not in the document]
NIC619: @ralexstokes I also come across the validator/beacon node config issue when I'm integrating eth1 monitor into beacon node. Currently I hardcoded `ETH1_FOLLOW_DISTANCE` for it to work and @ChihChengLiang and I were thinking about how to set up this validator/beacon node config thing.
Do you have anything in mind already? And if you do how should I handle these config parameters for now(left them as hardcoded?) so that it won't cause you too much trouble to reorganize and cleanup?

validator_client/clock.py
```diff
@@ -0,0 +1,84 @@
...
+        await trio.sleep(duration)
+
+    async def _wait_until_next_tick(self, tick) -> None:
+        next_tick_at = tick.t + self._tick_interval
+        t = self._time_provider()
+        duration = next_tick_at - t
+        if duration >= 0:
+            await trio.sleep(duration)
+
+    async def ticks(self):
```
pipermerriam: Maybe `stream_ticks`

```python
async for tick in thing.ticks():
    ...
async for tick in thing.stream_ticks():  # this one feels a little more intuitive.
    ...
```

validator_client/duty_scheduler.py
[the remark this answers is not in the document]
pipermerriam: So this pattern promotes the pattern of "call `_process_tick()`  multiple times per slot and treat it as an idempotent action".  This is *ok* but it feels inelegant/sloppy since the extra calls aren't actually needed.  If this were to be designed to raise an exception, it moves responsibility to the caller to ensure that it is called exactly once per slot.  The latter forces us to have better architecture.

So I guess it's less of a foot-gun and more of something that hints that there may be a *better* fix for this.  The current approach is likely *fine*, but I see this as a sort of minor *code smell*.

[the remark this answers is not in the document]
ralexstokes: ah gotcha -- yeah i was already leaning towards refactoring this to split apart the multiple responsibilities of this class at the time of this comment...

i've done that now and it removes the need to check which slot we are in which i agree is suboptimal :)

validator_client/beacon_client.py
```diff
@@ -0,0 +1,19 @@
...
+class BeaconClient:
+    def __init__(self, slots_per_epoch, endpoint):
+        self._slots_per_epoch = slots_per_epoch
+        self._endpoint = endpoint
+
+    async def fetch_duties(self, public_keys, epoch):
+        some_slot = (
+            random.randint(0, self._slots_per_epoch) + epoch * self._slots_per_epoch
+        )
+        return (Duty(some_slot, bytes(random.getrandbits(8) for _ in range(32))),)
```
ralexstokes: we just make something up here for now

validator_client/duty.py
```diff
@@ -0,0 +1,23 @@
...
+    A ``Duty`` represents some work that needs to be performed on behalf of some
+    validator, like signing an ``Attestation``.
+    """
+
+    slot: Slot
+    work: bytes
+
+    @property
+    def message(self):
+        return self.work
```
ralexstokes: these are temporary and will leave soon

validator_client/signatory.py
```diff
@@ -0,0 +1,70 @@
...
+
+    async def run(self):
+        try:
+            async for duties in self._duty_provider:
+                for duty in duties:
+                    print(duty)
+                    self.run_task(self._process_duty, duty)
+                # async with trio.open_nursery() as nursery:
+                #     for duty in duties:
+                #         nursery.start_soon(self._process_duty, duty)
```
ralexstokes: i previously had the commented out parts of this region instead of the non-commented parts.

w/ the previous code, ctrl-c would _not_ trigger the `finally` block.

w/ this code, we do trigger the `finally` but instead get a different crash that i will continue to look into but it definitely seems like i'm missing something obvious w/ the `Manager` API

validator_client/streams/tee.py
```diff
@@ -0,0 +1,40 @@
+import trio
```
ralexstokes: i haven't thought about if I can simplify the implementation here, yet

validator_client/signatory.py
```diff
@@ -0,0 +1,66 @@
...
+
+        await self._send_channel.send((duty, signature))
+
+    async def run(self):
+        try:
+            async for duties in self._duty_provider:
+                for duty in duties:
+                    self.manager.run_task(self._process_duty, duty)
+        finally:
+            await self._stop()
```
ralexstokes: update: the `finally` block here never runs on `ctrl-c`

validator_client/clock.py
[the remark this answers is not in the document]
jannikluhn: oh, you're right, sorry! I got confused, looks like `asyncio.sleep` works with negative times, `trio.sleep` does not.

validator_client/signatory.py
pipermerriam: I'm not sure if something changed between then and now but I just pulled this branch, added a logging statement to the finally block here, ran the client and did a CTRL+C and the logger statement was indeed executed....

ralexstokes: good to know, thanks for checking! my immediate thought is differences across operating systems (uh oh) but i'll play w/ it some more on my end

carver: Hm, like ctrl-c triggering a different signal on a different shell?

Looks like this is a way to look up what signal is triggered by ctrl-c: https://unix.stackexchange.com/a/362579/251234

ralexstokes: more just like differences in signal processing/delivery across OSs... 

but to report back -- there is some interplay b/t a `trio.Cancelled` exception and which outstanding coroutines are run...

i added the logging @pipermerriam mentioned and turns out that in the above code it goes into the `self._stop` and gets to the first `await`able which executes but then it seems the remainder of the `_stop` coroutine is not run

pipermerriam: you should read this full section: https://trio.readthedocs.io/en/stable/reference-core.html#cancellation-semantics

The problem you're running into is that your within a cancel scope that has already been cancelled.  That means that subsequent `await` calls end up raising `Cancelled`.  To prevent this you have to manually re-enter a new cancel scope like this.

```python
async def run(self):
    try:
    finally:
        with trio.CancelScope(shield=True):
            await self._stop()
```

I tinkered and it does appear that the `shield=True` is required.

I'll point out that the trio docs on this topic are an important read because what this pattern promotes is a shutdown that has the potential to not shut down.  *something* caused the `run(...)` to exit and then we halt it and throw another "blocking" await call in there.

So the general rule seems to be that to do reliable cleanup you need to be aware that trio will raise `Cancelled` at the next checkpoint unless you shield your code.

ralexstokes:
Closing this PR as it has gotten a bit stale and the current status is different enough to warrant a new PR.

Issue #1218, opened 2019-10-10
Pytest Failing (eth2spec not found, Namespace has no attribute, import file mismatch)


wschwab:
* Trinity version: 0.1.0a29
* OS: Linux (Debian 10)
* Python version (output of `python --version`): Python 3.7.3
* Environment (output of `pip freeze`):
aiohttp==3.6.0
alabaster==0.7.12
apipkg==1.5
argcomplete==1.10.0
argh==0.26.2
asn1crypto==1.0.1
async-exit-stack==1.0.1
async-generator==1.10
async-lru==0.1.0
async-timeout==3.0.1
asyncio-cancel-token==0.2.0
atomicwrites==1.3.0
attrdict==2.0.1
attrs==19.2.0
Babel==2.7.0
backcall==0.1.0
base58==1.0.3
bleach==3.1.0
bloom-filter==1.3
bumpversion==0.5.3
cached-property==1.5.1
cachetools==3.1.1
certifi==2019.9.11
cffi==1.12.3
chardet==3.0.4
Click==7.0
coincurve==10.0.0
colorama==0.4.1
coverage==4.5.4
cryptography==2.6.1
cytoolz==0.10.0
decorator==4.4.0
docopt==0.6.2
docutils==0.15.2
eth-abi==1.3.0
eth-account==0.2.3
eth-bloom==1.0.3
eth-hash==0.2.0
eth-keyfile==0.5.1
eth-keys==0.2.4
eth-rlp==0.1.2
eth-typing==2.1.0
eth-utils==1.7.0
execnet==1.7.1
factory-boy==2.11.1
Faker==2.0.2
fastecdsa==1.7.4
flake8==3.5.0
flake8-bugbear==18.8.0
hexbytes==0.2.0
hypothesis==4.40.0
idna==2.7
ifaddr==0.1.6
imagesize==1.1.0
incremental==17.5.0
ipython==7.8.0
ipython-genutils==0.2.0
jedi==0.15.1
Jinja2==2.10.3
jsonschema==3.0.1
lahja==0.14.5
libp2p==0.1.1
lru-dict==1.1.6
lxml==4.4.1
MarkupSafe==1.1.1
mccabe==0.6.1
milagro-bls-binding==0.1.3
more-itertools==7.2.0
multiaddr==0.0.8
multidict==4.5.2
mypy==0.730
mypy-extensions==0.4.2
netaddr==0.7.19
netdisco==2.6.0
netifaces==0.10.9
[...]

Issue #1195, opened 2019-09-25
Normalize casing of enum variants


cburgdorf:
### What is wrong?

We currently use all these different forms of enum casing when we'd ideally be only using a single form.

### Snake Case

```
class Thingy(Enum):
    some_thing = auto()
    other_thing = auto()
```

### Capitalized Snake Case

```
class Thingy(Enum):
    SOME_THING = auto()
    OTHER_THING = auto()
```

### Pascal Case

```
class Thingy(Enum):
    SomeThing = auto()
    OtherThing = auto()
```

### How can it be fixed

1. Decide on casing
2. Correct all places that derive from `Enum`
3. Put the rule into our style guide

cburgdorf:
My vote is on the third option (PascalCase). @ethereum/snake-charmers

pipermerriam:
I'm in favor of still allowing the variant of all caps for the enum values in cases where we mostly only handle the values (as opposed to code that actually instantiates instances of the enum).  It's nice to be able to treat the enum values as constants.

cburgdorf:
@pipermerriam maybe that's a good argument to just settle with `CAPITALIZED_SNAKECASE` then? I think it's good if all follow the same style without exceptions.

davesque:
Yeah, I don't mind the capitalized snake case style.  I recently ran into this problem when trying to write some code in Rust.  The official preferred style in Rust is PascalCase.  However, there are already lots of things that are already PascalCase and having enum variants the same way doesn't distinguish them visually very much.  So I opted to use CAP_SNAKE instead, especially since you can do `use Enum::*;` in Rust.  In that case, the CAP_SNAKE case makes it more obvious where particular identifiers are coming from.

wschwab:
Is this open? I'd like to give it a shot. (I assume there's consensus on CAP_SNAKE.)

pipermerriam:
Go for it

Sent from ProtonMail mobile

-------- Original Message --------

cburgdorf:
This was addressed in #1219

PR #1219, opened 2019-10-13
tests failing, seems unrelated

wschwab:
### What was wrong?

#1195 : Making one consistent syntax for Enums across the codebase.

### How was it fixed?

Using `git grep` to find instances of Enums, then updating so that all followed SNAKE_CAPS.

### To-Do

- [x] Clean up commit history
- [x] Currently breaking tests, see #1218 , needs fixing

### DIFF
--- a/eth2/beacon/tools/fixtures/test_case.py
+++ b/eth2/beacon/tools/fixtures/test_case.py
@@ -10,17 +10,17 @@ from .test_handler import Input, Output, TestHandler
 
 
 class BLSSetting(Enum):
-    Optional = 0
-    Enabled = 1
-    Disabled = 2
+    OPTIONAL = 0
+    ENABLED = 1
+    DISABLED = 2
 
 
 def _select_bls_backend(bls_setting: BLSSetting) -> None:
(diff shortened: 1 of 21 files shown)


pipermerriam:
If you run the lint tests I think you'll see there's at least one that wasn't updated under `StoreBlockBodies`

wschwab:
@pipermerriam I totally missed that - fixing up `StoreBlockBodies` took care of the lint tests and a couple of others, but now I'm getting a couple of fails that look sort of random to me (`py36-integration` and `py37-wheel-cli`). I tried going through the traces there, but couldn't extract something relevant to the codebase.

What do you suggest?

ralexstokes:
@wschwab thanks for your help! the two failures i can see in CI right now are known bugs we have in our tests and should be unrelated to your work. so you can ignore them :)

wschwab:
Great! So I'm all done from my side?

cburgdorf:
Unrelated to this PR but I just merged https://github.com/ethereum/trinity/pull/1216 which should reduce the (unrelated) failure that you've hit there

PR #1222, opened 2019-10-21
Fix up a couple p2p bugs

carver:
### What was wrong?

Resolve a couple little random quirks about peer interaction:
- occasional `AttributeError: _manager` exceptions escaping
- occasional `AttributeError: p2p_api` exceptions escaping
- sync to checkpoint can fail way too quickly if you happen to connect to a bad peer (because `disconnect()` exits while the peer is still in the peer pool, so it just gets re-selected and burns through all the attempts to get the checkpoint header from a peer)

### How was it fixed?

- catch if exchange manager is missing and raise `PeerConnectionLost` instead of `AttributeError`
- only disconnect from `p2p_api` if it has been started & attached
- wait for peer to close before exiting `disconnect()` (this just revives the previous behavior)
- plus a bonus doc improvement

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/exchange/exchange.py
+++ b/p2p/exchange/exchange.py
@@ -37,15 +37,12 @@ class BaseExchange(ExchangeAPI[TRequestPayload, TResponsePayload, TResult]):
             self.get_response_cmd_type(),
         )
 
-        try:
-            self._manager = ExchangeManager(
-                connection,
-                response_stream,
-            )
-            async with run_service(response_stream):
-                yield
-        finally:
-            del self._manager
+        self._manager = ExchangeManager(
+            connection,
+            response_stream,
+        )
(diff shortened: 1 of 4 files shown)


p2p/exchange/exchange.py
```diff
@@ -64,7 +65,12 @@ def __init__(self) -> None:
         # bind the outbound request payload to the payload validator
         message_validator = partial(payload_validator, request.command_payload)
 
-        return await self._manager.get_result(
+        try:
+            manager = self._manager
```
pipermerriam: Maybe it would be better to remove the deletion of the manager from line 49?  Might provide saner error handling or remove the need for extra handling here alltogether.

PR #1223, opened 2019-10-22
Fix #1221, add eth2-extra

### LINKED ISSUE #1221: milagro-bls-binding is not available for windows
* Trinity version: latest
* OS: ~~osx~~/~~linux~~/win
* Python version (output of `python --version`): any version
* Environment (output of `pip freeze`): anything

### What is wrong?
milagro-bls-binding library is not available for windows, I suspect there may be more which are not available for windows.
Please include information like:
This library is included here 
["milagro-bls-binding==0.1.3"](https://github.com/ethereum/trinity/blob/07d9692a36663c69b6f00331654501b4d928b001/setup.py#L116),

### How can it be fixed
I dont know what it used for but we can fix it by using another such library in place of it, or ask the author to push windows version to pypi.

ChihChengLiang:
### What was wrong?

Modules such as milagro-bls-binding that have no Windows build would make installation experience bad for users.

### How was it fixed?

Move those package to an optional eth2-extra section.

#### Cute Animal Picture

### DIFF
--- a/eth2/_utils/bls/backends/milagro.py
+++ b/eth2/_utils/bls/backends/milagro.py
@@ -2,6 +2,10 @@ from typing import Iterator, Sequence, Tuple
 
 from eth_typing import BLSPubkey, BLSSignature, Hash32
 from eth_utils import to_tuple
+from py_ecc.bls.typing import Domain
+
+from eth2._utils.bls.backends.base import BaseBLSBackend
+from eth2.beacon.constants import EMPTY_PUBKEY, EMPTY_SIGNATURE
 from milagro_bls_binding import (
     aggregate_pubkeys,
     aggregate_signatures,
@@ -10,10 +14,6 @@ from milagro_bls_binding import (
     verify,
     verify_multiple,
 )
-from py_ecc.bls.typing import Domain
-
-from eth2._utils.bls.backends.base import BaseBLSBackend
-from eth2.beacon.constants import EMPTY_PUBKEY, EMPTY_SIGNATURE
 
 
 def to_int(domain: Domain) -> int:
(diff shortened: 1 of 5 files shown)


eth2/_utils/bls/backends/milagro.py
```diff
@@ -2,6 +2,10 @@
 
 from eth_typing import BLSPubkey, BLSSignature, Hash32
 from eth_utils import to_tuple
+from py_ecc.bls.typing import Domain
+
+from eth2._utils.bls.backends.base import BaseBLSBackend
+from eth2.beacon.constants import EMPTY_PUBKEY, EMPTY_SIGNATURE
 from milagro_bls_binding import (
```
carver: You can add `milagro_bls_binding` to `known_third_party` in tox.ini so that isort puts it in the correct (3rd party) grouping of packages.

tests/eth2/core/beacon/tools/builder/test_validator.py
```diff
@@ -59,7 +58,7 @@ def test_aggregate_votes(votes_count, random, privkeys, pubkeys):
 
     aggregated_pubs = bls.aggregate_pubkeys(pubs)
 
-    if votes_count == 0 and bls.backend == MilagroBackend:
```
carver: Hm, does this mean there is no code coverage for `Milagro` now? It seems like you'd want to add `eth2-extra` to the tox eth2-core environments, so you could keep testing with it.

ChihChengLiang: Oops, Milagro should have been covered in `tests/eth2/utils-tests/bls-utils/test_backends.py`. But since the `AVAILABLE_BACKENDS` does not contain Milagro if Milagro is uninstalled. This is a bug, will fix it.

Issue #1161, opened 2019-09-18
Trinity logs should automatically rotate across runs


ralexstokes:
### What is wrong?

`trinity{,-beacon}` saves logs across runs but the default is to simply append to the same file. It is convenient for diagnosing particular issues if the logs are unique across runs of the client.

### How can it be fixed

Log rotation per instantiation of the client.

pipermerriam:
YES!

KimiWu123:
I want to take this one!

PR #1228, opened 2019-10-24
#1161, Trinity logs should automatically rotate across runs

### LINKED ISSUE #1221: milagro-bls-binding is not available for windows

KimiWu123:
### What was wrong?
#1161, Trinity logs should automatically rotate across runs

### How was it fixed?
Log rotation per instantiation of the client with format, YYYYMMDD_HHMMSS. File name would look like trinity_20191024_095854.log.


trinity/_utils/logging.py
```diff
@@ -105,6 +107,17 @@ def setup_trinity_file_and_queue_logging(
...
+        ".",
+        len(formatted_logfile) - len(os.path.basename(formatted_logfile))
+    )
+    if idx != -1:
+        formatted_logfile = formatted_logfile[0:idx] + str_timestamp + formatted_logfile[idx:]
+    else:
+        formatted_logfile = formatted_logfile + str_timestamp + ".log"
+
     handler_file = RotatingFileHandler(
         str(logfile_path),
```
mhchia: Could it be better to do it with `Path.suffix`? For example something like this:
```suggestion
         str(logfile_path)[:-len(logfile_path.suffix)] + str_timestamp + logfile_path.suffix,
```

trinity/_utils/logging.py
```diff
@@ -105,8 +107,9 @@ def setup_trinity_file_and_queue_logging(
 
     log_queue = ctx.Queue()
 
+    str_timestamp = datetime.now().strftime('_%Y%m%d_%H%M%S')
     handler_file = RotatingFileHandler(
```
pipermerriam: The `RotatingFileHandler` has a [`doRollover`](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.RotatingFileHandler.doRollover) method.  Maybe we can simplify this by simply calling that method conditionally on whether `logfile_path.exists()`?

KimiWu123: Using `doRollover`, log files would simply rollover without checking `logfile_path.exists()` in my test. The format of log files is simpler but is not extension friendly (extension becomes log.1, log.2... etc)

ralexstokes: i think the idea is:

```python
# previous logging setup
handler = RotatingFileHandler(logfile_path, ...)
if logfile_path.exists():
    handler.doRollover()
# continue w/ the logging setup
```


and if you don't like the naming scheme, you can set a custom `namer` on `RotatingFileHandler`, see the docs here: https://docs.python.org/3/library/logging.handlers.html#logging.handlers.BaseRotatingHandler

ralexstokes: as i understand the default naming, i'd personally elect to append a timestamp to the name like:

```python
handler.name = lambda name: f"{name}.{_get_timestamp_as_str()}"
```

which means the suffix will be the timestamp thing but i don't see a problem w/ this in terms of UX and means you can quickly see the time at which the logfile was created

KimiWu123: as my understanding, if using, `namer` it would be like,
```
    str_timestamp = datetime.now().strftime('_%Y%m%d_%H%M%S')
    handler_file = RotatingFileHandler( str(logfile_path), ...)
    handler_file.namer = lambda name: f"{str(logfile_path)}.{str_timestamp}"
```
and if so, it need to call `doRollover ()` to create a file with the new name.

why not just putting the name with timestamp as default name?
```
    str_timestamp = datetime.now().strftime('_%Y%m%d_%H%M%S')
    handler_file = RotatingFileHandler( str(logfile_path)+str_timestamp, ...)
```

tests/integration/test_trinity_cli.py
```diff
@@ -283,7 +284,8 @@ def contains_substring(iterable, substring):
             if contains_substring(stderr_logs, log):
                 assert False, f"Log should not contain `{log}` but does"
 
-        log_file_path = TrinityConfig(app_identifier="eth1", network_id=1).logfile_path
+        log_dir = TrinityConfig(app_identifier="eth1", network_id=1).log_dir
+        log_file_path = max(log_dir.glob('*'), key=os.path.getctime)
```
mhchia: Though `ctime` should be consistent with the filename since we have timestamps in the name of log files, is it more direct to get the log file whose name has the largest timestamp?

mhchia:
Re-ran the tests.

trinity/_utils/logging.py
KimiWu123: Just like you said, it's unlikely to happen, but why not. I'll add it.

tests/integration/test_trinity_cli.py
KimiWu123: Isn't it easier to get latest file directly? Does it mean I need to parse files name first to get timestamp to compare?

trinity/_utils/logging.py
KimiWu123: If I do this, there would be two log files. One is `trinity.log_xxx` and the other is `trinity.log_xxx.1`.  It seems logs file created at `RotatingFileHandler()`

l0rb: You can set `RotatingFileHandler(delay=True)` to stop the handler from creating the file on instantiation.

KimiWu123:
Hi @mhchia ,

Those errors seem not relative to my commitment. Could you help check it? Thanks :)

cburgdorf:
@KimiWu123 I believe these CI fails would vaporize if you rebase on top of latest master. The reason for those is (I believe) that the underlying `fixtures` submodule was recently updated so the tests in this PR does not work with them until you rebase on top of latest master.

PR #1233, opened 2019-10-25
Add `Eth1Monitor` to monitor and manage deposit data

### LINKED ISSUE #709: Add eth1 service to monitor eth1 deposit contract
### What is wrong?
Eth2 client should monitor the deposit contract, and get notified when there are events happening.

### How can it be fixed
Have an `Eth1Monitor` service(and possibly a component) monitoring event logs from `deposit_contract` through web3.py.

`Eth1Monitor`
- stores the received event logs.
- exposes API for beacon node to make requests
    - `get_eth1_data() -> Eth1Data`
    - `get_deposit(deposit_root) -> Deposit`

======
Edit: rewrote `How can it be fixed` section.

mhchia:
### What was wrong?
Fixes #709 . 

### How was it fixed?
Add `Eth1Monitor`, which keeps polling new logs from the deposit contract and saves them. When a new block `latest_block` polled through `web3`, it fetchs the block `block_confirmed` with block number `latest_block.number - num_blocks_confirmed` and parses the logs inside. We can be less concerned about forks in eth1 chain with a large enough `num_blocks_confirmed` selected. Thanks to @NIC619 for coming up with the strategy.

`Eth1Monitor` is run as a `p2p.trio_service.Service`, and is tested with `lahja.trio.endpoint.TrioEndpoint`.

lahja APIs
- `get_eth1_data(distance: int, eth1_voting_period_start_timestamp: Timestamp) -> Eth1Data` through the event `GetEth1DataRequest`.
- `get_deposit(deposit_count: int, deposit_index: int) -> Deposit` through the event `GetDepositRequest`.

### To-Do
- [ ] Change `Eth1Monitor._deposit_data` to a DB?
- [ ] Refine the tests with pytest-trio tricks.

#### Cute Animal Picture


tests/plugins/eth2/beacon/conftest.py
```diff
@@ -0,0 +1,51 @@
...
...10b009050602060c0825160208401600060025af1610e3c57600080fd5b60c0519050602082610c800101526020810190506000610300600880602084610c0001018260208501600060046012f15050805182019150506000601860208206610b800160208284011115610e9157600080fd5b602080610ba082610700600060046015f1505081815280905090509050601880602084610c0001018260208501600060046014f150508051820191505061084051602082610c0001015260208101905080610c0052610c009050602060c0825160208401600060025af1610f0457600080fd5b60c0519050602082610c8001015260208101905080610c8052610c809050602060c0825160208401600060025af1610f3b57600080fd5b60c0519050610ae052606435610ae05114610f5557600080fd5b6001805460018254011015610f6957600080fd5b6001815401815550600154610d0052610d2060006020818352015b60016001610d0051161415610fb957610ae051610d205160208110610fa857600080fd5b600060c052602060c020015561104d565b6000610d205160208110610fcc57600080fd5b600060c052602060c0200154602082610d40010152602081019050610ae051602082610d4001015260208101905080610d4052610d409050602060c0825160208401600060025af161101d57600080fd5b60c0519050610ae052610d00600261103457600080fd5b60028151048152505b8151600101808352811415610f84575b5050005b60006000fd5b61017f6111d60361017f60003961017f6111d6036000f3"}'  # noqa: E501
```
pipermerriam: @njgheorghita can you package this up as an ethpm package?  Deposit contract for eth2.0 seems like a nice asset to have laying around.

trinity/plugins/eth2/beacon/eth1_monitor.py
```diff
@@ -0,0 +1,60 @@
...
+        contract_address: bytes,
+        contract_abi: str,
+        since_block_height: int,
+        logs_lookback_period: int,
+    ) -> None:
+        self._w3 = w3
+        self._deposit_contract = w3.eth.contract(
+            address=contract_address, abi=contract_abi
+        )
+        self._log_filter = self._deposit_contract.events.DepositEvent.createFilter(
```
pipermerriam: I'm inclined to :-1: using `createFilter`.  I'd advocate for using a `getLogs` based approach instead and manually managing the `fromBlock/toBlock`.  This removes dependency on statefulness on the RPC-node in favor of making "stateless" requests and managing the state locall (the state being which block ranges we've already queried)

mhchia: I see. May I understand that we should also avoid blocks filters, and maintain the queried `blockhash`s on our side like [`LogHandler`](https://github.com/mhchia/py-evm/blob/3ee86a4505b28ecff227e23c5ee64b2e58aa7996/evm/vm/forks/sharding/log_handler.py#L8-L42) in py-evm?

tests/plugins/eth2/beacon/conftest.py
njgheorghita: Package is available [here](http://explorer.ethpm.com/manifest/QmVokjyU5AMVLdpihsVqJ1eDCkDzv8QUnwV3NEnBJLncay) or in it's [raw format](https://ipfs.io/ipfs/QmVokjyU5AMVLdpihsVqJ1eDCkDzv8QUnwV3NEnBJLncay). Do you think it would be useful to create an `eth2` registry and publish it there?

pipermerriam: Published to a registry seems like a good thing.  Maybe most valuable thing would be to make a gist that explains how to install the CLI and then how to use that to install the contracts locally to disk.  This would be aimed at the eth 2.x client devs and showing them how to access the contract assets via ethpm tooling.

trinity/plugins/eth2/beacon/eth1_monitor.py
pipermerriam: @mhchia yeah, roughly that.  Basically, I don't want to *trust* the RPC node we're connected to to properly maintain the filter (and in the event that they forget/lose it, we get left in an awkward state of needing to re-install it but with a new `fromBlock`, which is effectively the same amount of work to keep track of as just manually maintain block ranges ourselves)

mhchia:
IMO it's more ready now. Though there are 800+ lines, the main logics are in `eth1_monitor.py`. I can separate them into smaller PRs if it is more helpful for review.

Further optimization can be done later, including storing `deposit_data` in the database, caching for trees, etc.

trinity/components/eth2/eth1_monitor/eth1_monitor.py
```diff
@@ -0,0 +1,347 @@
...
+        Return `Eth1Data` at `distance` related to the eth1 block earlier and closest to the
+        timestamp `eth1_voting_period_start_timestamp`.
+        Ref: https://github.com/ethereum/eth2.0-specs/blob/61f2a0662ebcfb4c097360cc1835c5f01872705c/specs/validator/0_beacon-chain-validator.md#eth1-data  # noqa: E501
+
+        First, we find the `eth1_block` whose timestamp is the largest timestamp which is smaller
+        than `eth1_voting_period_start_timestamp`. Then, find the block `target_block` at number
+        `eth1_block.number - distance`. Therefore, we can return `Eth1Data` according to the
+        information of this block.
+        """
+        eth1_voting_period_start_block_number = self._get_closest_eth1_voting_period_start_block(
```
ralexstokes: probably want to cache these start block numbers for a given `eth1_voting_period_start_timestamp`... again a re-org means we have to invalidate everything...

trinity/components/eth2/eth1_monitor/eth1_monitor.py
```diff
@@ -0,0 +1,347 @@
...
+            )
+
+        # Initialize the block's `deposit_count` with the one of its parent.
+        parent_block_number = block.number - 1
+        if parent_block_number not in self._block_number_to_accumulated_deposit_count:
+            self._block_number_to_accumulated_deposit_count[block.number] = 0
+        else:
+            self._block_number_to_accumulated_deposit_count[
+                block.number
+            ] = self._block_number_to_accumulated_deposit_count[parent_block_number]
```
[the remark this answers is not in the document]
ralexstokes: although also see below -- we *may* be able to remove this...?

tests-trio/eth1-monitor/test_eth1_monitor.py
```diff
@@ -0,0 +1,271 @@
...
+        if expected_block_number_at_distance is None:
+            block_at_distance = w3.eth.getBlock(block_number - distance)
+        else:
+            block_at_distance = w3.eth.getBlock(expected_block_number_at_distance)
+        assert eth1_data.block_hash == block_at_distance["hash"]
+
+    def assert_get_eth1_data_raises(block_number, distance):
+        block = w3.eth.getBlock(block_number)
+        # `ValidationError` is raised due to `deposit_count == 0` because
+        #   `eth1_voting_period_start_timestamp` is earlier than `deposit_included_block`.
```
NIC619: Where is the `deposit_included_block` defined?

tests-trio/eth1-monitor/test_eth1_monitor.py
```diff
@@ -0,0 +1,271 @@
...
+
+    # Assert b2
+    assert_get_eth1_data(block_numbers[2], 0, 2)
+    assert_get_eth1_data(block_numbers[2], 1, 2)
+    assert_get_eth1_data(block_numbers[2], 2, 1)
+    assert_get_eth1_data_raises(block_numbers[2], 3)
+
+    # Assert b3, b4, b5.
+    # Since these blocks are still within `num_blocks_confirmed`,
+    # queries with their timestamps should get the result of the latest block after
```
NIC619:
```suggestion
    # queries with `distance == 0` will return eth1_data at b2.
```

tests-trio/eth1-monitor/test_eth1_monitor.py
```diff
@@ -0,0 +1,271 @@
...
+    func_do_deposit()
+    func_do_deposit()
+    tester.mine_blocks(num_blocks_confirmed + 1)
+    await trio.sleep(polling_period)
+    await wait_all_tasks_blocked()
+    # `2` is automined blocks by `deposit`,
+    # and `num_blocks_confirmed + 1` are mined later.
+    number_recent_blocks = 2 + num_blocks_confirmed + 1
+    current_height = w3.eth.blockNumber
+    block_numbers = [current_height - i for i in reversed(range(number_recent_blocks))]
```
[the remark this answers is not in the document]
mhchia: According to your suggestion, the corresponding code should be
```
    block_numbers = list(
        range(current_height - number_recent_blocks + 1, current_height + 1)
    )
```
I've changed to this. Thank you!

trinity/components/eth2/eth1_monitor/eth1_monitor.py
```diff
@@ -0,0 +1,347 @@
...
+        accumulated_deposit_count = self._block_number_to_accumulated_deposit_count[
+            target_block_number
+        ]
+        if accumulated_deposit_count == 0:
+            raise Eth1MonitorValidationError(
+                f"failed to make `Eth1Data`: `deposit_count = 0` at block #{target_block_number}"
+            )
+        _, deposit_root = _make_deposit_tree_and_root(
+            self._deposit_data[:accumulated_deposit_count]
+        )
```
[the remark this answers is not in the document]
mhchia: > we could also query via web3.py in this function to get the block_hash, rather than maintaining it ourselves.

Do you mean `call` with [`block_identifier`](
https://web3py.readthedocs.io/en/stable/contracts.html#web3.contract.ContractFunction.call)? That looks pretty promising. If it works well, indeed we can remove `self._block_number_to_accumulated_deposit_count`. I'm trying that now. Big thanks for the suggestions!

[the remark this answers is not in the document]
ralexstokes: i don't think we need to go thru `call`...

once we have the right block number, we can just call this: https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.getBlock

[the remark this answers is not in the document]
ralexstokes: the thing you linked is interesting.... seems like we can get the contract state at a certain block that way -- i'll let you resolve the details but i would first just considering the overall approach here to just reading the state on-demand. like i said above, i think this is the "easier to get correct, more often" approach :)

mhchia:
> am I correct that this PR assumes that the eth1 chain is (kind of) finalized after `num_blocks_confirmed` blocks?

Yes, it is what we assume.

> I recalled that in order to fix reorg, we have to know which block number is the reorg point so we can query the deposit data on the new canonical chain. So we still need to maintain something like `_block_number_to_hash` to know where should it be?

I still use `_block_number_to_hash` just for caching the blocks we handled, and in the worst case even we are delayed `num_blocks_confirmed` blocks and there is a fork happening, we can find it out by learning the fact `_block_number_to_hash` is not consistent with the record in eth1 node.
If we want to recover from a re-org, we need to know the block number where reorg happened. We need to keep all recent `blockhashes` to find out the fork point, like [this](https://github.com/mhchia/py-evm/blob/3ee86a4505b28ecff227e23c5ee64b2e58aa7996/evm/vm/forks/sharding/log_handler.py#L8-L42). However, does it seem too late for `Eth1Monitor` to recover from a fork after `num_blocks_confirmed`? It indicates "from a fork happening and before we detecting it `Eth1Monitor` are returning false `Eth1Data`". It seems to me that waiting for a proper `num_blocks_confirmed` is easier. Still, if it is deemed necessary I can add the fork recovery mechanism.

trinity/components/eth2/eth1_monitor/eth1_monitor.py
mhchia: Is it possible that the cache is not hit frequently given that the queried timestamps are easily different? For example, they have some small differences and are mapped to the same `start_block_number`, but they are different keys to `_get_closest_eth1_voting_period_start_block`.

trinity/components/eth2/eth1_monitor/eth1_monitor.py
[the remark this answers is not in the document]
mhchia: @ralexstokes 
I am removing `_block_number_to_hash` because we can get the hash of a block at the number through web3.

For `_block_number_to_accumulated_deposit_count`, currently, I see two ways I can get the accumulated deposit count:
1. Query the deposit contract through web3, by `call(block_identifier=target_block_number+1)`.
2. Get every block before and including the block `target_block_number` through `getBlock`, parse the transactions inside and get the resulting logs, and then aggregate them.

Is either the way what you are talking about? If not, could you elaborate more on this? Thank you :pray:

mhchia:
@hwwhww @ralexstokes @NIC619 

Really appreciate your reviews!

After your reviews, I've changed several things according to the feedback:
- Cleaned up(including several nits, rephrasing comments, and cleaning leftover code).
- Removed `_block_number_to_accumulated_deposit_count` and get it through calling `get_deposit_count` of the deposit contract.
- Handled errors when processing requests and propagate them back to the requester through lahja.
- Reused `make_deposit_proof` and `make_deposit_tree_and_root`, and refactor `create_mock_deposits_and_root`.

There are things that haven't been done but I lean towards putting them in a separate PR or issue:
- Put `_deposit_data` in DB.
- Cache the trees according to `deposit_count`.

Please add if there is anything missing :pray:, and it will be great to have your second review on the changes. I am starting on another PR to solve the issue "Put `_deposit_data` in DB".

trinity/components/eth2/eth1_monitor/eth1_monitor.py
```diff
@@ -0,0 +1,387 @@
...
+    _w3: Web3
+
+    _deposit_contract: "Web3.eth.contract"
+    _deposit_event_abi: Dict[str, Any]
+    _deposit_event_topic: str
+    # Number of blocks we wait to consider a block is "confirmed". This is used to avoid
+    # mainchain forks.
+    # We always get a `block` and parse the logs from it, where
+    # `block.number <= latest_block.number - _num_blocks_confirmed`.
+    _num_blocks_confirmed: BlockNumber
```
hwwhww: I think `num_blocks_confirmed` is not `BlockNumber`, i.e., 5 blocks != block 5. 😄

trinity/components/eth2/eth1_monitor/exceptions.py
```diff
@@ -0,0 +1,19 @@
+from eth_utils import ValidationError
+
+
+class Eth1MonitorError(Exception):
+    pass
```
hwwhww: nitpick: now Trinity prefers to use `...` for declaring exceptions.
```suggestion
    ...
```

trinity/components/eth2/eth1_monitor/exceptions.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Eth1MonitorValidationError(ValidationError, Eth1MonitorError):
+    pass
```
hwwhww: ditto
```suggestion
    ...
```

trinity/components/eth2/eth1_monitor/exceptions.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Eth1BlockNotFound(Eth1MonitorError):
+    pass
```
hwwhww: ditto
```suggestion
    ...
```

trinity/components/eth2/eth1_monitor/eth1_monitor.py
mhchia: Nice catch. I've changed that. Thank you for the review.

mhchia:
I'm merging this and proceeds TODOs in #1270 .

PR #1235, opened 2019-10-25
Fixed #1231, fix libp2p tests

### LINKED ISSUE #1231: Test failed for obvious reason but not reflected in CI
### What is wrong?

The test  `pytest tests/libp2p/test_peer_factories.py` should fail, but the [CI](https://circleci.com/gh/ethereum/trinity/131568) still shows it passing. 

### How can it be fixed

Investigating 🕵️‍♂️🐕

ChihChengLiang:
### What was wrong?

Rer #1231, the libp2p tests are not run due to the absence of test-asyncio in tox.ini. 

### How was it fixed?

- Add test-asyncio to tox.ini.
- Fix ConnectionPairFactory

#### Cute Animal Picture


tox.ini
```diff
@@ -136,7 +136,7 @@ commands =
     pytest -n 1 {posargs:tests/trinity_long_run/}
 
 [testenv:py37-libp2p]
-deps = .[p2p,trinity,eth2,test]
+deps = .[p2p,trinity,eth2,test,test-asyncio]
```
ChihChengLiang: This is the key mistake that the tox is not running all `tests/libp2p/` tests.

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -83,7 +83,7 @@
 )
 from libp2p.security.base_transport import BaseSecureTransport
 from libp2p.stream_muxer.abc import IMuxedConn
-from libp2p.stream_muxer.mplex.exceptions import MplexStreamEOF, MplexStreamReset
+from libp2p.network.stream.exceptions import StreamEOF, StreamReset
```
ChihChengLiang: Libp2p now all raise StreamEOF and StreamReset instead of MplexStreamEOF and MplexStreamReset https://github.com/libp2p/py-libp2p/pull/287

trinity/protocol/bcc_libp2p/servers.py
```diff
@@ -297,7 +297,7 @@ def _process_received_block(self, block: BaseBeaconBlock) -> None:
             # Successfully imported the block. See if any blocks in `self.orphan_block_pool`
             # depend on it. If there are, try to import them.
             # TODO: should be done asynchronously?
-            self._try_import_orphan_blocks(block.parent_root)
+            self._try_import_orphan_blocks(block.signing_root)
```
ChihChengLiang: Something wrong here, need to think more.

- signing_root:
  - should be the correct argument to this function by design.
  - tests pass
  - When running `python eth2/beacon/scripts/run_beacon_nodes.py`, chains fork
- parent_root
  - tests fail
  - When running `python eth2/beacon/scripts/run_beacon_nodes.py`, chains sync

ChihChengLiang: Leaning toward breaking `python eth2/beacon/scripts/run_beacon_nodes.py` now and fix it later.

tests/libp2p/bcc/test_receive_server.py
```diff
@@ -88,6 +89,7 @@ def get_blocks(
     server = ReceiveServerFactory(chain=chain, topic_msg_queues=topic_msg_queues)
     asyncio.ensure_future(server.run())
     await server.events.started.wait()
+    await asyncio.sleep(0.5)
```
ralexstokes: what does this `sleep` do here?

tests/libp2p/bcc/test_receive_server.py
```diff
@@ -106,6 +108,7 @@ def get_blocks(
...
```
ralexstokes: ditto

tests/libp2p/bcc/test_receive_server.py
```diff
@@ -373,7 +378,7 @@ def _try_import_orphan_blocks(parent_root):
         for orphan_block in (blocks[4],) + fork_blocks:
             receive_server.orphan_block_pool.add(orphan_block)
         # Wait for receive server to process the orphan blocks
-        await asyncio.sleep(0.5)
+        await asyncio.sleep(1)
```
ralexstokes: do we want to use explicit events or just wait what we hope is enough time?

trinity/protocol/bcc_libp2p/servers.py
ralexstokes: skimmed this code just now and it looks like use of `signing_root` is correct...

do you have a better idea of why the nodes fork? do they end up w/ different sets of blocks? problem w/ attestations? we should definitely keep the signing root here and ideally figure out why the `run_beacon_nodes` script is forking before merging this in.

bonus points if we can add a test that captures the failure...

tests/libp2p/bcc/test_receive_server.py
ChihChengLiang: When the event `started` is set, daemon tasks in the server`_run` are not running yet. That would fail some tests. I would suggest adding a ready state for this, can send a draft PR later.

tests/libp2p/bcc/test_receive_server.py
ChihChengLiang: I'm inclined to wait as long as the test is not flaky.

tests/libp2p/bcc/test_receive_server.py
ChihChengLiang: The concept of the ready event would look like this. https://github.com/ChihChengLiang/trinity/pull/1/files

ralexstokes: interesting... i would think that `started` implies that all of the child/sub services/tasks are already running and ready to go... i'd be curious to get @pipermerriam 's thoughts on this re: the `Service` abstraction....

i haven't communicated this well yet but a sticking point I have w/ my understanding of the `Service` abstraction is the semantics around this kind of stuff... seems like our move to `trio` is in part of force *some* semantics upon ourselves (esp. easy to use ones like in trio)... clarity on this would be great and i'd move to put in writing once it is sorted (as in, i'll update the docs w/ any info we get in this thread). i'm paging @pipermerriam as this is some of the stuff i was aiming to explore in #1217 and he was kind enough to review it :)

also worth noting is that this stuff could be clear as I'll admit I haven't dug as deeply into this part of the codebase as well I would have liked to so in short, don't worry to tell me to RTFM :)

pipermerriam: For both the legacy (asyncio) and trio service, the events are more about overall service lifecycle and have nothing to do with the service logic.  The `started` event is potentially the most confusing.  The intended meaning behing `service.events.started` is: *"All of the internal initialization code of the service has now been run and the `_run` method will be executed soon"*.

If you need something that is representative of the internal state of the `_run` method, you'll have to make your own event for that.

The trio-based service has the same potentially confusing API.  I looked into whether it made sense to include an event for the later concept but I found that it was better left up to the individual service classes as to whether they needed it.  I did try hard to define the lifecycle of the new trio-based service very concretely and it has what I feel are thorough tests to try and ensure that lifecycle is strictly enforced.  I suspect there is still room for better documentation of the lifecycle both in code and in more formal RST docs.

tests/libp2p/bcc/test_receive_server.py
pipermerriam: I believe that these sleeps would be better replaced by exposing an internal event on the service which can be awaited here.  The event would be set from somewhere within your `_run` method at whatever point things have properly initialized.

tests/libp2p/bcc/test_receive_server.py
pipermerriam: In that case, this should be removed and you should do something like:

```python
await asyncio.wait_for(asyncio.gather(peer_1_called_event.wait(), peer_2_called_event.wait()), timeout=1)
```

This will block until they are set and then exit immediately, timing out in the case something is broken.

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -576,11 +576,11 @@ def _compare_chain_tip_and_finalized_epoch(self,
         try:
             await write_resp(stream, hello_mine, ResponseCode.SUCCESS)
             has_error = False
-        except (WriteMessageFailure, MplexStreamEOF, MplexStreamReset) as error:
+        except (WriteMessageFailure, StreamEOF, StreamReset) as error:
```
pipermerriam: I think this is better done as:

```
try:
    try:
        await write_resp(...)
    except Exception:
        has_error = True
        raise
    else:
        has_error = False
except WriteMessageFailure:
    await stream.reset()
except StreamEOF:
    await stream.close()
finally:
    if has_error:
        ...
```

trinity/protocol/bcc_libp2p/servers.py
ChihChengLiang: Looks like #1236 solved this. Using `signing_root` in `run_beacon_nodes` script can sync now.
My theory of what happened before #1236 is that passing `parent_root` in `_try_import_orphan_blocks` moves the state to the right slot, as what is done in #1236.

trinity/protocol/bcc_libp2p/node.py
ChihChengLiang: Have a draft PR to handle this particularly #1243

trinity/protocol/bcc_libp2p/servers.py
```diff
@@ -198,6 +200,8 @@ def __init__(
             if message.from_id == self.p2p_node.peer_id:
                 continue
             await handler(message)
+            self.handle_message_done.set()
+            self.handle_message_done.clear()
```
mhchia: It's interesting to learn that this [wakes all coroutines](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event.set) which are waiting for this event. Would it be better to make these events stay in the tests since they are for testing?

trinity/protocol/bcc_libp2p/servers.py
```diff
@@ -247,6 +251,8 @@ def __init__(
                         continue
                     else:
                         self._process_received_block(block)
+            self.process_orphan_blocks_done.set()
+            self.process_orphan_blocks_done.clear()
```
mhchia: ditto

trinity/protocol/bcc_libp2p/servers.py
ChihChengLiang: Figured out another clean way to replace `handle_message_done` and `process_orphan_blocks_done`.

tests/libp2p/bcc/test_receive_server.py
```diff
@@ -27,9 +27,12 @@
 from trinity.tools.bcc_factories import (
     AsyncBeaconChainDBFactory,
     BeaconBlockFactory,
+    PeerFactory,
     ReceiveServerFactory,
 )
 
+from .helpers import wait_until_true
```
pipermerriam: I'm :-1: on doing cross imports within tests modules.

These things should either go under a `<lib>.tools` namespace in the main codebase or be inlined in the module itself.

By placing them under the main library namespace this ensures that these modules are subject to both linting and type checks which makes refactoring and debugging these things significantly easier in my experience.

tests/libp2p/bcc/helpers.py
```diff
@@ -0,0 +1,10 @@
+import asyncio
+
+
+async def wait_until_true(predicate, timeout=1.0, retry=5):
+    for _ in range(retry):
+        if predicate():
+            return True
+        else:
+            await asyncio.sleep(timeout / retry)
```
pipermerriam: I'm inclined to push for removing the `retry` argument and making the sleep a `sleep(0)` so that things can process as fast as possible.

PR #1236, opened 2019-10-28
Fix import fork chain block

NIC619:
### What was wrong?
When importing a fork chain block, the canonical `state` (`state` on canonical chain) is used to apply state transition and this leads to validation error due to mismatch `state.latest_block_header` and hence the block is not imported successfully.

### How was it fixed?
When importing a block, use the state at it's parent block (`get_state_by_root(parent_block.state_root)`) and apply state transition.
This should fix the forking issue mentioned at https://github.com/ethereum/trinity/pull/1235#discussion_r339435232 .

Also add some explanation for `test_demo.py` as suggested at https://github.com/ethereum/trinity/pull/1214#discussion_r339095159

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/eth2/beacon/chains/base.py
+++ b/eth2/beacon/chains/base.py
@@ -414,27 +414,9 @@ class BeaconChain(BaseBeaconChain):
                 )
             )
 
-        try:
-            canonical_root_at_slot = self.get_canonical_block_root(parent_block.slot)
-        except BlockNotFound:
-            # No corresponding block at this slot which means parent block is not
-            # on canonical chain.
-            is_new_canonical_block = False
-        else:
-            is_on_canonical_chain = parent_block.signing_root == canonical_root_at_slot
-            is_newer_than_head_state_slot = (
-                self.chaindb.get_head_state_slot() < block.slot
-            )
-            is_new_canonical_block = (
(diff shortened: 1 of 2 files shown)


tests/eth2/integration/test_demo.py
```diff
@@ -96,6 +96,37 @@ def test_demo(base_db, validator_count, keymap, pubkeys, fork_choice_scoring):
     assert state.slot == chain_length + genesis_slot
 
     # Justification assertions
+    # NOTE: why are the number `2` or `3` used in the checks below?
+    # Answer:
```
ralexstokes: nice! ^_^

eth2/beacon/chains/base.py
```diff
@@ -414,27 +414,9 @@ def import_block(
...
-                is_on_canonical_chain and is_newer_than_head_state_slot
-            )
-
-        if is_new_canonical_block:
-            state_machine = self.get_state_machine()
-            state = self.get_head_state()
-        else:
-            state_machine = self.get_state_machine(at_slot=parent_block.slot)
-            state = self.get_state_by_slot(parent_block.slot)
+        state_machine = self.get_state_machine(at_slot=parent_block.slot)
```
ralexstokes: this seems good

PR #1237, opened 2019-10-28
fix eth_syncing endpoint 

### LINKED ISSUE #765: Fix broken JSON-RPC eth_syncing API call
### What is wrong?

The functionality got removed in https://github.com/ethereum/trinity/commit/4c87a663a0b426d418045ece74c54cb4d6c1eeb4#diff-8a563f755f9f95bd5dbd1e7b6a0f3c22L160 (probably accidentially ping @carver )

We had no tests for this so that went unnoticed for a while.

### How can it be fixed

Add it back in and then expand this test to ensure it doesn't break again.

https://github.com/ethereum/trinity/blob/b35ddb9b90a440ec98d5d141e3fbc6e2f1f35b2b/tests/integration/test_trinity_cli.py#L142-L143

voith:
### What was wrong?

fixes #765 
`eth_syncing` was removed accidentally in https://github.com/ethereum/trinity/commit/4c87a663a0b426d418045ece74c54cb4d6c1eeb4#diff-8a563f755f9f95bd5dbd1e7b6a0f3c22L160.

### How was it fixed?
added back code that was removed accidentally.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -77,7 +77,7 @@ async def test_skeleton_syncer(request, event_loop, event_bus, chaindb_fresh, ch
     )
     async with peer_pair as (client_peer, server_peer):
 
-        client_peer_pool = MockPeerPoolWithConnectedPeers([client_peer])
+        client_peer_pool = MockPeerPoolWithConnectedPeers([client_peer], event_bus=event_bus)
         client = FastChainSyncer(LatestTestChain(chaindb_fresh.db), chaindb_fresh, client_peer_pool)
         server_peer_pool = MockPeerPoolWithConnectedPeers([server_peer], event_bus=event_bus)
 
@@ -190,7 +190,7 @@ async def test_beam_syncer(
         # manually add endpoint for trie data gatherer to serve requests
         gatherer_config = ConnectionConfig.from_name(f"GathererEndpoint-{unique_process_name}")
 
-        client_peer_pool = MockPeerPoolWithConnectedPeers([client_peer])
+        client_peer_pool = MockPeerPoolWithConnectedPeers([client_peer], event_bus=event_bus)
         server_peer_pool = MockPeerPoolWithConnectedPeers([server_peer], event_bus=event_bus)
 
         async with run_peer_pool_event_server(
@@ -260,7 +260,8 @@ async def test_regular_syncer(request, event_loop, event_bus, chaindb_fresh, cha
         client = RegularChainSyncer(
             ByzantiumTestChain(chaindb_fresh.db),
             chaindb_fresh,
-            MockPeerPoolWithConnectedPeers([client_peer]))
+            MockPeerPoolWithConnectedPeers([client_peer], event_bus=event_bus)
+        )
         server_peer_pool = MockPeerPoolWithConnectedPeers([server_peer], event_bus=event_bus)
 
         async with run_peer_pool_event_server(
@@ -381,7 +382,8 @@ async def test_regular_syncer_fallback(request, event_loop, event_bus, chaindb_f
         client = FallbackTesting_RegularChainSyncer(
             ByzantiumTestChain(chaindb_fresh.db),
             chaindb_fresh,
-            MockPeerPoolWithConnectedPeers([client_peer]))
+            MockPeerPoolWithConnectedPeers([client_peer], event_bus=event_bus)
+        )
         server_peer_pool = MockPeerPoolWithConnectedPeers([server_peer], event_bus=event_bus)
 
         async with run_peer_pool_event_server(
@@ -425,7 +427,8 @@ async def test_light_syncer(request,
         client = LightChainSyncer(
             LatestTestChain(chaindb_fresh.db),
             chaindb_fresh,
-            MockPeerPoolWithConnectedPeers([client_peer]))
+            MockPeerPoolWithConnectedPeers([client_peer], event_bus=event_bus)
+        )
         server_peer_pool = MockPeerPoolWithConnectedPeers([server_peer], event_bus=event_bus)
 
         async with run_peer_pool_event_server(
--- a/trinity/sync/common/headers.py
+++ b/trinity/sync/common/headers.py
@@ -9,6 +9,7 @@ from typing import (
     FrozenSet,
     Generic,
     Iterable,
+    Optional,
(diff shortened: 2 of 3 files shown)


voith:
marking this as WIP as I need to test this more.

tests/core/p2p-proto/test_sync.py
```diff
@@ -90,8 +90,16 @@ def small_header_batches(monkeypatch):
             client_peer.logger.info("%s is serving 1000 blocks", client_peer)
             server_peer.logger.info("%s is syncing up 1000 blocks", server_peer)
 
-            await asyncio.wait_for(client.run(), timeout=20)
```
voith: This particular line was getting timed out after this change no matter how much I'd increase the timeout. This is happening because `event_bus.stream(SyncingRequest)` waits indefinitely. This is the only way I could fix it.
The code below fixes this problem:
```python
asyncio.ensure_future(client.run())
 await wait_for_head(chaindb_fresh, chaindb_1000.get_canonical_head())
```
However, it doesn't seem right. The old code was nice as it would return as soon as it would catch up with its peer.
@carver Since you originally wrote this code, do you have any advise?

voith:
I have tested this with all sync-modes. However, I couldn't check for the edge case when the client catches up with the peer as my node is not fully synced.

Regarding the test failures, `py36-integration` and `py36-eth2-components ` look unrelated to this change. I don't have the privileges to restart tests.
Tests in `py36-lightchain_integration ` are failing locally on master too. But I'm not completely sure as its passing for other PR's. Can someone please check this locally?

I will squash the commits and add release notes once code review is done.

@pipermerriam @carver Can I get a review on this?

trinity/sync/common/headers.py
```diff
@@ -989,3 +1021,16 @@ def _is_syncing_skeleton(self) -> bool:
...
             pass
+
+    def _get_sync_status(self) -> Tuple[bool, Optional[SyncProgress]]:
+        if not self._is_syncing_skeleton or not self._meat.sync_progress:
+            return False, None
+        return True, self._meat.sync_progress
+
+    async def _handle_sync_status_requests(self) -> None:
+        event_bus = self._peer_pool.get_event_bus()
+        async for req in event_bus.stream(SyncingRequest):
```
carver: You want this loop to cancel out when `_build_skeleton()` exits, right? I think this will do what you're looking for:

```suggestion
        async for req in self.wait_iter(event_bus.stream(SyncingRequest)):
```

voith: Thank you very much @carver. I struggled a lot with this. `asyncio` is a very cryptic library 😆

voith:
Tests are green now and I've also added release notes. This PR is now ready for review.

trinity/sync/common/headers.py
```diff
@@ -989,3 +1021,23 @@ def _is_syncing_skeleton(self) -> bool:
                 "%s announced Head TD %d, which is higher than ours (%d), starting sync",
                 peer, peer.head_info.head_td, head_td)
             pass
+
+    def _run_handle_sync_status_requests(self) -> None:
+        try:
+            event_bus = self._peer_pool.get_event_bus()
+        except AttributeError:
+            pass
```
carver: Is it the peer pool or `get_event_bus()` that is missing? Why would it be missing? Some comments here would help.

It's probably at least worth a debug log. Or if this should only happen in testing, then maybe a warning log.

Is this the change you made instead of using `wait_iter()` below?

voith:
@carver Test failures are not related to this change. The last commit for which tests are failing only adds release notes. The previous commit is green. Can you please restart the tests?

trinity/sync/common/headers.py
voith: `event_bus` is not a strict requirement for creating a `peer pool`.  I  can definitely add a log message here. I took the `wait_iter` route but `lightchain_integration` tests were still failing.
At this point, I realized that `event_bus` is not a strict requirement and it should be handled using try except.

voith: In fact, the original version had a similar check: https://github.com/ethereum/trinity/blob/b3362edf350171acbaaa5774a5a75447a2762029/trinity/sync/common/chain.py#L100-L101
I initially thought that this check was unnecessary. But the tests made me realize that `event_bus` is not a strict requirement and I was approaching this the wrong way.

voith: I have added the note that you requested.

voith:
@carver tests are green now.

trinity/sync/common/headers.py
carver: Cool, thanks! Looks like the `pass` isn't necessary anymore.

I guess I want to take another look at lightchain integration. For integration tests in particular, I would expect everything to work without any special "test-friendly" constructions. Maybe there's a bug in the light chain that isn't setting the event bus correctly, and we're only now noticing?

Maybe rather than catching the `AttributeError`, we can call `if self._peer_pool.has_peer_pool:`. That seems to be the preferred approach in `p2p/peer_pool.py`. Also, catching the exception is ambiguous (since a missing `_peer_pool` would also cause the same error).

voith: > Looks like the pass isn't necessary anymore.

I saw some surrounding code and I thought this was some pattern 😆 . But now I realize that was a mistake too. I can clean that up in a separate PR.
https://github.com/ethereum/trinity/blob/3223ecc8c61a6bedd6ff2f2484e8514755d87691/trinity/sync/common/headers.py#L988-L991

> if self._peer_pool.has_peer_pool

I'm guessing that you meant `self._peer_pool.has_event_bus`. I have made this change.

> Maybe there's a bug in the light chain that isn't setting the event bus correctly, and we're only now noticing?

Just adding `event_bus` to `lightchain_integration` tests on the current master causes the tests to fail. So there's could be some bug there. But I thought that it was too much to handle in this PR.

I can add `event_bus` to tests in `tests/core/p2p-proto/test_sync.py` like I had added before in https://github.com/ethereum/trinity/commit/daebaf67203cc73d219b5374d6a0b94bd3b7df94

voith: I added `event_bus` to tests in `tests/core/p2p-proto/test_sync.py`. Also, added back `self.wait_iter`

carver:
Thanks!

PR #1189, opened 2019-09-20
Revamp Command and Protocol API

pipermerriam:
### What was wrong?

Why Notes:

- All command payloads are strongly typed.
- Since payloads are our own objects they can have things like computed fields.
- No more separation between command instance and payload instance (payload is now part of the command instance)
- Given a command instance and the protocol instance it belongs to you can have a generic `Protocol.send(command)` method (no more need for individual `send_thing` methods on the protocol class.
- Convenience APIs for sending protocol methods fit nicely on the new `Application` API
- Composition over inheritance. `Command` classes no longer rely on inheritance to extend their functionality.
- Better suited for re-usable code that serves protocol command over the event bus.  We only need a single lahja event type which only needs `session`, `protocol_type`, `command` to transmit **any** protocol message over the wire.  (this was some of the primary motivation for doing this)

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [x] Clean up commit history

[//]: # (For important changes that should go into the release notes please add a newsfragment file as explained here: https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)

[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)
- [ ] Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #1238, opened 2019-10-28
Explicit versioning for protocols with multiple supported versions

pipermerriam:
extracted from #1189 

### What was wrong?

Both the base `p2p` protocol and the `les` protocol used naming schemes that omitted the version on one of the implementations.  This lead to a bit of ambiguity when looking at the classes as to which version you were using.

### How was it fixed?

Changed both classes to use explicit versions in their names.

#### Cute Animal Picture

nice-pug-puppy

### DIFF
--- a/p2p/p2p_proto.py
+++ b/p2p/p2p_proto.py
@@ -11,7 +11,7 @@ import rlp
 from rlp import sedes
 
 from p2p.abc import TransportAPI
-from p2p.constants import P2P_PROTOCOL_COMMAND_LENGTH
+from p2p.constants import DEVP2P_V4, DEVP2P_V5, P2P_PROTOCOL_COMMAND_LENGTH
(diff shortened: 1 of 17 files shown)


p2p/p2p_proto.py
```diff
@@ -136,7 +136,7 @@ def send_pong(self) -> None:
 
 
 class P2PProtocolV4(BaseP2PProtocol):
-    version = 4
+    version = DEVP2P_V4
```
carver: Pushing this into a constant seems like overkill to me, the number is hard-coded in the name anyway. The actual value was more transparent (now it requires indirection to another file to see what it is). But I'm not swimming in these every day, so maybe it's helpful for a reason I'm not seeing.

pipermerriam: It's minor, but i'm inclined to keep the constant.  It's used in a bunch of other places, this was just an old leftover that never got updated.

PR #1239, opened 2019-10-28
Organize trinity tools factories

pipermerriam:
Builds on #1238 

Extracted from #1189 

### What was wrong?

The `trinity.tools.factories` module got too big while working on #1189 and needed to be divided up into smaller modules.

### How was it fixed?

Re-organized things by their natural categories.

#### Cute Animal Picture

hqdefault

### DIFF
--- a/trinity/tools/factories/db.py
+++ b/trinity/tools/factories/db.py
@@ -57,6 +57,9 @@ class AsyncHeaderDBFactory(factory.Factory):
 
         headerdb = model_class(*args, **kwargs)
 
+        # SIDE EFFECT!
+        # This uses the side effect of initializing a chain using the `builder`
+        # tool to populate the genesis state into the database.
         build(
             Chain,
             latest_mainnet_at(0),
--- a/trinity/tools/factories/db.py
+++ b/trinity/tools/factories/db.py
@@ -1,0 +1,65 @@
+try:
+    import factory
+except ImportError:
+    raise ImportError("The p2p.tools.factories module requires the `factory_boy` library.")
+
+from typing import (
+    Any,
+    Type,
+)
+
+from eth_utils import to_bytes
+
+from eth.db.header import HeaderDB
+from eth.db.backends.memory import MemoryDB
+from eth.db.atomic import AtomicDB
+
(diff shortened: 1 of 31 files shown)


trinity/tools/factories/db.py
```diff
@@ -0,0 +1,65 @@
...
+        genesis_params = kwargs.pop('genesis_params', None)
+
+        headerdb = model_class(*args, **kwargs)
+
+        build(
+            Chain,
+            latest_mainnet_at(0),
+            genesis(db=headerdb.db, params=genesis_params),
+        )
+        return headerdb
```
carver: So I guess this is building a chain to populate the database as a side effect?

trinity/tools/factories/les/proto.py
```diff
@@ -1,181 +1,42 @@
+try:
+    import factory
+except ImportError:
+    raise ImportError("The p2p.tools.factories module requires the `factory_boy` library.")
```
carver: Another option here is to import this friendly way in a `trinity/tools/thirdparty.py`. Then in all these files that want to import, it can drop to a one-liner, like: `from trinity.tools.thirdparty import factory`.

trinity/tools/factories/transactions.py
```diff
@@ -0,0 +1,46 @@
...
+                **kwargs: Any) -> BaseTransactionFields:
+        if 'vrs' in kwargs:
+            v, r, s = kwargs.pop('vrs')
+        else:
+            if 'private_key' in kwargs:
+                private_key = kwargs.pop('private_key')
+            else:
+                private_key = PrivateKeyFactory()
+
+            tx_for_signing = FrontierUnsignedTransaction(**kwargs)
```
carver: Similar to how we have an alias for the latest mainnet VM, maybe it would be good to have aliases for some of these things, like `ModernUnsignedTransaction` or `LatestMainnetUnsignedTransaction`.

pipermerriam: I'm going to leave this until we need it.  Right now the type of transaction matters very little here and I intend to migrate these to Py-EVM under `eth.tools.factories`.

Issue #1225, opened 2019-10-23
Beam Sync Metadata Propagation


carver:
### What is wrong?

Beam Sync (v0) can only keep up if you have at least one great connection to a peer. Most of the time is spent in the round-trip waiting for data, and which data to ask for is discovered in serial, after receiving the answer from the last request.

We need a more reliable BeamSync experience.

### How can it be fixed

Propagate metadata about which state is needed in a block, using a new devp2p subprotocol (`firehose0`?) with just one command: `WitnessMetadata`. A couple options for the ways to send metadata are considered in this sampling of mainnet witnesses: https://notes.ethereum.org/oqrAzyhhQ_Oczvw7OUqs4w

#### Format
I'm still undecided on the two possible metadata formats.

One would be:
- block hash
- block number
- the list of all hashes of trie nodes that make up the block witness

The other would be:
- block hash
- block number
- list of addresses that are read during the block (excluding transaction signers or recipients and coinbase, assume all recipients need to load their bytecode)
- an extra byte for whether the bytecode was loaded
- a list of storage slots read, by address
- an extra list of trie node hashes required for trie reorgs

The second one is more flexible for future ways of sending state, but the first one is so simple! The second one is a smaller amount of data, but the first one should still still be small enough to fit in a single request. (see the linked `notes.ethereum.org` above)

#### Directionality
Should this be broadcast (like `NewBlock`) or pulled data (like `GetBlockHeaders`)? I would argue that it depends on the probability that peers are mid-sync. If it's common for peers to be syncing, then broadcast seems ideal. There's a lot more tracking involved in guessing which peers have which witnesses available, so maybe broadcast is the simple solution, potentially at the cost of network bandwidth. We can tinker with the percent broadcast rate, but early on, I think it will be hard enough to find witness broadcasters, that we should just start at 100%. (Except if the peer has sent you the same witness, or a more recent one).

carver:
My default direction is toward simplicity: just node hash witnesses, and broadcast-style p2p.

pipermerriam:
I believe we need a way to signal what network we're on.  I think this would be best done with a handshake message exchanged during protocol handshakes.

Using the key/value format that `LES` uses seems appropriate because it allows for extension without having to update the base protocol  message formate.

carver:
> Should this be broadcast (like `NewBlock`) or pulled data (like `GetBlockHeaders`)?

[...]

PR #1250, opened 2019-11-05
[WIP] First pieces of Beam Sync phase 1

carver:
### What was wrong?

Some setup for #1225

### How was it fixed?

It's not. Just getting a rough start on introducing the new subprotocol to broadcast (and receive) the node hashes in the witness metadata.

~Some of it references APIs that probably won't be available until #1189 is merged.~
It also depends on things that aren't merged into py-evm yet. (Actually, they might not even be in a PR yet)

This is completely untested, just sharing early.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


trinity/sync/beam/witness.py
```diff
@@ -0,0 +1,353 @@
...
+        new_node_tasks = tuple(
+            task for task in node_tasks
+            if task.node_hash not in self._db and task not in self._witness_node_tasks
+        )
+
+        await self._witness_node_tasks.add(new_node_tasks)
+
+    async def _collect_peer_witness_metadata(self) -> None:
+        while self.is_operational:
+            peer, cmd, msg = await self.wait(self.msg_queue.get())
```
carver: @pipermerriam What would be the "new API" way of doing this? I want to collect all messages of a certain type, as they come in, across all peers. (Notice that this class is only listening to the one message: `NewBlockWitnessHashes`)

I'm thinking it's something like: attach a custom behavior at runtime to every peer. Roughly:

```py
    def register_peer(self, peer: BasePeer) -> None:
        super().register_peer(peer)
        peer.connection.add_behavior(NewBlockWitnessHashes, self._handle_new_witness)

    def _handle_new_witness(self, command: NewBlockWitnessHashes):
        # do thing with the command
```

pipermerriam: Hrm, we don't yet have a `PeerSubscriber` replacement which is the only current way to listen across all peers.  So yes, best way is to use the peer subscriber API to get access to each peer as it connects and then attach a `CommandHandler` for the `NewBlockWitnessHashes` to each peer as they come online.  I would not recommend using the `PeerSubscriber` APIs that are for subscribing to commands because those are probably about ready to be removed so make sure you use the `p2p.logic.CommandHandler` APIs for that part and only use the `PeerSubscriber` APIs to get access to the connected/disconnected events.

pipermerriam: Here's an equivalent from the `ETH` protocol that listens for `NewBlock`

https://github.com/ethereum/trinity/blob/39a2d1eddeff0b9313677b53037b48ebcd54c527/trinity/protocol/eth/api.py#L26-L41

carver: > make sure you use the `p2p.logic.CommandHandler` APIs for that part and only use the `PeerSubscriber` APIs to get access to the connected/disconnected events.

How do I find out when a new peer becomes available, so I can attach the new behavior using `add_behavior()`?

pipermerriam: You would still use this API: https://github.com/ethereum/trinity/blob/86e43338e43d0130d471ef76b7db61cc85bc3c15/p2p/peer.py#L348

carver: Hm, the more I read through the CommandHandler stuff the more I think it makes it difficult to add runtime handlers. If I do something like `peer.eth_api.add_child_behavior()` when I get a `register_peer` call, then I think this code will have already run:

https://github.com/ethereum/trinity/blob/992226b5036448b3e30f9638a6da73d1471e5a98/p2p/logic.py#L77-L89

So I think the *new* child behavior will be ignored, because the `Application` has already been applied to the connection. Do I have that right?

Maybe we need a new API on the peer (or connection or ?) to enable adding handlers at runtime.

pipermerriam: I'm all for improvements to these APIs.  They were developed in a semi-vacuum and definitely have room to be improved based on real use cases.

For the specific issue you are referencing, I think you are correct that using `add_child_behaviour()` would not work as expected.  I'm inclined towards making `Application` reject any calls to `add_child_behavior()` once it has already been attached to a connection.

trinity/sync/beam/importer.py
```diff
@@ -360,29 +362,67 @@ def _broadcast_import_complete(
...
+def slice_hashes(hash_list: bytes) -> Iterable[Hash32]:
+    """Split a concatenaded stream of hashes into an iterable list of hashes"""
+    for index in range(0, len(hash_list), 32):
+        next_hash = hash_list[index:index + 32]
+        if len(next_hash) != 32:
+            raise TypeError(
+                f"Value is not a hash, because it's only {len(next_hash)} long: {next_hash}"
+            )
+        else:
+            yield cast(Hash32, next_hash)
```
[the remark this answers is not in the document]
carver: I kind of like having it clear that this is only a type assertion and not an object instantiation, but only a mild preference.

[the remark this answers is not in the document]
cburgdorf: The thing about the `cast` is that it will allow *anything* to be passed. Think of it as just turning to  `mypy` saying *"Get away mypy, trust me, I know what I'm doing"*. So you can write `cast(Hash32, "foo")` and it will happily accept it even though it is obviously wrong and will lead to an error at runtime.

On the other hand you can not write `Hash32("foo")`. Mypy will reject this with.

>"Hash32" has incompatible type "str"; expected "bytes"

trinity/sync/beam/chain.py
```diff
@@ -142,12 +146,13 @@ def __init__(
             self.cancel_token,
         )
 
-        self._backfiller = BeamStateBackfill(db, peer_pool, token=self.cancel_token)
+        self._witness_collector = BeamStateWitnessCollector(db, peer_pool, self.cancel_token)
```
gsalgado, line 149: Did you actually intend to replace the backfiller with the witness collector or was this meant to be reverted before merging?

carver: I think this was a kind of a hack. I don't really remember, since it's from ~1 year ago. The basic idea was probably so that we could parallelize the requests for the witness data. But I think that's not so important, based on how many nodes we'll need to collect per block (maybe one or two full `GetNodeData` requests). So it would probably be best to simplify the witness collector a lot, and just trigger "urgent" requests for data.

PR #1251, opened 2019-11-05
Spec v0.9.0 updates - `eth2` part

### LINKED ISSUE #1240: Spec v0.9.0 updates - `eth2` part
### What is wrong?
https://github.com/ethereum/eth2.0-specs/releases/tag/v0.9.0
- features
	- alter beacon proposer selection logic (ethereum/eth2.0-specs#1413)
	- add seed domains (ethereum/eth2.0-specs#1415)
	- fix delay-based attestation inclusion reward (ethereum/eth2.0-specs#1434)
	- use eth1_block_hash for initial randao_mix (ethereum/eth2.0-specs#1447)
- simplifications
	- remove active active_index_roots and compact_committees_roots from phase 0 (ethereum/eth2.0-specs#1329)
	- remove shards/crosslinks from phase 0 (ethereum/eth2.0-specs#1428)
	- remove transfers from phase 0 (ethereum/eth2.0-specs#1443)
- Fork choice
	- verify state root in fork choice tests and other minor fixes (ethereum/eth2.0-specs#1454)

### How can it be fixed

Question: should the v0.9 progress on the new branch, so that `interop` tests can still work?
@ralexstokes

hwwhww:
### What was wrong?
Close #1240 

### How was it fixed?
1. Update to v0.9.0
2. Bump test suite, and it passed! 🎉

### To-Do
- [x] Clean up. There are many repeated codes in several functions.
- [x] Not quite sure the "verify state root in fork choice tests and other minor fixes" issue (ethereum/eth2.0-specs#1454). @ralexstokes Am I right that you implemented in a different scheme, and the block signatures are already verified before applying fork choice rule?

- [ ] Clean up commit history

#### Cute Animal Picture

wolf-2145949_640


eth2/beacon/tools/builder/validator.py
```diff
@@ -545,35 +531,33 @@ def _create_mock_signed_attestation(
...
-    )
-    for i in range(committees_per_slot):
-        shard = (slot_start_shard + i) % config.SHARD_COUNT
-        committee = get_crosslink_committee(
-            state, epoch, shard, CommitteeConfig(config)
-        )
-        results.append((committee, Shard(shard)))
-
-    return tuple(results)
+# # TODO(ralexstokes) merge in w/ ``get_committee_assignment``
```
hwwhww: Ah it can be removed now.

eth2/beacon/typing.py
```diff
@@ -16,10 +16,11 @@ def __str__(self) -> str:
...
 
-ValidatorIndex = NewType("ValidatorIndex", int)  # uint64
-CommitteeIndex = NewType(
-    "CommitteeIndex", int
-)  # uint64 The i-th position in a committee tuple
+CommitteeIndex = NewType("CommitteeIndex", int)  # uint64, a committee index at a slot
+ValidatorIndex = NewType("ValidatorIndex", int)  # uint64, a validator registry index
+CommitteeValidatorIndex = NewType(
+    "CommitteeValidatorIndex", int
+)  # uint64, the i-th position in a committee tuple
```
hwwhww: Note that, the *old* `CommitteeIndex` (not in the spec) has been renamed to `CommitteeValidatorIndex`: the i-th position in a committee, where committee is `Sequence[ValidatorIndex]`.
And this PR introduces the new `CommitteeIndex`: a slot may have multiple committees, `CommitteeIndex` is the index of a certain committee.

eth2/beacon/committee_helpers.py
```diff
@@ -235,68 +189,61 @@ def _compute_committee(
...
-        compact_balance = validator.effective_balance // effective_balance_increment
-        # `index` (top 6 bytes) + `slashed` (16th bit) + `compact_balance` (bottom 15 bits)
-        compact_validator = (index << 16) + (validator.slashed << 15) + compact_balance
-        compact_validators += (compact_validator,)
-
-    return CompactCommittee(pubkeys=pubkeys, compact_validators=compact_validators)
-
-
-def get_compact_committees_root(
+def iterate_committee_at_epoch(
```
hwwhww: I made `iterate_committee_at_epoch` and `iterate_committee_at_slot` to refactor the places we iterate the committees.

hwwhww:
@ralexstokes 
It's ready to review now! It's huge, but mostly about deletions. Some main points for the reviewers besides #1240:
1. https://github.com/ethereum/trinity/blob/3c7f5317949b0db15d264b7e7e704fdd04cb4e7e/eth2/beacon/committee_helpers.py:
    1. New committee grouping:
        * `get_committee_count_at_slot()` returns committees per slot. Note that if there are no enough validators, some slots may not have any committee.
        * If one slot has more than one committee, we use `CommitteeIndex` to identify them.
    2. proposer selection: the proposer no longer selected from the first committee of the slot.
    3. `iterate_committee_at_epoch` and `iterate_committee_at_slot` are helpers for iterating the committees.
2. Updated [`test_serenity_epoch_processing`](https://github.com/ethereum/trinity/blob/3c7f5317949b0db15d264b7e7e704fdd04cb4e7e/tests/eth2/core/beacon/state_machines/forks/test_serenity_epoch_processing.py#L188-L276) assertions.
3. Question: Not quite sure the "verify state root in fork choice tests and other minor fixes" issue (ethereum/eth2.0-specs#1454). @ralexstokes Am I right that you implemented in a different scheme, and the block signatures are already verified before applying fork choice rule?

Let me know if you have any questions and let's merge it ASAP! :D

tests/eth2/core/beacon/state_machines/forks/test_serenity_epoch_processing.py
```diff
@@ -351,139 +263,26 @@ def test_get_attestation_deltas(
...
         )
     state = state.copy(previous_epoch_attestations=prev_epoch_attestations)
 
     rewards_received, penalties_received = get_attestation_deltas(state, config)
-
-    # everyone attested, no penalties
-    assert sum(penalties_received) == 0
-    the_reward = rewards_received[0]
-    # everyone performed the same, equal rewards
-    assert sum(rewards_received) // len(rewards_received) == the_reward
```
hwwhww: Do you remember this assertion? The block proposers would get extra rewards in `get_attestation_deltas`, so I don't think this assertion is correct. Remove it now.

ralexstokes:
@hwwhww re: your question on the fork choice -- i will verify first thing in the morning after the eth2 call :) and then review this right after that!!

eth2/beacon/typing.py
```diff
@@ -45,6 +46,7 @@ class FromBlockParams(NamedTuple):
 default_slot = Slot(0)
 default_epoch = Epoch(0)
 default_shard = Shard(0)
```
ChihChengLiang: Do we also remove Shard and default_shard?

tests/eth2/core/beacon/conftest.py
```diff
@@ -297,15 +285,13 @@ def config(
     max_attestations,
     max_deposits,
     max_voluntary_exits,
-    max_transfers,
     deposit_contract_address,
 ):
     # adding some config validity conditions here
     # abstract out into the config object?
```
ralexstokes: this comment is no longer relevant and can be removed....

we may want to add configuration invariants (there is an issue somewhere on the specs repo) but that is a different conversation

tests/eth2/core/beacon/state_machines/forks/test_serenity_epoch_processing.py
ralexstokes: hmm you are right! but the test was passing -- which suggests everyone in the validator set in this test was also a proposer (the same number of times as everyone else) -- which isn't that far-fetched... but you are correct we should test this more thoroughly

tests/eth2/core/beacon/test_committee_helpers.py
```diff
@@ -1,158 +1,61 @@
...
-    epoch = random.randrange(config.GENESIS_EPOCH, 2 ** 64)
+def test_compute_proposer_index(genesis_validators, config):
     proposer_index = random.randrange(0, len(genesis_validators))
 
     validators = tuple()
-    # NOTE: validators supplied to ``_find_proposer_in_committee``
+    # NOTE: validators supplied to ``compute_proposer_index``
     # should at a minimum have 17 ETH as ``effective_balance``.
     # Using 1 ETH should maintain the same spirit of the test and
     # ensure we can know the likely candidate ahead of time.
```
ralexstokes: nitpick but this now really is:

```suggestion
     # ensure we can know the candidate ahead of time.
```

eth2/beacon/committee_helpers.py
```diff
@@ -108,69 +78,52 @@ def _find_proposer_in_committee(
...
     """
-    base = int(epoch)
-    committee_len = len(committee)
+    if len(indices) == 0:
+        raise ValidationError("There is no any active validator.")
+
     i = 0
-    while i < MAX_ROUNDS:
-        candidate_index = committee[(base + i) % committee_len]
+    while MAX_ROUNDS:
```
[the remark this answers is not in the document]
ralexstokes: this i think we should clean up before merging

ralexstokes:
@hwwhww i don't think we need to do anything directly related to this spec PR: https://github.com/ethereum/eth2.0-specs/pull/1454

the way the client is currently working we only _accept_ a block if:

1) we have its parent
2) we can successfully perform a state transition (validating signature along the way) and validate that the state root matches

if we do _accept_ a block, then we persist all of the new state (i.e. after applying the block) and the new block -- as we persist the new block, we run the fork choice to determine if there is a new canonical head

so, in short, we have already validated everything we are interested in about the block before the fork choice runs.

eth2/beacon/committee_helpers.py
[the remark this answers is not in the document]
hwwhww: Ahhh it should have been `while i < MAX_ROUNDS:` in the context, as the above docstring described.
I recalled that was by @ChihChengLiang because it did happend with some simple configuration cases (not enough validators). @ChihChengLiang What do you think? 

```suggestion
    while i < MAX_ROUNDS:
```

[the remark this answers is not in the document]
ChihChengLiang: I would say keeping `MAX_ROUNDS` to make the diffs in this PR smaller. The `MAX_ROUNDS` approach has the same effect of `while True` approach in the happy path, but if something goes wrong the former raises Error that helps to debug easier.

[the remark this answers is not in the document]
ChihChengLiang: Update: Reviewed the spec and I changed my position. The fact that it is possible that a validator has their balanced slashed but is still eligible to be a proposer breaks the probability assumption. It's very likely to go over 100 rounds, considering the case that you have 100 more slashed validators in a row.

Issue #1186, opened 2019-09-19
Include the new Istanbul fork in the Mainnet genesis json


carver:
This depends on: https://github.com/ethereum/py-evm/issues/1839 being closed and released.

PR #1255, opened 2019-11-06
Upgrade to py-evm 0.3.0a8

cburgdorf:
### What was wrong?

Trinity needs to be upgraded to the latest py-evm version to:

1. Support the Istanbul fork
2. Allow using the new Clique Consensus engine

### How was it fixed?

- `setup.py` voodoo
- `mainnet.json` vodoo
- Using new `IstanbulVM` where appropriate
- updated fixtures to latest version (because latest py-evm is incompatible with previous fixtures) and resolved some minor issues.

#### Cute Animal Picture


tests/json-fixtures-over-rpc/test_rpc_fixtures.py
```diff
@@ -54,21 +58,20 @@
...
-    'stQuadraticComplexityTest/Return50000_d0g1v0.json',
-    'stQuadraticComplexityTest/Callcode50000_d0g1v0.json',
-    'stQuadraticComplexityTest/Call50000_sha256_d0g1v0.json',
-    'stQuadraticComplexityTest/Call50000_ecrec_d0g1v0.json',
-    'walletReorganizeOwners',
-    'bcExploitTest/SuicideIssue.json',
+    'randomStatetest94_Homestead',
+    'randomStatetest94_Byzantium',
+    'randomStatetest94_Constantinople',
+    'randomStatetest94_ConstantinopleFix',
```
cburgdorf: These :point_up: four tests kept crashing CI (they work locally) and they were also put into the *slow tests* on the Py-EVM side.

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
```diff
@@ -466,6 +481,10 @@ class MainnetFullChain(FullChain):
             if get_in(['blockHeader', 'hash'], block_fixture) == chain_fixture['lastblockhash']:
                 await validate_last_block(rpc, block_fixture)
 
-    await validate_accounts(rpc, chain_fixture['postState'])
+    # Fixtures do not include the `postState` field when its size would be past a certain limit.
+    # https://github.com/ethereum/tests/issues/637#issuecomment-534072897
+    if chain_fixture.get('postState', None):
```
cburgdorf: @carver I was a bit worried if we are losing any safety here but it seems that we are doing the same on the Py-EVM side and I *think* that we are still safe here because even if we do not validate the `postState` for these special cases we do validate the `lastblockhash`.

trinity/_utils/eip1085.py
```diff
@@ -173,11 +173,8 @@ def _filter_vm_config(vm_config: VMConfiguration) -> Iterable[VMFork]:
     ByzantiumVM,
     ConstantinopleVM,
     PetersburgVM,
+    IstanbulVM,
 )
-ALL_VMS_BY_FORK = {
-    vm_class.fork: vm_class
-    for vm_class in ALL_VMS
-}
```
cburgdorf: This was just dead code

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
```diff
@@ -91,15 +94,20 @@
...
     'stQuadraticComplexityTest/Call50000_identity2_d0g1v0.json',
     'stQuadraticComplexityTest/Call50000_identity_d0g1v0.json',
     'stQuadraticComplexityTest/Call50000_rip160_d0g1v0.json',
     'stQuadraticComplexityTest/Call50000bytesContract50_1_d0g1v0.json',
     'stQuadraticComplexityTest/Call50000bytesContract50_2_d0g1v0.json',
     'stQuadraticComplexityTest/Create1000_d0g1v0.json',
-    'ShanghaiLove_Homestead',
-    'ShanghaiLove_Frontier',
-    'DelegateCallSpam',
+    'walletReorganizeOwners',
```
cburgdorf: Don't get confused, I just resorted this list alphabetically (in a separate commit), so there are really only four new tests in that list.

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
carver: IIRC, veox said something about the tests having high RAM usage.

voith: @cburgdorf That's correct. Those tests need more than 4gb of RAM(discovered [here](https://github.com/ethereum/py-evm/pull/1852#issuecomment-533900250)). And they were marked as slow in this [commit](https://github.com/ethereum/py-evm/pull/1852/commits/4b96de70d25591c64745e1421ece5fbdbb7d2e33).

carver:
Closed #1186

PR #1257, opened 2019-11-07
Deduplicate CI jobs

cburgdorf:
### What was wrong?

We currently run the `trinity_cli` tests as part of the `pyxx-wheel-cli` jobs as well as part of the `pyxx-integration` jobs. This is wasting resources especially since these tests can sometimes be a little flaky.

### How was it fixed?

Exclude them from the `py-xx-integration` jobs

#### Cute Animal Picture


carver:
> Exclude them from the `py-xx-wheel-cli` jobs

This description looks inconsistent with the PR code. The tests are excluded from the pyxx-integration jobs, instead, right? (The in-PR way is my preference, since the wheel test is the most "true to final" integration test)

PR #1258, opened 2019-11-07
Remove an one-off script that was accidentally merged via eth2 interop work

ralexstokes:
### What was wrong?

See: https://github.com/ethereum/trinity/issues/1218#issuecomment-545907237 and the containing issue.

### How was it fixed?

Remove a script we don't really need
#### Cute Animal Picture


ralexstokes:
as far as i know, this is a known failure for CI -- going ahead w/ merge

PR #1259, opened 2019-11-08
Experiment with an asyncio based service that doesn't use CancelToken

pipermerriam:
This is me running with an idea.  Here's an `asyncio` based service that doesn't use a `CancelToken` and yields clean shutdowns and resilient exception handling (the same pattern used in the `TrioService`).  I'm not really sure what this proves but it *might* be a way to reduce or eliminate our use of cancel tokens...

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/asyncio_service.py
+++ b/p2p/asyncio_service.py
@@ -1,0 +1,362 @@
+import asyncio
+import logging
+import sys
+from types import TracebackType
+from typing import (
+    Any,
+    AsyncIterator,
+    Awaitable,
+    Callable,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Type,
+)
+
+from async_generator import asynccontextmanager
+
+from trio import MultiError
+
+from p2p.trio_service import (
+    DaemonTaskExit,
+    LifecycleError,
+    ServiceAPI,
+    ManagerAPI,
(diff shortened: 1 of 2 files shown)


pipermerriam:
So this came from a few *connected* rabbit holes.

I wanted to experiment with using [`asyncio-run-in-process`](https://asyncio-run-in-process.readthedocs.io/en/latest/) to replace the usage of `multiprocessing` within our *Component* API.  This lead to me realizing that there wasn't a clean way for me to enter an `async` execution.

So, I went to look at the `trinity.extensibility.ComponentManager` to see if there was a clear way to cleanup the Component lifecycle, simplify the Component API and allow components to define async entry points.

This lead me to look into whether the `ComponentManager` could be replaced by a service since services already have mechanisms in place for managing the lifecycle of multiple concurrent background tasks....

And this brought me to re-examine the `TrioService` API since I wanted to use that same API for the `ComponentManager`.  After looking through the `TrioService` it seemed feasible to reproduce the same API.

So here it is, a *different* shaped service abstraction that is `asyncio` friendly and that I *think* might manage clean cascading shutdowns of hierarchical nested services without needing the cancel token....

PR #1265, opened 2019-11-12
[WIP] - Experiment with full removal of cancel token and multiprocessing.

pipermerriam:
### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


pipermerriam:
Just a brain dump of what I've learned here.

We can replace the cross-process logging mechanism which is actually based on the `multiprocessing` library with a custom implementation that uses roughly the same pattern as our `DBManager`.  This is necessary because you cannot pass a `multiprocessing.Queue` across a process boundary unless you are using the `multiprocessing` utils.

With the cross process logging tools replaced we can then use `asyncio-run-in-process`.  This opens the door for an improved `Component` API.

The Component API can be cleaned up significantly.  [This can be seen here](https://github.com/ethereum/trinity/blob/dda23ae64ec8d37dc3f2b191bb71efedc0a8e7a5/trinity/extensibility/component.py).  The component APIs end up being slimmed down to just the following two APIs:

```python
class BaseComponentAPI(ABC):
    def configure_parser(cls, arg_parser: ArgumentParser, subparser: _SubParsersAction) -> None:
        ...

    def validate_cli(cls, boot_info: TrinityBootInfo) -> None:
        ...

class ApplicationComponentAPI(BaseComponentAPI):
    name: str

    def __init__(self, trinity_boot_info: TrinityBootInfo, endpoint: EndpointAPI) -> None:
        ...

    @property
    def is_enabled(self) -> bool:
        ...

    async def run(self) -> None:
        ...
```

The base API is all that is needed for components like `fix-unclean-shutdown`.  The `ApplicationComponentAPI` is what is needed for things like the transaction pool or other components that run concurrently. 

The simplification to the component APIs also extends to the `ComponentManager` class which is easier modeled as just a plain `Service`.  This allows for easy concurrent spin-up of components, **however** this might expose a deficiency in our component APIs as some components are dependent on others and we probably need something to allow declaring dependencies between them so that they can be started in the correct order.

The new `Service` model doesn't use a `CancelToken`.  Assuming that this works, this allows removal of a *lot* of machinery like the `trinity._utils.shutdown` and things like `kill_process_gracefully`.  The main logic for the whole application can then be coded as just another `Service` and the startup logic for the whole application becomes as simple as running the service and installing some signal handlers at the outermost layer to catch `SIGTERM` or `KeyboardInterrupt` and trigger shutdowns in a clean manner.

pipermerriam:
Alright, after another day and a decent bit of hacking at the `async-run-in-process` to get it behaving correctly, I believe I've validated my hypothesis.

I can run the Trinity client and get clean shutdowns without any cancel tokens stuff.

There's still a lot of cleanup to be done but I'm leaning towards this being a viable path forward.

One cool thing that came from this is that we are now completely free of `multiprocessing` and any related magic.  All of the issues I ran into while debugging this branch were now in parts of code that I was in control of, meaning that I could actually fix them so that they shutdown cleanly rather than having to jump through hoops to try and fix them with indirection.  This included a decent bit of work in the `asyncio-run-in-process` to get it handling child processes correctly.

I feel pretty good about this.  I think it gives us a level of control over the application that we previously didn't have, specifically in managing the lifecycles of things which is important if we want users to have a nice UX.

PR #1266, opened 2019-11-12
Add export and import commands

cburgdorf:
### What was wrong?

In geth (and probably also in other clients there exist `export` and `import` commands to let users export blocks from the database into files or import blocks files into the canonical chain.

Trinity does currently lack this functionality and I recently [became aware](https://github.com/ethereum/trinity/pull/972#issuecomment-551863172) that it would help @adamschmideg to build something if Trinity would expose that functionality.

### How was it fixed?

Added an `export` command

```
trinity export --help
usage: trinity export [-h] [--append] [--overwrite] file_path block_number

positional arguments:
  file_path     Specify the file to export to
  block_number  Specify the block number to be exported

optional arguments:
  -h, --help    show this help message and exit
  --append      Disable peer discovery
  --overwrite   Disable peer discovery

```
 Added an `import` commands.

```
$ trinity import --help
usage: trinity import [-h] file_path

positional arguments:
  file_path   Specify the file to import from

optional arguments:
  -h, --help  show this help message and exit
 
```

Notice that the `import` command can handle to load multiple blocks from a single file and is compatible with geth. E.g. one can export a range of blocks from `geth` and import them in Trinity using this command.

### To-Do

- [x] Write some tests
- [ ] Update docs

#### Cute Animal Picture


adamschmideg:
It helps writing GraphQL tests across clients. What the tests do: init a new chain, import some blocks, run graphql queries, check the results.

cburgdorf:
@adamschmideg yep, makes perfect sense. I also think it's a great primitive to have for every client and I like us to catch up with other clients on these little details :)

From your linked files, it seems like you are importing a range of blocks from a single file. I still have to tackle that part. In the current state of the PR only a single block can be exported and imported per file. But I should have that finished soon :)

trinity/components/builtin/import_export/component.py
```diff
@@ -0,0 +1,174 @@
...
+        )
+
+        import_parser.set_defaults(func=cls.run_import)
+
+    @classmethod
+    def run_import(cls, args: Namespace, trinity_config: TrinityConfig) -> None:
+        logger = cls.get_logger()
+
+        with open(args.file_path, 'rb') as import_file:
+            file_bytes = import_file.read()
```
cburgdorf: Let me point out that we are consuming the whole file at once and that this will crumble with super large files. This could use a streaming approach but I also think it is better than nothing and ok to begin with :)

trinity/components/builtin/import_export/rlp_decode.py
```diff
@@ -0,0 +1,55 @@
...
+)
+
+import rlp
+from rlp.codec import _apply_rlp_cache, consume_item
+from rlp.exceptions import DecodingError
+from rlp.sedes.lists import is_sequence
+
+
+@to_tuple
+def decode_all(rlp: bytes,
```
cburgdorf: This could also be nicer and maybe even take in a stream. I kinda just did the minimal work that I needed for this PR but  I'm happy to create a more polished PR upstream if we decide what kind of support would be optimal for the py-rlp library.

trinity/components/builtin/import_export/component.py
pipermerriam: Maybe worth denoting with a comment here and opening an issue to track this (and linking to said issue in the comment)

trinity/components/builtin/import_export/component.py
```diff
@@ -0,0 +1,174 @@
...
+
+        logger.info("Importing %s blocks", len(blocks))
+
+        chain = get_chain(trinity_config)
+
+        for block in blocks:
+            try:
+                chain.import_block(block)
+            except (EVMMissingData, ValidationError) as exc:
+                logger.error(exc)
```
pipermerriam: Seems like we would want this action to be fully atomic which suggests that we should be using a the `atomic_batch` API for interacting with the database....

trinity/components/builtin/import_export/component.py
```diff
@@ -0,0 +1,174 @@
...
+        )
+
+        export_parser.add_argument(
+            "--append",
+            action="store_true",
+            help="Disable peer discovery",
+        )
+
+        export_parser.add_argument(
+            "--overwrite",
```
pipermerriam: I am inclined to suggest an alternate API which I *think* conforms more closely with expectations of CLI tools but this is of course open for debate.

- `$ trinity export 1`
    - exports block 1, dumps RLP to stderr/stdout
    - usage would be `trinity export 1 > block_1.rlp
- `$ trinity export 1 --outfile block_1.rlp
    - exports block 1 to the file `block_1.rlp`
```

Basically, I think the default behavior should be dumping to terminal, and that dumping to a file should be a flag based option.

File: trinity/components/builtin/import_export/component.py [hunk 5]
```diff
@@ -0,0 +1,174 @@
...
+        if args.file_path.exists() and not (args.append or args.overwrite):
+            logger.error(
+                "%s does exist. Must use `--append` or `--overwrite` to proceed.", args.file_path
+            )
+            return
+
+        parent_dir = args.file_path.parent
+
+        if not parent_dir.exists():
+            parent_dir.mkdir(parents=True)
```
pipermerriam: I don't think this is the right behavior.  If the directory doesn't exist I think it should be an error.

trinity/components/builtin/import_export/component.py
cburgdorf: I guess I might agree if we were talking about human readable data but since this is just binary I wonder about the practical purpose of dumping it on the terminal. In general, I think the current behavior has a better UX for the user (and is basically identical to `geth`).

trinity/components/builtin/import_export/component.py
cburgdorf: You mean, either import all blocks or none in case of errors? But that would be a new Py-EVM API I guess. I mean, I can not use the raw DB APIs here, since I need to actually execute the blocks. Or am I misreading your comment?

trinity/components/builtin/import_export/component.py
cburgdorf: I was thinking about it but went for the nicer UX. Like, what is the worst thing that can happen? It's not like we could overwrite any user data. We are just creating intermediate directories that the user told us to create.

We know that we have handled it like that in the past e.g. when you run Trinity with `--data-dir /somethign/custom` then the first thing you'll be greeted with is that Trinity can not write the logs because the log directory doesn't exist. So the user has to create it manually first. But, the more I think about this, the more I question this default. It feels *overly* cautious to the point where it just makes the user experience less pleasant I would say.

If we are not overwriting anything, what harm could we be doing?

trinity/components/builtin/import_export/component.py
pipermerriam: I don't feel strongly about this but here's a potential argument for dumping to terminal (albeit that the data is binary).

```python
$ trinity export 1-1000 | trinity --dev import
```

In the above example the data for the first 1k blocks would be dumped to the terminal and then read by the second trinity process from stdin and imported into some "dev" alternate dev chain.

My argument could maybe be summed up as trying to stick with Unix-y philosophies or something.... do as you please.

cburgdorf: Yes, good point. I generally agree that the Unix philosophy of composability of commands is great and to be embraced, but I'd also argue that we should optimize the UX for the *common* case. I guess exporting to a file is the common case when using `export` and it might even leave users confused when they'd run the command and the terminal fills with tons of weird characters. So in this case, I'd rather add an additional `--stdout` flag but leave the default to export to a file.

PR #1267, opened 2019-11-13
Improve stability and debug-ability of integration tests

cburgdorf:
### What was wrong?

1. @lithp figured out a while ago that some of our integration tests fail because a previous integration test may leave the Trinity default port blocked and so it is beneficial to always run these tests with random ports assigned. When that was discovered it was only applied to one of the integration tests but I just figured out that it is helpful to nearly all of them. E.g. the test that ensures our logging config works as expected may fail sporadically because it can not find the logging output from the `DiscoveryService` that is is expecting to be there. Surprise surprise, the `DiscoveryService` sometimes fails to boot when the port is taken.

2. I turned on full logging output for these tests so that everyone has an easier time figuring out what went wrong if one of these tests fails

3. One of the main problems why the web3 integration test usually fails is because requests time out. I believe this is a real problem not with the tests but with Trinity.

### How was it fixed?

1. Use an unused port for nearly all integration tests
2. Turn on logging output for the integration tests
3. I've cut down the work that Trinity performs in the web3 integration tests. I disabled sync, discovery and upnp leaving more resources to reliably answer these JSON-RPC requests.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/integration/test_trinity_cli.py
+++ b/tests/integration/test_trinity_cli.py
@@ -37,6 +37,13 @@ MAINNET_GENESIS_HASH = '0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c
 #         assert await contains_all(runner.stdout, {'bytes from'})
 
 
+def amend_command_for_unused_port(command, unused_tcp_port):
+    # use a random port each time, in case a previous run went awry and left behind a
+    # trinity instance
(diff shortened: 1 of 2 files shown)


cburgdorf:
Anyone up to review this? It saves us a bunch of manual CI triggering per day :)

PR #1268, opened 2019-11-13
Fix quickstart genesis

hwwhww:
### What was wrong?
The current `resources/genesis.ssz` is generated by our own key generator. It would be nice to reuse interop resources.

### How was it fixed?
1. Fix quickstart scripts
2. Generate `genesis.ssz` by using the common keys

### To-Do

- [ ] Clean up commit history

- [ ] ~~(not sure if it's related) `run_beacon_nodes.py` still doesn't finalize epoch. :'(~~

#### Cute Animal Picture

bull-walrus-1030287_640


eth2/beacon/scripts/quickstart_state/interop_keygen.py
```diff
@@ -0,0 +1,38 @@
...
+            "pubkey": encode_hex(bls.privtopub(privkey)),
+        }
+
+
+if __name__ == "__main__":
+    yaml = YAML(pure=True)
+    yaml.default_flow_style = None
+    n = 16
+    keypairs = generate_validator_keypairs(n)
+    key_file = KEY_DIR / Path(f"keygen_{n}_validators.yaml")
```
ralexstokes: this is a handy script to have :)

Issue #1209, opened 2019-09-30
Etherscan-based checkpoint is hard-coded to mainnet


carver:
### What is wrong?

AFAICT, if you beam sync with a checkpoint, the etherscan option always uses mainnet headers, even if you are syncing ropsten.

### How can it be fixed

Call a different API, depending on the active network.

In the short term, we could just quit out immediately, and say that you need to use the other style of checkpoint URI.

cc @cburgdorf

PR #1269, opened 2019-11-13
Properly handle different networks when resolving checkpoints

cburgdorf:
### What was wrong?

As mentioned in #1209, currently when using `--beam-from-checkpoint eth://block/byetherscan/latest` with non-mainnet networks it will silently try to do the wrong thing.

### How was it fixed?

Use the correct Etherscan API for every supported network. Also, tell the user when a network does not support this API.

**Bonus**: There were some leaking exceptions that I turned into a proper user facing error. That's in a separate commit but I sneaked it into this PR.

### To-Do

#### Cute Animal Picture

### DIFF
--- a/trinity/bootstrap.py
+++ b/trinity/bootstrap.py
@@ -103,7 +103,7 @@ def main_entry(trinity_boot: BootFn,
     args = parser.parse_args()
 
     if not args.genesis and args.network_id not in PRECONFIGURED_NETWORKS:
-        raise NotImplementedError(
+        parser.error(
             f"Unsupported network id: {args.network_id}. To use a network besides "
             "mainnet or ropsten, you must supply a genesis file with a flag, like "
             "`--genesis path/to/genesis.json`, also you must specify a data "
--- a/trinity/components/builtin/syncer/cli.py
+++ b/trinity/components/builtin/syncer/cli.py
@@ -106,5 +106,9 @@ class NormalizeCheckpointURI(argparse.Action):
                  namespace: argparse.Namespace,
                  value: Any,
                  option_string: str=None) -> None:
-        parsed = parse_checkpoint_uri(value)
+
+        try:
+            parsed = parse_checkpoint_uri(value)
+        except ValidationError as exc:
+            raise argparse.ArgumentError(self, str(exc))
         setattr(namespace, self.dest, parsed)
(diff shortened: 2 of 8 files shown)


trinity/bootstrap.py
```diff
@@ -100,10 +102,13 @@ def main_entry(trinity_boot: BootFn,
 
     argcomplete.autocomplete(parser)
 
-    args = parser.parse_args()
+    try:
+        args = parser.parse_args()
+    except ValidationError as exc:
+        parser.error(str(exc))
```
carver: Maybe we can get the whole stack trace in here, since an exception that bubbles up this far seems unintended.

cburgdorf: >since an exception that bubbles up this far seems unintended.

Oh, I understand! Let me roll this back. I didn't know there exists and actual `argparse.ArgumentError` that should be thrown when a custom argument parsing failed.

Now that I know I went to this other place and made sure it raises an `argsparse.Argumenterror` and then we don't need to catch `ValidationError` here :)

https://github.com/ethereum/trinity/pull/1269/commits/6b8d897aba1a1c2cc0e9febfe012137a0a06547c#diff-4398c8abe4c50ec44530591d4bbc1ee8R113

trinity/components/builtin/syncer/cli.py
```diff
@@ -106,5 +115,9 @@ def __call__(self,
                  namespace: argparse.Namespace,
                  value: Any,
                  option_string: str=None) -> None:
-        parsed = parse_checkpoint_uri(value)
+
+        try:
+            parsed = parse_checkpoint_uri(value, namespace.network_id)
```
pipermerriam: not part of this PR but what if we used our own type for `namespace` instances.

```python
class TrinityNamespace(argparse.Namespace):
    network_id: int
    disable_upnp: Optional[bool]
    ...
```

Then we could at least be explicit about what properties it has and have stronger types in our code that accesses them.

trinity/components/builtin/syncer/etherscan_api.py
```diff
@@ -9,19 +10,37 @@
...
-ETHERSCAN_PROXY_API_URL = f"{ETHERSCAN_API_URL}?module=proxy"
-
-
 class EtherscanAPIError(BaseTrinityError):
     pass
 
 
-def etherscan_post(action: str) -> Any:
-    response = requests.post(f"{ETHERSCAN_PROXY_API_URL}&action={action}")
+class Network(enum.Enum):
```
pipermerriam: if you use `enum.IntEnum` then you'll be able to do things like `if network_id == Network.MAINNET`

Issue #1241, opened 2019-10-29
Spec v0.9.0 updates - simple naive aggregation strategy 


hwwhww:
### What is wrong?

See https://github.com/ethereum/eth2.0-specs/pull/1440

### How can it be fixed

Fill this in if you know how to fix it.

hwwhww:
I started to implement the aggregator tools.

PR #1271, opened 2019-11-13
Add some aggregator helpers

hwwhww:
### What was wrong?
Add some aggregator helpers for https://github.com/ethereum/trinity/issues/1241
See https://github.com/ethereum/eth2.0-specs/blob/v09x/specs/validator/0_beacon-chain-validator.md#attestation-aggregation

### How was it fixed?
1. `slot_signature`: a function for the attester to sign on the slot. Then the attester would use this signature as the lottery ticket of aggregators selection of the given slot.
2. `is_aggregator`: a probabilistic function. The attester may get selected with the expected aggregator count per committee = `TARGET_AGGREGATORS_PER_COMMITTEE`. Here is the graph of the expected `len(selected_validators)`: https://docs.google.com/spreadsheets/d/1C7pBqEWJgzk3_jesLkqJoDTnjZOODnGTOJUrxUMdxMA/edit#gid=0
3. Add `AggregateAndProof` SSZ container.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

Dugong

### DIFF
--- a/eth2/beacon/types/defaults.py
+++ b/eth2/beacon/types/defaults.py
@@ -21,6 +21,7 @@ if TYPE_CHECKING:
 
 
 default_bls_pubkey = EMPTY_PUBKEY
+default_bls_pubkey = EMPTY_PUBKEY
 
 # NOTE: there is a bug in our current version of ``flake8`` (==3.5.0)
 # which does not recognize the inline typing:
--- a/eth2/beacon/types/defaults.py
+++ b/eth2/beacon/types/defaults.py
@@ -21,7 +21,6 @@ if TYPE_CHECKING:
 
 
 default_bls_pubkey = EMPTY_PUBKEY
-default_bls_pubkey = EMPTY_PUBKEY
 
 # NOTE: there is a bug in our current version of ``flake8`` (==3.5.0)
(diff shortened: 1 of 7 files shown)


eth2/beacon/types/defaults.py
```diff
@@ -20,6 +20,7 @@
     from typing import Any  # noqa: F401
 
 
+default_bls_pubkey = EMPTY_PUBKEY
```
hwwhww:
```suggestion
```

tests/eth2/core/beacon/tools/builder/test_aggregator.py
```diff
@@ -0,0 +1,41 @@
...
+            if index in committee:
+                signature = slot_signature(genesis_state, slot, privkeys[index], config)
+                attester_is_aggregator = is_aggregator(
+                    state, slot, committee_index, signature, config
+                )
+                if attester_is_aggregator:
+                    aggregator_count += 1
+        assert aggregator_count > 0
+        sum_aggregator_count += aggregator_count
+    assert (
```
ChihChengLiang: This is not 100% guaranteed, right? Should we leave a comment here?

PR #1272, opened 2019-11-13
Refactor and update `SlotTicker`

hwwhww:
### What was wrong?
Part of #1241: `SlotTicker` changes
- Before this PR,  `SlotTicker` should tick twice in one slot: one the beginning, the other is at half of the slot.
- For the new aggregation strategy, we need three ticks per slot:
    1. For validators to check if they are the proposer. If so, propose the block
    2. **[TODO in other PR]** For validators to attest the block. If so, broadcast the attestations to gossipsub attestation subnet
    3. **[TODO in other PR]** For validators to check if they are one of the aggregators of the subnet. If so, aggregate the attestations and then broadcast it to the global channel.

### How was it fixed?
1. Use `TickMessage` to refactor Event message
2. Now there are three ticks per slot:
    1. `SLOT_START`: at the beginning of the slot
    2. `SLOT_ONE_THIRD`: at 1/3 of the slot
    3. `SLOT_TWO_THIRD`: at 2/3 of the slot
3. For now, to implement the part 1 of the new mechanism without breaking things, make attesters broadcast to the global channel at 1/3 of the slot.
strategy.
4. Add a stub for 2/3 slot tick.

### To-Do

- [x] Clean up commit history

- [x] E2E test

#### Cute Animal Picture

Sea Cow2

### DIFF
--- a/trinity/components/eth2/beacon/slot_ticker.py
+++ b/trinity/components/eth2/beacon/slot_ticker.py
@@ -1,6 +1,6 @@
 import asyncio
 import time
-from typing import Set
+from typing import Optional, Set
 
 from cancel_token import (
     CancelToken,
@@ -28,10 +28,6 @@ from trinity._utils.shellart import (
 from trinity.components.eth2.misc.tick_type import TickType
 
 
-# Check twice per second: SECONDS_PER_SLOT * 2
-DEFAULT_CHECK_FREQUENCY = 24
-
-
 @dataclass
 class SlotTickEvent(BaseEvent):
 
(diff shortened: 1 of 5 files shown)


trinity/components/eth2/beacon/slot_ticker.py
```diff
@@ -24,16 +25,19 @@
 from trinity._utils.shellart import (
     bold_white,
 )
+from trinity.components.eth2.misc.tick_type import TickType
 
-DEFAULT_CHECK_FREQUENCY = 6
+
+# Check twice per second: SECONDS_PER_SLOT * 2
+DEFAULT_CHECK_FREQUENCY = 24
```
NIC619: nit: should we import `SECONDS_PER_SLOT` from config?

hwwhww: I changed it to be configurable by the `__init__` parameter (https://github.com/ethereum/trinity/pull/1272/commits/1885560f354db5c8bbbef9527d330a5eec5fa717).

PR #1273, opened 2019-11-13
Minor refactor of the peer connection

hwwhww:
### What was wrong?
Since `port = maddr.value_for_protocol(protocols.P_TCP)` in `Node` is actually `str` (!!!), the error message `self.logger.debug("Could not connect to peer %s at %s:%d;", peer_id, ip, port)` would be invalid, but it doesn't show any error message of the wrong format bug!

### How was it fixed?
1. IMO it's more common to use `port: int`, so this PR converts it to `int`.
2. Try to catch error from `self.host.connect()`.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

seal-1235138_640
(doge seal!?)

### DIFF
--- a/trinity/components/eth2/beacon/component.py
+++ b/trinity/components/eth2/beacon/component.py
@@ -117,8 +117,6 @@ class BeaconNodeComponent(AsyncioIsolatedComponent):
             chain=chain,
         )
 
-        self.logger.warning(f'Node listening: {libp2p_node.listen_maddr_with_peer_id}')
-
         receive_server = BCCReceiveServer(
             chain=chain,
             p2p_node=libp2p_node,
--- a/trinity/protocol/bcc_libp2p/node.py
+++ b/trinity/protocol/bcc_libp2p/node.py
@@ -314,6 +314,7 @@ class Node(BaseService):
         self._register_rpc_handlers()
         # TODO: Register notifees
         await self.host.get_network().listen(self.listen_maddr)
+        self.logger.warning("Node listening: %s", self.listen_maddr_with_peer_id)
         await self.connect_preferred_nodes()
         # TODO: Connect bootstrap nodes?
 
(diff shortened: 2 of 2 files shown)


trinity/components/eth2/beacon/component.py
```diff
@@ -117,8 +117,6 @@ def do_start(self) -> None:
             chain=chain,
         )
 
-        self.logger.warning(f'Node listening: {libp2p_node.listen_maddr_with_peer_id}')
```
hwwhww: Move this message to the right place in `BeaconNode`.

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -340,18 +341,24 @@ def _setup_topic_validators(self) -> None:
...
+        except Exception as e:
+            raise ConnectionRefusedError() from e
+
         try:
             # TODO: set a time limit on completing handshake
             await self.request_status(peer_id)
         except HandshakeFailure as e:
             self.logger.info("HandshakeFailure: %s", str(e))
-            # TODO: handle it
+            raise ConnectionRefusedError() from e
```
hwwhww: @NIC619 The error would be handled in `dial_peer_with_retries`. Do you think it's okay?

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -340,18 +341,24 @@ def _setup_topic_validators(self) -> None:
...
+            self.logger.debug("Dialing peer_id %s maddr %s", peer_id, maddr)
+            await self.host.connect(
+                PeerInfo(
+                    peer_id=peer_id,
+                    addrs=[maddr],
+                )
             )
-        )
+        except Exception as e:
+            raise ConnectionRefusedError() from e
```
hwwhww: Not sure what kind of error would be throw from libp2p `swarm` so I'm catching all of it. /cc @NIC619

NIC619: `connect` would raise `SwarmException` if failed to dial the peer.
```python
from libp2p.network.exceptions import SwarmException
```

trinity/protocol/bcc_libp2p/node.py
NIC619: Yeah! After updated to v090 and we cut a new py-libp2p release, I will update these altogether.

PR #1283, opened 2019-11-15
Hashable SSZ Tests

jannikluhn:
:construction: Please do not merge :construction: 

First step of transitioning to py-ssz hashables:

- change all types from serializables to hashable containers
- load the static SSZ tests into hashable lists/vectors/containers instead of tuples/serializables

The SSZ tests pass with the latest commit in https://github.com/ethereum/py-ssz/pull/97.

#### Cute Animal Picture

monkey-on-a-tree-2924649

### DIFF
--- a/tests/eth2/core/beacon/fork_choice/test_lmd_ghost.py
+++ b/tests/eth2/core/beacon/fork_choice/test_lmd_ghost.py
@@ -107,7 +107,7 @@ def _find_collision(state, config, validator_index, epoch, block_producer):
             attestation_data = AttestationData.create(
                 slot=slot,
                 index=committee_index,
-                target=Checkpoint.create(epoch=epoch),
+                target=Checkpoint.create(epoch=epoch, root=root),
                 beacon_block_root=root,
             )
             committee_count = len(committee)
@@ -176,7 +176,7 @@ def _mk_block_at_slot(block_template, slot):
     if slot in block_producer_cache:
         return block_producer_cache[slot]
     else:
-        block = block_template.copy(slot=slot)
+        block = block_template.set("slot", slot)
         block_producer_cache[slot] = block
         return block
 
--- a/tests/eth2/core/beacon/fork_choice/test_lmd_ghost.py
+++ b/tests/eth2/core/beacon/fork_choice/test_lmd_ghost.py
@@ -45,18 +45,13 @@ def _mk_attestation_inputs_in_epoch(epoch, block_producer, state, config):
             # empty committee this slot
             continue
 
-<<<<<<< HEAD
         block = block_producer(slot)
         root = block.signing_root
-        attestation_data = AttestationData(
+        attestation_data = AttestationData.create(
             slot=slot,
             index=committee_index,
-            target=Checkpoint(epoch=epoch, root=root),
+            target=Checkpoint.create(epoch=epoch, root=root),
             beacon_block_root=root,
-=======
-        attestation_data = AttestationData.create(
-            slot=slot, index=committee_index, target=Checkpoint.create(epoch=epoch)
->>>>>>> 34804039... Make most fork choice tests pass
         )
         committee_size = len(committee)
         aggregation_bits = bitfield.get_empty_bitfield(committee_size)
@@ -289,21 +284,6 @@ def test_store_get_latest_attestation(
         pool_attestations_by_index,
     )
 
-<<<<<<< HEAD
-=======
-    # ensure we get the expected results
-    state = state.mset(
-        "previous_epoch_attestations",
-        previous_epoch_attestations,
-        "current_epoch_attestations",
-        current_epoch_attestations,
-    )
-
-    pool = empty_attestation_pool
-    for attestation in pool_attestations:
-        pool.add(attestation)
-
->>>>>>> 34804039... Make most fork choice tests pass
     chain_db = None  # not relevant for this test
     context = Context.from_genesis(genesis_state, genesis_block)
     context.time = some_time
--- a/tests/eth2/core/beacon/fork_choice/test_lmd_ghost.py
+++ b/tests/eth2/core/beacon/fork_choice/test_lmd_ghost.py
@@ -45,6 +45,7 @@ def _mk_attestation_inputs_in_epoch(epoch, block_producer, state, config):
             # empty committee this slot
             continue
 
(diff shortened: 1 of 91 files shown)


jannikluhn:
All tests in `tests/eth2/fixtures` pass now. The transition is surprisingly painless so far.

jannikluhn:
This is mostly done. Changes are almost everywhere pretty simple:

- replace `Serializable` with `HashableContainer`
- replace `__init__` method with `create` method
- replace `SpecObject()` with `SpecObject.create`
- replace `spec_object.copy(a=b, c=d)` with `spec_object.mset("a", b, "c", d)` or for more complicated cases `spec_object.transform(["a", index, "b"], c)`
- remove the `update_validator` method and replace calls with `state.transform(["validators", index], new_validator)`)
- replace the tuple update functions with `hashable_list_or_vector.set` (for setting a value) or `transform` (for updating it with a function)

There are a few problems:

1) Serializables used to not check the size of vectors during initialization so that we could just pass an empty tuple as default for vectors (which would be illegal in the eventual SSZ object, but we usually set these later). This is doesn't work now anymore because we already create the hash tree during initialization, so we have to use the correct size from the very beginning. For now I just replaced them with arbitrary numbers that at least make the tests pass, but this is obviously not a proper solution. I see two ways to deal with it:
  
  - short term: just use the mainnet default for defaults, as proposed by @ralexstokes in another thread. This is somewhat dangerous, because if one overrides sizes but still uses a default or an object created before overriding, one gets weird and hard to debug errors (you end up with two competing sedes objects -- one in the class, the other one in the object).
  - longer term: implement the ssz size config stuff in some form discussed [here](https://github.com/ethereum/trinity/issues/786) and [here](https://github.com/ethereum/py-ssz/issues/98)

  I think I'm leaning towards the short term solution at least for this PR, to not blow up this PR and the corresponding one in py-ssz even more, but then implement the long term one very soon.

2) [This test](https://github.com/ethereum/trinity/blob/f94c1aa1642dd5d83eb6a89e48205abda234de79/tests/eth2/core/beacon/test_attestation_helpers.py#L143) is failing: It checks that indexed attestations with too many indices don't pass verification. It fails right now because it's not even possible to create such an object anymore (because the length is checked at initialization). So I think we can just remove this test, but wanted to double check.

3) [This test](https://github.com/ethereum/trinity/blob/f94c1aa1642dd5d83eb6a89e48205abda234de79/tests/eth2/core/beacon/fork_choice/test_lmd_ghost.py#L260-L263) and a few similar ones are failing: I think the test is just buggy: It tries to set `state.previous_epoch_attestations` with a list of `Attestations` instead of `PendingAttestations`. I guess this should be relatively easy to fix.

4) [This test](https://github.com/ethereum/trinity/blob/f94c1aa1642dd5d83eb6a89e48205abda234de79/tests/eth2/core/beacon/state_machines/forks/test_serenity_block_processing.py#L161) is failing. It tries to create a state with too many `eth1_data_votes`. Can I just remove that one?

ralexstokes:
>     1. [This test](https://github.com/ethereum/trinity/blob/f94c1aa1642dd5d83eb6a89e48205abda234de79/tests/eth2/core/beacon/test_attestation_helpers.py#L143) is failing: It checks that indexed attestations with too many indices don't pass verification. It fails right now because it's not even possible to create such an object anymore (because the length is checked at initialization). So I think we can just remove this test, but wanted to double check.

The test cases here are meant to mirror the checks in the corresponding validity function in the spec. This specific test case verifies that the length of the `attesting_indices` is less than the max. You are correct that this condition has already been verified w/ a _correct_ SSZ implementation. I think it is fine to remove this test -- especially if you are confident that this sort of validity condition is verified in our `py-ssz` unit testing.

>     2. [This test](https://github.com/ethereum/trinity/blob/f94c1aa1642dd5d83eb6a89e48205abda234de79/tests/eth2/core/beacon/fork_choice/test_lmd_ghost.py#L260-L263) and a few similar ones are failing: I think the test is just buggy: It tries to set `state.previous_epoch_attestations` with a list of `Attestations` instead of `PendingAttestations`. I guess this should be relatively easy to fix.

I'm happy to take a closer look -- in particular I've reworked some of that particular test in my #1326 fork choice PR so I'd ignore it for now; you'll be able to rebase on top of the fork choice updates soon! I'm not sure which other tests are acting up right now.

>     3. [This test](https://github.com/ethereum/trinity/blob/f94c1aa1642dd5d83eb6a89e48205abda234de79/tests/eth2/core/beacon/state_machines/forks/test_serenity_block_processing.py#L161) is failing. It tries to create a state with too many `eth1_data_votes`. Can I just remove that one?

I see -- yeah I think this is fine to remove; nice catch!

jannikluhn:
I've rebased this on master and removed the tests that don't make sense. The LMD tests are still failing though, all of them in [this line](https://github.com/ethereum/trinity/pull/1283/files#diff-f1ad25dfc555b4dbcbfe21a40cc4ae5fR260-R264) at the end of `test_store_get_latest_attestation`. The reason is that `previous_epoch_attestations` contains `Attestations` instead of `PendingAttestations`, so setting `state.previous_epoch_attestations` doesn't work. The attestation is created [here](https://github.com/ethereum/trinity/pull/1283/files#diff-f1ad25dfc555b4dbcbfe21a40cc4ae5fR74), so I think we just have to replace this with a `PendingAttestation`, but I don't know which parameters to use.

tests/eth2/core/beacon/fork_choice/test_lmd_ghost.py
```diff
@@ -166,8 +166,8 @@ def test_store_get_latest_attestation(
     find the latest ones for each validator?
     """
     some_epoch = 3
-    state = genesis_state.copy(
-        slot=compute_start_slot_at_epoch(some_epoch, config.SLOTS_PER_EPOCH)
+    state = genesis_state.set(
```
ChihChengLiang: Question: When do we `set` and when do we `mset`?

jannikluhn: Use `mset` when updating multiple attributes in a single step (which avoids calculating the intermediate merkle tree, but the efficiency gain should in most cases be marginal). Use `set` if you only want to set a single value. You can also use `mset` with only a single key-value pair which would be identical to `set`, but this feels slightly wrong to me.

PR #1285, opened 2019-11-15
Formalize new service api, move old service API to be "legacy"

pipermerriam:
Extracted from #1265 

### What was wrong?

The new service API that currently lives in `p2p/trio_service.py` is nice.

https://github.com/ethereum/trinity/pull/1259 demonstrates that the same API can be cleanly mapped into the `asyncio` ecosystem.

Currently everything lives under `p2p/trio_service.py` which isn't ideal.

### How was it fixed?

Relocated the existing service stuff to be under `p2p.legacy_service`

Moved the new service model under `p2p.service` and properly namespaced things that were trio specific.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


carver:
It feels backwards to name this the new "blessed" approach before we have anything running it.

I'm :+1: on a direction of getting rid of the cancel token, as long as we still:
- have the ability for services to run cleanups when they are cancelled externally
- keep some of the guarantees about methods not cancelling out in the middle of running

These are the main benefits of the cancel token that I know of. Anyway, I would expect roughly these 4-5 PRs in order:

1. Move the API out of trio into p2p.abc (this seems like a quick PR to give a :+1: to)
2. Build asyncio service using the new API (here we can talk about some of the things that we might lose by losing the cancel token)
3. Convert an interesting service to use the new API, updating/modifying the API along the way. (maybe a couple services)
4. "Bless" the new API, move the old one to legacy (this isolated will probably be a quick PR to :+1: )

This isn't really a process thing, so much as to say that good things come out of tighter reviews of smaller chunks, and I don't really trust the ability to review PR 2 in abstract until we do a couple from PR 3.

PR #1288, opened 2019-11-16
Update to libp2p 0.1.4

NIC619:
### How was it fixed?
Update to libp2p-0.1.2.
- [x] Catch `multiaddr` exceptions when dialing a peer for a more detailed error handling
- [x] dialing a peer could raise two type of errors: `SwarmException` and `HandshakeFailure`, so in `dial_peer` we catch these two exceptions and raise a `DialPeerError` exception from there
- [x] Handle `StreamFailure` when opening a `new_stream`
- There are errors regarding pubsub not handled in py-libp2p currently(version `0.1.2`) but is WIP in https://github.com/libp2p/py-libp2p/pull/351 .

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

(source: https://www.maxpixel.net/Cat-Feline-Kitty-Pets-Cute-Animal-Close-up-Animal-3402206)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ deps = {
         "argcomplete>=1.10.0,<2",
         "multiaddr>=0.0.8,<0.1.0",
         "pymultihash>=0.8.2",
-        "libp2p==0.1.1",
+        "libp2p==0.1.2",
     ],
     'test': [
         "async-timeout>=3.0.1,<4",
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ deps = {
         "argcomplete>=1.10.0,<2",
         "multiaddr>=0.0.8,<0.1.0",
         "pymultihash>=0.8.2",
-        "libp2p==0.1.2",
+        "libp2p==0.1.4",
     ],
     'test': [
         "async-timeout>=3.0.1,<4",
--- a/trinity/protocol/bcc_libp2p/node.py
+++ b/trinity/protocol/bcc_libp2p/node.py
@@ -55,6 +55,7 @@ from libp2p.security.secio.transport import Transport as SecIOTransport
 from libp2p.network.stream.net_stream_interface import (
     INetStream,
 )
+from libp2p.network.exceptions import SwarmException
 from libp2p.peer.id import (
     ID,
 )
@@ -78,6 +79,10 @@ from multiaddr import (
     Multiaddr,
     protocols,
 )
+from multiaddr.exceptions import (
+    BinaryParseError,
+    ProtocolLookupError,
+)
 
 import ssz
 
@@ -98,6 +103,7 @@ from .configs import (
     ResponseCode,
 )
 from .exceptions import (
+    DialPeerError,
     HandshakeFailure,
     ReadMessageFailure,
     RequestFailure,
@@ -352,15 +358,16 @@ class Node(BaseService):
                     addrs=[maddr],
                 )
             )
-        except Exception as e:
-            raise ConnectionRefusedError() from e
+        except SwarmException as e:
+            self.logger.debug("Fail to dial peer_id %s maddr %s, error: %s", peer_id, maddr, e)
+            raise DialPeerError from e
 
         try:
             # TODO: set a time limit on completing handshake
             await self.request_status(peer_id)
         except HandshakeFailure as e:
-            self.logger.info("HandshakeFailure: %s", str(e))
-            raise ConnectionRefusedError() from e
+            self.logger.debug("Fail to dial peer_id %s maddr %s, error: %s", peer_id, maddr, e)
+            raise DialPeerError from e
 
     async def dial_peer_with_retries(self, ip: str, port: int, peer_id: ID) -> None:
         """
@@ -372,7 +379,7 @@ class Node(BaseService):
                 await asyncio.sleep(2**i + random.random())
                 await self.dial_peer(ip, port, peer_id)
                 return
-            except ConnectionRefusedError:
+            except DialPeerError:
                 self.logger.debug(
                     "Could not connect to peer %s at %s:%d;"
                     " retrying attempt %d of %d...",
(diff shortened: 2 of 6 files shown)


trinity/protocol/bcc_libp2p/node.py
```diff
@@ -381,7 +393,7 @@ def _setup_topic_validators(self) -> None:
                     DIAL_RETRY_COUNT,
                 )
                 continue
-        raise ConnectionRefusedError
+        raise DialPeerError
 
     async def dial_peer_maddr(self, maddr: Multiaddr) -> None:
```
ChihChengLiang: Is it possible to clean up `dial_peer_maddr`, `dial_peer_with_retries`, and `dial_peer`?
We pass Multiaddress to `dial_peer_maddr`, then parse `ip`, `port`, and `p2p_id` and handle errors in `dial_peer_with_retries`,  and then again we turn `ip`, `port` into Multiaddress in `dial_peer`.

NIC619: @ChihChengLiang @mhchia 
I move the basic peer dialing functionalities (`host.connect()` and `request_status()`) into `_connect_and_handshake` and rename `dial_peer_with_retries` to `dial_peer_maddr_with_retries`.
`dial_peer` and `dial_peer_maddr` serve as the proxy to `_connect_and_handshake`

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -337,74 +347,80 @@ def _setup_topic_validators(self) -> None:
...
+        self.logger.debug("Dialing peer_id: %s, maddr: %s", peer_id, maddr)
+        await self._connect_and_handshake(peer_id, maddr)
+        self.logger.debug("Successfully connect to peer_id %s maddr %s", peer_id, maddr)
+
+    async def dial_peer_maddr(self, maddr: Multiaddr, peer_id: ID) -> None:
+        """
+        Dial the peer with given multi-address
+        """
+        self.logger.debug("Dialing peer_id: %s, maddr: %s", peer_id, maddr)
+        await self._connect_and_handshake(peer_id, maddr)
```
ChihChengLiang: So `dial_peer_maddr` is `_connect_and_handshake`? Should we just keep one?

trinity/protocol/bcc_libp2p/node.py
```diff
@@ -337,74 +347,80 @@ def _setup_topic_validators(self) -> None:
...
             # TODO: set a time limit on completing handshake
             await self.request_status(peer_id)
         except HandshakeFailure as e:
-            self.logger.info("HandshakeFailure: %s", str(e))
-            raise ConnectionRefusedError() from e
+            self.logger.debug("Fail to handshake with peer_id: %s, error: %s", peer_id, e)
+            raise DialPeerError from e
 
-    async def dial_peer_with_retries(self, ip: str, port: int, peer_id: ID) -> None:
+    async def dial_peer(self, ip: str, port: int, peer_id: ID) -> None:
```
ChihChengLiang: `dial_peer` is only used in the test `test_node_dial_peer`, and can be replaced with `dial_peer_maddr`

NIC619:
@ChihChengLiang I fix the `stream.close` condition in `Interaction.__aexit__` in 6bac555.

NIC619:
I'm going to wait for https://github.com/libp2p/py-libp2p/pull/355 and update to version `0.1.3` at once.

NIC619:
@carver After updating `libp2p` from version `0.1.2` to `0.1.3`, all `py36` tests fail because it can't find version `0.1.3`. Was wondering if you have any clue about what could be the cause.

NIC619:
Starting version `0.1.3`, py3.6 won't be supported so the py36 tests are failing.
Copied from [gitter](https://gitter.im/py-libp2p/Lobby?at=5ddf17fe05eec2433d937123):
> Note that if you are on py3.6, which py-libp2p doesn't officially support, then you will not be able to download this new version.

I will revert the libp2p version used back to `0.1.2` for now.

@pipermerriam @carver @ralexstokes @mhchia Should we support py3.6 in py-libp2p?

ralexstokes:
@NIC619 i'd think we want to look at supporting py36 in `py-libp2p`. if the changes are minimal then its probably the better route forward. given that the main user of `py-libp2p` at the moment afaik is `trinity` we want to maximize compatibility... we could theoretically have the eth2 bits of trinity only support 3.7+ but this is harder i suspect given how coupled the eth2 bits are to the rest of the codebase. i'll take a look at how tricky a `py-libp2p` "downgrade" is...

carver:
> i'd think we want to look at supporting py36 in `py-libp2p`.

Yeah, I think supporting py3.6 is the best option. We've been using Ubuntu LTS python3 distributions for a hint at how broadly available python versions are. Last I checked, the py3 package on the LTS release is 3.6.

:+1: for adding py3.6 support to py-libp2p

(and yay for catching this now via hard version constraints rather than later in some weird soft error about a missing API)

ralexstokes:
ok port to py36 is here: https://github.com/libp2p/py-libp2p/pull/372

 -- didn't take too long although i did find some stuff i'm not happy about (i.e. tests failing w/o the arbitrary sleeps) but i figure we hold off given a trio refactor...

NIC619:
@carver Can you cut a release for pylibp2p since we have merged https://github.com/libp2p/py-libp2p/pull/372 ? Thanks!

carver:
> @carver Can you cut a release for pylibp2p since we have merged [libp2p/py-libp2p#372](https://github.com/libp2p/py-libp2p/pull/372) ? Thanks!

Done!

veox:
PR title needs updating, too.

setup.py
```diff
@@ -49,7 +49,7 @@
         "argcomplete>=1.10.0,<2",
         "multiaddr>=0.0.8,<0.1.0",
         "pymultihash>=0.8.2",
-        "libp2p==0.1.1",
+        "libp2p==0.1.4",
```
carver: Maybe the next time we touch libp2p, or if we run into hard-pinning trouble, this can be relaxed to `libp2p>=0.1.4,<0.2.0` (semver for v0 releases)

NIC619: Sure! Last time there was a bug in 0.1.2 so we had to pin it to 0.1.1 for trinity beacon to work for the time being.

PR #1291, opened 2019-11-19
Correctly handle disconnections from maxed out peers

### LINKED ISSUE #1167: Base p2p handshake code doesn't like Disconnect as first message across the connection.
### What is wrong?

I'm seeing this line show up in logs a lot:

```
   DEBUG  09-18 13:18:03           ETHPeerPool  Could not complete handshake with <Node(0x0d51ceff21eb6eefd3e11417ea22358f16acec6dbfd97263339a7d4cabbd65bbb664aa2d7782ca5f79f734313e718ddd10c96a91d1b52022c453b3a104230e1c@95.216.227.170:30303)>: HandshakeFailure('First message across the DevP2P connection must be a Hello msg, got Disconnect (cmd_id=1), disconnecting',)
```

I *suspect* these are common because when we try to dial a peer that has a full peer pool, they *may* just immediately send the disconnect message rather than handshaking.

We should handle this gracefully as it happens a lot and it seems reasonable.

### How can it be fixed

https://github.com/ethereum/trinity/blob/83a2963050e7f7f30d846591d35937ebe6ac4f05/p2p/handshake.py#L126-L130

THis code should be updated to special case the `Disconnect` message.  The handshake still needs to fail so we could just raise a `HandshakeFailure` exception with a more appropriate message.

gsalgado:
A maxed out peer will disconnect during the P2P handshake by
sending us a disconnect msg with the TOO_MANY_PEERS reason.  We must
detect that and wait a while before attempting to connect again.

Closes #1167

### What was wrong?

We were not detecting the correct disconnection reason from maxed out
peers and because of that were attempting to connect again too
frequently.

### How was it fixed?

Raise the appropriate exception when a maxed out peer disconnect from us

### DIFF
--- a/p2p/abc.py
+++ b/p2p/abc.py
@@ -201,6 +201,7 @@ class TransportAPI(ABC):
     session: SessionAPI
     remote: NodeAPI
     read_state: TransportState
+    logger: ExtendedDebugLogger
 
     @property
     @abstractmethod
--- a/p2p/commands.py
+++ b/p2p/commands.py
@@ -94,6 +94,9 @@ class BaseCommand(CommandAPI[TCommandPayload]):
     def __init__(self, payload: TCommandPayload) -> None:
         self.payload = payload
 
+    def __repr__(self) -> str:
+        return f"{self.__class__}(payload={self.payload})"
+
     def encode(self, cmd_id: int, snappy_support: bool) -> MessageAPI:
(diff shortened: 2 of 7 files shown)


gsalgado:
The spec is actually ambiguous as it says 

> At any time, a Disconnect message may be sent.

I've filed https://github.com/ethereum/devp2p/issues/128 for that

PR #1289, opened 2019-11-18
Update trio to 0.13

pipermerriam:
### What was wrong?

Trio warnings galore!

### How was it fixed?

Updated to latest trio version.  This included an upstream bump to a new `lahja` version that supports the new `trio` version.

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [x] Clean up commit history

#### Cute Animal Picture

seal-xlarge_trans++RtvZRSV037_kYj9aGppl8KoSQRPSZoKNSSXquGEEDe0

PR #1294, opened 2019-11-20
fix #1161, Trinity logs should automatically rotate across runs

### LINKED ISSUE #1161: Trinity logs should automatically rotate across runs

KimiWu123:
### What was wrong?
#1161, Trinity logs should automatically rotate across runs.

PR #1228 is a mass so I re-PR again.

### How was it fixed?
Log rotation per instantiation of the client with format, YYYYMMDD_HHMMSS. File name would look like `trinity.20191024_095854.log`.

### DIFF
--- a/trinity/_utils/logging.py
+++ b/trinity/_utils/logging.py
@@ -1,4 +1,5 @@
 import copy
+from datetime import datetime
 import logging
 from logging import (
     StreamHandler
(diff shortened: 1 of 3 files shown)


cburgdorf:
@KimiWu123 sorry, the CI fail is unrelated to your code but I just merged https://github.com/ethereum/trinity/pull/1289 which should fix it if you rebase again.

trinity/_utils/logging.py
```diff
@@ -28,6 +28,8 @@
     bold_yellow,
 )
 
+from datetime import datetime
```
cburgdorf: This import should be on the first line right above `functools` 
More info: https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#overall-ordering

cburgdorf:
@KimiWu123 Thank you very much for your work on this. I can take care of the rebase and merge it!

pipermerriam:
Currently trinity keeps up to 10 logfiles at 1mb each which gives a total of up to 50mb of logs.

Now that the filename is part of the name, old logs won't get automatically cleaned up.

My preference is to take the timestamp out of the filename.

PR #1295, opened 2019-11-20
Enable py37-p2p-trio CI job and fix mypy error

hwwhww:
### What was wrong?
1. `py37-p2p-trio` job wasn't enabled in CI.
2. mypy check failed in CI

### How was it fixed?
1. Enable `py37-p2p-trio` job in CI
2. Due to https://github.com/ethereum/eth-utils/commit/ff14dae8eac5f17c38ec50606deff14bf1043d31 changes, the mypy check failed when installing `eth-utils>=1.8.1`. I set `eth-utils>=1.8.1,<2` and introduce `HexStr` to Trinity.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture
grey-seal-2164736_640

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -389,7 +389,7 @@ workflows:
 
       - py37-eth1-core
       - py37-p2p
-      - py36-p2p-trio
+      - py37-p2p-trio
       - py37-eth2-core
       - py37-eth2-utils
       - py37-eth2-fixtures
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -389,7 +389,7 @@ workflows:
 
       - py37-eth1-core
       - py37-p2p
-      - py36-p2p-trio
+      - py37-p2p-trio
```
mhchia: Nice catch!

PR #1298, opened 2019-11-20
Migrate to async-service library

pipermerriam:
### What was wrong?

I've worked with it enough that I'm confident enough in the new service abstraction to go ahead and pull it out into a stand alone library.

### How was it fixed?

Created the `async-service` library and migrated the existing code that uses the trio one to use the new library.

#### Cute Animal Picture

winter-seal-pup


pipermerriam:
Weird stuff going on with mypy....

pipermerriam:
A little confused why/how this is still installing `mypy==0.740` which is causing the strange assertion error.

pipermerriam:
I found a way around the odd mypy errors which was to run the individual modules separately...

p2p/discv5/packer.py
```diff
@@ -583,12 +583,15 @@ def register_peer_packer(self, remote_node_id: NodeID) -> None:
             enr_db=self.enr_db,
             message_type_registry=self.message_type_registry,
             incoming_packet_receive_channel=incoming_packet_channels[1],
-            incoming_message_send_channel=self.incoming_message_send_channel.clone(),
+            # trio doesn't expose a `MemoryXXXChannel` type that can be used
+            # for these and the `clone` method is only available on the memory
+            # channel interfaces.
```
carver: I'm having trouble parsing this. Is it the `.clone()` that mypy thinks you can't do. Or is the result a type that mypy thinks is invalid for the `incoming_message_send_channel` keyword arg?

setup.py
```diff
@@ -11,6 +11,7 @@
     'p2p': [
         "async-exit-stack==1.0.1",
         "async-generator==1.10",
+        "async-service==0.1.0a2",
```
carver: I added #1324 to track that this should reach beta stability before trinity is released as beta.

tests-trio/eth1-monitor/conftest.py
```diff
@@ -12,14 +12,14 @@
...
+from async_service import background_trio_service
 from web3 import Web3
 from web3.providers.eth_tester import EthereumTesterProvider
 from lahja.trio.endpoint import TrioEndpoint
 from lahja import ConnectionConfig
 
-from p2p.trio_service import background_service
-from trinity.components.eth2.eth1_monitor.configs import deposit_contract_json
 from trinity.components.eth2.eth1_monitor.eth1_monitor import Eth1Monitor
+from trinity.components.eth2.eth1_monitor.configs import deposit_contract_json
```
carver: *tiny nit* I think this line was in the correct place before (alphabetically)

PR #1316, opened 2019-11-21
Some typing cleanup

pipermerriam:
builds on #1310 

### What was wrong?

There is a newer `mypy` version (but it seems to break...)

### How was it fixed?

All of these changes are from running the codebase against `mypy==0.740`, however, I have started getting some very odd CI errors with mypy so this is just the changes without the library update.

#### Cute Animal Picture

cuteAnimal16

### DIFF
--- a/p2p/discv5/enr_db.py
+++ b/p2p/discv5/enr_db.py
@@ -35,7 +35,7 @@ class BaseEnrDb(EnrDbApi):
         """
         if enr.identity_scheme.id not in self.identity_scheme_registry:
             raise ValueError(
-                f"ENRs identity scheme with id {enr.identity_scheme.id} unknown to ENR DBs "
+                f"ENRs identity scheme with id {enr.identity_scheme.id!r} unknown to ENR DBs "
                 f"identity scheme registry"
             )
 
--- a/p2p/transport.py
+++ b/p2p/transport.py
@@ -381,7 +381,7 @@ class Transport(TransportAPI):
         expected_header_mac = self._ingress_mac.digest()[:HEADER_LEN]
         if not hmac.compare_digest(expected_header_mac, header_mac):
             raise DecryptionError(
-                f'Invalid header mac: expected {expected_header_mac}, got {header_mac}'
+                f'Invalid header mac: expected {expected_header_mac.hex()}, got {header_mac.hex()}'
             )
         return self._aes_dec.update(header_ciphertext)
 
@@ -401,7 +401,7 @@ class Transport(TransportAPI):
         expected_frame_mac = self._ingress_mac.digest()[:MAC_LEN]
         if not hmac.compare_digest(expected_frame_mac, frame_mac):
             raise DecryptionError(
-                f'Invalid frame mac: expected {expected_frame_mac}, got {frame_mac}'
+                f'Invalid frame mac: expected {expected_frame_mac.hex()}, got {frame_mac.hex()}'
             )
         return self._aes_dec.update(frame_ciphertext)[:body_size]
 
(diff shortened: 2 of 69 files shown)


p2p/discv5/enr_db.py
```diff
@@ -35,7 +35,7 @@ def validate_identity_scheme(self, enr: ENR) -> None:
         """
         if enr.identity_scheme.id not in self.identity_scheme_registry:
             raise ValueError(
-                f"ENRs identity scheme with id {enr.identity_scheme.id} unknown to ENR DBs "
+                f"ENRs identity scheme with id {enr.identity_scheme.id!r} unknown to ENR DBs "
```
cburgdorf: What's the reason that many places in this PR change expressions from evaluating to `str` to evaluating to `repr`?

p2p/transport.py
```diff
@@ -381,7 +381,7 @@ def _decrypt_header(self, data: bytes) -> bytes:
         expected_header_mac = self._ingress_mac.digest()[:HEADER_LEN]
         if not hmac.compare_digest(expected_header_mac, header_mac):
             raise DecryptionError(
-                f'Invalid header mac: expected {expected_header_mac}, got {header_mac}'
+                f'Invalid header mac: expected {expected_header_mac.hex()}, got {header_mac.hex()}'
```
cburgdorf: What's the reason for the`.hex()` changes?

tests/components/eth2/beacon/test_slot_ticker.py
```diff
@@ -26,7 +26,7 @@
             loop=event_loop,
         )
     except asyncio.TimeoutError:
-        assert False, "Slot not ticking"
+        raise AssertionError("Slot not ticking")
```
cburgdorf: I like that :+1:

trinity/rpc/modules/eth.py
```diff
@@ -121,13 +121,18 @@ def dict_to_spoof_transaction(
     return cast(SignedTransactionAPI, SpoofTransaction(unsigned, from_=sender))
 
 
+class SyncProgressDict(TypedDict):
```
cburgdorf: Why the rename from `SyncProcess` to `SyncProgressDict`? The `Dict` doesn't add any value to the name imho. Also I'd prefer it to derive from `NamedTuple` instead.

```python
class SyncProgress(NamedTuple):
    startingBlock: BlockNumber
    currentBlock: BlockNumber
    highestBlock: BlockNumber	
```

p2p/discv5/enr_db.py
pipermerriam: mypy `0.740` now complains anytime you use a bytes value in string formatting without explicitely declaring it to be `!r` since it will be embedded as `"b'somebytes'"` which may be surprising behavior.

p2p/transport.py
pipermerriam: For many of these that mypy `0.740` complains about I chose to use their hex representation over `!r` for improved readability.

tests/components/eth2/beacon/test_slot_ticker.py
pipermerriam: mypy `0.740` doesn't like `assert False`

trinity/rpc/modules/eth.py
pipermerriam: We already have a `SyncProgress` object that is a named tuple.  This one is specifically for RPC responses that are dictionaries.

p2p/discv5/enr_db.py
cburgdorf: Ah, that makes sense!

tests/components/eth2/beacon/test_slot_ticker.py
cburgdorf: Me neither 😅

PR #1325, opened 2019-11-26
Convert DiscoveryService to use the trio-based Service

gsalgado:

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -5,16 +5,13 @@ listening nodes.
 
 More information at https://github.com/ethereum/devp2p/blob/master/rlpx.md#node-discovery
 """
-import asyncio
 import collections
 import contextlib
 import random
-import socket
 import time
 from typing import (
     Any,
     Callable,
-    cast,
     DefaultDict,
     Dict,
     Hashable,
@@ -23,11 +20,12 @@ from typing import (
     List,
     Sequence,
     Set,
-    Text,
     Tuple,
     TYPE_CHECKING,
-    Union,
 )
+
+import trio
+import trio_typing
 
 import eth_utils.toolz
 from eth_utils import (
@@ -58,7 +56,7 @@ from eth_keys import datatypes
 
 from eth_hash.auto import keccak
 
-from cancel_token import CancelToken, OperationCancelled
+from async_service import Service
 
 from p2p import constants
 from p2p.abc import AddressAPI, NodeAPI
@@ -70,7 +68,7 @@ from p2p.events import (
 )
 from p2p.exceptions import AlreadyWaitingDiscoveryResponse, NoEligibleNodes
 from p2p.kademlia import Address, Node, RoutingTable, check_relayed_addr, sort_by_distance
-from p2p.service import BaseService
+from p2p import trio_utils
 
 if TYPE_CHECKING:
     # Promoted workaround for inheriting from generic stdlib class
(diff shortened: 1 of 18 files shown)


p2p/discovery.py
```diff
@@ -154,7 +158,7 @@ def update_routing_table(self, node: NodeAPI) -> None:
             # with the least recently seen node on that bucket. If the bonding fails the node will
             # be removed from the bucket and a new one will be picked from the bucket's
             # replacement cache.
-            asyncio.ensure_future(self.bond(eviction_candidate))
+            self.nursery.start_soon(self.bond, eviction_candidate)  # type: ignore
```
pipermerriam: I'm curious if this is something new with `trio_typing` to try and ensure that functions that are run using `start_soon` don't expect to return a value since `trio` doesn't expose any way to get the return value for such functions.

gsalgado: I don't think it's new, but I've reported it at https://github.com/python-trio/trio-typing/issues/16

Also, I'm talking to the developers on Gitter and they seem to agree it makes sense to allow a return value as the same function could be used in other contexts (e.g. trio.gather()) where the return value is preserved

p2p/discovery.py
```diff
@@ -388,36 +388,30 @@ def connection_made(self, transport: asyncio.BaseTransport) -> None:
...
+        # start_soon(), so ignore type checking here.
+        # https://github.com/python-trio/trio-typing/issues/16
+        self.nursery.start_soon(
+            self.socket.sendto, message, (node.address.ip, node.address.udp_port))  # type: ignore
+
+    async def consume_datagram(self) -> None:
+        datagram, (ip_address, port) = await self.socket.recvfrom(DATAGRAM_BUFFER_SIZE)
+        address = Address(ip_address, port)
+        self.logger.debug2("Received datagram from %s", address)
+        self.receive(address, datagram)
```
pipermerriam: Might be nice to wrap these up as something like:

```python
class Datagram(NamedTuple)
    payload: bytes
    address: Address
```

gsalgado:
@pipermerriam so, I think this is ready for a final review. there will be some conflicts to solve once you land #1328, but it shouldn't be anything major

p2p/discovery.py
gsalgado: I started doing that but then realized we'd be wrapping them here just to immediately unwrap in `receive()`. And also, in `receive()` we unpack the datagram to separate the payload from the signature and other stuff included in the datagram, so Datagram.payload would be confusing as here it still includes more than the real payload

p2p/discovery.py
```diff
@@ -770,21 +770,24 @@ def __init__(self,
...
-        # TODO: Support IPv6 addresses as well.
-        await loop.create_datagram_endpoint(
-            lambda: self.proto,
-            local_addr=('0.0.0.0', self.port),
-            family=socket.AF_INET)
+    @property
+    def proto(self) -> DiscoveryProtocol:
+        if self._proto is None:
+            # Must silence mypy here because we're accessing a private attribute of our manager.
+            nursery = self.manager._task_nursery  # type: ignore
```
pipermerriam: I'd like to understand what other options we have here as this seems likes something that is likely to cause problems.

Can we re-work this to be set during runtime and setup an independent nursery in the `Service.run` method?

tests-trio/p2p-trio/test_discovery.py
```diff
@@ -112,15 +108,16 @@ def test_find_node_neighbours_v4(alice, bob):
...
+async def test_wait_ping(nursery, echo):
     proto = MockDiscoveryProtocol([])
     node = NodeFactory()
 
     # Schedule a call to proto.recv_ping() simulating a ping from the node we expect.
-    recv_ping_coroutine = asyncio.coroutine(lambda: proto.recv_ping_v4(node, echo, b''))
-    asyncio.ensure_future(recv_ping_coroutine())
+    async def recv_ping() -> None:
+        proto.recv_ping_v4(node, echo, b'')
+    nursery.start_soon(recv_ping)
```
pipermerriam: I believe this could just be `nursery.start_soon(proto.recv_ping, node, echo, b'')`

tests-trio/p2p-trio/test_discovery.py
```diff
@@ -210,7 +208,9 @@ def test_find_node_neighbours_v4(alice, bob):
     proto.send_ping_v4 = lambda remote: token
 
     # Pretend we get a pong from the node we are bonding with.
-    proto.wait_pong_v4 = asyncio.coroutine(lambda n, t: t == token and n == node)
+    async def wait_pong_v4(remote, t) -> bool:
+        return t == token and remote == node
+    proto.wait_pong_v4 = wait_pong_v4
```
pipermerriam: maybe use the `monkeypatch` fixture to do this?  I think it is a *slight* improvement because it will throw an exception if the attribute isn't already on the object.

tests-trio/p2p-trio/test_discovery.py
```diff
@@ -232,19 +232,19 @@ def test_update_routing_table():
...
     bond_called = False
 
-    def bond(node):
+    async def bond(node):
         nonlocal bond_called
         bond_called = True
         assert node == old_node
 
-    proto.bond = asyncio.coroutine(bond)
+    proto.bond = bond
```
pipermerriam: same here for use of `monkeypatch` to set this.

trinity/components/builtin/peer_discovery/component.py
```diff
@@ -130,14 +78,27 @@ def configure_parser(cls,
...
-            self.boot_info.trinity_config
-        )
-        asyncio.ensure_future(exit_with_services(
-            discovery_bootstrap,
-            self._event_bus_service,
-        ))
-        asyncio.ensure_future(discovery_bootstrap.run())
+    async def run(self) -> None:
+        config = self.boot_info.trinity_config
+        external_ip = "0.0.0.0"
```
pipermerriam: Just noting that we should update the UPNP component to broadcast the external IP address and then consume that here to update our advertised address.

p2p/discovery.py
gsalgado: My plan was to try and redesign DiscoveryProtocol so that it has only sync methods and none of them go over the network, but that would be a lot of work and may not even be feasible. Another option, that seems a lot more reasonable, is to combine DiscoveryProtocol and DiscoveryService in a single class.

gsalgado: I could try and do the latter as part of this PR. What do you think?

pipermerriam: Just want to make sure it gets addressed promptly.  You can do that however you'd like.  :+1:

tests-trio/p2p-trio/test_discovery.py
gsalgado: Nope, that is a sync method. We run it asynchronously because we need to be `await`ing on `proto.wait_ping()` (a few lines below) when the ping gets sent

tests-trio/p2p-trio/test_discovery.py
gsalgado: Yeah, I need to remember to use that!

PR #1328, opened 2019-11-26
Update component APIs

pipermerriam:
### What was wrong?

The component APIs were complex.

### How was it fixed?

- The `ComponentManager` and `ComponentManagerService` have been combined into a single unit.
- The component APIs have been simplified
    - What was previously the `BaseMainProcessComponent` is now the `Application` component type only exposing `configure_argparser` and `validate_cli`.
    - The other component type is now just `ComponentAPI` with the `AsyncioIsolatedComponent` implementing that API.  The overall component API is now much simpler and the lifecycle management is better defined.
- Components now natively run asynchronously with `Component.run()`.

The other thing that this did was remove the `trinity._utils.shutdown` which are no longer needed.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

harp_seals_3


pipermerriam:
FYI, this should deliver on at least one major UX improvement which is faster startup/shutdown of the client.

pipermerriam:
I know this PR is larger than our normal size but I think that it makes sense as a single piece (and doesn't break up into smaller pieces well).  Mostly a lot of mundane changes to components to transform them to use the new APIs.

pipermerriam:
@gsalgado I got this branch ready if you want to take another look.  Hoping to merge tomorrow.

docs/guides/writing_components.rst
```diff
@@ -37,8 +37,6 @@ There are currently three different types of components that we'll all cover in
 
 - Components that overtake and redefine the entire ``trinity`` command
 - Components that spawn their own new isolated process
-- Components that run in the shared `networking` process
-
```
cburgdorf: Greetings from the past :ghost:

trinity/components/builtin/json_rpc/component.py
```diff
@@ -74,67 +111,33 @@ def configure_parser(cls, arg_parser: ArgumentParser, subparser: _SubParsersActi
...
         # Run HTTP Server
-        if self.boot_info.args.enable_http:
-            http_server = HTTPServer(rpc, port=self.boot_info.args.rpcport)
-            asyncio.ensure_future(http_server.run())
+        if boot_info.args.enable_http:
+            http_server = HTTPServer(rpc, port=boot_info.args.rpcport)
             services_to_exit += (http_server,)
 
-        asyncio.ensure_future(exit_with_services(*services_to_exit))
+        async with AsyncExitStack() as stack:
```
cburgdorf: Interesting, I'd never seen this before, but I just read up on it.

trinity/extensibility/component.py
```diff
@@ -180,47 +92,32 @@ class BaseIsolatedComponent(BaseComponent):
...
-        self._process = ctx.Process(
-            target=self._prepare_spawn,
-        )
-
-        self._process.start()
-        self.logger.info("Component started: %s (pid=%d)", self.name, self._process.pid)
-
-    def _prepare_spawn(self) -> None:
-        if self.boot_info.profile:
-            with profiler(f'profile_{self.normalized_name}'):
```
cburgdorf: The ability to do profiling for every Trinity process is a [documented feature](https://trinity-client.readthedocs.io/en/latest/cookbook.html#performance-profiling) (It seems it is missing a test though!). It looks like this feature is currently lost in this PR.

trinity/components/builtin/network_db/component.py
```diff
@@ -121,49 +117,43 @@ def configure_parser(cls,
...
-                self.start()
+    @classmethod
+    def validate_cli(cls, boot_info: BootInfo) -> None:
+        try:
+            get_tracking_database(get_networkdb_path(boot_info.trinity_config))
+        except BadDatabaseError as err:
+            raise ValidationError(
+                "Error loading network database.  Trying removing database "
+                f"with `remove-network-db` command:\n{err}"
+            ) from err
```
cburgdorf: I really really :heart: how that mechanism got transformed into just raising a `ValidationError` :sparkles:

trinity/components/builtin/network_db/component.py
```diff
@@ -121,49 +117,43 @@ def configure_parser(cls,
...
         if db_path.exists():
-            logger.info("Removing network database at: %s", db_path.resolve())
+            cls.logger.info("Removing network database at: %s", db_path.resolve())
             db_path.unlink()
         else:
-            logger.info("No network database found at: %s", db_path.resolve())
+            cls.logger.info("No network database found at: %s", db_path.resolve())
 
-    _session: Session = None
+    _session: ClassVar[Session] = None
```
cburgdorf: Nice tightening of constraints using types :+1:

trinity/components/builtin/peer_discovery/component.py
```diff
@@ -130,14 +123,13 @@ def configure_parser(cls,
...
+            boot_info.trinity_config
         )
-        asyncio.ensure_future(exit_with_services(
-            discovery_bootstrap,
-            self._event_bus_service,
-        ))
-        asyncio.ensure_future(discovery_bootstrap.run())
+
+        async with run_service(discovery_bootstrap):
+            await discovery_bootstrap.cancellation()
```
cburgdorf: This is another place that could become:

```python
await run_service_until_canceled(discovery_bootstrap)
```

trinity/components/builtin/request_server/component.py
```diff
@@ -48,37 +45,40 @@ def configure_parser(cls, arg_parser: ArgumentParser, subparser: _SubParsersActi
...
                 trinity_config.get_app_config(Eth1AppConfig),
                 base_db,
+                event_bus,
             )
         else:
             raise Exception("Trinity config must have eth1 config")
 
-        asyncio.ensure_future(exit_with_services(server, self._event_bus_service))
-        asyncio.ensure_future(server.run())
+        async with run_service(server):
```
cburgdorf: This is another place that could become:

```python
await run_service_until_canceled(discovery_bootstrap)
```

trinity/components/builtin/tx_pool/component.py
```diff
@@ -51,49 +47,39 @@ def configure_parser(cls, arg_parser: ArgumentParser, subparser: _SubParsersActi
...
-        if self.boot_info.trinity_config.network_id == MAINNET_NETWORK_ID:
+        if boot_info.trinity_config.network_id == MAINNET_NETWORK_ID:
             validator = DefaultTransactionValidator(chain, PETERSBURG_MAINNET_BLOCK)
-        elif self.boot_info.trinity_config.network_id == ROPSTEN_NETWORK_ID:
+        elif boot_info.trinity_config.network_id == ROPSTEN_NETWORK_ID:
             validator = DefaultTransactionValidator(chain, PETERSBURG_ROPSTEN_BLOCK)
         else:
-            raise ValueError("The TxPool component only supports MainnetChain or RopstenChain")
+            # TODO: this should cause the node to shut down maybe?
+            raise ValidationError("The TxPool component only supports MainnetChain or RopstenChain")
```
cburgdorf: This looks like a good candidate to move into `validate_cli` and shut down the node

trinity/components/builtin/tx_pool/component.py
```diff
@@ -51,49 +47,39 @@ def configure_parser(cls, arg_parser: ArgumentParser, subparser: _SubParsersActi
...
 
-    async def do_stop(self) -> None:
-        # This isn't really needed for the standard shutdown case as the TxPool will automatically
-        # shutdown whenever the `CancelToken` it was chained with is triggered. It may still be
-        # useful to stop the TxPool component individually though.
-        if self.tx_pool.is_operational:
-            await self.tx_pool.cancel()
-            self.logger.info("Successfully stopped TxPool")
+        async with run_service(tx_pool):
+            await tx_pool.cancellation()
```
cburgdorf: And another place that could become

```python
await run_service_until_canceled(tx_pool)
```

trinity/components/builtin/upnp/component.py
```diff
@@ -43,11 +36,10 @@ def configure_parser(cls,
...
-            self.upnp_service,
-            self._event_bus_service,
-        ))
-        asyncio.ensure_future(self.upnp_service.run())
+    @classmethod
+    async def do_run(cls, boot_info: BootInfo, event_bus: EndpointAPI) -> None:
+        port = boot_info.trinity_config.port
+        upnp_service = UPnPService(port)
+
+        async with run_service(upnp_service):
```
cburgdorf: And another place that could become

```python
await run_service_until_canceled(tx_pool)
```

trinity/extensibility/asyncio.py
```diff
@@ -1,45 +1,89 @@
...
+            except asyncio.CancelledError:
+                await event_bus_service.cancel()
+                raise
 
-        self.do_start()
+    @classmethod
+    @abstractmethod
+    async def do_run(self, boot_info: BootInfo, event_bus: EndpointAPI) -> None:
+        """
+        This is where subclasses should override
```
cburgdorf: How about:

>Define the entry point of the component. Should be overwritten in subclasses.

trinity/extensibility/component.py
```diff
@@ -6,168 +6,80 @@
...
 
-    @property
-    def status(self) -> ComponentStatus:
-        """
-        Return the current :class:`~trinity.extensibility.component.ComponentStatus`
-        of the component.
-        """
-        return self._status
+class ComponentAPI(BaseComponentAPI):
+    name: str
```
cburgdorf: It looks like this is making it easy for subclasses to forget to define a name if I'm not mistaking..

trinity/extensibility/component_manager.py
```diff
@@ -1,89 +1,121 @@
...
+                self._boot_info.trinity_config.ipc_dir
+            )
+        return self._main_endpoint_config
+
+    async def get_event_bus(self) -> EndpointAPI:
+        await self._endpoint_available.wait()
         return self._endpoint
 
-    def prepare(self, boot_info: BootInfo) -> None:
+    async def _run(self) -> None:
```
cburgdorf: This method looks a lot like vodoo but I like it :)

trinity/components/builtin/network_db/component.py
```diff
@@ -209,40 +203,43 @@ def _get_eth1_tracker(self) -> BaseEth1PeerTracker:
...
-                self._event_bus_service,
-                *tracker_services,
-            ))
+            cls.logger.exception(f"Unrecoverable error in Network Component: {err}")
+
+        async with AsyncExitStack() as stack:
             for service in tracker_services:
-                asyncio.ensure_future(service.run())
+                await stack.enter_async_context(run_service(service))
+            await tracker_services[0].cancellation()
```
gsalgado: ISTM that here we're assuming that all services are going to be cancelled together, which is the case when we get a SIGINT/SIGTERM, but is that the only case a running service might be cancelled? Or should we wait for the cancellation of *any* service here?

trinity/components/builtin/syncer/component.py
```diff
@@ -283,50 +276,55 @@ def extract_strategy_types(cls) -> Iterable[Type[BaseSyncStrategy]]:
...
-            self.event_bus,
+            event_bus,
             node.cancel_token
         )
 
-        if self.active_strategy.shutdown_node_on_halt:
-            self.logger.error("Sync ended unexpectedly. Shutting down trinity")
-            await self.event_bus.broadcast(ShutdownRequest("Sync ended unexpectedly"))
+        if strategy:
+            cls.logger.error("Sync ended unexpectedly. Shutting down trinity")
```
gsalgado: Hmmm, this looks weird. What does `if strategy` mean?

trinity/components/eth2/beacon/component.py
```diff
@@ -163,16 +160,10 @@ def do_start(self) -> None:
...
-        asyncio.ensure_future(slot_ticker.run())
-        asyncio.ensure_future(validator.run())
-        asyncio.ensure_future(syncer.run())
+        services = (libp2p_node, receive_server, slot_ticker, validator, syncer)
+
+        async with AsyncExitStack() as stack:
+            for service in services:
+                await stack.enter_async_context(run_service(service))
+
+            await libp2p_node.cancellation()
```
gsalgado: Same here about waiting for just one of the services

trinity/extensibility/component.py
```diff
@@ -180,47 +92,32 @@ class BaseIsolatedComponent(BaseComponent):
...
-        :class:`~trinity.extensibility.component_manager.ComponentManager` to tear down the process.
-        This allows isolated components to be taken down concurrently without depending on a running
-        event loop.
-        """
-        self._status = ComponentStatus.STOPPED
 
-    def _setup_logging(self) -> None:
-        setup_child_process_logging(self.boot_info)
+@asynccontextmanager
+async def run_component(component: ComponentAPI) -> AsyncIterator[None]:
```
gsalgado: Maybe `run_asyncio_component()`?

trinity/components/builtin/syncer/component.py
cburgdorf: Good catch, I guess that is supposed to be `strategy.shutdown_node_on_halt`

trinity/extensibility/component.py
gsalgado: Actually, no, that doesn't make sense. This method uses asyncio itself but the component can be anything that provides an async run() method

trinity/extensibility/event_bus.py
```diff
@@ -90,6 +22,7 @@ class AsyncioEventBusService(BaseService):
     def __init__(self,
                  trinity_config: TrinityConfig,
                  endpoint_name: str,
+                 main_endpoint_config: ConnectionConfig = None,
```
gsalgado: I'm curious about the reason for adding this to the constructor as it doesn't seem to be used by anyone other than the ComponentManager, which passes the MAIN_EVENTBUS_ENDPOINT anyway

docs/guides/writing_components.rst
```diff
@@ -188,108 +135,6 @@ its existence later.
...
-   :pyobject: PeerCountReporterComponent.on_ready
-
-In case of a :class:`~trinity.extensibility.asyncio.AsyncioIsolatedComponent`, this will cause the
-:meth:`~trinity.extensibility.asyncio.AsyncioIsolatedComponent.do_start` method to run on an entirely
-separated, new process. In other cases
-:meth:`~trinity.extensibility.asyncio.AsyncioIsolatedComponent.do_start` will simply run in the same
-process as the component manager that the component is controlled by.
-
-
-Communication pattern
```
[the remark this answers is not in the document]
pipermerriam: Probably.  I think I nuked it because the Component API no longer natively provides an event bus to the component so it would be up to the component to set one up.  Probably still deserves to be documented..

trinity/bootstrap.py
```diff
@@ -204,42 +214,43 @@ def main_entry(trinity_boot: BootFn,
...
-            asyncio.ensure_future(component_manager_service.run())
-            loop.add_signal_handler(signal.SIGTERM, lambda: kill_trinity_with_reason("SIGTERM"))
-            loop.run_forever()
-            loop.close()
+            task = asyncio.ensure_future(component_manager_service.run())
+            loop.add_signal_handler(
+                signal.SIGTERM,
+                lambda: loop.run_until_complete(kill_trinity_with_reason("SIGTERM")),
+            )
+            loop.run_until_complete(task)
```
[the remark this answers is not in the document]
pipermerriam: Took me a while to stumble on this pattern but it seems to work quite well as you can register the `task.cancel` in the signal handler for simple cancelations.

trinity/components/builtin/ethstats/component.py
```diff
@@ -81,52 +80,53 @@ def configure_parser(cls, arg_parser: ArgumentParser, subparser: _SubParsersActi
...
+        if not (args.ethstats_server_url or get_default_server_url(network_id)):
+            raise ValidationError(
                 'You must provide ethstats server url using the `--ethstats-server-url`'
             )
-            manager_eventbus.broadcast_nowait(ShutdownRequest("Missing EthStats Server URL"))
-            return
 
         if not args.ethstats_server_secret:
-            self.logger.error(
+            raise ValidationError(
```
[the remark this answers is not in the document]
pipermerriam: Yeah, it took a few iterations to get here but I feel pretty good about the new lifecycle management with `validate_cli` providing the proper hook where previously we had to operate over the bus.

docs/guides/writing_components.rst
```diff
@@ -188,108 +135,6 @@ its existence later.
   component.
 
 
-Defining a components starting point
-------------------------------------
```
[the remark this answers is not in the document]
pipermerriam: :+1: I didn't put much thought into the docs updates, nor did I really do much reading on what the final state was.  I'll give it another pass.

trinity/components/builtin/ethstats/component.py
```diff
@@ -81,52 +80,53 @@ def configure_parser(cls, arg_parser: ArgumentParser, subparser: _SubParsersActi
...
 
-        self.node_id = args.ethstats_node_id
-        self.node_contact = args.ethstats_node_contact
-        self.stats_interval = args.ethstats_interval
+        server_secret = args.ethstats_server_secret
 
-        self.start()
+        node_id = args.ethstats_node_id
+        node_contact = args.ethstats_node_contact
+        stats_interval = args.ethstats_interval
```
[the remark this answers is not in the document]
pipermerriam: I'll update to just pull them directly off of `args` rather than creating more local variables.

trinity/components/builtin/ethstats/component.py
```diff
@@ -81,52 +80,53 @@ def configure_parser(cls, arg_parser: ArgumentParser, subparser: _SubParsersActi
...
+            node_contact,
+            stats_interval,
         )
 
-        asyncio.ensure_future(exit_with_services(
-            service,
-            self._event_bus_service,
-        ))
-        asyncio.ensure_future(service.run())
+        async with run_service(service):
```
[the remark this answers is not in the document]
pipermerriam: The new service API has this so :+1: but I'm not going to go through modifying the old one since we'll be replacing it very soon.

trinity/extensibility/component.py
pipermerriam: I'll get it re-added.  didn't intend to drop it but I think I nuked it at some point to simplify while debugging.

trinity/extensibility/component.py
pipermerriam: I'll add a runtime check.  I like not having to define these as `@property` methods.  I don't feel strongly about this but I think it's a nice convenience.  An alternative would be to allow this to not be set and to derive it from the class name if it isn't explicitely set.

trinity/extensibility/component_manager.py
pipermerriam: I'll add more comments.  I think it deserves them.

tests/core/components/test_isolated_component.py
```diff
@@ -0,0 +1,114 @@
...
+    async def _run(self) -> None:
+        self.logger.error('Broadcasting `IsStarted`')
+        await self.event_bus.broadcast(IsStarted())
+        try:
+            self.logger.error('Waiting for cancellation')
+            await self.cancellation()
+        finally:
+            self.logger.error('Got cancellation: broadcasting `GotCancellation`')
+            await self.event_bus.broadcast(GotCancellation())
+            self.logger.error('EXITING')
```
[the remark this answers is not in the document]
pipermerriam: :+1: originally I think this was during some debugging where I wanted to be sure the messages were making it past handlers which didn't have an explicit level set but now it seems to be working fine at the `info` level.

trinity/components/builtin/network_db/component.py
pipermerriam: I'll update to use `asyncio.wait(..., return_when=asyncio.FIRST_COMPLETED)`

trinity/extensibility/event_bus.py
pipermerriam: I think this is a leftover from some testing code but ultimately I didn't end up needing it.  I'll back this change out.

pipermerriam:
omg finally.  Ok, this **should** be a nice step forward.

cburgdorf:
Nice! Just noticed this doesn't have a `newsfragment` but I think it deserves one. It might actually deserve two, a `bugfix` for the cleaner shutdowns and a `feature` to mention the component API overhaul.

pipermerriam:
@cburgdorf I'll either submit one as a standalone or include it in a follow up.

Issue #1011, opened 2019-08-29
Implement eth_getTransactionByHash


lithp:
From #26, [`eth_getTransactionByHash`](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionbyhash) is not implemented.

voith:
This is fairly simple to implement using graphql. I have added a test here: https://github.com/ethereum/trinity/blob/6b6d224ccca810575ad388bfe223ea3334c613c2/tests/core/graphql-rpc/test_grqphql.py#L418-L446
I haven't exposed all fields, but just writing all the attribute names in the query will fetch the needed attributes.
Maybe this is off topic as this needs to be implemented in `JSON-RPC`, but I'm just excited to share my progress : )

pipermerriam:
Some thoughts...

Initial implementation should be dead simple.  If we have the transaction we return it.

The later work is to figure out what we can do when we don't know about the hash yet.  If we want to be a *full* client in 5 minutes we need a solution for getting them on demand.  Ideas for how we might do that?

1. Swarm via the BZZETH protocol.  They're already doing it for block headers.  Why not transactions and receipts too.
2. Expose convenience API via another sub protocol that only trinity speaks (until other clients adopt it).

lithp:
As a note, Jason and I talked about this and had the idea that `GetNodeData` might *just work*, if transactions were stored in the database by hash it's possible that Geth or Parity don't check that they're returning a node before they return it. Don't know about Parity but I just skimmed the Geth code and transactions aren't store by hash, the entire block body is stored as one chunk, so `GetNodeData` won't work.

lithp:
> The later work is to figure out what we can do when we don't know about the hash yet. 

This RPC is less critical than the RPCs which involved state, I think I remember that syncing the chain + bodies only takes a few hours? But if we want to make the UX even better yeah, I don't see a way beyond what you've described here, somehow asking remote peers for these transactions.

As a kludge, each day we could generate a bunch of shards and upload them to s3 for clients to download-on-demand, then clients would be able to answer `getTransactionByHash` queries as soon as they've synced the last day's worth of blocks.

cburgdorf:
This has now landed as part of #1329. It doesn't do any clever beam sync related things to deal with the case were we try to act as a full node when in reality we haven't synced any historical data yet. But I don't think this is what the ticket is about so I'm closing this for now.

PR #1329, opened 2019-11-27
Implement eth_getTransactionByHash endpoint

cburgdorf:
### What was wrong?

Trinity does not yet serve the [eth_getTransactionByHash](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionbyhash) RPC.

### How was it fixed?

1. Add support in the `rpc/modules/eth.py`
2. Test it as part of the JSON-RPC fixture tests (simiar to how we test `eth_getTransactionByBlockHashAndIndex` and `eth_getTransactionByBlockNumberAndIndex` today)

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/rpc/modules/eth.py
+++ b/trinity/rpc/modules/eth.py
@@ -216,6 +216,12 @@ class Eth(Eth1ChainRPCModule):
         stored_val = state.get_storage(address, position)
         return encode_hex(int_to_big_endian(stored_val))
 
+    @format_params(decode_hex)
+    async def getTransactionByHash(self,
+                                   transaction_hash: Hash32) -> Dict[str, str]:
+        transaction = await self.chain.coro_get_canonical_transaction(transaction_hash)
+        return transaction_to_dict(transaction)
+
     @format_params(decode_hex, to_int_if_hex)
     async def getTransactionByBlockHashAndIndex(self,
                                                 block_hash: Hash32,
--- a/newsfragments/1329.feature.rst
+++ b/newsfragments/1329.feature.rst
@@ -1,0 +1,1 @@
+Add support for `eth_getTransactionByHash` JSON-RPC API
--- a/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
+++ b/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
@@ -362,11 +362,24 @@ def validate_rpc_transaction_vs_fixture(transaction, fixture):
 
 
 async def validate_transaction_by_index(rpc, transaction_fixture, at_block, index):
-    if is_by_hash(at_block):
+    block_by_hash = is_by_hash(at_block)
+    if block_by_hash:
         rpc_method = 'eth_getTransactionByBlockHashAndIndex'
     else:
         rpc_method = 'eth_getTransactionByBlockNumberAndIndex'
     result, error = await call_rpc(rpc, rpc_method, [at_block, hex(index)])
+    assert error is None
+    validate_rpc_transaction_vs_fixture(result, transaction_fixture)
+
+    if not block_by_hash:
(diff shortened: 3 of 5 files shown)


cburgdorf:
Interesting..this seems to fail for some of the fixtures. Investigating...

cburgdorf:
Okay, I deciphered that a little bit...a quick brain dump.

So, what happens is that we persist a block:

```python
    @classmethod
    def _persist_block(
            cls,
            db: DatabaseAPI,
            block: BlockAPI,
            genesis_parent_hash: Hash32) -> Tuple[Tuple[Hash32, ...], Tuple[Hash32, ...]]:
        header_chain = (block.header, )
        new_canonical_headers, old_canonical_headers = cls._persist_header_chain(
            db,
            header_chain,
            genesis_parent_hash
        )
```

But then it doesn't end up in `new_canonical_headers` (for reasons currently unknown to me).

So, the next thing I tried was to *always* store the transactions of the persisted block. But then I ran into an error where the transaction that was returned was a different transaction than we expected because we map transaction hashes to blocknumber + index and it seems that the block at that number has changed. Need to investigate that further tomorrow. This is mostly a brain dump for myself.

pipermerriam:
If this proves ugly I'd be fine with lowering it's priority as this is one of the functions of an ethereum node that I'd like to change, not guaranteeing that it can return all historical transactions by hash.

cburgdorf:
@pipermerriam Ok, so I ended up lowering the expectations of the test coverage for this method. Initially I thought there may be a bug in Py-EVM that needed to be addressed but now I'm actually not so sure anymore. There is a way to [address this on the Py-EVM side](https://github.com/ethereum/py-evm/pull/1884) so that we could always retrieve the tx by its hash but I ended up abandoning that train of thought, thinking that Py-EVM is purposefully designed so that one can only *directly* (meaning by it's hash) lookup transactions and receipts if they are part of the canonical chain.

Transactions or Receipts that are on blocks that aren't on the canonical chain can not be *directly* looked up (because internally we map them to blocknumbers, not blockhashes), they would only become *directly* retrievable if the block they belong to would become part of the canonical chain at some point.

cburgdorf:
Just linking to https://github.com/ethereum/trinity/issues/1011

trinity/rpc/modules/eth.py
```diff
@@ -216,6 +216,12 @@ class Eth(Eth1ChainRPCModule):
         stored_val = state.get_storage(address, position)
         return encode_hex(int_to_big_endian(stored_val))
 
+    @format_params(decode_hex)
+    async def getTransactionByHash(self,
+                                   transaction_hash: Hash32) -> Dict[str, str]:
+        transaction = await self.chain.coro_get_canonical_transaction(transaction_hash)
```
pipermerriam: Not for this PR but...

I'm inclined for us to migrate this stuff back to using a regular `Chain` and to use the `loop.run_in_executor(...)` approach.  I'm thinking it will be more maintainable long term.

Thoughts?

cburgdorf: Yes, getting rid of the `AsyncChainMixin` is probably a good idea :+1:

cburgdorf: On the other hand though, there are places where we want to make sure we don't accidentally call synchronous chain methods. Having a formalized `AsyncChainAPI` with only `coro_*` methods does help to prevent calling synchronous methods. With isn't exactly true today because `AsyncChainAPI` derives from `ChainAPI` so all the synchronous methods are still there and could accidentally be called.

I have mixed feelings about this...

pipermerriam:
https://github.com/ethereum/py-evm/issues/1807 seems somewhat related.

PR #1337, opened 2019-11-29
Implement eth_getTransactionReceipt API

cburgdorf:
### What was wrong?

We need to implement `eth_getTransactionReceipt` API according to the [spec](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt)

### How was it fixed?

1. Retrieve the Receipt, Header, Transaction from the chain
2. Calculate all computed fields
3. Serve via the `eth` RPC module

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
+++ b/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
@@ -22,6 +22,7 @@ from eth_utils import (
     is_string,
     to_checksum_address,
     to_tuple,
+    to_int,
 )
 from eth_utils.curried import (
     apply_formatter_if,
@@ -382,7 +383,7 @@ def validate_rpc_transaction_vs_fixture(transaction, fixture):
     assert actual_transaction == expected
 
 
-async def validate_transaction_by_index(rpc, transaction_fixture, at_block, index):
+async def validate_transaction_by_index(rpc, block_fixture, transaction_fixture, at_block, index):
     block_by_hash = is_by_hash(at_block)
     if block_by_hash:
         rpc_method = 'eth_getTransactionByBlockHashAndIndex'
@@ -397,12 +398,25 @@ async def validate_transaction_by_index(rpc, transaction_fixture, at_block, inde
         # that we can refer to by its number. Otherwise, it may be that we try to lookup
         # a transaction by its hash that is not in the canonical chain which isn't supported.
         await validate_transaction_by_hash(rpc, result['hash'], transaction_fixture)
+        await validate_transaction_receipt(rpc, result['hash'], block_fixture, transaction_fixture)
 
 
 async def validate_transaction_by_hash(rpc, tx_hash, transaction_fixture):
     result, error = await call_rpc(rpc, 'eth_getTransactionByHash', [tx_hash])
     assert error is None
     validate_rpc_transaction_vs_fixture(result, transaction_fixture)
+
+
+async def validate_transaction_receipt(rpc, tx_hash, block_fixture, transaction_fixture):
+    result, error = await call_rpc(rpc, 'eth_getTransactionReceipt', [tx_hash])
+    assert error is None
+    header_fixture = fixture_block_in_rpc_format(block_fixture['blockHeader'])
+    tx_index = to_int(hexstr=result['transactionIndex'])
+    tx_fixture = block_fixture['transactions'][tx_index]
+
+    assert result['to'] == add_0x_prefix(tx_fixture['to'])
+    assert result['blockHash'] == header_fixture['hash']
+    assert result['blockNumber'] == header_fixture['number']
 
 
 async def validate_block(rpc, block_fixture, at_block):
@@ -418,7 +432,8 @@ async def validate_block(rpc, block_fixture, at_block):
     assert len(result['transactions']) == len(block_fixture['transactions'])
 
     for index, transaction_fixture in enumerate(block_fixture['transactions']):
-        await validate_transaction_by_index(rpc, transaction_fixture, at_block, index)
+        await validate_transaction_by_index(
+            rpc, block_fixture, transaction_fixture, at_block, index)
 
     await validate_transaction_count(rpc, block_fixture, at_block)
 
--- a/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
+++ b/tests/json-fixtures-over-rpc/test_rpc_fixtures.py
@@ -1,6 +1,5 @@
 import asyncio
 import json
-import os
 from pathlib import Path
 
 import pytest
@@ -60,7 +59,7 @@ from trinity.rpc.modules import (
 ROOT_PROJECT_DIR = Path(__file__).parent.parent.parent
 
 
-BASE_FIXTURE_PATH = os.path.join(ROOT_PROJECT_DIR, 'fixtures', 'BlockchainTests')
+BASE_FIXTURE_PATH = ROOT_PROJECT_DIR / 'fixtures' / 'BlockchainTests'
 
 SLOW_TESTS = (
     'bcExploitTest/SuicideIssue.json',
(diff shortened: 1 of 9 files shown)


cburgdorf:
Just noticed the hacker PR number :sweat_smile:

trinity/rpc/format.py
```diff
@@ -35,6 +37,47 @@
...
+
+def to_receipt_response(receipt: ReceiptAPI,
+                        transaction: SignedTransactionAPI,
+                        index: int,
+                        header: BlockHeaderAPI) -> Dict[str, Any]:
+    return {
+        'blockHash': encode_hex(header.hash),
+        'blockNumber': hex(header.block_number),
+        # TODO: Add support in Py-EVM
+        # 'contractAddress': None,
```
cburgdorf: @pipermerriam @carver I've tried to decipher what `go-ethereum` does but I'm not entirely sure if I'm reading the code correctly. Am I right to assume that the only way for us to implement `contractAddress` (and maybe some other fields, too) is to store some additional meta data in the database in addition to the bare receipts? It seems like `contractAddress` is that kind of thing that we only have access to after the transaction was applied and we still have a hold of the `computation` (with the `contractAdress` being on `computation,msg.storage_address`)

Or is there some magical way to retrieve it later on the fly that I'm missing?

carver: Yup, we can regenerate the contract address using the sender and the nonce.

cburgdorf: I thought about that, too and I also found a line in the go-ethereum code that indicates something like that.

```go
        // if the transaction created a contract, store the creation address in the receipt.
	if msg.To() == nil {
		receipt.ContractAddress = crypto.CreateAddress(vmenv.Context.Origin, tx.Nonce())
	}
```
But I don't fully understand how this holds true as contracts can also be deployed using `CREATE2` which uses a different scheme.

Also the [go-ethereum](https://github.com/ethereum/go-ethereum/blob/c2d65d34d5c6f27b8d1a52280964023a3eefd66e/core/state_processor.go#L114-L126) seems to indicate that they create all this meta data at the time when the transaction is applied.

cburgdorf: I talked through this with Piper  and he explained to me that this isn't concerned with opcode based address generation which I wasn't aware of. :sweat_smile: 

In other words, it's clear to me now.

trinity/_utils/address.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Take any byte value and force it into becoming a valid address by padding it to 20 bytes
+    or returning the first 20 bytes in case the provided value is longer than 20 bytes.
+    """
+    trimmed_value = value[-20:]
+    padded_value = trimmed_value.rjust(20, b'\x00')
+    return Address(padded_value)
+
+
+def generate_contract_address(sender: Address, nonce: int) -> Address:
```
cburgdorf: I was surprised that we don't have this in `eth_utils` but then I noticed that `eth_utils` does currently not depend on RLP and I guess we want to keep it that way. I could add `force_bytes_to_address` to `eth_utils` to at least not duplicate that one. Not sure if it's super valuable to have it there though.

carver: So I *think* the ideal solution would be something like: modify py-evm to have each VM reference this method:
https://github.com/ethereum/py-evm/blob/f25d209d114e22c1efe6d5a3947f270d227c5469/eth/_utils/address.py#L17-L18

(Since it could theoretically change at a fork)

Then do something like: `VM.generate_contract_address(sender, nonce)`

---

I suppose we could also just import from `eth._utils.address`. But it seems dirty since it's "private" to py-evm, which maybe it *should* be, for the same reason that I think it belongs on the VM, above.

Whatever the solution, duplicating the logic here feels wrong. If the first suggestion above doesn't feel right, then just importing the private util is better than duplicating it, IMO.

cburgdorf: Yeah, I didn't want to import from `_utils` since I feel we kinda try to maintain a strong emphasis on these things being private. I didn't like the copy and paste either but wanted to include it at least temporary as the least of all evils. But I like your proposed solution and will go with that :+1:

cburgdorf: I implemented this with a tiny twist so that it can be used in the transaction executor where the VM is not accessible (https://github.com/ethereum/py-evm/pull/1897). I also created a cleanup PR so that the cleanup isn't forgotten in case we move on with this PR without waiting for the API to be released in Py-EVM (https://github.com/ethereum/trinity/pull/1380)

cburgdorf: Ok, looks like we are going to live with the copy and paste for now https://github.com/ethereum/py-evm/pull/1897#issuecomment-564259364

trinity/chains/light.py
```diff
@@ -195,9 +198,29 @@ def create_unsigned_transaction(cls,
...
         raise NotImplementedError("Chain classes must implement " + inspect.stack()[0][3])
 
+    async def coro_get_transaction_receipt(self, transaction_hash: Hash32) -> ReceiptAPI:
+        raise NotImplementedError("Chain classes must implement " + inspect.stack()[0][3])
+
+    async def coro_get_transaction_receipt_by_index(
+            self,
+            block_number: BlockNumber,
+            index: int) -> ReceiptAPI:
+        raise NotImplementedError("Chain classes must implement " + inspect.stack()[0][3])
```
cburgdorf: A bit more evidence that removing LES support might be good to reduce scope and complexity for now #1366

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
```diff
@@ -460,15 +474,96 @@ class MainnetFullChain(FullChain):
...
+async def setup_rpc_server(event_bus, chain_fixture, fixture_path):
     chain = MainnetFullChain(None)
     rpc = RPCServer(initialize_eth1_modules(chain, event_bus), chain, event_bus)
 
     setup_result, setup_error = await call_rpc(rpc, 'evm_resetToGenesisFixture', [chain_fixture])
     # We need to advance the event loop for modules to be able to pickup the new chain
     await asyncio.sleep(0)
-    assert setup_error is None and setup_result is True, "cannot load chain for {0}".format(fixture_data)  # noqa: E501
+    return rpc, setup_result, setup_error
+    assert setup_error is None and setup_result is True, f"cannot load chain for {fixture_path}"
```
carver: assert after return?

cburgdorf: Oops, looks like my brain went out halfway while I was writing this. :see_no_evil:

cburgdorf: Fixed it.

cburgdorf:
@gsalgado do you mind reviewing that one? If no bigger changes are needed this could potentially still make it into the release we have to cut for the Muir Glacier fork.

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
```diff
@@ -55,7 +55,7 @@
 ROOT_PROJECT_DIR = Path(__file__).parent.parent.parent
 
 
-BASE_FIXTURE_PATH = os.path.join(ROOT_PROJECT_DIR, 'fixtures', 'BlockchainTests')
+BASE_FIXTURE_PATH = ROOT_PROJECT_DIR / 'fixtures' / 'BlockchainTests'
```
gsalgado: What's the reason for this change?

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
```diff
@@ -384,6 +385,18 @@ def validate_rpc_transaction_vs_fixture(transaction, fixture):
...
 
 
+async def validate_transaction_receipt(rpc, tx_hash, block_fixture, transaction_fixture):
+    result, error = await call_rpc(rpc, 'eth_getTransactionReceipt', [tx_hash])
+    assert error is None
+    header_fixture = fixture_block_in_rpc_format(block_fixture['blockHeader'])
+    tx_index = to_int(hexstr=result['transactionIndex'])
+    tx_fixture = block_fixture['transactions'][tx_index]
+
+    assert result['to'] == "0x" if tx_fixture['to'] == "" else tx_fixture['to']
```
gsalgado: I was going to suggest a custom assert function here but it turns out we already have `eth_utils.add_0x_prefix`. Care to use that here?

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
cburgdorf: We are using it further down `fixture_path = BASE_FIXTURE_PATH / fixture[0]` and `pathlib.Path` allows for easy path joining via `/`. We've been trying to update all places in the code base to migrate to `pathlib.Path` but a few places in the test like this one are still left to replace.

tests/json-fixtures-over-rpc/test_rpc_fixtures.py
cburgdorf: Oh good point, will use that :+1:

PR #1338, opened 2019-11-29
Fix misc test error

ChihChengLiang:
### What was wrong?

`trinity-beacon` failed to run due to a wrong parameter (`trinity_config`) passed in `run_database_process`.

The deeper cause is that the integration tests are not detecting this exception. The test will still pass even if `contains_all` did not read all the keywords before timeout.

```python
async with AsyncProcessRunner.run(command, timeout_sec=30) as runner:
    assert await contains_all(runner.stderr, {"Validator", "BCCReceiveServer"})
```

### How was it fixed?

- Pass correct parameter to `run_database_process`
- Remove unused component test `test_beacon_component.py`, which is equivalently covered in `test_interops.py`
- Fix timeout in AsyncProcessRunner, thanks to @mhchia for the advice.

#### Cute Animal Picture


ChihChengLiang:
It turns out the py36 eth2 integration test won't work since a py37 feature `time_ns` is used in libp2p. Will disable eth2 integration in py36.

ralexstokes:
disabling the py36 variant is probably best for now although we should soon figure out what we want to do about the incompatible py-libp2p version

trinity/tools/async_process_runner.py
```diff
@@ -55,7 +55,11 @@ class AsyncProcessRunner():
             awaitable_bytes_fn: Callable[[], Awaitable[bytes]]) -> AsyncIterable[str]:
 
         while True:
-            line = await awaitable_bytes_fn()
+            try:
+                line = await awaitable_bytes_fn()
+            except Exception:
```
pipermerriam: Can this be tightened up to a more concrete exception.  What error was being raised?

tests/integration/test_trinity_cli.py
```diff
@@ -107,7 +107,7 @@ def amend_command_for_unused_port(command, unused_tcp_port):
         assert await contains_all(runner.stderr, {
             "Started DB server process",
             "Component started: Sync / PeerPool",
-            "Transaction pool does not support light mode",
+            "Transaction pool disabled.  Not supported in light mode.",
```
cburgdorf: Uh, so that was broken all the time but the test still passed because we had an error in the runner where when the runner was cancelled the assert here would still pass? That sounds really bad. Great find!

tests/integration/test_trinity_cli.py
```diff
@@ -264,7 +265,8 @@ def amend_command_for_unused_port(command, unused_tcp_port):
             # is resolved, the following should work.
             # {'DiscoveryProtocol  >>> ping'},
             # {'Started main process'},
-        )
+            marks=(pytest.mark.xfail),  # investigate in #1347
```
cburgdorf: I'm not a fan of disabling these but I assume that this PR does not actually break these cases but rather just *expose*  that they are currently broken/flaky. So, I'm ok to move the investigation off into another PR. Minor nitpick: Can we put an actual `TODO:` at the start of the line so that it stands out and is easy to search for?

trinity/tools/async_process_runner.py
```diff
@@ -55,7 +55,11 @@ class AsyncProcessRunner():
             awaitable_bytes_fn: Callable[[], Awaitable[bytes]]) -> AsyncIterable[str]:
 
         while True:
-            line = await awaitable_bytes_fn()
+            try:
+                line = await awaitable_bytes_fn()
+            except asyncio.CancelledError:
+                # Return to keep the consumer of the AsyncIterable running
+                return
```
cburgdorf: Great find :+1: Working with `asyncio` remains a black magic art :sparkles:

trinity/tools/async_process_runner.py
ChihChengLiang: (mask off 🐶✨👻) It was an `asyncio.CancelledError`.

tests/integration/test_trinity_cli.py
ChihChengLiang: Yes, your assumption is correct, they are not exposed prior to this PR. Added the TODO comments.

trinity/tools/async_process_runner.py
ChihChengLiang: I feel you now, doing asyncio is hard 😂.

PR #1342, opened 2019-12-02
bump pyrlp version to get rid of deprecation warnings

pipermerriam:
### What was wrong?

`rlp==1.1.0` raises deprecation warnings.

### How was it fixed?

updated to `>=1.2.0` that has the fix for the deprecations warnings.

#### Cute Animal Picture

2918709_1280x720


pipermerriam:
The warnings aren't actually gone.  Need to do one more PR to `pyrlp` to fix them (thought they were already fixed but I was wrong)

Issue #1312, opened 2019-11-21
Trinity is not 100% Istanbul-ready due to bug in current py-evm version


veox:
* Trinity version: [v0.1.0-alpha.30](https://github.com/ethereum/trinity/releases/tag/v0.1.0-alpha.30)
* `py-evm` version: [v0.3.0-alpha.8](https://github.com/ethereum/trinity/blob/074e887/setup.py#L7)

### What is wrong?

`py-evm` had [a stray change, a difficulty bomb delay](https://github.com/ethereum/py-evm/issues/1877) that didn't actually go into Istanbul. [The fix](https://github.com/ethereum/py-evm/pull/1878) has been applied, but is not yet in a tagged `py-evm` release, and so isn't in Trinity.

Come Istanbul, Trinity will apply the bomb delay, and start miscalculating block difficulty.

### How can it be fixed

Pin to the fixed version of `py-evm` when it is released, and make a new Trinity release.

veox:
`py-evm` `py-evm 0.3.0-alpha.9` has the fix:

https://py-evm.readthedocs.io/en/latest/release_notes.html#py-evm-0-3-0-alpha-9-2019-12-02

PR #1343, opened 2019-12-02
Upgrade py-evm to v0.3.0a9

carver:
### What was wrong?

A bug in the Istanbul implementation is in the pinned version of py-evm.

### How was it fixed?

Upgrade py-evm to latest version which fixes the bug:
https://py-evm.readthedocs.io/en/latest/release_notes.html#py-evm-0-3-0-alpha-9-2019-12-02

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1343.feature.rst
+++ b/newsfragments/1343.feature.rst
@@ -1,0 +1,2 @@
+Upgrade py-evm to v0.3.0-alpha.9, for Istanbul fix. See the `py-evm release notes
+<https://py-evm.readthedocs.io/en/latest/release_notes.html#py-evm-0-3-0-alpha-9-2019-12-02>`_
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ deps = {
         "coincurve>=10.0.0,<11.0.0",
         "dataclasses>=0.6, <1;python_version<'3.7'",
         "eth-utils>=1.8.1,<2",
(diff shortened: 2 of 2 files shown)


carver:
The wheel failures are real, from a recent ipython release. We were hooking into some non-public API which disappeared, and prevented us from attaching. So all the `trinity attach` commands were failing.

carver:
Ok, so ipython 7.10.1 gives a mysterious blinking `Press ENTER to continue...`. 7.10.0 shows what's going on by spitting out this warning over and over:

```
Press ENTER to continue...                                                                                                                                                                                                                    
Exception 'Loop' object has no attribute '_ready'

Unhandled exception in event loop:
  File "uvloop/cbhandles.pyx", line 68, in uvloop.loop.Handle._run
  File "/tmp/tmphvl_lsv2/package-smoke-test/lib/python3.6/site-packages/prompt_toolkit/eventloop/utils.py", line 69, in schedule
    if not loop2._ready:  # type: ignore
Press ENTER to continue...                                                                                                                                                                                                                    

Exception 'Loop' object has no attribute '_ready'
```

...(etc)

So it looks like it's our fault for hooking into some private API, which is allowed to change in a non-major release. I'll pin to <7.10 for now.

carver:
Closed #1312

PR #1344, opened 2019-12-02
Show more helpful information in logs

carver:
Especially:

- show peer info in more logs and exceptions
- add extra information about exception context & cause
- show eth_api stats in the Peer details again
- show which service is stuck waiting on subservices

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/kademlia.py
+++ b/p2p/kademlia.py
@@ -118,7 +118,7 @@ class Node(NodeAPI):
         return f'enode://{hexstring}@{self.address.ip}:{self.address.tcp_port}'
 
     def __str__(self) -> str:
-        return f"<Node({self.pubkey.to_hex()[:6]}@{self.address.ip})>"
+        return f"<Node({self.pubkey.to_hex()[:8]}@{self.address.ip})>"
 
     def __repr__(self) -> str:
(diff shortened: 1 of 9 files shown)


p2p/kademlia.py
```diff
@@ -118,7 +118,7 @@ def uri(self) -> str:
         return f'enode://{hexstring}@{self.address.ip}:{self.address.tcp_port}'
 
     def __str__(self) -> str:
-        return f"<Node({self.pubkey.to_hex()[:6]}@{self.address.ip})>"
+        return f"<Node({self.pubkey.to_hex()[:8]}@{self.address.ip})>"
```
carver: 2 bytes is a little too short to identify a peer, sometimes

p2p/peer_pool.py
```diff
@@ -520,6 +520,8 @@ def _peer_finished(self, peer: AsyncioServiceAPI) -> None:
                     "client_version_string='%s'",
                     peer.p2p_api.safe_client_version_string,
                 )
+                if not hasattr(peer, "eth_api"):
+                    self.logger.warning("Huh? %s doesn't have an eth API", peer)
```
carver: This seems like the wrong place for this, but non-eth peers seem to be slipping in here somehow...

Naturally, it should be removed sooner than later.

PR #1345, opened 2019-12-02
Speedups during launch and shutdown

carver:
### What was wrong?

Two unrelated slow components:
- Some service tasks were hanging and and to be force-killed (which adds a 5 second delay)
- When using a checkpoint to beam sync, if you relaunch with the same checkpoint, you try to download the header from peers no matter what

### How was it fixed?

- Add some `self.wait()` wrappers to make sure subtasks always shut down cleanly when service is cancelled
- Check if header is checkpoint header is present locally before trying to download it from peers

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/common/strategies.py
+++ b/trinity/sync/common/strategies.py
@@ -19,6 +19,9 @@ from eth.constants import (
     GENESIS_BLOCK_NUMBER,
     GENESIS_PARENT_HASH,
 )
+from eth.exceptions import (
+    HeaderNotFound,
+)
 
 from p2p.disconnect import DisconnectReason
 from p2p.exceptions import (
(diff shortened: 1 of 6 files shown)


trinity/sync/common/strategies.py
```diff
@@ -109,6 +112,15 @@ def __init__(self,
         self._peer_pool = peer_pool
 
     async def fulfill_prerequisites(self) -> None:
+        try:
+            checkpoint = await self._db.coro_get_block_header_by_hash(self._checkpoint.block_hash)
```
carver: I just hacked this together to speed up my launch. I guess I should check if the following `FULL_BLOCKS_NEEDED_TO_START_BEAM - 1` block headers are available as well. Does anything else stand out as problematic?

cburgdorf: >I guess I should check if the following `FULL_BLOCKS_NEEDED_TO_START_BEAM - 1` block headers are available as well

Sounds right :+1: 

>Does anything else stand out as problematic?

Nothing from the top of my head. LGTM

PR #1351, opened 2019-12-04
Add more flexible fork choice scoring

### LINKED ISSUE #782: LMD GHOST scoring is broken
### What is wrong?

See https://github.com/ethereum/trinity/pull/685/files/b4dbf6a58d444578ee1b882ca21256eaf1a5f6a4#r294035068 on issue #685.

The score is the sum of the attestation data and the block root data but this obscures the semantics of the fork choice scoring.

The attestation data yields a balance and only upon ties do we want to consider the block root data. The code in #685 just sums them which could result in mis-scoring a block where the balances are close and the block root puts the incorrect block as having a higher score.

### How can it be fixed

We need a more flexible score type for the fork choice rule (including a way to {de,}serialize from disk).

1. The `ForkChoiceScoringFn` type should return an opaque `Score` type (instead of just an `int`).
2. Different fork choice scorings can provide different implementations of `Score` as makes sense -- the main thing `Score` needs is to satisfy a comparator interface and a {de,}serializing ability for writing scores to the DB.

something like:

```python
class Score(abc.ABC):
    @abstractmethod
    def cmp(self, other: Score) -> int:
        """ Return [-1, 0, 1] if ``self`` is [less than, equal to, greater than] ``other``."""
        pass

    @abstractmethod
    def serialize(self) -> bytesOrSomething:
        pass

    @abstractmethod
    @classmethod
    def deserialize(cls, contents: bytesOrSomething) -> Score:
        pass
```

ralexstokes:
### What was wrong?

Fixes #782.

### How was it fixed?

Introduces an opaque "score" type so that different forks of the protocol can specify their own format for scoring a block tree when computing the fork choice.

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/chains/base.py
+++ b/eth2/beacon/chains/base.py
@@ -8,10 +8,11 @@ from eth.exceptions import BlockNotFound
 from eth.validation import validate_word
 from eth_utils import ValidationError, humanize_hash
 
-from eth2._utils.funcs import constantly
 from eth2._utils.ssz import validate_imported_block_unchanged
 from eth2.beacon.db.chain import BaseBeaconChainDB, BeaconChainDB
 from eth2.beacon.exceptions import BlockClassError, StateMachineNotFound
+from eth2.beacon.fork_choice.constant import ConstantScoring
+from eth2.beacon.fork_choice.scoring import BaseScore
 from eth2.beacon.types.attestations import Attestation
 from eth2.beacon.types.blocks import BaseBeaconBlock
 from eth2.beacon.types.states import BeaconState
@@ -139,7 +140,7 @@ class BaseBeaconChain(Configurable, ABC):
         ...
 
     @abstractmethod
-    def get_score(self, block_root: SigningRoot) -> int:
+    def get_score(self, block_root: SigningRoot) -> BaseScore:
         ...
 
     @abstractmethod
@@ -224,7 +225,7 @@ class BeaconChain(BaseBeaconChain):
         genesis_config: Eth2GenesisConfig,
     ) -> "BaseBeaconChain":
         """
-        Initialize the ``BeaconChain`` from a genesis state.
+        Initialize the ``BeaconChain`` from a genesis state and block.
         """
         sm_class = cls.get_state_machine_class_for_block_slot(genesis_block.slot)
         if type(genesis_block) != sm_class.block_class:
@@ -236,21 +237,13 @@ class BeaconChain(BaseBeaconChain):
 
         chaindb = cls.get_chaindb_class()(db=base_db, genesis_config=genesis_config)
         chaindb.persist_state(genesis_state)
-        return cls._from_genesis_block(base_db, genesis_block, genesis_config)
 
-    @classmethod
-    def _from_genesis_block(
-        cls,
-        base_db: AtomicDatabaseAPI,
-        genesis_block: BaseBeaconBlock,
-        genesis_config: Eth2GenesisConfig,
-    ) -> "BaseBeaconChain":
-        """
-        Initialize the ``BeaconChain`` from the genesis block.
-        """
-        chaindb = cls.get_chaindb_class()(db=base_db, genesis_config=genesis_config)
-        genesis_scoring = constantly(0)
+        genesis_scoring_class = sm_class.get_fork_choice_scoring_class()
+        genesis_score_class = genesis_scoring_class.get_score_class()
+        genesis_score = genesis_score_class.from_genesis(genesis_state, genesis_block)
+        genesis_scoring = ConstantScoring(genesis_score)
         chaindb.persist_block(genesis_block, genesis_block.__class__, genesis_scoring)
+
         return cls(base_db, genesis_config)
 
     #
@@ -375,13 +368,17 @@ class BeaconChain(BaseBeaconChain):
         block_class = self.get_block_class(block_root)
         return self.chaindb.get_block_by_root(block_root, block_class)
 
-    def get_score(self, block_root: SigningRoot) -> int:
+    def get_score(self, block_root: SigningRoot) -> BaseScore:
         """
         Return the score of the block with the given hash.
 
         Raise ``BlockNotFound`` if there is no matching black hash.
         """
-        return self.chaindb.get_score(block_root)
+        block = self.get_block_by_root(block_root)
+        slot = block.slot
+        state_machine = self.get_state_machine(at_slot=slot)
+        fork_choice_scoring = state_machine.get_fork_choice_scoring()
+        return self.chaindb.get_score(block_root, fork_choice_scoring.get_score_class())
 
     def get_canonical_block_by_slot(self, slot: Slot) -> BaseBeaconBlock:
         """
@@ -429,7 +426,9 @@ class BeaconChain(BaseBeaconChain):
             raise ValidationError(
                 "Attempt to import block #{}.  Cannot import block {} before importing "
                 "its parent block at {}".format(
-                    block.slot, block.signing_root, block.parent_root
+                    block.slot,
+                    humanize_hash(block.signing_root),
+                    humanize_hash(block.parent_root),
                 )
             )
 
--- a/eth2/beacon/db/chain.py
+++ b/eth2/beacon/db/chain.py
@@ -17,11 +17,11 @@ from eth2.beacon.db.exceptions import (
     HeadStateSlotNotFound,
     JustifiedHeadNotFound,
     MissingForkChoiceContext,
-    MissingForkChoiceScoringFns,
+    MissingForkChoiceScorings,
     StateNotFound,
 )
 from eth2.beacon.db.schema import SchemaV1
-from eth2.beacon.fork_choice.scoring import ScoringFn as ForkChoiceScoringFn
+from eth2.beacon.fork_choice.scoring import BaseForkChoiceScoring, BaseScore
 from eth2.beacon.helpers import compute_epoch_at_slot
 from eth2.beacon.types.blocks import BaseBeaconBlock, BeaconBlock  # noqa: F401
 from eth2.beacon.types.states import BeaconState  # noqa: F401
@@ -50,7 +50,7 @@ class BaseBeaconChainDB(ABC):
         self,
         block: BaseBeaconBlock,
         block_class: Type[BaseBeaconBlock],
-        fork_choice_scoring: ForkChoiceScoringFn,
+        fork_choice_scoring: BaseForkChoiceScoring,
     ) -> Tuple[Tuple[BaseBeaconBlock, ...], Tuple[BaseBeaconBlock, ...]]:
         ...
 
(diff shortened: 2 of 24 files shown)


ralexstokes:
CI failures seem unrelated to this PR

eth2/beacon/chains/base.py
```diff
@@ -236,21 +237,13 @@ def from_genesis(
...
-        genesis_config: Eth2GenesisConfig,
-    ) -> "BaseBeaconChain":
-        """
-        Initialize the ``BeaconChain`` from the genesis block.
-        """
-        chaindb = cls.get_chaindb_class()(db=base_db, genesis_config=genesis_config)
-        genesis_scoring = constantly(0)
+        genesis_scoring_class = sm_class.get_fork_choice_scoring_class()
+        genesis_score_class = genesis_scoring_class.get_score_class()
+        genesis_score = genesis_score_class.from_genesis(genesis_state, genesis_block)
```
ralexstokes: the scoring used for the `genesis_block` requires the `genesis_state` in the score's `from_genesis` method so i merged the two functions (for genesis block and genesis state) into one function.

eth2/beacon/chains/base.py
```diff
@@ -429,7 +426,9 @@ def import_block(
             raise ValidationError(
                 "Attempt to import block #{}.  Cannot import block {} before importing "
                 "its parent block at {}".format(
-                    block.slot, block.signing_root, block.parent_root
+                    block.slot,
+                    humanize_hash(block.signing_root),
+                    humanize_hash(block.parent_root),
```
ralexstokes: just formatting on error message

eth2/beacon/db/chain.py
```diff
@@ -412,18 +414,22 @@ def _get_block_signing_root_by_hash_tree_root(
...
+        db: DatabaseAPI, block_root: SigningRoot, score_class: Type[BaseScore]
+    ) -> BaseScore:
         try:
             encoded_score = db[SchemaV1.make_block_root_to_score_lookup_key(block_root)]
         except KeyError:
             raise BlockNotFound(
                 "No block with signing root {0} found".format(encode_hex(block_root))
             )
-        return ssz.decode(encoded_score, sedes=ssz.sedes.uint64)
+        return cast(BaseScore, score_class.deserialize(encoded_score))
```
ralexstokes: `cast` required to satisfy `mypy`

eth2/beacon/db/chain.py
```diff
@@ -437,31 +443,37 @@ def persist_block_chain(
...
         Return two iterable of blocks, the first containing the new canonical blocks,
         the second containing the old canonical headers
         """
         with self.db.atomic_batch() as db:
+            try:
+                first_block = next(iter(blocks))
+                if first_block.is_genesis:
+                    self._handle_exceptional_justification_and_finality(first_block)
+            except StopIteration:
+                pass
```
ralexstokes: there was actually a bit of a bug here in that we were only checking for the genesis block in `persist_block` and *not* in `persist_block_chain`.

a future refactoring could unify this somewhat, but for now we just check in both places.

eth2/beacon/fork_choice/constant.py
```diff
@@ -0,0 +1,29 @@
+from typing import Type
+
+from eth2.beacon.fork_choice.scoring import BaseForkChoiceScoring, BaseScore
+from eth2.beacon.types.blocks import BaseBeaconBlock
+from eth2.beacon.types.states import BeaconState
+
+
+class ConstantScoring(BaseForkChoiceScoring):
```
ralexstokes: this class is a bit of plumbing/proxy to satisfy the scoring API

eth2/beacon/fork_choice/lmd_ghost.py
```diff
@@ -21,6 +24,47 @@
...
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, LMDGHOSTScore):
+            return NotImplemented
+        return self._score == other._score
+
+    def serialize(self) -> bytes:
+        return ssz.encode(
+            self._score,
+            sedes=ssz.sedes.Vector(ssz.sedes.uint256, LMD_GHOST_SCORE_DATA_LENGTH),
+        )
```
ralexstokes: i like that we can now represent scores for LMD GHOST correctly, and not collapse to a single `int`.

eth2/beacon/chains/base.py
```diff
@@ -236,21 +237,13 @@ def from_genesis(
...
-        genesis_block: BaseBeaconBlock,
-        genesis_config: Eth2GenesisConfig,
-    ) -> "BaseBeaconChain":
-        """
-        Initialize the ``BeaconChain`` from the genesis block.
-        """
-        chaindb = cls.get_chaindb_class()(db=base_db, genesis_config=genesis_config)
-        genesis_scoring = constantly(0)
+        genesis_scoring_class = sm_class.get_fork_choice_scoring_class()
+        genesis_score_class = genesis_scoring_class.get_score_class()
```
ralexstokes: given that this is only place we are really using `get_fork_choice_scoring_class`, we could arguably collapse these two functions and only expose `get_score_class` on the `sm_class`.

eth2/beacon/tools/factories.py
```diff
@@ -21,33 +25,68 @@ class BeaconChainFactory(factory.Factory):
...
+                if block.is_genesis:
+                    continue
+                # NOTE: ideally we use the ``import_block`` method
+                # on ``chain`` but for the time being we skip some
+                # validation corresponding to assumptions made in clients of
+                # this class. A future refactoring should use the external API.
+                chain.chaindb.persist_block(
+                    block, SerenityBeaconBlock, HigherSlotScoring()
+                )
+
```
ralexstokes: needed more control over genesis and how we bootstrap a `Chain` object in the factory during the syncing tests... so added some plumbing here to do this...

tests/libp2p/bcc/test_receive_server.py
```diff
@@ -51,14 +51,16 @@ def import_block(
...
+            block, block.__class__, HigherSlotScoring()
         )
         return block, new_canonical_blocks, old_canonical_blocks
 
 
 async def get_fake_chain() -> FakeChain:
     genesis_config = Eth2GenesisConfig(MINIMAL_SERENITY_CONFIG)
     chain_db = AsyncBeaconChainDBFactory(genesis_config=genesis_config)
+    genesis_block = BeaconBlockFactory()
+    chain_db.persist_block(genesis_block, SerenityBeaconBlock, HigherSlotScoring())
```
ralexstokes: took essentially this logic out of `AsyncBeaconChainDBFactory` so replicating the missing pieces here

tests/libp2p/bcc/test_syncing.py
```diff
@@ -12,14 +14,20 @@
...
+    alice_chaindb = AsyncBeaconChainDBFactory()
+    bob_chaindb = AsyncBeaconChainDBFactory()
     peer_pair = ConnectionPairFactory(
-        alice_chaindb=alice_chaindb, bob_chaindb=bob_chaindb
+        alice_chaindb=alice_chaindb,
+        alice_branch=alice_branch,
+        bob_chaindb=bob_chaindb,
+        bob_branch=bob_branch,
+        genesis_state=genesis_state,
     )
```
ralexstokes: given the ordering of the objects here and the `asynccontextmanager` natuer of `ConnectionPairFactory`, had to do some modification to construct the genesis properly and in an order that set the rest of the assumptions used during syncing... like even having the genesis block (or our highest block) as the canonical head when resolving which blocks to sync

trinity/tools/bcc_factories.py
```diff
@@ -108,19 +111,29 @@ class Meta:
...
+    genesis_state: BeaconState = None,
     cancel_token: CancelToken = None,
     handshake: bool = True,
 ) -> AsyncIterator[Tuple[Node, Node]]:
     if cancel_token is None:
         cancel_token = CancelTokenFactory()
-    alice_kwargs = {}
-    bob_kwargs = {}
+    alice_kwargs: Dict[str, Any] = {}
+    bob_kwargs: Dict[str, Any] = {}
```
ralexstokes: `mypy` needed explicit typing here

eth2/beacon/fork_choice/lmd_ghost.py
```diff
@@ -21,6 +24,47 @@
...
 
+LMD_GHOST_SCORE_DATA_LENGTH = 2
+
+
+def score_block_by_root(root: SigningRoot) -> int:
+    return int.from_bytes(root, byteorder="big")
+
+
+class LMDGHOSTScore(BaseScore):
+    _score: Tuple[Gwei, int]
```
hwwhww: Perhaps add a comment like:
```suggestion
    # First score by the latest attesting balance, and then score by block root
    _score: Tuple[Gwei, int]  
```

eth2/beacon/fork_choice/lmd_ghost.py
```diff
@@ -21,6 +24,47 @@
...
+            data, sedes=ssz.sedes.Vector(ssz.sedes.uint256, LMD_GHOST_SCORE_DATA_LENGTH)
+        )
+        return cls(score)
+
+    @classmethod
+    def from_genesis(
+        cls, genesis_state: BeaconState, genesis_block: BaseBeaconBlock
+    ) -> BaseScore:
+        score = (Gwei(0), score_block_by_root(SigningRoot(ZERO_HASH32)))
+        return LMDGHOSTScore(score)
```
hwwhww: nitpick, just notice that you use `return cls(score)` in `deserialize`.
```suggestion
        return cls(score)
```

eth2/beacon/fork_choice/lmd_ghost.py
```diff
@@ -377,7 +417,7 @@ def on_attestation(
                     epoch=target.epoch, root=attestation.data.beacon_block_root
                 )
 
-    def scoring(self, block: BaseBeaconBlock) -> int:
+    def scoring(self, block: BaseBeaconBlock) -> LMDGHOSTScore:
         """
         Return the score of the target ``_block`` according to the LMD GHOST algorithm,
```
hwwhww:
```suggestion
        Return the score of the target ``block`` according to the LMD GHOST algorithm,
```

ralexstokes:
yep, good point about comparing scores across forks -- this seems like a case where we can just extend the scoring metrics for each class to account for any special rules required at the fork!

thanks for the review!!

eth2/beacon/fork_choice/lmd_ghost.py
ralexstokes: oh thank you! silly oversight on my part :)

eth2/beacon/fork_choice/lmd_ghost.py
ralexstokes: thanks! this got lost in a prior edit

PR #1354, opened 2019-12-05
Silence warnings of connection "already cancelled"

### LINKED ISSUE #1353: Log spam: "Tried to cancel <p2p.connection.Connection ...>, but it was already cancelled
* Trinity version: master

### What is wrong?

Related to #373

It looks like we try to shut down the connection twice, which generates some warnings. I think it's preferable if (usually) we have exactly one path to shutting down a service.

### How can it be fixed

After a brief look, it seems that we are cancelling the connection twice:
- via the cancel token passed into the peer factory
- via the peer's `_cleanup()` method

Can we just remove the cancel token and rely on the peer's cleanup? The opposite would not work, because the peer can shut down without the parent's cancel-token triggering.

cc @pipermerriam

carver:
### What was wrong?

Fixes #1353 

### How was it fixed?

I did the laziest possible thing and skipped the cancel if it's already cancelled.

I *think* one of the other options listed in #1353 would be better, but want some input from @pipermerriam before I go that route.

---

Bonus: I added the log that I used to track down where the second cancellation is happening.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/eth/commands.py
+++ b/trinity/protocol/eth/commands.py
@@ -13,9 +13,6 @@ from eth.rlp.headers import BlockHeader
 from eth.rlp.receipts import Receipt
 from eth.rlp.transactions import BaseTransactionFields
 
-from p2p.abc import (
-    SerializationCodecAPI,
-)
 from p2p.commands import BaseCommand, RLPCodec
 
 from trinity.protocol.common.payloads import BlockHeadersQuery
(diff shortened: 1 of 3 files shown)


carver:
This failure from py36-integration seems a little extreme as an integration test:
>   trinity.components.builtin.syncer.etherscan_api.EtherscanAPIError: Invalid status code: 502, Bad Gateway

I mean, I guess we want to know if their API changes. Hopefully we'll be able to move to more of an active monitoring system, rather than lump it in with the PRs. For now, I'll just ignore that error.

carver:
Etherscan appears to have been down (saw the same thing on ethernodes.org), which flagged a problem in CI. Everything else seems good.

carver:
ping @pipermerriam

trinity/protocol/eth/commands.py
```diff
@@ -51,7 +54,7 @@ class Status(BaseCommand[StatusPayload]):
 
 class NewBlockHashes(BaseCommand[Tuple[NewBlockHash, ...]]):
     protocol_command_id = 1
-    serialization_codec = RLPCodec(
+    serialization_codec: SerializationCodecAPI[Tuple[NewBlockHash, ...]] = RLPCodec(
```
pipermerriam: I know this was in another PR that is already merged but this should be `RLPCodec[Tuple[NewBlockHash, ...]]`

carver:
Unrelated #1362 is the CI error. Rage-merging...

PR #1355, opened 2019-12-05
Quiet some *really* noisy errors on exit

carver:
### What was wrong?

Sometimes, there is a `BrokenPipeError` on shutdown. Typically, a long stack trace for each transaction preview that the beam importer is in the middle of. Something recent (maybe the faster shutdowns) is causing this to show up more often.

### How was it fixed?

For now, especially for just the previews, which are basically throwaway actions, we can just completely suppress/ignore this error.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/importer.py
+++ b/trinity/sync/beam/importer.py
@@ -511,6 +511,9 @@ def partial_speculative_execute(
                 preview_time,
                 exc,
             )
+        except BrokenPipeError:
+            vm.logger.info("Got BrokenPipeError while Beam Sync speculatively executed %s", header)
+            vm.logger.debug("BrokenPipeError trace for %s", header, exc_info=True)
         else:
             preview_time = t.elapsed
 
--- a/trinity/sync/beam/importer.py
+++ b/trinity/sync/beam/importer.py
@@ -493,6 +493,7 @@ def partial_speculative_execute(
     Get an argument-free function that will trigger missing state downloads,
     by executing all the transactions, in the context of the given header.
     """
+
     def _trigger_missing_state_downloads() -> None:
         vm = beam_chain.get_vm(header)
         unused_header = header.copy(gas_used=0)
--- a/trinity/_utils/decorators.py
+++ b/trinity/_utils/decorators.py
@@ -1,3 +1,4 @@
+import functools
 from typing import (
     Any,
     Awaitable,
@@ -9,8 +10,14 @@ from typing import (
 
 TAsyncFn = TypeVar("TAsyncFn", bound=Callable[..., Awaitable[None]])
 
+# It's only permitted to ignore an exception and return None if
+#   the wrapped function always returns None
+TFn = TypeVar("TFn", bound=Callable[..., None])
 
-def async_suppress_exceptions(*exception_types: Type[BaseException]) -> TAsyncFn:
+
+def async_suppress_exceptions(
+        *exception_types: Type[BaseException]) -> Callable[[TAsyncFn], TAsyncFn]:
+
     def _suppress_decorator(func: TAsyncFn) -> TAsyncFn:
         async def _suppressed_func(*args: Any, **kwargs: Any) -> None:
             try:
@@ -21,4 +28,27 @@ def async_suppress_exceptions(*exception_types: Type[BaseException]) -> TAsyncFn
 
         return _suppressed_func  # type: ignore
 
-    return _suppress_decorator  # type: ignore
+    return _suppress_decorator
+
+
+def suppress_exceptions(*exception_types: Type[BaseException]) -> Callable[[TFn], TFn]:
+    """
+    Sometimes, rarely, it's okay for a function to just stop running,
+    in the case of certain exceptions. Use this decorator, defining the
+    exceptions that are allowed to simply exit.
+    """
+    def _suppress_decorator(func: TFn) -> TFn:
+
+        @functools.wraps(func)
+        def _suppressed_func(*args: Any, **kwargs: Any) -> None:
+            try:
+                func(*args, **kwargs)
+            except exception_types:
+                # these exceptions are expected, and require no handling
+                pass
+
+            return None
+
+        return _suppressed_func  # type: ignore
+
+    return _suppress_decorator
--- a/trinity/protocol/eth/peer.py
+++ b/trinity/protocol/eth/peer.py
@@ -143,7 +143,7 @@ class ETHPeerFactory(BaseChainPeerFactory):
         )
 
 
-async_fire_and_forget = async_suppress_exceptions(PeerConnectionLost, asyncio.TimeoutError)  # type: ignore  # noqa: E501
+async_fire_and_forget = async_suppress_exceptions(PeerConnectionLost, asyncio.TimeoutError)
 
 
 class ETHPeerPoolEventServer(PeerPoolEventServer[ETHPeer]):


carver:
py-36-integration:
Same issue as https://github.com/ethereum/trinity/pull/1354#issuecomment-561917316

carver:
So I think this means that one of the log lines is too long? I've never seen it before:

```
=================================== FAILURES ===================================
_ test_logger_configuration[command2-expected_stderr_logs2-unexpected_stderr_logs2-expected_file_logs2-unexpected_file_logs2] _

self = <StreamReader t=<ReadUnixTransport closed=False reading=True 0x7ff25e373d48>>

    @coroutine
    def readline(self):
        """Read chunk of data from the stream until newline (b'\n') is found.
    
        On success, return chunk that ends with newline. If only partial
        line can be read due to EOF, return incomplete line without
        terminating newline. When EOF was reached while no bytes read, empty
        bytes object is returned.
    
        If limit is reached, ValueError will be raised. In that case, if
        newline was found, complete line including newline will be removed
        from internal buffer. Else, internal buffer will be cleared. Limit is
        compared against part of the line without newline.
    
        If stream was paused, this function will automatically resume it if
        needed.
        """
        sep = b'\n'
        seplen = len(sep)
        try:
>           line = yield from self.readuntil(sep)

self       = <StreamReader t=<ReadUnixTransport closed=False reading=True 0x7ff25e373d48>>
sep        = b'\n'
seplen     = 1

/usr/local/lib/python3.6/asyncio/streams.py:488: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <StreamReader t=<ReadUnixTransport closed=False reading=True 0x7ff25e373d48>>
separator = b'\n'

    @coroutine
    def readuntil(self, separator=b'\n'):
        """Read data from the stream until ``separator`` is found.
    
        On success, the data and separator will be removed from the
        internal buffer (consumed). Returned data will include the
        separator at the end.
    
        Configured stream limit is used to check result. Limit sets the
        maximal length of data that can be returned, not counting the
        separator.
    
        If an EOF occurs and the complete separator is still not found,
        an IncompleteReadError exception will be raised, and the internal
        buffer will be reset.  The IncompleteReadError.partial attribute
        may contain the separator partially.
    
        If the data cannot be read because of over limit, a
        LimitOverrunError exception  will be raised, and the data
        will be left in the internal buffer, so it can be read again.
        """
        seplen = len(separator)
        if seplen == 0:
            raise ValueError('Separator should be at least one-byte string')
    
        if self._exception is not None:
            raise self._exception
    
        # Consume whole buffer except last bytes, which length is
        # one less than seplen. Let's check corner cases with
        # separator='SEPARATOR':
        # * we have received almost complete separator (without last
        #   byte). i.e buffer='some textSEPARATO'. In this case we
        #   can safely consume len(separator) - 1 bytes.
        # * last byte of buffer is first byte of separator, i.e.
        #   buffer='abcdefghijklmnopqrS'. We may safely consume
        #   everything except that last byte, but this require to
        #   analyze bytes of buffer that match partial separator.
        #   This is slow and/or require FSM. For this case our
        #   implementation is not optimal, since require rescanning
        #   of data that is known to not belong to separator. In
        #   real world, separator will not be so long to notice
        #   performance problems. Even when reading MIME-encoded
        #   messages :)
    
        # `offset` is the number of bytes from the beginning of the buffer
        # where there is no occurrence of `separator`.
        offset = 0
    
        # Loop until we find `separator` in the buffer, exceed the buffer size,
        # or an EOF has happened.
        while True:
            buflen = len(self._buffer)
    
            # Check if we now have enough data in the buffer for `separator` to
            # fit.
            if buflen - offset >= seplen:
                isep = self._buffer.find(separator, offset)
    
                if isep != -1:
                    # `separator` is in the buffer. `isep` will be used later
                    # to retrieve the data.
                    break
    
                # see upper comment for explanation.
                offset = buflen + 1 - seplen
                if offset > self._limit:
                    raise LimitOverrunError(
                        'Separator is not found, and chunk exceed the limit',
                        offset)
    
            # Complete message (with full separator) may be present in buffer
            # even when EOF flag is set. This may happen when the last chunk
            # adds data which makes separator be found. That's why we check for
            # EOF *ater* inspecting the buffer.
            if self._eof:
                chunk = bytes(self._buffer)
                self._buffer.clear()
                raise IncompleteReadError(chunk, None)
    
            # _wait_for_data() will resume reading if stream was paused.
            yield from self._wait_for_data('readuntil')
    
        if isep > self._limit:
            raise LimitOverrunError(
>               'Separator is found, but chunk is longer than limit', isep)
E           asyncio.streams.LimitOverrunError: Separator is found, but chunk is longer than limit

buflen     = 375539
isep       = 375538
offset     = 0
self       = <StreamReader t=<ReadUnixTransport closed=False reading=True 0x7ff25e373d48>>
separator  = b'\n'
seplen     = 1

/usr/local/lib/python3.6/asyncio/streams.py:585: LimitOverrunError

command = ('trinity', '-l=DEBUG2', '-l', 'p2p.discovery=ERROR', '--port=44773')
expected_stderr_logs = {'ConnectionTrackerServer  Running task <coroutine object', 'Started main process'}
unexpected_stderr_logs = {'DiscoveryProtocol  >>> ping'}
expected_file_logs = {'ConnectionTrackerServer  Running task <coroutine object', 'Started main process'}
unexpected_file_logs = {'DiscoveryProtocol  >>> ping'}, unused_tcp_port = 44773

    @pytest.mark.parametrize(
        'command,expected_stderr_logs,unexpected_stderr_logs,expected_file_logs,unexpected_file_logs',
        (
            (
                # Default run without any flag
                ('trinity',),
                # Expected stderr logs
                {'Started main process'},
                # Unexpected stderr logs
                {'DiscoveryProtocol  >>> ping'},
                # Expected file logs
                {'Started main process', 'Logging initialized'},
                # Unexpected file logs
                {'DiscoveryProtocol  >>> ping'},
            ),
            (
                # Enable DEBUG2 logs across the board
                ('trinity', '-l=DEBUG2'),
                {'Started main process', 'DiscoveryProtocol  >>> ping'},
                {},
                {'Started main process', 'DiscoveryProtocol  >>> ping'},
                {},
            ),
            (   # Enable DEBUG2 logs for everything except discovery which is reduced to ERROR logs
                ('trinity', '-l=DEBUG2', '-l', 'p2p.discovery=ERROR'),
                {'Started main process', 'ConnectionTrackerServer  Running task <coroutine object'},
                {'DiscoveryProtocol  >>> ping'},
                {'Started main process', 'ConnectionTrackerServer  Running task <coroutine object'},
                {'DiscoveryProtocol  >>> ping'},
            ),
            pytest.param(
                # Reduce stderr logging to ERROR logs but report DEBUG2 or higher for file logs
                ('trinity', '--stderr-log-level=ERROR', '--file-log-level=DEBUG2',),
                {},
                {'Started main process', 'DiscoveryProtocol  >>> ping'},
                {'Started main process', 'DiscoveryProtocol  >>> ping'},
                {},
                # TODO: investigate in #1347
                marks=(pytest.mark.xfail),
            ),
            pytest.param(
                # Reduce everything to ERROR logs, except discovery that should report DEBUG2 or higher
                ('trinity', '-l=ERROR', '-l', 'p2p.discovery=DEBUG2'),
                {'DiscoveryProtocol  >>> ping'},
                {'Started main process'},
                {},
                {},
                # Increasing per-module log level to a higher value than the general log level does
                # not yet work for file logging. Once https://github.com/ethereum/trinity/issues/689
                # is resolved, the following should work.
                # {'DiscoveryProtocol  >>> ping'},
                # {'Started main process'},
                # TODO: investigate in #1347
                marks=(pytest.mark.xfail),
            ),
        )
    )
    @pytest.mark.asyncio
    async def test_logger_configuration(command,
                                        expected_stderr_logs,
                                        unexpected_stderr_logs,
                                        expected_file_logs,
                                        unexpected_file_logs,
                                        unused_tcp_port):
    
        command = amend_command_for_unused_port(command, unused_tcp_port)
    
        def contains_substring(iterable, substring):
            return any(substring in x for x in iterable)
    
        # Saw occasional (<25%, >5%) failures in CI at 30s because of slow machines or bad luck
        async with AsyncProcessRunner.run(command, timeout_sec=45) as runner:
    
            stderr_logs = []
    
            # Collect logs up to the point when the sync begins so that we have enough logs
            # for assertions
            marker_seen_at = 0
>           async for line in runner.stderr:

command    = ('trinity', '-l=DEBUG2', '-l', 'p2p.discovery=ERROR', '--port=44773')
contains_substring = <function test_logger_configuration.<locals>.contains_substring at 0x7ff25e446950>
expected_file_logs = {'ConnectionTrackerServer  Running task <coroutine object', 'Started main process'}
expected_stderr_logs = {'ConnectionTrackerServer  Running task <coroutine object', 'Started main process'}
line       = '  DEBUG2  2019-12-05 01:03:30,714            Connection  Running task <coroutine object BasePeer._handle_subscriber_message at 0x7f537d328258>\n'
marker_seen_at = 0
runner     = <trinity.tools.async_process_runner.AsyncProcessRunner object at 0x7ff25e387ac8>
stderr_logs = ['    INFO  2019-12-05 01:02:51,982               trinity  \n', '      ______     _       _ __       \n', '     /_  __...__/ / __ \\/ / __/ / / /\n', '     / / / /  / / / / / / /_/ /_/ / \n', '    /_/ /_/  /_/_/ /_/_/\\__/\\__, /  \n', ...]
unexpected_file_logs = {'DiscoveryProtocol  >>> ping'}
unexpected_stderr_logs = {'DiscoveryProtocol  >>> ping'}
unused_tcp_port = 44773

tests/integration/test_trinity_cli.py:295: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
trinity/tools/async_process_runner.py:50: in stderr
    async for line in self._iterate_until_empty(self.proc.stderr.readline):
trinity/tools/async_process_runner.py:59: in _iterate_until_empty
    line = await awaitable_bytes_fn()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <StreamReader t=<ReadUnixTransport closed=False reading=True 0x7ff25e373d48>>

    @coroutine
    def readline(self):
        """Read chunk of data from the stream until newline (b'\n') is found.
    
        On success, return chunk that ends with newline. If only partial
        line can be read due to EOF, return incomplete line without
        terminating newline. When EOF was reached while no bytes read, empty
        bytes object is returned.
    
        If limit is reached, ValueError will be raised. In that case, if
        newline was found, complete line including newline will be removed
        from internal buffer. Else, internal buffer will be cleared. Limit is
        compared against part of the line without newline.
    
        If stream was paused, this function will automatically resume it if
        needed.
        """
        sep = b'\n'
        seplen = len(sep)
        try:
            line = yield from self.readuntil(sep)
        except IncompleteReadError as e:
            return e.partial
        except LimitOverrunError as e:
            if self._buffer.startswith(sep, e.consumed):
                del self._buffer[:e.consumed + seplen]
            else:
                self._buffer.clear()
            self._maybe_resume_transport()
>           raise ValueError(e.args[0])
E           ValueError: Separator is found, but chunk is longer than limit

self       = <StreamReader t=<ReadUnixTransport closed=False reading=True 0x7ff25e373d48>>
sep        = b'\n'
seplen     = 1

/usr/local/lib/python3.6/asyncio/streams.py:497: ValueError
```

cburgdorf:
>So I think the py36-wheel-cli error means that one of the log lines is too long? I've never seen it before:

Mmh, this is probably related to the recent changes in #1338

trinity/sync/beam/importer.py
```diff
@@ -493,6 +494,8 @@ def partial_speculative_execute(
     Get an argument-free function that will trigger missing state downloads,
     by executing all the transactions, in the context of the given header.
     """
+
+    @suppress_exceptions(BrokenPipeError)
```
pipermerriam: Can we do this in a way where we still get `ERROR` logs but just as a single line?

I don't believe silencing these exception is the right thing but I expect it to take a few iterations before we can really tackle the problem correctly .

trinity/sync/beam/importer.py
```diff
@@ -510,6 +511,9 @@ def _trigger_missing_state_downloads() -> None:
                 preview_time,
                 exc,
             )
+        except BrokenPipeError:
+            vm.logger.info("Got BrokenPipeError while Beam Sync speculatively executed %s", header)
+            vm.logger.debug("BrokenPipeError trace for %s", header, exc_info=True)
```
carver: I just explicitly grabbed and logged this error for now.

PR #1356, opened 2019-12-05
Cleaner txpool shutdown

carver:
### What was wrong?

The transaction pool was hanging on shutdown, because it had a bare `await` that doesn't get cancelled when the service is cancelled.

This shows up in an INFO log as:

```
Timed out waiting for <...TxPool...> to finish its cleanup, forcibly cancelling pending tasks and exiting anyway
```

### How was it fixed?

Added a couple `self.wait()`s. Now that log doesn't show up anymore, and the shutdown is a bit faster.

Bonus changes:
- ~~Linter fix, as in #1357~~
- Add a warning when beam sync skips ahead. This warning doesn't show up when beam sync is continued from the last synced block.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1356.internal.rst
+++ b/newsfragments/1356.internal.rst
@@ -1,0 +1,2 @@
+Beam Sync now warns in the logs when it skips ahead of headers (relying on light-client-style
+verification of older headers).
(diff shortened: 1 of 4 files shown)


carver:
Is there any chance #1362 is related to this PR? It seems unlikely, but wanted to get more :eyes: on it from @cburgdorf before merging.

carver:
#1362 Seems very unlikely to be related to this, since #1354 has the same error in CI.

So I'll go ahead and merge this.

PR #1357, opened 2019-12-05
Mysterious spontaneous linter issues

carver:
I have no idea why these just showed up. The mypy version (and
mypy-extensions) didn't change.
It first showed up in a version bump that didn't change any code.

Surprises make me grumpy. (Also couldn't reproduce the errors anywhere
besides tox, mypy passes in local runs)

Luckily the fixes were straightforward, and I don't *disagree* with the
lines that they are wrong (except serialization_codec needing an
explicit type, which seems odd).

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


cburgdorf:
Merged this because it unbreaks CI for everyone.

PR #1311, opened 2019-11-21
Basic aggregation strategy

hwwhww:
### What was wrong?
Fix #1241 

### How was it fixed?
1. Implement basic aggregation strategy
    * Spec reference: https://github.com/ethereum/eth2.0-specs/blob/v09x/specs/networking/p2p-interface.md#global-topics and https://github.com/ethereum/eth2.0-specs/blob/v09x/specs/validator/0_beacon-chain-validator.md#attestation-aggregation
2. Bugfix
    * Attester should add the attestation they just attested to the attestation pool! Added `import_attestation_fn` to `Validator` for importing the self-attested attestation.
3. Note: we didn't have the concept of "subnets" before this PR. This PR just simply subscribe **all** subnets (`ATTESTATION_SUBNET_COUNT = 64`) when bootstrapping the node. The real subnet logic will not be added to this PR.
4. Note: now the Trinity validators won't create `beacon_attestation` topic messages, but they can still process it. (`beacon_attestation` is still on the spec for interop).

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Clean up commit history
- [x] Add tests
- [x] Refactor and clean up `validator.py`

[//]: # (For important changes that should go into the release notes please add a newsfragment file as explained here: https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)

[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)
- [ ] Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)

#### Cute Animal Picture

grey-seal-1969517_640

PR #1359, opened 2019-12-05
Update Dockerfile

hwwhww:
### What was wrong?
1. eth2 dockerfile is broken.
2. Try to make multiple dockerfiles more organized.

### How was it fixed?
1. Move Dockerfiles under `./docker` folder
2. Rename `Dockerfile.eth` to `beacon.Dockerfile`. Reasons:
    - IMO it's better to make different software (eth1 trinity, beacon-node, valdiator-client) in different Dockerfiles, and use docker-compose to set them up together.
    - Use `.Dockerfile` file name for GitHub/editor highlight. ✨
3. Fix `beacon.Dockerfile` and add CI test.

The following command works for me to:

```sh
docker build -t="ethereum/trinity-beacon" -f ./docker/beacon.Dockerfile ./docker/
docker run ethereum/trinity-beacon -l=DEBUG interop --validators=0,1,2 --start-delay=10 --wipedb
```
p.s. I believe that a single beacon node can't finalize block because it doesn't import the attestaton to the pool. The [fix](https://github.com/ethereum/trinity/blob/8143ff74158aa396a6e8f25fb7cfcfdc7cecbdcb/trinity/components/eth2/beacon/validator.py#L437) is in #1311.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

mammoth-3599754_640


docker/beacon.Dockerfile
```diff
@@ -1,19 +1,18 @@
...
 # Install deps
 RUN apt-get update
 RUN apt-get -y install libsnappy-dev gcc g++ cmake
 
-ARG GITREF=interop
-
 RUN git clone https://github.com/ethereum/trinity.git .
-RUN git checkout $GITREF
-RUN pip install -e .[dev] --no-cache-dir
+RUN pip install -e .[eth2-dev] --no-cache-dir
```
ChihChengLiang: We might want milagro-bls-binding in 'eth2-extra'
```suggestion
RUN pip install -e .[eth2-dev,eth2-extra] --no-cache-dir
```

docker/beacon.Dockerfile
```diff
@@ -1,19 +1,18 @@
...
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
 # Install deps
 RUN apt-get update
 RUN apt-get -y install libsnappy-dev gcc g++ cmake
 
-ARG GITREF=interop
-
 RUN git clone https://github.com/ethereum/trinity.git .
```
ChihChengLiang: Note that this will build the master branch Trinity, instead of PR branch of Trinity. Not sure if this is what we want.

If we want to build the PR branch, we should use something like `COPY . trinity`

hwwhww: I read some docker best practice articles earlier. An argument is that `git clone` gives us some flexibility at the development stage, and  `COPY` is more common in production.

I think at the development stage, using `git clone` makes it a more isolated environment, which is good in some way.

cburgdorf: >If we want to build the PR branch, we should use something

That's an interesting point though. The Eth1 Dockerfile uses `COPY` and I actually wanted to change it to use `git clone` but your point makes me re-think that. We had a few times where the docker build exposed a general problem with Trinity (usually dependency issues) and as you say, if you use `git clone` then the CI run ends up always using the source from `master` which sounds like a weakness.

Another possible idea would be to pre-process the `Dockerfile` to replace the `$GITREF` like we do with the `DappNode` build.

https://github.com/ethereum/trinity/blob/0e3e67d2ceb52e124fcb866f11d3f9704cfe04bf/Makefile#L89-L92

docker/beacon.Dockerfile
```diff
@@ -1,19 +1,18 @@
...
-RUN git checkout $GITREF
-RUN pip install -e .[dev] --no-cache-dir
+RUN pip install -e .[eth2-dev,eth2-extra] --no-cache-dir
 RUN pip install -U trinity --no-cache-dir
 
+RUN echo "Type \`trinity-beacon\` to boot or \`trinity-beacon --help\` for an overview of commands"
+
 EXPOSE 30303 30303/udp
 # Trinity shutdowns aren't yet solid enough to avoid the fix-unclean-shutdown
-ENTRYPOINT trinity $EXTRA_OPTS fix-unclean-shutdown && trinity-beacon $EXTRA_OPTS
```
cburgdorf: Just pointing out that this removes the `fix-unclean-shutdown` but keeps the comment above that explains it.

docker/beacon.Dockerfile
```diff
@@ -1,19 +1,17 @@
...
 RUN apt-get update
 RUN apt-get -y install libsnappy-dev gcc g++ cmake
 
-ARG GITREF=interop
-
 RUN git clone https://github.com/ethereum/trinity.git .
-RUN git checkout $GITREF
-RUN pip install -e .[dev] --no-cache-dir
+RUN pip install -e .[eth2-dev] --no-cache-dir
 RUN pip install -U trinity --no-cache-dir
```
ChihChengLiang: Don't think we need this line. I think we'll install a pypi version trinity here instead of the master branch trinity.
```suggestion
```

PR #1361, opened 2019-12-05
Combine DiscoveryProtocol and DiscoveryService

gsalgado:
In #1325 I'd left a FIXME behind as we were accessing async_service's internal _task_nursery. This addresses that by combining DiscoveryProtocol and DiscoveryService. Another advantage of doing this is that callsites now need less boilerplate code as they no longer need to create a DiscoveryProtocol

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -25,7 +25,6 @@ from typing import (
 )
 
 import trio
-import trio_typing
 
 import eth_utils.toolz
 from eth_utils import (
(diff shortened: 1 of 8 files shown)


p2p/discovery.py
```diff
@@ -120,31 +119,83 @@ def __repr__(self) -> str:
...
+        self.socket = socket
+
+    async def consume_datagrams(self) -> None:
+        while self.manager.is_running:
+            await self.consume_datagram()
+
+    async def handle_get_peer_candidates_requests(self) -> None:
+        async for event in self._event_bus.stream(PeerCandidatesRequest):
+
+            self.manager.run_task(self.maybe_lookup_random_node)
```
pipermerriam: This strikes me as something that should just run in it's own loop similar to the `consume_datagram` above.  Is there a reason to embed it here beyond that being what was historically done?

p2p/discovery.py
```diff
@@ -574,25 +617,26 @@ def process_ping(self, remote: NodeAPI, hash_: Hash32) -> None:
             callback()
 
 
-class PreferredNodeDiscoveryProtocol(DiscoveryProtocol):
+class PreferredNodeDiscoveryService(DiscoveryService):
```
pipermerriam: Not for this PR but I'd like it if we could decouple the "preferred node" concept from discovery.  It would simplify the discovery code and work fine in isolation.  Take a look at the "PeerBackends" concept in the peer pool as I think that would be the way to do it.

p2p/discovery.py
gsalgado: I can't think of a reason not to do what you suggest, really, but since that's changing behaviour and this is just a refactoring I'm going to open a separate PR for it

p2p/discovery.py
gsalgado: I'll keep that in mind!

PR #1363, opened 2019-12-05
Migrate component isolation away from multiprocessing

pipermerriam:
### What was wrong?

Multiprocessing isn't async friendly.

### How was it fixed?

Convert `AsyncioIsolatedComponent` to use `asyncio-run-in-process` for process isolation.

#### Cute Animal Picture

Adorable two-week-old seal pups show playful side under watchful eye of mother 3

### DIFF
--- a/tests/core/components/test_isolated_component.py
+++ b/tests/core/components/test_isolated_component.py
@@ -2,15 +2,9 @@ from argparse import Namespace
 import asyncio
 import logging
 import os
-from pathlib import Path
-import tempfile
 
 from async_service import background_asyncio_service
 import pytest
-
-from lahja import EndpointAPI, BaseEvent
-
-from p2p.service import BaseService, run_service
 
 from trinity._utils.chains import (
     get_local_data_dir,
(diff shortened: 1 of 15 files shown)


pipermerriam:
Depends on https://github.com/ethereum/asyncio-run-in-process/pull/5 because `uvloop + pytest + async_run_in_process` aren't getting along due to `uvloop` not liking pytest's mocked out stdout/err/in capturing which I wasn't able to get around locally.

pipermerriam:
currently blocked by https://github.com/ethereum/asyncio-run-in-process/pull/6

tests/core/components/test_isolated_component.py
```diff
@@ -99,7 +48,7 @@ def log_listener(trinity_config):
 
 @pytest.mark.asyncio
 async def test_asyncio_isolated_component(boot_info,
-                                          log_listener):
+                                          log_listener, event_loop):
```
pipermerriam: looks like I can remove the `event_loop`

trinity/extensibility/asyncio.py
```diff
@@ -1,79 +1,67 @@
...
 
+import logging
+logger = logging.getLogger('trinity')
+
 
 class AsyncioIsolatedComponent(BaseIsolatedComponent):
-    async def run(self) -> None:
-        """
-        Call chain is:
+    def get_subprocess_kwargs(self) -> Optional[SubprocessKwargs]:
```
pipermerriam: This API is really only currently used for testing but it's needed to get around the unfortunate behavior of `asyncio.open_subprocess_exc + pytest + uvloop`.  Pytest replaces `sys.stdin/out/err` with versions that don't have a `fileno` attribute which `uvloop` gets very unhappy about when you try to open a subprocess.

trinity/extensibility/asyncio.py
```diff
@@ -1,79 +1,67 @@
...
-        loop.add_signal_handler(signal.SIGINT, task.cancel)
-        loop.add_signal_handler(signal.SIGTERM, task.cancel)
-        try:
-            loop.run_until_complete(task)
-        except asyncio.CancelledError:
-            pass
-        finally:
-            loop.close()
+                await proc.wait()
+            except asyncio.CancelledError as err:
```
pipermerriam: realizing this whole `try/except` can go away as this behavior is now actually native to `asyncio-run-in-process`

trinity/extensibility/component_manager.py
```diff
@@ -52,61 +54,62 @@ def __init__(self,
...
+                        '/'.join(component.name for component in enabled_components),
+                    )
+                    # Concurrently start the components.
+                    await asyncio.gather(*(
+                        stack.enter_async_context(run_component(component))
+                        for component in enabled_components
+                    ))
+                    self.logger.info("Components started")
+                    try:
+                        await self._trigger_component_exit.wait()
```
pipermerriam: This mechanism is still not ideal or bulletproof but its working for now.  I intend to continue iterating on this.

trinity/sync/beam/chain.py
```diff
@@ -191,7 +191,7 @@ def __init__(
                 "Timed out while trying to fulfill prerequisites of "
                 f"sync launch strategy: {exc} from {self._launch_strategy}"
             )
-            await self.cancel()
+            self.cancel_nowait()
```
pipermerriam: I'm not sure how much this matters but the way that `p2p.service.BaseService` can end up *waiting* for things to shutdown seemed to be causing problems.

trinity/extensibility/asyncio.py
```diff
@@ -1,79 +1,68 @@
...
+                proc.send_signal(signal.SIGINT)
+                try:
+                    await asyncio.wait_for(proc.wait(), timeout=2)
+                except asyncio.TimeoutError:
+                    pass
```
carver: Can we log that the process didn't shut down cleanly/quickly here? Or is it too tricky b/c the logger process might be dead.

pipermerriam: No, we should be able to log here, I'll add it.

trinity/extensibility/asyncio.py
```diff
@@ -1,79 +1,68 @@
...
+logger = logging.getLogger('trinity')
+
 
 class AsyncioIsolatedComponent(BaseIsolatedComponent):
-    async def run(self) -> None:
-        """
-        Call chain is:
+    def get_subprocess_kwargs(self) -> Optional[SubprocessKwargs]:
+        # Note that this method currently only exist to facilitate testing.
+        return None
```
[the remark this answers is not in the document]
pipermerriam: No, the unset case isn't problematic.  `None` is the default for this argument so it's fine to have the `Optional[..]` typing.

PR #1370, opened 2019-12-09
Discovery now refreshes buckets periodically

gsalgado:
We used to only refresh buckets when there was a peer candidate request,
but now we use a daemon task to refresh them periodically

As suggested here: https://github.com/ethereum/trinity/pull/1361#discussion_r355033899

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -120,8 +120,7 @@ CMD_ID_MAP = dict((cmd.id, cmd) for cmd in [CMD_PING, CMD_PONG, CMD_FIND_NODE, C
 
 
 class DiscoveryService(Service):
-    _last_lookup: float = 0
-    _lookup_interval: int = 30
+    _refresh_interval: int = 30
     _max_neighbours_per_packet_cache = None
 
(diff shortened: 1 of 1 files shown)


p2p/discovery.py
```diff
@@ -180,22 +174,16 @@ def __init__(self,
...
-        elif self._lookup_running.locked():
-            self.logger.debug("Node discovery lookup already in progress, not running another")
-            return
-        async with self._lookup_running:
-            try:
-                await self.lookup_random()
-            finally:
-                self._last_lookup = time.time()
+    async def periodically_refresh(self) -> None:
+        while self.manager.is_running:
```
pipermerriam: You can do this as:

```python
from p2p.trio_utils import every

for _ in every(self._refresh_interval):
    await self.lookup_random()
```

gsalgado: Nice, hadn't noticed that helper before. I guess I don't need the `while self.is_running:` either, as trio should take care of cancelling the task when the service is cancelled, right?

pipermerriam: Correct, we don't actually rely on this mechanism for the task to exit cleanly, especially in trio-land

PR #1372, opened 2019-12-09
Do not skip tests for IstanbulVM

cburgdorf:
### What was wrong?

We aren't testing Trinity against Istanbul yet.

### How was it fixed?

- This builds upon #1403 
- Wire tests up in tox/circle
- Remove skipping of Istanbul tests

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,7 @@ envlist=
     py{36,37}-{eth1-core,p2p,p2p-trio,eth1-monitor-trio,integration,lightchain_integration,eth2-core,eth2-fixtures,eth2-integration,eth1-components,eth2-utils}
     py36-long_run_integration
     py36-rpc-blockchain
-    py36-rpc-state-{frontier,homestead,tangerine_whistle,spurious_dragon,byzantium,constantinople,petersburg}
+    py36-rpc-state-{frontier,homestead,tangerine_whistle,spurious_dragon,byzantium,constantinople,petersburg,istanbul}
     py37-rpc-state-{quadratic,sstore,zero_knowledge}
     py37-{libp2p,eth2-components}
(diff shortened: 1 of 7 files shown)


veox:
`randomStatetest94` is OK to be tagged `SLOW`, IMO:

https://github.com/ethereum/py-evm/pull/1852#issuecomment-533989714

(It is slow, true, but it crashes the CI box because it consumes a lot of memory.)

cburgdorf:
@veox Thanks for chiming in!

>In fact, I wonder if `SLOWEST_TESTS` should just be imported from py-evm

Generally a good idea I'd say but I'm not sure if it's possible as is because the tests aren't real modules (as in, they don't have a `__init__.py` file) so we can't import from them and if I recall correctly there is some obscure pytest reason for that.

veox:
> ... so we can't import from them ...

Argh, I keep forgetting.

tox.ini
```diff
@@ -3,7 +3,7 @@ envlist=
     py{36,37}-{eth1-core,p2p,p2p-trio,eth1-monitor-trio,integration,lightchain_integration,eth2-core,eth2-fixtures,eth2-integration,eth1-components,eth2-utils}
     py36-long_run_integration
     py36-rpc-blockchain
-    py36-rpc-state-{frontier,homestead,tangerine_whistle,spurious_dragon,byzantium,constantinople,petersburg}
+    py36-rpc-state-{frontier,homestead,tangerine_whistle,spurious_dragon,byzantium,constantinople,petersburg, istanbul}
```
veox: Nitpick: extra space.

cburgdorf: :+1: fixed

PR #1378, opened 2019-12-10
Listen to 0.0.0.0 as default

ChihChengLiang:
### What was wrong?

Trinity can't listen to a remote node if we set 127.0.0.1.

Not sure if we really want it to be configurable. Correct me if I'm wrong, but in contrast to RPC functionality that we want to restrict access for only certain ip addresses, normally we want the p2p module listens to the world.

#### Cute Animal Picture

🐛

### DIFF
--- a/trinity/components/eth2/beacon/component.py
+++ b/trinity/components/eth2/beacon/component.py
@@ -106,7 +106,7 @@ class BeaconNodeComponent(AsyncioIsolatedComponent):
         # TODO: Handle `bootstrap_nodes`.
         libp2p_node = Node(
             key_pair=key_pair,
-            listen_ip="127.0.0.1",  # FIXME: Should be configurable
+            listen_ip="0.0.0.0",
             listen_port=boot_info.args.port,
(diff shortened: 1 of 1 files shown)


trinity/components/eth2/beacon/component.py
```diff
@@ -106,7 +106,7 @@ def _load_or_create_node_key(cls, boot_info: BootInfo) -> KeyPair:
         # TODO: Handle `bootstrap_nodes`.
         libp2p_node = Node(
             key_pair=key_pair,
-            listen_ip="127.0.0.1",  # FIXME: Should be configurable
+            listen_ip="0.0.0.0",  # FIXME: Should be configurable
```
cburgdorf: Just noting that the Eth1 part of Trinity does also listen on `0.0.0.0` and I also can't think of a reason why the beacon node would be different here.

ChihChengLiang: Oh, thanks for that information. Then we shouldn't bother with configuring it.

Issue #1362, opened 2019-12-05
test_web3_commands fails (flaky, I think), due to missing "EventBus Endpoint bjson-rpc-api connecting to other Endpoints"


carver:
* Trinity version: master
* OS: CircleCI

### What is wrong?

(Perhaps related to #1213 )

Looks like a flaky test, in py36-wheel-cli:

```
=================================== FAILURES ===================================
_ test_web3_commands_via_attached_console[command2-5-0x065fd78e53dcef113bf9d7732dac7c5132dcf85c9588a454d832722ceb097422] _

command = ('trinity', '--genesis=/home/circleci/repo/trinity/assets/eip1085/devnet.json', '--data-dir=/tmp/tmpsfoh77on/trinity/devnet', '--network-id=5', '--log-level=DEBUG', '--sync-mode=none', ...)
expected_network_id = 5
expected_genesis_hash = '0x065fd78e53dcef113bf9d7732dac7c5132dcf85c9588a454d832722ceb097422'
xdg_trinity_root = PosixPath('/tmp/tmpsfoh77on/trinity')
unused_tcp_port = 58073

    @pytest.mark.parametrize(
        'command, expected_network_id, expected_genesis_hash',
        (
            (('trinity', '--log-level=DEBUG'), 1, MAINNET_GENESIS_HASH),
            (('trinity', '--ropsten', '--log-level=DEBUG'), 3, ROPSTEN_GENESIS_HASH),
            (
                (
                    'trinity',
                    f'--genesis={ASSETS_DIR}/eip1085/devnet.json',
                    # We don't have a way to refer to the tmp xdg_trinity_root here so we
                    # make up this replacement marker
                    '--data-dir={trinity_root_path}/devnet',
                    '--network-id=5',
                    '--log-level=DEBUG',
                ), 5, '0x065fd78e53dcef113bf9d7732dac7c5132dcf85c9588a454d832722ceb097422'),
        )
    )
    @pytest.mark.asyncio
    async def test_web3_commands_via_attached_console(command,
                                                      expected_network_id,
                                                      expected_genesis_hash,
                                                      xdg_trinity_root,
                                                      unused_tcp_port):
    
        command = tuple(
            fragment.replace('{trinity_root_path}', str(xdg_trinity_root))
            for fragment
            in command
        )
        # The test mostly fails because the JSON-RPC requests time out. We slim down
        # services to make the application less busy and improve the overall answer rate.
[...]

PR #1382, opened 2019-12-11
Fix flaky test

cburgdorf:
### What was wrong?

We currently wait for the json-rpc Endpoint to connect
to other endpoints before we do certain assertions.
But sometimes we might end up in a situation where
al other endpoints have already connected to this
endpoint and so it does not actively connect to
others and the test fails.

### How was it fixed?

It is safer to simply wait for the announcement of the 
JSON-RPC endpoint instead.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1382.internal.rst
+++ b/newsfragments/1382.internal.rst
@@ -1,0 +1,9 @@
+Fix flaky interactive web3 console test
+
+Due to the way components cross connect to each
+other, not all logs of that startup routine are
+happening deterministically. We were waiting on
+a log that would sometimes never show up for all
+valid reason. This test is now based on a log that
+is guaranteed to show up unless a serious bug
+happened.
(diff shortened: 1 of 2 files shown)


cburgdorf:
@carver I think this fixes #1362

carver:
Maybe a release note in the `internal` section? Having less flaky tests is very satisfying to other devs :)

pipermerriam:
:+1: to ship this.  FYI, I'm very close to being ready to implement component dependencies which will make component startup order more deterministic for things like this and hopefully easier to read about.

cburgdorf:
>Maybe a release note in the `internal` section? Having less flaky tests is very satisfying to other devs :)

Good point. I keep forgetting we have `internal`! Added.

PR #1383, opened 2019-12-11
Update web3 to v5.4.0 (latest ATM)

veox:
### What was wrong?

A clean-room install of `[.dev]` produces a `pip` error:

``` sh
virtualenv -p python3.7 .virtualenv/`basename $(pwd)`
source .virtualenv/`basename $(pwd)`/bin/activate
pip install -e .\[dev\]
```

    ERROR: eth-tester 0.2.0b3 has requirement eth-abi<3.0.0,>=2.0.0b4, but you'll have eth-abi 1.3.0 which is incompatible.

### How was it fixed?

The older `eth-abi` is pulled in via `web3` v4. To resolve this, v5 is installed.

`websockets` (AFAIK only used in the `ethstats_client` component) got bumped, too: `web3` wants `>=8.1.0,<9`. Specifying `>5.0.1` would have worked, but installs v8.1.0 anyway.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

Source: ["Kukutis" by Irmantas Š. at efoto.lt](https://www.efoto.lt/node/1262181)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,11 +39,11 @@ deps = {
         "ipython>=7.8.0,<7.10.0",  # attach fails with v7.10.{0,1}
         "plyvel==1.1.0",
         PYEVM_DEPENDENCY,
-        "web3==4.4.1",
+        "web3==5.4.0",
         "lahja>=0.15.2,<0.16",
         "termcolor>=1.1.0,<2.0.0",
         "uvloop==0.14.0;platform_system=='Linux' or platform_system=='Darwin' or platform_system=='FreeBSD'",  # noqa: E501
-        "websockets==5.0.1",
+        "websockets>=8.1.0",
         "jsonschema==3.0.1",
         "mypy-extensions>=0.4.3,<0.5.0",
         "typing_extensions>=3.7.2,<4.0.0",
--- a/trinity/components/eth2/eth1_monitor/eth1_monitor.py
+++ b/trinity/components/eth2/eth1_monitor/eth1_monitor.py
@@ -326,7 +326,7 @@ class Eth1Monitor(Service):
         self, logs: Sequence[DepositLog], block_number: BlockNumber
     ) -> None:
         """
-        Simply store the deposit data from the log, and increase the corresponding block's
+        Store deposit data from the log in database, and increase the corresponding block's
         `deposit_count`.
         """
         seq_deposit_data = tuple(
--- a/newsfragments/1383.feature.rst
+++ b/newsfragments/1383.feature.rst
@@ -1,0 +1,1 @@
+Underlying ``web3`` module changed from v4 to v5, including in the console.
--- a/trinity/components/builtin/ethstats/ethstats_client.py
+++ b/trinity/components/builtin/ethstats/ethstats_client.py
@@ -1,7 +1,12 @@
 import asyncio
 import datetime
 import json
-import typing
+
+from typing import (
+    Any,
+    Dict,
+    NamedTuple,
+)
 
 import websockets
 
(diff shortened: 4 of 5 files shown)


setup.py
```diff
@@ -37,19 +37,20 @@
...
         "jsonschema==3.0.1",
         "mypy-extensions>=0.4.3,<0.5.0",
         "typing_extensions>=3.7.2,<4.0.0",
         "ruamel.yaml==0.15.98",
         "argcomplete>=1.10.0,<2",
         "multiaddr>=0.0.8,<0.1.0",
         "pymultihash>=0.8.2",
-        "libp2p==0.1.1",
+        "libp2p==0.1.1",  # try unpinning 'web3' above if you unpin this
+        "protobuf==3.9.0",  # not used directly, strictly due to libp2p/web3 conflict
```
veox: To clarify, I find this unacceptable; `trinity` is not using `protobuf` directly, so this will only exacerbate the problem.

I don't believe there'll come a time when we'll be able to remove this pinning, unless `pip` starts detecting `protobuf==3.9.0` as a viable solution on its own; or `libp2p` stops this "hard pinning" practice.

veox:
Also, consider this blocked by PR #1288.

setup.py
carver: A new libp2p is released

veox: Cool, that fixed the issue. Pinning `protobuf` no longer required.

trinity/components/eth2/eth1_monitor/eth1_monitor.py
```diff
@@ -341,7 +338,7 @@ def _process_logs(
             DepositData(
                 pubkey=log.pubkey,
                 withdrawal_credentials=log.withdrawal_credentials,
-                amount=log.amount,
+                amount=Web3.toInt(log.amount),  # FIXME: wrong format/amount?..
```
veox: Some local debug-printing shows the amount is already incorrect when being parsed in `_get_logs_from_block()` above.

veox: Yup! `log.amount` at this point has fiddled endianness.

Say, the actual deposit (`tx.value`) was `6471247182000000000`; the deposit contract converts that to shannon (gwei) first, so the value emitted in event is `6471247182`.

``` python
In [57]: w3.toHex(6471247182)                                                                                                                                  
Out[57]: '0x181b7654e'
```

But the data in the parsed log (passed to this function) is

``` python
In [53]: w3.toHex(b'Ne\xb7\x81\x01\x00\x00\x00')                                                                                                               
Out[53]: '0x4e65b78101000000'
```

`0x181b7654e` is the endian-opposite of `0x4e65b781`.

Now, where this happens is still a mystery to me.

Pinging @mhchia as file author.

veox: Ah! Perhaps I should still try to extract the event data properly, instead of trying to handle the raw log data.

trinity/components/eth2/eth1_monitor/eth1_monitor.py
```diff
@@ -334,14 +331,14 @@ def _process_logs(
...
-        Simply store the deposit data from the log, and increase the corresponding block's
+        Store deposit data from the log in database, and increase the corresponding block's
         `deposit_count`.
         """
         seq_deposit_data = tuple(
             DepositData(
                 pubkey=log.pubkey,
                 withdrawal_credentials=log.withdrawal_credentials,
-                amount=log.amount,
+                amount=Web3.toInt(log.amount[::-1]),  # reverse endianness, convert bytes to int
```
veox: There doesn't seem to be a `little_endian_to_int(...)` in `eth-utils`:

https://eth-utils.readthedocs.io/en/latest/utilities.html#encoding-utils

veox: There's an unused `DepositLog.from_contract_log_dict()` in the file; perhaps it should be used to avoid the cryptic `[::-1]` here. That would also fix a linting error.

veox:
Removed `[DEMO]` from title - I think this ended up better than demo-quality.

Issue #1381, opened 2019-12-10
Getting BrokenPipeError when trying to send logs on shutdown


carver:
### What is wrong?

Getting an absolute deluge of logs (about broken logs) on shutdown. It requires multiple `ctrl-c` before it finally stops. This is pretty recent.

One of: #1307 

### How can it be fixed

I guess the logging pipe is being closed before some other processes have finished shutting down?

pipermerriam:
I thought I had this clean but this was introduced in #1328 

I am pretty sure I have it fixed in #1363 which I think I'm almost done with.

carver:
Hmm, seems like it's still happening. Were you able to reproduce the failure and then fix locally?

pipermerriam:
I'm still narrowing in on this stuff.  I don't have anything concrete but the problem should be *better* now than it was before, and now I can narrow in on why the logging connection is closed even though it's still within the context block.

My two ideas right now are:

1. The parent process has actually already closed the logger server. (less likely)
2. There's still some asyncio loop stuff happening when it should already be shut down...   

Neither of these make any logical sense to me based on how the code *should* run, but I feel like I'm starting to gain some better intuition about how everything runs after throwing myself at it for a few weeks.  I continue to *feel* like the new service model is going to clean a lot of this up so I'm starting to chip away at that too...

cburgdorf:
I just came here to note that I see exactly the same flood of errors on shutdown.

gsalgado:
Should this be closed in favour of #1558?

carver:
> Should this be closed in favour of #1558?

Maybe the other way around, since this one has more commentary. Worth briefly noting that #1558 had a slightly different message, but probably the same cause:
```
--- Logging error ---
Traceback (most recent call last):
  File "/home/jcarver/code/trinity/trinity/db/manager.py", line 150, in serve_conn
    operation_byte = sock.read_exactly(1)
  File "/home/jcarver/code/trinity/trinity/_utils/socket.py", line 27, in read_exactly
    raise OSError("Connection closed")
OSError: Connection closed

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jcarver/code/trinity/trinity/_utils/logging.py", line 76, in emit
    self._socket.sendall(msg_length_data + msg_data)
OSError: [Errno 9] Bad file descriptor
```

(Bad file descriptor instead of `BrokenPipeError`)

gsalgado:
I'm no longer seeing this, so closing for now

PR #1384, opened 2019-12-11
Better handling for child process logging

pipermerriam:
fixes https://github.com/ethereum/trinity/issues/1381

### What was wrong?

The *child* processes hook into the main process over an IPC socket to pass logs through.

First, we weren't closing this IPC socket which resulted in lots of errors due to unclosed sockets.

Second, once the socket is closed, the handler was still left on the logger which meant that other things that might use logging, like things in the standard library, which do things during the python shutdown, would cause the handler to throw errors since it wouldn't be able to communicate with the parent process.

### How was it fixed?

Used two context managers, one to close the socket, and another to remove the handler.

#### Cute Animal Picture

IMG_0164-1024x768

### DIFF
--- a/trinity/_utils/logging.py
+++ b/trinity/_utils/logging.py
@@ -1,3 +1,4 @@
+import contextlib
 import copy
 import logging
 from logging import (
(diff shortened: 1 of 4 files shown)


carver:
Hah, looks like you just got hit by #1362 (#1382 is approved to fix it)

PR #1385, opened 2019-12-11
Ensure database socket connections are cleaned up

pipermerriam:
### What was wrong?

We use a socket to connect to the base database.  This socket was not being closed anywhere which resulted in a number of warnings during shutdown.

### How was it fixed?

Convert to use context management API to ensure that the database gets closed down as expected.

#### Cute Animal Picture

1411051378250_wps_4_A_baby_seal_cautiously_di

### DIFF
--- a/trinity/components/builtin/ethstats/ethstats_service.py
+++ b/trinity/components/builtin/ethstats/ethstats_service.py
@@ -1,5 +1,7 @@
 import asyncio
+import contextlib
 import platform
+from typing import Iterator
 
 import websockets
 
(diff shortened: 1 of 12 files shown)


trinity/components/builtin/ethstats/ethstats_service.py
```diff
@@ -61,26 +63,27 @@ def __init__(
...
-                self.run_daemon_task(self.server_handler(client))
-                self.run_daemon_task(self.statistics_handler(client))
-
-                await client.run()
-                if self.is_operational and not client.is_operational:
-                    self.logger.info('Connection to %s closed', self.server_url)
-                    self.logger.info('Reconnecting in 5s...')
-                    await self.sleep(5)
+        with self.get_chain() as chain:
+            self.chain = chain
```
carver: Yup, this example is safer, IMO.

pipermerriam:
I'm moving forward with this as-is.  I generally agree on your points about this being *able* to be used incorrectly but I'm not overly concerned about it right now and it's something I can cleanup in a subsequent sweep.

carver:
> I'm moving forward with this as-is. I generally agree on your points about this being _able_ to be used incorrectly but I'm not overly concerned about it right now and it's something I can cleanup in a subsequent sweep.

Mind adding a tracking issue?

PR #1386, opened 2019-12-11
Convert the ethstats service to use the new async-service API

pipermerriam:
### What was wrong?

The new `async-service` library is not being used everywhere.

### How was it fixed?

Converted the services in the `ethstats` component to use the new `async-service` APIs.

#### Cute Animal Picture

Adorable two-week-old seal pups show playful side under watchful eye of mother 4

### DIFF
--- a/trinity/components/builtin/ethstats/ethstats_client.py
+++ b/trinity/components/builtin/ethstats/ethstats_client.py
@@ -8,15 +8,9 @@ from typing import (
     NamedTuple,
 )
 
+from async_service import Service
+from eth_utils import get_extended_debug_logger
 import websockets
-
-from cancel_token import (
-    CancelToken,
-)
-
-from p2p.service import (
-    BaseService,
-)
 
 
 # Returns UTC timestamp in ms, used for latency calculation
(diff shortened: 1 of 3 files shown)


trinity/components/builtin/ethstats/ethstats_service.py
```diff
@@ -63,28 +61,30 @@ def __init__(
...
+                await client_manager.wait_finished()
 
-                await client.run()
-                if self.is_operational and not client.is_operational:
-                    self.logger.info('Connection to %s closed', self.server_url)
-                    self.logger.info('Reconnecting in 5s...')
-                    await self.sleep(5)
+                self.logger.info('Connection to %s closed', self.server_url)
+                self.logger.info('Reconnecting in 5s...')
+                await asyncio.sleep(5)
```
cburgdorf: I think this isn't well covered by tests. Did you manually test if this still recovers from a failed connection as it should?

trinity/components/builtin/ethstats/ethstats_client.py
```diff
@@ -31,35 +25,35 @@ class EthstatsException(Exception):
...
     # Get messages from websocket, deserialize them and put into queue
     async def recv_handler(self) -> None:
-        while self.is_operational:
+        while self.manager.is_running:
             try:
                 json_string: str = await self.websocket.recv()
             except websockets.ConnectionClosed as e:
                 self.logger.debug2("Connection closed: %s", e)
-                await self.cancel()
+                await self.manager.stop()
```
gsalgado: Do we really need to wait for manager.stop() here or can we just call manager.cancel() and forget about it?

gsalgado: Oh, maybe you do that because this is a daemon task and you don't want it to return before run()? If that's the case, I wonder if we should have a comment here? Also, it'd be nice if the docs made it clear that daemon tasks should probably always await for manager.stop() instead of manager.cancel()?

trinity/components/builtin/ethstats/ethstats_service.py
pipermerriam: I didn't...  I'm going to hold on merging this for now because I actually have some issues with how these two services are architected that's bugging me.  The fact that the outer service is responsible for driving the inner service feels wrong to me and I think makes this hard to reason about.

pipermerriam:
closing for now.  I want to do some minor cleanup to the underlying way these two services are architected to remove the awkwardness of the outer service needing to drive the client.

PR #1387, opened 2019-12-11
Update transaction pool component to use async-service APIs

pipermerriam:
### What was wrong?

The new `async-service` library is not being used everywhere.

### How was it fixed?

Converted the services in the `txpool` component to use the new `async-service` APIs.

#### Cute Animal Picture

PuppyDay_AntarcticFurSealPups_ChristopherWillis

### DIFF
--- a/trinity/components/builtin/tx_pool/pool.py
+++ b/trinity/components/builtin/tx_pool/pool.py
@@ -8,15 +8,14 @@ from typing import (
 )
 import uuid
 
+from async_service import Service
 from lahja import EndpointAPI
 
-from cancel_token import CancelToken
-
+from eth_utils import get_extended_debug_logger
 from eth_utils.toolz import partition_all
 from eth.abc import SignedTransactionAPI
 
 from p2p.abc import SessionAPI
-from p2p.service import BaseService
 
 from trinity._utils.bloom import RollingBloom
 from trinity.protocol.eth.events import (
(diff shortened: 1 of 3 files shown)


trinity/components/builtin/tx_pool/pool.py
```diff
@@ -96,15 +95,15 @@ def __init__(self,
     # now.
     msg_queue_maxsize: int = 2000
 
-    async def _run(self) -> None:
+    async def run(self) -> None:
```
cburgdorf: A small detail, but I like that we now get rid of this mistakenly private `_run` method.

trinity/components/builtin/tx_pool/pool.py
```diff
@@ -96,15 +95,15 @@ def __init__(self,
...
         self.logger.info("Running Tx Pool")
 
         # background process which aggregates transactions and relays them to
         # our other peers.
-        self.run_daemon_task(self._process_transactions())
+        self.manager.run_daemon_task(self._process_transactions)
 
-        async for event in self.wait_iter(self._event_bus.stream(TransactionsEvent)):
-            self.run_task(self._handle_tx(event.session, event.command.payload))
+        async for event in self._event_bus.stream(TransactionsEvent):
```
cburgdorf: And no `wait_iter` anymore. I must confess that I still have to get my feet wet to fully grok how we are now transitioning from all these cancel token APIs into the brave new world. This actually gets me excited about adding docs to the async-service APIs as it will force me to familiarize myself more with it.

PR #1390, opened 2019-12-12
Convert request server to use async-service

pipermerriam:
### What was wrong?

The Request Server component isn't using the nice new async-service APIs

### How was it fixed?

Converted the services used by the Request Server component to use the new `async-service` APIs

#### Cute Animal Picture

o-FANCY-DRESS-CHRISTMAS-facebook

### DIFF
--- a/tests/core/integration_test_helpers.py
+++ b/tests/core/integration_test_helpers.py
@@ -7,6 +7,7 @@ from zipfile import ZipFile
 from async_generator import (
     asynccontextmanager,
 )
+from async_service import background_asyncio_service
 from cancel_token import OperationCancelled
 from eth_keys import keys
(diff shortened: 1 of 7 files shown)


tests/core/integration_test_helpers.py
```diff
@@ -124,29 +122,8 @@ def load_fixture_db(db_fixture, db_class=LevelDB):
...
-    await event_server.events.started.wait()
-    try:
+    async with background_asyncio_service(event_server):
         yield event_server
-    finally:
-        await event_server.cancel()
-
-
-@asynccontextmanager
-async def run_request_server(event_bus, chaindb, server_type=None):
```
cburgdorf: Another demonstration of how this greatly simplifies code!

PR #1391, opened 2019-12-12
Convert UPNP component to use async-service

pipermerriam:
### What was wrong?

The UPNP component isn't using the nice new `async-service` APIs

### How was it fixed?

Converted the service to use the new `async-service` APIs

#### Cute Animal Picture

Pets-in-Christmas-costumes

### DIFF
--- a/trinity/components/builtin/upnp/nat.py
+++ b/trinity/components/builtin/upnp/nat.py
@@ -9,15 +9,13 @@ from typing import (
 )
 from urllib.parse import urlparse
 
-from cancel_token import (
-    CancelToken,
-    OperationCancelled,
-)
+from async_service import Service
+from eth_utils import get_extended_debug_logger
+
 from p2p.exceptions import (
     NoInternalAddressMatchesDevice,
 )
 import netifaces
-from p2p.service import BaseService
 
 from eth._warnings import catch_and_ignore_import_warning
 with catch_and_ignore_import_warning():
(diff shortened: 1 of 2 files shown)


trinity/components/builtin/upnp/nat.py
```diff
@@ -52,36 +50,37 @@ def find_internal_ip_on_device_network(upnp_dev: upnpclient.upnp.Device) -> str:
...
             try:
                 await self.add_nat_portmap()
                 # Wait for the port mapping lifetime, and then try registering it again
-                await self.wait(asyncio.sleep(self._nat_portmap_lifetime))
-            except (OperationCancelled, asyncio.CancelledError):
-                break
+                await asyncio.sleep(self._nat_portmap_lifetime)
+            except asyncio.CancelledError:
+                # Re-raise to prevent it being logged as an unexpected exception.
+                raise
```
cburgdorf: Am I right to assume that we should have re-raised that under the old model, too?

trinity/components/builtin/upnp/nat.py
```diff
@@ -155,16 +154,17 @@ def current_mapping(self) -> PortMapping:
...
-            devices = await self.wait(
-                loop.run_in_executor(ThreadPoolExecutor(max_workers=1), upnpclient.discover),
-                timeout=UPNP_DISCOVER_TIMEOUT_SECONDS,
-            )
-        except asyncio.TimeoutError:
-            self.logger.info("Timeout waiting for UPNP-enabled devices")
-            return
-        else:
-            self.logger.debug("Found %d candidate NAT devices", len(devices))
+        with ThreadPoolExecutor() as executor:
```
cburgdorf: Cool, I've set a trend :sweat_smile:

trinity/components/builtin/upnp/nat.py
pipermerriam: Yeah, it's generally considered bad mojo to ignore/silence it.

PR #1394, opened 2019-12-13
Hashablification

jannikluhn:
This is the same as #1283. CI got confused with dependency versions, so the tests were failing. I hope that a new PR makes them forget everything and install everything from scratch.

Edit: Didn't work :(

```
pip._vendor.pkg_resources.ContextualVersionConflict: (pyrsistent 0.15.6 (/home/circleci/repo/.tox/py36-eth2-core/lib/python3.6/site-packages), Requirement.parse('pyrsistent==0.15.4'), {'ssz'})
```

So it for some reason has `pyrsistent==0.15.6` installed even though `py-ssz` requires `pyrsistent==0.15.4`.

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/state_machines/forks/serenity/epoch_processing.py
+++ b/eth2/beacon/state_machines/forks/serenity/epoch_processing.py
@@ -2,6 +2,7 @@ from typing import Sequence, Set, Tuple
 
 from eth_typing import Hash32
 from eth_utils.toolz import curry
+from ssz.hashable_list import HashableList
 
 from eth2._utils.tuple import update_tuple_item, update_tuple_item_with_fn
 from eth2.beacon.constants import BASE_REWARDS_PER_EPOCH, FAR_FUTURE_EPOCH
@@ -138,7 +139,9 @@ def _determine_new_justified_checkpoint_and_bitfield(
         new_current_justified_root = state.current_justified_checkpoint.root
 
     return (
-        Checkpoint(epoch=new_current_justified_epoch, root=new_current_justified_root),
+        Checkpoint.create(
+            epoch=new_current_justified_epoch, root=new_current_justified_root
+        ),
         justification_bits,
     )
 
@@ -210,7 +213,7 @@ def _determine_new_finalized_checkpoint(
     else:
         new_finalized_root = state.finalized_checkpoint.root
 
-    return Checkpoint(epoch=new_finalized_epoch, root=new_finalized_root)
+    return Checkpoint.create(epoch=new_finalized_epoch, root=new_finalized_root)
 
 
 def process_justification_and_finalization(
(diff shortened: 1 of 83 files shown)


eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -473,7 +483,7 @@ def _determine_next_eth1_votes(
     state: BeaconState, config: Eth2Config
 ) -> Tuple[Eth1Data, ...]:
     if (state.slot + 1) % config.SLOTS_PER_ETH1_VOTING_PERIOD == 0:
-        return tuple()
+        return HashableList.from_iterable((), state.eth1_data_votes.sedes)
```
hwwhww: Just to confirm, is `HashableList(TElement)` is compatible with `Tuple[Eth1Data, ...]` in typing? I thought `HashableList` is `Sequence[TElement]`...?

tests/eth2/core/beacon/test_validator_status_helpers.py
```diff
@@ -34,7 +34,7 @@ def test_activate_validator(
         assert some_validator.activation_epoch == config.GENESIS_EPOCH
     else:
         some_validator = create_mock_validator(
-            pubkeys[: validator_count + 1], config, is_active=is_already_activated
```
hwwhww: nice catch

eth2/beacon/types/aggregate_and_proof.py
```diff
@@ -1,29 +1,34 @@
...
-        self,
+    @classmethod
+    def create(
+        cls: Type[TAggregateAndProof],
         index: ValidatorIndex = default_validator_index,
         selection_proof: BLSSignature = EMPTY_SIGNATURE,
         aggregate: Attestation = default_attestation,
-    ) -> None:
-        super().__init__(
+    ) -> TAggregateAndProof:
```
hwwhww: I'm curious, for `mypy`, is `TypeVar` *required* or we can simply use `"AggregateAndProof"` instead?

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
jannikluhn: Ah, no, that's a mistake. I'm quite surprised that mypy didn't catch this. I'll fix the type hint.

eth2/beacon/types/aggregate_and_proof.py
jannikluhn: It would make a difference if we subclass `AggregateAndProof` and don't override the classmethod. Then mypy would think `AggregateAndProofSubclass.create` would return a `AggregateAndProof` which wouldn't be correct.

PR #1403, opened 2019-12-18
Make sure eth_getStorageAt pads result to 32 byte

cburgdorf:
### What was wrong?

The return value of `eth_getStorageAt` is expected to be padded to 32 byte.
References:

[Spec](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getstorageat)
[Geth Example](https://api.etherscan.io/api?module=proxy&action=eth_getStorageAt&address=0x6e03d9cce9d60f3e9f2597e13cd4c54c55330cfd&position=0x0&tag=latest&apikey=YourApiKeyToken)

### How was it fixed?

1. Apply padding to the values before returning them
2. Normalize the state tests as they are inconsistent

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/rpc/modules/eth.py
+++ b/trinity/rpc/modules/eth.py
@@ -43,6 +43,9 @@ from eth.rlp.headers import (
 from eth.vm.spoof import (
     SpoofTransaction,
 )
+from eth._utils.padding import (
+    pad32,
+)
 
 from trinity.chains.base import AsyncChainAPI
 from trinity.constants import (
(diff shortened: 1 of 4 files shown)


trinity/rpc/modules/eth.py
```diff
@@ -43,6 +43,9 @@
 from eth.vm.spoof import (
     SpoofTransaction,
 )
+from eth._utils.padding import (
+    pad32,
```
cburgdorf: @pipermerriam I've heard kittens die and the devil will haunt me if I import from `eth._utils`. Would you be open to have `pad32` be added to `eth_utils` to fix this?

cburgdorf:
Notice that this is part of getting us to activate the Istanbul tests (#1372) but it is self contained enough to have it's own PR.

cburgdorf:
This landed as part of #1372

PR #1404, opened 2019-12-18
Use newer `cimg` docker images for faster builds

eddiewebb:
Will also improve determinism as new images from CircleCI provide more stable versioning.

Uses shared base layers with other languages to hit cache more and reduce build spinup times.

### What was wrong?
Using older CircleCI images that are larger and less deterministic.

### How was it fixed?
Switching docker images to new `cimg` type - https://github.com/CircleCI-Public/cimg-python

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

Cheetahs are fast, there are many since this image will be better at hitting cached layers.

### DIFF
--- a/.circleci/build_geth.sh
+++ b/.circleci/build_geth.sh
@@ -7,12 +7,12 @@ export GETH_BINARY="$HOME/.py-geth/geth-$GETH_VERSION/bin/geth"
 if [ ! -e "$GETH_BINARY" ]; then
   curl -O https://storage.googleapis.com/golang/go1.10.linux-amd64.tar.gz
   tar xvf go1.10.linux-amd64.tar.gz
-  sudo chown -R root:root ./go
-  sudo mv go /usr/local
-  sudo ln -s /usr/local/go/bin/go /usr/local/bin/go
-  sudo apt-get update;
-  sudo apt-get install -y build-essential;
+  chown -R root:root ./go
+  mv go /usr/local
+  ln -s /usr/local/go/bin/go /usr/local/bin/go
+  apt-get update;
+  apt-get install -y build-essential;
   python -m geth.install $GETH_VERSION;
 fi
-sudo ln -s $GETH_BINARY /usr/local/bin/geth
+ln -s $GETH_BINARY /usr/local/bin/geth
 geth version
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,7 @@ common: &common
         name: run tox
         command: ~/.local/bin/tox
     - save_cache:
+        when: always
         paths:
           - .hypothesis
           - .tox
@@ -63,6 +64,7 @@ geth_steps: &geth_steps
         name: run tox
         command: ~/.local/bin/tox -r
     - save_cache:
+        when: always
         paths:
           - .tox
           - ~/.cache/pip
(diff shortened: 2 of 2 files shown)


cburgdorf:
Sounds great! Thanks for sending the PR. What do we do about the failed snappy install though?

eddiewebb:
The snappy issue was due to a change that we now cleanup apt repo and package listings as advised by docker best practices. Adding the `apt update` command before installing tools got past that.  We can consider pinning the snappy version in the config as well.

Now seeing some test failures on 2 jobs, which could be genuine or related to the fact we dont run as circleci but as root.

This means paths changes too, which for example impacts caching. (that shouldn't cause failures, just slowness unless something is off with our caching strategy)

```
ERROR: InvocationError for command /root/repo/.tox/py37-wheel-cli/bin/pytest tests/integration/ -k trinity_cli --log-cli-level 0 (exited with code 1)
```

Earliest cause of failure I see is

```
tests/integration/test_trinity_cli.py:384: AssertionError
___ test_block_export[1-Block number 1 does not exist in the database-False] ___

block_number = 1, info = 'Block number 1 does not exist in the database'
success = False

    @pytest.mark.parametrize(
        'block_number, info, success',
        (
            (0, "Successfully exported Block #0", True),
            (1, "Block number 1 does not exist in the database", False),
        )
    )
    @pytest.mark.asyncio
    async def test_block_export(block_number, info, success):
    
        with tempfile.TemporaryDirectory() as export_path:
    
            export_file = pathlib.Path(export_path) / 'export.rlp'
    
            assert not export_file.exists()
    
            command = ('trinity', 'export', str(export_file), str(block_number))
            async with AsyncProcessRunner.run(command, timeout_sec=40) as runner:
>               assert await contains_all(runner.stderr, {info})
E               assert False

block_number = 1
command    = ('trinity', 'export', '/tmp/tmpyd7jcugq/export.rlp', '1')
export_file = PosixPath('/tmp/tmpyd7jcugq/export.rlp')
export_path = '/tmp/tmpyd7jcugq'
info       = 'Block number 1 does not exist in the database'
runner     = <trinity.tools.async_process_runner.AsyncProcessRunner object at 0x7f3fae862ed0>
success    = False
```

cburgdorf:
The `py37-wheel-cli` job got a little flaky with one of the recent commits I think. I kicked that build to run again. But the `py36-lightchain_integration` seems to fail because it does install some Go software which fails now. My guess is that the previous image had Go support (in addition to Python) out of the box whereas the new ones are intentionally slimmer and don't have that. If there's no quick resolution for that we might leave that one on the old image for now?

eddiewebb:
`py36-lightchain_integration` is passing, removed `sudo` from build_geth.sh since we're running as root now.

I think we'll want to rev the cache keys as well, looking now.

cburgdorf:
>removed sudo from build_geth.sh since we're running as root now.

Oh yeah, I see! Thanks for doing this!

cburgdorf (approved):

eddiewebb:
I'd like to update cache and squash all these commits before merging anything

cburgdorf:
You mentioned that this would bring a nice speedup and I'm trying to validate that. I'm just comparing some jobs with [another recent run](https://circleci.com/gh/NIC619/trinity/tree/set_up_eth1_monitor_component). I haven't compared all of the jobs but the ones I looked at actually took longer than before? I might just be looking at the wrong jobs. Is there a good way to compare the performance impact?

cburgdorf:
>I'd like to update cache 

Oh ok, so that might explain the performance issues?

PR #1406, opened 2019-12-18
Fix `InvalidGossipMessage` messages and co-attest

hwwhww:
### What was wrong?
Thanks to @ChihChengLiang for post feedback on #1311.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/bcc_libp2p/topic_validators.py
+++ b/trinity/protocol/bcc_libp2p/topic_validators.py
@@ -49,7 +49,7 @@ def get_beacon_block_validator(chain: BaseBeaconChain) -> Callable[..., bool]:
             block = ssz.decode(msg.data, BeaconBlock)
         except (TypeError, ssz.DeserializationError) as error:
             logger.debug(
-                bold_red("Failed to decode block=%s, error=%s"),
+                bold_red("Failed to deserialize BeaconBlock from %s, error=%s"),
                 encode_hex(msg.data),
                 str(error),
             )
@@ -77,7 +77,7 @@ def get_beacon_attestation_validator(chain: BaseBeaconChain) -> Callable[..., bo
         except (TypeError, ssz.DeserializationError) as error:
             # Not correctly encoded
             logger.debug(
-                bold_red("Failed to validate attestation=%s, error=%s"),
+                bold_red("Failed to deserialize Attestation from %s, error=%s"),
                 encode_hex(msg.data),
                 str(error),
             )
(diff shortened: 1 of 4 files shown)


trinity/protocol/bcc_libp2p/topic_validators.py
```diff
@@ -254,8 +251,6 @@ def run_validate_aggregate_and_proof(
             config,
         )
     except ValidationError as error:
-        InvalidGossipMessage(
-            "Failed to validate aggregate_and_proof=%s, error=%s",
-            aggregate_and_proof,
-            str(error),
+        raise InvalidGossipMessage(
+            f"Failed to validate aggregate_and_proof={aggregate_and_proof}, error={str(error)}",
```
ChihChengLiang: say we have
```python
class A(Exception):
    ...

class B(Exception):
    ...
```

if we do 
```python
try:
    raise A("Some A reason")
except A as a:
    raise B("Some B reason") from A
```
we get 

```
__main__.A

The above exception was the direct cause of the following exception:

  File "foo.py", line 12, in <module>
    raise B("Some B reason") from A
__main__.B: Some B reason
```

if we do 

```python
try:
    raise A("Some A reason")
except A as a:
    raise B("Some B reason", a)
```

We get 
```
  File "foo.py", line 10, in <module>
    raise A("Some A reason")
__main__.A: Some A reason

  File "foo.py", line 12, in <module>
    raise B("Some B reason", a)
__main__.B: ('Some B reason', A('Some A reason'))
```

If we want to keep the message from the ValidationError, we can do
```suggestion
            f"Failed to validate aggregate_and_proof={aggregate_and_proof}, error
```

hwwhww: Yeah but I just want to make `InvalidGossipMessage` only in `DEBUG` log instead of shouting loudly. `InvalidGossipMessage` will be caught in topic validator: https://github.com/ethereum/trinity/blob/cffd17f37ef0c7eb2c48fc34c4c91b0c076f4242/trinity/protocol/bcc_libp2p/topic_validators.py#L63-L65

ChihChengLiang: Ah, I see. Note that we can still get the full message in the log with the `B("Some B reason", a)` style. But both ways look good.
```python
def foo():
    try:
        raise A("Some A reason")
    except A as a:
        raise B("Some B reason", a)

import logging

try:
    foo()
except B as b:
    logging.getLogger("__main__").warning(b)
    # ('Some B reason', A('Some A reason'))
```

PR #1413, opened 2019-12-20
Rework Syncer

ChihChengLiang:
### What was wrong?

- Syncer not started on status
- we only query status once

### How was it fixed?

- syncer only starts on STATUS is acquired
- We ask STATUS from all peers every 10 secs

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


ChihChengLiang:
I tried a manual two nodes syncing a long-range of blocks and a few finalized epochs behind, the node that is behind will keep restarting the sync even if it has already imported all the blocks. The issue needs #1154 to fix.

PR #1415, opened 2019-12-20
EIP 868: Node Discovery v4 ENR Extension

gsalgado:
Once this is merged I'll submit separate PRs 
  * move the ENR-related stuff to a common are outside of the discv5 package
  * fetch the most recent ENR of a node before providing peer candidates to clients

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -11,13 +11,16 @@ import random
 import time
 from typing import (
     Any,
+    Awaitable,
     Callable,
+    cast,
     DefaultDict,
     Dict,
     Hashable,
     Iterable,
     Iterator,
     List,
+    Optional,
     Sequence,
     Set,
     Tuple,
@@ -37,6 +40,7 @@ from lahja import (
 )
 
 import rlp
+from rlp.exceptions import DeserializationError
 
 from eth_typing import Hash32
 
@@ -48,6 +52,7 @@ from eth_utils import (
     to_tuple,
     int_to_big_endian,
     big_endian_to_int,
+    ValidationError,
 )
 
 from eth_keys import keys
@@ -59,6 +64,15 @@ from async_service import Service
 
 from p2p import constants
 from p2p.abc import AddressAPI, NodeAPI
+from p2p.discv5.enr import ENR, UnsignedENR, IDENTITY_SCHEME_ENR_KEY
+from p2p.discv5.enr_db import MemoryEnrDb
+from p2p.discv5.identity_schemes import default_identity_scheme_registry, V4IdentityScheme
+from p2p.discv5.constants import (
+    IP_V4_ADDRESS_ENR_KEY,
+    TCP_PORT_ENR_KEY,
+    UDP_PORT_ENR_KEY,
+)
+from p2p.discv5.typing import NodeID
 from p2p.events import (
     PeerCandidatesRequest,
     RandomBootnodeRequest,
@@ -77,7 +91,7 @@ else:
     UserDict = collections.UserDict
 
 # V4 handler methods take a Node, payload and msg_hash as arguments.
-V4_HANDLER_TYPE = Callable[[NodeAPI, Tuple[Any, ...], Hash32], None]
+V4_HANDLER_TYPE = Callable[[NodeAPI, Tuple[Any, ...], Hash32], Awaitable[Any]]
 
 MAX_ENTRIES_PER_TOPIC = 50
 # UDP packet constants.
@@ -96,27 +110,25 @@ class WrongMAC(DefectiveMessage):
     pass
 
 
-class UnknownCommand(DefectiveMessage):
-    pass
-
-
 class DiscoveryCommand:
-    def __init__(self, name: str, id: int, elem_count: int) -> None:
+    def __init__(self, name: str, id: int) -> None:
         self.name = name
         self.id = id
-        # Number of required top-level list elements for this cmd.
-        # Elements beyond this length must be trimmed.
-        self.elem_count = elem_count
 
     def __repr__(self) -> str:
         return 'Command(%s:%d)' % (self.name, self.id)
 
 
-CMD_PING = DiscoveryCommand("ping", 1, 4)
-CMD_PONG = DiscoveryCommand("pong", 2, 3)
-CMD_FIND_NODE = DiscoveryCommand("find_node", 3, 2)
-CMD_NEIGHBOURS = DiscoveryCommand("neighbours", 4, 2)
-CMD_ID_MAP = dict((cmd.id, cmd) for cmd in [CMD_PING, CMD_PONG, CMD_FIND_NODE, CMD_NEIGHBOURS])
+CMD_PING = DiscoveryCommand("ping", 1)
+CMD_PONG = DiscoveryCommand("pong", 2)
+CMD_FIND_NODE = DiscoveryCommand("find_node", 3)
+CMD_NEIGHBOURS = DiscoveryCommand("neighbours", 4)
+CMD_ENR_REQUEST = DiscoveryCommand("enr_request", 5)
+CMD_ENR_RESPONSE = DiscoveryCommand("enr_response", 6)
+CMD_ID_MAP = dict(
+    (cmd.id, cmd)
+    for cmd in [
+        CMD_PING, CMD_PONG, CMD_FIND_NODE, CMD_NEIGHBOURS, CMD_ENR_REQUEST, CMD_ENR_RESPONSE])
 
 
 class DiscoveryService(Service):
@@ -137,6 +149,11 @@ class DiscoveryService(Service):
         self._event_bus = event_bus
         self.this_node = Node(self.pubkey, address)
         self.routing = RoutingTable(self.this_node)
+        self.enr_incoming_channels: Dict[Hash32, trio.MemorySendChannel[ENR]] = {}
+        # FIXME: Use a persistent EnrDb implementation.
+        self._enr_db = MemoryEnrDb(default_identity_scheme_registry)
+        # FIXME: Use a concurrency-safe EnrDb implementation.
+        self._enr_db_lock = trio.Lock()
         self.pong_callbacks = CallbackManager()
         self.ping_callbacks = CallbackManager()
         self.neighbours_callbacks = CallbackManager()
@@ -205,7 +222,7 @@ class DiscoveryService(Service):
         elif node == self.this_node:
             return False
 
-        token = self.send_ping_v4(node)
+        token = await self.send_ping_v4(node)
         log_version = "v4"
 
         try:
@@ -232,6 +249,57 @@ class DiscoveryService(Service):
         self.logger.debug("bonding completed successfully with %s", node)
         self.update_routing_table(node)
         return True
+
+    async def request_enr(self, remote: NodeAPI) -> ENR:
+        # No need to use a timeout because bond() takes care of that internally.
+        await self.bond(remote)
+        token = self.send_enr_request(remote)
+        send_chan: trio.MemorySendChannel[ENR]
+        recv_chan: trio.MemoryReceiveChannel[ENR]
+        send_chan, recv_chan = trio.open_memory_channel(1)
+        enr = None
+        try:
+            async with recv_chan:
+                self.enr_incoming_channels[token] = send_chan
+                with trio.move_on_after(constants.KADEMLIA_REQUEST_TIMEOUT):
+                    enr = await recv_chan.receive()
+        finally:
+            self.enr_incoming_channels.pop(token, None)
+        return enr
+
+    async def get_local_enr(self) -> ENR:
+        async with self._enr_db_lock:
+            try:
+                return await self._enr_db.get(cast(NodeID, self.this_node.id_bytes))
+            except KeyError:
+                pass
+
+            kv_pairs = {
+                IDENTITY_SCHEME_ENR_KEY: V4IdentityScheme.id,
+                V4IdentityScheme.public_key_enr_key: self.pubkey.to_compressed_bytes(),
+                IP_V4_ADDRESS_ENR_KEY: self.address.ip_packed,
+                UDP_PORT_ENR_KEY: self.address.udp_port,
+                TCP_PORT_ENR_KEY: self.address.tcp_port,
+            }
+            enr_seq = 1
+            unsigned_enr = UnsignedENR(enr_seq, kv_pairs)
+            enr = unsigned_enr.to_signed_enr(self.privkey.to_bytes())
+            await self._enr_db.insert(enr)
+            return enr
+
+    async def get_local_enr_seq(self) -> int:
+        enr = await self.get_local_enr()
+        return enr.sequence_number
+
+    async def get_enr(self, remote: NodeAPI) -> Optional[ENR]:
+        """Get the most recent ENR for the given node and update our local DB if necessary."""
+        enr = await self.request_enr(remote)
+        if enr is None:
+            self.logger.debug("Failed to get ENR for %s", remote)
+            return None
+        async with self._enr_db_lock:
+            await self._enr_db.insert_or_update(enr)
+            return await self._enr_db.get(cast(NodeID, remote.id_bytes))
 
     async def wait_ping(self, remote: NodeAPI) -> bool:
         """Wait for a ping from the given remote.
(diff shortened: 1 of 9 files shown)


p2p/discovery.py
```diff
@@ -233,6 +250,57 @@ def update_routing_table(self, node: NodeAPI) -> None:
         self.update_routing_table(node)
         return True
 
+    async def request_enr(self, remote: NodeAPI) -> ENR:
+        # No need to use a timeout because bond() takes care of that internally.
+        await self.bond(remote)
+        token = self.send_enr_request(remote)
+        send_chan: trio.MemorySendChannel[ENR]
+        recv_chan: trio.MemoryReceiveChannel[ENR]
+        send_chan, recv_chan = trio.open_memory_channel(1)
```
pipermerriam: Holy cow!  you figured out how to do it without an ugly `cast`!  This is so much better than what I've been doing for months.!

pipermerriam: I think you can use `0` for this channel size since it only intends to transfer a single value?

p2p/discovery.py
```diff
@@ -443,53 +515,89 @@ def send(self, node: NodeAPI, msg_type: DiscoveryCommand, payload: Sequence[Any]
...
-            self.logger.debug('received message already expired')
-            return
-
-        cmd = CMD_ID_MAP[cmd_id]
-        if len(payload) != cmd.elem_count:
-            self.logger.error('invalid %s payload: %s', cmd.name, payload)
+        try:
+            cmd = CMD_ID_MAP[cmd_id]
+        except KeyError:
+            self.logger.warning("Ignoring uknown msg type: %s; payload=%s", cmd_id, payload)
```
pipermerriam: assuming the payload here is bytes maybe log it as `payload.hex()`

p2p/discovery.py
```diff
@@ -443,53 +515,89 @@ def send(self, node: NodeAPI, msg_type: DiscoveryCommand, payload: Sequence[Any]
...
             return
+        self.logger.debug2("Received %s with payload: %s", cmd.name, payload)
         node = Node(remote_pubkey, address)
         handler = self._get_handler(cmd)
-        handler(node, payload, message_hash)
+        await handler(node, payload, message_hash)
+
+    def _is_msg_expired(self, rlp_expiration: bytes) -> bool:
+        expiration = rlp.sedes.big_endian_int.deserialize(rlp_expiration)
```
pipermerriam: I believe this can throw an exception if `rlp_expiration` isn't a valid encoded big endian integer (is that possible?)

p2p/discovery.py
gsalgado: How were you doing it?

pipermerriam:
```python
send_chan, recv_chan = cast(
    Tuple[trio.abc.SendChannel[...], trio.abc.ReceiveChannel[...]],
    trio.open_memory_channel(0),
)
```

gsalgado: But with no buffer the producer would block until the consumer reads, and I don't see why that would be beneficial -- I think it's fine for the producer to fire and forget in this case

p2p/discovery.py
```diff
@@ -233,6 +250,57 @@ def update_routing_table(self, node: NodeAPI) -> None:
...
+    async def get_local_enr_seq(self) -> int:
+        enr = await self.get_local_enr()
+        return enr.sequence_number
+
+    async def get_enr(self, remote: NodeAPI) -> Optional[ENR]:
+        """Get the most recent ENR for the given node and update our local DB if necessary."""
+        enr = await self.request_enr(remote)
+        if enr is None:
+            self.logger.debug("Failed to get ENR for %s", remote)
+            return None
```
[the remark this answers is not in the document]
gsalgado: Yeah, I guess an exception would be better here. Will change

p2p/discovery.py
gsalgado: No, it is actually a `Tuple[Any, ...]`

p2p/discovery.py
gsalgado: I guess so, but I've just moved this from `_unpack_v4()` to here, so if we haven't seen it yet in practice I don't think it will become a problem now. FWIW, had to move it here because after EIP-868 the expiration is not always the last element in the payload

gsalgado:
You haven't forgotten about me, right @pipermerriam? :-)

p2p/kademlia.py
```diff
@@ -77,6 +77,11 @@ def is_private(self) -> bool:
     def ip(self) -> str:
         return str(self._ip)
 
+    @property
```
pipermerriam: minor but we can use `cached_property` for these types of things.

p2p/discovery.py
pipermerriam: no complaints about the buffer size.  Only commenting on the improvement of how you typed the two channel variables.

p2p/discovery.py
```diff
@@ -851,11 +1013,11 @@ def _get_max_neighbours_per_packet() -> int:
...
     node_data = addr.to_endpoint() + [b'\x00' * (constants.KADEMLIA_PUBLIC_KEY_SIZE // 8)]
     neighbours = [node_data]
-    expiration = rlp.sedes.big_endian_int.serialize(_get_msg_expiration())
-    payload = rlp.encode([neighbours] + [expiration])
+    expiration = _get_msg_expiration()
+    payload = rlp.encode(tuple([neighbours]) + (expiration,))
     while HEAD_SIZE + len(payload) <= 1280:
         neighbours.append(node_data)
-        payload = rlp.encode([neighbours] + [expiration])
+        payload = rlp.encode(tuple([neighbours]) + (expiration,))
```
pipermerriam: This looks like a good use case for a `NamedTuple`

```python
class MaxNeighborsPacket(NamedTuple):
    neighbors: Tuple[bytes, ...]
    expiration: int
```

Not sure if I got the value types right but an instance of that can be passed directly into `rlp.encode` since it is also just a tuple.

ralexstokes: tangential question but would would `rlp.encode` handle this for `dataclass` instances as well?

p2p/discovery.py
gsalgado: I think our lines got crossed; that last comment of mine was in response to this:

> I think you can use 0 for this channel size since it only intends to transfer a single value?

p2p/discovery.py
gsalgado: Yeah, a NamedTuple sounds like a good idea here

gsalgado: I've also changed `send_neighbours_v4()` to use that

gsalgado: It looks like dataclass was added in 3.7? If that's the case I'd be surprised if rlp supported that

gsalgado:
> I think that the async_service.trio.external_api decorator is ready to use and this looks like the DiscoveryService exposes a few methods that should potentially start using this pattern?

I thought the idea was to have `DiscoveryService` do its thing independently and have other services talk to it via the event bus? I don't know which methods here would be called directly

pipermerriam:
> I thought the idea was to have `DiscoveryService` do its thing independently and have other services talk to it via the event bus? I don't know which methods here would be called directly.

The methods that should be decorated with `external_api` are ones that are called from outside of the `Service` instance.  Basically, if it is **only** called via `self.invoke_method(...)` then it can be normal. Otherwise if someone does `service.invoke_method(...)` then it should be decorated.  I'll have to improve the docs for this.

gsalgado:
>  The methods that should be decorated with external_api are ones that are called from outside of the Service instance

Right, that was my understanding, but here ISTM no APIs are used by external services -- the only ones I see talking to DiscoveryService are the `*PeerBackend`s that do so by sending `PeerCandidatesRequest`s over the event bus

p2p/discovery.py
ralexstokes: yeah i would think no -- i was just curious if `py-rlp` already supported `dataclasses` or not

p2p/discv5/enr.py
```diff
@@ -378,8 +383,8 @@ def __repr__(self) -> str:
     b"ip6": Binary.fixed_length(IP_V6_SIZE),
     b"tcp6": big_endian_int,
     b"udp6": big_endian_int,
+    b"eth": FORK_ID_SEDES,
```
ralexstokes: curious where the docs for this are

tests-trio/p2p-trio/test_discovery.py
```diff
@@ -33,9 +50,14 @@ def short_timeout(monkeypatch):
     alice, bob = manually_driven_discovery_pair
     # Collect all pongs received by alice in a list for later inspection.
     received_pongs = []
-    alice.recv_pong_v4 = lambda node, payload, hash_: received_pongs.append((node, payload))
 
-    token = alice.send_ping_v4(bob.this_node)
+    async def recv_pong(node, payload, hash_):
+        await trio.hazmat.checkpoint()
```
ralexstokes: any particular reason we need to `checkpoint` here?

tests-trio/p2p-trio/test_discovery.py
```diff
@@ -57,7 +134,12 @@ def short_timeout(monkeypatch):
 
     # Collect all neighbours packets received by alice in a list for later inspection.
     received_neighbours = []
-    alice.recv_neighbours_v4 = lambda node, payload, hash_: received_neighbours.append((node, payload))  # noqa: E501
+
+    async def recv_neighbours(node, payload, hash_):
+        await trio.hazmat.checkpoint()
```
ralexstokes: ditto question here

tests-trio/p2p-trio/test_discovery.py
```diff
@@ -309,12 +388,14 @@ def send(self, node, msg_type, payload):
         # network as that wouldn't work.
         raise ValueError("MockDiscoveryService must not be used to send network messages")
 
-    def send_ping_v4(self, node):
+    async def send_ping_v4(self, node):
+        await trio.hazmat.checkpoint()
```
ralexstokes: ditto

tests-trio/p2p-trio/test_discovery.py
```diff
@@ -309,12 +388,14 @@ def send(self, node, msg_type, payload):
...
         echo = hex(random.randint(0, 2**256))[-32:]
         self.messages.append((node, 'ping', echo))
         return echo
 
-    def send_pong_v4(self, node, echo):
+    async def send_pong_v4(self, node, echo):
+        await trio.hazmat.checkpoint()
```
ralexstokes: ditto

p2p/discovery.py
```diff
@@ -77,7 +92,7 @@
     UserDict = collections.UserDict
 
 # V4 handler methods take a Node, payload and msg_hash as arguments.
```
ralexstokes: update this comment? it would be nice if in isolation i had some idea what this `Awaitable[Any]` is doing

p2p/discovery.py
```diff
@@ -443,53 +519,89 @@ def send(self, node: NodeAPI, msg_type: DiscoveryCommand, payload: Sequence[Any]
...
+        elif len(payload) == 3:
+            _, token, expiration = payload[:3]
+            enr_seq = None
+        else:
+            _, token, expiration, enr_seq = payload[:4]
+        if self._is_msg_expired(expiration):
+            return
         self.logger.debug2('<<< pong (v4) from %s (token == %s)', node, encode_hex(token))
         self.process_pong_v4(node, token)
+        await trio.hazmat.checkpoint()
```
ralexstokes: does `trio` not insert checkpoints in between `async` function calls? if so, then it seems like you can drop the manual `checkpoint` here right?

p2p/discovery.py
```diff
@@ -443,53 +519,89 @@ def send(self, node: NodeAPI, msg_type: DiscoveryCommand, payload: Sequence[Any]
...
+        nodes, expiration = payload[:2]
+        if self._is_msg_expired(expiration):
+            return
         neighbours = _extract_nodes_from_payload(node.address, nodes, self.logger)
         self.logger.debug2('<<< neighbours from %s: %s', node, neighbours)
         self.process_neighbours(node, neighbours)
-
-    def recv_ping_v4(self, node: NodeAPI, _: Any, message_hash: Hash32) -> None:
-        self.logger.debug2('<<< ping(v4) from %s', node)
+        await trio.hazmat.checkpoint()
```
ralexstokes: same question for trailing `checkpoint` here...

p2p/discovery.py
ralexstokes: my 2 dai on this is that we should prefer to have as small a buffer as possible anywhere we can -- in this case, the "thread" of execution working on receiving the remote's ENR is blocked until that event occurs anyway so we can keep the channel to buffer size `0`. in my experience w/ channels, introducing buffers is usually a way to paper over bugs in concurrent code and over time leads to more memory usage than is strictly necessary.

i think we are fine to leave this alone for now but i just want to share my thinking on it

p2p/discovery.py
```diff
@@ -233,6 +251,60 @@ def update_routing_table(self, node: NodeAPI) -> None:
...
+        # No need to use a timeout because bond() takes care of that internally.
+        await self.bond(remote)
+        token = self.send_enr_request(remote)
+        send_chan: trio.MemorySendChannel[ENR]
+        recv_chan: trio.MemoryReceiveChannel[ENR]
+        send_chan, recv_chan = trio.open_memory_channel(1)
+        enr = None
+        try:
+            async with recv_chan:
+                self.enr_incoming_channels[token] = send_chan
```
ralexstokes: how are we synchronizing access to this `self.enr_incoming_channels` dict between the code here and the code in `recv_enr_request`?

is there any chance of a race condition where we go to look up that channel in the dict on line 650 and it is missing before we can insert it on line 264?

p2p/discv5/enr.py
gsalgado: for now that'd be geth's source code :-/

tests-trio/p2p-trio/test_discovery.py
gsalgado: I seem to remember trio complaining about an async function that never yielded control, but that no longer seems to be the case, so I removed those everywhere.

p2p/discovery.py
gsalgado: Updated to say they're async methods -- that's what the `Awaitable` means

p2p/discovery.py
gsalgado: Not really because we don't yield control between sending the ENR request and adding the channel to the dict here. This could become a problem, though, if we eventually have multiple coroutines sending ENR requests and two of them send to the same remote, but we shouldn't be doing that in the first place.

gsalgado:
> why do we remove the "maximum element count", which some comment suggests was due to EIP-8?

I didn't remove that, simply moved the validation of element count from `_unpack_v4()` to the msg handlers themselves. That was needed because before this change every command implementation supported only payloads of a certain length (by ignoring every item above that length), now we have some commands (e.g. ping/pong can now handle packets with or without an enr-seq as last element) that support varying lengths

> it looks like we are extending the existing set (before EIP-868) of discv4 messages with an expiration field 

The expiration was always there, but before it was added automatically (for all msgs) in `_pack_v4` as it was always the last element in any payload. That is no longer the case now (e.g. for ping/pong that have an optional enr-seq as last element), so this had to be moved to the msg handlers as well

p2p/discovery.py
gsalgado: I've just pushed a change to wrap some of this code in a class that will raise a more meaningful error if we try to sent simultaneous ENR requests to a certain node. I've made it generic as I intend to use it for other msg types, but I'll do that in a separate PR as they currently use an awkward callback mechanism

gsalgado:
Hey @pipermerriam and @ralexstokes, I think I've addressed everything here; are you guys happy for me to merge this?

pipermerriam:
I'm :+1: but I didn't re-review.  I'll delegate to @ralexstokes for final :+1: .

gsalgado:
@pipermerriam if we don't hear from @ralexstokes today, would you mind signing off on this? I've already submitted a couple other PRs that depend on this and have plenty more stuff in the pipeline, so this is kind of blocking me

ralexstokes:
@gsalgado thanks for the updates and answers! very helpful.

sorry for the delay w/ the holidays -- go ahead and merge this in when you are ready :)

PR #1417, opened 2019-12-20
Drain writer before closing transport

pipermerriam:
extracted from #1411 

builds on: https://github.com/ethereum/trinity/pull/1419

### What was wrong?

The closing of a `Transport` was just directly closing the writer without draining it.  This resulted in unwanted behavior like not actually transmitting the disconnect reason before closing connections with other peers.

### How was it fixed?

Changed the `Transport.close()` to be asynchronous and added a call to `writer.drain()` to ensure that any pending data gets written to the socket before closing it.

### To-Do

- [x] Clean up commit history


pipermerriam:
cc @carver I *think* this might have been why we weren't getting a disconnect reason when two trinity peers disconnect from each other.  The socket was likely being closed before the message had a chance to make it across the wire.

PR #1419, opened 2019-12-20
fix already used port when running tests concurrently

pipermerriam:
### What was wrong?

The tests in `./tests/core/p2p-proto/test_server.py` were using a port that was assigned during module loading.  Turing test runs this just started failing due to the port already being in use.

### How was it fixed?

Adjusted test to use a unique port for all runs.

#### Cute Animal Picture

FDCAPET20131-MSTR


tests/core/p2p-proto/test_server.py
```diff
@@ -43,15 +40,11 @@
 
 port = get_open_port()
```
gsalgado: Do we still need this or can we get rid of it to ensure nobody accidentally uses it in new tests?

PR #1427, opened 2019-12-24
Remove slot to state mapping

NIC619:
### What was wrong?
slot to state mapping is rarely used. We rarely use it to get state from past slot. Mostly used in `chain.get_head_state()`.

And also this mapping is not easy to maintain. We currently only `persist_state` when importing a new block or when validator skip a slot. So when re-org happens, outdated slot to state entry will not be updated. For example if old chain map slot 111 to state S' and new canonical chain skip slot 111, then `get_state_by_slot(111)` would return state S'.
Or another example if tip of old chain is at slot 100 but tip of new canonical chain is at slot 80, state from slot 80-100 are not pruned.

So IMO it's not worth the effort to keep track of this mapping.
This PR supersedes #1154 .

### How was it fixed?
- [x] Remove slot to state mapping
- [x] We still keep head state info, i.e., head state slot and head state root. And these info are updated when
    1. imported a new canonical block
    2. validator skip a slot
    - basically updated in the same circumstances as before but the difference is that these info are not updated in `chain.persist_state` but instead updated in `chaindb. update_head_slot_and_state_root`.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/tests/eth2/core/beacon/db/test_beacon_chaindb.py
+++ b/tests/eth2/core/beacon/db/test_beacon_chaindb.py
@@ -180,7 +180,7 @@ def test_chaindb_get_genesis_block_root(chaindb, genesis_block, fork_choice_scor
     assert block_root == genesis_block.signing_root
 
 
-def test_chaindb_get_head_state_slot(chaindb, state):
+def test_chaindb_get_head_state_slot_and_root(chaindb, state):
     # No head state slot stored in the beginning
     with pytest.raises(HeadStateSlotNotFound):
         chaindb.get_head_state_slot()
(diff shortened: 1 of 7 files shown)


tests/eth2/core/beacon/db/test_beacon_chaindb.py
```diff
@@ -196,8 +196,6 @@ def test_chaindb_get_head_state_slot(chaindb, state):
 def test_chaindb_state(chaindb, state):
     chaindb.persist_state(state)
     state_class = BeaconState
-    result_state_root = chaindb.get_state_root_by_slot(state.slot)
```
ChihChengLiang: We can put `get_head_state_root` here to protect it with this test.

NIC619: I added the tests for `get_head_state_root` and `update_head_slot_and_state_root`.

hwwhww:
Whoops, somehow my comment is lost between browser tabs. :sweat_smile: 
1. Two-steps state persistence might introduces some side effect.
2. We may need `get_state_by_slot` for HTTP APIs.

I think we can keep `BeaconChain.get_state_by_slot` and remove `slot -> state` mapping by using `BeaconChain.get_canonical_block_by_slot`.

```python
    def get_state_by_slot(self, slot: Slot) -> BeaconState:
        """
        Return the requested state as specified by slot number.
        Raise ``StateNotFound`` if there's no state with the given slot number in the db.
        """
        sm_class = self.get_state_machine_class_for_block_slot(slot)
        state_class = sm_class.get_state_class()
        try:
            block = self.get_canonical_block_by_slot(slot)
        except BlockNotFound:
            raise StateNotFound("No state root for slot #{0}".format(slot))
        state_root = block.state_root
        return self.chaindb.get_state_by_root(state_root, state_class)
```

NIC619:
@hwwhww  Yeah this works! I will update it tomorrow.

What do you think about applying state transition instead of returning `StateNotFound`?

hwwhww:
> applying state transition instead of returning `StateNotFound`?

Do you mean skipping the slot? Hmm, I wonder if using state transition here might expose a DOS attack vector on HTTP APIs endpoints.

NIC619:
@ChihChengLiang @hwwhww 
Added back `get_state_by_slot` and renamed `update_head_slot_and_state_root` to `update_head_state`.

PR #1428, opened 2019-12-24
Implement EIP-2124: Fork identifier for chain compatibility checks

gsalgado:


p2p/forkid.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def __init__(self, hash: bytes, next: BlockNumber) -> None:
+        # We often instantiate ForkIDs directly (rather than via rlp.decode()), so we validate the
+        # hash length here as well to catch mistakes before we attempt to serialize.
+        if len(hash) != 4:
+            raise ValueError("Hash {hash!r} length is not 4")
+        super().__init__(hash, next)
+
+
+def _extract_fork_blocks(
```
cburgdorf: I think we usually try to avoid these intermediate lists either by using comprehension syntax or `yield` + `to_tuple` syntax. In this case that would be.
```suggestion
@to_tuple
def _extract_fork_blocks(
        vm_configuration: Tuple[
            Tuple[BlockNumber, Type[VirtualMachineAPI]], ...]
) -> Tuple[BlockNumber, ...]:
    for block, vm in vm_configuration:
        if block == 0:
            continue
        yield block
        if issubclass(vm, HomesteadVM) and vm.support_dao_fork:
            yield vm.get_dao_fork_block_number()
```

p2p/forkid.py
```diff
@@ -0,0 +1,116 @@
...
+        if block_number > current_head:
+            next_fork = block_number
+            break
+        # As defined in the spec, we must checksum the 64-bit representation of the block number.
+        pre_hash_bytes += block_number.to_bytes(8, 'big')
+    checksum = _crc_to_bytes(binascii.crc32(pre_hash_bytes))
+    return ForkID(hash=checksum, next=BlockNumber(next_fork))
+
+
+def validate_forkid(
```
cburgdorf: Maybe add a link to https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2124.md#validation-rules here

gsalgado:
I'd rather hold back on the newsfragment until we are actually including the fork-id in our ENR (#1415) and validating the ones we receive

p2p/forkid.py
gsalgado: Makes sense

p2p/forkid.py
gsalgado: That's cool!

cburgdorf:
>I'd rather hold back on the newsfragment until we are actually including the fork-id in our ENR (#1415) and validating the ones we receive

Fair enough!

PR #1434, opened 2019-12-27
Rm signing root

### LINKED ISSUE #1369: rm signing root
### What is wrong?

https://github.com/ethereum/eth2.0-specs/pull/1491/

This will happen in the next spec release. Just do a quick overview of what could be affected

## SSZ

- remove SignedHashableContainer

### Add new type

- Add
    - SignedVoluntaryExit
    - SignedBeaconBlock
    - SignedBeaconBlockHeader
- Modify
    - DepositMessage
    - ProposerSlashing
    - BeaconBlockBody

### Helpers

- Add
    - verify_block_signature
- Modify
    - state_transition
    - process_slots
        - _process_slot
    - process_block_header
    - process_proposer_slashing
    - process_deposit
    - process_voluntary_exit
    - on_block
- Non-spec helper
    - create_block_on_state
        - add compute_new_state_root
        - add get_block_signature

### Unused API

- Remove
    - BeaconChain
        - get_block_by_hash_tree_root
    - ChainDB
        - get_block_signing_root_by_hash_tree_root
- Modify
    - OrphanBlockPool
        - Can use OperationPool directly now? We are using the `hash_tree_root(SignedBeaconBlock.message)`.

### Expect a lot of diff from API that has type SigningRoot

Mostly just rename, but some APIs in BeaconChain and BeaconChainDB need to get the correct root from correct object carefully.

ChihChengLiang:
### What was wrong?

Fix #1369

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/chains/base.py
+++ b/eth2/beacon/chains/base.py
@@ -18,7 +18,7 @@ from eth2.beacon.types.attestations import Attestation
 from eth2.beacon.types.blocks import BaseSignedBeaconBlock
 from eth2.beacon.types.nonspec.epoch_info import EpochInfo
 from eth2.beacon.types.states import BeaconState
-from eth2.beacon.typing import HashTreeRoot, SigningRoot, Slot, Timestamp
+from eth2.beacon.typing import HashTreeRoot, Slot, Timestamp
 from eth2.configs import Eth2Config, Eth2GenesisConfig
 
 if TYPE_CHECKING:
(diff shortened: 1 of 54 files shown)


ChihChengLiang:
We are reaching 1k diff now, but I'll break it to smaller PRs once this approach looks good to everyone.

The approach shown here is to operate the signed blocks almost all the time. Using signed blocks in some places might make us feel uncomfortable, but I'll justify here why we are biting these bullets.
- "Storing signed blocks in chaindb": One might think that once a node received a block from a peer, the signature of the block is verified and thus can be safely tossed away. But that's not the case, because the node must send signed blocks when other peer requests a branch. So signed blocks must be stored.
- "Genesis block is a signed block": Signing a genesis block makes no sense. But since we store all blocks with their signature in chaindb, genesis block should be no exception because the ssz container type for all blocks should be the same.
- "Use signed blocks in the state transition": Theoretically, the signature is validated when received on the wire, so no bother checking them again in state transition. Doing so is to minimize disparity from the spec.

hwwhww:
@ChihChengLiang 

- "Storing signed blocks in chaindb": sounds fine to me.
- "Genesis block is a signed block": hmm, I do feel it's equivocal when calling `get_genesis_block_root`. Maybe rename to `get_signed_genesis_block_root`?
- "Use signed blocks in the state transition": sounds fine to me. :)

ChihChengLiang:
#1472 is preferred

PR #1438, opened 2019-12-30
DiscoveryService now connects to the DB

gsalgado:
Something like this is needed in order for us to generate the ForkID that should be included in our local ENR record.

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -7,6 +7,7 @@ More information at https://github.com/ethereum/devp2p/blob/master/rlpx.md#node-
 """
 import collections
 import contextlib
+from pathlib import Path
 import random
 import time
 from typing import (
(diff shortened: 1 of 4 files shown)


gsalgado:
@pipermerriam I know you'd like to keep DiscoveryService unaware of the ETH chain, but to generate the forkid we need the vm_config, the current head and the genesis hash, so ISTM we'd either have to give it direct access to the chain DB or implement a service that provides the forkid over the event bus?

pipermerriam:
Can we do something like give the discovery service an opaque API mechanism for generating the ENR without actually having it be natively aware of the vm configuration.

```python
ENR_RecordProviderFN = Callable[[], Tuple[str, str]]

class DiscoveryService:
    def __init__(self, ..., enr_record_providers=Sequence[ENR_RecordProviderFN]):
        ...

    def get_enr_record_fields(self) -> Tuple[Tuple[str, str], ...]:
        return tuple(
            provider_fn() for provider_fn in self._enr_record_providers
        )
```

Something like this so that we can initialize the service with a function that will do what it needs without actually having it have any native awareness of the vm configuration.

gsalgado:
That sounds good to me, I'll give it a go

PR #1439, opened 2019-12-30
Fix discv4 lookup() to not return duplicate nodes

gsalgado:
That method could in some cases return duplicate nodes

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -368,7 +368,9 @@ class DiscoveryService(Service):
             results = await trio_utils.gather(*next_find_node_queries)
             for candidates in results:
                 closest.extend(candidates)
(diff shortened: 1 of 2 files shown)


gsalgado:
Well, that is an internal API so I'm not sure we want to pollute the changelog with that. What do you say?

cburgdorf:
>Well, that is an internal API so I'm not sure we want to pollute the changelog with that.

I'm assuming that this fixes an actual bug would have been visible to an end user (error or warning in logs etc). If not, then I agree that we don't need a changelog entry for this.

gsalgado:
> I'm assuming that this fixes an actual bug would have been visible to an end user (error or warning in logs etc). If not, then I agree that we don't need a changelog entry for this.

No, the items returned by `lookup()` are consumed only by DiscoveryService and used to populate the routing table, so those duplicates would be discarded when being inserted there.

PR #1448, opened 2020-01-03
Ensure Kbucket's replacement cache doesn't contain duplicates

gsalgado:
Also enforce a size limit on them.

### DIFF
--- a/tests/p2p/test_kademlia.py
+++ b/tests/p2p/test_kademlia.py
@@ -120,6 +120,36 @@ def test_kbucket_add():
     assert bucket.head == node2
 
 
+def test_kbucket_replacement_cache():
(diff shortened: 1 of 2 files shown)


tests/p2p/test_kademlia.py
```diff
@@ -120,6 +120,36 @@ def test_kbucket_add():
     assert bucket.head == node2
 
 
+def test_kbucket_replacement_cache():
+    """Check that the replacement cache has a limited size and doesn't contain duplicates."""
```
cburgdorf: *Nitpick*: Our styleguide discourages from putting the first doc line on the same line as the opening triple quotes: https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#exceptions-to-docstring-style-peps

PR #1452, opened 2020-01-07
Extend `BeaconChain.get_state_by_slot` by adding `get_state_at_slot` helper

hwwhww:
### What was wrong?
It's easy to get recent state roots from head state. This PR extends `BeaconChain.get_state_by_slot` with `BeaconState` helper.

### How was it fixed?
1. Add `get_state_root_at_slot` helper
2. Use `get_state_root_at_slot` helper in
   `BeaconChain.get_state_by_slot`
3. Add `get_state_by_root` for HTTP APIs

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

dark-night-river-forest-minimal-art

### DIFF
--- a/eth2/beacon/chains/base.py
+++ b/eth2/beacon/chains/base.py
@@ -13,6 +13,7 @@ from eth2.beacon.db.chain import BaseBeaconChainDB, BeaconChainDB, StateNotFound
 from eth2.beacon.exceptions import BlockClassError, StateMachineNotFound
 from eth2.beacon.fork_choice.constant import ConstantScoring
 from eth2.beacon.fork_choice.scoring import BaseScore
+from eth2.beacon.helpers import get_state_root_at_slot
 from eth2.beacon.types.attestations import Attestation
 from eth2.beacon.types.blocks import BaseBeaconBlock
 from eth2.beacon.types.nonspec.epoch_info import EpochInfo
@@ -105,6 +106,10 @@ class BaseBeaconChain(Configurable, ABC):
     #
     # State API
     #
+    @abstractmethod
+    def get_state_by_root(self, root: HashTreeRoot) -> BeaconState:
+        ...
+
     @abstractmethod
     def get_state_by_slot(self, slot: Slot) -> BeaconState:
         ...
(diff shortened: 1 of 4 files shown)


tests/eth2/core/beacon/test_helpers.py
```diff
@@ -85,6 +97,47 @@ def test_get_block_root_at_slot(
...
+    state = BeaconState.create(**sample_beacon_state_params).mset(
+        "slot", current_slot, "state_roots", state_roots
+    )
+    blocks[current_slot].state_root = state.hash_tree_root
+
+    if success:
+        state_root = get_state_root_at_slot(
+            state, target_slot, slots_per_historical_root
+        )
+        assert state_root == blocks[target_slot].state_root
```
NIC619: The logic looks correct but it seems like these blocks' state root are all the same? (So this assertion is actually trivial)

eth2/beacon/chains/base.py
```diff
@@ -302,22 +307,43 @@ def on_attestation(self, attestation: Attestation, slot: Slot = None) -> None:
...
-            # And so applying state transition could be potential DoS vector in this case.
-            raise StateNotFound(f"No state root for slot #{slot})")
-        state_root = block.state_root
+            try:
+                #  If the requested state_root is computed with skipped slot recently,
+                #  try to get recent state root from state.
+                head_state = self.get_head_state()
+                state_root = get_state_root_at_slot(
+                    head_state, slot, config.SLOTS_PER_HISTORICAL_ROOT
+                )
```
NIC619: This would result in we can only get state at a skipped slot for the past `config.SLOTS_PER_HISTORICAL_ROOT` slots since we are using `head_state`?

hwwhww: `get_state_root_at_slot` is getting state roots from `head_state.state_roots`, which contains *all* past `config.SLOTS_PER_HISTORICAL_ROOT` slots, not only the skipped ones.

The comment "If the requested state_root is computed with skipped slot recently" is describing the "except BlockNotFound:" case.

PR #1454, opened 2020-01-07
DiscoveryService.consume_datagram() now runs receive() in the bg

gsalgado:
That way we don't have to wait until a msg is processed before consuming 
the next

This depends on my discv4-enr branch, which has already been through review, so just the last
commit needs reviewing here

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -521,7 +521,7 @@ class DiscoveryService(Service):
             constants.DISCOVERY_DATAGRAM_BUFFER_SIZE)
         address = Address(ip_address, port)
         self.logger.debug2("Received datagram from %s", address)
-        await self.receive(address, datagram)
+        self.manager.run_task(self.receive, address, datagram)
 
     async def receive(self, address: AddressAPI, message: bytes) -> None:
         try:
--- a/tests-trio/p2p-trio/conftest.py
+++ b/tests-trio/p2p-trio/conftest.py
@@ -72,3 +72,10 @@ class ManuallyDrivenDiscoveryService(DiscoveryService):
     async def run(self) -> None:
         self.ready_to_drive.set()
         await self.manager.wait_finished()
+
+    async def consume_datagram(self) -> None:
+        await super().consume_datagram()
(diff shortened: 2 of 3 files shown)


p2p/discovery.py
```diff
@@ -519,7 +519,7 @@ def send(self, node: NodeAPI, msg_type: DiscoveryCommand, payload: Sequence[Any]
             constants.DISCOVERY_DATAGRAM_BUFFER_SIZE)
         address = Address(ip_address, port)
         self.logger.debug2("Received datagram from %s", address)
-        await self.receive(address, datagram)
+        self.manager.run_task(self.receive, address, datagram)
```
pipermerriam: Just realized this: https://github.com/ethereum/async-service/issues/43

Probably fine for now but wanted to point it out because it's likely to show up soon and should be fixed before we make a new release.

tests-trio/p2p-trio/conftest.py
```diff
@@ -72,3 +72,10 @@ def __init__(self, *args, **kwargs):
     async def run(self) -> None:
         self.ready_to_drive.set()
         await self.manager.wait_finished()
+
+    async def consume_datagram(self) -> None:
+        await super().consume_datagram()
+        # Our parent's consume_datagram() starts a background task to process the msg, so we yield
+        # control here to give that a chance to run. This avoid us having to do so in every test
+        # that calls consume_datagram().
+        await trio.sleep(0)
```
pipermerriam: *nitpick*

`trio.sleep(0)` is best written as `trio.hazmat.checkpoint()` .

gsalgado: I knew about it, but why is it preferable?

pipermerriam: I think for 2 reasons.

1. is that checkpoints in trio code are expected, but `sleep(...)` calls are usually anti-patterns.
2. `sleep(0)` will have some internal logic in addition to the call to `trio.hazmat.checkpoint()` that has to occur in there as well.  That means calling `trio.hazmat.checkpoint()` directly should be *slightly* more performant since it would bypass all of the code that checks the `seconds` argument.

Issue #674, opened 2019-05-26
Convert `p2p.discovery` callback based architecture to use futures and timeouts.


pipermerriam:
### What is wrong?

https://github.com/ethereum/trinity/blob/86b1bfb6f5fcc7b943fb408ae206c047aecb828d/p2p/discovery.py#L830-L843

The callback base approach used here, as well as the way the function behaves could be better.  First, the callback approach is a more complex but functionally equivalent way of implementing an `asyncio.Future`.

Second, the function accounts for not receiving the pong message in time by returning the default *empty* values along with the `got_pong` boolean indicating success/failure.  This is return value checking which is an inherently error prone API design.

### How can it be fixed

The callback part should be fixed by using an `asyncio.Future` instead of the callback.

The return value checking *should* be able to be replaced with an exception based API, raising something like a `TimeoutError` in the case that a pong is not received in time.  Call sites will need to be adjusted to account for this change.

gsalgado:
@pipermerriam now that we're using trio on DiscoveryService, I think the best alternative is to use channels here. How does https://github.com/gsalgado/trinity/commit/26809b09d396f298177664b5e405d1ab31a4c17a look like to you?

pipermerriam:
:+1: on the direction in the linked github commit.  Left a few comments on that commit.

PR #1456, opened 2020-01-08
discv4: use ExpectedResponseChannels for neighbours msgs

gsalgado:
Part of #674

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -11,6 +11,7 @@ import random
 import time
 from typing import (
     Any,
+    AsyncIterator,
     Awaitable,
     Callable,
     cast,
@@ -29,6 +30,8 @@ from typing import (
     TypeVar,
     TYPE_CHECKING,
 )
+
+from async_generator import aclosing
 
 import trio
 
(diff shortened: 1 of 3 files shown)


p2p/discovery.py
```diff
@@ -337,33 +340,28 @@ def update_routing_table(self, node: NodeAPI) -> None:
...
+                # mypy thinks wrapping our generator turns it into something else, so ignore.
+                async with aclosing(gen):  # type: ignore
+                    async for batch in gen:
+                        self.logger.debug2(
+                            f'got expected neighbours response from {remote}: {batch}')
+                        neighbours.extend(batch)
+                        if len(neighbours) >= constants.KADEMLIA_BUCKET_SIZE:
+                            break
+                self.logger.debug2(f'got expected neighbours response from {remote}')
+        except trio.TooSlowError:
```
pipermerriam: I think the following format might be more readable.

```python
with trio.move_on_after(...) as cancel_scope:
    ...
if cancel_scope.cancelled_caught:
    self.logger.debug(...)
```

Just a slight reduction in levels of indention.  It does have the slight downside of allowing for the silent failure if the `if cancel_scope.cancelled_caught` block is accidentally removed, but I think I prefer this over the large `try` block.

p2p/discovery.py
```diff
@@ -1110,20 +1102,53 @@ def already_waiting_for(self, remote: NodeAPI) -> bool:
...
+        on the given receive channel.
+
+        Closes the channel once a message is received.
+
+        Returns the read message.
+        """
+        gen = self.receive(remote, send_chan, recv_chan)
+        # mypy thinks wrapping our generator turns it into something else, so ignore.
+        async with aclosing(gen):  # type: ignore
+            return await gen.__anext__()
```
pipermerriam: Thoughts on doing this as:

```python
async for msg in self.receive(...):
    return msg
```

This should have the same effect of returning after the first value...

gsalgado: That's how I initially wrote it, but then mypy complains that the method is missing a return statement. Is there a way around that?

p2p/discovery.py
gsalgado: Yeah, I think I also prefer this over the large try and extra level of indentation

p2p/discovery.py
pipermerriam: I think this works:

```python
async for msg in self.receive(...):
    break
return msg
```

pipermerriam: it might also work to do `return await self.receive(...).asend(None)` but I'm not 100% sure that statement is equivalent....

gsalgado:
```
        gen = self.receive(remote, send_chan, recv_chan)
        async with aclosing(gen):  # type: ignore
            async for msg in gen:
                return msg
```

> p2p/discovery.py:1104: error: Missing return statement

PR #1460, opened 2020-01-08
convert event bus service to use async-service APIs

pipermerriam:
### What was wrong?

The services we use to manage the event bus were still using the legacy asyncio service.

### How was it fixed?

Updated it to use the new `async-service` APIs.

#### Cute Animal Picture

733290aee704addb395ea9b961a2e6dc

### DIFF
--- a/trinity/extensibility/event_bus.py
+++ b/trinity/extensibility/event_bus.py
@@ -1,6 +1,6 @@
 import asyncio
 import logging
-from typing import AsyncGenerator
+from typing import Type, Union
 
 import trio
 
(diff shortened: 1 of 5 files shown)


trinity/extensibility/event_bus.py
```diff
@@ -12,116 +12,43 @@
...
         self._trinity_config = trinity_config
-        self._endpoint_available = asyncio.Event()
+        self._endpoint_available = self.EventType()
         self._connection_config = ConnectionConfig.from_name(
             endpoint_name, self._trinity_config.ipc_dir
         )
-        super().__init__(cancel_token, loop)
 
-    async def wait_event_bus_available(self) -> None:
+    async def get_event_bus(self) -> None:
```
gsalgado: It feels weird to have a `get_something()` method that always returns None

trinity/extensibility/event_bus.py
```diff
@@ -12,116 +12,43 @@
...
-        self._endpoint_available = trio.Event()
-        self._connection_config = ConnectionConfig.from_name(
-            endpoint_name, self._trinity_config.ipc_dir
-        )
-
-    async def wait_event_bus_available(self) -> None:
-        await self._endpoint_available.wait()
-
-    def get_event_bus(self) -> TrioEndpoint:
         return self._endpoint
```
gsalgado: Oh, wait, it does return the endpoint! How did mypy not catch this?

pipermerriam:
No clue but in running this branch I pretty consistently am ending up with a bunch of timeouts and hard failures from the various endpoints failing to connect to each other... so need to figure that out before this moves forward

pipermerriam:
@gsalgado no substantial changes other than fixing the type hint you pointed out which I'm going to do separately because it is a larger problem with our mypy config that exposes about 100 more places where this is happening.

trinity/extensibility/event_bus.py
```diff
@@ -131,59 +58,64 @@ def get_event_bus(self) -> TrioEndpoint:
...
+            try:
+                await endpoint.connect_to_endpoints(*endpoints_to_connect_to)
+            except Exception as e:
+                self.logger.warning(
+                    "Failed to connect %s to one of %s: %s",
+                    endpoint.name,
+                    endpoint_names,
+                    e,
+                )
+                raise
```
gsalgado: As we're re-raising the exception, I wonder if it makes sense to log this warning here?

pipermerriam:
> LGTM, but I'm curious how you solved the above?

It solved itself :confused:  ... I don't believe there is much to be concerned about and I'm keeping an eye out for if it resurfaces but I currently think it was just something that was slightly out of place that got inadvertently fixed in some rebasing/refactoring...

trinity/extensibility/event_bus.py
pipermerriam: The logging statement ensures that we have all of the information about what endpoints we were trying to connect to (and what endpoint was trying to make the connection) which would not be present in the stacktrace otherwise.

Issue #1444, opened 2019-12-31
Need persistent storage for ENRs


gsalgado:
In order to fix #1299 we're probably going to need a persistent storage for ENRs retrieved as part of peer discovery. We need to figure out the requirements of said storage and then decide what backend we're going to use for it

pipermerriam:
I'd suggest something similar to the approach used in the `network-db` component, though there is some frustrating fragility in managing the database file itself.

jannikluhn:
I was thinking it would be nice to simply dump ENRs in a directory. This would make it very easy to manually share or compare them with other nodes.

Filename could be `f"enr_{encode_hex(node_id)}_{sequence_number}"`, content the standardized base64 representation.

cburgdorf:
>I was thinking it would be nice to simply dump ENRs in a directory.

I like the simplicity and flexibility of that approach.

PR #1461, opened 2020-01-08
File ENR DB

jannikluhn:
### What was wrong?

See #1444

### How was it fixed?

Implemented `FileEnrDb` which

- stores ENRs as files in a given directory
- ignores files with invalid contents or unexpected names
- reraises `IOError`s
- does not bother cleaning up old records
- has the same interface as the old `MemoryEnrDb`, except for a removed `remove` method

I'm not really sure anymore if the API is the right one or not. In particular, it doesn't allow getting ENRs by sequence number, it only ever provides the newest record for any given peer. But I think figuring this out is a separate issue.

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/enr_db.py
+++ b/p2p/discv5/enr_db.py
@@ -1,7 +1,18 @@
+import binascii
+from collections import defaultdict
 import logging
+import operator
+import pathlib
+import re
+import time
 from typing import (
+    DefaultDict,
     Dict,
 )
+
+from eth_utils import ValidationError
+
+import rlp
 
 import trio
 
@@ -11,6 +22,9 @@ from p2p.discv5.abc import EnrDbApi
 from p2p.discv5.enr import ENR
 from p2p.discv5.identity_schemes import IdentitySchemeRegistry
 from p2p.discv5.typing import NodeID
+
+
+ACCEPTABLE_ENR_LOAD_TIME = 1.0
 
 
 class BaseEnrDb(EnrDbApi):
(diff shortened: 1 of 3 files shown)


p2p/discv5/enr_db.py
```diff
@@ -39,6 +48,132 @@ def validate_identity_scheme(self, enr: ENR) -> None:
...
+
+
+def is_valid_enr_filename(filename: str) -> bool:
+    return bool(re.match(r"^enr_0x[0-9a-f]{64}_\d+$", filename))
+
+
+class FileEnrDb(BaseEnrDb):
+    def __init__(self,
+                 identity_scheme_registry: IdentitySchemeRegistry,
+                 directory: trio.Path) -> None:
```
pipermerriam: I would recommend just using `pathlib.Path`.  A conversation I had with the maintainer of trio suggested that the async file APIs trio exposes aren't the best and really just use threads under the hood to make things asynchronous.

jannikluhn: Good to know! Nice side effect is that we can load everything on disk in `__init__` already and don't need an additional async init to load stuff from disk.

p2p/discv5/enr_db.py
```diff
@@ -39,6 +49,133 @@ def validate_identity_scheme(self, enr: ENR) -> None:
...
+
+
+class FileEnrDb(BaseEnrDb):
+    def __init__(self,
+                 identity_scheme_registry: IdentitySchemeRegistry,
+                 directory: pathlib.Path) -> None:
+        super().__init__(identity_scheme_registry)
+        self.directory = directory
+        self.enrs: DefaultDict[NodeID, Dict[int, ENR]] = defaultdict(dict)
+        self.load_enrs()
```
gsalgado: I wonder how long this can take if we have a few thousand entries on disk.

p2p/discv5/enr_db.py
```diff
@@ -39,6 +49,133 @@ def validate_identity_scheme(self, enr: ENR) -> None:
...
+
+        valid_enr_paths = tuple(
+            path for path in enr_paths
+            if path not in invalid_enr_paths
+        )
+        for path in valid_enr_paths:
+            try:
+                enr = self.load_enr_file(path)
+            except (binascii.Error, rlp.DeserializationError, ValidationError):
+                self.logger.warning("Encountered invalid ENR in file {path}")
```
gsalgado: It'd be nice to include the exception msg in the log as well

p2p/discv5/enr_db.py
```diff
@@ -39,6 +49,133 @@ def validate_identity_scheme(self, enr: ENR) -> None:
...
+                continue
+
+            if get_enr_filename(enr) != path.name:
+                self.logger.warning(
+                    f"ENR in {path} has inconsistent name (actual node id: "
```
gsalgado: Looks like we prefer %-style formatting in logs: https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#formatting-log-messages

p2p/discv5/enr_db.py
```diff
@@ -39,6 +49,133 @@ def validate_identity_scheme(self, enr: ENR) -> None:
...
+                )
+
+            self.enrs[enr.node_id][enr.sequence_number] = enr
+
+    def load_enr_file(self, path: pathlib.Path) -> ENR:
+        try:
+            enr_base64 = path.read_text()
+        except IOError:
+            self.logger.exception("Failed to read ENR from disk")
+            raise
```
gsalgado: If we're propagating the exception there's not much point in logging it, right? But shouldn't you catch that in `load_enrs()` as well?

p2p/discv5/enr_db.py
```diff
@@ -39,6 +49,133 @@ def validate_identity_scheme(self, enr: ENR) -> None:
...
+                self.logger.warning("Encountered invalid ENR in file {path}")
+                continue
+
+            if get_enr_filename(enr) != path.name:
+                self.logger.warning(
+                    f"ENR in {path} has inconsistent name (actual node id: "
+                    f"{encode_hex(enr.node_id)}, actual seq num: {enr.sequence_number})"
+                )
+
+            self.enrs[enr.node_id][enr.sequence_number] = enr
```
gsalgado: So we'll be keeping all versions of a node's ENR we come across? Any reason for that as opposed to keeping just the latest?

p2p/discv5/enr_db.py
```diff
@@ -39,6 +49,133 @@ def validate_identity_scheme(self, enr: ENR) -> None:
...
+
+        enr = ENR.from_repr(enr_base64, self.identity_scheme_registry)
+        return enr
+
+    def write_enr_file(self, enr: ENR) -> None:
+        path = self.directory / get_enr_filename(enr)
+        try:
+            path.write_text(repr(enr))
+        except IOError:
+            self.logger.exception("Failed to write ENR to disk")
```
gsalgado: Again, probably not worth logging if we're propagating the exception, but it should probably be caught in places that call this method, no?

p2p/discv5/enr_db.py
```diff
@@ -39,6 +49,133 @@ def validate_identity_scheme(self, enr: ENR) -> None:
                 f"identity scheme registry"
             )
 
+    async def insert_or_update(self, enr: ENR) -> None:
+        try:
+            await self.update(enr)
+        except KeyError:
+            await self.insert(enr)
```
gsalgado: I wonder if it'd make sense to rename this to `update_or_insert()` given it first tries to update?

p2p/discv5/enr_db.py
```diff
@@ -39,6 +49,133 @@ def validate_identity_scheme(self, enr: ENR) -> None:
...
+            path.write_text(repr(enr))
+        except IOError:
+            self.logger.exception("Failed to write ENR to disk")
+            raise
+
+    async def insert(self, enr: ENR) -> None:
+        self.validate_identity_scheme(enr)
+
+        if await self.contains(enr.node_id):
+            raise ValueError("ENR with node id %s already exists", encode_hex(enr.node_id))
```
gsalgado: So, ISTM the assumption here is that this must never be called when we already have an ENR for a node and instead we should always use `insert_or_update()`? In that case, would it make sense to turn this into a private method and only expose `insert_or_update()`?

p2p/discv5/enr_db.py
```diff
@@ -39,6 +49,133 @@ def validate_identity_scheme(self, enr: ENR) -> None:
                 f"identity scheme registry"
             )
 
+    async def insert_or_update(self, enr: ENR) -> None:
+        try:
+            await self.update(enr)
+        except KeyError:
```
gsalgado: Instead of catching a KeyError here, which could actually be caused from something other than we not having an existing ENR with the same node ID (and is also wasteful because the first thing it does is validate the ENR), can't we use the `contains()` method to decide whether to insert or update?

p2p/discv5/enr_db.py
```diff
@@ -39,6 +49,133 @@ def validate_identity_scheme(self, enr: ENR) -> None:
...
+        enrs = self.enrs[node_id]
+        if not enrs:
+            raise KeyError(f"No ENR for node {encode_hex(node_id)} present in DB")
+        else:
+            # get enr with highest sequence number
+            _, enr = max(enrs.items(), key=operator.itemgetter(0))
+            return enr
+
+    async def contains(self, node_id: NodeID) -> bool:
+        await trio.hazmat.checkpoint()
```
gsalgado: Why do we yield control here and in `get()`, before doing anything?

p2p/discv5/enr_db.py
gsalgado: With 2k entries it's almost instantaneous, so no need to worry about it for now, I guess

jannikluhn: Nice, thanks for checking! My assumption was indeed that it wouldn't be too costly, especially considering that it would only run once during initialization. But maybe something to keep in mind, especially since ENRs will pile up over time.

p2p/discv5/enr_db.py
jannikluhn: No particular reason, just to stay flexible for now. One case I wasn't sure about for example is if we receive an updated ENR, but haven't verified its endpoint yet. We might want to keep the original ENR in this case, so that we can still reach the peer if it turns out the updated endpoint is wrong.

p2p/discv5/enr_db.py
jannikluhn: yes, I agree this would make sense. We could then even rename it to just `update`, because I can't think of a case in which the caller cares if there's already an older version or not. But since that's part of the API I didn't want to change it in this PR.

p2p/discv5/enr_db.py
jannikluhn: If we check with an `if`, there might be a race conditions, since someone else might insert just after we've checked but before we insert. So intuitively, catching the `KeyError` feels a bit safer.

I realize now though that the same can happen right now, so I guess the right way would be to do it the other way around (first try to insert, if that fails try to update)?

p2p/discv5/enr_db.py
jannikluhn: hmm, yes, maybe. But I didn't want to do any API changes in this PR.

p2p/discv5/enr_db.py
jannikluhn: Because `contains` is an async method, so it should contain a checkpoint. It's something the trio docs weakly recommend (see end of https://trio.readthedocs.io/en/stable/reference-core.html#checkpoints).

p2p/discv5/enr_db.py
pipermerriam: Maybe time the load here and issue a warning if it exceeds some reasonable lower bound for how long we think it's ok for this to block while data loads?

p2p/discv5/enr_db.py
```diff
@@ -39,6 +49,133 @@ def validate_identity_scheme(self, enr: ENR) -> None:
...
+            self.logger.exception("Failed to read ENR from disk")
+            raise
+
+        enr = ENR.from_repr(enr_base64, self.identity_scheme_registry)
+        return enr
+
+    def write_enr_file(self, enr: ENR) -> None:
+        path = self.directory / get_enr_filename(enr)
+        try:
+            path.write_text(repr(enr))
```
pipermerriam: I'm honestly not sure if this is required but I *think* that this can leave a partially written (corrupt) file if it gets interrupted mid-way.  It may not actually matter in this case but if you think we want/need atomic writes it is pretty easy to do so with this pattern:

https://github.com/ethpm/ethpm-cli/blob/ac8106464533efb9b8e5d3d2c8cfd9d72cb51d2b/ethpm_cli/_utils/filesystem.py#L50-L63

p2p/discv5/enr_db.py
gsalgado: fair enough

p2p/discv5/enr_db.py
gsalgado: I was under the impression no APIs here were meant to be safe from race conditions and that's why in discv4 I'm using a lock to serialize all access to the DB.

PR #1462, opened 2020-01-08
DiscoveryService now fetches ENRs for nodes it successfully bonds with

gsalgado:


p2p/discovery.py
```diff
@@ -204,6 +249,9 @@ def update_routing_table(self, node: NodeAPI) -> None:
             # with the least recently seen node on that bucket. If the bonding fails the node will
             # be removed from the bucket and a new one will be picked from the bucket's
             # replacement cache.
+            self.logger.debug2(
+                f"Routing table's bucket is full, couldn't add {node}. "
+                f"Checking if {eviction_candidate} is still responding, will evict if not")
```
pipermerriam: See this for why we always use `%s` formatting in logging messages:

https://github.com/ethereum/snake-charmers-tactical-manual/issues/19#issuecomment-442616682

p2p/discovery.py
```diff
@@ -196,6 +203,44 @@ def __init__(self,
...
+            async for (remote, enr_seq) in self.pending_enrs_consumer:
+                self.logger.debug2("Received request to fetch ENR for %s", remote)
+                self.manager.run_task(self._ensure_enr, remote, enr_seq)
+
+    async def _ensure_enr(self, node: NodeAPI, enr_seq: int) -> None:
+        # TODO: Check that we've recently bonded with the remote. For now it shouldn't be a
+        # problem as this is only triggered once we successfully bonded with a peer.
+        async with self._enr_db_lock:
+            try:
+                enr = await self._enr_db.get(cast(NodeID, node.id_bytes))
```
cburgdorf: *Nitpick*: `NodeID(node.id_bytes)` is slightly preferable because it checks the type. E.g. mypy will rightfully complain if you write `self._enr_db.get(NodeID("foo"))` whereas something like  `self._enr_db.get(cast(NodeID, "foo"))` will pass even though it will obviously blow up at runtime. In general, `cast` means: *"Step away `mypy`, take this piece of whatever and trust me that it is of `TypeX`".*

p2p/discovery.py
```diff
@@ -196,6 +203,44 @@ def __init__(self,
...
+            except KeyError:
+                enr = None
```
cburgdorf: How about we make :point_up: that `pass` and then add another `else` clause to the `except`?

```python
            except KeyError:
                pass
            else:
                if enr.sequence_number >= enr_seq:
                    self.logger.debug2("Already got latest ENR for %s", node)
                    return
```

I didn't find a hint in our styleguide but I think that is our preferred approach.

p2p/discovery.py
```diff
@@ -196,6 +203,44 @@ def __init__(self,
...
+
+        try:
+            await self.get_enr(node)
+        except CouldNotRetrieveENR as e:
+            self.logger.info("Failed to retrieve ENR for %s: %s", node, e)
+
+    async def report_stats(self) -> None:
+        async for _ in trio_utils.every(self._refresh_interval):
+            self.logger.debug("============================= Stats =======================")
+            full_buckets = [b for b in self.routing.buckets if b.is_full]
```
cburgdorf: We try to avoid single letter variables so I think this should be `bucket` instead of `b`

p2p/discovery.py
```diff
@@ -196,6 +203,44 @@ def __init__(self,
...
+            total_nodes = sum([len(b) for b in self.routing.buckets])
```
cburgdorf: same here and in the following code

p2p/discovery.py
gsalgado: Oh, that's good to know!

p2p/discovery.py
gsalgado: Sure, wfm

PR #1463, opened 2020-01-09
Add HTTP API Server and basic `beacon` endpoints

hwwhww:
### What was wrong?

Part of #1078, this PR added `ApiHandler(BaseHTTPHandler)`, `BaseResource`, and `Beacon(BaseResource)`.

### How was it fixed?

- `ApiHandler(BaseHTTPHandler)`: parse the requests and map them to corresponding the resource handler.
  - The URL parser is quite simple now. The request is case-sensitive. 
- `BaseResource`: resource is the domain of [RESTful HTTP API](https://github.com/ethereum/eth2.0-APIs)
  - classmethod name(cls): the lower-case name of the resource
  - route(self): the routing handler parses the request to the certain endpoint handler. e.g., `beacon/head` -> `Beacon.head(self, request)` function. 
- `Beacon(BaseResource)`
  - To display the API server, I added `beacon` APIs first as the [basic API spec](https://github.com/ethereum/eth2.0-APIs/blob/master/apis/beacon/basic.md). 

### To-Do

[//]: # "Stay ahead of things, add list items here!"

- [ ] Clean up commit history

[//]: # "For important changes that should go into the release notes please add a newsfragment file as explained here: https://github.com/ethereum/trinity/blob/master/newsfragments/README.md"

[//]: # "See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests"

#### Cute Animal Picture

cute-2500929_640

### DIFF
--- a/tests/components/eth2/apis/test_api_server.py
+++ b/tests/components/eth2/apis/test_api_server.py
@@ -108,6 +108,7 @@ async def test_restful_http_server(
         print('[ERROR]:', response.reason)
         raise
 
+    # The server may return 200 or 202 or others. 200 and 202 are both success.
     if str(status_code).startswith('2'):
         response_data = await response.json()
         print(f'[SUCCESS]: {request_path}: \t {response_data}\n')
--- a/tests/components/eth2/apis/test_api_server.py
+++ b/tests/components/eth2/apis/test_api_server.py
@@ -1,0 +1,115 @@
+import asyncio
+import pytest
+
+from aiohttp.test_utils import (
+    RawTestServer,
+    TestClient,
+)
+
+from eth2.beacon.types.attestations import Attestation
+from eth2.beacon.types.blocks import BeaconBlock
+from eth2.beacon.tools.factories import (
+    BeaconChainFactory,
+)
+from libp2p.crypto.secp256k1 import create_new_key_pair
+
+from trinity.protocol.bcc_libp2p.node import Node
+from trinity.http.handlers.api_handler import APIHandler
+
+
+GET_METHOD = 'GET'
+POST_METHOD = 'POST'
+
+
+@pytest.fixture()
+def chain(num_validators, base_db):
+    chain = BeaconChainFactory(num_validators=num_validators, base_db=base_db)
+    state_machine = chain.get_state_machine()
+    state = chain.get_head_state()
+    slot = 4
+    post_state = state_machine.state_transition.apply_state_transition(
+        state,
+        future_slot=slot,
+    )
+    chain.chaindb.persist_state(post_state)
+    chain.chaindb.update_head_state(post_state.slot, post_state.hash_tree_root)
+    return chain
+
+
+@pytest.fixture()
+async def libp2p_node(chain, event_bus):
+    key_pair = create_new_key_pair()
+    libp2p_node = Node(
+        key_pair=key_pair,
+        listen_ip="0.0.0.0",
+        listen_port=40000,
+        preferred_nodes=(),
+        chain=chain,
+        subnets=(),
+        event_bus=event_bus,
+    )
+    asyncio.ensure_future(libp2p_node.run())
+    await asyncio.sleep(0.01)
+    return libp2p_node
+
+
+@pytest.fixture
+async def http_server(chain, event_bus):
+    server = RawTestServer(APIHandler.handle(chain)(event_bus))
+    return server
+
+
+@pytest.fixture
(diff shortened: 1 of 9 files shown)


tests/components/eth2/apis/test_api_server.py
```diff
@@ -0,0 +1,115 @@
...
+    event_bus,
+    base_db,
+    method,
+    resource,
+    object,
+    json_data,
+    status_code,
+    num_validators,
+    chain,
+    libp2p_node,
```
hwwhww: `libp2p_node` will be used for `network` APIs.

tests/components/eth2/apis/test_api_server.py
```diff
@@ -0,0 +1,115 @@
...
+    request_path = resource + '/' + object
+    response = await http_client.request(method, request_path, json=json_data)
+
+    try:
+        assert response.status == status_code
+    except Exception:
+        print('[ERROR]:', response.reason)
+        raise
+
+    if str(status_code).startswith('2'):
```
hwwhww: The server may return `200` or `202` or others.

ChihChengLiang: Maybe explicitly add a comment?

trinity/http/api/resources/beacon.py
```diff
@@ -0,0 +1,49 @@
...
+    @get_method
+    async def state(self, request: web.Request) -> Dict[str, Any]:
+        if 'slot' in request.query:
+            slot = Slot(int(request.query['slot']))
+            state = self.chain.get_state_by_slot(slot)
+            state_class = self.chain.get_state_machine(slot).state_class
+        elif 'root' in request.query:
+            root = cast(HashTreeRoot, decode_hex(request.query['root']))
+            state = self.chain.get_state_by_root(root)
+            state_class = self.chain.get_head_state().__class__  # TODO: It's a simple workaround
```
ChihChengLiang: Question: is it possible to do `state_class = self.chain.get_state_machine(state.slot).state_class`?

trinity/http/api/resources/beacon.py
```diff
@@ -0,0 +1,49 @@
...
+        else:
+            raise APIServerError(f"Wrong querystring: {request.query}")
+
+        return to_formatted_dict(state, sedes=state_class)
```
ChihChengLiang: If state is already a serializable, do we need to pass sedes class here?

trinity/components/eth2/beacon/component.py
```diff
@@ -200,16 +212,24 @@ def _load_or_create_node_key(cls, boot_info: BootInfo) -> KeyPair:
...
             services: Tuple[BaseService, ...] = (
                 libp2p_node, receive_server, slot_ticker, validator, syncer
             )
+
             if boot_info.args.enable_metrics:
-                services += (http_server,)
+                services += (metrics_server,)
+
+            if boot_info.args.enable_api:
+                services += (api_server,)
```
ralexstokes: one thing to keep in mind w/ this general pattern is that resources like this could involve a substantial amount of setup/teardown code to build...

to avoid paying for things we don't need, i'd move the construction of these objects under the conditional that checks if we even need them... i don't think its super important here tbh but just mentioning it... may be helpful to keep in mind moving forward

trinity/http/api/resources/base.py
```diff
@@ -0,0 +1,55 @@
...
+
+    return wrapper
+
+
+def post_method(
+    func: Callable[[TBaseResource, web.Request], Any]
+) -> Callable[[TBaseResource, web.Request], Any]:
+    async def wrapper(self: TBaseResource, request: web.Request) -> Any:
+        if request.method != "POST":
+            raise APIServerError(f"Wrong HTTP method, should be GET, got {request.method}")
```
ralexstokes:
```suggestion
            raise APIServerError(f"Wrong HTTP method, should be POST, got {request.method}")
```

trinity/http/api/resources/beacon.py
ralexstokes: yeah -- can `py-ssz` introspect the sedes based on the class of the object? i feel like it can

ralexstokes:
another question: did we need the `event_bus` to be passed into the API server?

trinity/http/api/resources/beacon.py
hwwhww: Hmm, it requires the `root` -> `slot` mapping.

PR #1467, opened 2020-01-09
Update beam exec and preview plugins to use async-service

pipermerriam:
### What was wrong?

The beam sync "preview" and "exec" plugins were using the old service API

### How was it fixed?

Updated to use the new `async-service` API.

#### Cute Animal Picture

ewechair0306_468x492

### DIFF
--- a/trinity/extensibility/asyncio.py
+++ b/trinity/extensibility/asyncio.py
@@ -48,7 +48,6 @@ class AsyncioIsolatedComponent(BaseIsolatedComponent):
                         proc.pid,
                     )
                     proc.send_signal(signal.SIGTERM)
-                    pass
                 finally:
                     raise err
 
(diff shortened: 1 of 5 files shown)


pipermerriam:
still appear to have some test failures I need to sort out before this is ready for review.

trinity/extensibility/asyncio.py
```diff
@@ -48,8 +48,6 @@ def get_subprocess_kwargs(self) -> Optional[SubprocessKwargs]:
                         proc.pid,
                     )
                     proc.send_signal(signal.SIGTERM)
-                    pass
-                finally:
```
carver: Not re-raising `CancelledError` is unusual enough that it's worth commenting on why we shouldn't, IMO.

pipermerriam: Yeah, I'm puzzling over the *right* thing to do here, I'll give it one more pass.  This was originally to cleanup a place where the cancelled error was causing the `asyncio-run-in-process` library to log a warning which was noisy but I think the real problem needs to be fixed differently now that I've looked at it again.

pipermerriam: Ended up reverting this.  I think the thing it was fixing was benign and looking at the `asyncio-run-in-process` it looks like `CancelledError` should be handled fine there.

PR #1468, opened 2020-01-09
Discv5 Test Vectors

jannikluhn:
### What was wrong?

[There](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire-test-vectors.md) are official discv5 test vectors for encoding and crypto by now, which we're not using yet.

### How was it fixed?

Add tests using *some* of the test vectors:

- all packet encoding
- ping, pong, getnodes, and nodes messages
- ecdh and key derivation
- encryption

But not:

- the remaining message types (as both test vectors and ourselves are behind the current spec version)
- nonce signing tests (as we're behind the current spec version)
- authentication header construction test (includes nonce signing, so we can't do this at the moment, and I have the feeling we're doing something wrong there as well)

All tests except for the ECDH tests are passing, which I'm still trying to fix.

### To-Do

- [ ] Clean up commit history
- [x] Fix ECDH test

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,7 @@ deps = {
         "asyncio-cancel-token>=0.2,<0.3",
         "async_lru>=0.1.0,<1.0.0",
         "cached-property>=1.5.1,<2",
+        "coincurve>=10.0.0,<11.0.0",
         # cryptography does not use semver and allows breaking changes within `0.3` version bumps.
         "cryptography>=2.5,<2.7",
         "eth-hash>=0.1.4,<1",
--- a/tests/p2p/discv5/test_messages.py
+++ b/tests/p2p/discv5/test_messages.py
@@ -1,15 +1,25 @@
 import inspect
 
+from eth_utils import (
+    decode_hex,
+)
+
+import pytest
+
 import rlp
 from rlp.sedes import (
     big_endian_int,
 )
 
 from p2p.discv5 import messages
+from p2p.discv5.enr import ENR
 from p2p.discv5.messages import (
     default_message_type_registry,
     BaseMessage,
     PingMessage,
+    PongMessage,
+    NodesMessage,
+    FindNodeMessage,
 )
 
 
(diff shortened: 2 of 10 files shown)


jannikluhn:
I've added the remaining tests, excluding the message type tests, so this is ready for review.

Importantly, I also had to make some changes to the code due to recent changes to the spec:

- Update the signature format to include the ephemeral pubkey and a prefix: https://github.com/ethereum/trinity/pull/1468/commits/ba3d32a06aee92ca7665c515e8f3f1a4f5e66ea9
- Remove the auth header from from the authenticated data part in encryption: https://github.com/ethereum/trinity/pull/1468/commits/f3b87d1d7f00df2aa73b4c97a90bd17cc124be98
- [Potentially controversial] Use coincurve directly to do ECDH: https://github.com/ethereum/trinity/pull/1468/commits/89336855dd2a403fe732505182b1aa89aeb263a0 (see https://github.com/ethereum/devp2p/issues/134)

setup.py
```diff
@@ -16,6 +16,7 @@
         "asyncio-cancel-token>=0.2,<0.3",
         "async_lru>=0.1.0,<1.0.0",
         "cached-property>=1.5.1,<2",
+        "coincurve>=10.0.0,<11.0.0",
```
ChihChengLiang: Nitpick: `coincurve` is also referenced in `trinity` dep (line 37). May be we can remove the line 37 one, because coincurve is only imported in `p2p` module.

tests/p2p/discv5/test_packet_preparation.py
```diff
@@ -220,3 +226,167 @@ def test_auth_tag_packet_preparation(tag, auth_tag, key):
...
+    [
+        "auth_tag",
+        "id_nonce",
+        "ephemeral_public_key",
+        "auth_cipher_text",
+        "auth_header_rlp",
+    ],
+    [
+        [
+            decode_hex("0x27b5af763c446acd2749fe8e"),
```
ChihChengLiang: We should comment here something like `# official test vectors from the ethereum/devp2p repository` too. Preferably the link `https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire-test-vectors.md`

tests/p2p/discv5/test_messages.py
```diff
@@ -42,3 +52,76 @@ def test_message_encoding():
...
+        decode_hex("0x03c401820100"),
+    ],
+    [
+        NodesMessage,
+        {
+            "request_id": 0x01,
+            "total": 0x01,
+            "enrs": [],
+        },
+        decode_hex("04c30101c0"),
```
ChihChengLiang: nitpick add `0x` for consistency.

setup.py
jannikluhn: I think coincurve is used as a backend for eth-keys. So it's not explicitly imported, but still used.

PR #1469, opened 2020-01-10
DiscoveryService now supports a way to add extra fields to local ENR

gsalgado:
This allows us to add our ForkID to our ENR so that peers can decide whether or not to connect with us.

Also implement versioning of the local ENR so that a new one is generated with a higher sequence number every time anything in our ENR changes.

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -13,7 +13,6 @@ from typing import (
     AsyncIterator,
     Awaitable,
     Callable,
-    cast,
     Dict,
     Generic,
     Iterable,
@@ -24,8 +23,11 @@ from typing import (
     Sequence,
     Set,
     Tuple,
+    Type,
     TypeVar,
+    TYPE_CHECKING,
 )
+from typing_extensions import Literal
 
 from async_generator import aclosing
 
@@ -44,7 +46,7 @@ from lahja import (
 import rlp
 from rlp.exceptions import DeserializationError
 
-from eth_typing import Hash32
+from eth_typing import BlockNumber, Hash32
 
 from eth_utils import (
     encode_hex,
@@ -64,8 +66,11 @@ from eth_hash.auto import keccak
 
 from async_service import Service
 
+from eth.abc import VirtualMachineAPI
+from eth.constants import GENESIS_BLOCK_NUMBER
+
 from p2p import constants
-from p2p.abc import AddressAPI, NodeAPI
+from p2p.abc import AddressAPI, ENR_FieldProvider, NodeAPI
 from p2p.discv5.enr import ENR, UnsignedENR, IDENTITY_SCHEME_ENR_KEY
 from p2p.discv5.enr_db import MemoryEnrDb
 from p2p.discv5.identity_schemes import default_identity_scheme_registry, V4IdentityScheme
@@ -82,8 +87,12 @@ from p2p.events import (
     PeerCandidatesResponse,
 )
 from p2p.exceptions import AlreadyWaitingDiscoveryResponse, CouldNotRetrieveENR, NoEligibleNodes
+from p2p.forkid import ForkID, make_forkid
 from p2p.kademlia import Address, Node, RoutingTable, check_relayed_addr, sort_by_distance
 from p2p import trio_utils
+
+if TYPE_CHECKING:
+    from trinity.db.eth1.header import BaseAsyncHeaderDB
 
 
 # V4 handler are async methods that take a Node, payload and msg_hash as arguments.
@@ -132,6 +141,7 @@ class DiscoveryService(Service):
     # Maximum number of ENR retrieval requests active at any moment. Need to be a relatively high
     # number as during a lookup() we'll bond and fetch ENRs of many nodes.
     _max_pending_enrs: int = 20
+    _local_enr_refresh_interval: int = 60
 
     logger = get_extended_debug_logger('p2p.discovery.DiscoveryService')
 
@@ -140,7 +150,9 @@ class DiscoveryService(Service):
                  address: AddressAPI,
                  bootstrap_nodes: Sequence[NodeAPI],
                  event_bus: EndpointAPI,
-                 socket: trio.socket.SocketType) -> None:
+                 socket: trio.socket.SocketType,
+                 enr_field_providers: Sequence[ENR_FieldProvider] = tuple(),
+                 ) -> None:
         self.privkey = privkey
         self.address = address
         self.bootstrap_nodes = bootstrap_nodes
@@ -151,10 +163,14 @@ class DiscoveryService(Service):
         self.pong_channels = ExpectedResponseChannels[Tuple[Hash32, int]]()
         self.neighbours_channels = ExpectedResponseChannels[List[NodeAPI]]()
         self.ping_channels = ExpectedResponseChannels[None]()
+        self.enr_field_providers = enr_field_providers
         # FIXME: Use a persistent EnrDb implementation.
         self._enr_db = MemoryEnrDb(default_identity_scheme_registry)
         # FIXME: Use a concurrency-safe EnrDb implementation.
         self._enr_db_lock = trio.Lock()
+        self._local_enr: ENR = None
+        self._local_enr_next_refresh: float = time.monotonic()
+        self._local_enr_lock = trio.Lock()
         self.parity_pong_tokens: Dict[Hash32, Hash32] = {}
         if socket.family != trio.socket.AF_INET:
             raise ValueError("Invalid socket family")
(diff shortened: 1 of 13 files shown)


trinity/components/builtin/peer_discovery/component.py
```diff
@@ -2,57 +2,53 @@
...
 
-    eth1_config = trinity_config.get_app_config(Eth1AppConfig)
-    if eth1_config.database_mode is Eth1DbMode.LIGHT:
-        return LESProtocolV2
-    else:
-        return ETHProtocol
+async def generate_enr_forkid(
+        vm_config: Tuple[Tuple[BlockNumber, Type[VirtualMachineAPI]], ...],
+        headerdb: BaseAsyncHeaderDB
+) -> Tuple[bytes, bytes]:
```
pipermerriam: Minor improvement but I think you might be able to use `Literal[b'eth']` as the type of the first parameter but I'm not sure if it'll play nicely with the `ENRFieldProvider`...

trinity/components/builtin/peer_discovery/component.py
```diff
@@ -2,57 +2,53 @@
...
+    head = await headerdb.coro_get_canonical_head()
+    genesis_hash = await headerdb.coro_get_canonical_block_hash(
+        BlockNumber(GENESIS_BLOCK_NUMBER))
```
pipermerriam: You might check whether this wrapping is necessary, I think this was fixed upstream recently.

trinity/components/builtin/peer_discovery/component.py
```diff
@@ -88,20 +85,25 @@ def configure_parser(cls,
...
             await socket.bind((external_ip, config.port))
             discovery_service = PreferredNodeDiscoveryService(
-                boot_info.trinity_config.nodekey,
+                config.nodekey,
                 address,
                 config.bootstrap_nodes,
                 config.preferred_nodes,
                 event_bus,
                 socket,
+                [enr_forkid_provider],
```
pipermerriam: *nitpick* generally always prefer tuples over lists unless we have an explicit reason to need mutability.

pipermerriam:
Looks clean :+1:

p2p/forkid.py
```diff
@@ -46,7 +49,7 @@ def make_forkid(
         vm_config: Tuple[Tuple[BlockNumber, Type[VirtualMachineAPI]], ...]
 ) -> ForkID:
     fork_blocks = _extract_fork_blocks(vm_config)
-    pre_hash_bytes = to_bytes(hexstr=genesis_hash)
+    pre_hash_bytes = bytes(genesis_hash)
```
gsalgado: The function's signature specifies that `genesis_hash` is of type `Hash32` but that was not caught by mypy because this function is only used in a test

gsalgado: And we need to convert it into a `bytes` because further down we append another `bytes` variable to it

pipermerriam: `Hash32` is just a fancy wrapper around `bytes` so I think that the cast isn't even needed.

p2p/forkid.py
```diff
@@ -71,7 +74,7 @@ def validate_forkid(
     """
 
     fork_blocks = list(_extract_fork_blocks(vm_config))
-    checksums = [binascii.crc32(to_bytes(hexstr=genesis_hash))]
+    checksums = [binascii.crc32(genesis_hash)]
     for block_number in fork_blocks:
         block_number_as_bytes = block_number.to_bytes(8, 'big')
         checksums.append(binascii.crc32(block_number_as_bytes, checksums[-1]))
```
pipermerriam: FYI, I believe that this pattern is what `toolz.accumulate` is for.  I think this is one of the more complex functional tools but it is quite nice once you internalize how it works.

https://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.accumulate

Here is a really rough translation.

```python
block_numbers = tuple(block_number.to_bytes(8, 'big') for block_number in fork_blocks)
checksums = accumulate(binascii.crc32, block_numbers, initial=binascii.crc32(genesis_hash))
```

I'm not 100% sure that is equivalent but it should be close and a reasonably close approximation of how `accumulate` would be used.

scripts/discovery.py
```diff
@@ -35,6 +44,14 @@
...
     logging.getLogger('p2p').setLevel(log_level)
 
+    if args.mainnet:
+        default_bootnodes = constants.MAINNET_BOOTNODES
+        vm_config = MAINNET_VM_CONFIGURATION
+        genesis = MAINNET_GENESIS_HEADER
+    else:
+        default_bootnodes = constants.ROPSTEN_BOOTNODES  # type: ignore
+        vm_config = ROPSTEN_VM_CONFIGURATION
+        genesis = ROPSTEN_GENESIS_HEADER
```
pipermerriam: Dunno if it makes sense to do this but `trinity.network_configurations` has a data structure that attempts to aggregate some of this data in a more generic way.  Maybe it makes sense to go ahead and extend that with the `vm_config` and genesis file and then you could just use it here.

p2p/forkid.py
gsalgado: That doesn't work because crc32 expects a `bytes` argument and returns an int and the returned value is then re-fed into it in subsequent calls via its second argument. I believe there might be a way to make it work, but then I think we'd be sacrificing readability a lot for no real benefit

pipermerriam: :+1:

scripts/discovery.py
gsalgado: I think it's a good idea

p2p/forkid.py
gsalgado: without the cast, mypy complains when I append an actual `bytes` to it

gsalgado:
@pipermerriam you'd approved this when it was still a draft (a couple days ago) and since then I also added the versioning of the local ENR, so not sure if you're happy for me to merge or if you want to do another pass?

p2p/discovery.py
```diff
@@ -140,7 +150,9 @@ def __init__(self,
                  address: AddressAPI,
                  bootstrap_nodes: Sequence[NodeAPI],
                  event_bus: EndpointAPI,
-                 socket: trio.socket.SocketType) -> None:
+                 socket: trio.socket.SocketType,
+                 enr_field_providers: Optional[Sequence[ENR_FieldProvider]] = tuple(),
```
pipermerriam: I think the `Optional` can be removed from this type since there doesn't seem to be a good reason to allow a `None` value here.

p2p/discovery.py
```diff
@@ -330,31 +349,79 @@ def schedule_enr_retrieval(self, node: NodeAPI, enr_seq: int) -> None:
...
+        kv_pairs = {
+            IDENTITY_SCHEME_ENR_KEY: V4IdentityScheme.id,
+            V4IdentityScheme.public_key_enr_key: self.pubkey.to_compressed_bytes(),
+            IP_V4_ADDRESS_ENR_KEY: self.address.ip_packed,
+            UDP_PORT_ENR_KEY: self.address.udp_port,
+            TCP_PORT_ENR_KEY: self.address.tcp_port,
+        }
+        for field_provider in self.enr_field_providers:
+            key, value = await field_provider()
+            kv_pairs[key] = value
```
pipermerriam: Probably worth checking for key collisions and at least logging them here if a `key` is already present in the mapping.

p2p/discovery.py
```diff
@@ -330,31 +349,79 @@ def schedule_enr_retrieval(self, node: NodeAPI, enr_seq: int) -> None:
...
+
+        # Re-generate our ENR from scratch and compare it to the current one to see if we
+        # must create a new one with a higher sequence number.
+        current_enr = await self._generate_local_enr(self._local_enr.sequence_number)
+        if current_enr == self._local_enr:
+            return self._local_enr
+
+        # Either our node's details (e.g. IP address) or one of our ENR fields have changed, so
+        # generate a new one with a higher sequence number.
+        self._local_enr = await self._generate_local_enr(self._local_enr.sequence_number + 1)
```
pipermerriam: Maybe worth a logging statement here too to record that our ENR changed.

tests-trio/p2p-trio/conftest.py
```diff
@@ -40,8 +40,15 @@
         bootstrap_nodes=[],
         event_bus=None,
         socket=socket)
-    nursery.start_soon(TrioManager.run_service, discovery)
-    await discovery.ready_to_drive.wait()
+    manager = TrioManager(discovery)
+    nursery.start_soon(manager.run)
+    await manager.wait_started()
```
pipermerriam: These three lines of card are pretty much identical to:

```python
async with background_trio_service(discovery) as manager:
    ...
```

Any reason for the manual manager instantiation adn starting?

trinity/components/builtin/peer_discovery/component.py
```diff
@@ -88,20 +66,25 @@ def configure_parser(cls,
...
                 config.preferred_nodes,
                 event_bus,
                 socket,
+                (eth_cap_provider,),
             )
 
         try:
-            await TrioManager.run_service(discovery_service)
+            with db:
+                await TrioManager.run_service(discovery_service)
```
pipermerriam: This *could* be updated to just use `await async_service.run_trio_service` which is just an alias to `TrioManager.run_service` but I think it reads slightly better.

p2p/discovery.py
gsalgado: Good point, but I think I prefer an exception

p2p/discovery.py
gsalgado: Indeed

tests-trio/p2p-trio/conftest.py
gsalgado: Not really, changed it to use background_trio_service(), which also simplified the fixtures quite a bit!

PR #1470, opened 2020-01-10
Remove invalid use_develop option for wheel-cli tests

kclowes:
### What was wrong?
When I was updating the web3 script that I stole to do the same thing, I noticed tox doesn't have a use_develop configuration flag. The actual flag doesn't have an underscore (`usedevelop`) and it's default is false anyway, so I don't think you need it. Docs [here](https://tox.readthedocs.io/en/stable/config.html#conf-usedevelop) :)

### How was it fixed?
Removed it!

### To-Do

- [x] Clean up commit history

- Probably doesn't deserve a newsfragment, but let me know if you feel differently!

#### Cute Animal Picture
image

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -109,14 +109,12 @@ deps = {[common-wheel-cli]deps}
 whitelist_externals = {[common-wheel-cli]whitelist_externals}
 commands = {[common-wheel-cli]commands}
 skip_install=true
-use_develop=false
 
 [testenv:py37-wheel-cli]
 deps = {[common-wheel-cli]deps}
 whitelist_externals = {[common-wheel-cli]whitelist_externals}
 commands = {[common-wheel-cli]commands}
 skip_install=true
-use_develop=false
 
 [common-integration]
 deps = .[p2p,trinity,eth2,test,test-asyncio]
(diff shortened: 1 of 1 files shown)


kclowes:
I think the failing test is flaky because the py37-wheel-cli passed but if I just made it flaky, let me know. For some reason circle won't let me re-run the workflow ¯\_(ツ)_/¯

tox.ini
```diff
@@ -109,14 +109,12 @@ deps = {[common-wheel-cli]deps}
 whitelist_externals = {[common-wheel-cli]whitelist_externals}
 commands = {[common-wheel-cli]commands}
 skip_install=true
-use_develop=false
```
pipermerriam: Looking under the main `[testenv]` we do indeed have `usedevelop=True` there.  What I'm unsure about is whether we need it here as it isn't clear whether this environment inherits that setting (I am inclined to think that it does)...

kclowes: ohh, shoot. Didn't check that. I'm pretty sure it inherits too. I'll update to usedevelop.

kclowes:
@pipermerriam I am not sure why the 3.7 test is passing while the 3.6 test isn't. Also, looks like there are different failures, depending on the run. It looks like `test_shutdown_does_not_throw_errors` is pretty consistently `XPASS`ing, but I'm seeing failures sometimes on `test_does_not_throw_errors_on_short_run` and other times on `test_web3_commands_via_attached_console`. If you don't have any ideas off the top of your head, I may close this PR and raise an issue instead, since I don't think this is very high priority for anyone.

pipermerriam:
our wheel tests are flaky so I typically just re-run them unless they are both failing.

Issue #1443, opened 2019-12-31
Keep track of peers we've bonded, and when we did so


gsalgado:
Currently, when we bond of a peer as part of discovery, we try to add that peer to our routing table, but if a bucket is full the peer may never actually make it to our routing table. That is fine and works as expected, but we have some places in `DiscoveryService` that rely on the routing table to detect peers that we've bonded with and decide whether or not to accept ENR/FIND_NODE requests from them. That is obviously wrong; we should instead use something other than the routing table to keep track of nodes we've bonded before, and when we did so.

That would also allow us to avoid re-bonding with nodes that we've already bonded but never made it to the routing table

PR #1473, opened 2020-01-14
Use FileEnrDB in DiscoveryService

gsalgado:


scripts/discovery.py
```diff
@@ -51,6 +52,10 @@
         log_level = getattr(logging, args.l.upper())
     logging.getLogger('p2p').setLevel(log_level)
 
+    enr_db_path = Path(args.enrdb)
+    if not enr_db_path.exists():
+        enr_db_path.mkdir()
```
jannikluhn: maybe

```Py
enr_db_path.mkdir(parents=True, exists_ok=True)
```

p2p/discovery.py
```diff
@@ -151,6 +152,7 @@ def __init__(self,
                  bootstrap_nodes: Sequence[NodeAPI],
                  event_bus: EndpointAPI,
                  socket: trio.socket.SocketType,
+                 enr_db_path: pathlib.Path,
```
pipermerriam: I don't think this should take a path as that ties it to the file-based implementation of the EnrDB.  Rather, I would suggest defining an `ABC` base class `EnrDBAPI` that defines the interface and having this just be an instantiated `enr_db`.  This ensures that `mypy` will enforce only using the public API for the `enr_db`, ensuring that `DiscoveryService` remains agnostic about which implementation of the database it uses and also allows for testing scenarios to keep using the memory based database implemetnation.

scripts/discovery.py
```diff
@@ -40,6 +40,7 @@
         help="1 for mainnet, 3 for testnet"
     )
     parser.add_argument('-l', type=str, help="Log level", default="info")
+    parser.add_argument('-enrdb', type=str, help="Path to ENR database", default="/tmp/enrs")
```
pipermerriam: I think rather than hard coding this path you should use the `tempfile` standard library APIs so that the temporary directory is cleaned up after each run and so you don't leak data across multiple runs.

jannikluhn:
```suggestion
    parser.add_argument('--enrdb', type=str, help="Path to ENR database", default="/tmp/enrs")
```

Even more nitpicky: Looking through the existing Trinity options, the naming convention seems to prefer dashes to separate words (so `--enr-db`).

gsalgado:
So, we need to persist more data about peers (e.g. [last bond time](#1443)), and it'd make sense to have a single DB with all peer info instead of separate ones. Do you guys think it'd be ok to rename/extend `abc.EnrDbApi` into something like `abc.NodeDbApi` where we'd store ENRs as well as other node details?

jannikluhn:
Originally I had a separate class in mind (maybe with a common wrapper that provides a unified interface to both ENR and peer db). But thinking about it this is probably unnecessary. So I would say :+1: to rename and extend. Not sure though if we want to keep a separate interface for ENRs (probably not?).

scripts/discovery.py
gsalgado: The thing is that most of the time I do want to persist the ENRs across runs, and having a default was just a way to simplify things when I run the script. It's also tricky to make it so that we cleanup the ENRs when no enrdb arg is provided but to keep them when the arg is present, so for now I'll just make the argument required and never delete ENRs

gsalgado: Well, this is just a helper script and all other arguments here use single dash, so I'd rather stick to that for now

gsalgado:
> Originally I had a separate class in mind (maybe with a common wrapper that provides a unified interface to both ENR and peer db). But thinking about it this is probably unnecessary. So I would say +1 to rename and extend. Not sure though if we want to keep a separate interface for ENRs (probably not?).

I'm tempted to first try to keep the existing `Node` APIs but use an ENR internally, but better to discuss that in #1443. If you guys are happy I'd like to merge this while I'm working on another PR that will add a compatibility shim so that we can create/keep ENRs for all v4 nodes and not just the ones supporting the extension

PR #1475, opened 2020-01-14
Improve UX of some JSON-RPC endpoints

### LINKED ISSUE #1016: Implement mining JSON-RPCs
There are a few RPCs which Trinity is unlikely to fully support soon, but for which it might be a better UX to return error messages than do nothing:

- [`eth_getWork`](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getwork)
- [`eth_submitWork`](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_submitwork)
- [`eth_submitHashRate`](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_submithashrate)

Related to #20 and #26.

cburgdorf:
### What was wrong?

1. When a JSON-RPC API was called with the wrong number of parameters, the error wasn't as clear as it could be.

2. We currently return a generic not implemented error for `eth_getWork`, `eth_submitWork` and `eth_submitHashrate`. A more friendlier UX would be to hint that this is intended as Trinity does not support mining. This fixes #1016 

### How was it fixed?

1. Catch `TypeError` and suggest to check the count and type of the parameter and then concat the original exception.

2. Implement `eth_getWork`, `eth_submitWork` and `eth_submitHashrate` but raise `NotImplemtedError` with the message `Trinity does not support mining`.

3. Change existing `eth_hashrate` and `eth_coinbase` to also return errors instead of default values to be aligned with the rest of the mining related APIs.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1475.feature.rst
+++ b/newsfragments/1475.feature.rst
@@ -2,3 +2,5 @@ Improve error message for when the parameters to JSON-RPC calls are invalid
 
 Improve error message for ``eth_getWork``, ``eth_submitWork`` and ``eth_submitHashrate``
 APIs which Trinity does not support as it generally does not support mining operations.
+
+Change ``eth_hashrate`` and ``eth_coinbase`` to also return errors instead of default values.
--- a/newsfragments/1475.feature.rst
+++ b/newsfragments/1475.feature.rst
@@ -1,0 +1,1 @@
+Improve error message for when the parameters to JSON-RPC calls are invalid
--- a/newsfragments/1475.feature.rst
+++ b/newsfragments/1475.feature.rst
@@ -1,1 +1,4 @@
 Improve error message for when the parameters to JSON-RPC calls are invalid
+
+Improve error message for ``eth_getWork``, ``eth_submitWork`` and ``eth_submitHashrate``
+APIs which Trinity does not support as it generally does not support mining operations.
--- a/trinity/rpc/modules/eth.py
+++ b/trinity/rpc/modules/eth.py
@@ -8,6 +8,7 @@ from typing import (
     cast,
     Dict,
     List,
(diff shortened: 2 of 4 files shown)


newsfragments/1475.feature.rst
```diff
@@ -0,0 +1,6 @@
+Improve error message for when the parameters to JSON-RPC calls are invalid
+
+Improve error message for ``eth_getWork``, ``eth_submitWork`` and ``eth_submitHashrate``
+APIs which Trinity does not support as it generally does not support mining operations.
+
+Change ``eth_hashrate`` and ``eth_coinbase`` to also return errors instead of default values.
```
veox: AFAIR `eth_coinbase` is used in `ethstats` reporting?..

trinity/rpc/modules/eth.py
```diff
@@ -166,9 +167,7 @@ class Eth(Eth1ChainRPCModule):
         return encode_hex(result)
 
     async def coinbase(self) -> str:
-        # Trinity doesn't support mining yet and hence coinbase_address is default (ZERO_ADDRESS)
-        coinbase_address = ZERO_ADDRESS
-        return encode_hex(coinbase_address)
+        raise NotImplementedError("Trinity does not support mining")
```
veox: This will likely break when `--ethstats` is enabled.

cburgdorf: Thanks for checking. If it did, I'd argue we might be better off putting the defaults in the ethstats component but keeping consistency in returning the same error for all non implemented mining APIs. But I also double checked it and yeah it doesn't call into these APIs.

PR #1479, opened 2020-01-16
DiscV5 Component

jannikluhn:
### What was wrong?

A Trio component for discv5.

The PR builds on top of #1468, only the last two commits are new.

It successfully connects to https://github.com/sigp/rust-libp2p/tree/discv5. There is one [fix](https://github.com/ethereum/trinity/commit/1cb1438b9399040c0b728daa08b8e9d38d10c7bc) I had to apply to make it work, but in my opinion this is an inconsistency in the spec, so we will probably be able to revert it in the future.

There's still some things missing (most importantly: responding to find node requests properly, exposing peer information to other components, CLI parameter validation, NAT stuff), but going from here should be much easier now.

The component provides the following CLI options:

- `--enr-dir`: Path to the directory in which ENRs are stored (default `<data-dir>/enrs`)
- `--discovery-private-key`: The node's private key (default random)
- `--discovery-boot-enrs`: list of ENRs of peers to populate the ENR DB and the routing table with
- `--discovery-port`: the port

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/handshake.py
+++ b/p2p/discv5/handshake.py
@@ -9,6 +9,8 @@ from eth_utils import (
     ValidationError,
 )
 
+from eth_keys.datatypes import PublicKey
+
 from p2p.exceptions import (
     DecryptionError,
     HandshakeFailure,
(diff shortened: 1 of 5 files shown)


trinity/components/eth2/discv5/component.py
```diff
@@ -0,0 +1,266 @@
...
+            await enr_db.insert_or_update(enr)
+            routing_table.add(enr.node_id)
+
+        port = boot_info.args.discovery_port
+
+        socket = trio.socket.socket(
+            family=trio.socket.AF_INET,
+            type=trio.socket.SOCK_DGRAM,
+        )
+        outgoing_datagram_channels: ChannelPair[OutgoingDatagram] = trio.open_memory_channel(0)
```
pipermerriam: You can also do this as `send_chan, recv_chan = trio.open_memory_channel[OutgoingDatagram](0)`

p2p/discv5/handshake.py
```diff
@@ -132,9 +134,10 @@ def complete_handshake(self, response_packet: Packet) -> HandshakeResult:
             ephemeral_public_key,
         ) = self.identity_scheme.create_handshake_key_pair()
 
+        remote_public_key = PublicKey.from_compressed_bytes(self.remote_enr.public_key).to_bytes()
```
ChihChengLiang: nitpick: call it "remote_public_key_uncompressed" or add a comment to emphasize the intention of this line?

PR #1481, opened 2020-01-17
DiscV5 FindNode response

jannikluhn:
### What was wrong?

We only responded to the special case distance=0 (i.e. requesting our own ENR). The typical FindNode request with some non-zero distance was ignored.

### How was it fixed?

Plug in the kademlia routing table implementation in favor of the flat one and extend the FindNode handler. In addition, we computed the log distance between two node ids in a different way than other implementations, so I fixed this (even though I still think our implementation makes more sense).

Only the last three commits are new (extends #1468 #1479).

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/routing_table.py
+++ b/p2p/discv5/routing_table.py
@@ -85,7 +85,7 @@ def compute_log_distance(left_node_id: NodeID, right_node_id: NodeID) -> int:
     if left_node_id == right_node_id:
         raise ValueError("Cannot compute log distance between identical nodes")
     distance = compute_distance(left_node_id, right_node_id)
-    return distance.bit_length() - 1
+    return distance.bit_length()
 
 
 class KademliaRoutingTable:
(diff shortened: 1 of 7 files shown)


p2p/discv5/routing_table.py
```diff
@@ -231,11 +231,13 @@ def remove(self, node_id: NodeID) -> None:
 
     def get_nodes_at_log_distance(self, log_distance: int) -> Tuple[NodeID, ...]:
         """Get all nodes in the routing table at the given log distance to the center."""
-        if log_distance < 0:
-            raise ValueError("Log distance must not be negative")
-        elif log_distance >= len(self.buckets):
-            raise ValueError(f"Log distance must be smaller than {len(self.buckets)}")
-        return tuple(self.buckets[log_distance])
+        if log_distance <= 0:
+            raise ValueError("Log distance must be positive, got {log_distance}")
```
ChihChengLiang:
```suggestion
            raise ValueError(f"Log distance must be positive, got {log_distance}")
```

p2p/discv5/routing_table_manager.py
```diff
@@ -47,21 +55,60 @@
...
 
 
+@to_tuple
+def partition_enr_indices_by_size(enr_sizes: Sequence[int],
+                                  max_payload_size: int,
+                                  ) -> Generator[Tuple[int, ...], None, None]:
+    current_partition: Tuple[int, ...] = ()
+    current_partition_size = 0
+    for index, size in enumerate(enr_sizes):
+        if size > max_payload_size:
```
pipermerriam: Seems like this should raise an error as some kind of invariant condition since it is probably not reasonable for us to handle the properly inside this helper function.  This implies that an extra check should be added at some higher level that filters out ENR records that are too large and logs something to signal something is wrong.

PR #1485, opened 2020-01-20
Add eth1 monitor web3 endpoint option

NIC619:
### What was wrong?
Enable beacon node to request eth1 data from real web3 provider

### How was it fixed?
- [x] Add web3 provider options: HTTP, IPC, WebSocket
- [x] Read eth1 monitor configs from `eth1_monitor_config.yaml` instead of hardcoded them
    - Generate mockup `eth1_monitor_config.yaml` in `interop` setting if no `eth1_monitor_config.yaml` provided
- [x] Run a short-lived eth1 chain in `run_beacon_node.py`
    - use fake eth1 data instead if `USE_FAKE_ETH1_DATA == True`
    - there are a few steps to set up eth1 chain, but you only need to do it once
        - next time you run the script again it will use the same chain but deploy a new deposit contract
    - if these steps are too cumbersome, we can commit the initialized chain data and account data

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

source: https://www.maxpixels.net/Cavalier-King-Charles-Cute-Canine-Dog-Animal-Look-4256813


eth2/beacon/scripts/run_beacon_nodes.py
```diff
@@ -219,26 +440,87 @@ def sigint_handler(sig, frame):
...
+    # global variable `pw_dir` is the path to password of the geth account
+    # global variable `eth1_addr` is the account of eth1 chain miner
+
+    # Steps to set up eth1 chain with geth
+    # 1. create new account with `geth account new --datadir {data_dir}
+    # and replace `eth1_addr` with the account address you created
+    # and also replace the account in the bottom of genesis file, i.e., `poa_testnet.json`
+    # 2. store the password you input in step 1 into a file, e.g., pwd.txt`
+    # and update the path to password, i.e., `pw_dir`
+    # 3. init the chain with `geth init --datadir {data_dir} poa_testnet.json`
```
ChihChengLiang: What do you think we committing a default password to script dir, so that the `python eth2/beacon/scripts/run_beacon_node.py` just work? The steps here require many manual setups that are hard to get right.

NIC619: Yeah, indeed it's a problem... I commit the keystore and password file so now there's only one step needed.

ChihChengLiang:
We might also need to add the go ethereum version number some where.

eth2/beacon/scripts/run_beacon_nodes.py
NIC619: Yes, it's possible. Though since geth prevents you from re-init your chain, this command would fail.
I was thinking that since this is a one-time thing, it's better to leave it out of the script. I can add it into the script if that's more preferable.

ChihChengLiang: Hmmm, then can we remove `eth2/beacon/scripts/testnet/geth` every time we restarted? So we can always start with a fresh chain and geth won't complain about the re-init.

NIC619: Do you mean deleting the directory every time the script finishes?

PR #1486, opened 2020-01-20
Improve error handling on startup

cburgdorf:
### What was wrong?

1. We didn't use a string enum for `MiningMethod` when we should be (cleans up case matching)

2. We didn't validate the EIP1085 file early at startup which means errors would escape until a later point which isn't optimal

3. We run `validate_cli` on every component but we don't catch the errors to present them to the user properly

### How was it fixed?

1. Turned `MiningMethod` into an enum with string values

2. Call `boot_info.trinity_config.get_app_config(Eth1AppConfig).get_chain_config()` early in the syncer component

3. Catch validation errors on bootstrap and call `ArgumentParser().exit(message=str(exc))`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/bootstrap.py
+++ b/trinity/bootstrap.py
@@ -15,6 +15,7 @@ from typing import (
 )
 
 from async_service import AsyncioManager
+from eth_utils import ValidationError
 
 from trinity.exceptions import (
     AmbigiousFileSystem,
@@ -205,7 +206,10 @@ def main_entry(trinity_boot: BootFn,
 
     # Let the components do runtime validation
     for component_cls in component_types:
-        component_cls.validate_cli(boot_info)
+        try:
+            component_cls.validate_cli(boot_info)
+        except ValidationError as exc:
+            parser.exit(message=str(exc))
 
     # Components can provide a subcommand with a `func` which does then control
     # the entire process from here.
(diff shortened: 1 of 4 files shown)


trinity/bootstrap.py
```diff
@@ -205,7 +206,10 @@ def main_entry(trinity_boot: BootFn,
 
     # Let the components do runtime validation
     for component_cls in component_types:
-        component_cls.validate_cli(boot_info)
+        try:
+            component_cls.validate_cli(boot_info)
+        except ValidationError as exc:
+            ArgumentParser().exit(message=str(exc))
```
pipermerriam: Seems we have an instance of `ArgumentParser` already.  Can this just be `parser.exit(...)`

trinity/components/builtin/syncer/component.py
```diff
@@ -264,6 +264,10 @@ def validate_cli(cls, boot_info: BootInfo) -> None:
         # this will trigger a ValidationError if the specified strategy isn't known.
         cls.get_active_strategy(boot_info)
 
+        # This will trigger a ValidationError if the loaded EIP1085 file
+        # has errors such as an unsupported mining method
+        boot_info.trinity_config.get_app_config(Eth1AppConfig).get_chain_config()
```
pipermerriam: I don't :heart: the side-effect based nature of this but the code comment helps and maybe we can figure out a better way to do this as part of the `validate_cli` component API once we finish getting things properly into components.

cburgdorf: Yeah, I'm not crazy about it either but it didn't seem worth worrying too much about :sweat_smile:

trinity/bootstrap.py
cburgdorf: I was just checking if someone is paying attention :laughing:. Fixed!

PR #1487, opened 2020-01-20
DiscV5 Lookup

jannikluhn:
### What was wrong?

We didn't search for nodes.

### How was it fixed?

Implemented a lookup service that performs a recursive lookup in regular intervals.

Builds on top of #1468, #1479, and #1481.

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/abc.py
+++ b/p2p/discv5/abc.py
@@ -7,6 +7,7 @@ from typing import (
     AsyncIterable,
     Generic,
     Optional,
+    Tuple,
     Type,
     TypeVar,
 )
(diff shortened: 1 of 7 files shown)


pipermerriam:
Are you planning to divide this up into smaller PRs?

jannikluhn:
> Are you planning to divide this up into smaller PRs?

No, I think it doesn't make much sense to split it up further. Only the last three commits are new, everything comes from #1468, #1479, and #1481. The last commit is technically unrelated, but only a three character fix. The `request_nodes` method might be extractable, but it's essentially just a helper for lookups, so I think it belongs into this PR.

PR #1491, opened 2020-01-21
Add partial support for admin_peers JSON-RPC API

cburgdorf:
### What was wrong?

As part of the things that we want to ship with our first beta, we want to support the [`admin_peers` RPC endpoint](https://github.com/ethereum/go-ethereum/wiki/Management-APIs#admin_peers).

The response looks something like this in Geth:

```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": [
        {
            "enode": "enode://2ea727082ae93766342646f8ac1c98a20a4d14b1c1ca043e1b3c99dfa5a36767692eec2baf7b66474f6bdaca516322fcb45db58173c08a936f6ebab8928d4562@3.1.27.148:30303",
            "id": "e8191234978143d641f1df9898fde0a7a080ef3feb0f6149d857b1fa7d8a4549",
            "name": "Geth/source/linux/go1.10.4",
            "caps": [
                "eth/62",
                "eth/63"
            ],
            "network": {
                "localAddress": "172.17.0.2:56092",
                "remoteAddress": "3.1.27.148:30303",
                "inbound": false,
                "trusted": false,
                "static": false
            },
            "protocols": {
                "eth": {
                    "version": 63,
                    "difficulty": 729514544367193473735,
                    "head": "0x66f682d761aaabe7a98a1bc1b35303ef9c21b1b6c0a35c61965bce13f5499d28"
                }
            }
        },
        {
            "enode": "enode://c25a1d915f9cc87705690bd59af1776cf0ccc3fb7e8dfd31f5f37ef4a8b6a758a743a379bad6f0793c1413b10cdc26aa9aacb5bc1581c4625e22bb414af5840e@139.199.66.86:20182",
            "id": "e81f6a84023ea00c66bfc87857d14080136e91412afe2421816cd4c3be96e686",
            "name": "Geth/qk_client 1.0.1.202 0x4C1a41639D5bf12b8F5D1A83B2aFcbB6100174Eb/v1.9.11-stable-6a62fe39/windows-amd64/go1.13.8",
            "caps": [
                "eth/63",
                "eth/64",
                "eth/65",
                "les/2",
                "les/3"
            ],
            "network": {
                "localAddress": "172.17.0.2:51214",
                "remoteAddress": "139.199.66.86:20182",
                "inbound": false,
                "trusted": false,
                "static": false
            },
            "protocols": {
                "eth": "handshake"
            }
        }
    ]
}
```

### How was it fixed?

Actual Trintiy API response:

```python
{
    "id": 1,
    "jsonrpc": "2.0",
    "result": [
        {
            "enode": "enode://c25a56161448824a805edda3de2ca2de1e5ff498fa0ff6d6bc05939a075805d7217ff791bc003b8fe7cf05d8f1338ac578e1e68dc8e3358f5d27ff78799bcbb8@47.94.236.241:30303",
            "id": "40d4dbe8-675c-48c2-867f-b6a0ea919a94",
            "name": "Geth/v1.9.9-stable-01744997/linux-amd64/go1.13.5",
            "caps": [
                "eth/63",
                "eth/64",
                "shh/6"
            ],
            "network": {
                "localAddress": "0.0.0.0:30303",
                "remoteAddress": "47.94.236.241:30303"
            }
        }
    ]
}
```

Eagle eyes might spot that this is missing the `protocols` part which I aim to add in a different PR. The current implementation is based only on the information that we can obtain from the `BasePeerPool`.

### To-Do

- [x] Add tests

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/common/events.py
+++ b/trinity/protocol/common/events.py
@@ -5,6 +5,8 @@ from typing import (
     Any,
     Tuple,
     Type,
+    NamedTuple,
+    Dict,
 )
 
 from lahja import (
@@ -14,6 +16,7 @@ from lahja import (
 
 from p2p.abc import CommandAPI, NodeAPI, SessionAPI
 from p2p.disconnect import DisconnectReason
+from p2p.peer import BasePeer
 from p2p.typing import Capabilities
 
 
@@ -69,10 +72,28 @@ class PeerLeftEvent(BaseEvent):
     session: SessionAPI
 
 
+class PeerInfo(NamedTuple):
+    session: SessionAPI
+    capabilities: Capabilities
+    client_version_string: str
+    inbound: bool
+
+
 @dataclass
 class GetConnectedPeersResponse(BaseEvent):
 
-    sessions: Tuple[SessionAPI, ...]
+    peers: Tuple[PeerInfo, ...]
+
+    @staticmethod
+    def from_connected_nodes(peers: Dict[SessionAPI, BasePeer]) -> 'GetConnectedPeersResponse':
+        return GetConnectedPeersResponse(tuple(
+            PeerInfo(
+                session=session,
+                capabilities=peer.connection.remote_capabilities,
+                client_version_string=peer.connection.safe_client_version_string,
+                inbound=peer.inbound
+            ) for session, peer in peers.items()
+        ))
 
 
 class GetConnectedPeersRequest(BaseRequestResponseEvent[GetConnectedPeersResponse]):
--- a/trinity/rpc/modules/admin.py
+++ b/trinity/rpc/modules/admin.py
@@ -1,8 +1,10 @@
-from typing import Tuple, Iterable, Dict
+
+from typing import Tuple, Iterable, Dict, Sequence
 
 from eth.constants import GENESIS_BLOCK_NUMBER
 from eth_typing import BlockNumber
 from eth_utils import encode_hex, to_dict
+
 from lahja import EndpointAPI
 
 from p2p.kademlia import Node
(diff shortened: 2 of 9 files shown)


trinity/rpc/modules/admin.py
```diff
@@ -20,3 +30,21 @@ def __init__(self, event_bus: EndpointAPI) -> None:
...
+            'id': str(session.id),
+            'network': {
+                'localAddress': f"{BOUND_IP}:{self.trinity_config.port}",
+                'remoteAddress': f"{session.remote.address.ip}:{session.remote.address.tcp_port}"
+            }
+        }
+
+    async def peers(self) -> Sequence[RpcPeerResponse]:
+
+        response = await self.event_bus.request(GetConnectedPeersRequest())
```
cburgdorf: @pipermerriam So, our `GetConnectedPeersResponse` contains `SessionAPI`s now but all the interesting stuff that `geth` prints out is sitting on the peer. The obvious answer would be to change to not return a list of `SessionAPI` but something that includes more information (a subset of the actual peer) but I know you've been really deep into the p2p stack the last months and formed an idea of how these things could work in the future so I'm reaching out to here your thoughts before I just hack myself through until I have the missing data :)

pipermerriam: I think it would be very reasonable to have a new object returned that aggregates information about the session like:

```python
class PeerInfo(NamedTuple):
    session: SessionAPI
    capabilities: Capabilities
    ...
```

trinity/protocol/common/events.py
```diff
@@ -69,10 +72,28 @@ class PeerLeftEvent(BaseEvent):
     session: SessionAPI
 
 
+class PeerInfo(NamedTuple):
+    session: SessionAPI
+    capabilities: Capabilities
```
pipermerriam: Lol, it looks like you implemented pretty much the exact thing I typed up :+1:

PR #1493, opened 2020-01-22
Create ENRs for all nodes we find during discovery and use them as the backing data structure of Nodes

### LINKED ISSUE #1451: Consolidate discv4 and discv5 RoutingTable implementations
There's a lot of duplication between them both and we should be able to have a single implementation shared by both. The main issue is that the discv5 one stores only Node IDs (bytes) as it can assume all nodes will have a corresponding entry on the ENR DB, whereas we can't assume that on discv4 as not all peers support that.

gsalgado:
This will allow us to fix #1451, use a single DB for all nodes and by including an ENR in every Node instance, peer backends will have more info (e.g fork IDs) to decide which peers to connect to

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -155,11 +155,8 @@ class DiscoveryService(Service):
                  enr_field_providers: Sequence[ENR_FieldProvider] = tuple(),
                  ) -> None:
         self.privkey = privkey
-        self.address = address
         self.bootstrap_nodes = bootstrap_nodes
         self._event_bus = event_bus
-        self.this_node = Node(self.pubkey, address)
-        self.routing = RoutingTable(self.this_node)
         self.enr_response_channels = ExpectedResponseChannels[Tuple[ENR, Hash32]]()
         self.pong_channels = ExpectedResponseChannels[Tuple[Hash32, int]]()
         self.neighbours_channels = ExpectedResponseChannels[List[NodeAPI]]()
@@ -168,7 +165,6 @@ class DiscoveryService(Service):
         self._enr_db = enr_db
         # FIXME: Use a concurrency-safe EnrDb implementation.
         self._enr_db_lock = trio.Lock()
-        self._local_enr: ENR = None
         self._local_enr_next_refresh: float = time.monotonic()
         self._local_enr_lock = trio.Lock()
         self.parity_pong_tokens: Dict[Hash32, Hash32] = {}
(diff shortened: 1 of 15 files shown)


p2p/discovery.py
```diff
@@ -612,7 +612,15 @@ def send(self, node: NodeAPI, msg_type: DiscoveryCommand, payload: Sequence[Any]
             self.logger.warning("Ignoring uknown msg type: %s; payload=%s", cmd_id, payload)
             return
         self.logger.debug2("Received %s with payload: %s", cmd.name, payload)
-        node = Node(remote_pubkey, address)
+        try:
+            node = self.routing.get_node(node_id_from_pubkey(remote_pubkey))
+        except KeyError:
+            node = None
+        if node is None or node.address != address:
```
cburgdorf: Maybe this would be slightly clearer if the `KeyError` and non-equal address cases were separate.

```python
    try:
        node = self.routing.get_node(node_id_from_pubkey(remote_pubkey))
    except KeyError:
        node = Node.from_pubkey_and_addr(remote_pubkey, address)
    else:
        if node.address != address:
            # If this node's address changed since we last heard from it we create a new instance
            # as that will end up replacing the old one when we move it to the front of the bucket
            # (which is done every time we receive a msg from a node).
            node = Node.from_pubkey_and_addr(remote_pubkey, address)
```

p2p/kademlia.py
```diff
@@ -325,14 +372,22 @@ def add_node(self, node: NodeAPI) -> NodeAPI:
             return eviction_candidate
         return None  # successfully added to not full bucket
 
-    def get_bucket_for_node(self, node: NodeAPI) -> KBucket:
-        return binary_get_bucket_for_node(self.buckets, node)
+    def get_node(self, node_id: int) -> NodeAPI:
+        bucket = binary_get_bucket_for_node(self.buckets, node_id)
+        node = None
```
cburgdorf: :point_up: this `node` doesn't appear to be used. It is redefined in the `for` loop and then right after the loop a `KeyError` is raised.

gsalgado: Oops, indeed

p2p/discovery.py
gsalgado: Yeah, that's fine

p2p/discovery.py
```diff
@@ -180,6 +176,28 @@ def __init__(self,
...
             Tuple[NodeAPI, int]](self._max_pending_enrs)
 
+        # This is a stub Node instance (which will have an ENR with sequence number 0) that will
+        # be replaced in _init(). We could defer its creation until _init() is called by run(),
+        # but doing this here simplifies things as we don't have to store the Address in another
+        # instance attribute.
+        self.this_node = Node.from_pubkey_and_addr(self.pubkey, address)
+        self.routing = RoutingTable(self.this_node.id)
+
+    async def _init(self) -> None:
```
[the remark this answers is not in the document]
gsalgado: Well, this method performs initialization steps that can't live in `__init__` because they are async. That's why I chose to use a separate method and named it like this. I'd prefer to keep it separate from `_run()`

PR #1494, opened 2020-01-22
Implement Topic Table

jannikluhn:
Implements a topic table as specified [here](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md#topic-table).

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/topic_table.py
+++ b/p2p/discv5/topic_table.py
@@ -1,0 +1,91 @@
+import collections
+import math
+from typing import (
+    DefaultDict,
+    Deque,
+    NamedTuple,
+    Tuple,
+)
+
+from eth_utils import (
+    encode_hex,
+)
+
+from p2p.discv5.enr import ENR
+from p2p.discv5.typing import (
+    Topic,
+)
+
+
+class Ad(NamedTuple):
+    enr: ENR
+    registration_time: float
+
+
+class TopicTable:
+    def __init__(self, max_queue_size: int, max_total_size: int, target_ad_lifetime: float) -> None:
+        self.max_queue_size = max_queue_size
+        self.max_total_size = max_total_size
+        self.target_ad_lifetime = target_ad_lifetime
+
+        self.topic_queues: DefaultDict[Topic, Deque[Ad]] = collections.defaultdict(
+            lambda: collections.deque(maxlen=self.max_queue_size),
+        )
+        self.total_size = 0
+
+    def __len__(self) -> int:
+        """Return the total number of ads in the table across all queues."""
+        return self.total_size
+
+    def is_table_full(self) -> bool:
+        return len(self) >= self.max_total_size
+
+    def is_queue_full(self, topic: Topic) -> bool:
+        return len(self.topic_queues[topic]) >= self.max_queue_size
+
+    def get_enrs_for_topic(self, topic: Topic) -> Tuple[ENR, ...]:
+        """Get all ENRs registered for a given topic.
+
+        The result will be ordered from newest to oldest entry.
+        """
+        # reverse queue so that old entries come first
+        return tuple(ad.enr for ad in self.topic_queues[topic])
+
+    def get_wait_time(self, topic: Topic) -> float:
+        """Return the time at which the next ad for a given topic can be added."""
+        if self.is_table_full():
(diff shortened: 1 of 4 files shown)


p2p/discv5/topic_table.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def is_queue_full(self, topic: Topic) -> bool:
+        return len(self.topic_queues[topic]) >= self.max_queue_size
+
+    def get_enrs_for_topic(self, topic: Topic) -> Tuple[ENR, ...]:
+        """Get all ENRs registered for a given topic.
+
+        The result will be ordered from newest to oldest entry.
+        """
+        # reverse queue so that old entries come first
```
pipermerriam: This comment seems to not be accurate.

p2p/discv5/topic_table.py
```diff
@@ -0,0 +1,91 @@
...
+        self.topic_queues: DefaultDict[Topic, Deque[Ad]] = collections.defaultdict(
+            lambda: collections.deque(maxlen=self.max_queue_size),
+        )
+        self.total_size = 0
+
+    def __len__(self) -> int:
+        """Return the total number of ads in the table across all queues."""
+        return self.total_size
+
+    def is_table_full(self) -> bool:
```
pipermerriam: This could probably just be `is_full` since it would likely typically be called as `table.is_table_full` which makes the `table` part of the name a little redundant.

p2p/discv5/topic_table.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        # reverse queue so that old entries come first
+        return tuple(ad.enr for ad in self.topic_queues[topic])
+
+    def get_wait_time(self, topic: Topic) -> float:
+        """Return the time at which the next ad for a given topic can be added."""
+        if self.is_table_full():
+            oldest_ads = [queue[-1] for queue in self.topic_queues.values()]
+            oldest_ad_reg_time = max(ad.registration_time for ad in oldest_ads)
+        elif self.is_queue_full(topic):
```
pipermerriam: It seems to me this function's logic will return an incorrect value in the case that:

- both the table and the queue for the topic are full.
- the table's oldest add time is older than the topic queue's oldest time.

In this case it will return the time when the table allows registrations but at that time the table still won't allow registration because it will still have to wait for the individual queue element to expire.

Seems like we need a 3rd check that uses the newest of the two values in the event that both the table and the topic queue are full.

p2p/discv5/topic_table.py
```diff
@@ -0,0 +1,91 @@
...
+            queue = self.topic_queues[topic]
+            oldest_ad_reg_time = queue[-1].registration_time
+        else:
+            oldest_ad_reg_time = -math.inf
```
pipermerriam: This is odd to me.  Instead of returning `-math.inf` shouldn't we instead return `0` if there is no wait time?

p2p/discv5/topic_table.py
```diff
@@ -0,0 +1,91 @@
...
+    def register(self, topic: Topic, enr: ENR, current_time: float) -> None:
+        """Register a new ad.
+
+        A `ValueError` will be raised if the ad cannot be added because the table is full,
+        because the node already is present in the queue, or because the topic's wait time is
+        non-zero.
+        """
+        queue = self.topic_queues[topic]
+
+        if len(self) >= self.max_total_size:
```
pipermerriam: `if self.is_table_full`?

p2p/discv5/topic_table.py
jannikluhn: It's not the time/duration until, but the point in time at which registration is possible again.

jannikluhn:
Follow up PR: #1532

p2p/discv5/topic_table.py
pipermerriam: Hrm, seems like it would be better to do a delta or return `now()`.  I'm preferential to the delta (time until registration is possible) because it feels like there is a clearer conceptual distinction between:

- you may register now. (less ambiguity by using `0` or `None`) to represent this.
- you must wait X till you can register. (not critical but it doesn't suffer from leap second issues)

Both of these are not as clear when using absolute times.

PR #1498, opened 2020-01-23
[WIP] Wrap possible block_validation IndexErrors with ValidationError.

### LINKED ISSUE #1497: Beacon operation validation can raise `IndexError`s
### What is wrong?

Several block operation validation functions can raise `IndexError` in addition to a `ValidationError` e.g. [`validate_voluntary_exit()`](https://github.com/ethereum/trinity/blob/84381f4fe19b18dddf2d2b210bd0780038728acd/eth2/beacon/state_machines/forks/serenity/block_validation.py#L482), [`validate_proposer_slashing()`](https://github.com/ethereum/trinity/blob/84381f4fe19b18dddf2d2b210bd0780038728acd/eth2/beacon/state_machines/forks/serenity/block_validation.py#L140) when passed an `Operation` with an invalid validator index.

It is not obvious whether this is a documentation issue (that `IndexError`s are an expected result), or a crash-causing bug and DOS vector:

- Test handlers have `IndexError` as an expected exception type: https://github.com/ethereum/trinity/blob/84381f4fe19b18dddf2d2b210bd0780038728acd/eth2/beacon/tools/fixtures/test_types/operations.py#L135-L146
- Block importing doesn't appear to handle `IndexError`s: https://github.com/ethereum/trinity/blob/84381f4fe19b18dddf2d2b210bd0780038728acd/trinity/sync/beacon/chain.py#L158

### How can it be fixed

Catch `IndexError`s immediately and wrap them with a `ValidationError`. This is clearer and avoids potential bugs from forgetting to catch `IndexError`s.

### How it was found

Discovered via [`beacon-fuzz`](https://github.com/sigp/beacon-fuzz/) (initial testing of [`proposer_slashing` fuzzer](https://github.com/sigp/beacon-fuzz/tree/add_targets/files/fuzzers/proposer_slashing)).
Triggering case: [proposer_slashing-crash-35a9d8e810ef1c20f057ee4e6aa8a927dc2ed6dc](https://drive.google.com/file/d/17Zct2ozCvQA2wlmdPYNb1wETZr-gRjIr/view?usp=sharing) with the following [`beacon_state`s](https://github.com/sigp/beacon-fuzz-corpora/tree/1c2d95a67a2136b94511908250e8b21d550a09d0/0-9-1/mainnet/beaconstate)
Or the pre-processed input [propslash_preprocessed_indexerror.ssz](https://drive.google.com/file/d/1g7Am3n58LSna3ffGYOWPDLSVo-alaQZH/view?usp=sharing) can be directly passed to the [trinity harness](https://github.com/sigp/beacon-fuzz/blob/add_targets/files/fuzzers/proposer_slashing/trinity/harness.py)

gnattishness:
Fixes #1497.

Update test handlers to no longer expect IndexError

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

[Sam the Koala](https://commons.wikimedia.org/wiki/File:Taxidermied_Mount_-_Sam_the_Koala_1.jpg)
Benjamin Healley, Museums Victoria [CC BY](https://creativecommons.org/licenses/by/4.0)

### DIFF
--- a/eth2/beacon/state_machines/forks/serenity/block_validation.py
+++ b/eth2/beacon/state_machines/forks/serenity/block_validation.py
@@ -29,7 +29,7 @@ from eth2.beacon.types.proposer_slashings import ProposerSlashing
 from eth2.beacon.types.states import BeaconState
 from eth2.beacon.types.validators import Validator
 from eth2.beacon.types.voluntary_exits import SignedVoluntaryExit
-from eth2.beacon.typing import CommitteeIndex, Epoch, Root, Slot
+from eth2.beacon.typing import CommitteeIndex, Epoch, Root, Slot, ValidatorIndex
 from eth2.configs import CommitteeConfig, Eth2Config
 
 
@@ -150,13 +150,9 @@ def validate_proposer_slashing(
     Validate the given ``proposer_slashing``.
     Raise ``ValidationError`` if it's invalid.
     """
-    try:
-        proposer = state.validators[proposer_slashing.proposer_index]
-    except IndexError as error:
-        raise ValidationError(
-            f"Invalid ProposerSlashing index, proposer_index={proposer_slashing.proposer_index}",
-            error,
-        ) from error
+    proposer = _get_validator_by_index(
+        state.validators, proposer_slashing.proposer_index
+    )
 
     validate_proposer_slashing_epoch(proposer_slashing, slots_per_epoch)
 
@@ -491,6 +487,21 @@ def _validate_voluntary_exit_signature(
         )
 
 
+def _get_validator_by_index(
+    validators: Sequence[Validator], index: ValidatorIndex
+) -> Validator:
+    """
+    Return validators[index] but raise ValidationError if index is out of range.
+    Use this helper when the index can't be trusted.
+    """
+    try:
+        return validators[index]
+    except IndexError:
+        raise ValidationError(
+            f"Validator index ({index}) out of range ({len(validators)} validators)"
+        )
+
+
 def validate_voluntary_exit(
     state: BeaconState,
     signed_voluntary_exit: SignedVoluntaryExit,
@@ -498,13 +509,9 @@ def validate_voluntary_exit(
     persistent_committee_period: int,
 ) -> None:
     voluntary_exit = signed_voluntary_exit.message
-    try:
-        validator = state.validators[voluntary_exit.validator_index]
-    except IndexError as error:
-        raise ValidationError(
-            f"Invalid VoluntaryExit index, validator_index={voluntary_exit.validator_index}",
-            error,
-        ) from error
+    validator = _get_validator_by_index(
+        state.validators, voluntary_exit.validator_index
+    )
     current_epoch = state.current_epoch(slots_per_epoch)
 
     _validate_validator_is_active(validator, current_epoch)
--- a/eth2/beacon/state_machines/forks/serenity/block_validation.py
+++ b/eth2/beacon/state_machines/forks/serenity/block_validation.py
@@ -75,6 +75,8 @@ def validate_proposer_is_not_slashed(
     state: BeaconState, block_root: Root, config: CommitteeConfig
 ) -> None:
     proposer_index = get_beacon_proposer_index(state, config)
+    # NOTE: an IndexError should not occur here as get_beacon_proposer_index
+    # should return a valid index (provided a valid state)
     proposer = state.validators[proposer_index]
     if proposer.slashed:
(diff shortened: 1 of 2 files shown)


eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -75,6 +75,8 @@ def validate_proposer_is_not_slashed(
     state: BeaconState, block_root: Root, config: CommitteeConfig
 ) -> None:
     proposer_index = get_beacon_proposer_index(state, config)
+    # NOTE: an IndexError should not occur here as get_beacon_proposer_index
+    # should return a valid index (provided a valid state)
```
gnattishness: I didn't wrap these, as an IndexError here could indicate a bug in get_beacon_proposer_index or an invalid state.

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -89,6 +91,8 @@ def validate_proposer_signature(
 
     # Get the public key of proposer
     beacon_proposer_index = get_beacon_proposer_index(state, committee_config)
+    # NOTE: an IndexError should not occur here as get_beacon_proposer_index
+    # should return a valid index (provided a valid state)
```
gnattishness: As above

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -118,6 +122,8 @@ def validate_randao_reveal(
     randao_reveal: Hash32,
     slots_per_epoch: int,
 ) -> None:
+    # NOTE: an IndexError should not occur here provided proposer_index
+    # is the result of a call to get_beacon_proposer_index()
```
gnattishness: As above

gnattishness:
I also don't see where an IndexError is expected for `process_attestation`https://github.com/ethereum/trinity/blob/8f301c4e05174df1e85a1a204731580cdc194166/eth2/beacon/tools/fixtures/test_types/operations.py#L106

ChihChengLiang:
> I also don't see where an IndexError is expected for `process_attestation`
> 
> https://github.com/ethereum/trinity/blob/8f301c4e05174df1e85a1a204731580cdc194166/eth2/beacon/tools/fixtures/test_types/operations.py#L106

Feel free to remove that `expected_exceptions` too.

ralexstokes:
@gnattishness thanks for this!

the idea w/ the `expected_exceptions` was that there are tests in those suites that do (or did) raise the indicated exception and so receiving this condition when running those tests is expected. I can look at the latest w/ the test suite to see what is going on w/ the two suites you edited.

and as for how we want to handle the `IndexError` -- my intuition says we don't want to go down this road bc then we will forever be trapping every possible exception type at any possible site of error and wrapping it in `ValidationError`. i think it much better to just catch a bare `Exception` in the top-level block handler and log it as necessary. imo we don't want this sort of situation to crash the client, if anything, worst-case we would just fall out of consensus...

gnattishness:
Thanks for the review @ChihChengLiang and @ralexstokes!

The behavior of `expected_exceptions` in the test fixtures makes sense to me. My main uncertainty was whether the codebase itself expected the `IndexError` (e.g. [block syncing](https://github.com/ethereum/trinity/blob/84381f4fe19b18dddf2d2b210bd0780038728acd/trinity/sync/beacon/chain.py#L158) catches only a ValidationError)

My understanding is that catching `Exception` can be undesirable because it can mask errors that cannot be safely recovered from (e.g. `SystemError`, some cases of `MemoryError` or `OSError`) and could(?) result in the validator getting slashed.
Perhaps an alternative could be an exception heirarchy (e.g. inheriting from `BaseTrinityException`) that is handled at the top level, and can help differentiate an unexpected exception that indicates a bug (logged as `Critical` and asked to raise an Issue), or some other expected exception or system error.

Although verbose, I also prefer wrapping exceptions so the code is self-documenting whether the programmer took the error into account (rather than the documentation or tests needing to be maintained elsewhere)

(Obviously, you're welcome to do as you'd like and this is just my opinion :smiley:)

gnattishness:
Let me know before you merge, and I'll squish everything.

gnattishness:
@ChihChengLiang I prefer your approach and style here, so I grabbed it :smile: .
I like the reduced code repetition and my raising `from error` was unnecessary as the caught error is right nearby.
Please feel free to use #1502 instead.

ralexstokes:
Closing this for now as we are porting the eth2 code to another repo.

I'll link this PR against the issue in the new repo (where #1497 now lives).

PR #1501, opened 2020-01-23
discv4: Keep track of the last time we bonded with a remote

gsalgado:
Also use that to decide whether or not to reply to their FIND_NODEs and
ENR_REQUESTs

Closes: #1443

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -298,10 +298,13 @@ class DiscoveryService(Service):
         Bonding consists of pinging the node, waiting for a pong and maybe a ping as well.
         It is necessary to do this at least once before we send find_node requests to a node.
         """
-        if node in self.routing:
-            return True
-        elif node == self.this_node:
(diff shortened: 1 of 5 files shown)


p2p/discovery.py
```diff
@@ -298,11 +298,14 @@ def update_routing_table(self, node: NodeAPI) -> None:
...
         """
-        if node in self.routing:
-            return True
-        elif node == self.this_node:
+        if node == self.this_node:
+            # FIXME: We should be able to get rid of this check, but for now issue a warning.
+            self.logger.warning("Attempted to bond with self; this shouldn't happen")
             return False
 
+        if node.is_bond_valid():
```
pipermerriam: Existing convention in our codebase would be for this to be an `@property` method: `if node.is_bond_valid`

gsalgado: I haven't (re-)internalized all the conventions yet, will change

Issue #1303, opened 2019-11-21
Add some more admin endpoints


carver:
### What is wrong?

Would be good to be able to do things like list the currently-connected peers.

Related: #26 

### How can it be fixed

Implement the admin items from the rpc EIP (if there are any). If it doesn't define any way to list the peers, then I guess copy geth's `admin_` endpoints.

PR #1503, opened 2020-01-23
Implement admin_nodeInfo RPC endpoint

cburgdorf:
### What was wrong?

Related to #1303, we should support the [`admin_nodeInfo` RPC endpoint](https://github.com/ethereum/go-ethereum/wiki/Management-APIs#admin_nodeInfo).

### How was it fixed?

1. Added a new `GetSupportedProtocolsRequest` / `SupportedProtocolsResponse` pair to query the supported protocols from the peer pool

2. Added a new `nodeInfo` method to the `Admin` RPC module

3. Collect all relevant data and return it

Tada :tada: 

```
{
    "id": 1,
    "jsonrpc": "2.0",
    "result": {
        "enode": "enode://5796ec189cad54c801a8dab9790c6d5dad4378a128f336ed121f2edbb242b6bd1b31cb0d58f3c842626d10e5653a75e9a31a1831b969d27f3bdb45941e637f0f@0.0.0.0:30303",
        "ip": "::",
        "listenAddr": "[::]:30303",
        "name": "Trinity/v0.1.0a34/linux/cpython3.7.3",
        "ports": {
            "discovery": 30303,
            "listener": 30303
        },
        "protocols": {
            "eth": {
                "version": "eth/63",
                "difficulty": 2425203,
                "genesis": "0xbf7e331f7f7c1dd2e05159666b3bf8bc7a8a3a9eb1d518969eab529dd9b88c1a",
                "head": "0xd0902d3ef345f083011d18642b80629ba30b750d2ac4abe1ffe0a1a018098c6b",
                "network": 5,
                "config": {
                    "petersburgBlock": 0,
                    "istanbulBlock": 1561651,
                    "chainId": 5
                }
            }
        }
    }
}
```

### To-Do

- [x] Clean up commit history

- [x] Add tests

#### Cute Animal Picture

### DIFF
--- a/tests/integration/test_trinity_cli.py
+++ b/tests/integration/test_trinity_cli.py
@@ -131,10 +131,10 @@ async def test_expected_logs_for_light_mode(command):
 
 
 @pytest.mark.parametrize(
-    'command, expected_network_id, expected_genesis_hash',
+    'command, expected_network_id, expected_genesis_hash, expected_chain_id',
     (
-        (('trinity', '--log-level=DEBUG'), 1, MAINNET_GENESIS_HASH),
-        (('trinity', '--ropsten', '--log-level=DEBUG'), 3, ROPSTEN_GENESIS_HASH),
+        (('trinity', '--log-level=DEBUG'), 1, MAINNET_GENESIS_HASH, 1),
+        (('trinity', '--ropsten', '--log-level=DEBUG'), 3, ROPSTEN_GENESIS_HASH, 3),
         (
             (
                 'trinity',
@@ -145,13 +145,16 @@ async def test_expected_logs_for_light_mode(command):
                 '--network-id=4711',
                 '--disable-tx-pool',
                 '--log-level=DEBUG',
-            ), 4711, '0x065fd78e53dcef113bf9d7732dac7c5132dcf85c9588a454d832722ceb097422'),
+            ),
+            4711, '0x065fd78e53dcef113bf9d7732dac7c5132dcf85c9588a454d832722ceb097422', 3735928559
+        ),
     )
 )
 @pytest.mark.asyncio
 async def test_web3_commands_via_attached_console(command,
                                                   expected_network_id,
                                                   expected_genesis_hash,
+                                                  expected_chain_id,
                                                   xdg_trinity_root,
                                                   unused_tcp_port):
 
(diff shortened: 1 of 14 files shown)


tests/integration/test_trinity_cli.py
```diff
@@ -190,6 +193,12 @@ def amend_command_for_unused_port(command, unused_tcp_port):
             attached_trinity.expect(str(GENESIS_BLOCK_NUMBER))
             attached_trinity.sendline("w3.eth.getBlock(0).hash")
             attached_trinity.expect(expected_genesis_hash)
+            # The following verifies only a tiny subset of the node_info response. Making assertions
+            # regarding the full response is difficult with this test. This mainly exists here to
+            # ensure the API responds *at all* and contains *something* we would expect.
```
carver: Totally reasonable not to check every value. Maybe just check that the other fields exist and are not empty?

trinity/rpc/modules/admin.py
```diff
@@ -1,17 +1,46 @@
...
+    GetSupportedProtocolsRequest
+)
+from trinity.rpc.modules import Eth1ChainRPCModule
+from trinity.rpc.typing import RpcProtocolResponse, RpcNodeInfoResponse
+from trinity.server import BOUND_IP
+from trinity._utils.version import construct_trinity_client_identifier
+
+
+def format_enode(config: TrinityConfig) -> str:
+    return f"enode://{config.nodekey.public_key.to_hex()[2:]}@{BOUND_IP}:{config.port}"
```
carver: We'll probably want the bound IP to be configurable at some point, but it's fine to punt IMO.

trinity/rpc/modules/admin.py
```diff
@@ -1,17 +1,46 @@
...
+
 
+@to_dict
+def generate_chain_config(chain_config: Eth1ChainConfig) -> Iterable[Tuple[str, int]]:
+    for fork_block, vm in chain_config.vm_configuration:
+        yield f"{vm.fork}Block", fork_block
 
-class Admin(BaseRPCModule):
+    yield 'chainId', chain_config.chain_id
```
carver: Hm, maybe this belongs in `trinity/_utils/eip1085.py`?

trinity/rpc/modules/admin.py
```diff
@@ -20,3 +49,49 @@ def __init__(self, event_bus: EndpointAPI) -> None:
...
+        }
+
+    async def _generate_protocol_info(
+            self,
+            protocols: Tuple[Tuple[str, int], ...]) -> Dict[str, RpcProtocolResponse]:
+
+        return {
+            protocol: await self._generate_info_for(protocol, version)
+            for protocol, version in protocols
+        }
```
carver: It could be fine to in-line this, IMO, but it's fine either way.

trinity/rpc/modules/admin.py
```diff
@@ -20,3 +49,49 @@ def __init__(self, event_bus: EndpointAPI) -> None:
...
+
+    async def _generate_info_for(self, protocol: str, version: int) -> RpcProtocolResponse:
```
carver: I'm not sure it makes sense to model this as something that can be generically applied across different protocols. Seems like each protocol will have it's own fields that it will want to report.

Hard-coding it or doing some kind of formatter lookup by protocol are the other options that come to mind.

trinity/rpc/typing.py
```diff
@@ -1,11 +1,34 @@
 from typing import (
     Sequence,
     Union,
-)
+    Dict)
```
carver: *nitpick* trailing comma with paren on the next line?

trinity/protocol/common/events.py
```diff
@@ -90,3 +90,16 @@ class PeerPoolMessageEvent(BaseEvent):
     """
     session: SessionAPI
     command: CommandAPI[Any]
+
+
+@dataclass
+class SupportedProtocolsResponse(BaseEvent):
+
+    protocols: Tuple[Tuple[str, int], ...]
```
pipermerriam: I think we define an alias to this in [`p2p.typing`](https://github.com/ethereum/trinity/blob/master/p2p/typing.py#L12-L13).  You could go ahead and make that a `NamedTuple` if you are feeling ambitious but I think it's worth using that type either way in place of the `Tuple[str, int]`.

trinity/rpc/typing.py
cburgdorf: Ah, I assume that might be me trying out PyCharm professional with its auto import feature :)
Fixed :+1:

trinity/rpc/modules/admin.py
cburgdorf: I thought about that but the format is slightly different. The EIP1085 format defines block numbers and the `chainId` using `hex` values whereas the `admin_nodeInfo` response defines these using `int`s. Also, the EIP1085 format defines some of the fork blocks by their EIP numbers (e.g. `EIP158ForkBlock` whereas the `admin_nodeInfo` response defines that as `SpuriosDragonFork`.

So this method returns data in a format that is specific to the `admin_nodeInfo` endpoint and might be better suited here for now. That said, there's doesn't exist an actual spec that defines the `admin_nodeInfo` response so we could also consider to align the format of EIP1085 and `admin_nodeInfo` better. But that seems out of the scope of this PR.

trinity/rpc/modules/admin.py
cburgdorf: Yeah, you are right. I'm going to inline this for now and then we can revisit it when we support other protocols.

trinity/protocol/common/events.py
cburgdorf: Ah yeah, there is `Capabilities`. I updated to use this and also aligned the naming to use `capabilities` where appropriate.

PR #1510, opened 2020-01-24
Add libp2p to autodoc_mock_imports

cburgdorf:
### What was wrong?

Our live API docs are currently broken.

### How was it fixed?

While we had excluded `libp2p` for RTD bugs since ages, we did not put it in the `autodoc_mock_imports`. My theory is that we initially didn't need it because we had no direct imports to `libp2p` but started to have these at some point.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -51,7 +51,7 @@ autodoc_default_options = {
 }
 
 autodoc_member_order = 'groupwise'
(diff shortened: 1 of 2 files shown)


cburgdorf:
I'm going to merge this without review as an emergency fix to unbreak our live API docs.  I double checked that it works.

PR #1513, opened 2020-01-27
discv4: Use NodeID instead of int for Node.id

gsalgado:
We still use int IDs internally but this will make it easier for us to
share a single RoutingTable implementation between both discv5 and
discv4

Another step towards fixing #1451

### DIFF
--- a/p2p/kademlia.py
+++ b/p2p/kademlia.py
@@ -44,6 +44,7 @@ from p2p.discv5.constants import (
     UDP_PORT_ENR_KEY,
 )
 from p2p.discv5.identity_schemes import V4CompatIdentityScheme
+from p2p.discv5.typing import NodeID
 from p2p.validation import validate_enode_uri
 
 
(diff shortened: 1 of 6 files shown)


p2p/kademlia.py
```diff
@@ -245,10 +246,10 @@ def split(self) -> Tuple['KBucket', 'KBucket']:
         lower = KBucket(self.start, splitid)
         upper = KBucket(splitid + 1, self.end)
         for node in self.nodes:
-            bucket = lower if node.id <= splitid else upper
+            bucket = lower if node._id_int <= splitid else upper
```
pipermerriam: we shouldn't be accessing private attributes...  maybe just make it a public attribute?

gsalgado: My reasoning was that this would be ok here since `Node` is internal to this module, and we don't want that attribute used anywhere other than here. In #1515 I've changed `NodeAPI` to expose its attributes as read-only properties, so to merge this without waiting for another review round-trip and avoid conflicts I can make this one public there if you'd prefer

Issue #1449, opened 2020-01-03
Peer discovery strategy is too eager


gsalgado:
Currently we perform kademlia lookups for a random node ID every 30s, and each of those lookups consist of sending a lot of FIND_NODE msgs to peers whose ID are close to that random one we picked. These peers will then reply with lists of nodes and we'll then try to bond with them, finally adding them to our routing table when the bonding is successful. That allows us to populate the routing table pretty quickly, but what I'm seeing in practice is that once it's relatively full (less than 10 minutes running), most times we try the bond/add-to-table dance with nodes we get during a random lookup, the add-to-table fails because the bucket is full, and that immediately triggers a bond with the oldest node in the bucket, but we've bonded with that node just a few minutes ago so that's a complete waste. To make matters worse, if we get a list of nodes where many of them fall on the same (full) bucket, we'd fail to add them all and end up triggering multiple bonds with the same node (the oldest one in the bucket).

I think a better strategy would be to perform random lookups periodically until we have a certain number of nodes/full-buckets in the routing table and from then on switch to refreshing buckets (i.e. re-bonding with the oldest node, replacing it with one from the replacement cache if it fails) that have been idle for a certain period of time, and running random lookups when we start running out of entries in the replacement cache.

PR #1514, opened 2020-01-28
PeerPool now uses a filter function when asking for connection candidates

gsalgado:
One of the issues I've found with our peer discovery implementation is that PeerPool keeps asking the Discovery backend for a batch of random peers, and it only skips the nodes we are
already connected to, so we are likely to get duplicates across requests and if we have a big routing table with few compatible peers we're very unlikely to ever get to them.

This changes our backends to instead take a filter function and only return candidates that pass it. This way we can use ENR info (e.g. fork-id) to first try and connect to peers that are known to be compatible, and resort to random ones (that don't support ENR) if we do not have enough compatible ones. This will come in a separate PR (#1523)

Also, this allows us to make DiscoveryService's strategy less eager (#1449) by only triggering random lookups when we don't have enough compatible peers in the routing table.

Closes: #1449

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -168,6 +168,7 @@ class DiscoveryService(Service):
         self._enr_db_lock = trio.Lock()
         self._local_enr_next_refresh: float = time.monotonic()
         self._local_enr_lock = trio.Lock()
+        self._lookup_lock = trio.Lock()
         self.parity_pong_tokens: Dict[Hash32, Hash32] = {}
         if socket.family != trio.socket.AF_INET:
             raise ValueError("Invalid socket family")
@@ -203,12 +204,34 @@ class DiscoveryService(Service):
         while self.manager.is_running:
             await self.consume_datagram()
 
+    def get_peer_candidates(
+        self, should_skip_fn: Callable[[NodeAPI], bool], max_candidates: int,
+    ) -> Tuple[NodeAPI, ...]:
+        total_nodes = len(self.routing)
+        candidates = []
+        for candidate in self.get_nodes_to_connect(total_nodes):
+            if should_skip_fn(candidate):
+                continue
+            candidates.append(candidate)
+            if len(candidates) == max_candidates:
+                break
+        else:
+            self.logger.info(
+                "Not enough nodes in routing table passed PeerCandidatesRequest's "
+                "filter, triggering a random lookup in the background.")
+            if self._lookup_lock.locked():
+                self.logger.info(
+                    "Not performing random lookup as there is one in progress already")
+            else:
+                self.manager.run_task(self.lookup_random)
+        return tuple(candidates)
+
     async def handle_get_peer_candidates_requests(self) -> None:
         async for event in self._event_bus.stream(PeerCandidatesRequest):
-            nodes = tuple(self.get_nodes_to_connect(event.max_candidates))
-            self.logger.debug2("Broadcasting peer candidates (%s)", nodes)
+            candidates = self.get_peer_candidates(event.should_skip_fn, event.max_candidates)
+            self.logger.debug2("Broadcasting peer candidates (%s)", candidates)
             await self._event_bus.broadcast(
-                event.expected_response_type()(nodes),
+                event.expected_response_type()(candidates),
                 event.broadcast_config()
             )
 
@@ -232,15 +255,10 @@ class DiscoveryService(Service):
     def run_daemons_and_bootstrap(self) -> None:
         self.manager.run_daemon_task(self.handle_get_peer_candidates_requests)
         self.manager.run_daemon_task(self.handle_get_random_bootnode_requests)
-        self.manager.run_daemon_task(self.periodically_refresh)
         self.manager.run_daemon_task(self.report_stats)
         self.manager.run_daemon_task(self.fetch_enrs)
         self.manager.run_daemon_task(self.consume_datagrams)
         self.manager.run_task(self.bootstrap)
-
-    async def periodically_refresh(self) -> None:
-        async for _ in trio_utils.every(self._refresh_interval):
-            await self.lookup_random()
 
     async def fetch_enrs(self) -> None:
         async with self.pending_enrs_consumer:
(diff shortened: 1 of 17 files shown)


gsalgado:
@pipermerriam how do you feel about this approach?

gsalgado:
> TBH I _can_ be ok with this but I don't immediately like it. Something about the `filter_fn` approach feels like it couples these two things together more tightly than I'd like.
> 
> I'm curious about a completely different approach that comes to mind. It seems like it wouldn't be very expensive to transmit the entirety of the data from the routing table across the event bus. 

We'd be pickling/unpickling up to of 4k items (256 buckets * 16 nodes) only for the `DiscoveryBackend`, but then there's also the `Eth1PeerTracker` backend (and possibly others). Would it make sense to fetch all Nodes from all of them when, let's say, we have only a few free slots to fill? The time needed for pikcling/unpickling and sending them over the event bus would probably be insignificant when compared to the network roundtrips of attempting to connect to peers, but still, when we have only a few open slots to fill, it'd be extremely wasteful. Using the filter approach together with a limit on the number of nodes would be significantly less wasteful

Also, with the filter approach the `DiscoveryBackend` is able to trigger a lookup for more nodes when it detects there are not enough nodes to fulfill a request (which is the fix I intended for #1449) whereas otherwise the lookup would have to be triggered externally (probably by the `PeerPool`), and that means we'd have to special case the `DiscoveryBackend`.

> That would allow the peer pool to fully reconstruct the routing table locally which would allow it to have fine grained control over how it selects nodes.

FWIW, there'd be no use in reconstructing the RT as that's only meant to allow us to easily lookup nodes based on their "proximity" to other nodes, which is necessary when answering other peers' requests for nodes during the discovery protocol.

gsalgado:
Just ran a quick test and pickling/unpickling 4k `Node` instances takes over 1s (of a single core at 100%), so although it doesn't seem prohibitively expensive, if we take into account that most of the time we have difficulty finding enough peers to fill all our slots, we'd be performing a not-that-cheap operation quite regularly

pipermerriam:
@gsalgado I am adequately convinced `filter_fn` is the better route. :smile_cat:

p2p/discovery.py
```diff
@@ -203,12 +204,34 @@ def __init__(self,
         while self.manager.is_running:
             await self.consume_datagram()
 
+    def get_peer_candidates(
+        self, should_skip: Callable[[NodeAPI], bool], max_candidates: int,
```
pipermerriam: Maybe call this `should_skip_fn`.  More consistent with other names and makes it harder to confuse as something like a `bool`.

p2p/discovery.py
```diff
@@ -203,12 +204,34 @@ def __init__(self,
...
+
     async def handle_get_peer_candidates_requests(self) -> None:
         async for event in self._event_bus.stream(PeerCandidatesRequest):
-            nodes = tuple(self.get_nodes_to_connect(event.max_candidates))
-            self.logger.debug2("Broadcasting peer candidates (%s)", nodes)
+            candidates = self.get_peer_candidates(event.should_skip, event.max_candidates)
+            self.logger.debug2("Broadcasting peer candidates (%s)", candidates)
             await self._event_bus.broadcast(
-                event.expected_response_type()(nodes),
+                event.expected_response_type()(tuple(candidates)),
```
pipermerriam: `tuple` cast shouldn't be necessary here as `get_peer_candidates` returns a tuple.

p2p/peer_pool.py
```diff
@@ -539,3 +550,7 @@ def _peer_finished(self, peer: AsyncioServiceAPI) -> None:
                     self.logger.debug("    %s", line)
             self.logger.debug("== End peer details == ")
             await self.sleep(self._report_interval)
+
+
+def skip_candidate_if_on_list(skip_list: Tuple[NodeID, ...], candidate: NodeAPI) -> bool:
```
pipermerriam: Input type here should probably be `typing.Container` as it is the most generic version that support the `item in thing` API

trinity/components/builtin/network_db/connection/tracker.py
```diff
@@ -20,30 +21,32 @@
...
 from .events import (
     BlacklistEvent,
+    GetBlacklistedPeersRequest,
     ShouldConnectToPeerRequest,
 )
 
 
 class BlacklistRecord(Base):
-    __tablename__ = 'blacklist_records'
+    __tablename__ = 'blacklist_recordsv0.1'
```
pipermerriam: I'm :-1: on this approach.  I think you should bump the schema version in `trinity.db.orm. SCHEMA_VERSION`.  Everyone will be required to drop their network database but I don't think that is too big a deal.

trinity/components/builtin/network_db/connection/tracker.py
```diff
@@ -173,9 +182,14 @@ def record_blacklist(self, remote: NodeAPI, timeout_seconds: int, reason: str) -
             self.config,
         )
 
+    # XXX: Should I get rid of this?
```
pipermerriam: If it isn't used, yes.

p2p/discovery.py
gsalgado: I actually had it named like that at first, but decided to change because I thought the type annotations should make it clear it's a function. Happy to change it back, though

p2p/discovery.py
gsalgado: Indeed, changed

p2p/peer_pool.py
gsalgado: Didn't know about that one!

trinity/components/builtin/network_db/connection/tracker.py
gsalgado: Forgot to mention this in the PR description, but yeah, this should've had an XXX as it was really just me asking how to properly handle schema changes. Changed SCHEMA_VERSION

trinity/components/builtin/network_db/connection/tracker.py
gsalgado: :+1:

PR #1515, opened 2020-01-29
Convert EnrDb into NodeDB

gsalgado:


pipermerriam:
ping me for review when #1513 is merged and this is rebased.  Hard to separate the code diff across 4 commits.

gsalgado:
Just rebased this, @pipermerriam. As I mentioned in https://github.com/ethereum/trinity/pull/1513#discussion_r372709458, just let me know if you'd like me to add the `id_int` attribute to `NodeAPI`. Also, you might still prefer to review the commits individually because the first one is just mechanical changes (renames and changing APIs to take a `NodeAPI` instead of `ENR`), whereas the others introduce behaviour changes

p2p/kademlia.py
```diff
@@ -119,15 +119,33 @@ def __init__(self, enr: ENR) -> None:
...
-        self.pubkey = keys.PublicKey.from_compressed_bytes(enr.public_key)
-        self.id = NodeID(keccak(self.pubkey.to_bytes()))
+        try:
+            ip = enr[IP_V4_ADDRESS_ENR_KEY]
+        except KeyError:
+            self._address = None
+        else:
+            self._address = Address(
+                ip,
+                udp_port=enr.get(UDP_PORT_ENR_KEY),
```
jannikluhn: ENRs are not guaranteed to have a UDP port (even though almost all will, I guess), so we should handle this in some way. Setting `self._address` to `None` here as well is probably good enough.

p2p/kademlia.py
```diff
@@ -119,15 +119,33 @@ def __init__(self, enr: ENR) -> None:
...
+                tcp_port=enr.get(TCP_PORT_ENR_KEY, 0),
+            )
+        self._pubkey = keys.PublicKey.from_compressed_bytes(enr.public_key)
```
jannikluhn: ENRs may use different identity schemes which have different public key formats, so this line may break. For now there's just one, but this isn't unlikely to change at some point. Dealing with the public key object directly shouldn't be necessary because we have the `IdentityScheme` class which provides an interface for all the different crypto related functions.

gsalgado:
> I think having a single db instance for storing both ENRs and additional node attributes makes sense, but storing everything together as pickled objects seems like not the right way to go. 

Yeah, it didn't feel like the ideal solution but I thought it'd be worth trying to see how it'd look like

> One weird side effect this has for example is that, because we keep a history of all ENRs, we now also keep a history of the other attributes, but only snapshotted at the time we update the ENRs.

Well, the fact that we keep a history of all ENRs feels more like a bug to me as I see no reason why we need to do it since we should never have to roll back to a previous ENR version for any given node.

> It would cleaner if we stored them separately, for example ENRs in files as it is right now and everything else in a level db.

I don't see why we'd want to use two different storage backends; if we're going to use leveldb then why not store everything there? We could use one key for the ENR and another key for the other attributes that may change more frequently (or even one key for every attribute?) so that we don't have to unnecessarily serialize the ENR every time we update something else.

pipermerriam:
I think you may already be planning to move away from it but I'm a big :-1: on storing pickled values in a database.  Everytime I've ever done this it turned out to be a bad idea.  It should be pretty easy to define a simple schema for serializing/deserializing ENR records and it will save us a lot of future problems.

gsalgado:
> Right now we have a hierarchy of the ENR being a property of the `Node`. I think we have this relationship backwards. The ENR is a superset of the data a `Node` exposes which implies that the `ENR` should be the outer object and the `NodeAPI` would be a property of the ENR record and we should be looking at re-architecting our code so that we use ENR objects in all of the places that we currently use a `NodeAPI`.

`NodeAPI` actually stores things that can be derived from the ENR; for now only `last_pong`, but soon there should be other things, like liveness checks.

jannikluhn:
> Well, the fact that we keep a history of all ENRs feels more like a bug to me as I see no reason why we need to do it since we should never have to roll back to a previous ENR version for any given node.

> I don't see why we'd want to use two different storage backends; if we're going to use leveldb then why not store everything there? We could use one key for the ENR and another key for the other attributes that may change more frequently (or even one key for every attribute?) so that we don't have to unnecessarily serialize the ENR every time we update something else.

Yeah, I think I agree with both of your points. Using only the leveldb and getting simply overwriting outdated ENRs would get rid of a bunch of complexity. Using one key per attribute probably makes the most sense, then we can even get rid of special casing ENRs. They would become just another attribute, at least low-level.

pipermerriam:
@gsalgado 

> `NodeAPI` actually stores things that can't be derived from the ENR; for now only `last_pong`, but soon there should be other things, like liveness checks.

I think this still hints at what I am puzzling over.  

Prior to `last_pong`, the `NodeAPI` was a pure derivative of the ENR.  The `node_id` can be thought of as the *primary key* as it is the one piece of information that remains constant.  All of the other information like address, port, fork_id, are metadata that the node is in control of updating (via sending and updated ENR record).  Now we have a new class of metadata, `last_pong` which is contextual to the connection, so I'll call this "Session" data.

So that gives us 3 types of data.

- primary key (unique identifier): `node_id`
  - does not change
  - derived from ENR
- metadata: `ip`, `port`, `fork_id`
  - only changes with an updated ENR with new sequence number
  - derived from ENR
- session information: `last_pong`
  - ephemeral, only relevant within a single run, doesn't *need* to be persisted.
  - maybe ok to just store in memory

So my thought is that we should be cognizant of these different classifications because all of the data that needs to be persisted seems to be in the ENR.  Please correct me if I'm mistaken about any of this.

I believe it still makes sense to keep referring to this as the `NodeAPI` and keeping the current hierarchy, but I wanted to do a bit of a zoom out on the shape of this information to be sure we're handling it correctly.  Reading back over what I've written here there doesn't seem to be anything directly actionable :man_facepalming: so carry on and feel free to ignore my ramblings.

gsalgado:
> Also, I'm not a fan of the node objects because they are mutable and they would be hard to update if we want to add more fields (especially if they are pickled). What's wrong with accessing the db directly, keeping only the node id has representation of the node around? 

This would incur in a lot more DB queries as we'd have to lookup node details every time we send/receive a message. Most times only for the node we're dealing with, but for FIND_NODE requests we'd have to lookup info about all the nodes we include in the response. I'm not sure how much of a problem that would be, but ISTM it'd make sense to keep NodeAPI instances in the routing table as it's not prohibitively large so we can easily fit all nodes in it in memory -- that's what we do now anyway.

gsalgado:
> So my thought is that we should be cognizant of these different classifications because all of the data that needs to be persisted seems to be in the ENR. Please correct me if I'm mistaken about any of this.

That's correct, but we do actually need to persist metadata like `last_pong` because otherwise after a restart we'd start ignoring requests from nodes for which we have a valid bond, and since they don't know that they'd start treating us as an unresponsive node.

p2p/kademlia.py
gsalgado: I've added a FIXME comment and will deal with it as soon as we've agreed how to move forward with this PR

gsalgado:
Ok, so I don't think it's a good idea to squeeze extra changes in this PR to use a LevelDB to store the ENR and node metadata as that would make it a lot bigger and it'd be hard to distinguish the mechanical changes from the behaviour-changing ones. Given that, I've decided to morph this so that it simply renames the existing DB classes and updates their APIs to take NodeAPI instances instead of ENR. It no longer pickles the Node instance, nor it stores extra node metadata -- under the hoods we're storing only the ENR just like before. If you guys are ok, I'd like to get this through review first (as it touches a lot of files/APIs and has lots of potential for conflicts) and once it's merged I'll start working on using LevelDB and storing Node metadata apart from the ENR as well

jannikluhn:
> This would incur in a lot more DB queries as we'd have to lookup node details every time we send/receive a message. Most times only for the node we're dealing with, but for FIND_NODE requests we'd have to lookup info about all the nodes we include in the response. I'm not sure how much of a problem that would be, but ISTM it'd make sense to keep NodeAPI instances in the routing table as it's not prohibitively large so we can easily fit all nodes in it in memory -- that's what we do now anyway.

The db queries should be pretty well cachable.

I'm mostly worried about complexity and unnecessary entanglement in the routing table implementation. All the routing table cares about is node identities and insertion order, so I don't really see why it should deal with objects that also contain lots of additional data (public keys, capabilities, reputation, ...).

Issue #1442, opened 2019-12-31
How to actually benefit from ENR for better peer discovery?


gsalgado:
Soon we'll be fetching ENRs as part of node discovery, but in order to really benefit from that we will also need to change how `DiscoveryService` selects the nodes trinity should connect to.

I believe we still want to keep all nodes we contact (regardless of fork compatibility) in our routing table, so that we can be a decent p2p citizen. However, in our current implementation we use buckets of size 16 and when we try to add a node to a full bucket, those go into a replacement cache and will only get added to the routing table if the oldest node in that bucket is no longer responding. That means nodes that are compatible with us may never make it to the routing table and thus never be picked as connection candidates.

One way around that would be to get as many connection candidates we can from the ENR DB (which is separate from the routing table), and resort to the routing table only if we can't get enough from there. But I'm pretty sure that'd be a breach of protocol and may have security issues. Or maybe not, as long as we picked random ones from the ENR DB? One downside here is that we wouldn't try to connect to any non-ENR-enabled nodes that might be fork-compatible with us.

Another alternative would be to somehow prioritize nodes with ENRs when populating the routing table, and continue picking random ones from the routing table to use as connection candidates (although ignoring ones that are known not to be compatible).

A third option would be to not prioritize nodes with ENRs when populating the routing table and then, when picking connection candidates, use a mix of random known-to-be-fork-compatible entries from the ENR DB and random entries (with fork-compatibility unknown) from the routing table.

pipermerriam:
My *ideal* scenario would be to architect this so that the pieces are as independent as possible.  Here's where the lines seem to draw themselves.

## Discovery

This is the part that manages the kademlia routing table, bonds with peers, and retrieves their ENR records.  This part would expose an API for something like:

- fetching random nodes from the routing table
- broadcasting information about ENR records over the event bus.

## Network DB

> Not necessarily the same as the current `network-db` component which could be scrapped if we think rebuilding something simpler is better.

This part manages persisting ENR records between runs.  It listens for information about new ENR records (typically coming from the Discovery service) and keeps track of them.

- listen for ENR record information sent over the event bus.
- querying ENR database for records.

---------------------------------------------------------------


My thought on the above is that it aims to keep the discovery component simple and constrained without business logic, and lets us write something more specific to handle the business logic of consuming, querying, and persisting ENR record information.

I'll leave it to you to decide the right way forward.  My preference would be to try and keep our discovery service free of ETH specific business logic if at all possible.

[...]

PR #1523, opened 2020-02-03
Filter peer candidates by fork-id

gsalgado:
Closes: #1442

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -632,6 +632,26 @@ class DiscoveryService(Service):
         self.logger.debug2("Received datagram from %s", address)
         self.manager.run_task(self.receive, address, datagram)
 
+    async def _lookup_node_from_db(self, node_id: NodeID) -> NodeAPI:
+        # XXX: This is meant to be a temporary helper while we have NodeAPI instances in the RT
+        # and ENRs in the DB. Soon we'll store NodeAPI instances in the DB and this may become
+        # uneccesary as we'll have to do just a simple DB lookup.
+        try:
+            node = self.routing.get_node(node_id)
+        except KeyError:
+            node = None
+
+        if node is None or node.enr.sequence_number == 0:
+            # This means we still haven't received an ENR for the node, so try to look up an
+            # existing one from our DB.
+            try:
+                enr = await self._enr_db.get(node_id)
+            except KeyError:
+                pass
+            else:
+                node = Node(enr)
+        return node
+
     async def receive(self, address: AddressAPI, message: bytes) -> None:
         try:
             remote_pubkey, cmd_id, payload, message_hash = _unpack_v4(message)
@@ -644,17 +664,12 @@ class DiscoveryService(Service):
         except KeyError:
             self.logger.warning("Ignoring uknown msg type: %s; payload=%s", cmd_id, payload)
             return
-        self.logger.debug2("Received %s with payload: %s", cmd.name, payload)
-        try:
-            node = self.routing.get_node(node_id_from_pubkey(remote_pubkey))
-        except KeyError:
+
+        node = await self._lookup_node_from_db(node_id_from_pubkey(remote_pubkey))
+        if node is None or node.address != address:
             node = Node.from_pubkey_and_addr(remote_pubkey, address)
-        else:
-            if node.address != address:
-                # If this node's address changed since we last heard from it we create a new
-                # instance as that will end up replacing the old one when we move it to the front
-                # of the bucket (which is done every time we receive a msg from a node).
-                node = Node.from_pubkey_and_addr(remote_pubkey, address)
+
+        self.logger.debug2("Received %s from %s with payload: %s", cmd.name, node, payload)
         handler = self._get_handler(cmd)
         await handler(node, payload, message_hash)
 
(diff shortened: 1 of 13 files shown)


gsalgado:
@pipermerriam this doesn't work because the fork-id filter function includes a reference to a `vm_config` object, which can't be pickled

> _pickle.PicklingError: Can't pickle <class 'abc.TangerineWhistleVM'>: attribute lookup TangerineWhistleVM on abc failed\x1b[0m\n

Any suggestions on how to workaround that? I could change our fork-id APIs to take a list of fork-block numbers, but then the PeerPool would have to use the `_extract_fork_blocks()` function itself and use the return of that in the filter function. Not a big deal, but just thought I'd check in case maybe you've run into this situation before and have already created a pickle-able vm-config thing?

p2p/discovery.py
```diff
@@ -632,6 +632,26 @@ def send(self, node: NodeAPI, msg_type: DiscoveryCommand, payload: Sequence[Any]
...
+        except KeyError:
+            node = None
+
+        if node is None or node.enr.sequence_number == 0:
+            # This means we still haven't received an ENR for the node, so try to look up an
+            # existing one from our DB.
+            try:
+                enr = await self._enr_db.get(node_id)
+            except KeyError:
+                pass
```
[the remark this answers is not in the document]
gsalgado: This is expected to happen when it's the first time we get a message from a given peer, so it's ok for it to return None here. In that case, `receive()` will detect it and create a new `Node` instance for the peer

gsalgado:
> In the same veign, the BasePeerPool shouldn't have any awareness of the chain constructs. The new methods you added to the BasePeerPool should also be relocated to be somewhere in the trinity namespace, probably under trinity.protocol.common

Yeah, I wanted to add them only to `BaseChainPeerPool`, but they'd be used in `BasePeerPool.maybe_connect_more_peers()` so I thought it'd be ok to add them to `BasePeerPool` as abstract methods. Anyway, I made `BasePeerPool.maybe_connect_more_peers()` an abstractmethod now, and then was able to move the new methods to `BaseChainPeerPool`

gsalgado:
> Some guidelines for code organization. Anything that is specific to "Ethereum" should live under the trinity namespace. The only things that belong under the p2p namespace are things that are generic DevP2P protocol things (as well as a few shared coding constructs like p2p/resource_lock.py

Ok, my bad. I think I still saw the `p2p` namespace as the place for all the networking-related code, as I believe was the case before the split from py-evm to trinity. I'll submit a separate PR after this one moving the forkid code somewhere else

p2p/discovery.py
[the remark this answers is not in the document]
pipermerriam: I'd push for this to raise an exception rather than return `None`.  Return value checking is inherently error prone since it is easy to forget to do (though a little less easy now that we have `mypy` helping us out)

pipermerriam:
Just saw your question about the unpickleability of the `VMConfig`.  I don't have a good workaround for you beyond something like what you mentioned which side-steps the need to pass the whole thing and to pre-extract the fork blocks.

Probably something that would be nice to address eventually since I think the lack of pickle support hints at us leaning too heavily on what might be an antipattern of dynamic class construction and modification at runtime but that's a much bigger issue.

p2p/discovery.py
```diff
@@ -1199,11 +1217,12 @@ def get_channel(self, remote: NodeAPI) -> trio.abc.SendChannel[TMsg]:
 async def generate_eth_cap_enr_field(
         vm_config: Tuple[Tuple[BlockNumber, Type[VirtualMachineAPI]], ...],
         headerdb: 'BaseAsyncHeaderDB',
-) -> Tuple[Literal[b'eth'], Tuple[ForkID]]:
+) -> Tuple[Literal[b'eth'], Tuple[forkid.ForkID]]:
     head = await headerdb.coro_get_canonical_head()
     genesis_hash = await headerdb.coro_get_canonical_block_hash(GENESIS_BLOCK_NUMBER)
-    forkid = make_forkid(genesis_hash, head.block_number, vm_config)
-    return (b'eth', (forkid,))
+    fork_blocks = forkid.extract_fork_blocks(vm_config)
```
pipermerriam: I think this is already what you proposed but I think you should fix the pickle issue by changing this function to take `fork_blocks` instead of computing it internally from the `vm_config`

trinity/protocol/common/peer.py
```diff
@@ -187,3 +268,38 @@ def setup_peer_backends(self) -> Tuple[BasePeerBackend, ...]:
...
+    return False
+
+
+def skip_candidate_if_on_list_or_fork_mismatch(
+        genesis_hash: Hash32,
+        head: BlockNumber,
+        fork_blocks: Tuple[BlockNumber, ...],
+        skip_list: Container[NodeID],
+        candidate: NodeAPI) -> bool:
+    logger = get_extended_debug_logger('p2p')
```
pipermerriam: I would do this at the module level to avoid re-fetching it every time this function is called.  This isn't an expensive function but it isn't as cheap as `logging.getLogger` because it does some checks to be sure the returned logger is of the right type as well as swapping in/out the default logger class.

p2p/discovery.py
gsalgado: This function doesn't need to be pickle-able -- it was `make_forkid()` that had to be changed to take a list of fork_blocks instead of vm_config, so here I had to add the `extract_fork_blocks()` call

trinity/protocol/common/peer.py
gsalgado: Makes sense

PR #1532, opened 2020-02-04
Topic table fixes

jannikluhn:
### What was wrong?

As pointed out by @pipermerriam [here](https://github.com/ethereum/trinity/pull/1494), the topic table implementation was buggy:

- wait time was computed incorrectly if both queue and table are full
- registering ads in full tables wasn't possible

### How was it fixed?

- compute wait time properly (based on the registration time of the oldest ad in the queue or in the table, whatever was later
- if the table is full, remove the oldest ad in the table, no matter the queue
- rename `is_table_full` to `is_full`
- remove wrong comment on order of enrs in queue
- extend the tests so that the fixed issues are covered

#### Cute Animal Picture

### DIFF
--- a/p2p/discv5/topic_table.py
+++ b/p2p/discv5/topic_table.py
@@ -48,7 +48,6 @@ class TopicTable:
 
         The result will be ordered from newest to oldest entry.
         """
-        # reverse queue so that old entries come first
         return tuple(ad.enr for ad in self.topic_queues[topic])
 
     def get_wait_time(self, topic: Topic) -> float:
--- a/p2p/discv5/topic_table.py
+++ b/p2p/discv5/topic_table.py
@@ -1,5 +1,6 @@
 import collections
 import math
+import operator
 from typing import (
     DefaultDict,
     Deque,
@@ -10,6 +11,7 @@ from typing import (
 from eth_utils import (
     encode_hex,
 )
+from eth_utils import toolz
 
 from p2p.discv5.enr import ENR
 from p2p.discv5.typing import (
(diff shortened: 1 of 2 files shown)


p2p/discv5/topic_table.py
```diff
@@ -37,7 +39,7 @@ def __len__(self) -> int:
         """Return the total number of ads in the table across all queues."""
         return self.total_size
 
-    def is_table_full(self) -> bool:
+    def is_full(self) -> bool:
```
pipermerriam: I would make this a property method.  Anyone who accidentally does `if table.is_full` without calling it will *mistakenly* end up with a truthy response.

https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#booleans

p2p/discv5/topic_table.py
```diff
@@ -48,21 +50,30 @@ def get_enrs_for_topic(self, topic: Topic) -> Tuple[ENR, ...]:
 
         The result will be ordered from newest to oldest entry.
         """
-        # reverse queue so that old entries come first
         return tuple(ad.enr for ad in self.topic_queues[topic])
 
     def get_wait_time(self, topic: Topic) -> float:
```
pipermerriam: I'm still in favor of moving to use "time until registration is possible" rather than the current "time **at which** registration is possible".  However, if you choose to keep it this way, this method should probably be renamed as the current name suggests this would return the "time until ..." value due.

Maybe `get_allowed_registration_time`?

jannikluhn: Changed it, both are fine with me. The reason why I slightly preferred the absolute time is that it saves one parameter (the current time -- I'm trying to avoid an internal clock). But I agree that `get_wait_time` would be misleading if it returned an absolute time.

PR #1534, opened 2020-02-04
Convert main Node service to use async-service API

pipermerriam:
### What was wrong?

The `Node` concept in trinity is still using the legacy service API

### How was it fixed?

Updated it to use the new `async-service` API.  This includes a small patch to the old service API to easily wrap them in the new service API.

#### Cute Animal Picture

funny-cats-dogs-stuck-furniture-19

### DIFF
--- a/p2p/abc.py
+++ b/p2p/abc.py
@@ -20,6 +20,7 @@ from typing import (
 )
 import uuid
 
+from async_service import ServiceAPI
 from cancel_token import CancelToken
 
 from eth_utils import ExtendedDebugLogger
@@ -411,6 +412,10 @@ class AsyncioServiceAPI(ABC):
     events: ServiceEventsAPI
     cancel_token: CancelToken
 
+    @abstractmethod
+    def as_new_service(self) -> ServiceAPI:
+        ...
+
     @property
     @abstractmethod
     def logger(self) -> ExtendedDebugLogger:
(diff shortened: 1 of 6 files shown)


p2p/abc.py
```diff
@@ -445,6 +450,10 @@ def is_operational(self) -> bool:
     async def cancel(self) -> None:
         ...
 
+    @abstractmethod
+    async def cancellation(self) -> None:
```
gsalgado: A non-verb name for a method always strikes me as odd. Maybe after I've gone through the whole diff I'll have a suggestion about a better name

p2p/service.py
```diff
@@ -436,3 +440,12 @@ class EmptyService(BaseService):
...
                 pass
+
+
+class WrappedLegacyService(Service):
+    def __init__(self, service: AsyncioServiceAPI) -> None:
+        self._service = service
+
+    async def run(self) -> None:
+        async with run_service(self._service):
+            await self._service.cancellation()
```
gsalgado: Maybe `wait_cancellation()`, to be consistent with `wait_finished()` & co?

pipermerriam: This is the legacy API that I'm deprecating so I'm not going to do any updates to it.  The `abc.py` file change was just to appease `mypy` so it knew the method exists.

PR #1537, opened 2020-02-05
Piper/convert peer pool event server to use async service

pipermerriam:
builds on #1534 

### What was wrong?

The `PeerPoolEventServer` is not using the `async-service` API.

### How was it fixed?

Updated it to use the new API.

#### Cute Animal Picture

article-2154934-1374A08B000005DC-493_634x431


trinity/nodes/base.py
```diff
@@ -123,12 +123,13 @@ def base_db(self) -> AtomicDatabaseAPI:
...
-    async def _cleanup(self) -> None:
-        await self.event_bus.broadcast(ShutdownRequest("Node finished unexpectedly"))
+            self.manager.run_daemon_task(self.handle_network_id_requests)
+            self.manager.run_daemon_child_service(self.get_p2p_server().as_new_service())
+            self.manager.run_daemon_child_service(self.get_event_server())
+            try:
+                await self.manager.wait_finished()
+            finally:
+                self.master_cancel_token.trigger()
+                await self.event_bus.broadcast(ShutdownRequest("Node finished unexpectedly"))
```
carver: Why isn't this message being broadcast always, even during normal (Ctrl-C) termination?

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -60,22 +63,19 @@ class PeerPoolEventServer(BaseService, PeerSubscriber, Generic[TPeer]):
     This class bridges all common APIs but protocol specific communication can be enabled through
     subclasses that add more handlers.
     """
+    logger = get_extended_debug_logger('trinity.protocol.common.PeerPoolEventServer')
```
carver: Don't we have a nice mixin for this so we don't have to manually define it everywhere, like `HasExtendedDebugLogger`?

pipermerriam: I went down that rabbit hole a while back.  It requires a metaclass.  Python gets angry if you have a class with more than one disjoint set of metaclass parents.  The `typing.Generic` class and the other generics are all metaclasses.  This meant that while we can do `HasExtendedDebugLogger` in practice it requires a bunch of nasty boilerplate and shenanigans to get both python and the type system happy....

So I've resorted to just using `get_extended_debug_logger` and doing manual declaration of logger instances on classes.  I'm pretty *ok* with this approach.

trinity/nodes/base.py
pipermerriam: Maybe I translated incorrectly but I was just migrating the code from `_cleanup` so that it runs when the node exits.

Assuming I understand your comment, I think you are correct that this isn't ideal since it will broadcast the event even on a normal shutdown.  Currently I think this behavior ends up being benign as on a CTRL-C shutdown the broadcasting of this event will occur after everything is already on the way to being shutdown.

carver: Okay, yeah, the old construction seems similar. If it seems right to you, maybe we could update the text to something more general, like "Node component exiting".

PR #1538, opened 2020-02-05
Convert trinity server to use async service

pipermerriam:
builds on https://github.com/ethereum/trinity/pull/1537

### What was wrong?

The `trinity.server` constructs were not using the `async-service` API

### How was it fixed?

Converted them to use `async-service`.

#### Cute Animal Picture

5845719280_02d8a774e5_z


gsalgado:
This has been addressed in another PR

PR #1539, opened 2020-02-06
Helpers to run components in a standalone fashion

gsalgado:
So, these changes allow us to run a DBManager in one terminal and then easily run standalone components, connecting to that DB. My goal here was to be able to run the Discovery component and then use an event-bus client to request peer candidates to it. To do that one simply has to 

```
$ python trinity/db/manager.py -l debug --trinity-root-dir /tmp/discv4-driver
$ python trinity/components/builtin/peer_discovery/component.py -l p2p=debug --trinity-root-dir /tmp/discv4-driver
$ python scripts/discovery.py -ipc /tmp/discv4-driver/mainnet/ipcs-eth1/discovery.ipc
```

This doesn't have an event-bus service like when we run trinity, so all connections between components would have to be established manually.

### DIFF
--- a/trinity/components/builtin/peer_discovery/component.py
+++ b/trinity/components/builtin/peer_discovery/component.py
@@ -116,3 +116,12 @@ async def generate_eth_cap_enr_field(
     fork_blocks = forkid.extract_fork_blocks(vm_config)
     our_forkid = forkid.make_forkid(genesis_hash, head.block_number, fork_blocks)
     return (b'eth', sedes.List([forkid.ForkID]).serialize([our_forkid]))
+
+
+async def main() -> None:
+    from trinity.extensibility.component import run_standalone_eth1_component
+    await run_standalone_eth1_component(PeerDiscoveryComponent)
+
+
+if __name__ == "__main__":
+    trio.run(main)
--- a/trinity/extensibility/component.py
+++ b/trinity/extensibility/component.py
@@ -8,13 +8,22 @@ from argparse import (
 )
 import asyncio
 import logging
-from typing import AsyncIterator
+from typing import AsyncIterator, Iterable, Type, TYPE_CHECKING, Union
 
 from async_generator import asynccontextmanager
 
+from lahja import AsyncioEndpoint, ConnectionConfig, TrioEndpoint
+
 from trinity._utils.os import friendly_filename_or_url
 from trinity.boot_info import BootInfo
+from trinity.cli_parser import parser, subparser
+from trinity.config import BaseAppConfig, Eth1AppConfig, TrinityConfig
+from trinity.constants import APP_IDENTIFIER_ETH1, SYNC_FULL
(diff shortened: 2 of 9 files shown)


gsalgado:
@pipermerriam this is already quite useful to me when manually testing peer discovery strategies to find compatible ones, but I'd be happy to hear if you had other plans on how to architect this? I remember you mentioned you had some ideas about running components independently

trinity/components/builtin/peer_discovery/component.py
```diff
@@ -92,3 +90,12 @@ def configure_parser(cls,
...
             raise
+
+
+async def main() -> None:
+    from trinity.extensibility.component import run_standalone_eth1_component
+    await run_standalone_eth1_component(PeerDiscoveryComponent)
+
+
+if __name__ == "__main__":
+    trio.run(main)
```
cburgdorf: I like the idea of being able to run components standalone. Especially since multi process debugging isn't that great in Python.

I think this could be generalized into its own `Application` component (similar to e.g. [`fix-unclean-shutdown`](https://github.com/ethereum/trinity/blob/51df2b5e0c2b15863256fb70a8178c7d282fe020/trinity/components/builtin/fix_unclean_shutdown/component.py#L19). This type of component doesn't really boot `trinity`, it simply hijacks the `triniy` command to run any (main) function.

So, instead of having to prepare every component with a `main` function one could then say:

`trinity run-standalone-component PeerDiscoveryComponent -l p2p=debug --trinity-root-dir /tmp/discv4-driver`

This would become a generalized thing that should work with every component. Eventually, if the `DBManager` would also become a component one would be able to write:

```
trinity run-standalone-component DBManagerComponent
trinity run-standalone-component PeerDiscoveryComponent -l p2p=debug --trinity-root-dir /tmp/discv4-driver
```

trinity/extensibility/component.py
```diff
@@ -119,3 +128,57 @@ def _get_endpoint_name(cls) -> str:
...
+    component: Union[Type['TrioIsolatedComponent'], Type['AsyncioIsolatedComponent']],
+) -> None:
+    from trinity.extensibility.trio import TrioIsolatedComponent  # noqa: F811
+    from trinity.extensibility.asyncio import AsyncioIsolatedComponent  # noqa: F811
+    if issubclass(component, TrioIsolatedComponent):
+        endpoint_type = TrioEndpoint
+    elif issubclass(component, AsyncioIsolatedComponent):
+        endpoint_type = AsyncioEndpoint
+    else:
+        raise ValueError("Unknown component type")
```
pipermerriam: you might include `type(component)` in this message so that the exception message contains more useful info

gsalgado:
I've decided to go with this for now as the boilerplate we need to add to every component is really minimal and I think it'd make sense to see how this will work for other components before trying to make it work for all of them

PR #1526, opened 2020-02-03
Use manual garbage collection in Resourcelock

pipermerriam:
fixes #1405

### What was wrong?

The use of `weakref` to manage the locks for the `ResourceLock` has some sort of race condition that causes it to fail.

### How was it fixed?

Changed to use manual garbage collection.

#### Cute Animal Picture

Top-10-Images-of-Cats-With-Their-Heads-Stuck-8

PR #1540, opened 2020-02-06
Reproducing issue from 1526

carver:
### What was wrong?

#1526 has an issue. Buried in the latest fix is a bug that will be even harder to track it down and see what's going on.  I want to see if we can find a different solution.

### How was it fixed?

Checking to see if the builtin `asynccontextmanager` will solve it.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


carver:
Damn, still happens with the builtin version.

Hunting this down sucks, but at least we can reproduce it reliably right now. I really don't want to be stuck with this landmine when it happens occasionally and no one remembers how to reproduce it anymore.

PR #1541, opened 2020-02-07
Catch UnknownAPI error & log fixup

carver:
### What was wrong?

A bug in #1525 -- Not all `UnknownAPI` exceptions were replaced by `PeerConnectionLost`.

### How was it fixed?

We need to catch both exceptions.

Bonus change: the "lookup finished for target" log was moved to `debug`.

### To-Do

- [x] Clean up commit history

- [x] Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md) (nothing to add)

#### Cute Animal Picture


carver:
@gsalgado any objection to dropping the log severity?

PR #1542, opened 2020-02-10
Move the forkid package under trinity/protocol/eth

gsalgado:
This created circular dependencies so I've moved some globals into a classmethod
in ENRContentSedes and used a local import there to workaround it.

I was thinking that going forward we'd want to refactor the ENR class (and its parents)
into a base ENR and then extend that in trinity, but that's going to be annoying
because the current NodeAPI has an ENR instance attribute. Maybe the right way forward
is to have ENR expose any extra key/value pairs (like the 'eth' one) as raw values
and let callsites decode them? I think that's probably the most future-proof solution

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -24,11 +24,8 @@ from typing import (
     Sequence,
     Set,
     Tuple,
-    Type,
     TypeVar,
-    TYPE_CHECKING,
 )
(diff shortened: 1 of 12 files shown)


gsalgado:
I went ahead and removed the `eth` key from ENR's sedes mapping as that seemed like the right thing to do.

Issue #908, opened 2019-08-12
Beam Sync: pivot to latest block if the lag gets too high


carver:
### What is wrong?

If Beam Sync lags behind far enough, then it stops being able to sync. Sync fails because peers no longer have the necessary trie nodes.

### How can it be fixed

Unfortunately it's difficult to strictly define that this scenario has occurred, because peers are permitted to return empty results when requesting trie nodes. It could be a peer that is overloaded, or just recently started syncing, or that the local peer has fallen too far behind.

Likely, we'll need to measure the lag from the tip, or the time lag, and use that to decide that we have fallen too far behind. At that point, we would simply skip over importing the remaining blocks in between the last imported block and the tip, and re-start beam sync from the tip again. This should all happen without intervention.

carver:
I'm not certain this belongs in the MVP, just adding it for visibility, and maybe we will remove it again.

pipermerriam:
I think we could get reasonably reliable results if we define the need to pivot as:

1. no forward progress of syncing state after some timeout
    - all peers fail at returning needed nodes
    - maybe confirm by checking the actual block state-root node is also not available
2. state root of newer blocks is available
    - at least one peer does return non-empty state requests for "new blocks"

This allows us to a) detect the stall condition which I think can be done semi-reliably now and b) confirm that peers will return data for newer blocks (as defined as them being willing to return the node data for a fresh state root.

pipermerriam:
(this is probably more complex than is absolutely necessary but I want to document it).

If we define an integer `BEAM_WINDOW` which is the number of blocks from the known chain head, we can calculate the probability that we will fail to import or preview a given block.  It seems like this might be a good metric to determine if we should pivot.  This would allow us to pivot before starting a preview or import rather than timing out in the middle which might be uglier.

- `BLOCK_TIME` is the average block time measured over recent history.
- `BLOCK_STDEV` is the standard deviation of the block time over recent history
- `IMPORT_AVG` is the average time it takes to import a block
- `IMPORT_STDEV` is the standard deviation
- `PREV_AVG` is the average time it takes to import a block
- `PREV_STDEV` is the standard deviation.

We can compute an upper bound on block times with:

- `MAX_EXPECTED_BLOCK_TIME = BLOCK_TIME + 3 * BLOCK_STDEV`

We can determine an upper bound on import or preview time via with 99.9% likelyhood as:

[...]

PR #1545, opened 2020-02-12
Catch more sync errors

carver:
### What was wrong?

A variety of errors pop up sporadically. Occasionally, they event stop sync dead.

There is also an issue that causes the `BeamSyncer` to hang on shutdown. This was preventing pivot from working correctly, so is a blocker for #908 

### How was it fixed?

Mostly, just catch exceptions that were bubbling up too far.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/multiplexer.py
+++ b/p2p/multiplexer.py
@@ -396,6 +396,21 @@ class Multiplexer(CancellableMixin, MultiplexerAPI):
             *self._protocols,
             token=token,
         ), token=token)
+        try:
+            await self._handle_commands(msg_stream, stop)
+        except asyncio.TimeoutError as exc:
+            self.logger.warning(
+                "Timed out waiting for command from %s, Stop: %r, exiting...",
+                self,
+                stop.is_set(),
+            )
+            self.logger.debug("Timeout %r: %s", self, exc, exc_info=True)
+
+    async def _handle_commands(
+            self,
+            msg_stream: AsyncIterator[Tuple[ProtocolAPI, CommandAPI[Any]]],
+            stop: asyncio.Event) -> None:
+
         async for protocol, cmd in msg_stream:
             # track total number of messages received for each command type.
             self._msg_counts[type(cmd)] += 1
--- a/p2p/multiplexer.py
+++ b/p2p/multiplexer.py
@@ -33,6 +33,7 @@ from p2p.abc import (
 )
 from p2p.cancellable import CancellableMixin
 from p2p.exceptions import (
+    CorruptTransport,
     UnknownProtocol,
     UnknownProtocolCommand,
 )
@@ -365,6 +366,9 @@ class Multiplexer(CancellableMixin, MultiplexerAPI):
                         await asyncio.wait_for(fut, timeout=1)
                     except asyncio.TimeoutError:
                         pass
+                    except CorruptTransport as exc:
+                        self.logger.error("Corrupted transport found in %s: %r", self, exc)
+                        self.logger.debug("Corrupted transport trace in %s", self, exc_info=True)
 
                 # After giving the transport an opportunity to shutdown
                 # cleanly, we issue a hard shutdown, first via cancellation and
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -520,8 +520,6 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
                     "client_version_string='%s'",
                     peer.p2p_api.safe_client_version_string,
                 )
-                if not hasattr(peer, "eth_api"):
-                    self.logger.warning("Huh? %s doesn't have an eth API", peer)
                 for line in peer.get_extra_stats():
                     self.logger.debug("    %s", line)
             self.logger.debug("== End peer details == ")
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -53,6 +53,7 @@ from p2p.exceptions import (
     MalformedMessage,
     NoMatchingPeerCapabilities,
     PeerConnectionLost,
+    UnknownAPI,
     UnreachablePeer,
 )
 from p2p.peer import (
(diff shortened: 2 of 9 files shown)


trinity/sync/beam/importer.py
```diff
@@ -469,19 +469,36 @@ def _trigger_missing_state_downloads() -> None:
...
+                transactions,
+                header,
+                preview_time,
+                exc,
+            )
+        except OSError:
+            vm.logger.info("Got OSError while Beam Sync previewed execution of %s", header)
+        except BrokenPipeError:
+            vm.logger.info("Got BrokenPipeError while Beam Sync previewed execution of %s", header)
+            vm.logger.debug("BrokenPipeError trace for %s", header, exc_info=True)
```
carver: This seems to be the simplest way to handle when beam importing is cancelled externally. Someday, we can either handle it more explicitly/proactively, or we can just reduce it to debug or debug2 log, if we decide this catch-all approach isn't problematic.

trinity/sync/beam/queen.py
```diff
@@ -139,10 +143,28 @@ def _update_queen(self, peer: ETHPeer) -> ETHPeer:
         elif peer == self._queen_peer:
             # nothing to do, peer is already the queen
             return None
-        elif _peer_sort_key(peer) < _peer_sort_key(self._queen_peer):
```
carver: This change isn't nearly as big as it looks. Basically, we just have to catch it when a peer disappears at the moment we ask for it's quality. We want to do different things depending on which peer went away. So we have to make these lookups ahead of the if block.

p2p/peer_pool.py
```diff
@@ -520,9 +521,11 @@ def _peer_finished(self, peer: AsyncioServiceAPI) -> None:
...
                 )
-                if not hasattr(peer, "eth_api"):
-                    self.logger.warning("Huh? %s doesn't have an eth API", peer)
-                for line in peer.get_extra_stats():
-                    self.logger.debug("    %s", line)
+                try:
+                    for line in peer.get_extra_stats():
+                        self.logger.debug("    %s", line)
+                except (UnknownAPI, PeerConnectionLost) as exc:
+                    self.logger.debug("    Failure during stats lookup: %r", exc)
```
cburgdorf: Is there a reason for the whitespace in the beginning of this logging message? It's been present in the previous log so maybe it was just carried along.

trinity/sync/beam/chain.py
```diff
@@ -734,15 +734,15 @@ def __init__(
 
     async def _provide_missing_account_tries(self) -> None:
         async for event in self.wait_iter(self._event_bus.stream(CollectMissingAccount)):
-            self.run_task(self._serve_account(event))
+            self.run_task(self.wait(self._serve_account(event)))
```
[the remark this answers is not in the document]
cburgdorf: I'm just noting that the slightly better fix here is probably to migrate it to the new `async-service` which doesn't need the `wait` constructs anymore. But that shouldn't hold you back fixing it this way for now.

trinity/sync/beam/queen.py
```diff
@@ -139,10 +143,28 @@ def _update_queen(self, peer: ETHPeer) -> ETHPeer:
...
-            self._waiting_peers.put_nowait(old_queen)
-            return old_queen
         else:
-            # nothing to do, peer is slower than the queen
-            return None
+            try:
+                new_peer_quality = _peer_sort_key(peer)
+            except (UnknownAPI, PeerConnectionLost) as exc:
+                self.logger.debug("Ignoring %s, because we can't get speed stats: %r", peer, exc)
+                return None
```
cburgdorf: I know this has always returned `None` so no need to change it now but let me just impersonate @pipermerriam for a second to point out that it would be more aligned with the rest of our code base if this would raise exceptions (e.g. `QueenUnchanged`, `NoPreviousQueen`) instead of returning `None` here.

p2p/multiplexer.py
```diff
@@ -33,6 +33,7 @@
 )
 from p2p.cancellable import CancellableMixin
 from p2p.exceptions import (
+    TransportOverlap,
```
pipermerriam: I think that `CorruptTransport` would be more appropriate.   I think the definition of what happens here is that we lose our place in the stream due to a read operation being interrupted at some point.  At this stage I don't think there is a way to recover meaning that we should close the transport and terminate the peer connection when we detect this occuring.

trinity/sync/beam/chain.py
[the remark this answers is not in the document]
pipermerriam: I'll handle this separately.

p2p/peer_pool.py
carver: Yeah, it's an indentation to indicate that it's a sub-detail of the line that came before it. The line before it identifies which peer we are logging about.

For example:

```
   DEBUG  2020-02-11 15:51:17,318           ETHPeerPool  == Peer details == 
   DEBUG  2020-02-11 15:51:17,319           ETHPeerPool  ETHPeer <Session <Node(0x945b15@1.1.1.1)> 8bb73250-fecb-41c9-9e06-16b12f79cdcd>: uptime=58s, received_msgs=17
   DEBUG  2020-02-11 15:51:17,319           ETHPeerPool  client_version_string='Geth/v1.9.3-stable-cfbb969d/linux-amd64/go1.11.5'
   DEBUG  2020-02-11 15:51:17,319           ETHPeerPool      trinity.protocol.eth.commands.BlockBodies: None
   DEBUG  2020-02-11 15:51:17,319           ETHPeerPool      trinity.protocol.eth.commands.BlockHeaders: msgs=1  items=2  rtt=19.01/0.16/0.00  ips=0.62755  timeouts=0  quality=5
   DEBUG  2020-02-11 15:51:17,319           ETHPeerPool      trinity.protocol.eth.commands.NodeData: None
   DEBUG  2020-02-11 15:51:17,319           ETHPeerPool      trinity.protocol.eth.commands.Receipts: None
   DEBUG  2020-02-11 15:51:17,319           ETHPeerPool  ETHPeer <Session <Node(0xe61905@2.2.2.2)> 81a2d9fa-3dce-4562-80b7-82f5f9a1c0a4>: uptime=58s, received_msgs=6
   DEBUG  2020-02-11 15:51:17,319           ETHPeerPool  client_version_string='Geth/v1.9.2-stable-e76047e9/linux-amd64/go1.11.5'
   DEBUG  2020-02-11 15:51:17,319           ETHPeerPool      Failure during stats lookup: <exception details here>

cburgdorf: Ah, of course! Nevermind :)

trinity/sync/beam/queen.py
carver: It didn't feel right at first glance, so I poked around and decided it wasn't really important to return anything ever. So I just removed all returns :)

PR #1546, opened 2020-02-12
discv4: Address a few XXXs and FIXMEs I'd left behind

gsalgado:
- Drop Node.last_pong and instead keep track of valid bonds in
  DiscoveryService. This is how it should've been implemented in
  the first place because Nodes are not singletons and multiple
  coroutines may create separate instances during the initial bonding
  process, which can lead to some not realizing when the bonding
  completes and thus treating the remote as one without a valid bond

- Combined addition of nodes to DB and routing table in a single method,
  to ensure they're kept in sync

- Changed APIs for scheduling ENR retrieval to take NodeIDs instead of
  Nodes

- Got rid of _lookup_node_from_db(), which was meant to be temporary

- When receiving a NEIGHBOURS response we no longer need to load
  existing Node entries from the DB just to see if we've bonded with them
  already

- When we receive a ping from a Node for which we don't have a valid
  bond, we now will also trigger a bond in the background. This was
  is the behaviour defined in the spec, but we were not doing it

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -159,6 +159,7 @@ class DiscoveryService(Service):
         self._local_enr_next_refresh: float = time.monotonic()
         self._local_enr_lock = trio.Lock()
         self._lookup_lock = trio.Lock()
+        self.pong_times: Dict[NodeID, float] = {}
         self.parity_pong_tokens: Dict[Hash32, Hash32] = {}
         if socket.family != trio.socket.AF_INET:
             raise ValueError("Invalid socket family")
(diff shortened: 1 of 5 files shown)


p2p/discovery.py
```diff
@@ -253,25 +261,28 @@ def run_daemons_and_bootstrap(self) -> None:
 
     async def fetch_enrs(self) -> None:
         async with self.pending_enrs_consumer:
-            async for (remote, enr_seq) in self.pending_enrs_consumer:
-                self.logger.debug2("Received request to fetch ENR for %s", remote)
-                self.manager.run_task(self._ensure_enr, remote, enr_seq)
+            async for (remote_id, enr_seq) in self.pending_enrs_consumer:
+                self.logger.debug2("Received request to fetch ENR for %s", encode_hex(remote_id))
+                self.manager.run_task(self._ensure_enr, remote_id, enr_seq)
```
pipermerriam: doesn't have to be this PR, but for methods like this that are long running and spawn background tasks, I think it is appropriate to run a nursery and just use `nursery.start_soon`.  The `run_task` APIs have a decent bit of machinery that isn't necessary in cases like this which should reduce the overall load on the *even loop*.

gsalgado: Ok, I'll submit a separate PR for that

PR #1547, opened 2020-02-13
discv4: Fix the way we randomly iterate over peer candidates

gsalgado:
We recently changed DiscoveryService to iterate over the nodes in the
routing table, applying a caller-provided function to skip undersirable
ones, instead of simply blindly asking for a batch of random peers from
the RT. However, we were still using the get_random_nodes(count) API to
iterate over all nodes, which was insanely inneficient when iterating
over all the nodes, but also buggy because the RT's buckets can mutate
during iteration, leading to various sorts of issues.

This fixes that by creating a new API (iter_randmo()), which copies the
RT's buckets, shuffling all available nodes and then iterating over
them.

### DIFF
--- a/p2p/kademlia.py
+++ b/p2p/kademlia.py
@@ -349,28 +349,14 @@ class RoutingTable:
         self.this_node_id_int = big_endian_to_int(node_id)
         self.buckets = [KBucket(0, constants.KADEMLIA_MAX_NODE_ID)]
 
-    def get_random_nodes(self, count: int) -> Iterator[NodeAPI]:
-        if count > len(self):
-            if time.monotonic() - self._initialized_at > 30:
-                self.logger.warning(
-                    "Cannot get %d nodes as RoutingTable contains only %d nodes",
-                    count,
-                    len(self),
-                )
-            count = len(self)
-        seen: List[NodeAPI] = []
-        # This is a rather inneficient way of randomizing nodes from all buckets, but even if we
-        # iterate over all nodes in the routing table, the time it takes would still be
-        # insignificant compared to the time it takes for the network roundtrips when connecting
-        # to nodes.
-        while len(seen) < count:
-            bucket = random.choice(self.buckets)
-            if not bucket.nodes:
-                continue
-            node = random.choice(bucket.nodes)
-            if node not in seen:
-                yield node
-                seen.append(node)
+    def iter_random(self) -> Iterator[NodeAPI]:
+        # Create a new list with all available nodes as buckets can mutate while we're iterating.
+        # This shouldn't use a significant amount of memory as the new list will keep just
+        # references to the existing Node instances.
+        nodes = list(itertools.chain(*[bucket.nodes for bucket in self.buckets]))
+        random.shuffle(nodes)
+        for node in nodes:
+            yield node
 
     def split_bucket(self, index: int) -> None:
         bucket = self.buckets[index]
--- a/tests/p2p/test_kademlia.py
+++ b/tests/p2p/test_kademlia.py
@@ -127,20 +127,30 @@ def test_routingtable_neighbours():
         assert node_a == table.neighbours(node_b.id)[0]
 
 
-def test_routingtable_get_random_nodes():
+def test_routingtable_iter_random():
     table = RoutingTable(NodeFactory().id)
+    nodes_in_insertion_order = []
+    # Use a relatively high number of nodes here otherwise we could have two consecutive calls
+    # yielding nodes in the same order.
     for _ in range(100):
-        assert table.add_node(NodeFactory()) is None
+        node = NodeFactory()
+        assert table.add_node(node) is None
+        nodes_in_insertion_order.append(node)
 
-    nodes = list(table.get_random_nodes(50))
-    assert len(nodes) == 50
-    assert len(set(nodes)) == 50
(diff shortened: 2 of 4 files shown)


tests/p2p/test_kademlia.py
```diff
@@ -127,20 +129,24 @@ def test_routingtable_neighbours():
...
+        nodes_in_insertion_order.append(node)
+
+    nodes_in_iteration_order = []
+    for node in table.iter_random():
+        nodes_in_iteration_order.append(node)
+
+    # We iterate over all nodes
+    assert len(nodes_in_iteration_order) == len(table) == len(nodes_in_insertion_order)
+    # No repeated nodes are returned
+    assert len(list(toolz.unique(nodes_in_iteration_order))) == len(nodes_in_iteration_order)
```
carver: Just curious: Why the switch from `set()` to `toolz.unique()` here?

tests/p2p/test_kademlia.py
```diff
@@ -127,20 +129,24 @@ def test_routingtable_neighbours():
...
+    # The order in which we iterate is not the same as the one in which nodes were inserted.
+    assert nodes_in_iteration_order != nodes_in_insertion_order
```
carver: I guess this could technically fail randomly, but 1 in `100!` is small enough... Maybe worth a comment that the number of nodes created is important, because if someone decides to drop the number of created nodes to 3, it will start being flaky.

p2p/kademlia.py
```diff
@@ -349,28 +349,14 @@ def __init__(self, node_id: NodeID) -> None:
...
-                yield node
-                seen.append(node)
+    def iter_random(self) -> Iterator[NodeAPI]:
+        # Create a new list with all available nodes as buckets can mutate while we're iterating.
+        # This shouldn't use a significant amount of memory as the new list will keep just
+        # references to the existing Node instances.
+        nodes = list(itertools.chain(*[bucket.nodes for bucket in self.buckets]))
+        random.shuffle(nodes)
+        for node in nodes:
+            yield node
```
carver: Is there an upper bound on the total number of nodes? (I can't remember the Kademlia details)

Obviously, if the list of nodes gets long enough, then it could take a while to shuffle them all, and it would be wasted time, if only a few nodes are needed. By the time you hit 100k, it's noticeable, and 1M is problematic:
```
In [10]: %timeit random.shuffle(list(range(1000)))                                                                                                                                                                                         
444 µs ± 4.19 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [11]: %timeit random.shuffle(list(range(100000)))                                                                                                                                                                                       
53.4 ms ± 1.55 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [12]: %timeit random.shuffle(list(range(1000000)))                                                                                                                                                                                      
644 ms ± 6.09 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Anyway, this is clearly an improvement. No objection. Just wanted to understand what the worst case starts to look like.

tests/p2p/test_kademlia.py
gsalgado: Hmm, no idea why I didn't use a set, to be honest. Will change

tests/p2p/test_kademlia.py
gsalgado: I think maybe a better check here would be to assert that the result of two consecutive calls contain the same items but not in the same order? That'd make it less likely to fail randomly and check what we actually care about

p2p/kademlia.py
gsalgado: Yeah, the upper bound is 16*256, so pretty low

tests/p2p/test_kademlia.py
carver: You mean iterating through twice? Well, it would reduce the chances again, but not to 0. I'm not sure it's a big enough difference that it's worth the change. I really think it's fine. `100!` is gigantic. It's bigger than the 256-bit keyspace. Just a comment near the assertion or the `range(100)` to help people catch that subtlety seems like enough.

p2p/kademlia.py
carver: Yeah, which tops out at around 2ms on my machine. :+1: for now.

carver:
Oh, and don't forget the `newsfragments/` thing which ends up in the changelog. I guess it would go under `newsfragments/1547.performance.rst`

tests/p2p/test_kademlia.py
gsalgado: Right, but what we really want to ensure is that the iteration order varies between calls, otherwise we could introduce a bug that makes the order always the same, although different from the insertion order, and this test would never catch that.

carver: Ah, got it. :+1: for adding another test that tests iteration order varies between two iterations.

gsalgado:
> Oh, and don't forget the newsfragments/ thing which ends up in the changelog. I guess it would go under newsfragments/1547.performance.rst

Users of previous releases won't see any improvement in performance because of this, as until a few days ago we were only asking for small batches of candidates so the inefficiency of the iteration insignificant. And even in those few days since I changed how we get conn candidates it was not affecting the performance in any way that users could perceive.

carver:
Yup, no change since the last release is a classic "no notes" reason. :+1:

PR #1548, opened 2020-02-13
Catch more exceptions

carver:
### What was wrong?

Clean up random uncaught exceptions.

### How was it fixed?

Mostly just catching and logging them.

### To-Do

- [x] Clean up commit history
- [x] Add release notes for #1545 

#### Cute Animal Picture

### DIFF
--- a/p2p/auth.py
+++ b/p2p/auth.py
@@ -16,6 +16,7 @@ from eth_keys import (
 )
 
 from eth_hash.auto import keccak
+from eth_utils import ValidationError
 
 from cancel_token import CancelToken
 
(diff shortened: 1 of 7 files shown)


p2p/auth.py
```diff
@@ -119,6 +120,11 @@ class HandshakeBase:
     def __init__(
             self, remote: NodeAPI, privkey: datatypes.PrivateKey,
             use_eip8: bool, token: CancelToken) -> None:
+        if remote is None:
+            raise ValidationError("Cannot create handshake with None remote")
+        elif remote.address is None:
+            raise ValidationError("Cannot create handshake with remote address=None")
```
carver: @gsalgado Is this a reasonable constraint?

I had an issue later on in the handshake where `remote.address` was `None` and raised an uncaught exception. I wanted to find out earlier when this (seemingly) problematic remote was created.

gsalgado: The ENR spec does not require an IP address to be included, and I've seen some nodes send us ENRs without that, so yeah, they end up with `None` as their `.address`. Maybe I should change the `should_skip()` function used when looking up connection candidates to skip nodes with no address? Still, I think it'd make sense to guard against that here

carver: Yeah, seems like the ENR is unusable for connecting if it doesn't have an IP address. So :+1: for updating `should_skip()`

PR #1549, opened 2020-02-13
Beam pivot

### LINKED ISSUE #908: Beam Sync: pivot to latest block if the lag gets too high

carver:
### What was wrong?

Fixes #908 

### How was it fixed?

Track the difference between the currently importing block and the most recent announced block from peers. When the gap gets too big, cancel the active import, close down the entire `BeamSyncer`, and restart it from scratch.

### To-Do

- [x] Clean up commit history
- [x] ~Open an issue to make sure hanging imports get cancelled~ This was a requirement for the PR, anyway
- [x] Rebase on #1548

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/chain.py
+++ b/trinity/sync/beam/chain.py
@@ -8,7 +8,10 @@ from typing import (
 
 from lahja import EndpointAPI
 
-from cancel_token import CancelToken
+from cancel_token import (
+    CancelToken,
+    OperationCancelled,
+)
 from eth.abc import AtomicDatabaseAPI, DatabaseAPI
 from eth.constants import GENESIS_PARENT_HASH
 from eth.exceptions import (
@@ -35,6 +38,7 @@ from trinity.db.eth1.header import BaseAsyncHeaderDB
 from trinity.protocol.eth.peer import ETHPeerPool
 from trinity.protocol.eth.sync import ETHHeaderChainSyncer
 from trinity.sync.beam.constants import (
+    BEAM_PIVOT_BUFFER_FRACTION,
     ESTIMATED_BEAMABLE_SECONDS,
     FULL_BLOCKS_NEEDED_TO_START_BEAM,
 )
@@ -50,9 +54,9 @@ from trinity.sync.common.events import (
     CollectMissingStorage,
     DoStatelessBlockImport,
     DoStatelessBlockPreview,
-    MissingAccountCollected,
-    MissingBytecodeCollected,
-    MissingStorageCollected,
+    MissingAccountResult,
+    MissingBytecodeResult,
+    MissingStorageResult,
 )
 from trinity.sync.common.headers import (
     HeaderSyncerAPI,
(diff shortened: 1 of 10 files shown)


trinity/sync/beam/chain.py
```diff
@@ -228,6 +246,25 @@ def __init__(
         # run sync until cancelled
         await self.cancellation()
 
+    async def _trigger_pivot_on_lag(self) -> None:
```
cburgdorf: I think that conceptually the watching of pivot points would be better suited outside of this class. I find the `did_request_pivot` to be a bit of a code smell. I think this code could live in something like a `BeamPivotWatcher` that would be instantiated *next to* the `BeamSyncer` in the `BeamSyncService` and then *controls* the `BeamSyncer` which still just means tearing it down and spinning it up fresh. But you wouldn't need the `did_request_pivot` anymore and the `BeamSyncer` instance would not know why it was canceled.

That said, the logic is small enough that I also think it is ok to leave it as is for now.

cburgdorf: Oh just to maybe add more evidence: If we imagine we still had `fast` sync, then we would also need to pivot for the state sync and probably the logic would overlap. So it could be that one `StateSyncPivotWatcher` could be used for `beam` and `fast` sync. Obviously, this is just hypothetically as we have no plans to bring back `fast` sync.

trinity/sync/beam/constants.py
```diff
@@ -41,11 +41,19 @@
...
-# About how many seconds after a block can we request trie data from peers?
+# About how long after a block can we request trie data from peers?
 # The value is configurable by client, but tends to be around 120 blocks.
-# To make up for clients that are configured low, unusually low block times,
-# and other surprises, we pick half of the 15-second * 120 blocks time window:
-ESTIMATED_BEAMABLE_SECONDS = 900
+ESTIMATED_BEAMABLE_BLOCKS = 120
+
+# It's also useful to estimate the amount of time covered by those beamable blocks.
+PREDICTED_BLOCK_TIME = 14
```
cburgdorf: That's probably ok for now but might be better to actually calculate a moving average to react better to actual network conditions (e.g. ice age etc)

trinity/sync/beam/chain.py
carver: Yeah, this didn't necessarily feel like the best solution right away. Some other options, and considerations about them:
- independent `StateSyncPivotWatcher`
  - Does it make sense to expose the Beam Sync lag as a public API of `BeamSyncer`?
  - After cancellation, the `BeamSyncService` needs to know whether `BeamSyncer` was cancelled due to lag or other reasons, so I guess the pivot watcher has to save that flag and expose it? Starts to feel like a lot of machinery for a small win.
- independent `BeamPivoter` whose only job is to watch the lag and restart the syncer on too much lag
  - This would be another tiny class, and probably not reusable for Fast Sync directly anyway
- Monitor the lag in `BeamSyncService` and use that to decide whether to restart
  - BeamSyncService seems like a conceptually good spot for this. Doesn't add much indirection.
  - It's a small change
  - Still exposes a new lag API, but removes the `beam_syncer.did_request_pivot` API, so net neutral on new APIs created

I'm currently leaning toward the last option.

cburgdorf: Yeah, last option sounds good to me :+1:

PR #1550, opened 2020-02-13
NodeDB.update() now always updates when seq number is 0

gsalgado:
This is needed because we use ENRs with sequence number 0 for v4 nodes
that don't support the ENR extension, and whenever one of those changes
their address we'll create a new ENR with sequence number 0 and want
that to override the existing one. We can't increase the sequence number
ourselves for those stub ENRs because that'd become a problem when those
nodes start supporting the ENR extension


gsalgado:
@jannikluhn I believe this won't affect discv5, so should be ok?

PR #1554, opened 2020-02-14
Skip connection candidates that have no endpoint information

gsalgado:
As we now no longer keep ENRs with no endpoint info (#1553),
this is just a safety net and I'm not that thrilled about
writing a test for it. Hope you'll forgive me ;-)


carver:
*()&#$*% `EOFError`

PR #1559, opened 2020-02-17
New NodeDB implementation

gsalgado:
- Uses a key/value DatabaseAPI backend, provided by callsites
- No longer keeps multiple ENR versions for a given remote
- All APIs are now synchronous
- For now can store ENRs and pong times


gsalgado:
Do you see any issues with using this new implementation on discv5, @jannikluhn? This PR includes #1553 and a few other things, but https://github.com/ethereum/trinity/pull/1559/commits/24715290895ca60e22dc4a32cc6cc9422182dbcf has just the new nodedb implementation

gsalgado:
I've added a few more tests and removed the old MemoryNodeDB/FileNodeDB. Would you like to have another look at it, @jannikluhn ?

Issue #1561, opened 2020-02-17
Need support for API keys when making etherscan requests


gsalgado:
It looks like they have changed their policy and no longer accept requests without API keys, so tests/integration/test_etherscan_checkpoint_resolver.py is now failing (https://circleci.com/gh/ethereum/trinity/192445?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

@cburgdorf suggested an environment variable (https://github.com/ethereum/trinity/pull/1553#pullrequestreview-359608727), which sounds good to me

PR #1565, opened 2020-02-18
Add support for Etherscan API keys

cburgdorf:
### What was wrong?

From February 15th, Etherscan [has made API keys mandatory for all API access](https://medium.com/etherscan-blog/psa-for-developers-implementation-of-api-key-requirements-starting-from-february-15th-2020-b616870f3746) (See #1561).

Hence we need to support API keys for `--beam-from-checkpoint eth://block/byetherscan/latest` to continue to work.

### How was it fixed?

1. Read the API key from the `XDG_TRINITY_ETHERSCAN_API_KEY` (wasn't sure about the `XDG` prefix or even about the Trinity prefix. Should this be just `ETHERSCAN_API_KEY` instead?)

2. Refactor the loose etherscan api functions into a `Etherscan` class that cashes the API key.

This also moves the `request` and `idna` dependency from `dev` to `trinity`. The `request` library is directly used by Trinity and so having it only in `dev` breaks Trinity if it isn't installed with the `dev` extras (read, for the regular user). This issue was exposed just now because the `idna` library that `requests` depends on had a new [release](https://pypi.org/project/idna/#history) yesterday which now caused the integration tests to fail because the `idna` resolution was only specified in `dev`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/components/builtin/syncer/cli.py
+++ b/trinity/components/builtin/syncer/cli.py
@@ -1,4 +1,5 @@
 import argparse
+import os
 from pathlib import PurePosixPath
 import re
 import urllib
@@ -22,8 +23,7 @@ from eth_utils import (
 from trinity.sync.common.checkpoint import Checkpoint
 
 from .etherscan_api import (
-    get_block_by_number,
-    get_latest_block,
+    Etherscan,
     Network,
 )
 
@@ -64,9 +64,19 @@ def parse_byetherscan_uri(parsed: urllib.parse.ParseResult, network_id: int) -> 
             f"for network {network_id}. Network not supported"
         )
 
-    latest_block_number = get_latest_block(network)
+    try:
+        etherscan_api_key = os.environ['TRINITY_ETHERSCAN_API_KEY']
+    except KeyError:
+        raise RuntimeError(
+            "Etherscan API key missing. Assign your Etherscan API key "
+            "to the TRINITY_ETHERSCAN_API_KEY environment variable."
+        )
+
+    etherscan_api = Etherscan(etherscan_api_key)
+
+    latest_block_number = etherscan_api.get_latest_block(network)
     checkpoint_block_number = latest_block_number - BLOCKS_FROM_TIP
-    checkpoint_block_response = get_block_by_number(checkpoint_block_number, network)
+    checkpoint_block_response = etherscan_api.get_block_by_number(checkpoint_block_number, network)
     checkpoint_score = to_int(hexstr=checkpoint_block_response['totalDifficulty'])
     checkpoint_hash = checkpoint_block_response['hash']
 
@@ -119,6 +129,6 @@ class NormalizeCheckpointURI(argparse.Action):
 
         try:
             parsed = parse_checkpoint_uri(value, namespace.network_id)
-        except ValidationError as exc:
+        except (ValidationError, RuntimeError) as exc:
             raise argparse.ArgumentError(self, str(exc))
         setattr(namespace, self.dest, parsed)
--- a/trinity/components/builtin/syncer/etherscan_api.py
+++ b/trinity/components/builtin/syncer/etherscan_api.py
@@ -35,39 +35,44 @@ API_URLS = {
 }
 
 
-def etherscan_proxy_api_url(network: Network) -> str:
-    return f"{API_URLS[network]}?module=proxy"
+class Etherscan:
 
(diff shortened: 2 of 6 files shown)


cburgdorf:
Mmh..Circle CI says the environment var is set. Maybe it needs a bit to propagate. I'll kick the build after lunch and see if the error persists.

cburgdorf:
Ah, I needed to expose it through tox..

trinity/components/builtin/syncer/cli.py
```diff
@@ -64,9 +64,19 @@ def parse_byetherscan_uri(parsed: urllib.parse.ParseResult, network_id: int) ->
             f"for network {network_id}. Network not supported"
         )
 
-    latest_block_number = get_latest_block(network)
+    try:
+        etherscan_api_key = os.environ['XDG_TRINITY_ETHERSCAN_API_KEY']
+    except KeyError:
+        raise ValidationError(
```
gsalgado: ValidationError doesn't seem like the most appropriate exception here. It could be caught and ignored (i.e. logged as a debug msg) somewhere up the stack. Maybe a `RuntimeError`?

trinity/components/builtin/syncer/etherscan_api.py
```diff
@@ -35,39 +35,44 @@ class Network(enum.IntEnum):
 }
 
 
-def etherscan_proxy_api_url(network: Network) -> str:
-    return f"{API_URLS[network]}?module=proxy"
+class Etherscan:
 
+    def __init__(self, api_key: str):
```
gsalgado: How come mypy didn't complain that this method defines no return value?

trinity/components/builtin/syncer/etherscan_api.py
```diff
@@ -35,39 +35,44 @@ class Network(enum.IntEnum):
...
 
-def etherscan_post(action: str, network: Network) -> Any:
-    response = requests.post(f"{etherscan_proxy_api_url(network)}&action={action}")
+        self.api_key = api_key
 
-    if response.status_code not in [200, 201]:
-        raise EtherscanAPIError(
-            f"Invalid status code: {response.status_code}, {response.reason}"
-        )
+    def etherscan_proxy_api_url(self, network: Network) -> str:
```
gsalgado: Should this method have a `get_` prefix? I believe our tactical manual recommends that, no?

gsalgado: We should probably get rid of the `etherscan_` prefix now that it's under the `Etherscan` class

trinity/components/builtin/syncer/etherscan_api.py
```diff
@@ -35,39 +35,44 @@ class Network(enum.IntEnum):
...
+        return f"{API_URLS[network]}?module=proxy&apikey={self.api_key}"
 
-    try:
-        value = response.json()
-    except ValueError as err:
-        raise EtherscanAPIError(f"Invalid response: {response.text}") from err
+    def etherscan_post(self, action: str, network: Network) -> Any:
```
gsalgado: Same here about the `etherscan_` prefix

trinity/components/builtin/syncer/etherscan_api.py
cburgdorf: It's a [special `mypy` rule](https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods) that `__init__` doesn't need a return value annotation. But I'll add it anyway to be more in line with the rest of the code base.

trinity/components/builtin/syncer/etherscan_api.py
cburgdorf: Good point, I'll make that `get_proxy_api_url` and drop the `etherscan_` prefix everywhere.

trinity/components/builtin/syncer/cli.py
cburgdorf: Yeah, now that you point it out `ValidationError` doesn't seem correct. I turned this into a `RuntimeError` and make sure to catch it here.

https://github.com/ethereum/trinity/blob/476d6af2b7ec7733cb3aaa446ed2f4c6456d9adb/trinity/components/builtin/syncer/cli.py#L110-L124

The `RuntimeError` is caught among the other validation errors that can happen here and then turned into an `argparse.ArgumentError` to have Trinity shutdown and present the error to the user.

PR #1566, opened 2020-02-18
Use CouldNotRetrieveENR instead of AttributeError

carver:
### What was wrong?

When the address of a node is None, it was raising an AttributeError, because it tried to get node.address.ip. The AttributeError isn't caught, and crashes all of trinity.

### How was it fixed?

Raise a CouldNotRetrieveENR, which is handled much more gracefully, with a debug log.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -842,6 +842,9 @@ class DiscoveryService(Service):
             pass
 
     def send_enr_request(self, node: NodeAPI) -> Hash32:
+        if node.address is None:
+            raise CouldNotRetrieveENR(f"Cannot ask for ENR from {node} with no address")
+
         message = self.send(node, CMD_ENR_REQUEST, [_get_msg_expiration()])
         token = Hash32(message[:MAC_SIZE])
         self.logger.debug("Sending ENR request with token: %s", encode_hex(token))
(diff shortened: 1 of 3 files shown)


p2p/discovery.py
```diff
@@ -842,6 +842,9 @@ def _is_msg_expired(self, rlp_expiration: bytes) -> bool:
             pass
 
     def send_enr_request(self, node: NodeAPI) -> Hash32:
+        if node.address is None:
```
gsalgado: This should actually only happen if you have a node DB populated before https://github.com/ethereum/trinity/commit/0474fa707d0b4c2a42cd3c8ecc6ab37b1566037b was merged, so I'd prefer if we raised a different exception here because the `CouldNotRetrieveENR` would effectively mask a bug in our code

carver: How about logged WARNING instead? Without any change, it crashes trinity, which is a pretty harsh penalty.

gsalgado: Agreed it's a harsh penalty, but there was only a short time window where a node without an address could end up in the DB (and no releases during that, AFAICT), and there are maaaany other places that would crash if that happen, so I don't think it's worth checking it just here. Having said that, if you feel strong about handling that here, I think we should at least log it as an error and not a warning

gsalgado: And ideally the log message should instruct users to try removing their NodeDB and report a bug if the error persists

carver:
I'm fine handling like https://github.com/ethereum/trinity/issues/1570 -- since the "buggy" database was never released. We just tell people to drop their nodes database if they run into the same issue.

PR #1569, opened 2020-02-20
Collect runtime metrics and push them into InfluxDB to be used by Grafana

cburgdorf:
### What was wrong?

We don't have infrastructure in place to collect runtime metrics. Collecting various metrics helps tremendously to better understand the impact of changes ("Oh turns out our 2x improvement here causes a 4x regression there")

As an example, here's a [Geth PR with tons of shiny charts](https://github.com/karalabe/go-ethereum/pull/27) that demonstrate the impact of the PR.

I recently had a very constructive discussion with the Geth team and they have invited us to use their existing influxdb/grafana infrastructure to collect metrics for Trinity.

They went ahead and setup a DB and Grafana user for us and I decided to shift my attention there to get something off the ground.

### How was it fixed?

1. The general guideline for the implementation was "copy whatever geth does". This is based on the fact that they've already build up tons of knowledge in this area, invested many hours in the system including creating many different charts etc. and we can stand on the shoulders of giants if we just start copying what they do.

2. The implementation uses [pyformance](https://pyformance.readthedocs.io/en/latest/usage.html) which is the closest to [go-metrics](https://pkg.go.dev/github.com/rcrowley/go-metrics) that Geth uses. It provides Gauges, Counters, Meters etc as well as different reporters including one that reports straight to InfluxDB.

3. This adds a Metrics component that runs in its own process to collect system metrics similar to [this in Geth](https://github.com/ethereum/go-ethereum/blob/556888c4a971d4fd02be15f09cba7a65c5c0d930/metrics/metrics.go#L76-L126)

This does not yet add metrics that are concerned with more Trinity specific things such as peer connections etc. However, I think I have an idea about how that would work out. It's a little different for us compared to Geth because Trinity uses many different processes. I think that the `MetricsComponent` should be used to collect all *system wide*  metrics but that other metrics should just be collected and reported via additional `MetricService` instances in those other component processes. The alternative would be to send metrics cross processes via the eventbus to push them all at once but I see no benefit in that. I think that would just introduce lots of complexity and load in the form of inter process communication with the only benefit being a little better batching of data towards influxdb.

The following dashboard is a straight clone from one of the  Geth dashboards and it shows two connected charts with the rest of the charts just being a stub of metrics that we probably want to add in the future.

image

Also node that due to the nature of Trinitys multi process architecture there are quite a few interesting Trinity specific charts that we could also add in the future.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1569.feature.rst
+++ b/newsfragments/1569.feature.rst
@@ -1,0 +1,9 @@
+Implement basic metrics tracking. With this change Trinity supports collecting
+and writing metrics to InfluxDB to be further processed and visualized
+by Grafana.
+
+Run Trinity with ``--metrics --metrics-host "mymachine"`` and provide the
+Influx server and password via the following two environment variables:
+
+- ``TRINITY_METRICS_INFLUX_DB_SERVER``
+- ``TRINITY_METRICS_INFLUX_DB_PW``
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,9 @@ deps = {
         "argcomplete>=1.10.0,<2",
         "multiaddr>=0.0.8,<0.1.0",
         "prometheus-client==0.7.1",
+        "pyformance==0.4",
         "pymultihash>=0.8.2",
+        "psutil>=5.7.0, <6",
         "libp2p==0.1.4",
         # The direct dependency resolves a version conflict between multiaddr and libp2p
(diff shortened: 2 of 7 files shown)


cburgdorf:
This landed as part of #1595

PR #1572, opened 2020-02-24
UPnPService now broadcasts an event when it creates a portmap

gsalgado:
Also adds the boilerplate necessary to launch the UPnP component by
itself, as well as a helper script that listens for NewUPnPMapping
events and prints them to stdout.

We have no tests for UPnPService as that would probably require mocking
a ton of stuff, so there's no straightforward way to test that it emits
the events either, but at least now we can launch the component by
itself and there's a script to print the events it emits.

### DIFF
--- a/trinity/components/builtin/upnp/nat.py
+++ b/trinity/components/builtin/upnp/nat.py
@@ -1,32 +1,29 @@
 import asyncio
 from concurrent.futures import ThreadPoolExecutor
 import ipaddress
+import netifaces
 from typing import (
     AsyncGenerator,
-    NamedTuple,
 )
 from urllib.parse import urlparse
 
+import upnpclient
+
+from lahja import EndpointAPI
+
 from async_service import Service
+
 from eth_utils import get_extended_debug_logger
 
 from p2p.exceptions import (
     NoInternalAddressMatchesDevice,
 )
-import netifaces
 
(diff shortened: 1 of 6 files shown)


trinity/components/builtin/upnp/nat.py
```diff
@@ -73,7 +79,14 @@ def __init__(self, port: int) -> None:
         """
         while self.manager.is_running:
             try:
-                await self.add_nat_portmap()
+                external_ip = await self.add_nat_portmap()
+                if external_ip is not None:
+                    event = NewUPnPMapping(external_ip)
+                    self.logger.info(
+                        "NAT portmap created, broadcasting NewUPnPMapping event: %s", event)
+                    self.event_bus.broadcast_nowait(NewUPnPMapping(external_ip))
```
cburgdorf: Wondering if this should be `await self.event_bus.broadcast(NewUPnPMapping(external_ip))`. I think the recommendation is to only use the synchronous `no_wait` when we are in a synchronous context and can't easily work around it. The `async` version has the advantage of getting the backpressure and hence not running into a situation where we would broadcast more events than receivers can handle (not that this would be a problem for these rare events).

gsalgado:
@cburgdorf I've pushed another commit with the change you suggested plus the removal of some stuff that was no longer needed

Issue #1571, opened 2020-02-24
Must detect external IP address and use that in our ENR


gsalgado:
When we generate our local ENR we use the IP address in which DiscoveryService listens to (i.e. `0.0.0.0`). We need to somehow figure out the actual IP address that other nodes can use to connect to us, which can be either the external address of our NAT port-mapping (if one exists) or the address of any(?) externally-facing local interface.

Also, sice the NAT port-mapping is re-created periodically, we will probably need to change UPnPService to send events when that happens so that DiscoveryService can update our local ENR. 

Another complication is that DiscoveryService currently updates our local ENR upon startup, so it might make sense to change it to wait a bit for an event about a NAT port-forwarding being setup, and if one doesn't arrive, use the address of any externally-facing interface.

gsalgado:
@jannikluhn I see that in discv5 this problem has been avoided by not including any endpoint info in the local ENR. Have you thought about how to address it there?

jannikluhn:
> @jannikluhn I see that in discv5 this problem has been avoided by not including any endpoint info in the local ENR. Have you thought about how to address it there?

Yes, this is in fact specified by the v5 spec, see https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md#maintaining-the-local-node-record

We follow the `PONG` aproach: Whenever we receive a PONG, we emit an `EndpointVote` and send it to the [EndpointTracker](https://github.com/ethereum/trinity/blob/master/p2p/discv5/endpoint_tracker.py#L48). Once we have votes from enough peers for a particular endpoint, the tracker updates the ENR.

It looks like ip and port are included in [v4 pings](https://github.com/ethereum/devp2p/blob/master/discv4.md#ping-packet-0x01), so something similar might be possible for v4 as well.

gsalgado:
One problem with using the PING msg for that is the fact that in discv4, the only situation in which a remote will ping us is when the last pong they got from us is more than 12h old, so we may end up in a situation where we get no pings for a long time. I'll see what geth does on discv4 wrt maintaining the local ENR

PR #1574, opened 2020-02-25
Use internet-facing address in our local ENR

gsalgado:
Upon startup, DiscoveryService will now use the first internet-facing
address it finds on any network interfaces, falling back to the loopback
address if none is found.

It will also watch for any NewUPnPMapping and use that to update our
local ENR.

Closes: #1571

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -6,6 +6,8 @@ listening nodes.
 More information at https://github.com/ethereum/devp2p/blob/master/rlpx.md#node-discovery
 """
 import collections
+import ipaddress
+import netifaces
 import random
 import secrets
 import time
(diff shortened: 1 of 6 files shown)


p2p/discovery.py
```diff
@@ -403,14 +422,31 @@ def schedule_enr_retrieval(self, node_id: NodeID, enr_seq: int) -> None:
...
+    def get_external_ipaddress(self) -> ipaddress.IPv4Address:
+        for iface in netifaces.interfaces():
+            for family, addresses in netifaces.ifaddresses(iface).items():
+                if family != netifaces.AF_INET:
+                    continue
+                for item in addresses:
+                    iface_addr = ipaddress.ip_address(item['addr'])
+                    if iface_addr.is_global:
+                        return iface_addr
+        self.logger.info(
```
cburgdorf: Nothing in this method except the logging uses `self` which seems like hint that this method might be better suited as a standalone function either just at the top of this file or maybe in `p2p._utils`.

gsalgado: Fair enough, but as it is only used here I'd rather keep it in this file

PR #1575, opened 2020-02-25
A couple trivial discv4 changes

gsalgado:
- Log a warning if a node sends us an ENR with seq number prior to the
  one we have in the DB
- Replace a node's ENR from our DB if they send us a message using a
  different addr.

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -652,13 +652,19 @@ class DiscoveryService(Service):
         try:
             enr = self.node_db.get_enr(node_id_from_pubkey(remote_pubkey))
         except KeyError:
-            node = None
+            node = Node.from_pubkey_and_addr(remote_pubkey, address)
         else:
             node = Node(enr)
-        # If we don't have an entry for the sender's NodeID in our DB, or if the one we do is
-        # incomplete (i.e. without an Address), create a new one.
-        if node is None or node.address is None or node.address != address:
-            node = Node.from_pubkey_and_addr(remote_pubkey, address)
+            if node.address != address:
+                self.logger.warning(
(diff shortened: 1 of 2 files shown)


p2p/discovery.py
```diff
@@ -816,6 +822,16 @@ def _is_msg_expired(self, rlp_expiration: bytes) -> bool:
         self.logger.debug(
             "Received ENR %s (%s) with expected response token: %s",
             enr, enr.items(), encode_hex(token))
+        existing_enr = self.node_db.get_enr(enr.node_id)
+        if enr.sequence_number < existing_enr.sequence_number:
+            self.logger.warning(
```
gsalgado: @carver apart from a corrupt DB (as you seemed to have), this is the only other way we could get a ValueError when trying to store an ENR

p2p/discovery.py
```diff
@@ -652,13 +652,19 @@ def send(self, node: NodeAPI, msg_type: DiscoveryCommand, payload: Sequence[Any]
...
         except KeyError:
-            node = None
+            node = Node.from_pubkey_and_addr(remote_pubkey, address)
         else:
             node = Node(enr)
-        # If we don't have an entry for the sender's NodeID in our DB, or if the one we do is
-        # incomplete (i.e. without an Address), create a new one.
-        if node is None or node.address is None or node.address != address:
-            node = Node.from_pubkey_and_addr(remote_pubkey, address)
+            if node.address != address:
```
gsalgado: Now we can no longer have ENRs with no address info, so this doesn't change any behaviour but ensures we delete the ENR that has probably outdated info.

carver:
`EOFError`. Is it getting worse, or just more annoying?

PR #1580, opened 2020-02-27
discv4: Force new bonds with bootnodes when bootstrapping

gsalgado:
Before we started persisting bond information in our NodeDB we'd always
perform a new bond with our bootnodes during bootstrap, ensuring they
get added to our RT so that we can query them but also ensuring they
have our latest ENR. This change restores that behaviour by having
bootstrap() pass force_new_bond=True when bonding with bootnodes

Also turns DiscoveryService.bootstrap_nodes into an `@property` that
always loads the last version of our bootnodes from the DB instead
of the ones passed at instantiation time, which might get outdated.

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -152,7 +152,6 @@ class DiscoveryService(Service):
                  ) -> None:
         self.logger = get_extended_debug_logger('p2p.discovery.DiscoveryService')
         self.privkey = privkey
-        self.bootstrap_nodes = bootstrap_nodes
         self._event_bus = event_bus
         self.enr_response_channels = ExpectedResponseChannels[Tuple[ENR, Hash32]]()
         self.pong_channels = ExpectedResponseChannels[Tuple[Hash32, int]]()
@@ -171,6 +170,19 @@ class DiscoveryService(Service):
         self.socket = socket
         self.pending_enrs_producer, self.pending_enrs_consumer = trio.open_memory_channel[
             Tuple[NodeID, int]](self._max_pending_enrs)
+
+        # Ensure our bootstrap nodes are in our DB and keep only a reference to their IDs. This is
+        # to ensure self.bootstrap_nodes always returns the last known version of them.
+        self._bootstrap_node_ids: List[NodeID] = []
+        for node in bootstrap_nodes:
+            self._bootstrap_node_ids.append(node.id)
+            try:
+                self.node_db.get_enr(node.id)
+            except KeyError:
+                self.node_db.set_enr(node.enr)
+        if len(set(self._bootstrap_node_ids)) != len(self._bootstrap_node_ids):
+            raise ValueError(
+                "Multiple bootnodes with the same ID are not allowed: {self._bootstrap_node_ids}")
 
         # This is a stub Node instance (which will have an ENR with sequence number 0) that will
         # be replaced in _init(). We could defer its creation until _init() is called by run(),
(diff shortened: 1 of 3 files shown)


p2p/discovery.py
```diff
@@ -195,6 +204,18 @@ def __init__(self,
...
+    @property
+    def bootstrap_nodes(self) -> Tuple[NodeAPI, ...]:
+        bootnodes = []
+        for node_id in self._bootstrap_node_ids:
+            try:
+                enr = self.node_db.get_enr(node_id)
+            except KeyError:
+                self.logger.exception("Bootnode not found in our DB")
+            else:
+                bootnodes.append(Node(enr))
```
pipermerriam: You can change this to `yield Node(enr)` and add the `eth_utils.to_tuple` decorator inside the `@property` decorator to remove the accumulator list and tuple casting at the end.

tests-trio/p2p-trio/test_discovery.py
```diff
@@ -415,6 +418,30 @@ def should_skip(skip_list, candidate):
     assert not bonded
 
 
+@pytest.mark.trio
+async def test_bond_short_circuits(nursery, monkeypatch):
```
pipermerriam: The `nursery` doesn't appear to be used in this test body.

p2p/discovery.py
```diff
@@ -343,21 +364,23 @@ def run_daemons_and_bootstrap(self) -> None:
                 stack_info=True,
             )
 
-    async def bond(self, node: NodeAPI) -> bool:
+    async def bond(self, node: NodeAPI, force_new_bond: bool = False) -> bool:
```
pipermerriam: I'm not sure it is better but I tend to look for ways to not have functionality flags like this.

Could we have something like a `revoke_bond(node: NodeAPI)` method, and run all the bootnodes through that before re-bonding with them?  That keeps this function from having the additional logic... and may make your tests easier to write.

p2p/discovery.py
gsalgado: Something like this:
```
    @property
    def bootstrap_nodes(self) -> Tuple[NodeAPI, ...]:
        @to_tuple
        def inner() -> Iterator[NodeAPI]:
            for node_id in self._bootstrap_node_ids:
                try:
                    enr = self.node_db.get_enr(node_id)
                except KeyError:
                    self.logger.exception("Bootnode not found in our DB")
                else:
                    yield Node(enr)
        return inner()
```

I think it's less readable and provides no real benefit, but not a big deal so happy to go with that if you prefer

tests-trio/p2p-trio/test_discovery.py
gsalgado: Indeed, removed

p2p/discovery.py
gsalgado: It doesn't really make the test easier to write as I now need to check that bootstrap() calls yet another method to invalidate the bonds, but I agree it makes more sense to delete the bonds in bootstrap than having the extra flag on bond(), so I'll go with that

p2p/discovery.py
gsalgado: Is this what you had in mind, @pipermerriam ?

cburgdorf: No need to wrap it, this works fine (order of decorators is important)

```python
class Foo:

    @property
    @to_tuple
    def something(self):
        yield 1
        yield 2
```

cburgdorf: Oh, it does need a `type: ignore` on the line of `@property` though. https://github.com/python/mypy/issues/1362 
I'm going to make this change in this PR (I hope you don't mind) before I go and merge this for the upcoming release.

gsalgado: :+1:

cburgdorf:
Interesting, the [CI failure](https://circleci.com/gh/ethereum/trinity/196156) seems to come from a failed `test_logger_configuration` test.

Issue #1560, opened 2020-02-17
Logger hierarchy not guaranteed, leading to unexpected behaviour of log levels specified on the command-line


gsalgado:
Many of our loggers are defined at import time (i.e. module/class level), and although we use fully qualified names (e.g. `p2p.discovery.DiscoveryService`) to have a hierarchy of loggers, there is nothing to ensure the parent loggers exist, so we end up with most of our loggers descending directly from the root logger. That in turn leads to any package/module log level specified in the command line to have no effect. 

For example, if I pass `-l p2p=debug` it will have no effect on my `p2p.discovery.DiscoveryService` logger because it is defined at the class level and so descends directly from the root logger. 

This would be less of a problem if our loggers were defined at runtime (e.g. during class instantiation), but there would still be a chance some of them could be created before all of their parents, leading to the same problem as above.

Maybe we could change our eth_utils logging helpers to automatically create the hierarchies, and require that all our logger be defined using that instead of the stdlib APIs?

pipermerriam:
Can we do something dumb like having something we intentionally hit early in the import cycle which creates loggers for the various namespaces we know we use?

I'm not opposed to updating `eth-utils` but I'm unsure enough about unintended consequences that the dumb solution seems like an easy bandaid.

gsalgado:
The only way I could think of was to somehow traverse the filesystem looking for python packages and creating loggers for those, which would work fine when we run from a dev tree, but not sure how that'd work on an a regular install.

I think the cheapest solution would be to have a policy requiring that loggers be defined at instantiation time, then `main_entry()` can create loggers for any per-package log levels passed via the command line before our actual loggers are instantiated. That way we may not get the whole hierarchy, but at least in the example above the `DiscoveryService` logger would end up as a child of `p2p` and not the root logger

gsalgado:
Or we could wrap the `eth_utils` helpers in trinity to create the missing loggers in the hierarchy and once we're sure they don't have any unintended consequences we move that functionality over to `eth_utils`?

pipermerriam:
I'm comfortable with that approach (wrap locally, export layer once the dust settles)

Sent from ProtonMail mobile

-------- Original Message --------

PR #1583, opened 2020-02-28
New get_logger() helper that ensures logger hierarchy

gsalgado:
One annoying thing here is that even with the extra wrapper in `p2p._utils` (to avoid the module-level cyclic import), we still need to ensure all `get_logger()` calls in the `p2p` package only happen at runtime. Not a big deal as any calls made at the module level will cause test failures anyway

If you're happy with this approach I'll go ahead and update all callsites and add a test for the new wrapper

Closes: #1560

### DIFF
--- a/trinity/_utils/logging.py
+++ b/trinity/_utils/logging.py
@@ -20,6 +20,8 @@ from typing import (
     TypeVar,
 )
 
+import eth_utils
+
 from trinity._utils.shellart import (
     bold_red,
     bold_yellow,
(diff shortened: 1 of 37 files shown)


trinity/_utils/logging.py
```diff
@@ -35,8 +37,15 @@
 THandler = TypeVar("THandler", bound="IPCHandler")
 
 
+def get_logger(name: str) -> eth_utils.ExtendedDebugLogger:
```
gsalgado: I chose to have a single API that always returns an `ExtendedDebugLogger` because otherwise we can't be sure that passing `debug2` as the level in the cli will work for all loggers

pipermerriam:
> we still need to ensure all `get_logger()` calls in the `p2p` package only happen at runtime.

I don't understand why this is the case.

gsalgado:
Because the actual wrapper is in the `trinity` package, so any `get_logger()` calls executed at import time will trigger a cyclical import error. I don't think it's going to be problem as the only places where we have those are classes that define their loggers as class attributes, so we just have to turn them into instance attributes

gsalgado:
@pipermerriam are you happy with this approach?

Issue #1584, opened 2020-02-28
Crash with ListDeserializationError during peer discovery


carver:
* Trinity version: 08c678437 (master)

### What is wrong?

Node shut down with:
```
Traceback (most recent call last):
  File "/usr/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/home/jcarver/code/trinity/trinity/extensibility/trio.py", line 56, in run_process
    trio.run(cls._do_run, boot_info)
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/trio/_core/_run.py", line 1804, in run
    raise runner.main_task_outcome.error
  File "/home/jcarver/code/trinity/trinity/extensibility/trio.py", line 70, in _do_run
    nursery.cancel_scope.cancel()
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/trio/_core/_run.py", line 730, in __aexit__
    raise combined_error_from_nursery
  File "/home/jcarver/code/trinity/trinity/components/builtin/peer_discovery/component.py", line 101, in do_run
    await async_service.run_trio_service(discovery_service)
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/async_service/trio.py", line 156, in run_service
    await manager.run()
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/async_service/trio.py", line 208, in run
    for _, exc_value, exc_tb in self._errors
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/async_service/base.py", line 301, in _run_and_manage_task
    await task.run()
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/async_service/trio.py", line 76, in run
    await self._async_fn(*self._async_fn_args)
  File "/home/jcarver/code/trinity/p2p/discovery.py", line 239, in handle_get_peer_candidates_requests
    candidates = self.get_peer_candidates(event.should_skip_fn, event.max_candidates)
  File "/home/jcarver/code/trinity/p2p/discovery.py", line 222, in get_peer_candidates
    if should_skip_fn(candidate):
  File "/home/jcarver/code/trinity/trinity/protocol/common/peer.py", line 293, in skip_candidate_if_on_list_or_fork_mismatch
[...]

PR #1587, opened 2020-03-01
Do not crash if we get an ENR with an invalid `eth` entry

gsalgado:
Closes: #1584

### DIFF
--- a/tests/core/p2p-proto/test_chain_peer_pool.py
+++ b/tests/core/p2p-proto/test_chain_peer_pool.py
@@ -15,7 +15,7 @@ from p2p.kademlia import Node
 from p2p.tools.factories.discovery import ENRFactory
 from p2p.tools.factories.kademlia import IPAddressFactory, NodeFactory
 
-from trinity.exceptions import BaseForkIDValidationError
+from trinity.exceptions import BaseForkIDValidationError, ENRMissingForkID
 from trinity.protocol.common.peer import skip_candidate_if_on_list_or_fork_mismatch
 from trinity.protocol.eth import forkid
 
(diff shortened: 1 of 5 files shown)


cburgdorf:
Merged this to get it in the upcoming release

tests/core/p2p-proto/test_chain_peer_pool.py
```diff
@@ -76,7 +76,8 @@ def test_skip_candidate_if_on_list_or_fork_mismatch():
     assert functools.partial(should_skip_fn, skip_list)(no_forkid_nodes[1]) is True
 
     # It returns False for candidates with no fork-id that are not on the skip list.
-    assert forkid.extract_forkid(no_forkid_nodes[0].enr) is None
+    with pytest.raises(ENRMissingForkID):
+        assert forkid.extract_forkid(no_forkid_nodes[0].enr)
```
carver: Looks like the comment is out of date now? Also, no point in running the assert now, I suppose.

gsalgado: The comment is about the `should_skip_fn` call below, so it still applies. The assert is just to show that the candidate has no fork-id

carver: Since this line raises an exception on a passing test, the `assert` should never evaluate, right?

I think it would be a confusing failure if `extract_forkid` stops raising the exception, and the `assert` fails. The logging would warn about the assert instead of a the exception not being raised.

gsalgado: Oops, indeed, I should've removed the assert from that line. I'll sneak that change in another branch

Issue #1586, opened 2020-03-01
Trinity crashes during ENR lookup


carver:
* Trinity version: 08c6784

### What is wrong?

```
Traceback (most recent call last):
  File "/usr/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/home/jcarver/code/trinity/trinity/extensibility/trio.py", line 56, in run_process
    trio.run(cls._do_run, boot_info)
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/trio/_core/_run.py", line 1804, in run
    raise runner.main_task_outcome.error
  File "/home/jcarver/code/trinity/trinity/extensibility/trio.py", line 70, in _do_run
    nursery.cancel_scope.cancel()
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/trio/_core/_run.py", line 730, in __aexit__
    raise combined_error_from_nursery
  File "/home/jcarver/code/trinity/trinity/components/builtin/peer_discovery/component.py", line 101, in do_run
    await async_service.run_trio_service(discovery_service)
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/async_service/trio.py", line 156, in run_service
    await manager.run()
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/async_service/trio.py", line 208, in run
    for _, exc_value, exc_tb in self._errors
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/async_service/base.py", line 301, in _run_and_manage_task
    await task.run()
  File "/home/jcarver/code/trinity/venv/lib/python3.6/site-packages/async_service/trio.py", line 76, in run
    await self._async_fn(*self._async_fn_args)
  File "/home/jcarver/code/trinity/p2p/discovery.py", line 706, in receive
    await handler(node, payload, message_hash)
  File "/home/jcarver/code/trinity/p2p/discovery.py", line 860, in recv_enr_response
    existing_enr = self.node_db.get_enr(enr.node_id)
  File "/home/jcarver/code/trinity/p2p/discv5/enr_db.py", line 58, in get_enr
    return rlp.decode(self.db[self._get_enr_key(node_id)], sedes=ENR)
[...]

PR #1588, opened 2020-03-02
discv4: Replace ENR in our DB when we detect an address change

gsalgado:
If a node sends us a message from an address different than what is
in their last ENR, we replace that with a stub ENR using the new
address.

Closes: #1586

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -88,7 +88,8 @@ from p2p.exceptions import (
     CouldNotRetrieveENR,
     NoEligibleNodes,
 )
-from p2p.kademlia import Address, Node, RoutingTable, check_relayed_addr, sort_by_distance
+from p2p.kademlia import (
+    Address, Node, RoutingTable, check_relayed_addr, create_stub_enr, sort_by_distance)
 from p2p import trio_utils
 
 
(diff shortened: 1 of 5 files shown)


carver:
> I'm not sure if @carver is on the way to EthCC or not. I'll leave it up to you to decide if you want to wait for his review or merge already.

Nope, I'm not going this year. I'll take a look today.

tests-trio/p2p-trio/test_discovery.py
```diff
@@ -458,6 +459,43 @@ def should_skip(skip_list, candidate):
     assert bob_enr == await bob.get_local_enr()
 
 
+@pytest.mark.trio
+async def test_lookup_and_maybe_update_enr():
```
carver: Looks like this could be split into three isolated tests.

p2p/discovery.py
```diff
@@ -854,24 +868,28 @@ def _is_msg_expired(self, rlp_expiration: bytes) -> bool:
...
-                "have (%d). Ignoring it",
-                encode_hex(enr.node_id),
-                enr.sequence_number,
-                existing_enr.sequence_number,
-            )
-            return
+        try:
+            existing_enr = self.node_db.get_enr(enr.node_id)
+        except KeyError:
+            self.logger.warning("No existing ENR for %s, this shouldn't happen", node)
```
carver: Hm, can we supply any more info here? Would be nice to have something that will help us when we get the inevitable bug report "Thing that shouldn't happen, happened". :)

tests-trio/p2p-trio/test_discovery.py
gsalgado: Indeed!

p2p/discovery.py
gsalgado: We could include the exception info in the log, but that wouldn't help at all as this is the handler for `ENR_RESPONSE` msgs, so the call stack will be `receive() -> recv_enr_response()` -- the buggy code that corrupted the DB would have executed long before that in another task

PR #1589, opened 2020-03-02
discv4: Don't let some task crashes propagate to run_task()

gsalgado:
Some discv4 coroutines executed with run_task() could leak exceptions
causing the whole component to stop, even though a failure in any of
them do not affect the component's operation. These coroutines are now
wrapped and any exceptions they leak are logged as errors.

This was motivated by the following crash, which I saw today and didn't
give me any clues about what was the invalid address nor how it ended up
in our DB

```
  File "trinity/components/builtin/peer_discovery/component.py", line 124, in <module>
    trio.run(main)
  File "/home/salgado/virtualenvs/trinity-trio/lib/python3.7/site-packages/trio/_core/_run.py", line 1804, in run
    raise runner.main_task_outcome.error
  File "trinity/components/builtin/peer_discovery/component.py", line 120, in main
    await run_standalone_eth1_component(PeerDiscoveryComponent)
  File "/home/salgado/src/trinity/trinity/extensibility/component.py", line 185, in run_standalone_eth1_component
    await run_standalone_component(component, endpoint_type, APP_IDENTIFIER_ETH1, (Eth1AppConfig,))
  File "/home/salgado/src/trinity/trinity/extensibility/component.py", line 171, in run_standalone_component
    await component.do_run(boot_info, endpoint)
  File "trinity/components/builtin/peer_discovery/component.py", line 101, in do_run
    await async_service.run_trio_service(discovery_service)
  File "/home/salgado/src/async-service/async_service/trio.py", line 156, in run_service
    await manager.run()
  File "/home/salgado/src/async-service/async_service/trio.py", line 208, in run
    for _, exc_value, exc_tb in self._errors
  File "/home/salgado/src/async-service/async_service/base.py", line 299, in _run_and_manage_task
    await task.run()
  File "/home/salgado/src/async-service/async_service/trio.py", line 76, in run
    await self._async_fn(*self._async_fn_args)
  File "/home/salgado/virtualenvs/trinity-trio/lib/python3.7/site-packages/trio/_socket.py", line 753, in sendto
    args[-1] = await self._resolve_remote_address(args[-1])
  File "/home/salgado/virtualenvs/trinity-trio/lib/python3.7/site-packages/trio/_socket.py", line 564, in _resolve_remote_address
    return await self._resolve_address(address, 0)
  File "/home/salgado/virtualenvs/trinity-trio/lib/python3.7/site-packages/trio/_socket.py", line 539, in _resolve_address
    host, port, self._sock.family, self.type, self._sock.proto, flags
  File "/home/salgado/virtualenvs/trinity-trio/lib/python3.7/site-packages/trio/_socket.py", line 159, in getaddrinfo
    host, port, family, type, proto, flags | _NUMERIC_ONLY
  File "/usr/lib/python3.7/socket.py", line 748, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -9] Address family for hostname not supported
```

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -658,10 +658,22 @@ class DiscoveryService(Service):
             return
         await self.lookup_random()
 
+    async def _sendto(self, msg: bytes, ip: str, port: int) -> None:
+        try:
+            await self.socket.sendto(msg, (ip, port))
+        except OSError:
+            self.logger.exception("Unexpected error sending msg to %s", (ip, port))
+
     def send(self, node: NodeAPI, msg_type: DiscoveryCommand, payload: Sequence[Any]) -> bytes:
+        """
+        Pack the given payload using the given msg type and fire a background task to try and
+        send it over our socket.
+
+        If we get an OSError from our socket when attempting to send it, that will be logged
+        and the message will be lost.
+        """
         message = _pack_v4(msg_type.id, payload, self.privkey)
-        self.manager.run_task(
-            self.socket.sendto, message, (node.address.ip, node.address.udp_port))
+        self.manager.run_task(self._sendto, message, node.address.ip, node.address.udp_port)
         return message
 
     async def consume_datagram(self) -> None:
--- a/scripts/discovery.py
+++ b/scripts/discovery.py
@@ -12,6 +12,7 @@ from p2p.events import PeerCandidatesRequest
 from p2p.constants import DISCOVERY_EVENTBUS_ENDPOINT
 from p2p.discv5.typing import NodeID
 
+from trinity.exceptions import ENRMissingForkID
 from trinity.network_configurations import PRECONFIGURED_NETWORKS
 from trinity.protocol.eth.forkid import (
     extract_fork_blocks,
@@ -62,10 +63,12 @@ async def main() -> None:
             with trio.fail_after(1):
                 response = await client.request(PeerCandidatesRequest(MAX_PEERS, should_skip))
             candidates = response.candidates
-            missing_forkid = [
-                candidate.id for candidate in candidates
-                if extract_forkid(candidate.enr) is None
-            ]
+            missing_forkid = []
+            for candidate in candidates:
+                try:
+                    extract_forkid(candidate.enr)
+                except ENRMissingForkID:
+                    missing_forkid.append(candidate.id)
             logger.info(
                 "Got %d connection candidates, %d of those with a matching ForkID",
                 len(candidates),


p2p/discovery.py
```diff
@@ -658,10 +669,15 @@ def _get_max_neighbours_per_packet(cls) -> int:
             return
         await self.lookup_random()
 
+    async def _sendto(self, msg: bytes, ip: str, port: int) -> None:
+        try:
+            await self.socket.sendto(msg, (ip, port))
+        except Exception:
```
cburgdorf: I'm not sure if catching `Exception` is the right thing to do here. It feels like this could unintentionally make it harder for us to notice certain errors. Maybe instead we should just handle the `socket.gaierror` that is mentioned in the stack trace but leave other errors crash the component (and consequently crash Trinity). That way, if there's an error that we need to handle we will get to know about it sooner.

I will leave that one to be reviewed by @pipermerriam or @carver

gsalgado: AFAICT, If the component crashes it won't crash trinity. And in this case it's the DiscoveryService so trinity will continue to run but most likely fail to connect to any peers. The point about the blanket except is that these tasks don't really affect the operation of the DiscoveryService, so having them crash the whole component and forcing users to restart (and maybe even patch) trinity is not nice. To give you another example, #1586 is another crash that would've been captured here instead of crashing the whole component. I'd argue that an ERROR in the logs **should** draw our (devs) attention and be a good indication of a bug, in which case this change wouldn't be hiding bugs, but given the whole situation with the failures when shutting down, this may not be the case.

carver: This basic approach of wrapping `socket.sendto` before calling `run_task()` makes sense to me. It will also help us by making the stack trace clearer. I also don't think we have a good enough reason to catch all `Exception`s here. Looks to me like catching `OSError` should be sufficient: https://docs.python.org/3/library/socket.html#socket.error

The internal method could be called something like `_best_effort_send()` to try to indicate to the caller when a typical socket failure happens, there won't be anything raised. But things like `MemoryError` being caught and silenced here would be odd.

> AFAICT, If the component crashes it won't crash trinity. And in this case it's the DiscoveryService so trinity will continue to run but most likely fail to connect to any peers.

I think the goal is to eventually have a single (critical) component crash cause all of trinity to crash, so we don't get these weird partially-working situations.

> To give you another example, #1586 is another crash that would've been captured here instead of crashing the whole component. I'd argue that an ERROR in the logs should draw our (devs) attention and be a good indication of a bug, in which case this change wouldn't be hiding bugs, but given the whole situation with the failures when shutting down, this may not be the case.

Yeah, I seem to be the one noticing a lot of these because I'm running a full node in the background most of the time, and it's annoying when it crashes. :) But I probably *wouldn't* notice a random error that prints out once every few hours. So I still think it's better to explicitly catch exceptions like `KeyError` in #1586, but let it crash for unexpected exceptions. We just need to keep getting better at predicting and catching the expected exceptions.

gsalgado:
Fair enough, let's go with just catching OSErrors from socket.sendto(), then

cburgdorf:
I'm merging this to get it into the release (I think the fundamental point has been addressed). In case anyone is wondering [CI is green](https://circleci.com/gh/ethereum/trinity/196104?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), it's just that the session is still open at the time of merging.

PR #1591, opened 2020-03-04
discv4: Use the same routing table implementation as discv5

gsalgado:
Closes: #1451


p2p/discv5/routing_table.py
```diff
@@ -255,10 +263,23 @@ def get_least_recently_updated_log_distance(self) -> int:
...
-    def iter_nodes_around(self, reference_node_id: NodeID) -> Generator[NodeID, None, None]:
+    def iter_nodes_around(self, reference_node_id: NodeID) -> Iterator[NodeID]:
         """Iterate over all nodes in the routing table ordered by distance to a given reference."""
         all_node_ids = itertools.chain(*self.buckets)
         distance_to_reference = functools.partial(compute_distance, reference_node_id)
         sorted_node_ids = sorted(all_node_ids, key=distance_to_reference)
         for node_id in sorted_node_ids:
             yield node_id
+
+    def iter_all_random(self) -> Iterator[NodeID]:
```
gsalgado: So, the reason why I iterate over both buckets and replacement caches here is described in https://github.com/ethereum/trinity/issues/1451#issuecomment-592322858, but basically it is to ensure we can provide enough compatible candidates to DiscoveryService

cburgdorf:
@gsalgado I might not get to look at this before tomorrow but based on the discussion in #1451 maybe @jannikluhn wants to review this. If not, I'll take a look tomorrow.

gsalgado:
@cburgdorf Yeah, I wanted to hear @jannikluhn 's opinion, but this is not changing discv5 in any way, and I haven't managed to get his attention, so yeah, it'd be nice if you could review it when you have some time

p2p/discovery.py
```diff
@@ -886,8 +912,20 @@ def _is_msg_expired(self, rlp_expiration: bytes) -> bool:
                 "Ignoring find_node request from node (%s) we haven't bonded with", node)
             return
         target_id = NodeID(keccak(target))
-        found = self.routing.neighbours(target_id)
-        self.send_neighbours_v4(node, found)
+        self.send_neighbours_v4(node, self.get_neighbours(target_id))
+
+    @to_tuple
+    def get_neighbours(self, target_id: NodeID) -> Iterator[NodeAPI]:
```
jannikluhn: Not important, but it seems random that this function returns `constants.KADEMLIA_BUCKET_SiZE` nodes. Maybe take a parameter `max_nodes`.

p2p/discv5/routing_table.py
jannikluhn: :+1: sounds reasonable

tests-trio/p2p-trio/test_discovery.py
```diff
@@ -50,10 +50,14 @@
...
-# Force our tests to fail quickly if they accidentally make network requests.
+# Force our tests to fail quickly if they accidentally get stuck waiting for a response that will
+# never come. Notice that we can't use a too low value here because this constant is also used
+# when waiting for things like waiting for a msg handler coroutine to forward the received msg
+# over the appropriate channel (e.g. when request_enr() waits for the ENR to be sent over the
+# enr_response_channels).
 @pytest.fixture(autouse=True)
 def short_timeout(monkeypatch):
-    monkeypatch.setattr(constants, 'KADEMLIA_REQUEST_TIMEOUT', 0.05)
+    monkeypatch.setattr(constants, 'KADEMLIA_REQUEST_TIMEOUT', 0.2)
```
jannikluhn: this seems quite fragile. I'm not sure if it's applicable here, but trio testing provides a fixture `autojump_clock`. If you use that timeouts shouldn't matter at all anymore.

cburgdorf:
Thanks for jumping in @jannikluhn

p2p/discovery.py
gsalgado: Well, that's defined in the spec: the response to a FIND_NODE request should contain the 16 nodes closes to the target, and I suppose the fact that it also recommends 16 as the bucket size is not a coincidence. Also, since it is only used here, I'd prefer to just document that in a docstring instead of adding an extra parameter.

gsalgado: Although it does feel weird to use the bucket size constant here, so I created a new one

tests-trio/p2p-trio/test_discovery.py
gsalgado: Hmm, I didn't know about that. I'll investigate if we can benefit from that, but I imagine it might require changes to quite a few tests, so I'll do that in a separate PR

PR #1593, opened 2020-03-09
Skip DAO fork block check on eth/64

gsalgado:
As of eth/64, the handshake also ensures the remote's ForkID
is compatible with the local one, so there's no need to perform
the DAO fork block check.

I did consider a more generic API on `Protocol` that would allow us to specify multiple types of checks/validations performed during handshake, but ultimately decided to go with a simpler alternative until we have an actual use case for different types of checks.

### DIFF
--- a/p2p/abc.py
+++ b/p2p/abc.py
@@ -477,8 +477,20 @@ class AsyncioServiceAPI(ABC):
         ...
 
 
+class HandshakeCheckAPI(ABC):
+    """
+    A certain type of check performed during a handshake with another peer.
+
+    e.g. Checking that we're on the same network, ForkID validation, etc.
+    """
+
+
 class HandshakeReceiptAPI(ABC):
     protocol: ProtocolAPI
+
+    @abstractmethod
+    def was_check_performed(self, check_type: Type[HandshakeCheckAPI]) -> bool:
+        ...
 
 
 THandshakeReceipt = TypeVar('THandshakeReceipt', bound=HandshakeReceiptAPI)
--- a/p2p/receipt.py
+++ b/p2p/receipt.py
@@ -1,4 +1,7 @@
+from typing import Type
+
 from p2p.abc import (
+    HandshakeCheckAPI,
     HandshakeReceiptAPI,
     ProtocolAPI,
 )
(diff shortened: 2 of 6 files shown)


p2p/abc.py
```diff
@@ -288,6 +288,7 @@ def send(self, message: MessageAPI) -> None:
 class ProtocolAPI(ABC):
     name: ClassVar[str]
     version: ClassVar[int]
+    validates_forkid_on_handshake: ClassVar[bool]
```
cburgdorf: I'm a little bit worried about this making the whole `ProtocolAPI` slightly less generic. I think you mention this yourself in the description of the PR.

>I did consider a more generic API on Protocol that would allow us to specify multiple types of checks/validations performed during handshake, but ultimately decided to go with a simpler alternative until we have an actual use case for different types of checks.

I'm wondering if the situation would be slightly better if in `boot.py` we would not check for `receipt.protocol.validates_forkid_on_handshake` but `receipt.validated_forkid_on_handshake`  instead? So, making it a property of the receipt rather than the protocol`. In fact, the `fork_id` is already a protocol of the `ETHHandshakeReceipt` anyway.

gsalgado: Yeah, it might make more sense to have that attribute on the receipt classes, but I also just realized that it should not actually be in either of the generic `ProtocolAPI` / `HandshakeReceiptAPI` classes as those are meant to have no awareness of any bits specific to the Ethereum protocols. But that is also a problem because `Peer.connection.handshake_receipts` is declared as a `Tuple[HandshakeReceiptAPI, ...]` so unless we have the new attribute there we'd lose static type checking and have to silence mypy everywhere we access it. I know @pipermerriam feels pretty strong about keeping everything under `p2p` ETH-agnostic, so I think it's better to hold back on this and wait to hear what he'd prefer here

gsalgado: One way around this would be to have a method on `HandshakeReceiptAPI` that takes as argument an identifier for a type of check (e.g. `ForkIDValidation`, `GenesisHashValidation etc) and returns true if that check is performed during handshake. For now, as this will only be used for the fork-id validation, we could have a default implementation in `p2p.HandshakeReceipt` that always returns False and in `ETHHandshakeReceipt` we'd overwrite that and return True for `ForkIDValidation`.

gsalgado: I went with the solution above as Piper said he was good with anything that didn't introduce eth knowledge to the p2p package. Can you have another look when you have a chance, @cburgdorf

cburgdorf: Yep, I think that's a good solution :+1:

PR #1595, opened 2020-03-09
Monitor total number of incoming and outgoing packets

cburgdorf:
### What was wrong?

This is building up on #1569. Nothing was wrong but we want to track more metrics. 

### How was it fixed?

This adds metrics tracking for total number of incoming and outgoing packets.

image

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1569.feature.rst
+++ b/newsfragments/1569.feature.rst
@@ -1,0 +1,9 @@
+Implement basic metrics tracking. With this change Trinity supports collecting
+and writing metrics to InfluxDB to be further processed and visualized
+by Grafana.
+
+Run Trinity with ``--enable-metrics --metrics-host "mymachine"`` and provide
+the Influx server and password via the following two environment variables:
+
+- ``TRINITY_METRICS_INFLUX_DB_SERVER``
+- ``TRINITY_METRICS_INFLUX_DB_PW``
--- a/trinity/components/builtin/metrics/metrics_service.py
+++ b/trinity/components/builtin/metrics/metrics_service.py
@@ -1,0 +1,54 @@
+from async_service import Service
+from eth_utils import get_extended_debug_logger
+from pyformance.reporters import InfluxReporter
+
+from p2p import trio_utils
+
+from trinity.components.builtin.metrics.registry import HostMetricsRegistry
+
+
+class MetricsService(Service):
+    """
+    A service to provide a registry where metrics instruments can be registered and retrieved from.
+    It continuously reports metrics to the specified InfluxDB instance.
+    """
+
+    def __init__(self,
+                 influx_server: str,
+                 influx_user: str,
+                 influx_password: str,
+                 influx_database: str,
+                 host: str,
+                 reporting_frequency: int = 10):
+
+        self._influx_server = influx_server
+        self._reporting_frequency = reporting_frequency
+        self._registry = HostMetricsRegistry(host)
+        self._reporter = InfluxReporter(
+            registry=self._registry,
+            protocol='https',
+            port=443,
+            database=influx_database,
+            username=influx_user,
+            password=influx_password,
+            server=influx_server
+        )
+
+    logger = get_extended_debug_logger('trinity.components.builtin.metrics.MetricsService')
+
+    @property
+    def registry(self) -> HostMetricsRegistry:
+        """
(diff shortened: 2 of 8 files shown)


cburgdorf:
@gsalgado I pinged you for a review. If you want access to the grafana backend I can ask Martin or Peter from the Geth team to create a login for you.

newsfragments/1569.feature.rst
```diff
@@ -0,0 +1,9 @@
+Implement basic metrics tracking. With this change Trinity supports collecting
+and writing metrics to InfluxDB to be further processed and visualized
+by Grafana.
+
+Run Trinity with ``--metrics --metrics-host "mymachine"`` and provide the
```
gsalgado: At first I thought `--metrics-host` was for the remote host where the influx DB is running. Maybe `--metrics-node-id` as there may me multiple instances running on the same host?

Also, maybe it should be `--enable-metrics` instead of just `--metrics`, to be consistent with other args?

trinity/components/builtin/metrics/metrics_service.py
```diff
@@ -0,0 +1,51 @@
...
+
+        self._influx_server = influx_server
+        self._reporting_frequency = reporting_frequency
+        self._registry = HostMetricsRegistry(host)
+        self._reporter = InfluxReporter(
+            registry=self._registry,
+            protocol='https',
+            port=443,
+            database='trinity',
+            username='trinity',
```
gsalgado: So the InfluxDB has to be pre-configured with a trinity DB and user? We need to document that somehow... If we have cli args for those as well (with `trinity` as the default value), I think that'd be enough

trinity/components/builtin/metrics/metrics_service.py
```diff
@@ -0,0 +1,51 @@
+from async_service import Service
+from eth_utils import get_extended_debug_logger
+from p2p import trio_utils
+from pyformance.reporters import InfluxReporter
```
gsalgado: Hmm, the ordering here is not right

trinity/components/builtin/metrics/system_metrics_collector.py
```diff
@@ -0,0 +1,116 @@
+from typing import (
+    NamedTuple,
+)
+
+from async_service import (
+    as_service,
+    ManagerAPI,
+)
+
+import psutil
```
gsalgado: This should be at the top

newsfragments/1569.feature.rst
cburgdorf: Yeah, `--metrics-host` is a little bit ambiguous. The geth team introduced that name and I'd like to keep it to keep the grafana charts as compatible with theirs as possible. But renaming the CLI parameter to `--metrics-node-id` while at the same time sticking to `host` in grafana might not improve the situation. I tried to reduce the confusion by naming the actual influx host `--metrics-influx-server` so that at least we do not have multiple *hosty* things here :sweat_smile: 

>Also, maybe it should be `--enable-metrics` instead of just `--metrics`, to be consistent with other args?

That sounds like a good idea. I'll also send a follow up PR to remane `--ethstats` to `--enable-ethstats`

trinity/components/builtin/metrics/metrics_service.py
cburgdorf: >if we have cli args for those as well (with trinity as the default value), I think that'd be enough

Good call, will add those.

trinity/components/builtin/metrics/metrics_service.py
cburgdorf: I guess I'm treating `p2p` like an external package here and then numbers go before letters thus `p2` is higher ranked than `py` but I'm fine treating it like an internal package for the time being.

trinity/components/builtin/metrics/system_metrics_collector.py
cburgdorf: `psutil` is a 3rd party package wheras `typing` isn't. But I guess I should bring an empty line between it and `p2p` to treat `p2p` as internal (same issue as above).

So then we have:

```python
from typing import (
    NamedTuple,
)

from async_service import (
    as_service,
    ManagerAPI,
)
import psutil

from p2p import trio_utils

from trinity.components.builtin.metrics.registry import HostMetricsRegistry
```

PR #1598, opened 2020-03-10
Update quickstart.rst

Lilith1410:
### What was wrong?
There was no warning that starting trinity will lead to syncing (==downloading) the whole mainnet chain. 

### How was it fixed?
Added a warning in the docs that this will take time, a good internet connection + enough disk space to store the chain, as well as flags + reference to the guide on how to run a light node.

### DIFF
--- a/docs/guides/quickstart.rst
+++ b/docs/guides/quickstart.rst
@@ -126,10 +126,7 @@ Running Trinity
 
 .. note::
 
-    Execution of the following comand will result in the downloading of a considerable amount of data. We need to make sure to have enough disk space available as well as a data plan that allows such traffic. 
-    
-    If you want to run Trinity as a light node, check out :ref:`Run Trinity as a light client<cookbook_recipe_running_as_a_light_client>`
-    or use the ``--sync-mode=light`` flag. 
+    Execution of the following comand will cause downloading of a considerable amount of data. We need to make sure to have enough disk space available as well as a data plan that allows such traffic.
 
 After Trinity is installed we should have the ``trinity`` command available to start it.
 
--- a/docs/guides/quickstart.rst
+++ b/docs/guides/quickstart.rst
@@ -124,6 +124,15 @@ the package and it will immediately start syncing. Head over to the
 Running Trinity
 ~~~~~~~~~~~~~~~
 
+.. warning::
+
+    Executing the following command will download the entire ethereum mainnet chain. Please be sure that
+    this is what you want to do. It will probably take a long time, as well as a good internet connection and 
+    enough disk space on your machine to store the whole blockchain. 
+    
+    If you want to run Trinity as a light node, check out :ref:`Run Trinity as a light client<cookbook_recipe_running_as_a_light_client>`
+    or use the ``--sync-mode=light`` flag. 
+
 After Trinity is installed we should have the ``trinity`` command available to start it.
 
 .. code:: sh
--- a/docs/guides/quickstart.rst
+++ b/docs/guides/quickstart.rst
@@ -124,11 +124,9 @@ the package and it will immediately start syncing. Head over to the
 Running Trinity
 ~~~~~~~~~~~~~~~
 
-.. warning::
+.. note::
 
-    Executing the following command will download the entire ethereum mainnet chain. Please be sure that
-    this is what you want to do. It will probably take a long time, as well as a good internet connection and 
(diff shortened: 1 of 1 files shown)


docs/guides/quickstart.rst
```diff
@@ -124,6 +124,15 @@ the package and it will immediately start syncing. Head over to the
 Running Trinity
 ~~~~~~~~~~~~~~~
 
+.. warning::
```
cburgdorf: I think this could be `note` instead of `warning` to be a little less alarming.

docs/guides/quickstart.rst
```diff
@@ -124,6 +124,15 @@ the package and it will immediately start syncing. Head over to the
...
+
+    Executing the following command will download the entire ethereum mainnet chain. Please be sure that
+    this is what you want to do. It will probably take a long time, as well as a good internet connection and
```
cburgdorf: The narrative perspective that we use in our docs is written in the *first-person "we" form*. [We have some guidelines on how to write these docs](https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/documentation.md#narrative-perspective).


https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/documentation.md#narrative-perspective

docs/guides/quickstart.rst
```diff
@@ -124,6 +124,15 @@ the package and it will immediately start syncing. Head over to the
 Running Trinity
 ~~~~~~~~~~~~~~~
 
+.. warning::
+
+    Executing the following command will download the entire ethereum mainnet chain. Please be sure that
```
cburgdorf: >will download the entire ethereum mainnet chain

Trinity uses `beam` sync by default and at the time of writing we do not actually perform a full backfill of all the historical blockchain data. So the `will download the entire ethereum mainnet chain` part isn't entirely correct here. In contrast, it would be correct for `--sync-mode full` but that isn't the default sync mode we use.

So, I wonder if we can just rephrase that to be a little bit more generic. Something like:

*"Execution of the following command will result in the downloading of a considerable amount of data. We need to make sure to have enough disk space available as well as a data plan that allows such traffic."* (notice the change of narrative perspective)

docs/guides/quickstart.rst
```diff
@@ -124,6 +124,15 @@ the package and it will immediately start syncing. Head over to the
...
+    this is what you want to do. It will probably take a long time, as well as a good internet connection and 
+    enough disk space on your machine to store the whole blockchain. 
+    
+    If you want to run Trinity as a light node, check out :ref:`Run Trinity as a light client<cookbook_recipe_running_as_a_light_client>`
```
cburgdorf: We will likely remove `light` mode soon. A more future proof thing to point the reader to would be to use the `--beam-from-checkpoint <checkpoint-uri>` which avoids pre-downloading the header chain entirely. We do not have a cookbook receipt for that yet but it would be super cool if you could add one that covers "Running beam sync from a trusted checkpoint" and then link to that.

docs/guides/quickstart.rst
Lilith1410: Changed `` warning `` to `` note``

docs/guides/quickstart.rst
Lilith1410: Ah, yes, it makes sense to keep it consistent. Sorry, will definitely follow the guidelines in the future!

docs/guides/quickstart.rst
Lilith1410: I changed the text to your suggestion.

docs/guides/quickstart.rst
Lilith1410: Maybe we can merge this pull request with the `` light`` mode for now and make `` add cookbook recipe for beam sync from trusted checkpoint`` into its own issue? I have no idea how beam works right now so writing a recipe for that feels like out of scope for this pull request :)

cburgdorf: The problem is that even with `light` mode we will first download the entire chain of headers. In contrast, starting beam sync from a trusted checkpoint will just start right from that header.

But I agree that we don't need to put that beam sync node in there just now if you don't feel comfortable writing it. I suggest that we just remove the reference to `light` sync then and just keep the pure *space/traffic* warning.

I'm making this tiny adjustment right before merging.

cburgdorf:
Landed! Thank you for your contribution @Lilith1410

PR #1600, opened 2020-03-12
Make it possible to connect standalone components to one another

gsalgado:
When running standalone components one can now specify a list of IPC
files the component should connect to.

Also add the boilerplate code needed to run the chain syncer standalone.

### DIFF
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -184,7 +184,9 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
                 self.connection_tracker.get_blacklisted(),
                 timeout=1)
         except asyncio.TimeoutError:
-            self.logger.warning("ConnectionTracker.get_blacklisted() request timed out.")
+            self.logger.warning(
+                "Timed out getting blacklisted peers from connection tracker, pausing peer "
+                "addition until we can get that.")
             return 0
 
         skip_list = connected_node_ids.union(blacklisted_node_ids)
--- a/scripts/discovery.py
+++ b/scripts/discovery.py
@@ -1,6 +1,7 @@
 import argparse
 import functools
 import logging
+import pathlib
 from typing import Set
 import uuid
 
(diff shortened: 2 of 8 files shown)


p2p/peer_pool.py
```diff
@@ -185,7 +185,7 @@ def available_slots(self) -> int:
                 timeout=1)
         except asyncio.TimeoutError:
             self.logger.warning("ConnectionTracker.get_blacklisted() request timed out.")
-            return 0
+            blacklisted_node_ids = tuple()
```
gsalgado: This is inside `_add_peers_from_backend()`, so this would effectively ignore all peer candidates if the conn tracker times out, which doesn't seem like the right thing to do to me.

gsalgado: @carver ^

carver: Ah, is the idea here that you don't start the blacklist component, so it will always time out?

I think some explicit way to discover that there is no blacklist would be better. Maybe something in here could detect it?
https://github.com/ethereum/trinity/blob/df92df488c6ade7a168483075c1d4f76843d3090/trinity/protocol/common/peer.py#L119-L127

gsalgado: > My thought process: It depends on how important the blacklist is to the user.

I was thinking of the blacklist exclusively in terms of peers we want to temporarily avoid connecting to (e.g. different networks or handshake failures because they are full), but if a node operator manually adds a permanent blacklist, then it's a different story. I'll revert that change for now

> Ah, is the idea here that you don't start the blacklist component, so it will always time out?

Yeah, it'd be better if we could run the syncer without having to run the network-db as well, but it's trivial to run it anyway.

> I think some explicit way to discover that there is no blacklist would be better. Maybe something in here could detect it?

AFAIK, that is just so that we can run Peer services without an event bus in tests. When running the real component I don't see a way other than catching Timeouts from the event bus, but that doesn't allow us to tell for sure that it's not running

gsalgado:
@carver I've reverted the blacklist change but changed the warning to make it clear we pause adding new peers. Do I have your :+1: ?

PR #1602, opened 2020-03-12
Clean up some temporary components

ralexstokes:
### What was wrong?

These bits were useful for earlier interop experiments but are no longer.

### How was it fixed?
Removing stuff we don't need.

#### Cute Animal Picture


ralexstokes:
closing this in lieu of a fresh start :)

Issue #1605, opened 2020-03-13
Flaky test: py36-p2p-trio test_wait_neighbors


cburgdorf:
### What is wrong?

Test sometimes fails:

```
@pytest.mark.trio
    async def test_wait_neighbours(nursery):
        service = MockDiscoveryService([])
        node = NodeFactory()
    
        # Schedule a call to service.recv_neighbours_v4() simulating a neighbours response from the
        # node we expect.
        neighbours = tuple(NodeFactory.create_batch(3))
        expiration = _get_msg_expiration()
        neighbours_msg_payload = [
            [n.address.to_endpoint() + [n.pubkey.to_bytes()] for n in neighbours],
            expiration]
        nursery.start_soon(service.recv_neighbours_v4, node, neighbours_msg_payload, b'')
    
        received_neighbours = await service.wait_neighbours(node)
    
>       assert neighbours == received_neighbours
E       assert (<Node(0xb575...7.177:42975)>) == (<Node(0xb575...168.6:38939)>)
E         Left contains one more item: <Node(0xd27861f1651dd644af1e8dbc6beef38af74a753b8046474f119dda0c79c50cb355a25a003162f2c109700f82aaafd797126afef1b6bb912e099a2bc4cd1d71c0@10.209.67.177:42975)>
E         Use -v to get the full diff
```
https://circleci.com/gh/ethereum/trinity/198563

### How can it be fixed

:hocho:

gsalgado:
I couldn't reproduce it locally even when running hundreds of times in parallel, but the only possible cause I could think of is the fact that the extra node has an IP address is in a private network, and in that case `wait_neighbours()` will ignore it unless the sender is on the same private network

PR #1607, opened 2020-03-16
Fix test_discovery::test_wait_neighbours

gsalgado:
That test could fail when one of the randomly generated Nodes had
an IP address that was not reachable. Now it uses hard-coded
IP addresses.

Closes: #1605


cburgdorf:
Merged this as it will potentially make CI a little smoother so it's good to have it in as soon as possible.

PR #1608, opened 2020-03-16
RFC: Peer now sends a disconnect msg when terminating

gsalgado:
In order to do so it has to ensure the connection stays open until its
cleanup() method is called, and it does that by running the connection with
an unchained cancel token, and then manually cancels the connection once it
has used it to send the disconnect msg

Before this change, the connection would run with the PeerPool's cancel token (and
as a child service of the PeerPool), so almost always it'd be closed before the Peer
had a chance to run its _cleanup()

### DIFF
--- a/p2p/p2p_api.py
+++ b/p2p/p2p_api.py
@@ -32,8 +32,13 @@ class CancelOnDisconnect(CommandHandler[Disconnect]):
     disconnect_reason: DisconnectReason = None
 
     async def handle(self, connection: ConnectionAPI, cmd: Disconnect) -> None:
+        self.logger.warning("Received disconnect on %s", self.connection)
         self.disconnect_reason = cmd.payload
 
+        # XXX: This is probably necessary to trigger the peer's cancellation as well, although
+        # ideally we should cancel the peer
+        # Maybe the Peer should itself subscribe to Disconnect msgs so that it can cancel itself
+        # then. It doesn't seem right to have the P2PAPI cancel the connection...
         if connection.is_operational:
             connection.cancel_nowait()
 
@@ -72,7 +77,7 @@ class P2PAPI(Application):
         """
         return self._disconnect_handler.disconnect_reason
 
-    def _disconnect(self, reason: DisconnectReason) -> None:
+    def disconnect(self, reason: DisconnectReason) -> None:
         self.logger.debug(
             "Sending Disconnect to remote peer %s; reason: %s",
             self.connection,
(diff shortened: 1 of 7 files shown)


p2p/p2p_api.py
```diff
@@ -81,16 +86,6 @@ def _disconnect(self, reason: DisconnectReason) -> None:
         self.send_disconnect(reason)
         self.local_disconnect_reason = reason
 
-    async def disconnect(self, reason: DisconnectReason) -> None:
-        self._disconnect(reason)
-        if self.connection.is_operational:
-            await self.connection.cancel()
```
carver: There are other places that call this API that require the peer to be disconnected (and gone from the peer pool) at the end of an `await peer.disconnect()`. The first place that comes to mind is the checkpoint resolution stuff here:

https://github.com/ethereum/trinity/blob/54304e30e9b871055a178c3ce601a4715043e181/trinity/sync/common/strategies.py#L196

Can we keep that API guarantee?

gsalgado: Good catch, I'd missed that. We could have it wait for the connection's `cleaned_up` event, I guess? But I wonder if it should also trigger the peer cancellation?

gsalgado: How was this not caught by mypy, btw?

gsalgado: Oh, actually, scratch all of the above! This is `P2PAPI.disconnect()`. `Peer.disconnect()` is still async and waits until the peer is finished, which in any case will also mean the connection is closed

p2p/p2p_api.py
```diff
@@ -32,8 +32,13 @@ class CancelOnDisconnect(CommandHandler[Disconnect]):
     disconnect_reason: DisconnectReason = None
 
     async def handle(self, connection: ConnectionAPI, cmd: Disconnect) -> None:
+        self.logger.warning("Received disconnect on %s", self.connection)
         self.disconnect_reason = cmd.payload
 
+        # XXX: This is probably necessary to trigger the peer's cancellation as well, although
+        # ideally we should cancel the peer
+        # Maybe the Peer should itself subscribe to Disconnect msgs so that it can cancel itself
+        # then. It doesn't seem right to have the P2PAPI cancel the connection...
```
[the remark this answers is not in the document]
pipermerriam: I don't have any objections.  I agree that the `P2PAPI` cancelling the connection is a code smell.

p2p/peer_pool.py
```diff
@@ -263,7 +263,10 @@ def unsubscribe(self, subscriber: PeerSubscriber) -> None:
         self.run_child_service(peer.connection)
         await self.wait(peer.connection.events.started.wait(), timeout=1)
 
-        self.run_child_service(peer)
+        if not isinstance(peer.connection, BaseService):
+            raise ValueError("FIXME")
+
+        cast(BaseService, peer.connection).run_child_service(peer)
```
gsalgado: This is where I really wanted your feedback, @pipermerriam... Running the peer as a child service of the connection doesn't feel right to me, but at the same time the alternative I could think of ([overwriting the connection's cancel token](#1609)) doesn't feel right either. Can you think of any reasons why we'd prefer one over the other? Or maybe a third option?

gsalgado:
Sorry, I'd messed things up and this was identical to #1609. Now this one reflects what is in the description. Do you have a preference for any of these approaches, @pipermerriam?

pipermerriam:
I think it makes the most sense to run the peer as a child of the connection. Overriding the cancel token seems worse.

gsalgado:
Closing as we decided to go with #1609

PR #1609, opened 2020-03-16
Peer now sends a disconnect msg when terminating

gsalgado:
In order to do so we need to ensure the connection stays open until the peer's cleanup() method is called, and we do that by running the peer as a child service of the connection.

Also moves the logic to cancel the peer/connection from P2PAPI to BasePeer

### DIFF
--- a/p2p/connection.py
+++ b/p2p/connection.py
@@ -9,6 +9,7 @@ from typing import (
     Set,
     Tuple,
     Type,
+    TYPE_CHECKING,
     Union,
 )
 
@@ -31,6 +32,7 @@ from p2p.abc import (
     TLogic,
     TProtocol,
 )
+from p2p.constants import PEER_READY_TIMEOUT
 from p2p.disconnect import DisconnectReason
 from p2p.exceptions import (
     DuplicateAPI,
@@ -45,6 +47,9 @@ from p2p.subscription import Subscription
 from p2p.service import BaseService
 from p2p.p2p_proto import BaseP2PProtocol, DevP2PReceipt, Disconnect
 from p2p.typing import Capabilities
+
+if TYPE_CHECKING:
+    from p2p.peer import BasePeer  # noqa: F401
 
 
 class Connection(ConnectionAPI, BaseService):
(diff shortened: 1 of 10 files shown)


p2p/peer_pool.py
```diff
@@ -263,7 +263,10 @@ def unsubscribe(self, subscriber: PeerSubscriber) -> None:
         self.run_child_service(peer.connection)
         await self.wait(peer.connection.events.started.wait(), timeout=1)
 
-        self.run_child_service(peer)
+        if not isinstance(peer.connection, BaseService):
+            raise ValueError("FIXME")
+
+        cast(BaseService, peer.connection).run_child_service(peer)
```
gsalgado: This is where I really wanted your feedback, @pipermerriam... Running the peer as a child service of the connection doesn't feel right to me, but at the same time the alternative I could think of (overwriting the connection's cancel token) doesn't feel right either. Can you think of any reasons why we'd prefer one over the other? Or maybe a third option?

pipermerriam: I think I prefer this approach to the one in #1608 which overrides the cancel token.  It seems much more straight forward.

tests/p2p/test_peer_pair_factory.py
```diff
@@ -26,3 +27,22 @@
...
+async def test_disconnect_on_cancellation(monkeypatch):
+    disconnects_received = []
+
+    orig_handle = p2p_api.CancelOnDisconnect.handle
+
+    async def handle(self, connection, cmd) -> None:
+        disconnects_received.append(connection)
+        await orig_handle(self, connection, cmd)
+
+    monkeypatch.setattr(p2p_api.CancelOnDisconnect, 'handle', handle)
```
pipermerriam: You shouldn't have to monkey patch this.  Instead you should use the [`ConnectionAPI.add_command_handler`](https://github.com/ethereum/trinity/blob/cd0f290276963cb07ee0771a7bd8bf8d2164f2cf/p2p/connection.py#L194) to just register a function to that will then be called if a disconnect is received.

tests/p2p/test_peer_pair_factory.py
```diff
@@ -26,3 +27,22 @@
...
+    async with ParagonPeerPairFactory() as (alice, bob):
+        await alice.cancel()
+        await alice.events.cleaned_up.wait()
+        await asyncio.sleep(0.1)
```
pipermerriam: Rather than using a mutable list and appending to it to detect that something happened, I recommend using an `asyncio.Event` and just setting it in the handler.  That allows this test to do an `asyncio.wait_for(event.wait(), timeout=1)` which will ensure that the test runs as fast as the event loop will allow on the success case, and hangs for 1 second on the failure case.

p2p/connection.py
```diff
@@ -89,6 +93,17 @@ def __repr__(self) -> str:
...
+    async def run_peer(self, peer: 'BasePeer') -> None:
+        """
+        Run the peer as a child service.
+
+        A peer must always run as a child of the connection so that it has an open connection
+        until it finishes its cleanup.
+        """
+        self.run_child_service(peer)
+        await self.wait(peer.events.started.wait(), timeout=1)
+        await self.wait(peer.ready.wait(), timeout=1)
```
cburgdorf: Wondering if these timeouts should use descriptive constants defined in `p2p.constants.py`

gsalgado: Sure, I went with `PEER_READY_TIMEOUT`

PR #1610, opened 2020-03-16
Do not send a double SIGTERM to components that timeout when terminating

gsalgado:
AsyncioIsolatedComponent would catch a CancelledError from its
sub-process, send a SIGTERM to it and then re-raise the exception,
causing asyncio-run-in-process to send yet another SIGTERM to the
process, causing asyncio to raise a InvalidStateError. Now AIC
does not re-raise the CancelledError after sending a SIGTERM

```
   ERROR  2020-03-16 15:24:38,928               asyncio  Task was destroyed but it is pending!
task: <Task pending coro=<AsyncioIsolatedComponent._do_run() running at /home/salgado/src/trinity/trinity/extensibility/asyncio.py:70> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f153ccb3410>()]>>
   ERROR  2020-03-16 15:24:41,181               asyncio  Exception in callback <built-in method set_result of _asyncio.Future object at 0x7f53f6e18290> handle: <Handle Future.set_result>
  File "uvloop/cbhandles.pyx", line 70, in uvloop.loop.Handle._run
asyncio.base_futures.InvalidStateError: invalid state
```

### DIFF
--- a/trinity/extensibility/asyncio.py
+++ b/trinity/extensibility/asyncio.py
@@ -1,6 +1,4 @@
 from abc import abstractmethod
-import asyncio
-import signal
 from typing import Optional
 
 from asyncio_run_in_process import open_in_process
@@ -32,23 +30,7 @@ class AsyncioIsolatedComponent(BaseIsolatedComponent):
             subprocess_kwargs=self.get_subprocess_kwargs(),
         )
         async with proc_ctx as proc:
-            try:
-                await proc.wait_result()
-            except asyncio.CancelledError as err:
-                logger.debug('Component %s exiting. Sending SIGINT to pid=%d', self, proc.pid)
-                proc.send_signal(signal.SIGINT)
-                try:
-                    await asyncio.wait_for(proc.wait(), timeout=2)
-                except asyncio.TimeoutError:
-                    logger.debug(
-                        'Component %s running in process pid=%d timed out '
-                        'during shutdown. Sending SIGTERM and exiting.',
-                        self,
-                        proc.pid,
-                    )
-                    proc.send_signal(signal.SIGTERM)
-                finally:
-                    raise err
+            await proc.wait_result_or_raise()
 
     @classmethod
     async def _do_run(cls, boot_info: BootInfo) -> None:


pipermerriam:
If I'm understanding correctly, the `asyncio.CancelledError` is coming from the other process and being propogated across the process boundary.  This doesn't seem like good behavior.

The `asyncio.CancelledError` is one of those exceptions that should only be silenced when you know where it is coming from like having previously called `task.cancel`.  Otherwise you typically always want it to bubble.  I however don't think this logic holds true for things that happen across process boundaries.  By allowing `asyncio.CancelledError` to bubble from the child process into the parent process it introduces ambiguity over what the proper way to handle a the error in this type of context.  Since you don't know if the error came from the child process or the current process, you can't know how to handle it...

I'm not sure the best way to address this right now but I wanted to write my thoughts down.

trinity/extensibility/asyncio.py
```diff
@@ -35,21 +35,19 @@ def get_subprocess_kwargs(self) -> Optional[SubprocessKwargs]:
...
-                    logger.debug(
+                    logger.info(
                         'Component %s running in process pid=%d timed out '
                         'during shutdown. Sending SIGTERM and exiting.',
                         self,
                         proc.pid,
                     )
                     proc.send_signal(signal.SIGTERM)
-                finally:
-                    raise err
```
pipermerriam: See my comment about how it may be incorrect to not re-raise this since we don't know if it came from the child process or whether it originated locally.  Maybe an intermediate solution would be to only re-raise if `proc.error` is an `asyncio.CancelledError`.

pipermerriam:
If you feel sure about silencing the `CancelledError` :+1:

trinity/extensibility/asyncio.py
```diff
@@ -34,21 +34,19 @@ def get_subprocess_kwargs(self) -> Optional[SubprocessKwargs]:
         async with proc_ctx as proc:
             try:
                 await proc.wait_result()
```
gsalgado, line 36: I wasn't sure but I just realized the CancelledError is actually from the component itself.  When the ComponentManager leaves the `run_component()` context, the component is cancelled, causing the CancelledError here, and we then send a sigint/sigterm combo to the sub process. Here's what I see on the logs: 

```
   INFO  2020-04-03 21:11:33,383  AsyncioIsolatedComponent  Component <Component[Upnp]> running in process pid=24423 timed out during shutdown. Sending SIGTERM and exiting: 
    INFO  2020-04-03 21:11:33,384  asyncio_run_in_process  Got CancelledError while running subprocess pid=24423.  Sending SIGTERM.

```

But now I'm not sure how this happens since `run_component()` swallows the CancelledError from here (`AIOIC.run()`).

gsalgado: Actually, since the call chain is: `run_component() -> open_in_process() -> AsyncioIsolatedComponent.run()`, if AIOC.run() re-raises the CancelledError, it will first be caught in `open_in_process()`, causing the double sigterm. So yeah, I'm now sure we don't want to re-raise it here. @pipermerriam

gsalgado: But now I'm thinking that the correct fix is to get rid of the whole `except asyncio.CancelledError:` from here, letting the exception propagate to `open_in_process()`, which handles it. The only problem is that `open_in_process()` currently sends just a SIGTERM when there's a CancelledError. Can you think of any reason why `open_in_process()` shouldn't use the same logic as here to handle a CancelledError?

pipermerriam:
If we approach this from the `asyncio-run-in-process` side then I think it needs to have the following properties.

- locally raised `CancelledError` should terminate the process and bubble.
- a `CancelledError` from within the process **should not** bubble.  The process should terminate with an error but we probably need to wrap it with a custom exception so that you can differentiate when handling errors.

gsalgado:
Ok, so I was hoping that we'd be able to consolidate the sending of signals to child processes in `asyncio_run_in_process` with https://github.com/ethereum/asyncio-run-in-process/pull/13 but it turns out one of our tests (https://github.com/ethereum/trinity/blob/00dc1a032851ef830ab04cedc4ce39070c83b18a/tests/core/components/test_isolated_component.py#L49) only passes because we have those two places sending signals to the child, so the child ends up receiving a SIGINT (from `AsyncIOIsolatedComponent`) followed immediately by a SIGTERM (from `asyncio_run_in_process`), as can be seen below:

```
   DEBUG  04-06 10:59:39            asyncio.py  Component <Component[component-test]> exiting. Sending SIGINT to pid=14938                                                                                         
   DEBUG  04-06 10:59:39               base.py  <Manager[ComponentManager] flags=SRCfe>: task _handle_shutdown_request[daemon=True] raised CancelledError.                                                         
   DEBUG  04-06 10:59:39  _component_isolation.py  Exiting `do_run`                                                                                                                                                
   DEBUG  04-06 10:59:39               base.py  <Manager[ComponentManager] flags=SRCfe>: task _track_and_propagate_available_endpoints[daemon=True] raised CancelledError.                                         
 WARNING  04-06 10:59:39            asyncio.py  <Manager[AsyncioComponentService] flags=SRcfe>: Waiting to stop                                                                                                    
   DEBUG  04-06 10:59:39     run_in_process.py  Got CancelledError while running subprocess pid=14938.  Sending SIGTERM.              
```

I doubt it is intentional that components only terminate when sent a SIGINT/SIGTERM combo, and it looks like there might be some race condition here because when actually running trinity we only seem to send the SIGTERM after the 2s timeout, and that coupled with the sequential termination of components (#1650) is why we take forever to shutdown. Sometimes it feels like I'll never get to the bottom of all this mess

gsalgado:
Closing this as what we need is https://github.com/ethereum/trinity/pull/1663

Issue #1507, opened 2020-01-23
Stopping trinity-beacon doesn't work


jannikluhn:
* Trinity version: master
* OS: Arch Linux
* Python version (output of `python --version`):

### What is wrong?

Attempting to stop `trinity-beacon` by hitting `CTRL-c` fails on my machine. Here's the log output:

```
    INFO  2020-01-23 17:54:57,778      ComponentManager  Stopping components
    INFO  2020-01-23 17:54:57,779  TrioIsolatedComponent  Sent SIGINT to process 20572, waiting 10 seconds for it to terminate
    INFO  2020-01-23 17:54:57,785           Eth1Monitor  Eth1 Monitor got new eth1 block: Eth1Block(block_hash=b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xb2\x81', number=45697, timestamp=396045), number of logs contained in the block: 1
Task was destroyed but it is pending!
task: <Task pending coro=<Event.wait() running at /home/jannik/.pyenv/versions/3.7.5/lib/python3.7/asyncio/locks.py:293> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f72c9bbb710>()]> cb=[_release_waiter(<Future pending>)() at /home/jannik/.pyenv/versions/3.7.5/lib/python3.7/asyncio/tasks.py:392]>
    INFO  2020-01-23 17:54:59,011  TrioIsolatedComponent  Process 20572 has already terminated
    INFO  2020-01-23 17:54:59,011  TrioIsolatedComponent  Process 20572 has already terminated
 WARNING  2020-01-23 17:55:03,017  asyncio_run_in_process  Child process pid=20570 failed to exit cleanly.  Sending SIGKILL
Traceback (most recent call last):
  File "/home/jannik/.pyenv/versions/3.7.5/envs/trinity/lib/python3.7/site-packages/asyncio_run_in_process/run_in_process.py", line 298, in _open_in_process
    raise err
  File "/home/jannik/.pyenv/versions/3.7.5/envs/trinity/lib/python3.7/site-packages/asyncio_run_in_process/run_in_process.py", line 263, in _open_in_process
    yield proc
  File "/home/jannik/Projects/Ethereum/trinity/trinity/extensibility/asyncio.py", line 52, in run
    raise err
[...]

PR #1612, opened 2020-03-17
Use threadsafe cancel to prevent hanging cleanup on ctrl-c

njgheorghita:
### What was wrong?
Issue: #1507  
Suggested solution: https://github.com/ethereum/asyncio-run-in-process/issues/11
> I don't think this issue addresses the solution, I was not able to reproduce the failing tests locally. 

Problem:
- `trinity-beacon` command hangs when terminated with `ctrl-c`. 
- After process is killed manually - launching `trinity-beacon` again results in an `Address already in use` error - caused by dangling ipc files in the xdg directory. These ipc files need to be deleted manually.

### How was it fixed?
I'm new to this library - so this might not be the best way to solve this problem. But, the solution in this pr works locally. When sending a `SIGKILL` the `trinity-beacon` process (via `ctrl-c`), it successfully kills all processes, and removes all dangling ipc files from the xdg: (`ipcs-beacon`) directory.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/trinity/components/eth2/beacon/component.py
+++ b/trinity/components/eth2/beacon/component.py
@@ -261,6 +261,6 @@ class BeaconNodeComponent(AsyncioIsolatedComponent):
                     await stack.enter_async_context(run_service(service))
 
                 await asyncio.gather(*(
-                    service.cancellation()
+                    service.threadsafe_cancel()
                     for service in services
                 ))


njgheorghita:
Any guidance on what would be the best approach to testing this would be lovely!

ralexstokes:
oh awesome! didn't see you were working on this!!

i'm not as familiar w/ all of this machinery yet either... but i can say my vague intention here was to look at just using the `trio` services as much as possible -- i don't think we have too much `asyncio`-dependent stuff w/ eth2 yet that we can't heavily bias that way...

ralexstokes:
btw you can run `trinity-beacon fix-unclean-shutdown` which has been reliably cleaning up after a hung process ime

njgheorghita:
I'll just close this pr for now - and circle back to it if necessary down the road

Issue #1435, opened 2019-12-27
wheel-cli/long-run-integration tests fail intermittently


gsalgado:
https://circleci.com/api/v1.1/project/github/ethereum/trinity/168014/output/108/0?file=true&allocation-id=5e04f8648b44031be926bab5-0-build%2F290499C8

Not clear to me which tests are failing, but there are a few tracebacks for an EOFError in the logs for the test above at https://circleci.com/api/v1.1/project/github/ethereum/trinity/168014/output/108/0?file=true&allocation-id=5e04f8648b44031be926bab5-0-build%2F290499C8

cburgdorf:
```
Traceback (most recent call last):\n'
  File "<string>", line 1, in <module>\n'
  File "/usr/local/lib/python3.7/multiprocessing/spawn.py", line 105, in spawn_main\n'
    exitcode = _main(fd)\n'
  File "/usr/local/lib/python3.7/multiprocessing/spawn.py", line 113, in _main\n'
    preparation_data = reduction.pickle.load(from_parent)\n'
EOFError: Ran out of input\n'
```

cburgdorf:
This seems to be the most frequent cause of error currently. It seems pretty recent so I suspect it has something to do with our recent change in how we spawn processes. /cc @pipermerriam

gsalgado:
I've had a look at this again and it seems like the error can happen on any of the trinity-cli tests. And some of those tests don't even fail when that error happens (e.g https://circleci.com/api/v1.1/project/github/ethereum/trinity/176696/output/108/0?file=true&allocation-id=5e1f2c6db9b49522ebe888d1-0-build%2F67431DD8 was a successful run but `test_expected_logs_for_full_mode` did raise that EOFError)

I haven't managed to reproduce it locally, so am giving up for now. Has anybody managed to reproduce it locally?

cburgdorf:
The only tests that would fail are those that actively scan for errors via `run_command_and_detect_errors` or `scan_for_errors` directly. 

>was a successful run but test_expected_logs_for_full_mode did raise that EOFError)

Right, that test does only care if the right logs are written but doesn't care about any unrelated logs/errors coming up.

>I haven't managed to reproduce it locally, so am giving up for now.

I tried to reproduce it and I also tried to google the hell out of it. I would bet it is something inside the `asyncio_run_in_process` library that recently took over the launch of the processes.

[...]

PR #1615, opened 2020-03-18
Implement beacon node API for the validator client

ralexstokes:
### What was wrong?

The eth2 validator client was missing its half of the API to interact with the beacon node.

### How was it fixed?

This PR adds a first pass at adding this API, noting that some details are likely to change as the rest of the repo is updated to later spec versions.

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/typing.py
+++ b/eth2/beacon/typing.py
@@ -1,6 +1,6 @@
 from typing import Any, NamedTuple, NewType, Sequence, Tuple
 
-from eth_typing import Hash32
+from eth_typing import BLSSignature, Hash32
 from typing_extensions import Protocol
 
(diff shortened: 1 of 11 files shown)


ralexstokes:
CI failure seem unrelated so I'll merge this in so I can move ahead w/ the beacon node half of this API

cburgdorf:
>CI failure seem unrelated

It's the infamous `EOFError` that has been pestering us ever since we started using `asyncio-run-in-process` but nobody really knows why exactly. Because the bug itself *seems* to be related to our trio components. I recommend to read this issue from this [comment to the next](https://github.com/ethereum/trinity/issues/1435#issuecomment-586974614) if you are curious.

PR #1616, opened 2020-03-19
Introduce utility to generate network genesis data

ralexstokes:
### What was wrong?

We need a single point of entry to generate a (self-contained) piece of data that describes a network genesis for both the beacon node and the validator client.

### How was it fixed?

Introduce an `Application` to accomplish this task; will deprecate the `Interop` component in a subsequent PR. NOTE: The genesis data is written in a custom JSON format that only Trinity nodes will know how to handle.

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/tools/builder/initializer.py
+++ b/eth2/beacon/tools/builder/initializer.py
@@ -1,23 +1,29 @@
-from typing import Any, Dict, Sequence, Tuple, Type
+from typing import Any, Callable, Dict, Sequence, Tuple, Type
 
 from eth.constants import ZERO_HASH32
 from eth_typing import BLSPubkey, Hash32
 from eth_utils import decode_hex
+from eth_utils.toolz import curry
 from py_ecc.optimized_bls12_381.optimized_curve import (
     curve_order as BLS12_381_CURVE_ORDER,
 )
 
+from eth2._utils.bls import bls
 from eth2._utils.hash import hash_eth2
+from eth2._utils.merkle.common import MerkleTree, get_merkle_proof
 from eth2.beacon.constants import ZERO_TIMESTAMP
 from eth2.beacon.genesis import get_genesis_block, initialize_beacon_state_from_eth1
-from eth2.beacon.tools.builder.validator import create_mock_deposit_data
+from eth2.beacon.tools.builder.validator import (
+    create_deposit_data,
+    create_mock_deposit_data,
+)
 from eth2.beacon.types.blocks import BaseBeaconBlock
 from eth2.beacon.types.deposit_data import DepositData  # noqa: F401
 from eth2.beacon.types.deposits import Deposit
 from eth2.beacon.types.eth1_data import Eth1Data
 from eth2.beacon.types.states import BeaconState
 from eth2.beacon.types.validators import Validator
-from eth2.beacon.typing import Timestamp
+from eth2.beacon.typing import Gwei, Timestamp
 from eth2.beacon.validator_status_helpers import activate_validator
 from eth2.configs import Eth2Config
 
@@ -36,6 +42,10 @@ def generate_privkey_from_index(index: int) -> int:
 def create_keypair_and_mock_withdraw_credentials(
     config: Eth2Config, key_set: Sequence[Dict[str, Any]]
 ) -> Tuple[Tuple[BLSPubkey, ...], Tuple[int, ...], Tuple[Hash32, ...]]:
+    """
+    NOTE: this function mixes the parsing of keying material with the generation of derived values.
+    Prefer other functions in this module that do the derivation directly.
+    """
     pubkeys: Tuple[BLSPubkey, ...] = ()
     privkeys: Tuple[int, ...] = ()
     withdrawal_credentials: Tuple[Hash32, ...] = ()
@@ -52,6 +62,64 @@ def create_keypair_and_mock_withdraw_credentials(
         withdrawal_credentials += (withdrawal_credential,)
 
     return (pubkeys, privkeys, withdrawal_credentials)
+
+
+def create_key_pairs_for(validator_count: int) -> Dict[BLSPubkey, int]:
+    """
+    Generates ``validator_count`` key pairs derived in a deterministic manner based
+    on the validator index in the ``range(validator_count)``.
+
+    Returns a second map associating a public key with the validator's index in the set.
+    """
+    key_pairs = {}
+    for i in range(validator_count):
+        private_key = generate_privkey_from_index(i)
+        public_key = bls.privtopub(private_key)
+        key_pairs[public_key] = private_key
+    return key_pairs
+
+
+@curry
+def mk_withdrawal_credentials_from(prefix: bytes, public_key: BLSPubkey) -> Hash32:
+    return Hash32(prefix + hash_eth2(public_key)[1:])
+
+
+def create_deposit_proof(
+    tree: MerkleTree, index: int, leaf_count: int
+) -> Tuple[Hash32, ...]:
+    length_mix_in = Hash32(leaf_count.to_bytes(32, byteorder="little"))
+    proof = get_merkle_proof(tree, index)
+    return proof + (length_mix_in,)
+
+
+def create_deposit(
+    deposit_data: DepositData, tree: MerkleTree, index: int, leaf_count: int
+) -> Deposit:
+    proof = create_deposit_proof(tree, index, leaf_count)
+    return Deposit.create(proof=proof, data=deposit_data)
+
+
+def create_genesis_deposits_from(
+    key_pairs: Dict[BLSPubkey, int],
+    withdrawal_credentials_provider: Callable[[BLSPubkey], Hash32],
+    amount_provider: Callable[[BLSPubkey], Gwei],
+) -> Tuple[Deposit, ...]:
+    deposit_data = tuple(
+        create_deposit_data(
+            public_key,
+            private_key,
+            withdrawal_credentials_provider(public_key),
+            amount_provider(public_key),
+        )
+        for public_key, private_key in key_pairs.items()
+    )
+
+    deposits: Tuple[Deposit, ...] = tuple()
+    for index, data in enumerate(deposit_data):
+        data_to_index = deposit_data[: index + 1]
+        tree, _ = make_deposit_tree_and_root(data_to_index)
+        deposits += (create_deposit(data, tree, index, len(data_to_index)),)
+    return deposits
 
 
 def create_mock_deposits_and_root(
--- a/eth2/beacon/tools/builder/initializer.py
+++ b/eth2/beacon/tools/builder/initializer.py
@@ -1,8 +1,8 @@
-from typing import Any, Callable, Dict, Sequence, Tuple, Type
+from typing import Any, Callable, Dict, Iterable, Sequence, Tuple, Type
 
 from eth.constants import ZERO_HASH32
 from eth_typing import BLSPubkey, Hash32
-from eth_utils import decode_hex, encode_hex
+from eth_utils import decode_hex, encode_hex, to_dict, to_tuple
 from eth_utils.toolz import curry
 from py_ecc.optimized_bls12_381.optimized_curve import (
     curve_order as BLS12_381_CURVE_ORDER,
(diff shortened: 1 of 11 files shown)


eth2/beacon/tools/builder/initializer.py
```diff
@@ -54,6 +106,64 @@ def create_keypair_and_mock_withdraw_credentials(
...
+    Generates ``validator_count`` key pairs derived in a deterministic manner based
+    on the validator index in the ``range(validator_count)``.
+
+    Returns a second map associating a public key with the validator's index in the set.
+    """
+    key_pairs = {}
+    for i in range(validator_count):
+        private_key = generate_privkey_from_index(i)
+        public_key = bls.privtopub(private_key)
+        key_pairs[public_key] = private_key
```
pipermerriam: We have `eth_utils.to_dict` which allows you to do:

```python
@to_dict
def create_key_pairs_for(...):
    for i in range(validator_count):
        ...
        yield public_key, private_key
```

eth2/beacon/tools/builder/initializer.py
```diff
@@ -54,6 +106,64 @@ def create_keypair_and_mock_withdraw_credentials(
...
+            amount_provider(public_key),
+        )
+        for public_key, private_key in key_pairs.items()
+    )
+
+    deposits: Tuple[Deposit, ...] = tuple()
+    for index, data in enumerate(deposit_data):
+        data_to_index = deposit_data[: index + 1]
+        tree, _ = make_deposit_tree_and_root(data_to_index)
+        deposits += (create_deposit(data, tree, index, len(data_to_index)),)
```
pipermerriam: This should also be trivial to replace with `eth_utils.to_tuple` and just `yield create_deposit(...)` on each loop.

eth2/configs.py
```diff
@@ -60,6 +62,31 @@
 )
 
 
+def _serialize(item: Any) -> Union[int, str]:
+    if isinstance(item, bytes):
+        return encode_hex(item)
+    else:
+        return int(item)
```
pipermerriam: this jumped out to me as confusing.  We have an `Any` type anytime it isn't `bytes` we just blindly cast it to an integer?  Maybe worth a comment?

eth2/configs.py
```diff
@@ -60,6 +62,31 @@
...
+
+
+def serialize(config: Eth2Config) -> Tuple[Union[int, str], ...]:
+    return tuple(_serialize(item) for item in config)
+
+
+def _deserialize(item: Union[int, str], typ: Any) -> Any:
```
pipermerriam: I think you want to write this function using a `TypeVar`

```python
T = TypeVar('T')
def _deserialize(items: ..., typ: Type[T]) -> T:
    ...
```

Should let you get rid of the `Any` type.

eth2/configs.py
```diff
@@ -60,6 +62,31 @@
...
+    if typ is bytes:
+        return decode_hex(cast(str, item))
+    else:
+        return typ(item)
+
+
+def deserialize(config: Tuple[Union[int, str], ...]) -> Eth2Config:
+    values: Tuple[Any, ...] = ()
+    for index, (_, typ) in enumerate(Eth2Config._field_types.items()):
```
pipermerriam: Looks like `_field_types` should be made into or exposed as a public accessor to avoid accessing the private value.

trinity/components/eth2/beacon/component.py
```diff
@@ -104,9 +67,7 @@ def is_enabled(self) -> bool:
     @classmethod
     def _load_or_create_node_key(cls, boot_info: BootInfo) -> KeyPair:
         if boot_info.args.beacon_nodekey:
-            privkey = Secp256k1PrivateKey.new(
-                decode_hex(boot_info.args.beacon_nodekey)
-            )
+            privkey = Secp256k1PrivateKey.new(decode_hex(boot_info.args.beacon_nodekey))
```
pipermerriam: This module looks like the changes are mostly `black` code reformatting?  I'd suggest separating those changes from any actual changes since it isn't trivial to separate the two.  Should also reduce the total diff size of this PR to a more manageable size.

trinity/components/eth2/interop/component.py
```diff
@@ -159,23 +126,23 @@ def munge_all_args(cls, args: Namespace, trinity_config: TrinityConfig) -> None:
             start_time = now + args.start_delay
             cls.logger.info("Genesis time is %d", start_time)
 
-            state = state.set(
-                "genesis_time", start_time
-            )
+            state = state.set("genesis_time", start_time)
```
pipermerriam: Same here, lots of pure formatting changes that make it hard to see what has actually changed.

eth2/beacon/tools/builder/initializer.py
ralexstokes: ok this is nice

eth2/beacon/tools/builder/initializer.py
ralexstokes: also nice

eth2/configs.py
ralexstokes: the broader context here is that i'm leaning towards reworking how this config is defined so these functions are temporary just to get the "genesis data" working for now in this PR...

i agree this may not be clear but if you look at the fields of the config they are all integers except for one `bytes`field. so on `serialization` we can be lossy.

i think defining the config as an `ssz` object is overkill but we do have a very nice way to go back/forth to `json` in `py-ssz` which i'm taking advantage of... if you can point me to something that can automatically derive this serialization code that is lighter than pulling in `py-ssz.tools` that would be great :)

eth2/configs.py
ralexstokes: was just going for something quick as i'm thinking to rework how the config is defined (and ideally get "automatic" {de,}serialization out of it)...

it seems like this `TypeVar` would fail the type checker when `typ` is `bytes` (as the return type is `str` in that case...)

eth2/configs.py
ralexstokes: these are in the stdlib docs... referring to the above i think `NamedTuple` is not exactly what we want here and the serialization change for the `Eth2Config` in this PR are temporary

trinity/components/eth2/beacon/component.py
ralexstokes: agreed, i didn't expect anyone to review this particular PR --- thanks for looking it over :) i'll pull out the formatting changes into a separate PR and rebase on top

eth2/configs.py
pipermerriam: `decode_hex` returns a `bytes` value so I believe that this should still behave correctly using `TypeVar` approach.

ralexstokes: ah, you are right; i had the direction of the transformation backwards!

trinity/components/eth2/beacon/component.py
```diff
@@ -2,15 +2,15 @@
 import asyncio
```
ralexstokes: changes in this file are the beginning steps to remove the in-process validator in lieu of the validator client (which this PR is setting the ground work for)

trinity/components/eth2/discv5/component.py
```diff
@@ -39,8 +40,9 @@
 
 DEFAULT_NODEDB_DIR_NAME = "nodes"
 
-
-logger = logging.getLogger(__name__)
+logger = logging.getLogger(
```
ralexstokes: somewhat unrelated to this PR, but the way we current have logging set up, it would just log `component` as the name of the thing dumping `discv5` logs which was very apparent when we only have the `BeaconNode` and `discv5` components.

this set of changes patches up the logger name to be more descriptive

trinity/components/eth2/interop/component.py
```diff
@@ -139,11 +138,11 @@ def munge_all_args(cls, args: Namespace, trinity_config: TrinityConfig) -> None:
 
         # Save the genesis state to the data dir!
         yaml = YAML(typ="unsafe")
-        with open(trinity_config.trinity_root_dir / GENESIS_FILE, "w") as f:
+        with open(trinity_config.trinity_root_dir / "genesis", "w") as f:
```
ralexstokes: in line w/ the message of the commit that introduced these changes, this is temporary to clean up these constants we don't need for directories/files that won't exist anymore

trinity/components/eth2/network_generator/component.py
```diff
@@ -0,0 +1,113 @@
...
+            ),
+            "genesis_state": to_formatted_dict(genesis_state),
+        }
+        cls.logger.info(
+            "configuration generated; genesis state has root %s",
+            humanize_hash(genesis_state.hash_tree_root),
+        )
+        output_file_path.parent.mkdir(parents=True, exist_ok=True)
+        with open(output_file_path, "w") as output_file:
+            output_file.write(json.dumps(output))
```
ralexstokes: the point of this component is to generate a **single** config file that has everything a beacon node or validator client would need to start from genesis.

trinity/components/registry.py
```diff
@@ -111,4 +111,4 @@ def get_components_for_eth1_client() -> Tuple[Type[BaseComponentAPI], ...]:
 
 
 def get_components_for_beacon_client() -> Tuple[Type[BaseComponentAPI], ...]:
-    return get_all_components(*BEACON_NODE_COMPONENTS)
+    return BEACON_NODE_COMPONENTS
```
ralexstokes: the peer discovery component was not playing nicely and we have the discv5 component that should replace it either way -- drop down to specifically specify eth2 components for now

trinity/config.py
```diff
@@ -19,7 +19,6 @@
     Any,
```
ralexstokes: changes in this file are also somewhat WIP-y but moving towards launching the beacon node from the aforementioned config file this PR generates

ralexstokes:
rebased away the formatting changes, added some clarifying comments that will be helpful for anyone reviewing... i'll incorporate the comments above soon and then this will be ready for a final review before merge

eth2/configs.py
ralexstokes: this approach caused more issues w/ `mypy` so i'm electing to just keep the `Any` type for now with the understanding that an already typed solution will land in a PR landing soon... i can prioritize this PR if the `Any` is really bothersome

ralexstokes:
CI failure is unrelated so I'm going to move ahead w/ merge.

@pipermerriam if you want to make another pass, please do! i left some comments to clarify where i'm heading... but i would like to merge this in so other PRs on top don't become a huge chain dependent on this one

eth2/beacon/tools/builder/initializer.py
```diff
@@ -1,29 +1,77 @@
...
+    key_pairs: Dict[BLSPubkey, int],
+    genesis_state: BeaconState,
+    public_key_codec: Callable[[BLSPubkey], str] = encode_hex,
+    private_key_codec: Callable[[int], str] = _encode_private_key_as_hex,
+) -> Tuple[Dict[str, str], ...]:
+    key_map: Tuple[Dict[str, str], ...] = ()
+    for _, validator in enumerate(genesis_state.validators):
+        public_key = validator.pubkey
+        private_key = key_pairs[public_key]
+        key_map += (
```
pipermerriam: This is another place where the `to_tuple` + `yield` pattern would remove mutability.

eth2/beacon/tools/builder/initializer.py
```diff
@@ -1,29 +1,77 @@
...
+    public_key_codec: Callable[[str], BLSPubkey] = lambda key: BLSPubkey(
+        decode_hex(key)
+    ),
+    private_key_codec: Callable[[str], int] = _decode_private_key_from_hex,
+) -> Dict[BLSPubkey, int]:
+    key_map = {}
+    for key_pair in encoded_key_map:
+        public_key = public_key_codec(key_pair["public_key"])
+        private_key = private_key_codec(key_pair["private_key"])
+        key_map[public_key] = private_key
```
pipermerriam: This could also use the `to_dict` to remove mutability.

eth2/beacon/tools/builder/initializer.py
```diff
@@ -54,6 +106,62 @@ def create_keypair_and_mock_withdraw_credentials(
...
+
+@curry
+def mk_withdrawal_credentials_from(prefix: bytes, public_key: BLSPubkey) -> Hash32:
+    return Hash32(prefix + hash_eth2(public_key)[1:])
+
+
+def create_deposit_proof(
+    tree: MerkleTree, index: int, leaf_count: int
+) -> Tuple[Hash32, ...]:
+    length_mix_in = Hash32(leaf_count.to_bytes(32, byteorder="little"))
```
pipermerriam: Given that this is just a `tools` module it doesn't really matter, but a while back @hwwhww figured out that there is a measurable performance difference between specifying the `byteorder` argument via keyword or positionally (with the positional approach being non-trivially faster).

trinity/components/eth2/discv5/component.py
```diff
@@ -39,8 +40,9 @@
...
+    re.sub(re.escape("component") + "$", "DiscV5Component", __name__)
```
pipermerriam: This seems weird when you could just specify the path fully.  This only seems to make sense if we assume `__name__` is going to change often (such as if this file moved around a bunch).  I think just giving this a static name might be a better choice as this line isn't the easiest to grok.

eth2/beacon/tools/builder/initializer.py
ralexstokes: hmmm good to know

trinity/components/eth2/discv5/component.py
ralexstokes: fair enough... i do agree that this component shouldn't move much so the thrash on a static name should be minimal

PR #1620, opened 2020-03-20
Integrate validator client into trinity platform

ralexstokes:
### What was wrong?

The validator client was developed separately from the rest of the Trinity platform so it was lacking some of the niceties the platform provides.

### How was it fixed?

Integrate the validator client into the platform, taking care to avoid some troublesome areas that it seems like the validator client *can* avoid for now (i.e. the multiprocessing elements that have historically lead to hard-to-control processes).

#### Cute Animal Picture

### DIFF
--- a/eth2/validator_client/cli_parser.py
+++ b/eth2/validator_client/cli_parser.py
@@ -2,16 +2,19 @@ import argparse
 import getpass
 import logging
 import signal
+from typing import Dict
 
 import argcomplete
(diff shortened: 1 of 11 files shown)


ralexstokes:
This PR builds off the work in #1616 to consume the beacon node/validator client-shared configuration.

ralexstokes:
happy to address post-merge feedback!

PR #1621, opened 2020-03-21
Fix an API issue with KeyStore creation

ralexstokes:
### What was wrong?

A `KeyStore` was (awkwardly) a property of a `Config` as the `Config` kept a record of any `KeyPair`s the user supplied to the validator client.

NOTE: this is mostly a quick fix intended to patch up an API deficiency from an earlier PR :)

### How was it fixed?

By separating the `key_pairs` away from the `KeyStore` construction we gain the flexibility to make `KeyStore`s as required independently of the collection of key pairs.

#### Cute Animal Picture

### DIFF
--- a/trinity/main_validator.py
+++ b/trinity/main_validator.py
@@ -1,9 +1,8 @@
 import json
 import logging
 from pathlib import Path
-from typing import Any, Callable, Dict
+from typing import Any, Dict
 
-from eth_typing import BLSPubkey
 from ssz.tools.parse import from_formatted_dict
 import trio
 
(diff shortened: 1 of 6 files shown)


trinity/main_validator.py
```diff
@@ -69,12 +57,12 @@ def main_validator() -> None:
...
     arguments = parse_cli_args()
     root_data_dir = (
-        Path(os.environ["HOME"]) /
-        ".local" /
-        "share" /
-        "trinity" /
-        "eth2" /
-        "validator_client"
+        Path(os.environ["HOME"])
+        / ".local"
```
pipermerriam: You should be using the [`trinity._utils.xdg.get_xdg_data_home`](https://github.com/ethereum/trinity/blob/11d2fa1a58dfad3baa8c86a1f06ca8b3928e7fb5/trinity/_utils/xdg.py#L31) utility function to build the base path here since that one respects the `$XDG_DATA_HOME` environment variable.

ralexstokes: agreed; and this is part of the idea of #1620 to integrate w/ the configuration machinery of the trinity platform; this specific change is here: https://github.com/ethereum/trinity/pull/1620/files#diff-9d3f000652992dc39b0df738cbccc05bR56-R67

Issue #1397, opened 2019-12-13
Convert `DBClient.connect()` to directly be a context manager.


pipermerriam:
### What is wrong

from https://github.com/ethereum/trinity/pull/1385#pullrequestreview-332041499

The following pattern is used in a lot of places:

```python
db = DBClient.connect(...)
with db:
    # do things....
```

This pattern is error prone because it allows you to directly use the `db` object even if you aren't using the context management APIs.

### How to fix.

Look into just directly embedding the context management API into `DBClient.connect()` so that you can do:

```python
with DBClient.connect(...) as db:
    ...
```

PR #1622, opened 2020-03-22
Convert DBClient.connect() to be a context manager

onyb:
### What was wrong?

Ref: #1397 

### How was it fixed?

The new API for `DBClient.connect` is as follows:

```py
with DBClient.connect(...) as db:
    ...
```

#### Cute Animal Picture

### DIFF
--- a/trinity/db/manager.py
+++ b/trinity/db/manager.py
@@ -348,12 +348,17 @@ class DBClient(BaseAtomicDB):
         self._socket.close()
 
     @classmethod
-    def connect(cls, path: pathlib.Path, timeout: int = 5) -> "DBClient":
+    @contextlib.contextmanager
+    def connect(cls, path: pathlib.Path, timeout: int = 5) -> Iterator["DBClient"]:
         wait_for_ipc(path, timeout)
         s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         cls.logger.debug("Opened connection to %s: %s", path, s)
         s.connect(str(path))
-        return cls(s)
+        db = cls(s)
+        try:
+            yield db
+        finally:
+            db.close()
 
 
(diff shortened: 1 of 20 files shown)


trinity/db/manager.py
```diff
@@ -348,12 +348,15 @@ def close(self) -> None:
...
+    @contextlib.contextmanager
+    def connect(cls, path: pathlib.Path, timeout: int = 5) -> Iterator["DBClient"]:
         wait_for_ipc(path, timeout)
         s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         cls.logger.debug("Opened connection to %s: %s", path, s)
         s.connect(str(path))
-        return cls(s)
+        db = cls(s)
+        yield db
+        db.close()
```
pipermerriam: These two lines should typically be written using a `try/finally` to ensure the `close` happens even in the case of an exception.

```python
try:
    yield
finally:
    db.close()
```

onyb: Fixed

onyb:
@pipermerriam Not sure if the `py{36,37}-wheel-cli` tests are expected to fail. Are the errors on the CI something you recognize?

cburgdorf:
@onyb all CI jobs are expected to succeed. That said, the `py{36,37}-wheel-cli` wheel jobs have been unstable for some weeks. It is mostly [the pesty `EOFError` described here](https://github.com/ethereum/trinity/issues/1435) when it fails. But in your case, it is something different. It looks like the tests that spin up Trinity and then attach to it via `trinity attach` to run assertions against the `web3` console fail. I checked it locally and the error persists so it seems legit.

You can run `pytest tests/integration/test_trinity_cli.py -k 'test_web3_commands_via_attached_console'` to dig in locally.

PR #1625, opened 2020-03-22
Cleanly factor IO into `ContextManager` of validator `KeyStore`

ralexstokes:
### What was wrong?

There were two types of `KeyStore`, in-memory or one that persists keys to disk, and by factoring the IO into the `ContextManager` part of the `KeyStore`, the two classes could be merged.

This is cleaning up the `KeyStore` so there is a cleaner separation of handling keys and persisting keys to disk.

#### Cute Animal Picture

### DIFF
--- a/eth2/validator_client/abc.py
+++ b/eth2/validator_client/abc.py
@@ -70,7 +70,7 @@ class KeyStoreAPI(ContextManager["KeyStoreAPI"]):
         ...
 
     @abstractmethod
-    def import_private_key(self, encoded_private_key: str, password: bytes) -> None:
+    def import_private_key(self, encoded_private_key: str) -> None:
         ...
 
     @abstractmethod
--- a/eth2/validator_client/cli_parser.py
+++ b/eth2/validator_client/cli_parser.py
@@ -1,5 +1,4 @@
 import argparse
-import getpass
 import logging
 import signal
 
@@ -43,11 +42,8 @@ async def _import_key(
     logger.info("importing private key...")
     try:
         key_store = KeyStore.from_config(config)
-        logger.warn(
-            "please enter a password to protect the key on-disk (can be empty):"
-        )
-        password = getpass.getpass().encode()
-        key_store.import_private_key(arguments.private_key, password)
+        with key_store.persistence():
+            key_store.import_private_key(arguments.private_key)
     except Exception:
         logger.exception("error importing key")
 
--- a/eth2/validator_client/cli_parser.py
+++ b/eth2/validator_client/cli_parser.py
@@ -6,15 +6,14 @@ import argcomplete
 from async_service.trio import background_trio_service
 import trio
 
+from eth2.validator_client.beacon_node import MockBeaconNode as BeaconNode
 from eth2.validator_client.client import Client
+from eth2.validator_client.clock import Clock
 from eth2.validator_client.config import Config
 from eth2.validator_client.key_store import KeyStore
+from eth2.validator_client.tools.password_providers import terminal_password_provider
 from trinity.cli_parser import parser, subparser
 
-DEMO_MODE_HELP_MSG = (
-    "set configuration suitable for demonstration purposes (like ignoring a password)."
-    " Do NOT use in production."
-)
 IMPORT_PARSER_HELP_MSG = (
     "import a validator private key to the keystore discovered from the configuration"
 )
@@ -30,10 +29,16 @@ async def _wait_for_interrupts() -> None:
 async def _main(
     logger: logging.Logger, config: Config, arguments: argparse.Namespace
 ) -> None:
-    client = Client.from_config(config)
-    async with background_trio_service(client):
-        await _wait_for_interrupts()
-        logger.info("received interrupt; shutting down...")
+    key_store = KeyStore.from_config(config)
+    clock = Clock.from_config(config)
+    beacon_node = BeaconNode.from_config(config)
+
+    # with key_store.persistence():
+    async with beacon_node:
+        client = Client(key_store, clock, beacon_node)
+        async with background_trio_service(client):
+            await _wait_for_interrupts()
+            logger.info("received interrupt; shutting down...")
 
 
 async def _import_key(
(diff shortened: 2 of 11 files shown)


eth2/validator_client/abc.py
```diff
@@ -70,7 +70,7 @@ def public_keys(self) -> Collection[BLSPubkey]:
         ...
 
     @abstractmethod
-    def import_private_key(self, encoded_private_key: str, password: bytes) -> None:
+    def import_private_key(self, encoded_private_key: str) -> None:
```
ralexstokes: do not leak concerns of "IO" into the abstract interface

eth2/validator_client/cli_parser.py
```diff
@@ -31,30 +29,36 @@
...
 ) -> None:
-    client = Client.from_config(config)
-    async with background_trio_service(client):
-        await _wait_for_interrupts()
-        logger.info("received interrupt; shutting down...")
+    key_store = KeyStore.from_config(config)
+    clock = Clock.from_config(config)
+    beacon_node = BeaconNode.from_config(config)
+
+    # with key_store.persistence():
```
ralexstokes: for now, keys are generated in memory and passed via `config.key_pairs`... eventually will want to only take keys provided by the user (e.g. via `trinity-beacon import-key`) -- disable the disk check for now

eth2/validator_client/cli_parser.py
```diff
@@ -31,30 +29,36 @@
...
-        logger.warn(
-            "please enter a password to protect the key on-disk (can be empty):"
+        key_store = KeyStore(
+            key_store_dir=config.key_store_dir,
+            password_provider=terminal_password_provider,
         )
-        password = getpass.getpass().encode()
-        key_store.import_private_key(arguments.private_key, password)
+        with key_store.persistence(should_load_existing_key_pairs=False):
+            key_store.import_private_key(arguments.private_key)
```
ralexstokes: this is the setting to get 1 password from the terminal and ignore anything else on-disk (there could be many keys in the key store the user wants to ignore in this configuration).

eth2/validator_client/client.py
```diff
@@ -123,17 +111,9 @@ def from_config(cls, config: Config) -> "Client":
...
-
-        with self._key_store:
-            self.logger.info(
-                "found %d validator key pair(s) for public key(s) %s",
-                len(self._key_store.public_keys),
-                tuple(map(humanize_bytes, self._key_store.public_keys)),
-            )
-            async with self._beacon_node:
-                await self._verify_client_state_at_boot()
-                await self._run_client()
```
ralexstokes: let the caller decide how they want to initialize the various dependencies

tests-trio/eth2/validator_client/conftest.py
```diff
@@ -14,7 +14,7 @@ def sample_bls_public_key(sample_bls_private_key):
 
 
 @pytest.fixture
-def sample_bls_key_pair(sample_bls_private_key, sample_bls_public_key):
+def sample_bls_key_pairs(sample_bls_private_key, sample_bls_public_key):
```
ralexstokes: just fixing the name

ralexstokes:
also noting a similar unification could happen with `BeaconNode` and `MockBeaconNode` classes; or possibly that we just move the `Mock` class into testing code as the validator client codebase matures...

ralexstokes:
going to go ahead and merge in; happy to address post-merge feedback :)

eth2/validator_client/key_store.py
```diff
@@ -14,48 +22,91 @@
...
 
     @classmethod
     def from_config(cls, config: Config) -> "KeyStore":
-        return cls(config.key_store_dir, config.key_pairs, config.demo_mode)
+        return cls(config.key_pairs, config.key_store_dir, insecure_password_provider)
+
+    def _ensure_dirs(self) -> None:
+        did_create = create_dir_if_missing(self._key_store_dir)
+        if did_create:
+            self.logger.warning(
```
pipermerriam: Seems like this *might* be more appropriate at the `INFO` level.

eth2/validator_client/key_store.py
```diff
@@ -14,48 +22,91 @@
...
-        self._key_files: Dict[str, Dict[str, Any]] = {}
+        self._key_pairs = key_pairs if key_pairs else {}
+        self._key_store_dir = key_store_dir
+        self._password_provider = password_provider
 
-        self._ensure_dirs()
+        self._key_files: Dict[str, Dict[str, Any]] = {}
+        # Mapping of public key to key file ID
+        self._key_file_index: Dict[BLSPubkey, str] = {}
+        self._should_load_existing_key_pairs = False
```
pipermerriam: This *could* be set at the class level.

eth2/validator_client/tools/password_providers.py
```diff
@@ -0,0 +1,14 @@
+import getpass
+
+from eth_typing import BLSPubkey
+from eth_utils import encode_hex
+
+
+def terminal_password_provider(public_key: BLSPubkey) -> bytes:
+    return getpass.getpass(
+        f"Please enter password for keyfile with public key {encode_hex(public_key)}:"
+    ).encode()
```
pipermerriam: cc @njgheorghita You've done a lot of "prompting" in the `ethpm-cli`.  Any suggestions here?

njgheorghita: I've been using [`input()`](https://docs.python.org/3/library/functions.html#input) to capture user input. Though since `getpass` hides the password when it's typed (if that's the desired ux) it's probably the better option for this scenario.

pipermerriam: I was intending to refer to any mechanisms you may have used for allowing the user to retry on a failed response or whatnot, but they may not apply in this case which is ok.

ralexstokes: yeah -- which calls to mind we should do a confirmation flow where the passwords have to match.

this all being said, i'm not sure how much we want to commit to making a complete solution here -- bc in the limit we are just building a signing wallet which imo to do well is its own project.

i'd hope before production we have hardware wallets for BLS that we would plug into the validator client.

i'll make a tracking issue for the double confirmation flow

PR #1627, opened 2020-03-23
Partly implement eth/65

cburgdorf:
### What was wrong?

Before we can add support for `eth/66` we need to add support for [`eth/65`](https://github.com/ethereum/EIPs/pull/2464/files). That said, `eth/65` is all about optimizing transaction propagation, given that we do not have a fully fledged transaction pool yet, this PR will only partly implement what `eth/65` specifies.

### How was it fixed?

1. Add support for `NewPooledTransactionHashes`, `GetPooledTransactions` and `PooledTransactions`.

2. Wire up all the infrastructure to support these commands across the event bus and the proxy peer. 

3. Make our "Transaction Pool" answer `GetPooledTransactions` with empty responses (valid response)

#### What does that mean exactly?

The PR lays the infrastructure to support `eth/65` and recognize the new command types. Further changes to fully support `eth/65` would only need to happen in the transaction pool directly. Put another way, a 3rd party transaction pool component could be written that fully supports `eth/65` without any further needed changes to the core of Trinity.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/eth/commands.py
+++ b/trinity/protocol/eth/commands.py
@@ -8,7 +8,7 @@ from eth_utils.curried import (
 from eth_utils.toolz import compose
 from rlp import sedes
 
-from eth.abc import BlockHeaderAPI, ReceiptAPI
+from eth.abc import BlockHeaderAPI, ReceiptAPI, SignedTransactionAPI
 from eth.rlp.headers import BlockHeader
 from eth.rlp.receipts import Receipt
 from eth.rlp.transactions import BaseTransactionFields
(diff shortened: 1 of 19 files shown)


trinity/protocol/eth/commands.py
```diff
@@ -159,6 +159,27 @@ class NewBlock(BaseCommand[NewBlockPayload]):
...
+class GetPooledTransactionHashes(BaseCommand[Tuple[Hash32, ...]]):
+    protocol_command_id = 9
+    serialization_codec: RLPCodec[Tuple[Hash32, ...]] = RLPCodec(
+        sedes=sedes.CountableList(sedes.List([hash_sedes])),
+    )
+
+
+class PooledTransactions(BaseCommand[Tuple[BaseTransactionFields, ...]]):
+    protocol_command_id = 10
+    serialization_codec: RLPCodec[Tuple[BaseTransactionFields, ...]] = RLPCodec(
```
pipermerriam: Grumble, did they fail to put a `request_id` in these commands?!

cburgdorf: That is coming with `eth/66`

trinity/protocol/eth/validators.py
```diff
@@ -88,3 +89,26 @@ def validate_result(self, response: BlockBodyBundles) -> None:
...
+    def validate_result(self, response: Tuple[SignedTransactionAPI, ...]) -> None:
+        if not response:
+            # an empty response is always valid
+            return
+
+        tx_hashes = tuple(tx.hash for tx in response)
+        tx_hash_set = set(tx_hashes)
+
+        if len(tx_hashes) != len(tx_hash_set):
+            raise ValidationError("Response may not contain duplicate tx hashes")
```
pipermerriam: It always takes more work but actually reporting the duplicates in a case like this always seems worth it.

Issue #1302, opened 2019-11-21
Trinity <-> Trinity test infrastructure


carver:
### What is wrong?

We need a full-featured, reliable test suite to make sure mainnet trinity connects to mainnet trinity smoothly and consistently.

### How can it be fixed

:man_shrugging: 

Maybe very similar to existing integration tests, but instead of launching one trinity, we launch two nodes and link them together. We might consider running a dedicated node to connect to, so that we don't run as much of a risk of flaky tests, when we happen to not find any good nodes quickly.

gsalgado:
I'm having a go at this and the first issue I run into is addressed in #1608

PR #1630, opened 2020-03-24
Fix peer disconnection when idle for too long

gsalgado:
Also as part of my work on #1302, I noticed sometimes peer would get stuck in the pool even though they were no longer reachable. After spending some time implementing something to periodically ping idle peers and disconnect if they don't reply, I found the `Transport.read()` method uses a timeout of 30s, exactly so that we close idle connections. However, that was not working because the TimeoutError was being silenced. This fixes that and adds a test to ensure it works, but I've left a few XXXs because there are things I don't completely understand here.

### DIFF
--- a/p2p/multiplexer.py
+++ b/p2p/multiplexer.py
@@ -335,14 +335,13 @@ class Multiplexer(CancellableMixin, MultiplexerAPI):
         # multiplex so that we can reliably cancel it without requiring the
         # master token for the multiplexer to be cancelled.
         async with self._multiplex_lock:
-            multiplex_token = CancelToken(
+            self._multiplex_token = CancelToken(
                 'multiplex',
                 loop=self.cancel_token.loop,
             ).chain(self.cancel_token)
 
             stop = asyncio.Event()
-            self._multiplex_token = multiplex_token
-            fut = asyncio.ensure_future(self._do_multiplexing(stop, multiplex_token))
+            fut = asyncio.ensure_future(self._do_multiplexing(stop))
             # wait for the multiplexing to actually start
             try:
                 yield
@@ -398,10 +397,10 @@ class Multiplexer(CancellableMixin, MultiplexerAPI):
                     except asyncio.CancelledError:
                         pass
 
-                multiplex_token.trigger()
+                self._multiplex_token.trigger()
                 del self._multiplex_token
 
-    async def _do_multiplexing(self, stop: asyncio.Event, token: CancelToken) -> None:
+    async def _do_multiplexing(self, stop: asyncio.Event) -> None:
         """
         Background task that reads messages from the transport and feeds them
         into individual queues for each of the protocols.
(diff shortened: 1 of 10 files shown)


p2p/multiplexer.py
```diff
@@ -407,12 +408,20 @@ def stream_protocol_messages(self,
...
+            # shouldn't we cancel the peer/connection here as well?
             self.logger.warning(
                 "Timed out waiting for command from %s, Stop: %r, exiting...",
                 self,
                 stop.is_set(),
             )
             self.logger.debug("Timeout %r: %s", self, exc, exc_info=True)
+        except PeerConnectionLost:
+            # XXX: This is the multiplex_token created in multiplex(), so I don't understand how
+            # triggering it cancels the connection, but it seems to cause that.
```
pipermerriam: This is the line you are looking for: https://github.com/ethereum/trinity/blob/347e2beac23c5c1bb4aab136bb44c162467f6ff7/p2p/connection.py#L71

p2p/multiplexer.py
```diff
@@ -407,12 +408,20 @@ def stream_protocol_messages(self,
         try:
             await self._handle_commands(msg_stream, stop)
         except asyncio.TimeoutError as exc:
+            # XXX: Transport.read() used to let TimeoutErrors (from peers that have been idle to
+            # long) bubble up and those were silenced here. That is why peers could get stuck
+            # in the pool forever. I'm not sure if we still need this, and in case we do,
+            # shouldn't we cancel the peer/connection here as well?
```
pipermerriam: Yes, this whole `except` block looks suspect.  I'd experiment with removing it entirely if you're wanting the exception to bubble.

p2p/multiplexer.py
gsalgado: I don't think it is. The token here is created in https://github.com/ethereum/trinity/blob/347e2beac23c5c1bb4aab136bb44c162467f6ff7/p2p/multiplexer.py#L330, and it would get triggered if the connection/multiplexer token was triggered, but triggering `multiplex_token` would not trigger the connection/multiplexer one.

gsalgado: Ok, what happens is that when we trigger `multiplex_token` here, that causes [Multiplexer.stream_protocol_messages()](https://github.com/ethereum/trinity/blob/347e2beac23c5c1bb4aab136bb44c162467f6ff7/p2p/multiplexer.py#L288-L290) to raise an `OperationCancelled`, which in turn causes the `feed_protocol_handlers()` daemon (https://github.com/ethereum/trinity/blob/347e2beac23c5c1bb4aab136bb44c162467f6ff7/p2p/connection.py#L131) to exit, finally causing the connection to be cancelled

gsalgado:
@pipermerriam, the test failures here are actually indicating a regression. Triggering the multiplex_token causes the next async call on it to raise an `OperationCancelled`, and that bubbles up all the way to `PeerPool.connect()`, which re-raises it to stop the main loop: https://github.com/ethereum/trinity/blob/347e2beac23c5c1bb4aab136bb44c162467f6ff7/p2p/peer_pool.py#L346-L352

I'm starting to wonder if it's worth attempting to fix this or if it'd be better to use a different approach, as you suggested in https://github.com/ethereum/trinity/pull/1631#issuecomment-604187946

pipermerriam:
We've never had a good solution for being able to differentiate an `OperationCancelled` that raises from an unrelated chain of cancel tokens so I'm not sure what the right way to go about that would be.  It couldn't hurt to try and run with the [#1631 comment approach](https://github.com/ethereum/trinity/pull/1631#issuecomment-604187946) and see if it lets you solve this via a different mechanism.

gsalgado:
This has been abandoned in favour of #1648

PR #1631, opened 2020-03-24
p2p.Multiplexer now keeps track of the time the last msg was received

gsalgado:
I've implemented this because I intended to have the PeerPool
periodically check for peers that haven't sent us anything
in a long time, ping them, and if they still didn't send us
anything back after a timeout, disconnect. However, it turns
out we already have something that could achieve the same, but
it was not working as expected (see #1630).

I believe this is still useful to have anyway, and even after #1630
is merged, we might still want to periodically ping peers that we
haven't heard from in a while, to not hit the idle timeout on active
connections

### DIFF
--- a/p2p/abc.py
+++ b/p2p/abc.py
@@ -344,6 +344,11 @@ class MultiplexerAPI(ABC):
     def get_total_msg_count(self) -> int:
         ...
 
+    @property
+    @abstractmethod
+    def last_msg_time(self) -> float:
+        ...
+
     #
     # Proxy Transport properties and methods
     #
--- a/p2p/multiplexer.py
+++ b/p2p/multiplexer.py
@@ -1,5 +1,6 @@
 import asyncio
 import collections
+import time
 from typing import (
     Any,
     AsyncIterator,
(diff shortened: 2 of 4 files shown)


pipermerriam:
@gsalgado I would recommend putting the logic for pinging peers that we haven't heard from in a while into the `P2PAPI`.

gsalgado:
> @gsalgado I would recommend putting the logic for pinging peers that we haven't heard from in a while into the P2PAPI.

I think it will have to be in Peer/Connection as the P2PAPI is not a service so cannot run daemon tasks

pipermerriam:
> I think it will have to be in Peer/Connection as the P2PAPI is not a service so cannot run daemon tasks

If you take a look at:

https://github.com/ethereum/trinity/blob/75f9eccc1971cbf3fa16729237f25c35a40bbe4a/p2p/behaviors.py

You'll see that the `BehaviorAPI.apply` method is an async context manager.  That allows you to do something like run a service or even just a manually run background task and manage the lifecycle using the context management API.

gsalgado:
Oh, cool, I hadn't noticed that!

pipermerriam:
To extend on my previous comment:

https://github.com/ethereum/trinity/blob/75f9eccc1971cbf3fa16729237f25c35a40bbe4a/p2p/logic.py

You would do something like this using the `p2p.logic` APIs:

https://github.com/ethereum/trinity/blob/75f9eccc1971cbf3fa16729237f25c35a40bbe4a/p2p/logic.py

```python
class MonitorIdleConnection(BaseLogic):
    async def apply(self, connection):
        # Encode the following logic in a Service and run it as a background service within this context block.
        # 1. register a handler with the connection to handle any incoming message.
        # 2. use the handler to continually reset a timer of some sort that if it ever runs down all the way, then a ping/pong round trip is initiated.
        # 3. disconnect if the ping/pong fails
```

PR #1633, opened 2020-03-24
Add missing API docs for proxy events

cburgdorf:
### What was wrong?

As I'm working on `eth/65` I noticed our documentation on the various proxy events could be improved.

### How was it fixed?

Added API docs for various missing proxy events.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/common/events.py
+++ b/trinity/protocol/common/events.py
@@ -81,7 +81,9 @@ class PeerInfo(NamedTuple):
 
 @dataclass
 class GetConnectedPeersResponse(BaseEvent):
-
+    """
+    The response class to answer a :class:`trinity.protocol.common.events.GetConnectedPeersRequest`
+    """
     peers: Tuple[PeerInfo, ...]
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


trinity/protocol/common/events.py
```diff
@@ -116,12 +121,17 @@ class PeerPoolMessageEvent(BaseEvent):
 
 @dataclass
 class ProtocolCapabilitiesResponse(BaseEvent):
-
+    """
+    The response class to answer a ``GetProtocolCapabilitiesRequest``
```
pipermerriam: Will these properly cross-link the classes or do you need to use `:class:` style syntax?

cburgdorf: These just highlight but don't cross link...but I can run the extra mile to do that ;)

cburgdorf: Updated! But you could also review it in one go with the (partly) implemented `eth/65` https://github.com/ethereum/trinity/pull/1627

cburgdorf:
Landed as part of #1627

PR #1634, opened 2020-03-24
Add validator APIs, using high-level `aiohttp`

ralexstokes:
The beacon node currently does not support the validator APIs.

We implement a version of them in this PR.

Additionally, we add some HTTP infrastructure to take advantage of the higher-level functionality in `aiohttp`. There was some earlier work using the lower-level primitives in `aiohttp` but that leads us to reimplementing a lot of logic that is already in the `aiohttp` library.

#### Cute Animal Picture

put a cute animal picture link inside the parentheses/adorable-white-pomeranian-puppy-spitz-921029690-5c8be25d46e0fb000172effe.jpg)

### DIFF
--- a/trinity/http/apps/base_handler.py
+++ b/trinity/http/apps/base_handler.py
@@ -1,0 +1,49 @@
+import functools
+import inspect
+from typing import Iterable
+
+from aiohttp import web
+from eth_utils import to_tuple
+
+from eth2.validator_client.beacon_node import BeaconNodePath as APIEndpoint
+
+WEB_APP_HANDLER_KEY = "__web_app_handler_data"
+
+
+# NOTE: ignoring the typing until this stabilizes a bit
+def _mk_method_decorator(action):  # type: ignore
+    def _method_func(endpoint: APIEndpoint):  # type: ignore
+        def _decorator(method):  # type: ignore
+            @functools.wraps(method)
+            def _inner(*args, **kwargs):  # type: ignore
+                return method(*args, **kwargs)
+
+            setattr(
+                _inner,
+                WEB_APP_HANDLER_KEY,
+                {"_action": action, "_path": endpoint.value},
+            )
+            return _inner
+
+        return _decorator
+
+    return _method_func
+
+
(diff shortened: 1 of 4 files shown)


trinity/http/apps/base_handler.py
```diff
@@ -0,0 +1,49 @@
+import functools
```
ralexstokes: everything in this file just exists to generate the mapping from route to handler according to the "decorator magic"; there is example of the decorator usage in the validator API file

trinity/http/apps/validator_api.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def __init__(
+        self, chain: BaseBeaconChain, event_bus: EndpointAPI, genesis_time: int
+    ):
+        self._chain = chain
+        self._event_bus = event_bus
+        self._genesis_time = genesis_time
+        self._client_identifier = construct_trinity_client_identifier()
+
+    @get(APIEndpoint.node_version)
```
ralexstokes: not super attached to this decorator usage but it is here for now...

the other candidate is to just define a function like

```python
def _mk_routes(handler):
    return (web.get(APIEndpoint.foo, handler._get_foo), ...)
```

ralexstokes: the thinking w/ the decorator approach is that it is easier to maintain this mapping if the definition is right  next to the handler definition

trinity/http/server.py
```diff
@@ -0,0 +1,46 @@
...
+        routes: Collection[web.RouteDef],
+        host: str = "127.0.0.1",
+        port: int = 8545,
+        token: CancelToken = None,
+        loop: asyncio.AbstractEventLoop = None,
+    ) -> None:
+        super().__init__(token=token, loop=loop)
+        self.host = host
+        self.port = port
+        app = web.Application()
```
ralexstokes: the `web.Application` is much higher-level than the `web.Server` approach in `trinity.http.main` and would deprecate a lot of code under the `trinity.http` package in lieu of using the (assumedly better maintained) code in `aiohttp`

PR #1638, opened 2020-03-25
Change default name of ENR directory

ralexstokes:
### What was wrong?

The default name of this directory is `nodes` which is not very specific and happens to be an overloaded symbol (full node, light node, etc.)

### How was it fixed?

Change the name.

#### Cute Animal Picture

### DIFF
--- a/trinity/constants.py
+++ b/trinity/constants.py
@@ -27,7 +27,7 @@ IPC_DIR = 'ipcs'
 LOG_DIR = 'logs'
 LOG_FILE = 'trinity.log'
 PID_DIR = 'pids'
-NODE_DB_DIR = 'nodes'
+NODE_DB_DIR = 'discovery-node-records'
 
 # sync modes
 SYNC_FULL = 'full'
--- a/trinity/components/eth2/discv5/component.py
+++ b/trinity/components/eth2/discv5/component.py
@@ -35,9 +35,8 @@ from p2p.identity_schemes import default_identity_scheme_registry
 from p2p.kademlia import KademliaRoutingTable
 from p2p.node_db import NodeDB
 from trinity.boot_info import BootInfo
+from trinity.constants import NODE_DB_DIR as DEFAULT_NODEDB_DIR_NAME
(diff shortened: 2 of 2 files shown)


trinity/constants.py
```diff
@@ -27,7 +27,7 @@
 LOG_DIR = 'logs'
 LOG_FILE = 'trinity.log'
 PID_DIR = 'pids'
-NODE_DB_DIR = 'nodes'
+NODE_DB_DIR = 'discv5-node-records'
```
gsalgado: This is also used for discv4. Both use the same implementation so could even share the same NodeDB dir if we wanted, although I'm not sure we do

ralexstokes: ah, good point!

i'd think we do actually want to use the same database...

the idea is to eventually have nodes of both types on the same DHT, if bc aiui you get better privacy/security out of the DHT the more nodes that are in it

gsalgado:
Actually, now that I think of it we probably don't want to share the same NodeDB between discv4 and discv5 because v4 will add _compat_ ENR entries (with a seq number 0 and empty sig) to the DB, for those nodes that don't support the ENR extension. And that might break discv5

ralexstokes:
interesting...  ok i'll change it back for now

ralexstokes:
this or we just have two classes, one that filters the `compat` ENRs -- what do you prefer @gsalgado ?

gsalgado:
I think for now we should keep them separate. But in case you want to experiment sharing the same, you'd have to change the TrinityConfig logic that combines `NODE_DB_DIR` with the app identifier to get the actual path for the DB (e.g. `~/.local/share/trinity/mainnet/nodes-eth1/`)

ralexstokes:
in which case we want the prefix to still be a generic identifier -- let me verify that the directory will be suffixed across eth1 vs eth2 (discv4 vs discv5) uses and i'll merge this in

ralexstokes:
yep it does, happy to update this further post-merge if we need to

PR #1639, opened 2020-03-25
Use the trinity `nodekey` for eth2

ralexstokes:
### What was wrong?

We were not piping the `nodekey` into eth2's components.

### How was it fixed?

Connect the bytes.

#### Cute Animal Picture

### DIFF
--- a/trinity/components/eth2/beacon/component.py
+++ b/trinity/components/eth2/beacon/component.py
@@ -13,7 +13,7 @@ from libp2p.crypto.secp256k1 import Secp256k1PrivateKey, create_new_key_pair
 from eth2.beacon.typing import SubnetId
 from p2p.service import BaseService, run_service
 from trinity.boot_info import BootInfo
-from trinity.config import BeaconAppConfig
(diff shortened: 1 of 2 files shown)


trinity/config.py
```diff
@@ -791,6 +791,16 @@ def get_chain_config(self) -> BeaconChainConfig:
         """
         return BeaconChainConfig.from_genesis_config()
 
+    @property
+    def discovery_private_key(self) -> PrivateKey:
+        return self.trinity_config.nodekey
+
+    @property
+    def node_private_key(self) -> PrivateKey:
```
pipermerriam: Since it doesn't appear you are using this property anywhere, maybe just remove it until it is actually needed?

ralexstokes: i could either delete it or add to this PR to use this property in the relevant location...

the main hesitation atm is whether to re-use the key pair or not...

ralexstokes: i'll elect to keep it simple and just use the one key pair in both places now :)

PR #1640, opened 2020-03-25
Raise error in `AsyncioIsolatedComponent`

ralexstokes:
### What was wrong?

Errors happening in `AsyncioIsolatedComponent`s are silently swallowed.

### How was it fixed?

Raise the error if there is one as reported by the process.

It is possible this should be part of the `asyncio-open-in-process` library -- i'll ping @pipermerriam  and @gsalgado to chime in further.

Also note: with the current multiprocess machinery, a user does not know about this error until shutting down the `trinity` platform (although they do eventually learn of it). Ideally individual components run with `run_component` can log the status of their run if they terminate early but it looks like this would introduce more substantial changes the `ComponentManager`

#### Cute Animal Picture

### DIFF
--- a/trinity/extensibility/asyncio.py
+++ b/trinity/extensibility/asyncio.py
@@ -33,7 +33,7 @@ class AsyncioIsolatedComponent(BaseIsolatedComponent):
         )
         async with proc_ctx as proc:
             try:
-                await proc.wait()
+                await proc.wait_result()
             except asyncio.CancelledError as err:
                 logger.debug('Component %s exiting. Sending SIGINT to pid=%d', self, proc.pid)
                 proc.send_signal(signal.SIGINT)


trinity/extensibility/asyncio.py
```diff
@@ -49,6 +49,8 @@ def get_subprocess_kwargs(self) -> Optional[SubprocessKwargs]:
                     proc.send_signal(signal.SIGTERM)
                 finally:
                     raise err
+        if proc.error:
```
pipermerriam: I'm not 100% sure but I think it is possible for this line to raise an `AttributeError`

https://github.com/ethereum/asyncio-run-in-process/blob/74f11319739c64c0fe8ff7c2aebee3ecf5a125e8/asyncio_run_in_process/process.py#L181-L189

It hints at a missing API which is the ability to check for an error.  Maybe something like `Process.did_error`?  or to just fix/adjust the `Process.error` property to return a `None` value...

ralexstokes: ah good catch -- i was just looking at the fn signature...

i agree about being able to check for errors...

in this case, i think we can get away with just `await .proc.wait_result()`, which is not very informative but should raise the exception if there is one...

trinity/extensibility/asyncio.py
```diff
@@ -33,7 +33,7 @@ def get_subprocess_kwargs(self) -> Optional[SubprocessKwargs]:
         )
         async with proc_ctx as proc:
             try:
-                await proc.wait()
+                await proc.wait_result()
```
ralexstokes: note to reviewers: `wait_result` will raise any exceptions in `proc`, rather than `wait` which did not.

PR #1647, opened 2020-03-26
Extract genesis config generation into reusable library function

njgheorghita:
#### Context
To facilitate genesis of a testnet, we define a self-contained configuration file that any software in the network can read and find the relevant information to begin their job at genesis.

There is a trinity `Application` defined under `trinity.components.eth2.network_generator` that generates this configuration and writes the data to disk. It would be nice to pull out the configuration generation code into a library function so that it can be used in other places.

The main use case is if the configuration on disk is missing for some reason:
```python
try:
    config = _read_genesis_config()
except FileNotFoundError:
    config = _generate_config_anyway()
finally:
    ...  # do stuff with config
```

### How was it fixed?
Extracted genesis config logic into `_generate_genesis_config()`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/components/eth2/network_generator/component.py
+++ b/trinity/components/eth2/network_generator/component.py
@@ -3,11 +3,12 @@ import json
 import logging
 import pathlib
 import time
-from typing import Optional
+from typing import Any, Dict, Optional
 
 from eth.constants import ZERO_HASH32
 from eth_utils import humanize_hash
 from ssz.tools.dump import to_formatted_dict
+from typing_extensions import Literal
 
 from eth2.beacon.genesis import initialize_beacon_state_from_eth1
 from eth2.beacon.state_machines.forks.serenity.configs import SERENITY_CONFIG
@@ -102,9 +103,10 @@ class NetworkGeneratorComponent(Application):
     def _generate_network_as_json(
         cls, args: Namespace, trinity_config: TrinityConfig
     ) -> None:
-        config = _get_eth2_config(args.config_profile)
-        override_lengths(config)
-        validator_count = config.MIN_GENESIS_ACTIVE_VALIDATOR_COUNT
+        eth2_config = _get_eth2_config(args.config_profile)
+        override_lengths(eth2_config)
+        validator_count = eth2_config.MIN_GENESIS_ACTIVE_VALIDATOR_COUNT
+
         output_file_path = args.output
 
         cls.logger.info(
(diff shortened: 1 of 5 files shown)


trinity/components/eth2/network_generator/component.py
```diff
@@ -102,9 +103,10 @@ def configure_parser(
     def _generate_network_as_json(
         cls, args: Namespace, trinity_config: TrinityConfig
     ) -> None:
-        config = _get_eth2_config(args.config_profile)
-        override_lengths(config)
-        validator_count = config.MIN_GENESIS_ACTIVE_VALIDATOR_COUNT
+        eth2_config = _get_eth2_config(args.config_profile)
+        override_lengths(eth2_config)
+        validator_count = eth2_config.MIN_GENESIS_ACTIVE_VALIDATOR_COUNT
```
njgheorghita: this part is now kinda redundant, but it made sense to me to leave it in order to preserve the logging messages

trinity/main_validator.py
```diff
@@ -55,9 +56,13 @@ def main_validator() -> None:
...
 
-    genesis_config = _load_genesis_config_at(
-        validator_client_app_config.genesis_config_path
-    )
+    try:
+        genesis_config = _load_genesis_config_at(
+            validator_client_app_config.genesis_config_path
+        )
+    except FileNotFoundError:
+        genesis_config = _generate_genesis_config("minimal")
```
njgheorghita: I was unsure if this is a fair example of the use case you were thinking of or not - and if so, whether it should default to `"minimal"` or `"mainnet"`

trinity/components/eth2/network_generator/component.py
```diff
@@ -113,37 +115,51 @@ def _generate_network_as_json(
...
+            genesis_config["genesis_state"].genesis_time,
         )
         output_file_path.parent.mkdir(parents=True, exist_ok=True)
         with open(output_file_path, "w") as output_file:
-            output_file.write(json.dumps(output))
+            output_file.write(json.dumps(genesis_config))
+
+
+def _generate_genesis_config(
+    config_profile: Literal["minimal", "mainnet"],
```
ralexstokes: didn't know `mypy` has dependent-typing!

trinity/components/eth2/network_generator/component.py
```diff
@@ -113,37 +115,51 @@ def _generate_network_as_json(
...
+    genesis_time: Timestamp = None,
+    genesis_delay: int = None,
```
ralexstokes: i guess we don't have to type these as `Optional[T]`?

trinity/components/eth2/network_generator/component.py
```diff
@@ -113,37 +115,51 @@ def _generate_network_as_json(
...
-            eth1_block_hash=eth1_block_hash,
-            eth1_timestamp=Timestamp(eth1_timestamp),
-            deposits=deposits,
-            config=config,
-        )
-        genesis_state = _adjust_genesis_time(
-            initial_state, args.genesis_time, args.genesis_delay
+
+        genesis_config = _generate_genesis_config(
+            args.config_profile, args.genesis_time, args.genesis_delay
```
ralexstokes: i would actually suggest simplifying this as the "real" inputs here are just the `genesis_time`.

this utility lets the user specify a time either directly as `genesis_time` or as a relative offset to "now" as `genesis_delay`.

so possibly:

```python
genesis_time = args.genesis_time if args.genesis_time or int(time.time())+args.genesis_delay
c = _generate_genesis_config(profile, genesis_time)
# NOTE: unconditionally set `genesis_time` on the `genesis_state` _inside_ `_generate_genesis_config`
```

then we could just remove `_adjust_genesis_time`

trinity/components/eth2/network_generator/component.py
ralexstokes: see above comment about separating the concerns here

trinity/main_validator.py
```diff
@@ -16,6 +16,7 @@
 from eth2.validator_client.tools.directory import create_dir_if_missing
 from trinity._utils.logging import LOG_FORMATTER
 from trinity.bootstrap import load_trinity_config_from_parser_args
+from trinity.components.eth2.network_generator.component import _generate_genesis_config
```
ralexstokes: don't import private functions!

trinity/main_validator.py
ralexstokes: yes exactly!

spectacular

"minimal" is fine for now... theoretically this is a command line flag the user could pass but defaulting to `minimal` is preferred and good enough for now

ralexstokes: btw we would want to do the same thing in `trinity.config` where we load this in the `BeaconChainConfig`

trinity/components/eth2/network_generator/component.py
ralexstokes: i think we get rid of it and remove the first logging message (possibly merging the statements if you think the information is important) -- a big thing to know are the genesis_state hash_tree_root and the genesis_time -- i guess it's helpful to get the output path as well just so you know where this thing just destructively changed your computer

this isn't obvious but we kind of only want to call `override_lengths` once as it causes a global change to the entire interpreter state -- no problem calling it multiple times but it is the kind of thing we want to call once as soon as we can and then leave it alone

trinity/components/eth2/network_generator/component.py
```diff
@@ -113,37 +115,51 @@ def _generate_network_as_json(
...
+            humanize_hash(genesis_config["genesis_state"].hash_tree_root),
+            genesis_config["genesis_state"].genesis_time,
         )
         output_file_path.parent.mkdir(parents=True, exist_ok=True)
         with open(output_file_path, "w") as output_file:
-            output_file.write(json.dumps(output))
+            output_file.write(json.dumps(genesis_config))
+
+
+def _generate_genesis_config(
```
ralexstokes: this is stylistic but i prefer to only have code in "application" level modules that directly pertains to the module -- this means `_generate_genesis_config` would go into the "library code", i'd imagine under `eth2` namespace

njgheorghita: sounds good - I moved it to the `eth2.configs` namespace... I'm not sure if this is the best place for it or not? or I can import it into the top-level `eth2` module if that'd be a better way to access it

trinity/components/eth2/network_generator/component.py
njgheorghita: nice! what's the desired behavior if neither `genesis_time` or `genesis_delay` are provided? right now it just sets `genesis_time` to `None` before passing it off to `generate_genesis_config`

njgheorghita:
@ralexstokes let me know if there's anything else that needs changing. I also don't have merge access in `trinity` so if you want to merge this - or I can ping whoever's an admin (piper?) to get merge access.

trinity/components/eth2/network_generator/component.py
ralexstokes: `genesis_delay` should default to `0` -- passing `None` may be just fine -- i'll go look

ralexstokes: in the PR it would end up as `0` which is an equally fine default value for now

trinity/components/eth2/network_generator/component.py
ralexstokes: `eth2.genesis` seems fine to me -- to me it is more of a utility than things core to the `eth2` protocol but lets' not get picky right now ;)

ralexstokes:
@njgheorghita sorry for the delay on this and thanks a bunch!!

PR #1648, opened 2020-03-27
Monitor peer connections and disconnect if idle for too long

gsalgado:
After CONN_IDLE_TIMEOUT/2 seconds without receiving any messages on a
connection, we send a ping. If CONN_IDLE_TIMEOUT/2 seconds pass and we
still haven't received any messages, cancel the connection.

### DIFF
--- a/p2p/connection.py
+++ b/p2p/connection.py
@@ -57,6 +57,7 @@ class Connection(ConnectionAPI, BaseService):
         Type[ProtocolAPI],
         Set[HandlerFn]
     ]
+    _msg_handlers: Set[HandlerFn]
     _command_handlers: DefaultDict[
         Type[CommandAPI[Any]],
         Set[HandlerFn]
@@ -76,6 +77,7 @@ class Connection(ConnectionAPI, BaseService):
 
         self._protocol_handlers = collections.defaultdict(set)
         self._command_handlers = collections.defaultdict(set)
+        self._msg_handlers = set()
 
         # An event that controls when the connection will start reading from
         # the individual multiplexed protocol streams and feeding handlers.
@@ -182,6 +184,7 @@ class Connection(ConnectionAPI, BaseService):
                 )
                 self.run_task(proto_handler_fn(self, cmd))
             command_handlers = set(self._command_handlers[type(cmd)])
+            command_handlers.update(self._msg_handlers)
             for cmd_handler_fn in command_handlers:
                 self.logger.debug2(
                     'Running command handler %s for protocol=%s command=%s',
@@ -205,6 +208,11 @@ class Connection(ConnectionAPI, BaseService):
             self._protocol_handlers[protocol_class].remove,
             handler_fn,
         )
+        return Subscription(cancel_fn)
+
+    def add_msg_handler(self, handler_fn: HandlerFn) -> SubscriptionAPI:
+        self._msg_handlers.add(handler_fn)
+        cancel_fn = functools.partial(self._msg_handlers.remove, handler_fn)
         return Subscription(cancel_fn)
 
     def add_command_handler(self,
--- a/p2p/p2p_api.py
+++ b/p2p/p2p_api.py
@@ -1,8 +1,21 @@
+import asyncio
+from typing import Any, AsyncIterator, cast
+
+from async_generator import asynccontextmanager
+
 from cached_property import cached_property
 
-from p2p.abc import ConnectionAPI
-from p2p.logic import Application, CommandHandler
+from cancel_token import OperationCancelled
+
+from async_service import (
+    background_asyncio_service,
+    Service,
+)
+
+from p2p.abc import CommandAPI, ConnectionAPI, HandlerFn
+from p2p import constants
 from p2p.disconnect import DisconnectReason
+from p2p.logic import Application, BaseLogic, CommandHandler
 from p2p.p2p_proto import Disconnect, Ping, Pong
 from p2p.qualifiers import always
 from p2p._utils import get_logger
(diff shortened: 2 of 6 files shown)


p2p/p2p_api.py
```diff
@@ -15,9 +29,60 @@ class PongWhenPinged(CommandHandler[Ping]):
...
+    while connection.is_operational:
+        try:
+            await connection.wait_first(msg_received.wait(), asyncio.sleep(half_timeout))
+        except OperationCancelled:
+            return
+        if msg_received.is_set():
+            connection.logger.debug2("Received msg on %s, restarting idle monitor", connection)
+            msg_received.clear()
+            continue
+        _send_ping(connection)
```
pipermerriam: This is *ok* but did you happen to look into passing a reference to the `P2PAPI` down into here so that you could just call `P2PAPI.send_ping`?  If it's overly complex to do so not a big deal to keep it this way.

tests/p2p/test_peer.py
```diff
@@ -41,3 +44,31 @@
...
+
+
+@pytest.mark.asyncio
+async def test_sends_ping_when_conn_idle(monkeypatch):
+    got_ping = asyncio.Event()
+
+    async def _handle_ping(conn, cmd):
+        got_ping.set()
+
+    monkeypatch.setattr(constants, 'CONN_IDLE_TIMEOUT', 0.1)
```
pipermerriam: Suggestion for an alternate way to test this.  The `DisconnectIfIdle` logic could take this timeout value as an argument.

Then, instead of testing this at the `P2PAPI` level, you could instead test the `DisconnectIfIdle` directly.  Rather than using `ParagonPeerPairFactory` you can just use raw `ConnectionPairFactory` and manually attach the behavior to the connections and then test that the connections behave as expected. This would allow you to reduce the timeout without monkey patching and to directly test this behavior in isolation.

gsalgado: I like the idea of testing this directly with a connection pair, but I think it'd make more sense to still have the `P2PAPI` applied to the connections, as that ensures any connection with that applied will get the disconnect-if-idle behavior we want, whereas if I test just `DisconnectIfIdle` directly, we wouldn't ensure all connections on which `P2PAPI` is applied behave that way. In order to do that I could either add the extra argument to both `P2PAPI`, `DisconnectIfIdle` and `PingAndDisconnectIfIdle`, or turn `PingAndDisconnectIfIdle` into a class with a `get_idle_timeout()` method, and then I can monkeypatch that method in the test? If you feel strong against monkeypatching I can add the extra args

pipermerriam: @gsalgado all I'm suggesting is testing this in isolation.  I think the `P2PAPI` should still be applied to all connections during trinity runs.

gsalgado: Right, I understood that, but if we test the `DisconnectIfIdle` behavior directly, we'd have no guarantees that `P2PAPI` correctly applies that behavior, so if in a future change/refactoring `P2PAPI` stops applying that behavior for any reason connections would no longer be disconnected when idle and we'd have no tests to catch the regression. It's actually the common dilemma of choosing between writing unit or integration (in a loose sense) tests for a feature

pipermerriam: Sorry, I misunderstood.  I'd propose testing in isolation and a minimal integration test at the `P2PAPI` level.

p2p/connection.py
```diff
@@ -207,6 +210,11 @@ def add_protocol_handler(self,
         )
         return Subscription(cancel_fn)
 
+    def add_msg_handler(self, handler_fn: HandlerFn) -> SubscriptionAPI:
+        self._command_handlers[_AnyCommand].add(handler_fn)
```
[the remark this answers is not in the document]
gsalgado: Yeah, that makes sense!

p2p/p2p_api.py
```diff
@@ -15,9 +29,60 @@ class PongWhenPinged(CommandHandler[Ping]):
...
 
 
+@as_service
+async def PingAndDisconnectIfIdle(manager: ManagerAPI, connection: ConnectionAPI) -> None:
+    msg_received = asyncio.Event()
+
+    async def set_msg_received(connection: ConnectionAPI, cmd: CommandAPI[Any]) -> None:
+        msg_received.set()
+
+    connection.add_msg_handler(cast(HandlerFn, set_msg_received))
```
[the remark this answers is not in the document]
gsalgado: :+1:

p2p/p2p_api.py
gsalgado: That is actually how I implemented it initially but it annoyed me that we had to store a `P2PAPI` instance there just so that we could use one of its methods that is nothing more than a thin wrapper around a connection method.

gsalgado:
@pipermerriam last commit has the changes you suggested and a couple new tests

p2p/p2p_api.py
```diff
@@ -33,38 +32,44 @@ class PongWhenPinged(CommandHandler[Ping]):
...
+        with conn.add_msg_handler(cast(HandlerFn, set_msg_received)):
+            while conn.is_operational:
+                try:
+                    await conn.wait_first(msg_received.wait(), asyncio.sleep(half_timeout))
+                except OperationCancelled:
+                    return
+                if msg_received.is_set():
+                    conn.logger.debug2("Received msg on %s, restarting idle monitor", conn)
+                    msg_received.clear()
+                    continue
```
pipermerriam: Just noting that the only difference between this and lines 60-67 is the timeout parameter.  Maybe drop it into a small re-usable function to make the overally logic of this loop easier to read?

gsalgado: This is what I came up with, but we still need to catch the OperationCancelled outside and handle the functions return value to decide whether or not to continue, so in the end I don't think it's more readable. What do you think?

```
    async def wait_msg(self) -> bool:
        msg_received = asyncio.Event()

        async def set_msg_received(connection: ConnectionAPI, cmd: CommandAPI[Any]) -> None:
            msg_received.set()

        conn = self.connection
        half_timeout = self.idle_timeout / 2
        await conn.wait_first(msg_received.wait(), asyncio.sleep(half_timeout))
        if msg_received.is_set():
            conn.logger.debug2("Received msg on %s, restarting idle monitor", conn)
            return True
        return False

    async def run(self) -> None:
        msg_received = asyncio.Event()

        async def set_msg_received(connection: ConnectionAPI, cmd: CommandAPI[Any]) -> None:
            msg_received.set()

        conn = self.connection
        with conn.add_msg_handler(cast(HandlerFn, set_msg_received)):
            while conn.is_operational:
                try:
                    if await self.wait_msg():
                        continue
                except OperationCancelled:
                    return

                _send_ping(conn)

                try:
                    if await self.wait_msg():
                        continue
                except OperationCancelled:
                    return

                conn.logger.info(
                    "Reached idle limit (%.2f) on %s, disconnecting", self.idle_timeout, conn)
                conn.cancel_nowait()
                return
```

pipermerriam: Yeah, that's gross.  Carry on...

PR #1649, opened 2020-03-27
Fix broken Transaction propagation

cburgdorf:
### What was wrong?

Turns out Trinity did not propagate transactions to peers since we moved the tx pool into an isolated plugin a loooong time ago. 

#### Wait, I'm sure we had tests for that, no?

Kind of, unfortunately, the tests we had only ensured that the transaction pool would send out the correct transactions to the correct peers. In fact there was nothing wrong with the transaction pool directly.

Because of Trinitys multiprocess nature peer interactions from outside of the peer pool are a bit more involved. When the transaction pool wants to send out transactions to a peer, it doesn't act on the peer directly but rather dispatches a message on the event bus which is then later "replayed" on the actual peer in the peer pool. Turns out, that part was broken, there was no code to listen for the `SendTransactionsEvent` that the transaction pool was sending. They ended in the void.

### How was it fixed?

1. The `TransactionSendEvent` is now picked up and the transactions are forwarded to the peer

2. The tests were rewritten to use two actual connected transaction pools and do real assertions about what transaction pool B receives from transaction pool a and vice versa.

3. Bonus, I added a little refactoring to delete much of the code that handles these simple `SendXEvent`'s which hopefully will reduce the chance of such error in the future.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/tx-pool/test_tx_pool.py
+++ b/tests/core/tx-pool/test_tx_pool.py
@@ -8,35 +8,41 @@ from eth._utils.address import (
     force_bytes_to_address
 )
 
-from p2p.tools.factories import SessionFactory
 from p2p.service import run_service
 
-from trinity.constants import TO_NETWORKING_BROADCAST_CONFIG
 from trinity.components.builtin.tx_pool.pool import (
     TxPool,
 )
 from trinity.components.builtin.tx_pool.validators import (
     DefaultTransactionValidator
 )
-from trinity.protocol.common.events import (
-    GetConnectedPeersRequest,
-)
-from trinity.protocol.eth.commands import (
-    Transactions
-)
+from trinity.constants import TO_NETWORKING_BROADCAST_CONFIG
 from trinity.protocol.eth.events import (
     TransactionsEvent,
-    SendTransactionsEvent,
 )
-from trinity.tools.event_bus import mock_request_response
-from trinity.tools.factories.events import GetConnectedPeersResponseFactory
-
 from trinity.protocol.eth.peer import (
     ETHProxyPeerPool,
+    ETHPeerPoolEventServer
 )
+from trinity.tools.factories import ETHPeerPairFactory, ChainContextFactory
+
+from tests.core.integration_test_helpers import run_peer_pool_event_server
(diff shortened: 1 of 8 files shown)


pipermerriam:
This suggests we should prioritize the `lahja` updates we've previously talked about which throw an error in the case that you're sending an event *into the void*.  Basically, adjust things so that any event that would go nowhere raises an exception unless maybe something specific is set on the broadcast config.

tests/core/tx-pool/test_tx_pool.py
```diff
@@ -8,182 +8,192 @@
...
+@pytest.mark.asyncio
+async def test_does_not_propagate_invalid_tx(two_connected_tx_pools,
+                                             funded_address_private_key,
+                                             chain_with_block_validation):
+    (
+        (peer1, peer1_event_bus, peer1_tx_pool),
+        (peer2, peer2_event_bus, peer2_tx_pool)
+    ) = two_connected_tx_pools
+
+    peer1_incoming_tx, peer1_got_tx = observe_incoming_transactions(peer1_event_bus)
```
pipermerriam: I know this is just a carry over of the previous naming but it would probably improve readability a lot to change these from `peer1/peer2` to `alice/bob`.

cburgdorf: Yep, good call, I replaced all of these in this file with `alice` and `bob`.

cburgdorf:
>This suggests we should prioritize the `lahja` updates we've previously talked about which throw an error in the case that you're sending an event into the void. Basically, adjust things so that any event that would go nowhere raises an exception unless maybe something specific is set on the broadcast config.

Good point. In general, `lahja` already provides the building blocks with the `wait_until_any_endpoint_subscribed_to` API. But it's opt-in to use that. We could bake it into the `request` and `broadcast` and allow to side step. There are certainly valid use cases where you would not want that check e.g.

- events that are raised for extensibility without any existing use case in core
- events that only have subscribers based on which features are activated (e.g. think `--disable-tx-pool`
- ...

I'm inclined to send a follow up PR to fix it on the Trinity side with the existing APIs to see how that would look and feel first.

PR #1653, opened 2020-03-31
Upgrade to lahja 0.16.0

cburgdorf:
### What was wrong?

It is easy to end up in a scenario where important events are forgotten to get subscribed to. This can lead to fatal bugs such as [Trinity not propagating any transactions to its peers](https://github.com/ethereum/trinity/pull/1649)

### How was it fixed?

1. The latest lahja release [raises `NoSubscriber` exceptions](https://github.com/ethereum/lahja/pull/176) by default whenever events or requests are send into the void.

2. Events/Requests can still be fired if there are no subscribers but it needs to be made explicit via `BroadcastConfig(require_subscriber=False)`. 

This PR updates to lahja `0.16.0` and adjusts some tests and sets `require_subscriber=False` where ever it is appropriate.

### To-Do

- [x] Clean up commit history

[//]: # (For important changes that should go into the release notes please add a newsfragment file as explained here: https://github.com/ethereum/lahja/blob/master/newsfragments/README.md)

[//]: # (See: https://lahja.readthedocs.io/en/latest/contributing.html#pull-requests)
- [x] Add entry to the [release notes](https://github.com/ethereum/lahja/blob/master/newsfragments/README.md)

- [x] Change `setup.py` when the Lahja release is cut

#### Cute Animal Picture

### DIFF
--- a/trinity/constants.py
+++ b/trinity/constants.py
@@ -40,6 +40,8 @@ NETWORKDB_EVENTBUS_ENDPOINT = 'network-db'
 NETWORKING_EVENTBUS_ENDPOINT = 'networking'
 UPNP_EVENTBUS_ENDPOINT = 'upnp'
 TO_NETWORKING_BROADCAST_CONFIG = BroadcastConfig(filter_endpoint=NETWORKING_EVENTBUS_ENDPOINT)
+FIRE_AND_FORGET_BROADCASTING = BroadcastConfig(require_subscriber=False)
+
 
 # Network IDs: https://ethereum.stackexchange.com/questions/17051/how-to-select-a-network-id-or-is-there-a-list-of-network-ids/17101#17101  # noqa: E501
 MAINNET_NETWORK_ID = 1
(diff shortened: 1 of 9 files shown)


trinity/constants.py
```diff
@@ -40,6 +40,8 @@
 NETWORKING_EVENTBUS_ENDPOINT = 'networking'
 UPNP_EVENTBUS_ENDPOINT = 'upnp'
 TO_NETWORKING_BROADCAST_CONFIG = BroadcastConfig(filter_endpoint=NETWORKING_EVENTBUS_ENDPOINT)
+FIRE_AND_FORGET_BROADCASTING = BroadcastConfig(require_subscriber=False)
```
pipermerriam, line 43: This looks like something that should be included in `lahja` for easy convenient import.  Mind making an issue.  This would be a good "Good First Issue" to leave for a 3rd party contributor.

cburgdorf: Good idea: https://github.com/ethereum/lahja/issues/178

PR #1656, opened 2020-04-02
Use a nursery for short-lived txpool tasks

pipermerriam:
### What was wrong?

The `async-service` API for `Manager.run_task` involves a decent bit of extra machinery for task tracking and clean shutdown.  For functions that need to run lots of short lived tasks, there should be some performance gain from using a bare `trio.Nursery` to do the job.

### How was it fixed?

Used a nursery instead of `run_task` inside the transaction pool.

#### Cute Animal Picture

9dd6a26fc920381f1cb60587e66fe072

### DIFF
--- a/trinity/components/builtin/tx_pool/pool.py
+++ b/trinity/components/builtin/tx_pool/pool.py
@@ -106,8 +106,9 @@ class TxPool(Service):
         # Process GetPooledTransactions requests
         self.manager.run_daemon_task(self._process_get_pooled_transactions_requests)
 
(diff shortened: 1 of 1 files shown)


cburgdorf:
>The `async-service` API for `Manager.run_task` involves a decent bit of extra machinery for task tracking and clean shutdown.  For functions that need to run lots of short lived tasks, there should be some performance gain from using a bare `trio.Nursery` to do the job.

Good to know! However, the transaction pool isn't running on Trio yet (but is probably easy to migrate over)

PR #1658, opened 2020-04-03
Fix broken eth/64 support

cburgdorf:
### What was wrong?

Turns out as I added support for `eth/65` recently, I accidentally broke support for `eth/64`. 

### How was it fixed?

- Added `ETHProtocolV64` to `supported_sub_protocols` on `ETHPeer`
- Added `ETHV64API` to `get_behaviors`
- Created `ETHV64PeerPairFactory` and added it to the parameters of `test_eth_api` so that we have test coverage. 

### To-Do

- [x] Clean up commit history

- [x]  No need for release notes as the bug was never released

#### Cute Animal Picture

### DIFF
--- a/tests/core/p2p-proto/test_eth_api.py
+++ b/tests/core/p2p-proto/test_eth_api.py
@@ -25,9 +25,9 @@ from trinity.protocol.eth.commands import (
 )
 from trinity.protocol.eth.handshaker import ETHHandshakeReceipt, ETHV63HandshakeReceipt
 from trinity.protocol.eth.proto import (
-    ETHProtocol,
     ETHProtocolV63,
     ETHProtocolV64,
+    ETHProtocolV65,
 )
 
 from trinity.tools.factories.common import (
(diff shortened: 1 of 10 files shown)


tests/core/p2p-proto/test_eth_api.py
```diff
@@ -24,7 +24,11 @@
     StatusV63,
 )
 from trinity.protocol.eth.handshaker import ETHHandshakeReceipt, ETHV63HandshakeReceipt
-from trinity.protocol.eth.proto import ETHProtocolV63, ETHProtocol
+from trinity.protocol.eth.proto import (
+    ETHProtocol,
```
pipermerriam, line 28: I would advocate against having an unversioned one of these and instead just make all of them versioned.  You can still do something like `from ... import EthProtocolV65 as EthProtocol` if you want to cut down on some of the renaming overhead...

cburgdorf: Yeah, after implementing a few of these different protocol versions I have to agree that maintaining an unversioned variant that constantly implicitly moves the version it represents is a source of confusion and error. I have renamed this and also the `ETHAPI` to contain versions.

I also noticed a slight inconsistency with the naming.

We have:

- `ETHV63HandshakeReceipt`
- `ETHV63PeerFactory`
- `ETHV63API`

But also 

- `ETHProtocolV63`
- `ETHProtocolV64`
- ...

- We could be consistent and either rename:

- `ETHProtocolV63` and friends to `ETHV63Protocol`

or rename

- `ETHV63HandshakeReceipt`, `ETHV63PeerFactory` and friends to `ETHHandshakeReceiptV63`, `ETHPeerFactoryV63`

In general I think keeping `ETHVXX` together feels more natural to me compared to `ETHPeerFactoryV63` except for the `ETHProtocol` itself where I somehow think it feels more natural to have the `V63` as a suffix. Maybe because in my mind I read `ETHProtocol` as a single word.

So, long story short, I just wanted to point out that the current naming inconsistency is on purpose but I'm open to adjust if you feel differently about the naming.

pipermerriam:
Regarding the naming inconsistency, if you want to clean that up, I'd suggest doing it as a follow up.  I don't feel strongly about fixing the inconsistency.

PR #1663, opened 2020-04-09
AsyncioIsolatedComponent no longer handles CancelledErrors

gsalgado:
Those are handled in open_in_process(); there should be no need
to duplicate that behaviour here

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ deps = {
     ],
     'trinity': [
         "aiohttp==3.6.0",
-        "asyncio-run-in-process==0.1.0a4",
+        "asyncio-run-in-process==0.1.0a5",
         "bloom-filter==1.3",
         "cachetools>=3.1.0,<4.0.0",
         "coincurve>=10.0.0,<11.0.0",
--- a/trinity/extensibility/asyncio.py
+++ b/trinity/extensibility/asyncio.py
@@ -1,6 +1,4 @@
 from abc import abstractmethod
-import asyncio
-import signal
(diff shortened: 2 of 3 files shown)


gsalgado:
This test failure shows the `CancelledError` I mention in https://github.com/ethereum/asyncio-run-in-process/pull/14: https://circleci.com/gh/ethereum/trinity/210502

pipermerriam:
I assume this is now waiting for an `asyncio-run-in-process` release?

gsalgado:
Yes, but it depends on a release that includes something like https://github.com/ethereum/asyncio-run-in-process/pull/14

gsalgado:
The fact that we have both `AsyncioIsolatedComponent` and `asyncio-run-in-process` sending SIGINTs/SIGTERMs to terminate children makes debugging the whole shutdown mess a lot more complicate than it already is. But this change requires an ugly hack to `asyncio-run-in-process` just to please one of our tests, so maybe as an interim solution we can disable that test until we have a better shutdown story and then, if it's still needed, we cut a new `asyncio-run-in-process` release with the ugly hack and re-enable the test. @pipermerriam are you ok with that?

gsalgado:
Actually, the test passes even without the hack, but if you run it by itself, pytest hangs forever after reporting success, similar to our shutdown. Checking the logs emitted by the test one can see the component's `do_run()` method never returns, possibly because it hangs forever when trying to exit the `background_asyncio_service()` context, just like what happens during our shutdown

PR #1666, opened 2020-04-11
Every component's process is now its own process group leader

gsalgado:
IOW, for every process we start, we create a new process group.
This way a Ctrl-C in the terminal won't cause a SIGINT to be sent
to each of our processes, instead a single one will be sent to the
parent process, which will then send SIGINTs to every child (via
asyncio-run-in-process' open_in_process()).

Turns out our integration [tests rely on all processes belonging to a single process group so that they can be correctly terminated](https://github.com/ethereum/trinity/blob/5937306437735be2bcea5c6e054101299a994a24/trinity/tools/async_process_runner.py#L69-L73) between tests, so I had to add this new env var to control whether or not to use separate process groups.

### DIFF
--- a/trinity/extensibility/asyncio.py
+++ b/trinity/extensibility/asyncio.py
@@ -1,4 +1,5 @@
 from abc import abstractmethod
+import os
 from typing import Optional
 
 from asyncio_run_in_process import open_in_process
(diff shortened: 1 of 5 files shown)


trinity/extensibility/asyncio.py
```diff
@@ -23,13 +26,23 @@
 class AsyncioIsolatedComponent(BaseIsolatedComponent):
     def get_subprocess_kwargs(self) -> Optional[SubprocessKwargs]:
         # Note that this method currently only exist to facilitate testing.
-        return None
+        return {}
 
     async def run(self) -> None:
+        start_new_session = True
+        if os.getenv('TRINITY_SINGLE_PROCESS_GROUP') == "1":
```
pipermerriam, line 33: Maybe rather than putting this logic in the `run` method here it could just live inside the` get_subprocess_kwargs` method?

gsalgado: Makes sense! I'd like to think a bit more about the implications about making this behaviour the default in asyncio-run-in-process, though, so I've filed an issue for that

PR #1670, opened 2020-04-14
Convert BaseServer into a new-style Service

gsalgado:


tests/core/p2p-proto/test_server.py
```diff
@@ -87,15 +90,16 @@ def receiver_remote():
...
+    async with background_asyncio_service(server):
+        try:
+            yield server
+        finally:
+            # XXX: Need to manually cancel the token here but in production the server's caller
+            # would do that.
+            server.cancel_token.trigger()
+    # Now that the server is finished, sleep a bit to give old-style services a chance to detect
+    # the token was triggered and finish their cleanup.
+    await asyncio.sleep(0.1)
```
gsalgado, line 102: @cburgdorf I've pulled those changes from https://github.com/ethereum/trinity/pull/1671 after your review, as they actually belong here

PR #1673, opened 2020-04-15
More careful handling of UnknownAPI and PeerConnectionLost

carver:
### What was wrong?

Uncaught `UnknownAPI` exception
problem trace

### How was it fixed?

In progress. Main theme:
- Switch from `UnknownAPI` to `PeerConnectionLost` (which feels more appropriate for a peer that had the API and then dropped it during disconnection).
- Make sure all `head_info` usages are prepared to handle the connection lost exception

Note that the whole class `PeerHeaderSyncer` is unused, so I removed it to make maintenance easier. (It includes several `head_info` usages which needed more explicit handling). The last reference to it that I could find was removed in 4c87a663a0b426d -- about 1.5 years ago.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


p2p/connection.py
```diff
@@ -247,16 +249,21 @@ def add_logic(self, name: str, logic: LogicAPI) -> SubscriptionAPI:
...
             raise PeerConnectionLost("Cannot look up subprotocol when connection is closing")
-        return name in self._logics
+        else:
+            return name in self._logics
 
     def get_logic(self, name: str, logic_type: Type[TLogic]) -> TLogic:
         if not self.has_logic(name):
-            raise UnknownAPI(f"No API registered for the name '{name}'")
+            if name in self._disabled_logics:
+                raise PeerConnectionLost(f"Subprotocol {name} has been disabled on the peer")
```
carver, line 264: Without this change, callers get this confusing behavior of sorting peers based on available API and then later getting an `UnknownAPI` exception, even though it was previously verified.

trinity/sync/common/headers.py
```diff
@@ -648,8 +648,6 @@ def fail_task() -> None:
             self._filler_header_tasks.complete(batch_id, tuple())
 
         peer = await self._waiting_peers.get_fastest()
-        if not self.sync_progress:
-            await self._init_sync_progress(parent_header, peer)
```
carver, line 652: This feels like an awkward place to check if the peer connection is lost, because no other direct communication with the peer happens in this method. That's why I moved it below to where other messages to the peer happen.

trinity/sync/common/headers.py
```diff
@@ -891,12 +898,26 @@ async def new_sync_headers(
                 self._buffer_capacity.set()
 
     def get_target_header_hash(self) -> Hash32:
-        if not self._is_syncing_skeleton and self._last_target_header_hash is None:
-            raise ValidationError("Cannot check the target hash before the first sync has started")
-        elif self._is_syncing_skeleton:
-            return self._skeleton.peer.head_info.head_hash
```
carver, line 897: Nothing really changed about this logic except for handling the `PeerConnectionLost` and returning the last saved value, if possible.

trinity/sync/light/service.py
```diff
@@ -383,6 +384,10 @@ async def _retry_on_bad_response(self, make_request_to_peer: Callable[[LESPeer],
                 self.logger.warning("Disconnecting from peer, because: %s", exc)
                 await peer.disconnect(DisconnectReason.SUBPROTOCOL_ERROR)
                 # reattempt after removing this peer from our pool
+            except PeerConnectionLost:
+                self.logger.debug("%s disconnected during LES request, moving on...", peer)
+                # Let the event loop cycle, then try again
+                await self.sleep(0)
```
carver, line 390: If we don't sleep here, then highest_td_peer may not get a chance to refresh? I seem to remember problems like this with the checkpoint syncer. I guess I can write that in a comment in-code.

carver:
integration run `EOFError` again :(

p2p/connection.py
```diff
@@ -247,16 +249,21 @@ def add_logic(self, name: str, logic: LogicAPI) -> SubscriptionAPI:
         return Subscription(cancel_fn)
 
     def remove_logic(self, name: str) -> None:
+        self._disabled_logics.add(name)
```
gsalgado, line 252: At first I thought `_disabled_logics` would contain all logics not enabled, but in fact it's the ones that were once enabled but for some reason were removed, right? Maybe rename it to make that clear?

p2p/connection.py
gsalgado: Should we also `cancel()` the connection in this case or do we want to keep it active as other callers may not care about this logic and so the peer could still be useful?

gsalgado: Still, it doesn't feel right to raise a PeerConnectionLost here. If this was only supposed to happen when the peer is disconnecting, the `has_logic()` call on line 262 would raise a `PeerConnectionLost` and we wouldn't need this, right? That suggests to me we're getting this while the peer is still running

trinity/protocol/common/peer.py
```diff
@@ -242,12 +245,24 @@ def highest_td_peer(self) -> BaseChainPeer:
...
+    def get_peers(self, min_td: int) -> Iterator[BaseChainPeer]:
         # TODO: Consider turning this into a method that returns an AsyncIterator, to make it
         # harder for callsites to get a list of peers while making blocking calls, as those peers
         # might disconnect in the meantime.
         peers = tuple(self.connected_nodes.values())
-        return [peer for peer in peers if peer.head_info.head_td >= min_td]
+        for peer in peers:
+            try:
+                peer_td = peer.head_info.head_td
+            except PeerConnectionLost as err:
```
gsalgado, line 257: Again, it feels weird to expect an exception from a simple attribute access, but that kind of suggests a deeper issue (too many obscure abstractions, maybe?) than what you're trying to fix here. Just thought I'd mention.

gsalgado: Also, this is all synchronous code, just iterating over the connected peers, so they couldn't have disconnected since we copied the list in line 253

trinity/sync/light/service.py
```diff
@@ -383,6 +384,13 @@ async def _retry_on_bad_response(self, make_request_to_peer: Callable[[LESPeer],
...
+                # This is critical, because `highest_td_peer` is not awaitable, so it can
+                #   get caught in an infinite loop of getting the same "lost connection"
+                #   peer without the pool getting a chance to update.
```
gsalgado, line 393: The sleep is critical, you mean? Maybe move the comment above it, then?

p2p/connection.py
carver: > If this was only supposed to happen when the peer is disconnecting, the `has_logic()` call on line 262 would raise a `PeerConnectionLost` and we wouldn't need this, right?

Yeah, frankly this is a patching approach because I don't understand all the intricacies of how the connections work. As best I can tell, there is some kind of state where the logic has been removed but `is_closing` has not been set, yet.

If there was some way to make sure that the logic stays enabled until after `is_closing` is set, that would probably be better.

carver: So here's how I think it happens:
1. Peer gets cancelled for whatever reason: `await peer.cancel()`
2. This releases the peer's `run()` method:
https://github.com/ethereum/trinity/blob/27bbf61209d81b894292e36904a8b634fcb0094a/p2p/peer.py#L285
3. `AsyncExitStack` triggers `await _.aclose()` on all the internal contexts, like `p2p_api`:
https://github.com/ethereum/trinity/blob/27bbf61209d81b894292e36904a8b634fcb0094a/p2p/peer.py#L264-L265
4. When the behavior's `apply()` context manager exits, it triggers an exit on the `Subscription` returned by `Connection.add_logic`:
https://github.com/ethereum/trinity/blob/27bbf61209d81b894292e36904a8b634fcb0094a/p2p/logic.py#L90-L91
5. The subscription exit had been set to `connection.remove_logic`
https://github.com/ethereum/trinity/blob/27bbf61209d81b894292e36904a8b634fcb0094a/p2p/connection.py#L246-L247
6. Now the logic has been removed from the `Connection`
7. Steps 3-6 repeat for all the other behaviors on the peer (losing the hold on the event loop each time)
8. In between one of the behaviors exiting above ^, the event loop runs some other coro that wants to use the peer. It tries a `peer.eth_api`, which triggers a `has_logic` check, which will return `False`, eventually raising an `UnknownAPI`. (Note that `self`/`Connection` has *not* been closed yet):
https://github.com/ethereum/trinity/blob/27bbf61209d81b894292e36904a8b634fcb0094a/p2p/connection.py#L252-L255
9. Only after Step 7 completes for all behaviors, then the peer's `finally` block actually closes the connection:
https://github.com/ethereum/trinity/blob/27bbf61209d81b894292e36904a8b634fcb0094a/p2p/peer.py#L296

Okay, now to decide what to do about it... Is treating a removed logic as a "pending" cancellation sufficient? (That's effectively what this PR does) Or should we do something else, like... *not* remove the logic, and just allow other calls to the API all the way up until the connection has cancelled.

That last option is kind of appealing at first glance.

----

Mildly related: I don't think this disconnect is ever working...
https://github.com/ethereum/trinity/blob/27bbf61209d81b894292e36904a8b634fcb0094a/p2p/peer.py#L291

It happens after the p2p api has been removed from the connection. How is it not raising an exception? Am I missing something about how the behavior APIs exit?

gsalgado: > Only after Step 7 completes for all behaviors, then the peer's finally block actually closes the connection:

This is only a safety net, really. The connection most likely has been cancelled already, but its `.is_closing` will only be set to True at the very end of its `run()` method (when we **close** the multiplexer/transport), and that only happens after the Peer has finished its cleanup and returned as well -- this is achieved by running the peer runs as a child daemon of the connection.

I think the problem here is the fact that we want to keep the connection open/running until the Peer has finished its cleanup, but the connection behaviors are applied/removed in the `Peer.run()` method. Maybe all that should happen in `Connection.start_peer()` (or a new method that runs the peer, waits for it to finish and then removes the behaviors)? Maybe @pipermerriam has some ideas as well as he's more familiar with those abstractions

> Mildly related: I don't think this disconnect is ever working...

ISTM that works because we keep a reference to the P2PAPI instance:

> self.p2p_api = self.connection.get_logic('p2p', P2PAPI)

carver:
A lot of this is irrelevant now, but there are probably a few things worth extracting before closing it down.

gsalgado:
A lot of the code touched by this has been rewritten and it has a bunch of conflicts now, so I think it's probably not worth trying to resurrect this

PR #1675, opened 2020-04-16
Remove support for Python 3.6

cburgdorf:
### What was wrong?

Python 3.7 has been out for almost two years and continuing to support Python 3.6 is starting to feel unnecessary especially when we consider that the latest Ubuntu LTS ships with Python 3.8.

### How was it fixed?

Removed all Python 3.6 jobs from CI and adjusted some references in texts.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


cburgdorf:
This landed in another PR

PR #1677, opened 2020-04-17
p2p: Convert remaining BaseServices to new-style Services

gsalgado:


p2p/peer.py
```diff
@@ -162,6 +165,16 @@ def __init__(self,
...
         self.ready = asyncio.Event()
 
+    @property
+    def uptime(self) -> float:
+        if self._start_time is None:
+            return 0.0
+        else:
+            return time.monotonic() - self._start_time
+
+    def add_finished_callback(self, finished_callback: Callable[['BasePeer'], None]) -> None:
```
cburgdorf, line 175: I don't see this being used anywhere in this PR...

p2p/peer_pool.py
```diff
@@ -269,18 +262,19 @@ def unsubscribe(self, subscriber: PeerSubscriber) -> None:
...
     async def start_peer(self, peer: BasePeer) -> None:
-        self.manager.run_child_service(peer.connection.as_new_service())
-        await asyncio.wait_for(peer.connection.events.started.wait(), timeout=PEER_READY_TIMEOUT)
+        self.manager.run_child_service(peer.connection)
+        await asyncio.wait_for(
+            peer.connection.get_manager().wait_started(), timeout=PEER_READY_TIMEOUT)
 
         await peer.connection.run_peer(peer)
-        if peer.is_operational:
+        if peer.manager.is_running:
```
cburgdorf, line 270: Should this be `peer.get_manager().is_running`? I'm under the impression that `.manager` should only be used on `self` from *within*  a service, not to retrieve the manager of a different service (because `.manager` returns an `InternalManagerAPI` whereas `get_manager()` returns a `ManagerAPI`.

p2p/peer_pool.py
```diff
@@ -269,18 +262,19 @@ def unsubscribe(self, subscriber: PeerSubscriber) -> None:
...
             self._add_peer(peer, ())
         else:
             self.logger.debug("%s was cancelled immediately, not adding to pool", peer)
 
         try:
             await asyncio.wait_for(
-                peer.boot_manager.events.finished.wait(),
+                peer.boot_manager.manager.wait_finished(),
```
cburgdorf, line 277: Same here, should this be `peer.boot_manager.get_manager().wait_finished()`?

p2p/peer_pool.py
```diff
@@ -291,7 +285,7 @@ async def start_peer(self, peer: BasePeer) -> None:
             self.connection_tracker.record_failure(peer.remote, err)
             raise
         else:
-            if not peer.is_operational:
+            if not peer.manager.is_running:
```
cburgdorf, line 288: same here about `get_manager()`

p2p/peer_pool.py
```diff
@@ -493,7 +483,7 @@ async def __aiter__(self) -> AsyncIterator[BasePeer]:
             # Yield control to ensure we process any disconnection requests from peers. Otherwise
             # we could return peers that should have been disconnected already.
             await asyncio.sleep(0)
-            if peer.is_operational and not peer.is_closing:
+            if peer.manager.is_running and not peer.is_closing:
```
cburgdorf, line 486: same here about `get_manager()`

p2p/peer_pool.py
```diff
@@ -514,7 +504,7 @@ async def _periodically_report_stats(self) -> None:
             # make a copy, because we might edit the original during iteration
             peers = tuple(self.connected_nodes.values())
             for peer in peers:
-                if not peer.is_operational:
+                if not peer.manager.is_running:
```
cburgdorf, line 507: same here about `get_manager()`

p2p/tools/factories/connection.py
```diff
@@ -48,10 +45,9 @@ async def ConnectionPairFactory(*,
         bob_private_key=bob_private_key,
         bob_client_version=bob_client_version,
         bob_p2p_version=bob_p2p_version,
-        cancel_token=cancel_token,
         start_streams=start_streams
     )
-    async with run_service(alice_connection), run_service(bob_connection):
+    async with background_asyncio_service(alice_connection), background_asyncio_service(bob_connection):  # noqa: E501
```
cburgdorf, line 50: This is just a nitpick, feel free to ignore, but I usually prefer doing what `black` does to get these shorter.

```python
async with background_asyncio_service(alice_connection), background_asyncio_service(
    bob_connection
):
    if start_streams:
        alice_connection.start_protocol_streams()
        bob_connection.start_protocol_streams()

```

trinity/nodes/base.py
```diff
@@ -6,7 +6,7 @@
 )
 
 from async_service import Service, ServiceAPI
-from cancel_token import CancelToken
+from cancel_token.token import CancelToken
```
cburgdorf, line 9: Is there anything holding us back from getting rid of this and the `master_cancel_token`?

trinity/sync/beam/backfill.py
```diff
@@ -51,11 +51,7 @@ def __init__(
             self,
             db: AtomicDatabaseAPI,
             peer_pool: ETHPeerPool,
-            token: CancelToken = None) -> None:
-
-        # in case there is no token set, make sure this gets cancelled when the peer pool does
-        if token is None:
-            token = peer_pool.cancel_token
+            token: CancelToken) -> None:
```
cburgdorf, line 54: Just curious: Is this just a "I didn't want to make the PR any bigger" type of thing or is there anything other reason why `BeamStateBackFill` still uses the old service / cancel token model?

p2p/peer.py
gsalgado: This is part of the `BaseService` API, but something we don't have in `async_service.Service`, so I added it here. It's not used in any of our other `BaseService`s so I saw no point in trying to add that to the new `Service`

p2p/peer_pool.py
gsalgado: Yeah, it should!

p2p/peer_pool.py
gsalgado: Yep

trinity/nodes/base.py
gsalgado: Yes, the `SyncerComponent` still needs to pass that to the syncer, which is a `BaseService`

trinity/sync/beam/backfill.py
gsalgado: BeamStateBackFill is one of the `BaseService`s left under `trinity`. I'm only converting the ones under the `p2p` package here.

p2p/peer.py
cburgdorf: Ah! Makes sense!

trinity/sync/beam/backfill.py
cburgdorf: Alright!

pipermerriam:
Awesome!

PR #1679, opened 2020-04-17
Add support for Python 3.8

cburgdorf:
### What was wrong?

When we get rid of Python 3.6 support we should make sure we support Python 3.8 by testing against it in CI.

### How was it fixed?

Add CI jobs for Python 3.8

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -198,6 +198,7 @@ setup(
         'License :: OSI Approved :: MIT License',
         'Natural Language :: English',
         'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
     ],
     # trinity
     entry_points={
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ deps = {
         # idna 2.7 is not supported by requests 2.18
         "requests>=2.20,<3",
         "ipython>=7.8.0,<7.10.0",  # attach fails with v7.10.{0,1}
-        "plyvel==1.1.0",
+        "plyvel==1.2.0",
         PYEVM_DEPENDENCY,
         "web3>=5.10.0,<6",
         "lahja>=0.16.0,<0.17",
@@ -52,7 +52,7 @@ deps = {
         "jsonschema==3.0.1",
         "mypy-extensions>=0.4.3,<0.5.0",
         "typing_extensions>=3.7.4,<4.0.0",
-        "ruamel.yaml==0.15.98",
+        "ruamel.yaml==0.16.10",
         "argcomplete>=1.10.0,<2",
         "multiaddr>=0.0.8,<0.1.0",
         "prometheus-client==0.7.1",
@@ -76,7 +76,7 @@ deps = {
         # xdist pinned at <1.29 due to: https://github.com/pytest-dev/pytest-xdist/issues/472
         "pytest-xdist>=1.29.0,<1.30",
         # only for eth2
-        "ruamel.yaml==0.15.98",
+        "ruamel.yaml==0.16.10",
         "eth-tester==0.4.0b2",
     ],
     # We have to keep some separation between trio and asyncio based tests
@@ -125,7 +125,7 @@ deps = {
         "eth-keyfile",  # validator client
     ],
     'eth2-extra': [
-        "milagro-bls-binding==0.1.3",
+        "milagro-bls-binding==0.1.4",
     ],
     'eth2-lint': [
         "black==19.3b0",
--- a/setup.py
+++ b/setup.py
@@ -188,7 +188,7 @@ setup(
     url='https://github.com/ethereum/trinity',
     include_package_data=True,
     py_modules=['trinity', 'p2p', 'eth2'],
-    python_requires=">=3.6,<4",
+    python_requires=">=3.7,<4",
     install_requires=install_requires,
     extras_require=deps,
     license='MIT',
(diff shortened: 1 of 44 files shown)


cburgdorf:
@ChihChengLiang do you have any idea what's preventing the `milagro-bls-binding` from getting properly installed under a Python 3.8 environment?

cburgdorf:
@pipermerriam No you're right, there was a huge chunk of things that we could get rid off. I updated the PR.

ChihChengLiang:
Hi @cburgdorf, I think I missed some release of `milagro-bls-binding` binding for 3.8 version. Just fixed it now, would you like to give it another try?

cburgdorf:
@ChihChengLiang Thanks for fixing the Python 3.8 builds! Unfortunately, it seems like the API has changed substantially enough to throw me off :sweat_smile: For example, `sign` has no `domain` anymore, and `verify_multiple` is gone. Would you mind handling the upgrade to the latest version in a standalone PR?

ChihChengLiang:
@cburgdorf I'm terribly sorry for saying it's ready for the upgrade and I'm so sorry for your time on trying to upgrade that 🙈, I totally forgot the breaking API change. The API change would require a lengthy PR because it touches eth2 code and eth2-fixture tests.

Here are some band-aids for that. 

- The quickest possible way is to drop the `eth2-extra` requirement in CI, the tests would fall back to using `py-ecc` as the bls library. 
- Meanwhile, I'll try to release an old API compatible py 3.8 release of `milagro-bls-binding`, under version `0.1.x`. If it works, it makes existing Trinity codebase works. I'll yank out the `0.2.0`, `0.3.0`, and `0.4.0`, as those versions are using the new API. The new API will be released under `2.0.x`. I'll update in this thread when it's done.

I'm not able to commit to the upgrade of Trinity codebase to the new API yet, due to the other ongoing tasks. But happy to do it a while later if no one else is on it.

ChihChengLiang:
@cburgdorf I just released `milagro-bls-binding==0.1.4` for py3.8 and old API support. The installation in Trinity looks fine and CI tests depend on that pass [here](https://github.com/ChihChengLiang/trinity/pull/2). Hope this time works 😂.

cburgdorf:
@ChihChengLiang Awesome! That fixes it. Also, don't worry too much, I haven't wasted much time on the issue and it isn't urgent. Thank you very much for cutting a new release :pray:

cburgdorf:
@carver Are we good to trade Python 3.6 against Python 3.8 support by now?

carver:
> @carver Are we good to trade Python 3.6 against Python 3.8 support by now?

What is the urgency to drop 3.6 again? Can't we add 3.8 support without dropping 3.6?

cburgdorf:
1. We can get rid of a bunch of Python 3.6 shims. I also suspect there may be some other features we could use that do not have shims available.

2. Python 3.6 does not support pickleing of generic classes (Something that our [pending `eth/66` implementation uses](https://github.com/ethereum/trinity/pull/1672/ but which I suspect might also be useful for various lahja events e.g.  `eventbus.stream(Something[PayloadX])`)

3. Adding Python 3.8 support next to Python 3.6 would also increase our CI times.

setup.py
```diff
@@ -200,7 +197,7 @@ def filter_dependencies(package_list, *package_name):
         'Intended Audience :: Developers',
         'License :: OSI Approved :: MIT License',
         'Natural Language :: English',
-        'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
```
carver, line 200: I guess we can add a line for 3.8, too.

tox.ini
```diff
@@ -1,15 +1,15 @@
...
     py37-rpc-state-{quadratic,sstore,zero_knowledge}
     py37-{libp2p,eth2-components}
-    py{36,37}-lint
-    py{36,37}-lint-eth2
-    py{36,37}-wheel-cli
-    py36-docs
+    py{38,37}-lint
+    py{38,37}-lint-eth2
+    py{38,37}-wheel-cli
+    py38-docs
```
carver, line 12: IMO, any of the single-run environments should use the oldest supported version, to help verify that we correctly operate in that environment. Things like `docs`, `long_run_integration`, etc.

cburgdorf: That's a good point. I'll fix that before the merge.

tox.ini
```diff
@@ -1,15 +1,15 @@
 [tox]
 envlist=
-    py{36,37}-{eth1-core,p2p,p2p-trio,integration,lightchain_integration,eth2-core,eth2-fixtures,eth2-integration,eth1-components,eth2-utils,eth2-trio}
-    py36-long_run_integration
-    py36-rpc-blockchain
-    py36-rpc-state-{frontier,homestead,tangerine_whistle,spurious_dragon,byzantium,constantinople,petersburg,istanbul}
+    py{37,38}-{eth1-core,p2p,p2p-trio,integration,lightchain_integration,eth2-core,eth2-fixtures,eth2-integration,eth1-components,eth2-utils,eth2-trio}
+    py37-long_run_integration
+    py37-rpc-blockchain
+    py38-rpc-state-{frontier,homestead,tangerine_whistle,spurious_dragon,byzantium,constantinople,petersburg,istanbul}
```
carver, line 6: Looks like this one got missed, I expected: `py37`

veox:
So this added proper Python 3.8 support (for running the node, etc.), or does it only make CI pass?

Because if the former, then the `newsfragments` doesn't reflect it, and the fact will not be in the release notes.

cburgdorf:
@veox oh you're right. I somehow missed to mention Python 3.8 support. I should work but afaik there are still some quirks with Python 3.8 that need to be resolved.

carver:
We can add the release note when we upgrade to a py-evm version that includes https://github.com/ethereum/py-evm/pull/1940

PR #1682, opened 2020-04-19
Support blockchain stats for grafana

njgheorghita:
### What was wrong?
Some endpoints were missing for influxdb/grafana metrics reporting. 

### How was it fixed?
First stab at supporting the `blockchain` stats from the geth dashboard in trinity. 
- `latest block`
- `latest header`
- `latest receipt`

I wanted to play around / experiment with implementing these endpoints (partially inspired by how the `attach` component works), but I'm sure there might be a better solution for grabbing this data. I'm also a little confused by what exactly `latest receipt` and `latest header` means - and if they would ever differ from `latest block` (which they don't seem to on the geth dashboard).

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/trinity/_utils/connect.py
+++ b/trinity/_utils/connect.py
@@ -9,7 +9,10 @@ from trinity.db.eth1.header import AsyncHeaderDB
 from trinity.chains.light_eventbus import (
     EventBusLightPeerChain,
 )
-from trinity.config import TAppConfig
+from trinity.config import (
+    Eth1AppConfig,
+    TAppConfig,
+)
 from trinity.constants import (
     SYNC_LIGHT,
 )
@@ -17,12 +20,10 @@ from trinity.db.manager import DBClient
 
 
 @contextlib.contextmanager
-def get_chain(app_config_type: Type[TAppConfig],
-              boot_info: BootInfo,
-              event_bus: EndpointAPI) -> Iterator[ChainAPI]:
-    app_config = boot_info.trinity_config.get_app_config(app_config_type)
-    # ignore b/c TAppConfig has no attribute get_chain_config
-    chain_config = app_config.get_chain_config()  # type: ignore
+def get_eth1_chain_with_remote_db(boot_info: BootInfo,
+                                  event_bus: EndpointAPI) -> Iterator[ChainAPI]:
+    app_config = boot_info.trinity_config.get_app_config(Eth1AppConfig)
+    chain_config = app_config.get_chain_config()
 
     chain: ChainAPI
     base_db = DBClient.connect(boot_info.trinity_config.database_ipc_path)
--- a/trinity/_utils/connect.py
+++ b/trinity/_utils/connect.py
@@ -1,0 +1,39 @@
+from typing import Iterator, Type
+import contextlib
+
+from eth.abc import ChainAPI
+
+from lahja import EndpointAPI
+from trinity.boot_info import BootInfo
+from trinity.db.eth1.header import AsyncHeaderDB
+from trinity.chains.light_eventbus import (
+    EventBusLightPeerChain,
+)
+from trinity.config import TAppConfig
+from trinity.constants import (
+    SYNC_LIGHT,
+)
+from trinity.db.manager import DBClient
+
+
+@contextlib.contextmanager
+def get_chain(app_config_type: Type[TAppConfig],
+              boot_info: BootInfo,
+              event_bus: EndpointAPI) -> Iterator[ChainAPI]:
+    app_config = boot_info.trinity_config.get_app_config(app_config_type)
+    # ignore b/c TAppConfig has no attribute get_chain_config
+    chain_config = app_config.get_chain_config()  # type: ignore
+
+    chain: ChainAPI
+    base_db = DBClient.connect(boot_info.trinity_config.database_ipc_path)
+    with base_db:
+        if boot_info.args.sync_mode == SYNC_LIGHT:
+            header_db = AsyncHeaderDB(base_db)
+            chain = chain_config.light_chain_class(
+                header_db,
+                peer_chain=EventBusLightPeerChain(event_bus)
+            )
+        else:
+            chain = chain_config.full_chain_class(base_db)
+
+        yield chain
--- a/trinity/_utils/connect.py
+++ b/trinity/_utils/connect.py
@@ -1,4 +1,4 @@
-from typing import Iterator, Type
+from typing import Iterator
 import contextlib
 
 from eth.abc import ChainAPI
@@ -11,7 +11,6 @@ from trinity.chains.light_eventbus import (
 )
 from trinity.config import (
     Eth1AppConfig,
-    TAppConfig,
 )
 from trinity.constants import (
     SYNC_LIGHT,
--- a/trinity/components/builtin/metrics/blockchain_metrics_collector.py
+++ b/trinity/components/builtin/metrics/blockchain_metrics_collector.py
@@ -15,9 +15,7 @@ from trinity._utils.connect import get_eth1_chain_with_remote_db
 
 
 class BlockchainStats(NamedTuple):
-    latest_header: int
     latest_block: int
-    latest_receipt: int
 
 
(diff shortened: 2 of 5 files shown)


trinity/components/builtin/metrics/component.py
```diff
@@ -141,7 +144,18 @@ async def do_run(cls, boot_info: BootInfo, event_bus: EndpointAPI) -> None:
             frequency_seconds=boot_info.args.metrics_system_collector_frequency
         )
 
-        services_to_exit = (metrics_service, system_metrics_collector,)
+        # types ignored due to https://github.com/ethereum/async-service/issues/5
+        blockchain_metrics_collector = collect_blockchain_metrics(  # type: ignore
+            boot_info.trinity_config,
+            metrics_service.registry,
+            frequency_seconds=boot_info.args.metrics_system_collector_frequency
```
cburgdorf, line 151: Notice that you probably want the blockchain metrics on a slower frequency. It doesn't need to be higher than the average blocktime. I think something like `15` seconds should be good.

trinity/components/builtin/metrics/blockchain_metrics_collector.py
```diff
@@ -0,0 +1,67 @@
...
+    # PersistentSocket must be closed here to avoid unclosed socket error
+    socket = ipc_provider._socket.__enter__()
+    socket.close()
+
+    block = to_int(hexstr=response['result'])
+    return BlockchainStats(
+        latest_header=block,  # ?
+        latest_block=block,
+        latest_receipt=block,  # ?
+    )
```
cburgdorf, line 45: Am I right to assume that you are trying to connect this widget from the dashboard?

image

If so then, all I can say is that I'd also have to dig into the geth source code to find out how exactly they are implemented. But here's my guess. When a node syncs, all the different pieces of the block chain are fetched in different sub processes. So, you might have already the latest header right at the tip of the chain, but you still have to catch up with the blocks which might still be 2000 blocks away from the tip. And receipts may be yet another storry, mayb you are 50000 blocks away from the tip when it comes to how far you've already caught up downloading receipts. I think that is, what this dashboard element is supposed to show.

trinity/components/builtin/metrics/blockchain_metrics_collector.py
```diff
@@ -0,0 +1,67 @@
...
+    if not ipc_path.exists():
+        return BlockchainStats(
+            latest_header=0,
+            latest_block=0,
+            latest_receipt=0,
+        )
+
+    # Is there a better way to get this data and avoid importing web3
+    ipc_provider = web3.IPCProvider(ipc_path)
+    response = ipc_provider.make_request('eth_blockNumber', None)
```
cburgdorf, line 34: Yeah, that's a very indirect way of getting the latest block. What you should be doing here is to work with the chain directly.

```python
latest_header = chain.get_canonical_head()
```

In fact, that is also what happens when you hit the `eth_blockNumber` RPC API.

https://github.com/ethereum/trinity/blob/8b4c56b363ebae885cb981411d2d14a85018a2c2/trinity/rpc/modules/eth.py#L157

Now you're probably thinking "ok, where does the chain come from?". Check this example of the ethstats service getting a hold of the chain.

https://github.com/ethereum/trinity/blob/8b4c56b363ebae885cb981411d2d14a85018a2c2/trinity/components/builtin/ethstats/ethstats_service.py#L157-L174

In fact, I think we could either put this in a helper function or even implement a method directly on the `TrinitiyConfig` to not duplicate the logic everywhere.

njgheorghita:
@cburgdorf Thanks for the pointers! I decided to go for a  utility function, since it felt clunkier to move it onto `TrinityConfig`. Just a couple questions...

- I'm not sure I have a great context for what the name of the utility file (`connect.py`) and utility function `get_chain` should be. I'm not sure if those are great names. When I dropped the `get_chain` into `_utils/chains.py` it introduced a circular import so I opted for a separate module.

- It wasn't obvious to me the best way to access the most recently downloaded receipt and block number - can you point me in the right direction?

- Is there any strategy for testing ethstats / grafana? I've just been running  them locally to make sure things are hooked up correctly.

trinity/_utils/connect.py
```diff
@@ -0,0 +1,39 @@
...
+from trinity.db.manager import DBClient
+
+
+@contextlib.contextmanager
+def get_chain(app_config_type: Type[TAppConfig],
+              boot_info: BootInfo,
+              event_bus: EndpointAPI) -> Iterator[ChainAPI]:
+    app_config = boot_info.trinity_config.get_app_config(app_config_type)
+    # ignore b/c TAppConfig has no attribute get_chain_config
+    chain_config = app_config.get_chain_config()  # type: ignore
```
cburgdorf, line 25: What do you think about just basing this on `Eth1AppConfig`. As you noticed `get_chain_config` isn't generalized and the other code that follows is even less generalized. If we base this on `Eth1AppConfig` we won't have to `# type ignore` anything.

trinity/_utils/connect.py
```diff
@@ -0,0 +1,39 @@
...
+)
+from trinity.config import TAppConfig
+from trinity.constants import (
+    SYNC_LIGHT,
+)
+from trinity.db.manager import DBClient
+
+
+@contextlib.contextmanager
+def get_chain(app_config_type: Type[TAppConfig],
```
cburgdorf, line 20: Disclaimer, I usually come up with bad names and I *really*, *really* try not to. But since you asked, I'd go with something like `get_eth1_chain_with_remote_db` (assuming you do the change to restrict this to `Eth1AppConfig`).

This would convey that we get a chain that uses some form of IPCing to a database (rather than having a local connection to the database which LevelDB only allows for a single process) and it squeezes in an `eth1` to allow for a similar function with a `beacon` tag if needed in the future.

trinity/components/builtin/metrics/blockchain_metrics_collector.py
```diff
@@ -0,0 +1,51 @@
...
+
+@as_service
+async def collect_blockchain_metrics(manager: ManagerAPI,
+                                     boot_info: BootInfo,
+                                     event_bus: EndpointAPI,
+                                     registry: HostMetricsRegistry,
+                                     frequency_seconds: int) -> None:
+    blockchain_header_gauge = registry.gauge('trinity.blockchain/head/header.gauge')
+    blockchain_block_gauge = registry.gauge('trinity.blockchain/head/block.gauge')
+    blockchain_receipt_gauge = registry.gauge('trinity.blockchain/head/receipt.gauge')
```
cburgdorf, line 45: Just to double check, you also configure these in the shared dashboard that you have access to (not just in a local grafana install). I just want to make sure we don't get out of sync anywhere.

cburgdorf:
>It wasn't obvious to me the best way to access the most recently downloaded receipt and block number 

I *think* that there's no straightforward way to retrieve the blocknumber up to which we have all blocks and receipts. At least not straight from the db. I think we don't have an index for that to look it up. I believe that this information may be present to the syncing process (depending on the actual syncing strategy) but not outside it. But I also don't think our default sync (`beam`) does backfilling of blocks and receipts already. @carver would be the best to know that.

>Is there any strategy for testing ethstats / grafana? I've just been running them locally to make sure things are hooked up correctly.

No, not really. As far as Grafana goes, Trinity doesn't even know it exists. It just writes metrics into the InfluxDB and Grafana happens to be a tool to visualize them but we could also use different tools for that as long as they can pull the data out of InfluxDB.

njgheorghita:
danke @cburgdorf ! This pr is ready for final :eyes: whenever you get the chance.

> Just to double check, you also configure these in the shared dashboard that you have access to (not just in a local grafana install). I just want to make sure we don't get out of sync anywhere.

Yeah, I'm having some issues with the `geth-bench` dashboard right now. But, I'll be sure to keep the shared dashboard synced with any new endpoints.

cburgdorf:
>Yeah, I'm having some issues with the geth-bench dashboard right now.

Oh, interesting, I had trouble, too. Because I wanted to check if it was updated and it kept prompting me a login screen.

PR #1684, opened 2020-04-20
Use background_asyncio_service instead of run_asyncio_service

gsalgado:
run_asyncio_service is broken (https://github.com/ethereum/async-service/issues/65)

### DIFF
--- a/trinity/components/builtin/tx_pool/component.py
+++ b/trinity/components/builtin/tx_pool/component.py
@@ -4,7 +4,7 @@ from argparse import (
 )
 import logging
 
-from async_service import run_asyncio_service
+from async_service import background_asyncio_service
 from eth_utils import ValidationError
 from eth.chains.mainnet import ISTANBUL_MAINNET_BLOCK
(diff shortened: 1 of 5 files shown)


trinity/components/builtin/tx_pool/component.py
```diff
@@ -94,7 +94,5 @@ async def do_run(cls, boot_info: BootInfo, event_bus: EndpointAPI) -> None:
 
             tx_pool = TxPool(event_bus, proxy_peer_pool, validator)
 
-            try:
-                await run_asyncio_service(tx_pool)
-            finally:
-                cls.logger.info("Stopping Tx Pool...")
```
carver, line 100: Do we definitely want to drop this log, or was it an oversight?

gsalgado: I don't think we need an INFO log for that sort of thing -- we don't do that for any other components, either. And we have plenty of DEBUG ones for the same effect

Issue #1668, opened 2020-04-13
No DEBUG logs from asyncio_run_in_process shows up in our log files


gsalgado:
None of the DEBUG logs emitted by asyncio_run_in_process shows up in our log files. I suspect it's because the logger instance used there is created at import time, so it gets the default level (INFO) and not what trinity sets up during startup.

gsalgado:
Same goes for any DEBUG logs from `trinity/extensibility/component.py`, which also creates the logger instance at import time

pipermerriam:
Some important context.

`asyncio-run-in-process` will be using `loggng.getLogger` to setup it's loggers.

The `eth_utils` logging utility functions do some fancy stuff under the hood.  Specifically this:

https://github.com/ethereum/eth-utils/blob/b3146fc7c283208dee29a549372eac1631c5ea3b/eth_utils/logging.py#L79-L81

gsalgado:
It doesn't look like this is caused by the loggers being created at the module level. Any ideas?

Issue #1347, opened 2019-12-03
test_logger_configuration is failing


ChihChengLiang:
### What is wrong?

After the fix of async runner in #1338, these tests turn out to be failing and were marked as `xfail`

https://github.com/ethereum/trinity/blob/0e3e67d2ceb52e124fcb866f11d3f9704cfe04bf/tests/integration/test_trinity_cli.py#L247-L270

### How can it be fixed

Need to investigate further

cburgdorf:
I wanted to understand this better and looked into it today. Here's a quick brain dump.

It seems that this is a regression that came with #1277 when the new logging was introduced that doesn't rely on `multiprocessing.Queue` anymore. It seems that the regression made it's way in because the `AsyncProcessRunner` had a bug where a test would sometimes automatically pass when the timeout of the test was hit. That bug was fixed by #1338 which now exposes this regression caused by #1277.

I wanted to prove that so I went back to an older commit and just cherry-picked the fix from #1388 to see if all the logging tests would still pass which is the case.

I tried to come up with a fix but couldn't find the root cause yet so I wanted to at least did a quick brain dump.

/cc @pipermerriam

pipermerriam:
Nothing immediately comes to mind.  I believe someone will just need to dig in and see where things aren't being correctly done.

Also, I think that it would be better if we could streamline how these are tested since the current mechanism is very indirect...

gsalgado:
ISTM that there's something wrong with the code responsible for capturing trinity's output and matching it against the expected/unexpected patterns. I say that because for `test_logger_configuration[command1,...]` we expect only to see `{'Started main process', '>>> ping'}` patterns, and in two subsequent runs, both containing the expected pattern, one failed and the other passed. http://sprunge.us/PS4BLy is the one that failed, and http://sprunge.us/uDdS1U is the one that passed

gsalgado:
And another clue that might lead us to to the buggy code is the fact that all tests that fail intermittently are the ones that have empty sets for one or more of the expected/unexpected patterns

Issue #689, opened 2019-06-04
Make file logging respect individual log level increase


cburgdorf:
### What is wrong?

On current master, it is not possible to reduce the general log level to e.g. `ERROR` but increase the log level for one specific module e.g. set `p2p.discovery` to `DEBUG2`.

With #682 this becomes possible at least for stderr logs. E.g. the following works as expected:

`trinity -l ERROR -l 'p2p.discovery=DEBUG2'`.

For `stderr` this does only print `ERROR` logs for everything except `p2p.discovery` where also `DEBUG2` logs or higher are printed.

For the file log this will only print `ERROR` because it is currently not possible to increase log level of individual modules above the general file log level.

### How can it be fixed

1. Probably some sort of re-attaching log handler magic inside https://github.com/ethereum/trinity/blob/9fe8fa87689428608f76eefb4415c85dc28ff646/trinity/_utils/logging.py#L103-L108

2. Enable the test currently commented test case https://github.com/ethereum/trinity/blob/38cc2bb2119104030f3ec54e2022932fc73c293f/tests/integration/test_trinity_cli.py#L208-L219

jannikluhn:
For some tests I did a while ago I used a simple YAML file in conjunction with [logging.config.dictConfig](https://docs.python.org/3.8/library/logging.config.html#logging.config.dictConfig) to configure logging (see https://github.com/ethereum/trinity/pull/1033/files#diff-efaf22e24a01e85cc46fa436ca244817R11-R25 and https://github.com/ethereum/trinity/pull/1033/files#diff-69674a1dd689a764b5421acc5f5b24b2R1-R35).

This turned out to be quite useful for debugging because editing a YAML file is much easier than a multi-line CLI command if one wants to specify lots of options (at least for me). Also, one can quickly switch between multiple very different configurations (e.g. for a client or a server). Because of `dictConfig` it's also maximally powerful, e.g. one can also change the formatting.

Not suggesting this to be the only option, but as an additional option this could be very helpful.

pipermerriam:
I could see allowing this via something like `--logging-config`.  We'd probably want some logic that would detect duplicates between the supplied logging config and the command line flags for logging.  Also probably allow command line flags to override things in the config file.  And last, if there are duplicates between the config file and the CLI flags, we probably want to log those during application startup.  :+1:

cburgdorf:
I think that at some point we want a general config file for Trinity anyway. Not only for logging also for all kind of other things like preferred peers, max peers etc.

PR #1688, opened 2020-04-20
Fix logging on the main process

gsalgado:
The root logger of the main process was configured with the level
used for the stderr handler, so any logs from the main process with
severity lower than that would be filtered before reaching the file
handler. Now we use the minimum level on the root logger of the
main process, like we do for the root logger on child processes.

Closes: #1668

### DIFF
--- a/tests/integration/test_trinity_cli.py
+++ b/tests/integration/test_trinity_cli.py
@@ -270,16 +270,17 @@ async def test_does_not_throw_errors_on_short_run(command, unused_tcp_port):
             {'Started main process', '<Manager[ConnectionTrackerServer] flags=SRcfe>: running root task run[daemon=False]'},  # noqa: E501
             {'>>> ping'},
         ),
-        pytest.param(
-            # Reduce stderr logging to ERROR logs but report DEBUG2 or higher for file logs
-            ('trinity', '--stderr-log-level=ERROR', '--file-log-level=DEBUG2',),
-            {},
-            {'Started main process', '>>> ping'},
-            {'Started main process', '>>> ping'},
-            {},
-            # TODO: investigate in #1347
-            marks=(pytest.mark.xfail),
-        ),
+        # Commented out because sometimes it passes and sometimes it fails.
+        # pytest.param(
+        #     # Reduce stderr logging to ERROR logs but report DEBUG2 or higher for file logs
+        #     ('trinity', '--stderr-log-level=ERROR', '--file-log-level=DEBUG2',),
+        #     {},
+        #     {'Started main process', '>>> ping'},
+        #     {'Started main process', '>>> ping'},
+        #     {},
+        #     # TODO: investigate in #1347
+        #     marks=(pytest.mark.xfail),
+        # ),
         pytest.param(
             # Reduce everything to ERROR logs, except discovery that should report DEBUG2 or higher
             ('trinity', '-l=ERROR', '-l', 'p2p.discovery=DEBUG2'),
@@ -323,9 +324,7 @@ async def test_logger_configuration(command,
                 break
             if "DiscoveryService" in line:
                 marker_seen_at = time.time()
-                stderr_logs.append(line)
-            else:
-                stderr_logs.append(line)
+            stderr_logs.append(line)
 
         for log in expected_stderr_logs:
             if not contains_substring(stderr_logs, log):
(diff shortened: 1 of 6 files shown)


tests/integration/test_trinity_cli.py
```diff
@@ -319,7 +319,7 @@ def contains_substring(iterable, substring):
         # for assertions
         marker_seen_at = 0
         async for line in runner.stderr:
-            if marker_seen_at != 0 and time.time() - marker_seen_at > 3:
+            if marker_seen_at != 0 and time.time() - marker_seen_at > 5:
```
gsalgado, line 322: @carver so, this was the cause of the failure. we now have more log lines, so need to capture a few more seconds' worth of stderr to ensure we get what we're looking for.

tests/integration/test_trinity_cli.py
```diff
@@ -319,7 +319,7 @@ def contains_substring(iterable, substring):
         # for assertions
         marker_seen_at = 0
         async for line in runner.stderr:
-            if marker_seen_at != 0 and time.time() - marker_seen_at > 3:
+            if marker_seen_at != 0 and time.time() - marker_seen_at > 8:
```
gsalgado, line 322: So, I had to make this change because we now have more log lines and so need to collect more seconds' worth of stderr to get to the point we want. I first tried 5s, but it wasn't enough, and with 8 we get a `asyncio.streams.LimitOverrunError` (https://circleci.com/api/v1.1/project/github/ethereum/trinity/215477/output/107/0?file=true&allocation-id=5e9e3f21e36e38143fff989d-0-build%2F30EF02D2). @pipermerriam, @cburgdorf any suggestions on how to go about this? these CLI tests are giving me a really hard time lately :-(

gsalgado:
I've increased the stream limit so that test now passes, but then one of our xfail tests passed as well:

https://circleci.com/api/v1.1/project/github/ethereum/trinity/215727/output/107/0?file=true&allocation-id=5e9e8b4a3324b2678fd57c86-0-build%2F1F22A450
```
_ test_logger_configuration[command3-expected_stderr_logs3-unexpected_stderr_logs3-expected_file_logs3-unexpected_file_logs3] _
[XPASS(strict)] 
```

But it doesn't seem to always pass, so I can't just remove the xfail marker as suggested in #689

tests/integration/test_trinity_cli.py
cburgdorf: I'm really puzzled what kind of limit we are hitting here. This is supposed to be streaming line for line so if I'm not reading this wrongly it suggests that there's a line that is too large so that it hits `LimitOverrunError` before it had a chance to get to the line separator of that gigantic line. That's how I interpret it but it doesn't make sense to me.

>because we now have more log lines

Could there be a specific log line that we haven't had before that is causing this?

gsalgado: I don't think we have any log lines that could cause that, but I've recently seen crashes with tracebacks that exhausted my tmux's scrollback, so it could be that

gsalgado: So, it turns out this change is not needed, and without it we don't get the LimitOverrunError and thus don't need the bigger limit. However, one of the xfail tests in `test_logger_configuration()` now passes sometimes. I'll just comment it out and maybe one day it will be fixed as part of #1347

tests/integration/test_trinity_cli.py
```diff
@@ -270,16 +270,17 @@ async def test_does_not_throw_errors_on_short_run(command, unused_tcp_port):
...
-            {'Started main process', '>>> ping'},
-            {'Started main process', '>>> ping'},
-            {},
-            # TODO: investigate in #1347
-            marks=(pytest.mark.xfail),
-        ),
+        # Commented out because sometimes it passes and sometimes it fails.
+        # pytest.param(
+        #     # Reduce stderr logging to ERROR logs but report DEBUG2 or higher for file logs
+        #     ('trinity', '--stderr-log-level=ERROR', '--file-log-level=DEBUG2',),
```
pipermerriam, line 276: I recently had the thought that we should rewrite these tests such that they use custom components that emit known logging statements.  It would add a level of indirection with the benefit of less fragile tests and the cost of potentially still having a failure at some other level of our logging infrastructure setup that might not be caught due to the test loggers being setup differently.

PR #1695, opened 2020-04-24
Add a minimal trio-based JSON HTTP server

ralexstokes:
Supersedes #1645.

Add a very minimal JSON HTTP server that works w/ `trio`.

There doesn't seem to be a good HTTP server option for `trio`. This PR adapts an example from the `trio` maintainers and may need to be enhanced or replaced as we move towards production.

The server in this PR will do for now in terms of a minimal JSON API for the validator client <-> beacon node connection.

#### Cute Animal Picture

### DIFF
--- a/trinity/_utils/trio_utils.py
+++ b/trinity/_utils/trio_utils.py
@@ -1,5 +1,13 @@
+from itertools import count
+import json
+import logging
 import signal
+from typing import Any, Awaitable, Callable, Dict, Tuple
+from urllib.parse import urlparse
+from wsgiref.handlers import format_date_time
 
+from eth_utils.toolz import curry
+import h11
 import trio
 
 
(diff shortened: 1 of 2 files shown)


ralexstokes:
note: if we don't care about validator interop, we can simply use an IPC transport and not worry about HTTP and HTTP performance.

This being said there will be other instances where we want the beacon node to be discoverable via HTTP...

trinity/_utils/trio_utils.py
```diff
@@ -1,9 +1,405 @@
...
+    response_body_bytes = response_body_unicode.encode("utf-8")
+    await send_simple_response(
+        wrapper, 200, "application/json; charset=utf-8", response_body_bytes
+    )
+
+
+class JSONHTTPServer:
+    logger = logging.getLogger("trinity._utils.trio_utils.JSONHTTPServer")
+
+    def __init__(self, handlers: Handlers):
```
ralexstokes, line 400: very simple, but wraps the above functionality... all it wants is a `handlers` object like:

```python
handlers = {"/node/version": {"GET": lambda _request: "$CLIENT_VERSION"}
```

PR #1697, opened 2020-04-27
Ensure trio components terminate when trinity crashes

gsalgado:
I recently changed TrioIsolatedComponent to not send a SIGINT to the
subprocess as they run on the same process group of the main process and
thus get that when there's a Ctrl-C in the terminal, but that doesn't
happen when trinity crashes, so trio components were left running after
everything else had terminated. This fixes that (in a suboptimal way),
although hopefully it will be temporary

Another issue I thought would be straightforward to fix, but in the
end spent quite some time trying to figure out the RuntimeError
mentioned in one of the comments.

### DIFF
--- a/tests/core/components/test_isolated_component.py
+++ b/tests/core/components/test_isolated_component.py
@@ -6,14 +6,16 @@ import os
 from async_service import background_asyncio_service
 import pytest
 
-from trinity._utils.chains import (
-    get_local_data_dir,
-)
+from trinity._utils.chains import get_local_data_dir
 from trinity._utils.logging import IPCListener
 from trinity.boot_info import BootInfo
 from trinity.config import TrinityConfig
 from trinity.extensibility import ComponentManager
-from trinity.tools._component_isolation import IsStarted, AsyncioComponentForTest
+from trinity.tools._component_isolation import (
+    AsyncioComponentForTest,
+    IsStarted,
+    TrioComponentForTest,
+)
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


trinity/extensibility/trio.py
```diff
@@ -49,14 +53,36 @@ async def run(self) -> None:
...
+    def _send_signal_and_join(self, process, sig, timeout) -> None:
+        try:
+            os.kill(process.pid, sig)
+        except ProcessLookupError:
+            self.logger.debug("Process %d has already terminated", process.pid)
+            return
+        # XXX: Using process.join() here is far from ideal as it blocks the main process,
+        # forcing us to wait in sequence for every trio components, but try and run it
+        # asynchronously (using an executor, like above) and you'll get a
+        #   RuntimeError: cannot reuse already awaited coroutine
```
gsalgado, line 82: FWIW, we were using `kill_process_gracefully()` with a blocking call to `process.join()` before, so there's no change in behavior here.

tests/core/components/test_isolated_component.py
```diff
@@ -6,14 +6,13 @@
...
-    get_local_data_dir,
-)
+from trinity._utils.chains import get_local_data_dir
 from trinity._utils.logging import IPCListener
 from trinity.boot_info import BootInfo
 from trinity.config import TrinityConfig
 from trinity.extensibility import ComponentManager
-from trinity.tools._component_isolation import IsStarted, AsyncioComponentForTest
+from trinity.tools._component_isolation import (
+    AsyncioComponentForTest, IsStarted, TrioComponentForTest)
```
cburgdorf, line 15: formatting doesn't look correct here

tests/core/components/test_isolated_component.py
```diff
@@ -46,12 +45,12 @@ def log_listener(trinity_config):
         yield
 
 
+@pytest.mark.parametrize("component", (AsyncioComponentForTest, TrioComponentForTest))
```
cburgdorf, line 48: An aligned test that works across both component types, eu gosto disso :heart:

PR #1699, opened 2020-04-27
Collect and visualize the current head of peers

njgheorghita:
### What was wrong?
Re: https://github.com/ethereum/trinity/pull/1682#pullrequestreview-397441551

Visualize the current heads of our peers in grafana. 

### How was it fixed?
Added a `PeerReporter` class and `PeerReporterRegistry` to track the current head for each connected peer. I'm still quite unfamiliar with the trinity/p2p codebase - so if there's a cleaner  solution or a better place to put this code, i'm happy to make any changes.

Here are some screenshots of the metrics reporting locally. Though these were captured using a `_reporting_frequency` of `15` rather than the default `60`. I'd be curious if anybody has thoughts on the frequency - and whether it should be left to `60` (which I did in this pr) to match the frequency with which peer connection info is logged to the terminal - or perhaps if it should be increased? 

Screen Shot 2020-04-27 at 9 41 57 AM

Screen Shot 2020-04-27 at 9 36 39 AM

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/p2p/metrics.py
+++ b/p2p/metrics.py
@@ -1,0 +1,62 @@
+from typing import Dict, Generic
+from pyformance import MetricsRegistry
+
+from p2p.abc import TPeer
+from p2p.exceptions import PeerReporterRegistryError
+
+
+class PeerReporterRegistry(Generic[TPeer]):
+    """
+    Registry to track active peers and report metrics to influxdb/grafana.
+    """
+    def __init__(self, metrics_registry: MetricsRegistry) -> None:
+        self.metrics_registry = metrics_registry
+        self._peer_reporters: Dict[TPeer, int] = {}
+
+    def assign_peer_reporter(self, peer: TPeer) -> None:
+        if peer in self._peer_reporters.keys():
+            raise PeerReporterRegistryError(
+                f"Cannot reassign peer: {peer} to PeerReporterRegistry. "
+                "Peer already assigned."
+            )
+        min_unclaimed_id = self._get_min_unclaimed_gauge_id()
+        self._peer_reporters[peer] = min_unclaimed_id
+        self.reset_peer_meters(min_unclaimed_id)
+
+    def unassign_peer_reporter(self, peer: TPeer) -> None:
+        try:
+            gauge_id = self._peer_reporters[peer]
+        except AttributeError:
+            raise PeerReporterRegistryError(
+                f"Cannot unassign peer: {peer} to PeerReporterRegistry. "
+                "Peer not found in peer reporter registry."
+            )
+        self.reset_peer_meters(gauge_id)
+        del self._peer_reporters[peer]
+
+    def _get_min_unclaimed_gauge_id(self) -> int:
+        try:
+            max_claimed_id = max(self._peer_reporters.values())
+        except ValueError:
+            return 0
+
+        unclaimed_ids = [
+            num for num in range(0, max_claimed_id + 1) if num not in self._peer_reporters.values()
+        ]
+        if len(unclaimed_ids) == 0:
+            return max_claimed_id + 1
+
+        return min(unclaimed_ids)
+
+    def trigger_peer_reports(self) -> None:
+        for peer, peer_id in self._peer_reporters.items():
+            if peer.get_manager().is_running:
+                self.make_periodic_update(peer, peer_id)
+
+    def reset_peer_meters(self, peer_id: int) -> None:
+        # subclasses are responsible for implementing method that resets all implemented meters
+        pass
+
+    def make_periodic_update(self, peer: TPeer, peer_id: int) -> None:
+        # subclasses are responsible for implementing method that updates all implemented meters
+        pass
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -3,6 +3,7 @@ import asyncio
 import functools
 import operator
 from typing import (
+    Any,
     AsyncContextManager,
     AsyncIterator,
     AsyncIterable,
@@ -57,6 +58,7 @@ from p2p.exceptions import (
     UnknownAPI,
     UnreachablePeer,
 )
+from p2p.metrics import PeerReporterRegistry
 from p2p.peer import (
     BasePeer,
     BasePeerFactory,
@@ -102,11 +104,13 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
     """
     PeerPool maintains connections to up-to max_peers on a given network.
     """
-    _report_interval = 60
+    _report_stats_interval = 60
+    _report_metrics_interval = 15  # for influxdb/grafana reporting
     _peer_boot_timeout = DEFAULT_PEER_BOOT_TIMEOUT
     _event_bus: EndpointAPI = None
 
     _handshake_locks: ResourceLock[NodeAPI]
+    peer_reporter_registry_class: Type[PeerReporterRegistry[Any]] = PeerReporterRegistry[BasePeer]
 
     def __init__(self,
                  privkey: datatypes.PrivateKey,
@@ -131,6 +135,7 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
             # This is so that we don't need to pass a MetricsRegistry in tests and mocked pools.
             metrics_registry = MetricsRegistry()
         self._active_peer_counter = metrics_registry.counter('trinity.p2p/peers.counter')
+        self._peer_reporter_registry = self.get_peer_reporter_registry(metrics_registry)
 
         # Restricts the number of concurrent connection attempts can be made
         self._connection_attempt_lock = asyncio.BoundedSemaphore(MAX_CONCURRENT_CONNECTION_ATTEMPTS)
@@ -230,6 +235,11 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
             event_bus=self._event_bus,
         )
 
+    def get_peer_reporter_registry(
+        self, metrics_registry: MetricsRegistry
+    ) -> PeerReporterRegistry[BasePeer]:
+        return self.peer_reporter_registry_class(metrics_registry)
+
     async def get_protocol_capabilities(self) -> Tuple[Tuple[str, int], ...]:
         return tuple(
             (handshaker.protocol_class.name, handshaker.protocol_class.version)
@@ -303,6 +313,7 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
         logger("Adding %s to pool", peer)
         self.connected_nodes[peer.session] = peer
         self._active_peer_counter.inc()
+        self._peer_reporter_registry.assign_peer_reporter(peer)
         peer.add_finished_callback(self._peer_finished)
         for subscriber in self._subscribers:
             subscriber.register_peer(peer)
@@ -319,6 +330,7 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
             self.manager.run_daemon_task(self.maybe_connect_more_peers)
 
         self.manager.run_daemon_task(self._periodically_report_stats)
+        self.manager.run_daemon_task(self._periodically_report_metrics)
         await self.manager.wait_finished()
 
     async def connect(self, remote: NodeAPI) -> BasePeer:
@@ -477,6 +489,7 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
             subscriber.deregister_peer(peer)
 
         self._active_peer_counter.dec()
+        self._peer_reporter_registry.unassign_peer_reporter(peer)
 
     async def __aiter__(self) -> AsyncIterator[BasePeer]:
         for peer in tuple(self.connected_nodes.values()):
@@ -485,6 +498,11 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
             await asyncio.sleep(0)
             if peer.get_manager().is_running and not peer.is_closing:
                 yield peer
+
+    async def _periodically_report_metrics(self) -> None:
+        while self.manager.is_running:
+            self._peer_reporter_registry.trigger_peer_reports()
+            await asyncio.sleep(self._report_metrics_interval)
 
     async def _periodically_report_stats(self) -> None:
         while self.manager.is_running:
@@ -525,4 +543,4 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
                     self.logger.debug("    Failure during stats lookup: %r", exc)
 
             self.logger.debug("== End peer details == ")
-            await asyncio.sleep(self._report_interval)
+            await asyncio.sleep(self._report_stats_interval)
--- a/trinity/protocol/common/peer.py
+++ b/trinity/protocol/common/peer.py
@@ -16,6 +16,7 @@ from async_service import Service
 from cached_property import cached_property
 
 from lahja import EndpointAPI
+from pyformance.meters import SimpleGauge
 
 from cancel_token import OperationCancelled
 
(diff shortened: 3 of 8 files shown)


pipermerriam:
Those are some nice looking graphs.  :heart: 

No opinions about the reporting interval.  I think @cburgdorf is probably the person with the most context on this subject.

cburgdorf:
I think `15` seconds is a good default for this so that it more or less tracks the block times and brings the monitoring closer to real time. We track system resources (CPU, memory, disk) in 3 second intervals (same as geth) so I think 15 seconds for block metrics should be fine.

p2p/metrics.py
```diff
@@ -0,0 +1,44 @@
...
+        self.gauge: SimpleGauge = gauge
+
+
+class PeerReporterRegistry:
+    def __init__(self, metrics_registry: MetricsRegistry, max_peers: int) -> None:
+        self._max_peers = max_peers
+        self._peer_reporter_store: List[PeerReporter] = []
+        self._peer_reporters: Dict[str, PeerReporter] = {}
+        for peer_slot in range(0, self._max_peers):
+            gauge = metrics_registry.gauge(f'trinity.p2p/peer{peer_slot}.gauge')
```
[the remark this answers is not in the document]
cburgdorf: But maybe, one thing we could do is to reset the value to `0` whenever the gauge is reassigned to another peer. That would help to identify in the chart that the gauge has lost the relation to its previous peer.

p2p/metrics.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class PeerReporter:
+    def __init__(self, claimed: bool, gauge: SimpleGauge) -> None:
+        self.claimed: bool = claimed
+        self.gauge: SimpleGauge = gauge
+
+
+class PeerReporterRegistry:
+    def __init__(self, metrics_registry: MetricsRegistry, max_peers: int) -> None:
```
cburgdorf, line 15: I think we might be better off if we don't couple this against `max_peers` and just let it grow dynamically with the pool size.

What do you think about this:

Instead of pre-creating all gauges and keeping them in `self._peer_reporter_store`, we just create them on demand whenever `assign_peer_reporter` needs one.

I think what we could do is to hold a `Dict[BasePeer, int]` that just maps peers to the number of the gauge. And whenever a number for a gauge is needed we call a method on `PeerReporterRegistry` such as `get_new_gauge_id` which loops through the **sorted** gauge ids of the dictionary and returns the first free gap that it finds. If it does not find a free gap it will simply return the highest number + 1.

And then we can simply get the gauge on demand with `metrics_registry.gauge(f'trinity.p2p/peer{peer_slot}.gauge')` as before.

Performance wise, this isn't so much different from the current solution where we loop through the fixed-length list of peer reporters to find an unclaimed one. But performance shouldn't worry us too much here anyway as the set of peers is rather small and the code only runs when peer join or drop.

The benefit I see is:

- operate free from a max peer concept
- do not pre-allocate objects in memory when they aren't needed (e.g. if we run this with `--max-peers 1000` it would pre-allocate 1000 `PeerReporter` no matter if those end up needed or not.
- reduced code complexity.

p2p/metrics.py
```diff
@@ -0,0 +1,44 @@
...
+        unclaimed_gauge.claimed = True
+
+    def unassign_peer_reporter(self, peer: BasePeer) -> None:
+        if str(peer) not in self._peer_reporters:
+            raise AttributeError(f"Peer: {peer} not found in peer reporter registry.")
+        gauge = self._peer_reporters[str(peer)]
+        del self._peer_reporters[str(peer)]
+        gauge.claimed = False
+
+    def update_peer_head(self, peer: BasePeer) -> None:
```
cburgdorf, line 38: We can make this a bit more generic and get rid of the `# type: ignore`.

1. Instead of naming this `update_peer_head`, we might give it a name such as `make_periodic_update`

2. Instead of using `BasePeer` as the type, we use `TPeer` defined as `TPeer = TypeVar('TPeer', bound=BasePeer)`

3. Next, we rip this whole class `PeerReporterRegistry` apart to become two: One base class that the peer pool can use as a default. But the implementation of `make_periodic_update` will just be a `pass`. But then there's another one such as `ChainPeerReporterRegistry` that is derived from `PeerReporterRegistry` and overwrites the method appropriately. And since it will be operating on `BaseChainPeer` it will have access to the `head_info`. But this can be made even more specific to cater to specific needs of e.g. the `ETHPeerPool`.

4. The `PeerPool` reads the specific class that it should instantiate from `peer_reporter_registry_class`. You can see the same concept already in play for different things.

https://github.com/ethereum/trinity/blob/e7960639d9701d7cb2ed8e79ea22b2409dda8523/trinity/protocol/eth/peer.py#L332-L333

p2p/metrics.py
njgheorghita: Yeah, I definitely like your solution better. The reason I coupled my code to `max_peers` so closely was because it seemed simpler from the grafana end of things. In grafana, I have 25 queries (one for each of the gauges) - and each one is filtered to show if its value is `> 0`. 

Though, I guess this strategy also breaks anytime `max_peers` of more than 25 is used...

It would be really nice if we can use your solution and transmit the data to grafana so that grafana can dynamically register/display peer gauges as they are created (even above 25) - and not simply instantiate them all and filter when their value is `> 0`. But, from what I can tell with the pyformance registry - the best option available is a single gauge / peer? 

But, I'd be happy to update this code to your solution - and if anybody wants to report metrics with more than 25 peers - they'll be responsible for adding those queries to grafana themselves. And since `trinity.p2p/peer{peer_slot}.gauge` is deterministic, instantiating all 25 queries in grafana but dynamically creating the gauges in trinity should still work fine.

p2p/metrics.py
njgheorghita: Quick question, does this imply that I should implement a `peer_reporter_registry_class` on all the various peer pools? ie. `LESPeerPool`, `ParagonPeerPool`, etc. ?

cburgdorf:
They could just choose not to set it and go with what the BasePeerPool
provides as a base class.

p2p/metrics.py
cburgdorf: Expanding on my brief fat finger mobile answer:

You can set `peer_reporter_registry_class` in `BasePeerPool` to `PeerReporterRegistry` directly which means `LESPeerPool` (which is on the path to removal anyway) as well as `ParagonPeerPool` do not have to set anything further. However, `ETHPeerPool` (or maybe `BaseChainPeerPool` :thinking: ) would set the `peer_reporter_registry_class` to something different.

njgheorghita:
@cburgdorf Thanks for your help with this -  ready for :eyes: when you get the chance. Also - if you have any tips for the mypy error in the failing lint ci build, I'm all ears. I tried using generics to solve the problem - but was never able to land on a solid solution...

p2p/metrics.py
```diff
@@ -0,0 +1,61 @@
...
+    Registry to track active peers and report metrics to influxdb/grafana.
+    """
+    def __init__(self, metrics_registry: MetricsRegistry) -> None:
+        self.metrics_registry = metrics_registry
+        self._peer_reporters: Dict[BasePeer, int] = {}
+
+    def assign_peer_reporter(self, peer: TPeer) -> None:
+        if peer in self._peer_reporters.keys():
+            raise PeerReporterRegistryError(
+                "Cannot reassign peer: {str(peer)} to PeerReporterRegistry. "
```
cburgdorf, line 21: If you write `{ peer }` it will already call `str(peer)` behind the scenes so the `str(...)` can be omitted.

trinity/protocol/common/peer.py
```diff
@@ -152,6 +156,24 @@ async def disconnect(self, reason: DisconnectReason) -> None:
         await self.manager.stop()
 
 
+class BaseChainPeerReporterRegistry(PeerReporterRegistry[BaseChainPeer]):
+    def make_periodic_update(self, peer: BaseChainPeer) -> None:
+        try:
+            gauge_id = self._peer_reporters[peer]
```
cburgdorf, line 162: I think we need to re-think something here. Derived classes such as `BaseChainPeerReporterRegistry` should not reach into private properties of the `PeerReporterRegistry` such as `_peer_reporters`. But more importantly, I think we can tame down the responsibility of this method. E.g. it's unfortunate that it even has to worry about the `AttributeError`.

At the same time it is unfortunate that the `BasePeerPool` has to loop through all peers in `_periodically_report_metrics` even though the `PeeReporterRegistry` has knowledge of all the peers itself!

What I propose goes as follows:

1. Inside the `BasePeerPool` in `_periodically_report_metrics` we call `trigger_peer_reports` on `PeerReporterRegistry`

2. `trigger_peer_reports` will take care of looping through all registered peers and looking up their `peer_id`.

3. While it is looping through all the peers it will call `make_periodic_update(peer, peer_id)` for every peer.

This means that `make_periodic_update` will already have access to both the `peer` and the `peer_id` and doesn't have to worry about retrieving the `peer_id` anymore.

I further suggest that it then takes the `peer_id` to name gauges, counter etc as it makes most sense for the reported item (after all, we might be reporting different things per surrounding peer)

- self.metrics_registry.gauge(`trinity.p2p/peer_{peer_id}_blockheight.gauge`)
- self.metrics_registry.gauge(`trinity.p2p/peer_{peer_id}_total_difficulty.gauge`)
...

Related to this change, I think that `assign_peer_reporter` and `unassign_peer_reporter` should not directly try to reset a specific gauge but rather call a method `reset_peer_meters(peer, peer_id)`. Then it is left to derived classes to reset all the different meters they may have created (e.g. `peer_2_blockheight.gauge`, `peer_2_total_difficulty.gauge` etc).

What do you think?

p2p/metrics.py
```diff
@@ -0,0 +1,61 @@
...
+        unclaimed_ids = [
+            num for num in range(0, max_claimed_id + 1) if num not in self._peer_reporters.values()]
+
+        if len(unclaimed_ids) == 0:
+            return max_claimed_id + 1
+
+        return min(unclaimed_ids)
+
+    def _get_tag_for_peer_gauge(self, number: int) -> str:
+        return f"trinity.p2p/peer{number}.gauge"
```
cburgdorf, line 45: See my other comment why I think this is problematic. In short, this doesn't tell us *what* it is what we are tracking and we might want to track different things from our surrounding peers (blockheight, total difficulty etc)

p2p/metrics.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def assign_peer_reporter(self, peer: TPeer) -> None:
+        if peer in self._peer_reporters.keys():
+            raise PeerReporterRegistryError(
+                "Cannot reassign peer: {str(peer)} to PeerReporterRegistry. "
+                "Peer already assigned."
+            )
+        min_unclaimed_id = self._get_min_unclaimed_gauge_id()
+        self._peer_reporters[peer] = min_unclaimed_id
+        tag = self._get_tag_for_peer_gauge(min_unclaimed_id)
```
cburgdorf, line 26: I think just assigning the `peer_id` is all we should do in this method and then call `reset_peer_meters(peer, peer_id)` which might be implemented in subclasses to set up multiple gauges, counters etc per peer.

p2p/metrics.py
```diff
@@ -0,0 +1,61 @@
...
+        try:
+            gauge_id = self._peer_reporters[peer]
+        except AttributeError:
+            raise PeerReporterRegistryError(
+                "Cannot unassign peer: {str(peer)} to PeerReporterRegistry. "
+                "Peer not found in peer reporter registry."
+            )
+        tag = self._get_tag_for_peer_gauge(gauge_id)
+        gauge = self.metrics_registry.gauge(tag)
+        gauge.set_value(0)
```
cburgdorf, line 57: Similar to `assign_peer_reporter` I think this just just remove the peer from the mapping and call `reset_peer_meters(peer, peer_id)` to leave it up to others to reset gauges, counter etc.

p2p/peer_pool.py
```diff
@@ -486,6 +498,15 @@ async def __aiter__(self) -> AsyncIterator[BasePeer]:
             if peer.get_manager().is_running and not peer.is_closing:
                 yield peer
 
+    async def _periodically_report_metrics(self) -> None:
+        while self.manager.is_running:
+            peers = tuple(self.connected_nodes.values())
+            for peer in peers:
```
cburgdorf, line 504: Because the registry does now about the peers, this could just become a call to `_peer_reporter_registry.trigger_peer_reports()` (see my longer comment)

trinity/protocol/common/peer.py
```diff
@@ -152,6 +156,24 @@ async def disconnect(self, reason: DisconnectReason) -> None:
         await self.manager.stop()
 
 
+class BaseChainPeerReporterRegistry(PeerReporterRegistry[BaseChainPeer]):
```
cburgdorf, line 159: The `mypy` error is saying that you are trying to provide a generic type parameter (`[BaseChainPeer]`) but `PeerReporterRegistry` is not generic. The class definition needs to be `class PeerReporterRegistry(Generic[TPeer]):` for that to work.

trinity/protocol/common/peer.py
njgheorghita: Great, thanks for the write-up!! This is a much cleaner setup. I updated the code to follow this - and it's ready for another :eyes:. I also added stats to track the `total_difficulty` for peers. (displaying the `block_hash` didn't seem that useful to me, since we already have the block number .. but I could be wrong on that). Just one last question about the types.. I couldn't figure out how to set the correct type for `PeerPool.peer_reporter_registry_class`... any ideas?

Screen Shot 2020-05-05 at 2 29 10 PM

p2p/metrics.py
```diff
@@ -16,7 +16,7 @@ def __init__(self, metrics_registry: MetricsRegistry) -> None:
     def assign_peer_reporter(self, peer: TPeer) -> None:
         if peer in self._peer_reporters.keys():
             raise PeerReporterRegistryError(
-                "Cannot reassign peer: {peer} to PeerReporterRegistry. "
+                f"Cannot reassign peer: {peer} to PeerReporterRegistry. "
```
cburgdorf, line 19: Without the `f` before the `"` the interpolation won't be recognized as such and will simply print out `Cannot reassign peer: {peer} to PeerReporterRegistry.` as the final text.

p2p/peer_pool.py
```diff
@@ -109,7 +110,7 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
     _event_bus: EndpointAPI = None
 
     _handshake_locks: ResourceLock[NodeAPI]
-    peer_reporter_registry_class: Type[PeerReporterRegistry] = PeerReporterRegistry
+    peer_reporter_registry_class: Type[PeerReporterRegistry[Any]] = PeerReporterRegistry[BasePeer]
```
cburgdorf, line 113: It was complaining because you can not refer to a generic type without specifying the type parameters. Now, the way that I fixed it is a little lame because we bind to `Type[PeerReporterRegistry[Any]]` which means:

1. We care about it being a type
2. We care about it being a type or subtype of `PeerReporterRegistry`
3. But we don't care about the type it is acting on.

It would be much better if that was scoped to `TPeer` but that would mean, we have to make the whole `BasePeerPool` generic over `TPeer` which would be a good thing (the peer pool is among the oldest places in the code base, it would be generic over `TPeer` if someone had written it today) but out of scope for this PR I think.

trinity/protocol/common/peer.py
cburgdorf: Cool! Yeah, I think the block hashes don't convey much meaning on their own. I also pushed two more commits to fix the type hints and the f-strings. I think this is ready to merge (preferable with squashing everything into a single commit)

p2p/metrics.py
njgheorghita: ahhh good catch, yeah this was just a brain fart on my part. it would be nice if linters could catch this... but from what i've read there's no mention of supporting such a feature in flake8 and it's still an [open issue](https://github.com/PyCQA/pylint/issues/2507) in pylint 😢

Issue #1188, opened 2019-09-19
BrokenPipe bubbling out of database manager on shutdown.


pipermerriam:
### What is wrong?
Our DBManager process still runs on the same process group as the main process, so when there's a Ctrl-C in the terminal that process is likely to terminate before we send a SIGINT to our components' processes, leading to BrokenPipeErrors like below

```
   ERROR  2019-09-19 16:07:34,941        BeamDownloader  Unexpected error in <trinity.sync.beam.state.BeamDownloader object at 0x7f4ae1aea4e0>, exiting
Traceback (most recent call last):
  File "/home/piper/projects/trinity/p2p/service.py", line 118, in run
    await self._run()
  File "/home/piper/projects/trinity/trinity/sync/beam/state.py", line 516, in _run
    await self.wait(self._match_node_requests_to_peers())
  File "/home/piper/projects/trinity/p2p/cancellable.py", line 20, in wait
    return await self.wait_first(awaitable, token=token, timeout=timeout)
  File "/home/piper/projects/trinity/p2p/cancellable.py", line 43, in wait_first
    return await token_chain.cancellable_wait(*awaitables, timeout=timeout)
  File "/home/piper/python-environments/trinity/lib/python3.6/site-packages/cancel_token/token.py", line 174, in cancellable_wait
    task.result()
  File "/home/piper/projects/trinity/trinity/sync/beam/state.py", line 333, in _match_node_requests_to_peers
    predictive_batch_id,
  File "/home/piper/projects/trinity/trinity/sync/beam/state.py", line 396, in _get_nodes_from_peer
    batch[node_hash] = node
  File "/home/piper/.pyenv/versions/3.6.5/lib/python3.6/contextlib.py", line 88, in __exit__
    next(self.gen)
  File "/home/piper/projects/trinity/trinity/db/manager.py", line 446, in atomic_batch
    ATOMIC_BATCH.value + kv_pair_count_and_size_data + kv_and_delete_data
BrokenPipeError: [Errno 32] Broken pipe
```

### How can it be fixed

Not sure.  There's no clear place that I see that handling this makes sense.  Ideally what we want to do is have the `DBManager` shutdown after all of the other processes have shutdown, or to have it give all of the connected clients some amount of time to disconnect before shutting down.

gsalgado:
[...]

PR #1701, opened 2020-04-28
Use asyncio-run-in-process to run the DBManager

gsalgado:
Closes: #1188

### DIFF
--- a/trinity/bootstrap.py
+++ b/trinity/bootstrap.py
@@ -9,16 +9,19 @@ import signal
 from typing import (
     Callable,
     Dict,
-    Iterable,
     Sequence,
     Tuple,
     Type,
 )
 
+from asyncio_run_in_process import open_in_process
 from async_service import AsyncioManager
-from eth_utils import ValidationError
-from eth_utils.toolz import curry
 
+from eth.db.backends.level import LevelDB
+
+from eth_utils import ValidationError
+
+from trinity.db.manager import DBManager
 from trinity.exceptions import (
     AmbigiousFileSystem,
     MissingPath,
@@ -45,10 +48,11 @@ from trinity.network_configurations import (
     PRECONFIGURED_NETWORKS,
 )
 from trinity._utils.ipc import (
-    kill_process_gracefully,
     remove_dangling_ipc_files,
+    wait_for_ipc,
 )
 from trinity._utils.logging import (
+    child_process_logging,
     enable_warnings_by_default,
     set_logger_levels,
     setup_file_logging,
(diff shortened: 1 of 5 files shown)


pipermerriam:
What is keeping us from going ahead and getting this running in its own process (as its own process leader)?

gsalgado:
> What is keeping us from going ahead and getting this running in its own process (as its own process leader)?

I recently tried to do that for `TrioIsolatedComponent` (which uses `multiprocessing` like the `DBManager`), and realized it'd take a lot of work because we can't use `asyncio-run-in-process` there, but I guess we could turn `trinity.main.run_database_process()` into an async function so that we can use `asyncio-run-in-process` to run the `DBManager`, and that way we can make it run on a separate process group?

pipermerriam:
Are you aware of https://github.com/ethereum/trio-run-in-process

It's less mature than `asyncio-run-in-process` but should be in a state that is viable to start off with.

gsalgado:
I had seen that but was under the impression it was still in an embryonic state, don't know why. Have you tried using it on `TrinityIsolatedComponent`?

Anyway, do you think it makes sense to wrap the `DBManager` runner into an async API just so that we can use `asyncio-run-in-process`?

gsalgado:
It seemed to me like introducing the concept of dependency between components to the ComponentManager would be a significant undertaking, with the possibility of opening up a few more cans of worms, so I decided to just try to use asyncio-run-in-process (and refactor a few things) to run the `DBManager` process. I've managed to get it working, but now we're getting an asyncio warning because asyncio-run-in-process seems to be ignoring the exception of a task, even though it attempts to wait for it twice (as we can see in the RuntimError that is the task's exception that is being ignored);

```
    INFO  2020-04-29 14:48:31,605             DBManager  Got KeyboardInterrupt, returning
Task exception was never retrieved
future: <Task finished coro=<run_db_manager() done, defined at /home/salgado/src/trinity/trinity/bootstrap.py:351> exception=RuntimeError('cannot reuse already awaited coroutine')>
RuntimeError: cannot reuse already awaited coroutine
```

This should not really come as a surprise as we still haven't figured out what's going on in https://github.com/ethereum/asyncio-run-in-process/pull/12, and `asyncio-run-in-process` has some code that awaits for a coroutine multiple times (https://github.com/ethereum/asyncio-run-in-process/blob/87e3cf4303790bcf7a965c372999f81ec585598f/asyncio_run_in_process/_child.py#L94-L101), which asyncio doesn't seem to like

I'll get back to it tomorrow, but thought I'd post this update just in case

pipermerriam:
> It seemed to me like introducing the concept of dependency between components to the ComponentManager would be a significant undertaking, with the possibility of opening up a few more cans of worms

Fair, but I'm still somewhat convinced that this is the direction we need to move towards.  Otherwise, we need a different mechanism to solve the fundamental base problem which is:

- we have components that depend on each other.
- if `A` depends on functionality of `B` then:
    - `A` should not be started until `B` starts
    - `B` should not be stopped until `A` is stopped

In my head the structure is simple and the complexity is not that different from the current complexity since a component that doesn't shutdown cleanly today will also not shutdown cleanly in this new dependency based model as the shutdown mechanism shouldn't be that different, only the order in which the shutdowns are issued.

The optimist in me thinks that it may actually result in things being cleaner since it institutes more control and order to the lifecycles and may eliminate weird bugs like when the `DBManager` shuts down before the other components which are still using it.  In the new model this type of error shouldn't occur.

pipermerriam:
I'll point out that we could in theory solve this problem at a more granular level and figure out how to have the `DBManager` wait for all of it's connections to close before shutting down, but I feel like this approach is more likely to result in more unclean shutdowns since it means that some components would have to *block* shutdown while waiting for things like this and it's not unreasonable to expect us to accidentally bake in some sort of accidental circular dependency where `A` waits on `B` which waits on `C` which waits on `A`, all via implicit mechanisms within the component that aren't obviously visible.

No need to prove me wrong or argue against the complexity.  I'm very ok with you taking whatever path forward you feel is best and most actionable.

gsalgado:
This is working fine now, and I've also refactored a few things to get rid of some code duplication. BTW, since we're now using `asyncio_run_in_process` to run the DBManager *and* the only place that deals with the DB manager/process is the new `_run()` method (https://github.com/ethereum/trinity/pull/1701/files#diff-5c2d4b3cd13a63d555a9b179b7bc741dR244), it would actually be easier to turn it into an isolated component and use the `ComponentManager` to run it, if we want to go that way

Issue #1683, opened 2020-04-20
Connection leaks BrokenPipeError, leading to shutdown


gsalgado:
I'm getting frequent shutdowns because of those. Guess we just need to handle the BrokenPipeError when closing a transport?

```
   DEBUG  2020-04-20 07:50:15,998               Manager  <Manager[FullNode] flags=SRcfe>: task FullServer[daemon=True] exited with error: [Errno 32] Broken pipe
Traceback (most recent call last):
  File "/home/salgado/src/async-service/async_service/base.py", line 301, in _run_and_manage_task
    await task.run()
  File "/home/salgado/src/async-service/async_service/base.py", line 166, in run
    await self.child_manager.run()
  File "/home/salgado/src/async-service/async_service/asyncio.py", line 217, in run
    for _, exc_value, exc_tb in self._errors
  File "/home/salgado/src/async-service/async_service/base.py", line 301, in _run_and_manage_task
    await task.run()
  File "/home/salgado/src/async-service/async_service/base.py", line 166, in run
    await self.child_manager.run()
  File "/home/salgado/src/async-service/async_service/asyncio.py", line 217, in run
    for _, exc_value, exc_tb in self._errors
  File "/home/salgado/src/async-service/async_service/base.py", line 301, in _run_and_manage_task
    await task.run()
  File "/home/salgado/src/async-service/async_service/base.py", line 166, in run
    await self.child_manager.run()
  File "/home/salgado/src/async-service/async_service/asyncio.py", line 217, in run
    for _, exc_value, exc_tb in self._errors
  File "/home/salgado/src/async-service/async_service/base.py", line 301, in _run_and_manage_task
    await task.run()
  File "/home/salgado/src/async-service/async_service/asyncio.py", line 35, in run
    await self._async_fn(*self._async_fn_args)
  File "/home/salgado/src/trinity/p2p/connection.py", line 154, in run
    await self._multiplexer.close()
  File "/home/salgado/src/trinity/p2p/multiplexer.py", line 185, in close
    await self._transport.close()
  File "/home/salgado/src/trinity/p2p/transport.py", line 333, in close
    await self._writer.drain()
  File "/usr/lib/python3.7/asyncio/streams.py", line 339, in drain
    raise exc
BrokenPipeError: [Errno 32] Broken pipe
```

gsalgado:
And here's a nearly identical traceback, but with a ConnectionResetError instead. Might make sense to fix both at the same time

``` 
[...]

PR #1702, opened 2020-04-28
Do not leak BrokenPipeError/ConnectionResetError when closing transport

gsalgado:
Our Transport.close() method will first try to drain the writer, but
that may raise BrokenPipeError/ConnectionResetError (just like when
reading from the transport's reader), so we need to catch them to
ensure we still close the writer.

Closes: #1683

### DIFF
--- a/p2p/transport.py
+++ b/p2p/transport.py
@@ -330,7 +330,10 @@ class Transport(TransportAPI):
 
         This will cause the peer to stop in case it is running.
         """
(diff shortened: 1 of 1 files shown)


gsalgado:
Well, this is specifically for the BrokenPipeErrors when dealing with peers. The BrokenPipeErrors from the IPCLogger, on shutdown (#1381), are still going to be with us, though :-(

PR #1703, opened 2020-04-28
Blacklist peers that send us a malformed message during handshake

gsalgado:
I am getting malformed messages during handshake with one specific peer all the time, and since we
were not recording failures for those in  the connection tracker, the PeerPool would keep
attempting to connect to that peer. Now we blacklist those for 1h

```
Got malformed response from <Node(0xf974a7082d8bdeb3753fbe2c2f705c419bc762a699860e856f8bacbc2a3185aa94f005da76cd1217285e82fd3af8abbe3ec48d38398339245db5e27409c1e0f7@104.248.49.16:30303)> during handshake
```

### DIFF
--- a/p2p/tracking/connection.py
+++ b/p2p/tracking/connection.py
@@ -29,7 +29,7 @@ def register_error(exception: Type[BaseP2PError], timeout_seconds: int) -> None:
 register_error(HandshakeFailure, 10)  # 10 seconds
 register_error(HandshakeFailureTooManyPeers, 60)  # one minute
 # A MalformedMessage is usually not a transient issue, so blacklist the remote for a long time.
-register_error(MalformedMessage, 3600)  # 1 hour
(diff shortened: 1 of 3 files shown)


p2p/tracking/connection.py
```diff
@@ -27,6 +28,8 @@ def register_error(exception: Type[BaseP2PError], timeout_seconds: int) -> None:
 
 register_error(HandshakeFailure, 10)  # 10 seconds
 register_error(HandshakeFailureTooManyPeers, 60)  # one minute
+# A MalformedMessage is usually not a transient issue, so blacklist the remote for a long time.
+register_error(MalformedMessage, 3600)  # 1 hour
```
carver, line 32: Guidance from geth was to be very cautious about blacklisting, because it's so easy to accidentally cause a network split between clients. Maybe reduce to about 10 minutes? That should still remove most of the garbage from the logs.

PR #1696, opened 2020-04-25
Add validator api server to beacon node

ralexstokes:
Supersedes #1645.

Adds the trio-powered HTTP JSON API server to the beacon node to facilitate the validator API.

To demo the server, we install the "client version" API endpoint.

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #1705, opened 2020-04-29
Add remainder of minimum validator APIs

ralexstokes:
There is a minimum set of validator APIs we need to build a (minimum) beacon chain: blocks and attestations.

This API builds on top of #1696 to add the rest of the APIs not introduced in that PR.

#### Cute Animal Picture


eth2/api/http/validator.py
```diff
@@ -17,17 +55,209 @@ class Paths(Enum):
...
+class SyncStatus:
+    is_syncing: bool
+    starting_slot: Slot
+    current_slot: Slot
+    highest_slot: Slot
+
+
+class SyncerAPI(ABC):
+    async def get_status(self) -> SyncStatus:
+        ...
```
ralexstokes, line 68: this likely lives in a place w/ the syncing logic but we keep it here for now...

eth2/validator_client/beacon_node.py
```diff
@@ -53,9 +60,9 @@ async def _get_duties_from_beacon_node(
         await session.get(
             url,
             params={
-                "validator_pubkeys": [
+                "validator_pubkeys": ",".join(
```
ralexstokes, line 63: `asks` was not serializing this correctly so we update to what we expect on the beacon node's side

eth2/validator_client/signatory.py
```diff
@@ -68,7 +68,7 @@ async def sign_and_broadcast_operation_if_valid(
     try:
         await _validate_duty(duty, operation, signature_store)
     except ValidationError as e:
-        logger.warn("a duty %s was not valid: %s", duty, e)
+        logger.warning("a duty %s was not valid: %s", duty, e)
```
ralexstokes, line 71: fixes `DeprecationWarning`

trinity/components/registry.py
```diff
@@ -49,7 +49,6 @@
 from trinity.components.builtin.upnp.component import (
     UpnpComponent,
 )
-from trinity.components.eth2.beacon.component import BeaconNodeComponent
```
ralexstokes, line 52: finally hit circular import errors across the trio vs asyncio beacon node so we (temporarily?) disable for now

trinity/nodes/beacon/full.py
```diff
@@ -35,6 +37,14 @@ def _mk_clock(
     )
 
 
+def _mk_syncer() -> SyncerAPI:
+    class _sync(SyncerAPI):
```
ralexstokes, line 41: this function would build the object in charge of syncing; we create a mock object for now

PR #1707, opened 2020-04-29
Add --sync-mode header + a few other goodies

cburgdorf:
### What was wrong?

Once upon a time...eh no, let's skip this part.

### How was it fixed?

1. This adds support for `--sync-mode header` to only sync the header chain on the ETH network. Notice that this is almost what `--sync-mode light` does but a.) it does it on the ETH network, not LES and b.) light sync is on the road to removal but it is nice to keep this feature. This does also support launching header sync from a checkpoint via `--sync-from checkpoint <uri>`

2. `--beam-from-checkpoint` is now `--sync-from-checkpoint` to not have two different names for the exact same concept

3. Sync related flags such as `--beam-from-checkpoint` or `--force-beam-block-number` are now grouped under `sync-mode` in `--trinity-help`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/components/builtin/syncer/component.py
+++ b/trinity/components/builtin/syncer/component.py
@@ -5,6 +5,7 @@ from abc import (
 from argparse import (
     ArgumentParser,
     Namespace,
+    _ArgumentGroup,
     _SubParsersAction,
 )
 import logging
@@ -69,7 +70,7 @@ from trinity.sync.light.chain import (
 from trinity.components.builtin.syncer.cli import NormalizeCheckpointURI
 
 
-def add_sync_from_checkpoint_arg(arg_parser: ArgumentParser):
+def add_sync_from_checkpoint_arg(arg_parser: _ArgumentGroup) -> None:
     arg_parser.add_argument(
         '--sync-from-checkpoint',
         action=NormalizeCheckpointURI,
@@ -80,6 +81,7 @@ def add_sync_from_checkpoint_arg(arg_parser: ArgumentParser):
         ),
         default=None,
     )
+
 
 class BaseSyncStrategy(ABC):
 
@@ -92,7 +94,7 @@ class BaseSyncStrategy(ABC):
         return True
 
     @classmethod
-    def configure_parser(cls, arg_parser: ArgumentParser) -> None:
+    def configure_parser(cls, arg_group: _ArgumentGroup) -> None:
         """
         Configure the argument parser for the specific sync strategy.
         """
@@ -166,15 +168,15 @@ class BeamSyncStrategy(BaseSyncStrategy):
         return SYNC_BEAM
 
     @classmethod
-    def configure_parser(cls, arg_parser: ArgumentParser) -> None:
-        arg_parser.add_argument(
+    def configure_parser(cls, arg_group: _ArgumentGroup) -> None:
+        arg_group.add_argument(
             '--force-beam-block-number',
             type=int,
             help="Force beam sync to activate on a specific block number (for testing)",
             default=None,
         )
 
-        add_sync_from_checkpoint_arg(arg_parser)
+        add_sync_from_checkpoint_arg(arg_group)
 
     async def sync(self,
                    args: Namespace,
@@ -246,9 +248,6 @@ class SyncerComponent(AsyncioIsolatedComponent):
         if type(cls.default_strategy) not in cls.extract_strategy_types():
             raise ValidationError(f"Default strategy {cls.default_strategy} not in strategies")
 
-        for sync_strategy in cls.strategies:
-            sync_strategy.configure_parser(arg_parser)
-
         syncing_parser = arg_parser.add_argument_group('sync mode')
         mode_parser = syncing_parser.add_mutually_exclusive_group()
         mode_parser.add_argument(
@@ -256,6 +255,9 @@ class SyncerComponent(AsyncioIsolatedComponent):
             choices=cls.extract_modes(),
             default=cls.default_strategy.get_sync_mode(),
         )
+
+        for sync_strategy in cls.strategies:
+            sync_strategy.configure_parser(syncing_parser)
 
     @classmethod
     def validate_cli(cls, boot_info: BootInfo) -> None:
--- a/trinity/components/builtin/syncer/component.py
+++ b/trinity/components/builtin/syncer/component.py
@@ -3,6 +3,7 @@ from abc import (
     abstractmethod,
 )
 from argparse import (
+    ArgumentError,
     ArgumentParser,
     Namespace,
     _ArgumentGroup,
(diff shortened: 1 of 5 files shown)


trinity/sync/header/chain.py
```diff
@@ -0,0 +1,70 @@
...
+                head,
+                self._checkpoint,
+            )
+        else:
+            self.logger.info("Initializing header-sync; current head: %s", head)
+
+        self.manager.run_daemon_child_service(self._header_syncer)
+        self.manager.run_daemon_task(self._persist_headers)
+        # run sync until cancelled
+        await self.manager.wait_finished()
```
carver, line 57: From what I understand, this line isn't necessary anymore. The new `async-service` run method doesn't immediately exit when it returns, you have to manually cancel it if you want to exit.

trinity/sync/header/chain.py
```diff
@@ -0,0 +1,70 @@
...
+
+    async def _persist_headers(self) -> None:
+        async for headers in self._header_syncer.new_sync_headers():
+
+            self._header_syncer._chain.validate_chain_extension(headers)
+
+            timer = Timer()
+            await self._db.coro_persist_header_chain(headers)
+
+            head = await self._db.coro_get_canonical_head()
```
carver, line 67: Not for this PR, but it seems like it might make sense to add a return value to `persist_header_chain()`, like `import_block()`, that returns the old and new canonical headers. Here, it would be a minor win of saving a DB read.

trinity/components/builtin/syncer/component.py
```diff
@@ -243,17 +279,17 @@ def configure_parser(cls, arg_parser: ArgumentParser, subparser: _SubParsersActi
         if type(cls.default_strategy) not in cls.extract_strategy_types():
             raise ValidationError(f"Default strategy {cls.default_strategy} not in strategies")
 
-        for sync_strategy in cls.strategies:
-            sync_strategy.configure_parser(arg_parser)
-
-        syncing_parser = arg_parser.add_argument_group('sync mode')
+        syncing_parser = arg_parser.add_argument_group('sync mode', conflict_handler='resolve')
```
carver, line 282: Hm, the `conflict_handler` keyword isn't even documented for this method AFAICT: https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_argument_group

Maybe add a comment explaining how it works and why we need it. I'm guessing it's something like: `--sync-from-checkpoint` will get added twice, once for header sync and once for beam sync, and we need that to not crash.  Sound right?

An alternative that means we get to stop using an undocumented API would be to check if the argument is already there before adding, in `add_sync_from_checkpoint_arg()`.

cburgdorf: >Sound right?

That's correct. The reasoning here is that it would be bad if only one calls `add_sync_from_checkpoint_arg` while other strategies that also depend on it rely on the side effect of another strategy making the call.

>An alternative that means we get to stop using an undocumented API would be to check if the argument is already there before adding

Yes, that was my favorite way of operation but it doesn't seem as if  `argsparse` exposes such an API to check it. That said, maybe another acceptable option is to catch the exception in `add_sync_from_checkpoint_arg` and ignore it with a comment. Because there's another downside to the `conflict_handler` solution: It is not bound to this specific case, meaning it might silence other cases where one would actually want to get an error.

trinity/sync/header/chain.py
cburgdorf: I tried it but it does exit immediately. The other sync modes also still use it the same way.

https://github.com/ethereum/trinity/blob/ece990ac710eeef3e0fb1d3cf52590a09a2e00ec/trinity/sync/full/chain.py#L979-L983

trinity/sync/header/chain.py
cburgdorf: Good idea!

trinity/components/builtin/syncer/component.py
cburgdorf: I rewrote it with the exception solution but I'll hold off merging to confirm back with you if you think that's an acceptable solution.

trinity/sync/header/chain.py
carver: I'm confused, that sounds like the opposite of this comment: https://github.com/ethereum/trinity/pull/1692/files#r414904955

cc @gsalgado

trinity/sync/header/chain.py
carver: lol, wait a minute. It does! So we should probably use that instead of the canonical head lookup...

https://github.com/ethereum/py-evm/blob/6fef9c25eff570471d147a4f4a6d4eba00262f57/eth/db/header.py#L140-L143

trinity/sync/header/chain.py
gsalgado: If you have any tasks scheduled, it is fine for `run()` to exit, but the service will terminate as soon as all your scheduled tasks complete. This test shows it:
```
@pytest.mark.asyncio
async def test_foo():
    task_event = asyncio.Event()

    class S(Service):
        async def daemon_task_fn(self):
            await task_event.wait()

        async def run(self):
            self.manager.run_task(self.daemon_task_fn, name="daemon_task_fn")

    loop = asyncio.get_event_loop()
    loop.call_later(2, task_event.set)
    await AsyncioManager.run_service(S())
```

cburgdorf: @gsalgado that test uses `run_task`. I wonder if the same holds true for when `run_task` or `run_daemon_child_service` is being used as it is the case in this PR?

Anyway, I'm happy to address this post-merge if it turns out there's a better way to write this under the new service paradigm. All I can say for now is that simply removing `await self.manager.wait_finished()` does not work here.

pipermerriam: I believe this is a deficiency in the `async-service` APIs.  I think there are edge cases whre you have a `run` method like this which schedules some background things and then immediately returns.  I believe that the internals don't necessarily always handle this case correctly and sometimes the service actually exits and cancels itself before the background tasks/services actually get started.  I ran into this recently as well and used this pattern to fix it.

trinity/sync/header/chain.py
cburgdorf: You're right, fixed it. Also there are a few places that could benefit but I actually have some local word that refactors the header persistence logic to be shared broader across places so I'm holding off fixing the other places for now.

trinity/sync/header/chain.py
carver: I opened https://github.com/ethereum/async-service/issues/66 -- but don't feel very confident about it, so would be happy to have more :eyes: on it.

pipermerriam: thanks for opening that.  I also opened https://github.com/ethereum/async-service/issues/67 once I saw your issue as I think it is related and may improve the library UX.

PR #1708, opened 2020-04-29
Fix a new intermittent Beam Sync test crash

### LINKED ISSUE #1704: Beam sync test fails intermittently
`tests/core/p2p-proto/test_sync.py::test_beam_syncer()` fails intermittently after conversion to new-style Service: https://circleci.com/gh/ethereum/trinity/219875

I had never seen that one failing intermittently before, and since the exception is coming from `async_service`, it looks like something introduced with the recent conversion from BaseService to Service. @carver any idea what could be causing it?

carver:
### What was wrong?

Fix #1704 

### How was it fixed?

This bug was introduced in the switchover to async-service. The old BaseService didn't require a service to be active in order to call run_task(). The new one does, so it exposes a situation where a run_task was being triggered at the time that BeamBlockImporter was closing.

The fix? Exit early from the method if you're about to run a task when the service is shutting down.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/chain.py
+++ b/trinity/sync/beam/chain.py
@@ -605,6 +605,10 @@ class BeamBlockImporter(BaseBlockImporter, Service):
             parent_state_root: Hash32,
             lagging: bool = True) -> None:
 
+        if not self.manager.is_running:
+            # If the service is shutting down, we can ignore preview requests
+            return
+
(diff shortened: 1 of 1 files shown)


carver:
`EOFError` :(

PR #1713, opened 2020-05-05
Report trinity process and thread count to grafana

njgheorghita:
### What was wrong?
Metrics needed for number of processes spawned from main `trinity` process and number  of threads.

### How was it fixed?
Added metrics for total number of processes and threads via `psutil`.

Screen Shot 2020-05-04 at 10 11 43 PM

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/trinity/components/builtin/metrics/system_metrics_collector.py
+++ b/trinity/components/builtin/metrics/system_metrics_collector.py
@@ -1,11 +1,14 @@
 from typing import (
+    Iterator,
     NamedTuple,
+    Optional,
 )
 
 from async_service import (
     as_service,
     ManagerAPI,
 )
+from eth_utils import to_tuple
 import psutil
 
 from p2p import trio_utils
@@ -41,10 +44,16 @@ class NetworkStats(NamedTuple):
     in_packets: int
 
 
+class ProcessStats(NamedTuple):
+    process_count: int
+    thread_count: int
+
+
 class SystemStats(NamedTuple):
     cpu_stats: CpuStats
     disk_stats: DiskStats
     network_stats: NetworkStats
+    process_stats: ProcessStats
 
 
 def read_cpu_stats() -> CpuStats:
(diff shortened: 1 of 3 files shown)


trinity/components/builtin/metrics/system_metrics_collector.py
```diff
@@ -78,6 +87,44 @@ def read_network_stats() -> NetworkStats:
...
+def read_process_stats() -> ProcessStats:
+    trinity_main_process = get_main_trinity_process()
+    if not trinity_main_process:
+        num_processes = 0
+        num_threads = 0
+    else:
+        child_processes = trinity_main_process.children(recursive=True)
+        num_processes = len(child_processes) + 1
+        num_child_threads = sum([p.num_threads() for p in child_processes])
+        num_threads = num_child_threads + trinity_main_process.num_threads()
```
njgheorghita, line 120: I'm pretty sure I did this correctly. From what I can make out of the [docs](https://psutil.readthedocs.io/en/latest/#psutil.Process.num_threads) calling `num_threads` on the main process is "non-cumulative" - which I'm guessing means it doesn't include the threads of it's child processes?

trinity/components/builtin/metrics/system_metrics_collector.py
```diff
@@ -78,6 +87,44 @@ def read_network_stats() -> NetworkStats:
     )
 
 
+@to_tuple
+def get_all_python_processes() -> Iterator[psutil.Process]:
+    for p in psutil.process_iter():
```
cburgdorf, line 92: We try to avoid single letter variable names. Suggestion: `process_info`

trinity/components/builtin/metrics/system_metrics_collector.py
```diff
@@ -78,6 +87,44 @@ def read_network_stats() -> NetworkStats:
...
+            continue
+        except psutil.ZombieProcess:
+            continue
+        if 'python' in p.name():
+            yield p
+
+
+def get_main_trinity_process() -> Optional[psutil.Process]:
+    python_processes = get_all_python_processes()
+    for p in python_processes:
```
cburgdorf, line 105: Same here about single letter variables

trinity/components/builtin/metrics/system_metrics_collector.py
```diff
@@ -78,6 +87,44 @@ def read_network_stats() -> NetworkStats:
...
+        if 'trinity' in p.cmdline():
+            return p
+    return None
```
cburgdorf, line 108: Returning `None` will summon Piper haha. Joke aside, we try to avoid returning `None` and Piper has left [some words about it in our tactical manual](https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/antipatterns.md#return-value-checking).

I think this should raise an exception.

trinity/components/builtin/metrics/system_metrics_collector.py
```diff
@@ -78,6 +87,44 @@ def read_network_stats() -> NetworkStats:
...
+
+
+def read_process_stats() -> ProcessStats:
+    trinity_main_process = get_main_trinity_process()
+    if not trinity_main_process:
```
cburgdorf, line 113: Is there a valid reason why this could happen? If not, I think the exception (see my previous comment) should bubble up, ultimately bringing the client down. The only reason I can think of that this could happen is if we renamed the whole client but forgot to adjust the value here and in this case we rather want the whole program to crash to make it obvious. Or are there valid cases where we hit this code while the `trinity` process can not be found?

trinity/components/builtin/metrics/system_metrics_collector.py
```diff
@@ -78,6 +87,44 @@ def read_network_stats() -> NetworkStats:
...
+        num_processes = 0
+        num_threads = 0
+    else:
+        child_processes = trinity_main_process.children(recursive=True)
+        num_processes = len(child_processes) + 1
+        num_child_threads = sum([p.num_threads() for p in child_processes])
```
cburgdorf, line 119: Single letter variable

trinity/components/builtin/metrics/system_metrics_collector.py
cburgdorf: Yeah, understand it the same way :+1:

trinity/components/builtin/metrics/system_metrics_collector.py
pipermerriam: I laughed when I read this.

PR #1714, opened 2020-05-06
Fixup gaps in header chain due to checkpoint syncing

cburgdorf:
### What was wrong?

Whenever someone uses `--sync-from-checkpoint` we are introducing gaps in the chain of headers (and as a consequence also in the chain of blocks for beam sync) as a trade off to get a fully functional node faster. This is a compromise between security and convenience and the goal should be to close these gaps eventually.

### How was it fixed?

This introduces two very simple new sync primitives:

1. `HeaderChainGapSyncer` -> Sync a specific gap
2. `SequentialChainGapSyncer` -> Spin up instances of `HeaderChainGapSyncer` to work through the list of gaps.

It utilizes the `SequentialChainGapSyncer` for the sync modes `header` and `beam` but allows to disable the job via the `--disable-backfill` and `--disable-header-backfill` flags. The names for the flags are different because for the `header` sync there is nothing else to be filled except headers and for `beam` sync I would imagine that we want to allow fine grained control over which parts to backfill or not (header, blocks, state).

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


cburgdorf:
@carver This is a baby steps approach to deliver something useful in terms of backfilling headers. There's nothing sophisticated like using a specific peer etc. and it mostly builds on top of existing primitives. II'm not deep enough into sync to judge if this passes the minimal viable header backfill threshold or not which is why I'm pinging you for a review here :)

trinity/sync/header/chain.py
```diff
@@ -62,24 +72,101 @@ async def run(self) -> None:
             self.manager.cancel()
             return
 
+        # Running the backfiller at this point means there can not be any new gaps introduced
+        # after this point. This means, we can let this service finish and call the work done.
```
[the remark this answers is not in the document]
cburgdorf: >New gaps can be created during different persist scenarios

What I wanted to emphasis on is that this is being made *after* resolving the checkpoint on purpose because (with the current code at least) there can be no new gaps introduced after this point. Do you foresee a scenario where that assumption wouldn't hold?

Oh, ok, maybe I could imagine one. We could *theoretically* pivot to new checkpoints automatically resolved from etherscan under certain conditions if the user wants that which would in fact introduce new gaps on a periodic basis. But I don't really see us doing that. In that case, we could have a periodic check for gaps that would spin up a gap fill syncer but it sounds like a YAGNI case. Therefore, I decided to delay running the `SequentialHeaderChainGapsSyncer` up until a point where (currently) no new gaps can be introduced.

trinity/sync/header/chain.py
```diff
@@ -62,24 +72,101 @@ async def run(self) -> None:
...
             self.logger.info(
-                "Imported %d headers in %0.2f seconds, new head: %s",
-                len(headers), timer.elapsed, head)
+                "Trying to fill an invalid gap %s. Available: %s", self._gap, self._gaps
+            )
+            return
+
+        launch_block_number = BlockNumber(self._gap[0] - 1)
+        self.logger.info(f"Launching from {launch_block_number}")
+        launch_strategy = FromDefinedBlockNumberLaunchStrategy(launch_block_number)
```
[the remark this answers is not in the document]
cburgdorf: >This strategy never attempts to reach back to recover from that scenario.

Ah, you are right of course! I kinda thought that we would have more luck in the next attempt to find the right *closing header*  but of course it might very well be that it isn't just a single final header that was incorrect which we are now missing. So, we might have to roll back some of the already persisted headers instead.

I follow your proposed solution of subtracting `MAX_SKELETON_REORG_DEPTH` as a way to mitigate that. :+1:  

>It might make us accidentally cross over a checkpoint, but I think dealing with that particular scenario could wait until a follow-up PR.

I'm not exactly sure I follow that. Since we attempt to fill gaps **sequentially** we do always work on the **first gap**, hence when we look back there can't be a case where we would accidentally land on a gap imho. What (theoretically) *could*  happen is that `MAX_SKELETON_REORG_DEPTH` is still within the faulty range of `B` headers but I guess if that would happen, Ethereum would have bigger problems to solve :sweat_smile:

trinity/sync/header/chain.py
```diff
@@ -62,24 +72,101 @@ async def run(self) -> None:
...
 
+    async def run(self) -> None:
+        if self._gap not in self._gaps:
+            # User defined gaps (e.g. subsets) are currently not supported
             self.logger.info(
-                "Imported %d headers in %0.2f seconds, new head: %s",
-                len(headers), timer.elapsed, head)
+                "Trying to fill an invalid gap %s. Available: %s", self._gap, self._gaps
+            )
+            return
```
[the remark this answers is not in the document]
cburgdorf: >should not pass any gaps in and this class figures out gaps for itself

I wanted to keep the `HeaderChainGapSyncer` flexible enough to allow passing in a specific gap in the constructor. E.g. that would allow to build things like `ConcurrentHeaderChainGapSyncer` on top which runs a few instances of  `HeaderChainGapSyncer` each assigned to a different gap.

But at the same time, I think it is a convenient if `HeaderChainGapSyncer` can simply default to the first gap if no gap is given. I also think we want to validate that any passed in gap is fact a valid gap (a future version might allow gaps that are subsets of gaps though).

I rewrote it to eliminate `self._gaps` and not directly fetch the gaps in the constructor, maybe this eliminates your main concern?

trinity/sync/beam/chain.py
```diff
@@ -428,37 +437,9 @@ async def _persist_headers_if_tip_too_old(self) -> None:
...
-                headers[-1],
-                old_canon_headers[0] if len(old_canon_headers) else None,
-                old_canon_headers[-1] if len(old_canon_headers) else None,
-                new_canon_headers[0] if len(new_canon_headers) else None,
-                new_canon_headers[-1] if len(new_canon_headers) else None,
-            )
+        await persist_headers(self.logger, self._db, self._header_syncer, self._exit_if_checkpoint)
 
-    async def _exit_if_checkpoint(self, headers: Tuple[BlockHeader, ...]) -> bool:
+    async def _exit_if_checkpoint(self, headers: Sequence[BlockHeader]) -> bool:
```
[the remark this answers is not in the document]
cburgdorf: Good call :+1:

trinity/sync/common/headers.py
```diff
@@ -1019,3 +1025,47 @@ async def _handle_sync_status_requests(self, event_bus: EndpointAPI) -> None:
...
+            head = new_canon_headers[-1]
+        else:
+            head = await db.coro_get_canonical_head()
+
+        logger.info(
+            "Imported %d headers in %0.2f seconds, new head: %s",
+            len(headers),
+            timer.elapsed,
+            head,
+        )
```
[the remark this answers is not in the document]
cburgdorf: Makes sense. Since we also need to expose the elapsed time I felt it looks cleaner to introduce a `HeaderPersistInfo` type to have a more concise `async for persist_info in` as well as a cleaner `AsyncIterator[HeaderPersistInfo` return type.

trinity/sync/header/chain.py
```diff
@@ -62,24 +72,101 @@ async def run(self) -> None:
...
+            self.manager.run_child_service(backfiller)
+
         self.manager.run_daemon_child_service(self._header_syncer)
         self.manager.run_daemon_task(self._persist_headers)
         # run sync until cancelled
         await self.manager.wait_finished()
 
     async def _persist_headers(self) -> None:
-        async for headers in self._header_syncer.new_sync_headers():
+        await persist_headers(self.logger, self._db, self._header_syncer)
```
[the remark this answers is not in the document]
cburgdorf: This turned now into an `async for` so it's staying in place.

trinity/sync/common/strategies.py
```diff
@@ -249,3 +249,22 @@ async def fulfill_prerequisites(self) -> None:
...
+     block number already exist, otherwise the sync will fail with an exception.
+    """
+
+    logger = logging.getLogger('trinity.sync.common.strategies.FromMissingHeaderLaunchStrategy')
+
+    def __init__(self, missing_block_number: BlockNumber) -> None:
+        self._missing_block_number = missing_block_number
+
+    async def fulfill_prerequisites(self) -> None:
+        pass
```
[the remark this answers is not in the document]
cburgdorf: Good point. Also renamed `missing_block_number` to `launch_block_number` as the name was a leftover from an earlier version and doesn't actually make sense now.

trinity/components/builtin/syncer/component.py
```diff
@@ -95,6 +90,49 @@ def add_sync_from_checkpoint_arg(arg_parser: _ArgumentGroup) -> None:
             raise
 
 
+def add_sync_from_checkpoint_arg(arg_group: _ArgumentGroup) -> None:
+    arg_name = '--sync-from-checkpoint'
+    add_shared_argument(
+        lambda: arg_group.add_argument(
```
cburgdorf, line 96: I hope this is an allowed lambda case :sweat_smile: I didn't want to copy the error handling over and over

trinity/sync/header/chain.py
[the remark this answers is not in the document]
carver: > there can be no new gaps introduced after this point. Do you foresee a scenario where that assumption wouldn't hold?

Yeah, but maybe we're talking about different things? I am saying that filling the gaps can actually add a new gap also. Like it's possible to start with gaps `(((3,3),), 6)`, then fill a header at 3, then have gaps `(((4,4),), 6)`. Anyway, the statement "there can not be any new gaps introduced after this point" looks not quite right.

I'm not trying to suggest the functionality needs to change, just that maybe the comment could be misleading (because of the above scenario).

Maybe a slightly more precise statement is something like:
> Because checkpoints are only set at startup (for now): once all gaps are filled, no new ones will be created. So we can simply run this service till completion and then exit.

trinity/sync/header/chain.py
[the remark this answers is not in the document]
carver: Yeah, I still think `YAGNI` applies here for building features that aren't used yet (like trying to support a concurrent gap syncer). That it gets a lot simpler if we don't try to support the use case where we don't quite know the API yet. (Since we don't have to pass in a gap at all, which means we don't have to verify it or comment on it). But it's not a big deal, and it's better than it used to be.

trinity/sync/header/chain.py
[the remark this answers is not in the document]
carver: > > It might make us accidentally cross over a checkpoint, but I think dealing with that particular scenario could wait until a follow-up PR.
> 
> I'm not exactly sure I follow that. Since we attempt to fill gaps **sequentially** we do always work on the **first gap**, hence when we look back there can't be a case where we would accidentally land on a gap imho.

_(BTW, yet another reason to not worry about supporting concurrent parallel gap filling)_

The concern isn't that we cross into an old gap, but that we cross before an old checkpoint. So if:
- A[3] is a checkpoint
- A[100] is a checkpoint
- A[4:10] is persisted
- we try to fill the (10,99) gap with a new peer who is sending uncle chains.
- we start asking the peer for headers from genesis (because of the lookback)
- the peer gives us alternative headers, including one that would replace (decanonicalize) the checkpoint at A[3]
- are we catching this exception properly? should we do a temporary ban on the peer? etc.

> What (theoretically) _could_ happen is that `MAX_SKELETON_REORG_DEPTH` is still within the faulty range of `B` headers but I guess if that would happen, Ethereum would have bigger problems to solve

These long uncle chains are actually happening! Not so much on the tip of the chain, but definitely in history. (I have personally witnessed it on Ropsten). It's not so hard to build 35000 orphan headers from genesis. So we definitely have to think about how to deal with it for filling in old headers.

Right now, our main mechanism to prevent this is that two peers have to agree on the uncle chain: the skeleton peer and another peer that fills the meat on the skeleton. We are hoping that we don't land on the naughty peer for building the skeleton, and if we do, that we don't find yet another peer willing to fill in the skeleton gaps.

trinity/sync/header/chain.py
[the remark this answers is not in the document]
cburgdorf: Ah gotcha :+1:

trinity/sync/header/chain.py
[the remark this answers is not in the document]
cburgdorf: Thank you very much for the detailed write up! I think I understand the issue now. Let me try to put it in my own words.

I actually had to look up the most recent gap tracking rules again and it looks that we have settled with the rule that if we declare `A3` a checkpoint than there's no way that we can ever overtake that checkpoint with a chain B even if chain B turns out to be the heavier chain and even after the gap had already disappeared before. To let Py-EVM tests speak, this is the status quo:

```python
(StepAction.PERSIST_CHECKPOINT, 2),
# First attempt to overtake with B fails because we can not close the gap with B headers
(StepAction.VERIFY_PERSIST_RAISES, ('b', CheckpointsMustBeCanonical, lambda headers: headers[:5])),
(StepAction.VERIFY_GAPS, (((1, 1),), 3)),
(StepAction.PERSIST_HEADERS, ('a', lambda headers: headers[:3])),
(StepAction.VERIFY_GAPS, ((), 4)),
# Even after the gap is closed we still can't overtake the checkpoint with chain B
(StepAction.VERIFY_PERSIST_RAISES, ('b', CheckpointsMustBeCanonical, lambda headers: headers[:5])),
```

So, I guess what you are saying is that we should check that our `launch_block_number` does not reach further back than our previous checkpoint which we could validate if Py-EVM exposes a `get_checkpoints()` API.

This got me thinking about the following scenario:

1. The user runs `trinity --sync-from-checkpoint <checkpoint>`

Trinity runs, and manages to backfill the header chain and we might even print something in the logs that says:

>:tada: the header chain is now connected back to genesis

2. The user shuts down the client
3. The user runs Trinity again, this time without the `--sync-from-checkpoint` flag.

But what happens now? We are still bound to the rules of Py-EVM and as the test proves, Py-EVM will hold on to these checkpoints no matter what.

If the default launch strategy (`FromGenesisLaunchStrategy` ) stays as is, we could still cross over the previous checkpoint and run into the scenario that you described and eventually hit a Py-EVM exception. So, we are still ruled by our checkpoint even if that may not be our intention?

One way to deal with that could be to call `get_checkpoints()` early and if checkpoints exists we would tell the user that these are still taken into consideration and if they **intent not to use these** they manually have to run something like `trinity clear-checkpoints` to fix the database. That would call a new Py-EVM API `clear_checkpoints()` to get rid of all checkpoints to be  exclusively governed by common chain rules again.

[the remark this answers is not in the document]
carver: > But what happens now? We are still bound to the rules of Py-EVM and as the test proves, Py-EVM will hold on to these checkpoints no matter what.
> 
> If the default launch strategy (`FromGenesisLaunchStrategy` ) stays as is, we could still cross over the previous checkpoint and run into the scenario that you described and eventually hit a Py-EVM exception. So, we are still ruled by our checkpoint even if that may not be our intention?

Yup! Although, if you chose a checkpoint that ends up reverting, then that's user error. We are okay crashing and requiring user intervention at that point. I want to keep hammering home the idea that a checkpoint should really be something we never expect to revert.

I think the `FromGenesisLaunchStrategy` is a misnomer now. It's more like the `UseExistingCheckpointsOrGenesis` strategy.

> One way to deal with that could be to call `get_checkpoints()` early and if checkpoints exists we would tell the user that these are still taken into consideration ...

Yup, this seems like a very reasonable thing to print into INFO level logs.

> ... and if they **intent not to use these** they manually have to run something like `trinity clear-checkpoints` to fix the database. That would call a new Py-EVM API `clear_checkpoints()` to get rid of all checkpoints to be exclusively governed by common chain rules again.

Yup, I think that's okay, but it's also okay to push into an issue (probably in the beta milestone instead of the MVP one). It would require more discussion. Maybe/probably we want a tool to clear just one checkpoint instead of all of them, etc.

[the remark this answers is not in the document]
carver: > So, I guess what you are saying is that we should check that our `launch_block_number` does not reach further back than our previous checkpoint which we could validate if Py-EVM exposes a `get_checkpoints()` API.

Oh and yes, this is what I had in mind. :+1:

trinity/components/builtin/syncer/component.py
cburgdorf: Yeah, that also works for me :+1:

tests/core/p2p-proto/test_sync.py
```diff
@@ -519,6 +518,50 @@ async def test_sequential_header_gapfill_syncer(request,
...
+
+            async with background_asyncio_service(client):
+                try:
+                    await wait_for_head(
+                        # We check for 499 because 500 exists from the
+                        # very beginning (the checkpoint)
+                        chaindb_with_gaps, chaindb_1000.get_canonical_block_header_by_number(499)
+                    )
+                except asyncio.TimeoutError:
+                    assert "Attempted to fill gap with invalid header" in caplog.text
```
cburgdorf, line 562: @carver This test doesn't yet prove that we recover from that scenario but maybe it would be ok to do that in a follow up PR.

cburgdorf:
@carver I think I addressed all your comments. This doesn't yet include a sophisticated test that proves we recover from an "uncle attack" but maybe that can wait till a follow up PR.

tests/core/p2p-proto/test_sync.py
```diff
@@ -425,6 +442,126 @@ async def test_header_syncer(request,
...
+        ), background_asyncio_service(ETHRequestServer(
+            event_bus, TO_NETWORKING_BROADCAST_CONFIG, AsyncChainDB(chaindb_1000.db),
+        )):
+
+            server_peer.logger.info("%s is serving 1000 blocks", server_peer)
+            client_peer.logger.info("%s is syncing up 1000", client_peer)
+
+            async with background_asyncio_service(client):
+                await wait_for_head(
+                    # We check for 499 because 250 is there from the very beginning (the checkpoint)
```
carver, line 479: Should be "check for 249" I suppose?

tests/core/integration_test_helpers.py
```diff
@@ -56,7 +56,7 @@ def load_mining_chain(db):
 class DBFixture(Enum):
     TWENTY_POW_HEADERS = '20pow_headers.ldb'
     THOUSAND_POW_HEADERS = '1000pow_headers.ldb'
-
+    UNCLE_CHAIN = '1000pow_uncle_chain.ldb'
```
carver, line 59: Maybe a quick summary of what's happening here, and mention the arguments that `build_uncle_chain_to_existing_db()` was called with to generate this? Similar to lines 60-62 below.

trinity/components/builtin/syncer/component.py
```diff
@@ -206,6 +234,7 @@ async def sync(self,
             event_bus,
             args.sync_from_checkpoint,
             args.force_beam_block_number,
+            not args.disable_backfill
```
carver, line 237: *nit* trailing comma

trinity/sync/common/headers.py
```diff
@@ -8,24 +8,28 @@
...
+    Awaitable,
     Callable,
     FrozenSet,
     Generic,
     Iterable,
     Optional,
     Sequence,
     Tuple,
-    Type,
+    Type, NamedTuple,
```
carver, line 19: *nit* to its own line

trinity/sync/common/strategies.py
```diff
@@ -247,3 +247,28 @@ async def fulfill_prerequisites(self) -> None:
...
+
+
+class FromDefinedBlockNumberLaunchStrategy(SyncLaunchStrategyAPI):
+    """
+    A launch strategy that is starting to sync from a specified block number. In contrast to
+     ``FromCheckpointLaunchStrategy``, this strategy assumes all previous parents to the defined
+     block number already exist, otherwise the sync will fail with an exception.
+    """
+
+    logger = logging.getLogger('trinity.sync.common.strategies.FromMissingHeaderLaunchStrategy')
```
carver, line 259: *nit* logger name mismatch

trinity/sync/header/chain.py
```diff
@@ -62,24 +73,124 @@ async def run(self) -> None:
...
+
+        available_gaps, _ = self._db.get_header_chain_gaps()
+
+        if self._gap is None:
+            # If no gap was provided fall back to the first gap in the chain
+            if len(available_gaps) > 1:
+                self._gap = available_gaps[0]
             else:
-                head = await self._db.coro_get_canonical_head()
+                self.logger.info("No gaps to fill. Aborting")
```
carver, line 124: *nit* "Abort" means to me that something went wrong, but this seems like a "happy termination" case. I would expect something like "Stopping" or "Exiting".

trinity/sync/header/chain.py
```diff
@@ -62,24 +73,124 @@ async def run(self) -> None:
...
-            new_canonical_headers, _ = await self._db.coro_persist_header_chain(headers)
 
-            if len(new_canonical_headers) > 0:
-                head = new_canonical_headers[-1]
+class HeaderChainGapSyncer(Service):
+    def __init__(self,
+                 chain: AsyncChainAPI,
+                 db: BaseAsyncChainDB,
+                 peer_pool: ETHPeerPool,
+                 gap: Tuple[BlockNumber, BlockNumber] = None) -> None:
```
carver, line 107: Maybe worth mentioning in a comment/docstring that (as currently implemented), the `gap` should always be the first gap of the chain gaps.

trinity/sync/header/chain.py
```diff
@@ -62,24 +73,124 @@ async def run(self) -> None:
...
+                    self.logger, self._db, self._header_syncer, self._is_at_end_of_gap):
+                self.logger.info(
+                    "Imported %d gap headers from %s to %s in %0.2f seconds,",
+                    len(persist_info.imported_headers),
+                    persist_info.imported_headers[0],
+                    persist_info.imported_headers[-1],
+                    persist_info.elapsed_time,
+                )
+        except CheckpointsMustBeCanonical as err:
+            # TODO: we should have a test case for this
```
carver, line 169: I guess this is the `TODO` that can be removed when the new issue is added to the MVP milestone.

trinity/sync/header/chain.py
cburgdorf: Well, the way I wanted to see the `HeaderChainGapSyncer` is as a low level class that will simply attempt to fix the gap that you tell it to fix. But now that we subtract the `MAX_SKELETON_REORG_DEPTH` internally, that assumption doesn't really hold.

If we ended up at a place where you can not use this primitive in the intended way anymore then I'd be ready to go in any of these directions:

- Drop `HeaderChainGapSyncer` entirely and inline all the code in the `SequentialHeaderChainGapSyncer`
- Keep `HeaderChainGapSyncer` but drop the `gap` argument to `__init__`
- Keep the `gap` argument to `__init__` but don't alter it and move the alteration with `MAX_SKELETON_REORG_DEPTH` into the `SequentialHeaderChainGapSyncer`

trinity/sync/header/chain.py
cburgdorf: Zounts gudd. Inglisch nodd mai fürst languitsch!

trinity/sync/common/strategies.py
cburgdorf: :+1: I also shortened the whole class name to `FromBlockNumberLaunchStrategy`

trinity/sync/common/headers.py
cburgdorf: I blame it on PyCharm :see_no_evil:

tests/core/p2p-proto/conftest.py
```diff
@@ -19,6 +19,26 @@ def leveldb_1000():
...
+@pytest.fixture
+def leveldb_uncle_chain():
+    yield from load_fixture_db(DBFixture.UNCLE_CHAIN)
+
+
+@pytest.fixture
+def chaindb_uncle(leveldb_1000, leveldb_uncle_chain):
+    canoncical_chain = load_mining_chain(AtomicDB(leveldb_1000))
+    uncle_chain = load_mining_chain(AtomicDB(leveldb_uncle_chain))
+    # Shared history from genesis till header 474
```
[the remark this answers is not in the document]
cburgdorf: Correct. I'll add a comment.

trinity/sync/header/chain.py
cburgdorf: Ok, I rewrote the `HeaderChainGapSyncer` to drop the `gap` parameter.

tests/core/p2p-proto/test_sync.py
cburgdorf: I created an issue and assigned it to the milestone https://github.com/ethereum/trinity/issues/1778

PR #1720, opened 2020-05-11
Convert validator API to use quart-trio

pipermerriam:
### What was wrong?

`quart-trio` provides us with a flask-like API for running a trio-friendly asynchronous webserver.

### How was it fixed?

Converted the previous h11 based approach to use quart-trio instead.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


pipermerriam:
@ralexstokes here's a preliminary draft of the server component.  I've stubbed out all the methods that were not implemented and converted the existing methods.  None of this has been tested yet.  I'm going to shift focus to the client now so that I can get some basic end-to-end tests in place.

ralexstokes:
great @pipermerriam i'll review the draft so far.

there are a few unit tests touching this logic i can point you to and you should definitely take a look at a rudimentary integration test i started here: https://github.com/ethereum/trinity/blob/master/tests-trio/integration/test_validator_api.py

eth2/api/http/app.py
```diff
@@ -0,0 +1,7 @@
+from quart_trio import QuartTrio
+
+from .converters import HexadecimalPublicKeyConverter
+
+
+app = QuartTrio(__name__)
+app.url_map.converters['pubkey'] = HexadecimalPublicKeyConverter
```
ralexstokes, line 7: nice sugar, `quart-trio` :)

eth2/api/http/converters.py
```diff
@@ -0,0 +1,13 @@
+from eth_utils import encode_hex, decode_hex
+
+from werkzeug.routing import BaseConverter
+
+
+class HexadecimalPublicKeyConverter(BaseConverter):
+    regex = r"0x?([0-9a-fA-F]{96})"
```
ralexstokes, line 7: i think you want:
```suggestion
    regex = r"(0x)?([0-9a-fA-F]{96})"
```

eth2/api/http/ir.py
```diff
@@ -129,13 +113,12 @@ def get_block_proposal(
             slot, parent_block_root, randao_reveal, parent_state, state_machine
         )
 
-    async def broadcast_block(self, signed_block: SignedBeaconBlock) -> bool:
+    async def broadcast_block(self, signed_block: SignedBeaconBlock):
```
ralexstokes, line 116: we will want to signal to the client if broadcast fails -- we can throw an exception here but will need a way to transmit that across the HTTP boundary -- food for thought.

eth2/api/http/status.py
```diff
@@ -0,0 +1,2 @@
+HTTP_200_OK = 200
+HTTP_201_CREATED = 201
```
ralexstokes, line 2: i'm not sure if we ultimately want stronger typing here. i'm very ok w/ integer types, but we could consider putting them under an `enum`. shouldn't block merging but a point to consider on improving type safety.

eth2/api/http/views.py
```diff
@@ -0,0 +1,151 @@
...
+from eth_utils import decode_hex, encode_hex
+from quart import request, jsonify, Response, make_response
+from ssz.tools.dump import to_formatted_dict
+from ssz.tools.parse import from_formatted_dict
+
+from eth2.beacon.types.blocks import SignedBeaconBlock
+from eth2.beacon.typing import Slot
+
+from .app import app
+from .ir import ValidatorDuty
```
ralexstokes, line 14: just top of mind -- i don't know what an `ir` is.

"intermediate", like "i'll change this later before we merge this PR?"

eth2/api/http/views.py
```diff
@@ -0,0 +1,151 @@
...
+from .status import HTTP_201_CREATED
+
+
+GET = "GET"
+POST = "POST"
```
ralexstokes, line 19: another place to consider stronger typing... may not add much beyond just having dedicated constants though

eth2/api/http/views.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+#
+# Routes: `/beacon`
+#
+@app.route('/beacon/genesis')
+async def beacon_genesis() -> Response:
+    context = app.config["context"]
```
ralexstokes, line 27: i don't think the symbol `context` is that special here -- we could save some indirection by setting these properties directly on the `app.config` although i'm not sure if `quart` has conventions around putting user-level data in this `config` or if they prefer specific namespaces (e.g. like our `context` we have here)

eth2/api/http/views.py
```diff
@@ -0,0 +1,151 @@
...
+    return jsonify({
+        "genesis_time": str(context.genesis_time),
```
ralexstokes, line 29: not sure why this is a string but i double-checked the spec and it is so let's go w/ it

eth2/api/http/views.py
```diff
@@ -0,0 +1,151 @@
...
+        "genesis_validators_root": "TODO",
```
ralexstokes, line 30: this will be static data per-chain that we can set much like the `genesis_time` once we have it

eth2/api/http/views.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+@app.route('/beacon/fork')
+async def beacon_fork() -> Response:
+    raise NotImplementedError
+
+
+@app.route('/beacon/fork/stream')
+async def beacon_fork_stream() -> Response:
+    # TODO: websockets and streaming response....?
```
ralexstokes, line 41: in general, websockets ability would be great.

the general validator APIs are all "polling", in part to simplify implementation for those wanting to conform to a standard.

but nothing is stopping us from having special APIs based on our knowledge of the client, and e.g. stream duties to the validator client if that is something supported by trinity beacon nodes/validator client.

eth2/api/http/views.py
```diff
@@ -0,0 +1,151 @@
...
+#
+@app.route('/node/version')
+async def node_version() -> Response:
+    return jsonify(app.config["context"].client_identifier)
+
+
+@app.route("/node/syncing")
+async def sync_status() -> Response:
+    context = app.config["context"]
+    status = await context.get_sync_status()
```
ralexstokes, line 56: again not sure how we want to handle getting the "dynamic" information from the chain object to each of these API endpoints but in lieu of other demands i'd say we just keep it simple and build on top of this `Context` thing

eth2/api/http/views.py
```diff
@@ -0,0 +1,151 @@
...
+        decode_hex(request["randao_reveal"]).ljust(96, b"\x00")
+    )
+    block = context.get_block_proposal(slot, randao_reveal)
+    return jsonify(to_formatted_dict(block))
+
+
+async def _post_validator_block(epoch) -> Response:
+    context = app.config['context']
+    raw_payload = await request.get_json()
+    # TODO: should validate post-data
```
ralexstokes, line 148: do you mean consensus/protocol level validations?

or something specific to HTTP?

if the `raw_payload` is not a properly written `SignedBeaconBlock` then the next line will fail.

and while we can chat exact API, i'd figure we put the consensus validations ("is this a sensical state transition, is this the proposer for this slot, etc") behind `context.broadcast_block`.

this is also a good time to acknowledge we are generally trusting the software on the other side of this API (i.e. "beacon node trusts validator client" and "validator client trusts beacon node"), otherwise we cant profitably capitalize on the separation b/t the two chunks of software

trinity/nodes/beacon/full.py
```diff
@@ -139,13 +122,9 @@ async def _iterate_clock(
...
-        server = self._validator_api_server
-        async with trio.open_nursery() as nursery:
-            self.validator_api_port = await nursery.start(server.serve)
-            self.logger.info(
-                "validator HTTP API server listening on %d", self.validator_api_port
-            )
-            task_status.started()
+        app.config["context"] = self._api_context
+        task_status.started()
+        await app.run_task(port=self._validator_api_port)
```
ralexstokes, line 127: something that may not be obvious here is that we in fact want to give the API server a chance to boot in a blocking fashion as this gives the caller the option to let the kernel select a free port (by passing in `_validator_api_port == 0`). we want to block so the caller can let the port be allocated and then read downstream to know where to connect to -- this is currently being used in the tests and was the motivation for introducting the `task_status` machinery from `trio` land.

 i'm assuming `quart` wraps the same socket machinery so ultimately we can do a similar thing here (they may even have a special API for it??). if there is a way to avoid `task_status` then that would be preferable as this may be the only task/component that needs to block in this fashion so if we can skip the overhead in all of the other places, that would be great!

eth2/api/http/status.py
pipermerriam: Yeah, these make a lot of sense to be an `IntEnum`

eth2/api/http/views.py
pipermerriam: I'm game to come up with different name.  "Intermediate Representation".  These are the representations of this data that we use internally to represent the data structures we need to serve the REST API.

eth2/api/http/views.py
pipermerriam: From what I can find in the quart/flask documentation this is the place to store something like this.  In theory we could do something similar to `quart.request` or `quart.current_app` using thread locals to access this value.  It would be cleaner in the sense that there would just be a *magic* `context` variable that was always the appropriate context being used by the REST API, but ultimately that is just *UX sugar* on top of something like this.  If you see an alternate approach I'm all ears.

eth2/api/http/views.py
pipermerriam: I'll probably leave this un-implemented at first.  It does look quite trivial to implement as quart has native support for websockets.

eth2/api/http/views.py
pipermerriam: This was simply about returning 4xx level responses with informative error messages.

trinity/nodes/beacon/full.py
pipermerriam: :+1:

pipermerriam:
The pairing I did with @marcgarreau yesterday was really positive. Quart gives us nice tools for testing the server in isolation, as well as it being pretty easy to orchestrate tests that actually interface with the HTTP API.

eth2/api/http/views.py
ralexstokes: makes sense -- i like the idea, can't think of something better than `ir` at the moment

eth2/api/http/views.py
ralexstokes: yeah then just doing what we have here works for me

eth2/api/http/views.py
ralexstokes: good to hear -- i agree we can skip wss support for now

PR #1721, opened 2020-05-13
Use asyncio-run-in-process to run trio components

gsalgado:
Closes: #1435

### DIFF
--- a/trinity/extensibility/trio.py
+++ b/trinity/extensibility/trio.py
@@ -1,17 +1,13 @@
 from abc import abstractmethod
-import asyncio
-from multiprocessing import Process
-import os
-import signal
 
 import trio
 from async_service import background_trio_service
-from asyncio_run_in_process.constants import SIGINT_TIMEOUT_SECONDS, SIGTERM_TIMEOUT_SECONDS
 
 from lahja import EndpointAPI
 
+from asyncio_run_in_process import open_in_process_with_trio
+
 from trinity._utils.logging import child_process_logging, get_logger
-from trinity._utils.mp import ctx
 from trinity._utils.profiling import profiler
 from trinity.boot_info import BootInfo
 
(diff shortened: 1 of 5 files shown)


trinity/extensibility/trio.py
```diff
@@ -31,86 +27,42 @@ class TrioIsolatedComponent(BaseIsolatedComponent):
...
-            # SIGINTs), and if we don't send one at all the subprocess never terminates when
-            # trinity exits because of a crash.
-            self.logger.debug("Waiting for process %d to terminate", process.pid)
-            await loop.run_in_executor(None, process.join, SIGINT_TIMEOUT_SECONDS / 4)
-            if process.is_alive():
-                self.logger.debug("Process %d did not terminate, sending SIGINT", process.pid)
-                self._send_signal_and_join(process, signal.SIGINT, SIGINT_TIMEOUT_SECONDS)
-            if process.is_alive():
-                self.logger.debug("Process %d did not terminate, sending SIGTERM", process.pid)
-                self._send_signal_and_join(process, signal.SIGTERM, SIGTERM_TIMEOUT_SECONDS)
```
cburgdorf, line 72: Yay, I like to see a lots of red lines :heart:

trinity/tools/_component_isolation.py
```diff
@@ -60,42 +60,50 @@ def get_subprocess_kwargs(self) -> SubprocessKwargs:
     def is_enabled(self) -> bool:
         return True
 
-    @classmethod
-    async def do_run(cls, boot_info: BootInfo, event_bus: EndpointAPI) -> None:
-        cls.logger.info('Entered `do_run`')
+    async def do_run(self, boot_info: BootInfo, event_bus: EndpointAPI) -> None:
+        self.logger.info('Entered `do_run`')
```
cburgdorf, line 64: What's the reason that became an instance method?

trinity/tools/_component_isolation.py
```diff
@@ -60,42 +60,50 @@ def get_subprocess_kwargs(self) -> SubprocessKwargs:
...
+
     @property
     def is_enabled(self) -> bool:
         return True
 
-    @classmethod
-    async def do_run(cls, boot_info: BootInfo, event_bus: EndpointAPI) -> None:
-        cls.logger.info('Entered `do_run`')
+    async def do_run(self, boot_info: BootInfo, event_bus: EndpointAPI) -> None:
+        self.logger.info('Entered `do_run`')
```
cburgdorf, line 99: Same here, what's the reason this became an instance method?

trinity/tools/_component_isolation.py
```diff
@@ -60,42 +60,50 @@ def get_subprocess_kwargs(self) -> SubprocessKwargs:
...
     name = "component-test-trio"
     endpoint_name = 'component-test-trio'
 
+    def get_subprocess_kwargs(self) -> SubprocessKwargs:
+        return merge(
+            super().get_subprocess_kwargs(),
+            {
+                'stdin': subprocess.PIPE,
+                'stdout': subprocess.PIPE,
+                'stderr': subprocess.PIPE,
```
pipermerriam, line 90: This seems like it deserves a comment as it isn't clear to me why we need to pass these for only the `trio` version of this.

carver:
I wasn't able to tell from scanning the patch: what was making the `EOFError` happen and how does this fix it?

trinity/tools/_component_isolation.py
gsalgado: We have that in the asyncio version as well, I copied it from there. Without that pytest crashes with a horrible traceback. I thought you were the one who figured out this was needed as you wrote these tests originally, no? Will add a comment, anyway

I considered moving it to a common base class but it would be annoying because it'd require multiple inheritance, so I decided to keep the small duplication.

gsalgado:
> I wasn't able to tell from scanning the patch: what was making the EOFError happen and how does this fix it?

We still don't know what was causing it, but it comes from the multiprocessing module, and this is the last place we were using that in trinity.

trinity/tools/_component_isolation.py
gsalgado: It had no reason for being a classmethod, but also to be consistent with `AsyncioIsolatedComponent`

trinity/tools/_component_isolation.py
gsalgado: This is the traceback, btw

```
self = <_pytest.capture.DontReadFromInput object at 0x7f2e41a635d0>

    def fileno(self):
>       raise UnsupportedOperation("redirected stdin is pseudofile, has no fileno()")
E       io.UnsupportedOperation: redirected stdin is pseudofile, has no fileno()

self       = <_pytest.capture.DontReadFromInput object at 0x7f2e41a635d0>

../../virtualenvs/trinity-asyncio/lib/python3.7/site-packages/_pytest/capture.py:711: UnsupportedOperation
```

pipermerriam:
Huzzah,  the end of multiprocessing.

trinity/tools/_component_isolation.py
pipermerriam: :+1: I didn't see what module I was in (and thus missed the fact this was for testing purposes).

PR #1726, opened 2020-05-14
Add basic block import to beacon node

ralexstokes:
The beacon node now handles basic block import with a variety of validation outcomes.

#### TODO

- [x] test that injects some blocks and yields the (expected) canonical head

#### Cute Animal Picture


eth2/beacon/chains/base.py
```diff
@@ -471,15 +472,27 @@ def import_block(
...
             parent_block = self.get_block_by_root(block.parent_root)
         except BlockNotFound:
-            raise ValidationError(
-                "Attempt to import block #{}.  Cannot import block {} before importing "
-                "its parent block at {}".format(
-                    block.slot,
-                    humanize_hash(block.message.hash_tree_root),
-                    humanize_hash(block.parent_root),
-                )
+            raise ParentNotFoundError(
```
ralexstokes, line 475: we want an explicit error here to capture failed import due to a missing parent as we will process orphan blocks differently than invalid blocks

eth2/beacon/chains/base.py
```diff
@@ -471,15 +472,30 @@ def import_block(
...
+                # NOTE: we want to keep the block but avoid heavy state transition for now...
+                # Rationale: this block may simply be a slashable block. It could also be on
+                # a fork. Choose to defer the state materialization until we re-org via fork choice.
+                self.chaindb.write_signed_block(block)
+
+                raise SlashableBlockError(
+                    block,
+                    f"attempt to import {block} but canonical chain"
+                    " already has a block at this slot",
+                )
```
ralexstokes, line 491: we found a duplicate block, raise it as "slashable" while storing the block in the event that we re-org to another chain (given a race b/t an "attack" chain and a block slashing getting on-chain)

eth2/beacon/chains/base.py
```diff
@@ -471,15 +472,30 @@ def import_block(
...
+            else:
+                # NOTE: block already imported!
+                return block, (), ()
```
ralexstokes, line 494: we already have this block, skip the heavy state transition

eth2/beacon/chains/base.py
```diff
@@ -471,15 +472,30 @@ def import_block(
...
+        except BlockNotFound:
+            # NOTE: block has not been imported for ``block.slot``
+            pass
```
ralexstokes, line 497: new block w/ parent: let's keep moving on!

eth2/beacon/db/chain.py
```diff
@@ -466,6 +470,16 @@ def _set_block_score_to_db(
     def set_score(self, block: BaseBeaconBlock, score: BaseScore) -> None:
         self.__class__._set_block_score_to_db(self.db, block, score)
 
+    @classmethod
+    def _write_signed_block(cls, db: DatabaseAPI, block: BaseSignedBeaconBlock) -> None:
+        # TODO key by block root or signed block root?
```
ralexstokes, line 475: should likely make this an issue to make sure there are no edge cases where we need to key by block root or signed block root

eth2/beacon/db/chain.py
```diff
@@ -466,6 +470,16 @@ def _set_block_score_to_db(
...
+        db.set(block.message.hash_tree_root, ssz.encode(block))
```
ralexstokes, line 476: another issue: this should likely go through some DB schema to get a key

eth2/beacon/db/chain.py
```diff
@@ -854,7 +864,7 @@ def _get_canonical_epoch_info(db: DatabaseAPI) -> EpochInfo:
     def _add_attestations_root_to_block_lookup(
         db: DatabaseAPI, block: BaseBeaconBlock
     ) -> None:
-        root = block.message.hash_tree_root
```
ralexstokes, line 857: we found a `mypy` failure... the tests did bring it up though!

trinity/nodes/beacon/full.py
```diff
@@ -86,6 +88,8 @@ def __init__(
 
         self._chain = chain_class(self._base_db, eth2_config)
 
+        self._block_pool: Set[SignedBeaconBlock] = set()
```
ralexstokes, line 91: future issue: decide how to integrate w/ `eth2.beacon.operations.pool` types

trinity/nodes/beacon/full.py
```diff
@@ -147,6 +151,50 @@ async def _run_validator_api(
             )
             task_status.started()
 
+    def on_block(self, block: SignedBeaconBlock) -> None:
```
ralexstokes, line 154: top-level block handler: i give a beacon node a block and it runs this.

the block could come from the gossip network, the local syncing apparatus, or an attached validator client.

will want to look a validations in the gossip layer and see if they make sense to apply to the other layers (e.g. the validator client may be more trusted than a random peer)

trinity/nodes/beacon/full.py
```diff
@@ -147,6 +151,50 @@ async def _run_validator_api(
...
+    def _on_block_imported(self, block: SignedBeaconBlock) -> None:
+        self.logger.debug("successfully imported block %s", block)
+        self._try_import_orphan(block.message.hash_tree_root)
+        # TODO: synchronize any operations from pools that are now on-chain
+
+    def _on_block_import_failure(
+        self, block: SignedBeaconBlock, exc: Exception
+    ) -> None:
+        self.logger.exception("failed to import block %s: %s", block, exc)
+        # TODO: do not drop block?
```
ralexstokes, line 182: not sure what we would want to do here but it seems that if we get a well-formed block we should *not* just drop it on the floor. it could become slashable later or....? likely depends on the nature of the `exc`

ralexstokes:
merging to keep the dev flow simpler on top of this PR -- happy to address post-merge feedback!

PR #1727, opened 2020-05-15
Log unexpected component errors as soon as we detect them

gsalgado:


trinity/extensibility/asyncio.py
```diff
@@ -61,10 +60,20 @@ async def _do_run(self, boot_info: BootInfo) -> None:
...
                     return
+                except BaseException:
+                    # XXX: Logging (and re-raising!) the exception here is important because
+                    # an unexpected component crash won't take down trinity so the error would go
+                    # unnoticed until the user terminates trinity. Also, because of an issue in
+                    # the ComponentManager (see comment in _cleanup_component_task), when a
+                    # component crashes and requests trinity to shutdown, there's still a chance
+                    # its exception could be lost.
+                    self.logger.exception("Unexpected error on %s, exiting", self)
+                    # XXX: Should we also send a ShutdownRequest here?
```
pipermerriam, line 71: I think *yes* we do, or at least, it seems like the conservative option (though it isn't clear to me whether the client will already crash the client in this case....).

gsalgado: It won't crash trinity, and that's why logging it here becomes so important. Took me a while to track this down and I think it's probably better if we send a `ShutdownRequest` here as leaving trinity running after a component has crashed can lead to unexpected behavior that would be really hard to reproduce/debug.

gsalgado:
@pipermerriam I've also experimented with another approach to try and not discard the component's exception in `run_component()`: https://github.com/ethereum/trinity/pull/1727/commits/a6c62bfd237d9420471906d9c2dc23c5bddc6e19#diff-890a0d799e95b21dc6edfddd4da71e7eR123-R134

However, that introduces an unnecessary delay when terminating well-behaved components, without guaranteeing that we won't discard an uncaught exception, so I think the logging+shutdown-request approach is better. Just thought I'd mention this anyway

PR #1729, opened 2020-05-16
Variety of fixes to get beacon node accepting blocks from validator client

ralexstokes:
A medley of fixes/updates to get the validator client driving the injection of blocks into the beacon node across our existing HTTP API.

(Just hit slot 123!!)

#### Cute Animal Picture

### DIFF
--- a/eth2/api/http/validator.py
+++ b/eth2/api/http/validator.py
@@ -186,9 +186,8 @@ class Context:
         return Attestation.create(aggregation_bits=aggregation_bits, data=data)
 
     async def broadcast_attestation(self, attestation: Attestation) -> bool:
-        # logger.info(
-        #     "broadcasting attestation with root %s",
-        #     humanize_hash(attestation.hash_tree_root),
+        # logger.debug(
+        #     "broadcasting attestation with root %s", attestation.hash_tree_root.hex()
         # )
         # TODO the actual brodcast
         self._broadcast_operations.add(attestation.hash_tree_root)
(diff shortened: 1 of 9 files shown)


eth2/api/http/validator.py
```diff
@@ -102,6 +102,9 @@ async def get_sync_status(self) -> SyncStatus:
     def get_validator_duties(
         self, public_keys: Collection[BLSPubkey], epoch: Epoch
     ) -> Iterable[ValidatorDuty]:
+        if epoch < self.eth2_config.GENESIS_EPOCH:
+            return ()
+
```
ralexstokes, line 107: ignore eager clients

eth2/validator_client/beacon_node.py
```diff
@@ -308,6 +329,12 @@ async def fetch_duties(
         public_keys: Collection[BLSPubkey],
         target_epoch: Epoch,
     ) -> Collection[Duty]:
+        if target_epoch < 0:  # GENESIS_EPOCH == 0
+            # avoid fetching duties before genesis
+            # NOTE: we do want to fetch duties from the genesis epoch
+            # with at least 1 epoch of lookahead.
+            return ()
```
ralexstokes, line 336: we get attestation duties in epoch -1 for epoch 0. the duty fetcher asks for duties in the current epoch and the next epoch.

this means we end up asking for duties in epoch -1 but that is nonsensical so return nothing in this case

trinity/nodes/beacon/full.py
```diff
@@ -36,6 +36,8 @@ def _mk_clock(
         config.SLOTS_PER_EPOCH,
         config.SECONDS_PER_SLOT * config.SLOTS_PER_EPOCH,
         time_provider,
+        # node can wait until genesis
+        genesis_lookahead=0,
```
ralexstokes, line 40: the beacon node does not need the lookahead that the validator client desires... avoids filter those ticks w/in the beacon node down below in this module

Issue #1732, opened 2020-05-19
ETHPeerPool._periodically_report_metrics crashes (with a PeerConnectionLost) and takes trinity with it


gsalgado:
### What is wrong?

Got this crash today.

```
   DEBUG  2020-05-19 15:59:46,855               Manager  <Manager[ETHPeerPool] flags=SRcfe>: task _periodically_report_metrics[daemon=True] exited with error: Cannot look up subprotocol when connection is closing
Traceback (most recent call last):
  File "/home/salgado/src/async-service/async_service/base.py", line 300, in _run_and_manage_task
    await task.run()
  File "/home/salgado/src/async-service/async_service/asyncio.py", line 35, in run
    await self._async_fn(*self._async_fn_args)
  File "/home/salgado/src/trinity/p2p/peer_pool.py", line 505, in _periodically_report_metrics
    self._peer_reporter_registry.trigger_peer_reports()
  File "/home/salgado/src/trinity/p2p/metrics.py", line 54, in trigger_peer_reports
    self.make_periodic_update(peer, peer_id)
  File "/home/salgado/src/trinity/trinity/protocol/common/peer.py", line 171, in make_periodic_update
    head_gauge.set_value(peer.head_info.head_number)
  File "/home/salgado/virtualenvs/trinity-asyncio/lib/python3.7/site-packages/cached_property.py", line 35, in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
  File "/home/salgado/src/trinity/trinity/protocol/common/peer.py", line 105, in head_info
    return self.connection.get_logic(HeadInfo.name, HeadInfo)
  File "/home/salgado/src/trinity/p2p/connection.py", line 265, in get_logic
    if not self.has_logic(name):
  File "/home/salgado/src/trinity/p2p/connection.py", line 261, in has_logic
    raise PeerConnectionLost("Cannot look up subprotocol when connection is closing")
p2p.exceptions.PeerConnectionLost: Cannot look up subprotocol when connection is closing
   DEBUG  2020-05-19 15:59:46,866               Manager  <Manager[ETHPeerPool] flags=SRCfE>: _handle_cancelled triggering task cancellation
```

After that, `EthPeerPool` terminates and takes trinity with it

### How can it be fixed

Guess we need to catch `PeerConnectionLost` in yet another place, @cburgdorf?

gsalgado:
[...]

PR #1736, opened 2020-05-20
Fix PeerConnectionLost related crash in metrics reporting

cburgdorf:
### What was wrong?

See #1732, the whole node crashes when we try to read `head_td` and friends from a peer that is in the process of shutting down. 

### How was it fixed?

Catch the `PeerConnectionLost` error and set gauge to `0`.

### To-Do

- [x] Clean up commit history

- [ ] ~~Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)~~ Unreleased bug

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/common/peer.py
+++ b/trinity/protocol/common/peer.py
@@ -166,15 +166,16 @@ class BaseChainPeerReporterRegistry(PeerReporterRegistry[BaseChainPeer]):
     def make_periodic_update(self, peer: BaseChainPeer, peer_id: int) -> None:
         head_gauge = self._get_blockheight_gauge(peer_id)
         td_gauge = self._get_td_gauge(peer_id)
+        expected_errors = (AttributeError, PeerConnectionLost)
         # set to 0 if head_number unavailable on head_info
         try:
             head_gauge.set_value(peer.head_info.head_number)
-        except AttributeError:
+        except expected_errors:
             head_gauge.set_value(0)
         # set to 0 if head_td unavailable on head_info
         try:
             td_gauge.set_value(peer.head_info.head_td)
-        except AttributeError:
+        except expected_errors:
             td_gauge.set_value(0)
 
     def _get_blockheight_gauge(self, peer_id: int) -> SimpleGauge:
--- a/trinity/protocol/common/peer.py
+++ b/trinity/protocol/common/peer.py
@@ -166,17 +166,28 @@ class BaseChainPeerReporterRegistry(PeerReporterRegistry[BaseChainPeer]):
     def make_periodic_update(self, peer: BaseChainPeer, peer_id: int) -> None:
         head_gauge = self._get_blockheight_gauge(peer_id)
         td_gauge = self._get_td_gauge(peer_id)
-        expected_errors = (AttributeError, PeerConnectionLost)
-        # set to 0 if head_number unavailable on head_info
+
         try:
-            head_gauge.set_value(peer.head_info.head_number)
-        except expected_errors:
+            head_info = peer.head_info
+        except PeerConnectionLost:
             head_gauge.set_value(0)
-        # set to 0 if head_td unavailable on head_info
-        try:
-            td_gauge.set_value(peer.head_info.head_td)
-        except expected_errors:
             td_gauge.set_value(0)
+        else:
+            try:
+                head_number = head_info.head_number
+            except AttributeError:
+                # set to 0 if head_number unavailable on head_info
+                head_gauge.set_value(0)
+            else:
+                head_gauge.set_value(head_number)
+
+            try:
+                head_td = head_info.head_td
+            except AttributeError:
+                # set to 0 if head_td unavailable on head_info
+                td_gauge.set_value(0)
+            else:
+                td_gauge.set_value(head_td)
 
     def _get_blockheight_gauge(self, peer_id: int) -> SimpleGauge:
         return self.metrics_registry.gauge(f"trinity.p2p/peer_{peer_id}_blockheight.gauge")
--- a/trinity/protocol/common/peer.py
+++ b/trinity/protocol/common/peer.py
@@ -173,6 +173,8 @@ class BaseChainPeerReporterRegistry(PeerReporterRegistry[BaseChainPeer]):
             head_gauge.set_value(0)
             td_gauge.set_value(0)
         else:
+            td_gauge.set_value(head_info.head_td)
+
             try:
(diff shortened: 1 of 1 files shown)


cburgdorf:
@gsalgado This should fix #1732 but it will continue to run the periodic check (with noop gauges). We currently do not have information whether the metrics system is enabled or not at this place and I'm not entirely sold that we need it.

trinity/protocol/common/peer.py
```diff
@@ -166,15 +166,16 @@ def reset_peer_meters(self, peer_id: int) -> None:
     def make_periodic_update(self, peer: BaseChainPeer, peer_id: int) -> None:
         head_gauge = self._get_blockheight_gauge(peer_id)
         td_gauge = self._get_td_gauge(peer_id)
+        expected_errors = (AttributeError, PeerConnectionLost)
         # set to 0 if head_number unavailable on head_info
         try:
             head_gauge.set_value(peer.head_info.head_number)
-        except AttributeError:
```
gsalgado, line 172: Catching `AttributeError` is not usually a good idea, but if we really need that, can we at least protect only the specific attribute access that we expect to cause it? I suppose in this case that'd be `peer.head_info`? Otherwise an attribute error from any of the others (`.set_value`, `.head_number`) would be caught here and I'm sure we wouldn't want that

trinity/protocol/common/peer.py
```diff
@@ -166,15 +166,16 @@ def reset_peer_meters(self, peer_id: int) -> None:
...
+        except expected_errors:
             head_gauge.set_value(0)
         # set to 0 if head_td unavailable on head_info
         try:
             td_gauge.set_value(peer.head_info.head_td)
-        except AttributeError:
+        except expected_errors:
```
gsalgado, line 178: ditto

trinity/protocol/common/peer.py
cburgdorf: Yeah, I agree this isn't a bright spot in our code base and I'm not entirely familiar with the underlying mechanics but after further investigation I found that:

1. We have quite a few places in sync that catch the `AttributeError` when accessing `head_number` and friends

2. The `PeerConnectionLost` is raised from 

https://github.com/ethereum/trinity/blob/5741eb12c940200127dd8af21ea6a384a7f63e02/p2p/connection.py#L252-L255

which means it is raised when we try to access `HeadInfo`.

So, if I understand it correctly we only need to catch the `AttributeError`s for `head_number` and `head_td`.

I updated the code accordingly.

trinity/protocol/common/peer.py
```diff
@@ -166,17 +166,28 @@ def reset_peer_meters(self, peer_id: int) -> None:
...
         td_gauge = self._get_td_gauge(peer_id)
-        expected_errors = (AttributeError, PeerConnectionLost)
-        # set to 0 if head_number unavailable on head_info
+
         try:
-            head_gauge.set_value(peer.head_info.head_number)
-        except expected_errors:
+            head_info = peer.head_info
+        except PeerConnectionLost:
             head_gauge.set_value(0)
```
gsalgado, line 173: Do we want to set a value of 0 here as well, or simply skip it as the peer is gone anyway?

trinity/protocol/common/peer.py
```diff
@@ -166,17 +166,28 @@ def reset_peer_meters(self, peer_id: int) -> None:
...
+            try:
+                head_number = head_info.head_number
+            except AttributeError:
+                # set to 0 if head_number unavailable on head_info
+                head_gauge.set_value(0)
+            else:
+                head_gauge.set_value(head_number)
+
+            try:
+                head_td = head_info.head_td
```
gsalgado, line 185: The `head_td` property won't raise an AttributeError like `head_info` does

cburgdorf:
>As we're seeing here, a seemingly innocuous attribute access can raise a `PeerConnectionLost`, which is quite unexpected and I doubt it's documented anywhere.

Yeah, I agree that isn't ideal and I think I've recently seen some discussion around the very same issue just surfaced from a different area of the code base.

>I'm sure there will be more cases like this in other metric-gathering bits, so I think it'd be better if we found a way to avoid executing any of this code unless the user has enabled metrics

I'm not opposed, in fact, I agree we could face places where the *gathering* of metrics is costly and we want to avoid that. I think ideally we would pass the `MetricsServiceAPI` all the way down instead of the `MetricsRegistry` but we can not do that because `MetricsServiceAPI` is defined in trinity and I think it would be akward to move it to `p2p` just for the sake of giving `p2p` access to it (kinda like how the `BaseService` used to live inside of `p2p` before it became its own module).

The other idea I would have is to define a `MetricsConfig`  in `p2p`.

```python

MetricsConfig(NamedTuple):
    enabled: bool
    registry: MetricsRegistry
```

This would give the `p2p` code access to that information without forcing all the metrics code to be moved into `p2p`.

If you'd be happy with that solution or know an alternative I'm happy to address that in a subsequent PR.

trinity/protocol/common/peer.py
cburgdorf: Are you saying only `peer.head_info` raises `AttributeError` (and `PeerConnectionLost`) or did you mean to say `head_info.head_number` can raise `AttributeError` but `head_info.head_td` won't? I lost count :sweat_smile:

gsalgado: You should only need to catch an `AttributeError` from `head_info.head_number`: https://github.com/ethereum/trinity/blob/dfe7b4c227888583744f3901958b8855f6dbf472/trinity/protocol/eth/api.py#L97-L101

cburgdorf: Oh, I see! It wasn't entirely clear to me where the `AttributeError` exactly originated. Thanks for clearing up my confusion!

trinity/protocol/common/peer.py
cburgdorf: I *think* we want to set it to `0`. These metrics are dealing with a fixed number of surrounding peers displayed as `peer_1` - `peer_10`  in the dashboard. That means, if `peer_2` drops of, id `2` will be reassigned to the next peer joining which may be an entirely different peer (yeah, not ideal but we can only have a fixed set of hardcoded charts). In that case it would be extra confusing if `head_number` would jump from e.g `1m` to `2m` directly and so we want it to drop to `0` as soon as a peer disconnects to reflect that the next time `peer_2` starts reporting it may not be the same peer after all.

PR #1743, opened 2020-05-25
Validator client doesn't crash due to unresponsive beacon node.

### LINKED ISSUE #1737: The validator client currently crashes if a request is made against a beacon node that goes down.
A common workflow during development is to have the validator client running and then restart an attached beacon node.

This currently raises an `OSError` which is not caught in the client and crashes the entire process.

Ideally the validator client drops requests that cannot be made when the beacon node is unreachable and keeps moving along.

g-r-a-n-t:
Closes #1737

### What was wrong?

The validator client would crash if a request to a beacon node failed.

### How was it fixed?

The validator client catches these exceptions and logs a warning instead of crashing.

### To-Do

- [ ] Clean up commit history


eth2/validator_client/duty_scheduler.py
```diff
@@ -22,13 +22,19 @@ async def resolve_duty(
...
-        if attestation:
-            await resolved_duties.send((duty, attestation))
+
+        try:
+            attestation = await beacon_node.fetch_attestation(
+                duty.validator_public_key,
+                duty.tick_for_execution.slot,
+                duty.committee_index,
+            )
+        except OSError as err:
```
cburgdorf, line 32: I'm not very familiar with the eth2 code but I just want to leave that generally catching `OSError` in that way accidentally catch a whole range of unrelated errors which can make debugging harder.

If we have to catch the `OSError` then maybe we just catch it here and re-raise it as a more specific error.

https://github.com/ethereum/trinity/blob/32f523128780fc7e24e1582b4bd8d87813396763/eth2/validator_client/beacon_node.py#L50-L51

Something like `BeaconNodeUnresponsive` or I actually wonder if that one would even qualify as a `TimeoutError` :thinking:

g-r-a-n-t: Makes sense. I'm assuming this applies to `_post_json` as well, so I've updated that too.

ralexstokes: `OSError` is along the exception chain when the tcp connection cannot be made -- if there is a more specific exception then let's use that.

ralexstokes: the context here is that the validator client will try to dial the beacon node host and if the port isn't open (or also i imagine the ip is not reachable) then we get an exception. it is definitely worth looking into a more specific exception for this case

cburgdorf: >the context here is that the validator client will try to dial the beacon node host and if the port isn't open (or also i imagine the ip is not reachable) then we get an exception

Correct, which is why something like `BeaconNodeUnreachable` or `TimeoutError` seems more suitable to be used across multiple callsites. Because catching `OSError` among many callsites may accidentially catch unrelated errors.

The current state of the PR reflects the idea of turning the `OSError` early into a `TimeoutError` but I'm not 100 % sold on the idea (even though I brought it up). I think creating a custom `BeaconNodeUnavailable` (derived from `BaseTrinityError`) may be the most straight forward.

g-r-a-n-t: I suppose `TimeoutError` would be misleading. As Alex noted, the node could be offline or running at a different address and therefore unreachable, which is probably going to be the case most of the time. I think `BeaconNodeUnreachable` is clearer, but might be misleading if the request has actually timed out or returned a server error. Maybe we should go with `BeaconNodeRequestFailed`?

cburgdorf: I'm fine with `BeaconNodeRequestFailed` :+1:

ralexstokes:
i'm also honestly not sure if there aren't other places we need to catch this exception if we keep it external the `BeaconNode` class. and we can update the API to always return the object (rather than `Optional[Object]`) if we stick w/ this "exception external to the class" approach.

trinity/exceptions.py
```diff
@@ -148,3 +148,10 @@ class MetricsReportingError(BaseTrinityError):
     Raised when there is an error while reporting metrics.
     """
     pass
+
+
+class BeaconNodeRequestFailure(BaseTrinityError):
```
ralexstokes, line 153: there may be some python conventions i'm missing but i'd think we want to define an exception as close to its possible source of creation as possible w/in the broader module/file hierarchy.... if we moved this concern to be entirely inside the scope of the `BeaconNode` then we can even keep this exception private to _that file/module_

ralexstokes: the general reason to do this is that i think it tends towards better maintainability and discoverability of the code base over time -- imagine you are exploring the trinity codebase and you suddenly have N exception types that tie you across not only different parts of the client stack, but different clients! it would quickly overflow my cognitive stack :)

g-r-a-n-t:
Yeah, I think this makes sense to me for the most part. By dispatcher you mean a class that basically consumes the `_post_json`, `_get_json` and `_get_...` methods in beacon_node.py, right?

PR #1744, opened 2020-05-25
Generalized Pool

### LINKED ISSUE #1331: Consolidate implementation of `OperationPool` and `OrphanBlockPool`
### What is wrong?

We can likely just have one solid implementation of some kind of "object" pool that is used for `Attestations`,  orphan `BeaconBlock`s and whatever else may arise in the future.

### How can it be fixed

Continue the work along the lines of #783 and #1330 but for the `OrphanBlockPool` abstraction.

g-r-a-n-t:
Closes #1331 

### What was wrong?
There is alot of duplicated functionality between `OrphanBlockPool` and `OperationPool`.

### How was it fixed?
I've moved the `OperationPool` class into a new class `HashableContainerPool`, which can be used to create pools for `HashableContainer` objects. By default, the containers are keyed in the pool's `_pool_storage` dict by the container's `hash_tree_root` value. This behavior can be changed though by overriding the  `_get_root` method in a `HashableContainerPool` child class.

In the case of `OrphanBlockPool`, we want to key pool values using the root found in `block.message`, so we override the `_get_root` method like so:

```python
def _get_root(self, block: BaseSignedBeaconBlock) -> Root:
  return block.message.hash_tree_root
```

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


eth2/beacon/operations/pool.py
```diff
@@ -1,50 +0,0 @@
-from typing import Callable, Dict, Generic, Iterable, Iterator, Tuple, TypeVar, Union
-
-from eth2.beacon.typing import Operation, Root
-
-T = TypeVar("T", bound=Operation)
```
ralexstokes, line 5: anything wrong w/ the narrowly-scoped type `Operation`?

eth2/beacon/hashable_container_pool.py
```diff
@@ -0,0 +1,55 @@
...
+
+from ssz.hashable_container import HashableContainer
+
+from eth2.beacon.typing import Root
+
+T = TypeVar("T", bound=HashableContainer)
+
+
+class HashableContainerPool(Generic[T]):
+    _pool_storage: Dict[Root, T]
```
ralexstokes, line 11: i made another pass at this in the latest iteration of the beacon node and i'm quite tempted to just subclass `set`.

we are basically rewriting a `set` impl w/ this approach.... something to consider

eth2/beacon/hashable_container_pool.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def batch_add(self, containers: Iterable[T]) -> None:
+        self._batch_do(self.add, containers)
+
+    def remove(self, container: T) -> None:
+        root = self._get_root(container)
+        if root in self._pool_storage:
+            del self._pool_storage[root]
+
+    def batch_remove(self, containers: Iterable[T]) -> None:
```
ralexstokes, line 54:
```suggestion
    def batch_remove(self, containers: *T) -> None:
```

i believe this will typecheck and in an option to consider towards a more ergonomic api (accepting variadic operations)... if you want to update this here then we will want to do it everywhere in this class defn that we accept an` Iterable[T]` as a collection of operations

ralexstokes:
also i'm not entirely sure we need to override the `key` functionality, i.e. direct to the inner message on a "signed" operation -- in which case we can drop that for now -- we would need to double-check the spec but so far we have been getting along fine working directly w/ the signed operation and deferring to the inner operation as needed...

eth2/beacon/operations/pool.py
g-r-a-n-t: no, some confusion over how python worked led me to make this unnecessary change

eth2/beacon/hashable_container_pool.py
g-r-a-n-t: wait.. `set` or `dict`? we still want the key-value behavior, right

g-r-a-n-t: I was running into some trouble making `AttestationPool` and `OrphanBlockPool` subclasses of `Dict[Root, T]`. For example, the iter signature for these pools is defined as `__iter__(self) -> Iterator[Tuple[Root, T]]` , which is different than dict's iter method. So if we subclass dict, we would either need to refactor all of the places where iter is used or override the iter method (which upsets mypy, so not really an option). The same goes for `__contains__`, as it deviates from the method defined in dict.

I would be willing to go and make these changes, but wanted to run this by you before getting into it.

eth2/beacon/hashable_container_pool.py
g-r-a-n-t: Do you mean this?

```python
def batch_remove(self, *T) -> None:
  ...

# So that we can do this
batch_remove(operation_set1, operation_list2, operation_dict3.values())
```

g-r-a-n-t:
Hey Alex, I was able to tighten this up quite a bit.

eth2/beacon/hashable_container_pool.py
ralexstokes: yeah if those args are collections have to be unpacked at the call site but this is what i mean :)

```python
batch_remove(*op_set1, *op_list2, *op_dict3.values())
```

ralexstokes: the part i was fuzzy on was the typing and now that i'm looking at it it is the other way around:

```python
def batch_remove(self, *containers: T) -> None:
```

tests/eth2/core/beacon/fork_choice/test_lmd_ghost.py
```diff
@@ -594,16 +593,6 @@ def test_lmd_ghost_fork_choice_scoring(
...
-    attestations = tuple(
-        _get_attestation_from_block(block)
-        for block in _iter_block_tree_by_block(block_tree)
-        if _get_attestation_from_block(block)
-    )
-
-    attestation_pool = empty_attestation_pool
-    for attestation in attestations:
-        attestation_pool.add(attestation)
-
```
ralexstokes, line 606: i'm confused why this was here to begin with as it looks like it is not being used.. definitely delete!

eth2/beacon/hashable_container_pool.py
ralexstokes: yeah i'm not sure we _need_ any sort of mapping functionality given the equivalence between a `T: Operation` and a `Root` (to get the `Root`, just call `t.hash_tree_root` for some `t: T`).

we may need to add it back but i'd say we can drop it for now

g-r-a-n-t:
Someone with write access will need to merge this for me :smile:

PR #1750, opened 2020-05-27
Multiplexer now consumes transport messages in a background task

gsalgado:
This way we don't have to deal with pausing/resuming, which led to issues like #1001, and it also fixes a subtle bug I'd recently identified in `Connection`, in which any errors from the multiplexer were only reported once the `Connection` was cancelled. Now the connection starts its daemon tasks and blocks on an `await multiplexer.wait_streaming_finished()` which will raise any multiplexing errors as soon as they happen

Closes: #1001

### DIFF
--- a/p2p/handshake.py
+++ b/p2p/handshake.py
@@ -280,14 +280,28 @@ async def negotiate_protocol_handshakes(transport: TransportAPI,
     # the `Transport` and feeds them into protocol specific queues.  Each
     # protocol is responsible for reading its own messages from that queue via
     # the `Multiplexer.stream_protocol_messages` API.
-    async with multiplexer.multiplex():
-        # Concurrently perform the handshakes for each protocol, gathering up
-        # the returned receipts.
+    await multiplexer.stream_in_background()
+    # Concurrently perform the handshakes for each protocol, gathering up
+    # the returned receipts.
+    try:
         protocol_receipts = cast(Tuple[HandshakeReceiptAPI, ...], await asyncio.gather(*(
             handshaker.do_handshake(multiplexer, protocol)
             for handshaker, protocol
             in zip(selected_handshakers, selected_protocols)
         )))
+    except BaseException as handshake_err:
+        # If the multiplexer has a streaming error, that will certainly be the cause of
+        # whatever handshake error we got, so raise that instead.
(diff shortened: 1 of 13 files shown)


p2p/handshake.py
```diff
@@ -280,14 +280,32 @@ async def negotiate_protocol_handshakes(transport: TransportAPI,
...
+    # Concurrently perform the handshakes for each protocol, gathering up
+    # the returned receipts.
+    try:
         protocol_receipts = cast(Tuple[HandshakeReceiptAPI, ...], await asyncio.gather(*(
             handshaker.do_handshake(multiplexer, protocol)
             for handshaker, protocol
             in zip(selected_handshakers, selected_protocols)
         )))
+    except BaseException as handshake_err:
+        multiplex_err = multiplexer.get_streaming_error()
```
pipermerriam, line 293: I think this code block could be simplified to:

```python
multiplexer.raise_if_streaming_error()
await multiplexer.stop_streaming()
raise handshake_err
```

I think the code above is logically equivalent to the current `if/else` construct.

p2p/multiplexer.py
```diff
@@ -290,96 +287,57 @@ async def _stream_protocol_messages(self,
...
-                self,
-                stop.is_set(),
-            )
+                "Timed out waiting for command from %s, exiting...", self,)
             self.logger.debug("Timeout %r: %s", self, exc, exc_info=True)
-        except CorruptTransport as exc:
-            self.logger.error("Corrupt transport, while multiplexing %s: %r", self, exc)
-            self.logger.debug("Corrupt transport, multiplexing trace: %s", self, exc_info=True)
+        finally:
+            self._is_streaming.clear()
```
pipermerriam, line 335: Why do we clear the event?  Is the multiplexer now re-entrant (can be started, stopped, re-started)?

gsalgado: No, that was only because I was using it in the `@is_streaming` property, but I've changed that to `return not self_streaming_task.done()` and renamed `_is_streaming` to `_started_streaming` so now it is only used in `stream_in_background()` and we no longer need to clear it

p2p/handshake.py
gsalgado: It is indeed!

PR #1753, opened 2020-05-27
Beacon Chain spec update v0.9.4 -> 0.11.3

g-r-a-n-t:
### What was wrong?
The Beacon Node needs to be updated according to the v0.11.3 spec.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


eth2/beacon/genesis.py
```diff
@@ -52,8 +52,15 @@ def initialize_beacon_state_from_eth1(
     deposits: Sequence[Deposit],
     config: Eth2Config
 ) -> BeaconState:
+    fork = Fork(
+        previous_version=config.GENESIS_FORK_VERSION,
+        current_version=config.GENESIS_FORK_VERSION,
+        epoch=config.GENESIS_EPOCH,
```
ralexstokes, line 58:
```suggestion
        # as a constant, right?
        epoch=GENESIS_EPOCH,
```

eth2/beacon/helpers.py
```diff
@@ -224,3 +231,14 @@ def compute_fork_digest(
     return ForkDigest(
         compute_fork_data_root(current_version, genesis_validators_root)[:4]
     )
+
+
+def compute_signing_root(container: HashableContainer, domain: Domain) -> Root:
```
ralexstokes, line 236: `HashableContainer` is fine here but this could more generally just be anything that yields a `hash_tree_root` and i think our nicest type for that is something like `ssz.Serializeable`

eth2/_utils/bls/backends/noop.py
```diff
@@ -10,32 +10,21 @@
 
 class NoOpBackend(BaseBLSBackend):
     @staticmethod
-    def privtopub(k: int) -> BLSPubkey:
+    def SkToPk(k: int) -> BLSPubkey:
         return BLSPubkey(k.to_bytes(48, "little"))
 
     @staticmethod
-    def sign(message_hash: Hash32, privkey: int, domain: Domain) -> BLSSignature:
+    def Sign(**kwargs) -> BLSSignature:
```
ralexstokes, line 17: its a bit annoying but prolly better in the long run to use the concrete interface here, rather than just smear over them w/ `**kwargs`... i had intended to give this entire module some thought as i think we can simplify here a bit.... i think a lot of the places where we needed to (defensively) subclass bits of behavior are due to legacy behavior that dissolves away as we move towards multiple implementations of a single standard...

eth2/beacon/helpers.py
ralexstokes: speaking of.... we did define an `Operation` type in iirc `eth2.beacon.typing`... can we use this protocol type here?

while the spec may not do this, we could definitely pass in a valid ssz object that has a `hash_tree_root` that is not a `Container` type

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -144,6 +162,7 @@ def validate_proposer_slashing(
     """
     proposer = state.validators[proposer_slashing.proposer_index]
 
+    # NOTE: spec has: assert header_1.slot == header_2.slot
```
ralexstokes, line 165: ohh nice catch! we want to update this to compare by slot, rather than epoch -- i can dig up the spec issue concerning this if you'd like...

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -185,6 +204,13 @@ def validate_proposer_slashing_epoch(
 def validate_proposer_slashing_headers(proposer_slashing: ProposerSlashing) -> None:
     header_1 = proposer_slashing.signed_header_1
     header_2 = proposer_slashing.signed_header_2
+
+    if header_1.proposer_index != header_2.proposer_index:
+        raise ValidationError(
+            f"proposer_slashing.signed_header_1.proposer_index ({header_1.proposer_index}) != "
+            f"proposer_slashing.signed_header_2.proposer_index ({header_2.proposer_index})"
+        )
+
```
ralexstokes, line 213: stylistically (i.e. keeping w/ the organization of the rest of the code here as mapped from the spec), this would be a separate function corresponding to the spec condition

so `def validate_proposer_slashing_proposer_index` or something that raises `ValidationError` upon the opposite of the spec condition.

eth2/beacon/state_machines/forks/serenity/configs.py
```diff
@@ -21,21 +24,21 @@
     EJECTION_BALANCE=Gwei(2 ** 4 * GWEI_PER_ETH),  # (= 16,000,000,000) Gwei
     EFFECTIVE_BALANCE_INCREMENT=Gwei(2 ** 0 * GWEI_PER_ETH),  # (= 1,000,000,000) Gwei
     # Initial values
-    GENESIS_SLOT=Slot(0),
-    GENESIS_EPOCH=Epoch(0),
+    GENESIS_FORK_VERSION=Version(b"\x00" * 4),
     BLS_WITHDRAWAL_PREFIX=0,
```
ralexstokes, line 28: noting that we will want to make this a `byte` and that has some typing implications for the places where we use it in the state transition function...

eth2/beacon/tools/builder/initializer.py
```diff
@@ -280,6 +280,6 @@ def create_mock_validator(
         pubkey, withdrawal_credentials, config.MAX_EFFECTIVE_BALANCE, config
     )
     if is_active:
-        return activate_validator(validator, config.GENESIS_EPOCH)
+        return activate_validator(validator, GENESIS_EPOCH)
```
ralexstokes, line 283: general comment on this edit (making the GENESIS_{SLOT, EPOCH} constants) --- is that we can now do some "constant folding" and just import them at use, rather than passing an extra fn arg

i'd be fine if we want to punt on this and do it in a follow up PR -- but would also want to make a tracking issue here for it if so

eth2/beacon/tools/builder/state.py
```diff
@@ -56,7 +57,7 @@ def create_mock_genesis_state_from_validators(
     _check_distinct_pubkeys(genesis_validators)
     _check_no_missing_balances(genesis_validators, genesis_balances)
     _check_sufficient_balance(genesis_balances, config.MAX_EFFECTIVE_BALANCE)
-    _check_activated_validators(genesis_validators, config.GENESIS_EPOCH)
+    _check_activated_validators(genesis_validators, GENESIS_EPOCH)
```
ralexstokes, line 60: another example for "constant folding"

eth2/beacon/tools/misc/ssz_vector.py
```diff
@@ -25,7 +25,6 @@ def _mk_overrides(config: Eth2Config) -> Dict[HashableContainer, Dict[str, int]]
             "block_roots": config.SLOTS_PER_HISTORICAL_ROOT,
             "state_roots": config.SLOTS_PER_HISTORICAL_ROOT,
             "historical_roots": config.HISTORICAL_ROOTS_LIMIT,
-            "eth1_data_votes": config.SLOTS_PER_ETH1_VOTING_PERIOD,
```
ralexstokes, line 28: we removed this bc we are re-using another of the limits? or should there be a `config.EPOCHS_PER_ETH1_VOTING_PERIOD` now?

eth2/beacon/types/block_headers.py
```diff
@@ -28,12 +29,14 @@ def create(
         cls: Type[TBeaconBlockHeader],
         *,
         slot: Slot = default_slot,
-        parent_root: Root = ZERO_ROOT,
-        state_root: Hash32 = ZERO_HASH32,
-        body_root: Hash32 = ZERO_HASH32,
```
ralexstokes, line 33: nice clean up!

eth2/beacon/types/signing_root.py
```diff
@@ -0,0 +1,37 @@
...
+from eth_utils import humanize_hash
+from ssz.hashable_container import HashableContainer
+from ssz.sedes import bytes32
+
+from eth2.beacon.typing import Root, Domain, default_root
+
+TSigningRoot = TypeVar("TSigningRoot", bound="SigningRoot")
+
+
+class SigningRoot(HashableContainer):
```
ralexstokes, line 12: while we are here -- this was changed to `SigningData` in v0.12.0 and should not affect interop w/ `v0.11.3` clients so we can save some future work by doing this now!

ralexstokes: ref: https://github.com/ethereum/eth2.0-specs/pull/1740

eth2/beacon/typing.py
```diff
@@ -34,6 +34,7 @@ def __str__(self) -> str:
 
 Root = NewType("Root", Hash32)  # a Merkle root
 
+Domain = NewType("Domain", Hash32)  # a signature domain
 
 #
 #  Networkinig
```
ralexstokes, line 40:
```suggestion
#  Networking
```

setup.py
```diff
@@ -114,15 +114,15 @@
...
+        "py-ecc==3.0.0",
         "rlp>=1.1.0,<2.0.0",
         PYEVM_DEPENDENCY,
         "ssz==0.2.4",
         "asks>=2.3.6,<3",  # validator client
         "eth-keyfile",  # validator client
     ],
     'eth2-extra': [
-        "milagro-bls-binding==0.1.4",
+        "milagro-bls-binding==1.3.0",
```
ralexstokes, line 125: awesome -- wasn't sure which versions correspond to `v0.11.3` bls!

eth2/beacon/tools/misc/ssz_vector.py
ralexstokes: FYI looks like we need to update the `BeaconState` type and this function will be updated accordingly as well

eth2/beacon/helpers.py
g-r-a-n-t: I think `Operation` makes sense... that's what I'll go with for now

eth2/beacon/typing.py
```diff
@@ -76,3 +77,17 @@ class Operation(Protocol):
...
+
+    def __init__(self, slot: Slot):
+        self.hash_tree_root = get_hash_tree_root(slot, sedes=uint64)
+
+
+class EpochOperation(Operation):
+    hash_tree_root: Root
+
+    def __init__(self, epoch: Epoch):
+        self.hash_tree_root = get_hash_tree_root(epoch, sedes=uint64)
```
ralexstokes, line 93: when we say `Operation` we are referring to one of the "system-level" types of state transitions in eth2 that can be included in a block: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#operations

so it is a little confusing to make these into operations in my opinion as someone who isn't as familiar w/ the spec may not understanding why we are doing this

given that we do pass `slot` and `epoch` as arguments to `get_signing_data` i'd suggest we just punt the "can yield a hash tree root" property to the caller in those scenarios by constructing an ssz object in place:

```python
slot = Slot(223432)
return get_signing_data(ssz.sedes.uint64(slot), domain)
```

a broader refactoring we have been playing around with for some time is to make these types full ssz-types (like in the spec, actually) so that `class Slot(ssz.sedes.uint64)` but for historical reasons haven't done that yet.

i'll agree that we are still muddying what we mean in the code w/ the `Operation` typing when we do it this way; a better typing for the signing data function may just be `ssz.Serializable`, or whichever class we have that confers the `hash_tree_root` property (which is broader than the space of possible "operations")

eth2/_utils/bls/backends/milagro.py
```diff
@@ -61,7 +61,7 @@ def FastAggregateVerify(
             )
 
         return FastAggregateVerify(
-            PKs,
+            list(PKs),
```
ralexstokes, line 64: random but we generally prefer immutable structures across the entire codebase where possible as a stylistic thing -- with the rationale that it gives us all the benefits of data structures that don't change under our feet :)

so i'd suggest `tuple(PKs)`

i'd also have to look further at this API but something that could be nicer is taking advantage of varargs:

```python
def FastAggregateVerify(message: bytes32, signature: BLSSignature, *public_keys: BLSPubKey) -> bool:
```

tests/eth2/core/beacon/chains/test_beacon_chain.py
```diff
@@ -100,6 +100,7 @@ def test_get_state_by_slot(valid_chain, genesis_block, genesis_state, config, ke
     )
 
 
+"""
```
ralexstokes, line 103: let's be sure to get these passing before we merge this in

eth2/_utils/bls/backends/milagro.py
g-r-a-n-t: Anything other than a list type being passed into this [function](https://github.com/ChihChengLiang/milagro_bls_binding/blob/v1.0.2/src/lib.rs#L87) throws a TypeError, so I guess there's not really an option here.

I think the suggested signature looks good. For now I've just stuck with what's written in the spec, but I can change them.

I'm wondering if you'd agree with these other two changes:
1. rename all SK and PK parameters to their full names like in your example
2. make all of the function names snake case instead of upper camel case

eth2/_utils/bls/backends/py_ecc.py
```diff
@@ -1,54 +1,41 @@
...
+    @staticmethod
+    def Verify(PK: BLSPubkey, message: Hash32, signature: BLSSignature) -> bool:
+        return G2ProofOfPossession.Verify(PK, message, signature)
 
     @staticmethod
-    def aggregate_signatures(signatures: Sequence[BLSSignature]) -> BLSSignature:
+    def Aggregate(signatures: Sequence[BLSSignature]) -> BLSSignature:
         # py_ecc use a different EMPTY_SIGNATURE. Return the Trinity one here:
         if len(signatures) == 0:
             return EMPTY_SIGNATURE
```
ralexstokes, line 34: while i'm looking at this, the "zero" signature is `0xc0` with 47 trailing `0x0` bytes.

we should probably update the `EMPTY_SIGNATURE` although i'm not super sure where we are using this either way

ralexstokes:
this wasn't linked into the PR for whatever reason:

https://github.com/ethereum/trinity/commit/0f9064e5b269b8dc3fdeaded62ddb4f8f730b6db#r39868216

eth2/_utils/bls/backends/milagro.py
ralexstokes: well we could probably work to change that rust type to something more acceptable but it is probably not that important all things considered.

yeah -- in general we can use whatever functions internal to this `bls` module but i think we should be idiomatic for consumers of the `bls` module

eth2/beacon/tools/fixtures/loading.py
```diff
@@ -69,6 +69,9 @@ def load_config_at_path(p: Path) -> Eth2Config:
         return config_cache[p]
 
     config_data = load_yaml_at(p)
+    config_data["GENESIS_FORK_VERSION"] = config_data["GENESIS_FORK_VERSION"].to_bytes(
+        4, byteorder="big"
+    )
```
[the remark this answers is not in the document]
g-r-a-n-t: Just `BLS_WITHDRAWAL_PREFIX` and the domain type values. The domain type values are ignored in this config file and hardcoded [somewhere else](https://github.com/g-r-a-n-t/trinity/blob/0.11.3-spec/eth2/beacon/signature_domain.py) as an integer enum named SignatureDomain. This enum is missing two domain type values that are included in the spec: `DOMAIN_SELECTION_PROOF` and `DOMAIN_AGGREGATE_AND_PROOF`. They're converted to bytes later on when computing domains. I can add these domain values to the signature domain enum, but it doesn't look like they're used anywhere.

The withdrawal prefix config value was specified as bytes in v0.9.4 and was still treated as an integer in Trinity. Is this something we want to keep, or should withdrawal prefix be changed to bytes as well? Or would we rather just treat `GENESIS_FORK_VERSION` as an integer?

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -406,7 +407,11 @@ def _validate_aggregation_bits(
 ) -> None:
     data = attestation.data
     committee = get_beacon_committee(state, data.slot, data.index, config)
-    if not (len(attestation.aggregation_bits) == len(committee)):
+
+    if not any(attestation.aggregation_bits):
+        raise ValidationError("The attestation bit length is empty.")
```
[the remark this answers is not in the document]
g-r-a-n-t: Ok, thanks for providing some clarity here. Let me update the exception.

eth2/beacon/typing.py
g-r-a-n-t: how does this refactor look https://github.com/ethereum/trinity/pull/1753/commits/f9eae27de49c932f2a0924928ed3d06e1f947636?

`ssz.sedes.uint64` on its own is not callable, so I created a new serializable object for both slots and epochs that wraps around an int value and makes it serializable.

eth2/_utils/bls/backends/py_ecc.py
g-r-a-n-t: > the "zero" signature is 0xc0 with 47 trailing 0x0 bytes

I'm not quite sure what you're referencing here?

eth2/beacon/tools/fixtures/loading.py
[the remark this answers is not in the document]
ralexstokes: > DOMAIN_SELECTION_PROOF and DOMAIN_AGGREGATE_AND_PROOF

yeah we will want these -- technically we would read these from the config but i think it is fine to keep them as constants in the `SignatureDomain` type

eth2/_utils/bls/backends/py_ecc.py
ralexstokes:
```python
b"\x00" * 48
```

is not a valid signature -- the closest thing would be the encoding of the formal zero in G2 which ends up being 

```python
b"\xc0" + b"\x00" * 47
```

The complication here is that i'm not sure how we are currently using `EMPTY_SIGNATURE` and if it matters... if we are using it in a way that some consumer of this code expects a valid BLS signature then we should, in fact, use a valid signature, otherwise we will hit an edge case and the client will fall over :(

eth2/api/http/validator.py
```diff
@@ -192,11 +194,17 @@ def get_block_proposal(
         parent_state = self.chain.advance_state_to_slot(target_slot, parent_state)
```
ralexstokes, line 194: `target_slot` may actually not be a very clear name here... it is more of the "last slot with a block".

eth2/_utils/bls/backends/py_ecc.py
g-r-a-n-t: Ok, I see. Do you mean `b"\xc0" + b"\x00" * 95`? There are a few more bytes in the BLSSignature type.

```
BLSSignature | Bytes96 | a BLS12-381 signature
```

> The complication here is that i'm not sure how we are currently using EMPTY_SIGNATURE and if it matters...

Yeah, if it's not needed, I'd like to just remove this stuff :) However, I'm not really sure why it was added or what could have changed to make it unnecessary.

eth2/_utils/bls/backends/milagro.py
g-r-a-n-t: Cleaned up here https://github.com/ethereum/trinity/pull/1753/commits/0fead5a52d87efb4c9244db09d230ef9ecec9bba

eth2/beacon/typing.py
```diff
@@ -67,6 +61,17 @@ class FromBlockParams(NamedTuple):
 default_version = Version(b"\x00" * 4)
 default_fork_digest = ForkDigest(b"\x00" * 4)
 default_root = Root(Hash32(b"\x00" * 32))
+default_domain = Domain(Hash32(b"\x00" * 32))
```
ralexstokes, line 64:
```suggestion
default_domain = Domain(ZERO_HASH32)
```

ralexstokes: and it would be nicer to use this for `default_root` as well

eth2/validator_client/client.py
```diff
@@ -108,7 +108,6 @@ async def _verify_client_state_at_boot(self) -> None:
         """
         # TODO: sanity check the validator key pairs we find are valid
         # perhaps by fetching their indices in the state
-        pass
```
ralexstokes, line 111: ?

setup.py
```diff
@@ -113,15 +113,15 @@
         "cytoolz>=0.9.0,<1.0.0",
         "eth-typing>=2.1.0,<3.0.0",
         "lru-dict>=1.1.6",
-        "py-ecc==1.7.1",
+        "py-ecc==3.0.0",
```
ralexstokes, line 116: at the risk of breaking everything -- what happens if this is version `4.0.0`?

tests-trio/trinity/nodes/beacon/test_full.py
```diff
@@ -19,18 +20,24 @@ async def test_beacon_node_can_count_slots(autojump_clock, eth2_config, beacon_n
...
 @to_tuple
 def _mk_minimum_viable_signed_beacon_blocks(
-    slots, state_machine_provider, state, block
+    slots, state_machine_provider, state, block, config
 ):
     for slot in slots:
+        future_state = state_machine_provider(slot)\
+            .state_transition\
+            .apply_state_transition(state, future_slot=block.slot + 1)
+        proposer_index = get_beacon_proposer_index(future_state, config)
```
ralexstokes, line 29: did `black` introduce these line continuations `\`? just curious...

tests/eth2/core/beacon/test_epoch_processing_helpers.py
```diff
@@ -222,6 +222,7 @@ def test_get_matching_head_attestations(genesis_state, config):
     attestations = get_matching_head_attestations(state, some_epoch, config)
 
     assert attestations == target_attestations
+"""
```
ralexstokes, line 226: ?

tests/eth2/core/beacon/tools/builder/test_validator.py
```diff
@@ -47,19 +43,11 @@ def test_aggregate_votes(votes_count, random, privkeys, pubkeys):
...
     ]
     assert len(voted_index) == len(votes)
-
-    aggregated_pubs = bls.aggregate_pubkeys(pubs)
-
-    if votes_count == 0:
-        with pytest.raises(ValidationError):
-            bls.validate(message_hash, aggregated_pubs, sigs, domain)
-    else:
-        bls.validate(message_hash, aggregated_pubs, sigs, domain)
```
ralexstokes, line 65: are we not checking the BLS validation anymore?

tests/eth2/utils-tests/bls-utils/test_backends.py
```diff
@@ -96,103 +68,6 @@ def test_invalid_private_key(backend, privkey, domain):
...
-    pubs_2 = [bls.privtopub(k) for k in privkeys_2]
-    aggpub_2 = bls.aggregate_pubkeys(pubs_2)  # sig_2 to msg_2
-
-    message_hashes = [msg_1, msg_2]
-    pubs = [aggpub_1, aggpub_2]
-    aggsig = bls.aggregate_signatures(sigs_1 + sigs_2)
-
-    assert bls.verify_multiple(
-        pubkeys=pubs, message_hashes=message_hashes, signature=aggsig, domain=domain
-    )
```
ralexstokes, line 198: did we just delete all these tests?

eth2/beacon/tools/fixtures/loading.py
```diff
@@ -52,6 +69,12 @@ def load_config_at_path(p: Path) -> Eth2Config:
         return config_cache[p]
 
     config_data = load_yaml_at(p)
+    config_data["GENESIS_FORK_VERSION"] = config_data["GENESIS_FORK_VERSION"].to_bytes(
+        4, byteorder="big"
```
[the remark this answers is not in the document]
g-r-a-n-t: looks like youre correct

tests-trio/trinity/nodes/beacon/test_full.py
g-r-a-n-t: My editor just added them automatically... I guess it's preferred to break up lines in Python with parens

eth2/beacon/tools/fixtures/loading.py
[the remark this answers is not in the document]
g-r-a-n-t: maybe not, this caused fixture test failures. Ill dig a bit more before resolving these

tests/eth2/core/beacon/test_epoch_processing_helpers.py
g-r-a-n-t: yeah, missed this

tests/eth2/utils-tests/bls-utils/test_backends.py
g-r-a-n-t: I had removed them earlier after changing the BLS module and planned on rewriting them. It seems like they duplicate the testing done with the fixtures, so I didn't bother readding them. Maybe it's still worth having them since the `validate` function isn't actually being tested in fixtures.

eth2/beacon/tools/fixtures/loading.py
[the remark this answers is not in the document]
g-r-a-n-t: I guess in this case, `YAML(..).load(path)` will load hex values as big-endian integers, so we need to convert it back to bytes as big-endian.

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -388,7 +390,13 @@ def _validate_aggregation_bits(
 ) -> None:
     data = attestation.data
     committee = get_beacon_committee(state, data.slot, data.index, config)
-    if not (len(attestation.aggregation_bits) == len(committee)):
+
+    if not any(attestation.aggregation_bits):
+        raise ValidationError(
+            f"No bits set in attestation's aggregation bits, attestation={attestation}."
+        )
```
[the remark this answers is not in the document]
g-r-a-n-t: It was added for a similar [test](https://github.com/ethereum/eth2.0-specs/blob/v0.11.3/tests/core/pyspec/eth2spec/test/phase_0/block_processing/test_process_attestation.py#L261) in v0.11.3 , which is using the correct aggregation_bits length.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -484,12 +493,17 @@ def _determine_next_eth1_votes(
 def _update_effective_balances(
     state: BeaconState, config: Eth2Config
 ) -> Tuple[Validator, ...]:
-    half_increment = config.EFFECTIVE_BALANCE_INCREMENT // 2
+    hysteresis_increment = (
+        config.EFFECTIVE_BALANCE_INCREMENT // config.HYSTERESIS_QUOTIENT
+    )
+    downward_threshold = hysteresis_increment * config.HYSTERESIS_DOWNWARD_MULTIPLIER
+    upward_threshold = hysteresis_increment * config.HYSTERESIS_UPWARD_MULTIPLIER
     new_validators = state.validators
```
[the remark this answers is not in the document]
g-r-a-n-t: linked here: https://github.com/ethereum/trinity/issues/1794

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -288,16 +289,21 @@ def get_attestation_deltas(
...
                     index,
                     lambda balance, delta: balance + delta,
-                    get_base_reward(state, index, config)
-                    * attesting_balance
-                    // total_balance,
+                    (
+                        get_base_reward(state, index, config)
+                        * (attesting_balance // increment)
+                    )
+                    // (total_balance // increment),
```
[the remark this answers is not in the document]
g-r-a-n-t: linked here: https://github.com/ethereum/trinity/issues/1794

eth2/beacon/state_machines/forks/serenity/block_validation.py
```diff
@@ -61,6 +60,25 @@ def validate_block_slot(state: BeaconState, block: BaseBeaconBlock) -> None:
...
+def validate_proposer_index(
+    state: BeaconState, block: BaseBeaconBlock, config: Eth2Config
+) -> None:
+    expected_proposer = get_beacon_proposer_index(state, config)
+    if block.proposer_index != expected_proposer:
+        raise ValidationError(
+            f"block.proposer_index "
+            f"({block.proposer_index}) does not equal expected_proposer ({expected_proposer}) "
+            f"at block.slot {state.slot}"
+        )
```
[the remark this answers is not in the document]
g-r-a-n-t: linked here: https://github.com/ethereum/trinity/issues/1794

tests/eth2/core/beacon/tools/builder/test_validator.py
g-r-a-n-t: readded here with the new BLS module https://github.com/ethereum/trinity/pull/1753/commits/644443cdb38aba5838930240ed503cc96401841f

setup.py
g-r-a-n-t: Did some more digging on this, we actually want `2.0.0`... The py_ecc backend doesn't get tested by default

g-r-a-n-t: changing to `4.0.0` or `3.0.0` breaks sign/verify

eth2/beacon/state_machines/forks/serenity/block_validation.py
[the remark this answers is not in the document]
ralexstokes: my hesitation here is that this is ultimately handled in `is_valid_indexed_attestation` in v0.12.x -- i'm having trouble finding where this condition is actually implemented in the spec in v0.11.3 and it is possible that it wasn't -- refer to spec issue 1780.

we can keep this here but let's just remember to clean it up when we implement v0.12.x

eth2/beacon/tools/fixtures/loading.py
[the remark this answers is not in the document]
ralexstokes: ah ok

tests/eth2/utils-tests/bls-utils/test_backends.py
ralexstokes: i'm fine dropping tests if we know there are tests already in the fixture suite, although there is argument to be made for having tests specific to our BLS implementation that could go beyond what the fixtures are looking at or could care to know about...

if we know there is stuff we aren't testing that was tested here, let's restore those tests (and at a certain point it may be less work to just update the APIs in this file? -- your call)

PR #1754, opened 2020-05-28
Fix incorrect tox env setup

cburgdorf:
### What was wrong?

I introduced a bug when I swapped Py36 against Py38 support where we ended up creating the wrong tox environments for some jobs. 

### How was it fixed?

Swapped `py37` against `py38`

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,7 @@ envlist=
     py{37,38}-{eth1-core,p2p,p2p-trio,integration,lightchain_integration,eth2-core,eth2-fixtures,eth2-integration,eth1-components,eth2-utils,eth2-trio}
     py37-long_run_integration
     py37-rpc-blockchain
-    py38-rpc-state-{frontier,homestead,tangerine_whistle,spurious_dragon,byzantium,constantinople,petersburg,istanbul}
+    py37-rpc-state-{frontier,homestead,tangerine_whistle,spurious_dragon,byzantium,constantinople,petersburg,istanbul}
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -3,7 +3,7 @@ envlist=
     py{37,38}-{eth1-core,p2p,p2p-trio,integration,lightchain_integration,eth2-core,eth2-fixtures,eth2-integration,eth1-components,eth2-utils,eth2-trio}
     py37-long_run_integration
     py37-rpc-blockchain
-    py38-rpc-state-{frontier,homestead,tangerine_whistle,spurious_dragon,byzantium,constantinople,petersburg,istanbul}
+    py37-rpc-state-{frontier,homestead,tangerine_whistle,spurious_dragon,byzantium,constantinople,petersburg,istanbul}
```
cburgdorf, line 6: @carver I'm wondering why we need this `envlist` section at all? Is this for when one runs `tox` without specifying any specific environment to run so that it knows the full list of environments to run? Wondering if we could simply drop it.

carver: > Is this for when one runs `tox` without specifying any specific environment to run so that it knows the full list of environments to run?

Yup, also to show up in `tox -l`.



> Wondering if we could simply drop it.

I occasionally use `tox -l`, so I'm a soft :+1: to keep it.

cburgdorf:
This was fixed from another PR.

Issue #1733, opened 2020-05-19
Daemon task Connection._feed_protocol_handlers exits while connection is still running


gsalgado:
### What is wrong?

Got this crash because a daemon task exited while the service was still running:

```
DEBUG  2020-05-19 08:15:44,319               Manager  <Manager[Connection-<Session <Node(0x2591b6@13.228.72.242)> 5188e42d-65ee-4ebb-9e77-8cc447c61b66>] flags=SRcfe>: task _feed_protocol_handlers[daemon=True] exited with error: Daemon task _feed_protocol_handlers[daemon=True] exited
...
DEBUG  2020-05-19 08:15:44,397               Manager  <Manager[ETHPeerPool] flags=SRcfe>: task Connection-<Session <Node(0x2591b6@13.228.72.242)> 5188e42d-65ee-4ebb-9e77-8cc447c61b66>[daemon=False] exited with error: Daemon task _feed_protocol_handlers[daemon=True] exited
...
DEBUG  2020-05-19 08:16:14,691               Manager  <Manager[FullServer] flags=SRcfe>: task ETHPeerPool[daemon=True] exited with error: Daemon task _feed_protocol_handlers[daemon=True] exited
...
DEBUG  2020-05-19 08:16:14,692               Manager  <Manager[FullNode] flags=SRcfe>: task FullServer[daemon=True] exited with error: Daemon task _feed_protocol_handlers[daemon=True] exited

```

_feed_protocol_handlers() will indeed return if `Multiplexer.stream_protocol_messages` stops producing items: https://github.com/ethereum/trinity/blob/dfe7b4c227888583744f3901958b8855f6dbf472/p2p/connection.py#L174

And that is what happens when the transport is closed: https://github.com/ethereum/trinity/blob/dfe7b4c227888583744f3901958b8855f6dbf472/p2p/multiplexer.py#L272-L280

However, AFAICT, the only place we explicitly close the transport (by calling `Multiplexer.close()` is at the end of `Connection.run()`, and ISTM we can only reach that code path when the connection has been cancelled: https://github.com/ethereum/trinity/blob/dfe7b4c227888583744f3901958b8855f6dbf472/p2p/connection.py#L153-L154

In that case, if `_feed_protocol_handlers()` exited, its DaemonTaskExit exception would be silenced because the service was cancelled, so there must be something else that is closing the transport while the connection is still running

### How can it be fixed

Need to keep digging

gsalgado:
[...]

PR #1757, opened 2020-05-29
Workaround to prevent a DaemonTaskExit error in Connection

gsalgado:
One of the daemons started when we run a Connection service will exit as
soon as the transport is closed and that may happen immediately after we
received a Disconnect+EOF from a remote, but before we've had a chance
to process the disconnect (and cancel the Connection), which would cause
a DaemonTaskExit error (https://github.com/ethereum/trinity/issues/1733)
That is somewhat tricky to fix because the daemon relies on a stream
from the multiplexer, which only has access to the transport, so as a
workaround the daemon (_feed_protocol_handlers) will now wait a bit for
the connection to be cancelled when the msg stream ends.

Closes: #1733

### DIFF
--- a/p2p/connection.py
+++ b/p2p/connection.py
@@ -165,6 +165,8 @@ class Connection(ConnectionAPI, Service):
                     "%s went away while trying to disconnect for MalformedMessage",
                     self,
                 )
+        finally:
+            self.manager.cancel()
 
     #
     # Subscriptions/Handler API
@@ -204,6 +206,19 @@ class Connection(ConnectionAPI, Service):
                     type(cmd),
                 )
                 self.manager.run_task(cmd_handler_fn, self, cmd)
+
+        # XXX: This ugliness is needed because Multiplexer.stream_protocol_messages() stops as
+        # soon as the transport is closed, and that may happen immediately after we received a
+        # Disconnect+EOF from a remote, but before we've had a chance to process the disconnect,
+        # which would cause a DaemonTaskExit error
+        # (https://github.com/ethereum/trinity/issues/1733).
+        if self.is_closing:
+            try:
+                await asyncio.wait_for(self.manager.wait_finished(), timeout=2)
+            except asyncio.TimeoutError:
+                self.logger.error(
+                    "stream_protocol_messages() terminated but Connection was never cancelled, "
+                    "this will cause the Connection to crash with a DaemonTaskExit")
 
     def add_protocol_handler(self,
                              protocol_class: Type[ProtocolAPI],
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -472,7 +472,7 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
         This is passed as a callback to be called when a peer finishes.
         """
         if peer.session in self.connected_nodes:
-            self.logger.debug(
+            self.logger.info(
                 "Removing %s from pool: local_reason=%s remote_reason=%s",
                 peer,
                 peer.p2p_api.local_disconnect_reason,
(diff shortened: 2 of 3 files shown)


p2p/connection.py
```diff
@@ -205,6 +207,25 @@ async def _feed_protocol_handlers(self, protocol: ProtocolAPI) -> None:
                 )
                 self.manager.run_task(cmd_handler_fn, self, cmd)
 
+        # XXX: This ugliness is needed because Multiplexer.stream_protocol_messages() stops as
+        # soon as the transport is closed, and that may happen immediately after we received a
+        # Disconnect+EOF from a remote, but before we've had a chance to process the disconnect,
+        # which would cause a DaemonTaskExit error
+        # (https://github.com/ethereum/trinity/issues/1733).
+        self.logger.debug(
```
pipermerriam, line 215: Maybe instead of this we could just do something simple like:

```python
if self.transport.is_closing():
    self.logger.debug(...)
```

Basically, if the stream exits: error via `DaemonTaskExit` if the transport isn't closing, otherwise, just wait a bit since a shutdown of the connection should be imminent.

gsalgado: That's a good idea, but we need to ensure the `wait_finished()` has a timeout or else we might hang forever there since the transport closing doesn't guarantee the connection will be cancelled

pipermerriam: Sure, we could wrap that in something long like 20-30 seconds and blow up a big loud exception if the timeout is hit.  It's a little ugly, but maybe still cleaner than the current approach.

gsalgado: I don't see why we'd want such a crazy long timeout -- if the remote sent us a Disconnect+EOF the connection should be cancelled as soon as the disconnect msg handler executes. Also, if we want that long, the idle connection monitor will eventually kick in, cancelling the connection and if there is an actual bug causing the transport to be closed prematurely, we'd never notice it. I think the best approach here is to use a short timeout (a couple secs at most) and let the daemon exit, causing a `DaemonTaskExit` if we hit the timeout.

gsalgado:
@pipermerriam would you like to have another look at this? if you have your plate full I could ask somebody else to review as well

p2p/peer_pool.py
```diff
@@ -472,7 +472,7 @@ def _peer_finished(self, peer: BasePeer) -> None:
         This is passed as a callback to be called when a peer finishes.
         """
         if peer.session in self.connected_nodes:
-            self.logger.debug(
+            self.logger.info(
```
pipermerriam, line 475: This seems like it is going to be a noisy logger at `INFO` level.

p2p/transport.py
```diff
@@ -244,11 +244,8 @@ async def recv(self) -> MessageAPI:
         header_bytes = await self.read(HEADER_LEN + MAC_LEN)
         try:
             padded_header = self._decrypt_header(header_bytes)
-        except DecryptionError as err:
-            self.logger.debug(
-                "Bad message header from peer %s: Error: %r",
-                self, err,
-            )
+        except (ValueError, DecryptionError) as err:
+            self.logger.info("Bad message header from peer %s: Error: %r", self, err)
```
pipermerriam, line 248: Are these left at `INFO` level on purpose.  If so, why?

p2p/peer_pool.py
gsalgado: IME we don't seem to get added/removed peers frequently (once every few minutes, at most), and having this shown by default on the console logs will draw our attention to regressions that cause peers to disconnect from us.

p2p/transport.py
gsalgado: This is likely to indicate a bug on our side, so I think it is important that we make it more visible. If this wasn't at DEBUG level maybe we wouldn't have been running with https://github.com/ethereum/trinity/issues/1001#issuecomment-633364583 for such a long time, and that regression significantly reduced the number of peers we could connect to

PR #1758, opened 2020-05-29
Updating Contributing page (docs) to Python3.7

VMLVaske:
### What was wrong?

Docs (Contributing Page) wasn't accurate anymore since dropping Python 3.6 Support. 

### How was it fixed?

Updated the command to install the dependencies with `pip3.7 install -e .[dev]` instead of `pip install -e .[dev]`

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -20,7 +20,7 @@ After we have activated our virtual environment, installing all dependencies tha
 
 .. code:: sh
 
-    pip3.7 install -e .[dev]
+    pip3 install -e .[dev]
 
 
 Running the tests
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -20,7 +20,7 @@ After we have activated our virtual environment, installing all dependencies tha
 
 .. code:: sh
 
-    pip install -e .[dev]
+    pip3.7 install -e .[dev]
 
 
 Running the tests
--- a/docs/fragments/virtualenv_explainer.rst
+++ b/docs/fragments/virtualenv_explainer.rst
@@ -10,7 +10,7 @@ Then, we can initialize a new virtual environment ``venv``, like:
 
 .. code:: sh
 
-  virtualenv -p python3 venv
+  virtualenv -p python3.7 venv
 
 This creates a new directory ``venv`` where packages are installed isolated from any other global
 packages.


docs/contributing.rst
```diff
@@ -20,7 +20,7 @@ After we have activated our virtual environment, installing all dependencies tha
 
 .. code:: sh
 
-    pip install -e .[dev]
+    pip3.7 install -e .[dev]
```
cburgdorf, line 23: I'm on the latest ubuntu and I do not have a `pip3.7` alias. Where did you get that from? Even if some systems might create that alias, I don't think we should put it here as we usually target multiple Python versions (e.g. we currently do also support Python 3.8) and hence the un-versioned `pip` command seems better suited.

Lilith1410: I tried following the guide yesterday after I messed up the trinity version running on my local machine. 

I tried installing the dependencies with `pip install -e .[dev]` and it didn't work. 
Then I tried installing the dependencies with `pip3 install -e .[dev]` and it didn't work, because I ran on Python 3.6.9. 

Then I went to upgrade my Python3 to Python 3.7.5 and only then was I able to install the dependencies.

Lilith1410: Oh, and am on Ubuntu 18.04 LTS

cburgdorf: >Then I went to upgrade my Python3 to Python 3.7.5 and only then was I able to install the dependencies.

Correct, you'll need Python 3.7 on your machine to install Trinity but I assume your `pip` command will now be linked to Python 3.7.

E.g. 

```
$ pip --version
pip 18.1 from /usr/lib/python3/dist-packages/pip (python 3.7)
```

In contrast, `pip3.7` simply is not an available command on my system.

```
$ pip3.7 --version
pip3.7: command not found
```

So, I'm wondering do you really have a `pip3.7` command on your machine?

Lilith1410: I really do have a pip3.7 command on my machine (after installing it with `sudo apt install python3.7` ). 
`$ pip --version
pip 9.0.1 from /usr/lib/python2.7/dist-packages (python 2.7)`
`$ pip3 --version
pip 19.2.3 from /usr/local/lib/python3.7/site-packages/pip (python 3.7)`
`$ pip3.7 --version
pip 19.2.3 from /usr/local/lib/python3.7/site-packages/pip (python 3.7)`

This seems like it is only a problem when there are several versions of python on a local machine. Maybe it would be better to mention this in a warning / alter on the page instead of altering the `pip install -e .[dev]` command?

cburgdorf: I guess the `pip` command is linked to the default version that the system uses. So, if you are on a system where the default Python version is `2.7` and then additionally install another Python version such as `3.7` you'll probably have to use `pip3.7` if you want to refer to the `pip` that comes with Python `3.7`.

But notice that all this should not matter once you are in your `virtualenv`. E.g. I can create a `virtualenv` that is linked to Python `3.8` and then `pip` will be linked to Python 3.8.

E.g.

```sh
$ virtualenv v38 -p python3.8
Running virtualenv with interpreter /usr/bin/python3.8
Already using interpreter /usr/bin/python3.8
Using base prefix '/usr'
New python executable in /home/cburgdorf/Documents/hacking/ef/py-evm/v38/bin/python3.8
Also creating executable in /home/cburgdorf/Documents/hacking/ef/py-evm/v38/bin/python
Installing setuptools, pip, wheel...
done.
cburgdorf at snakebot in ~/Documents/hacking/ef/py-evm on christoph/tests/fixtures*
$ . v38/bin/activate
(v38) cburgdorf at snakebot in ~/Documents/hacking/ef/py-evm on christoph/tests/fixtures*
$ pip --version
pip 20.1.1 from /home/cburgdorf/Documents/hacking/ef/py-evm/v38/lib/python3.8/site-packages/pip (python 3.8)
```

Same for Python `3.7` with `pip` linked to `3.7`

```sh
cburgdorf at snakebot in ~/Documents/hacking/ef/py-evm on christoph/tests/fixtures*
$ virtualenv v38 -p python3.7
Running virtualenv with interpreter /usr/bin/python3.7
Already using interpreter /usr/bin/python3.7
Using base prefix '/usr'
New python executable in /home/cburgdorf/Documents/hacking/ef/py-evm/v38/bin/python3.7
Not overwriting existing python script /home/cburgdorf/Documents/hacking/ef/py-evm/v38/bin/python (you must use /home/cburgdorf/Documents/hacking/ef/py-evm/v38/bin/python3.7)
Installing setuptools, pip, wheel...
done.
cburgdorf at snakebot in ~/Documents/hacking/ef/py-evm on christoph/tests/fixtures*
$ pip --version
pip 18.1 from /usr/lib/python3/dist-packages/pip (python 3.7)
cburgdorf at snakebot in ~/Documents/hacking/ef/py-evm on christoph/tests/fixtures*
$ 
```

So, what I'm saying is that at the point where you are running `pip install -e .[dev]` your `pip` command should be linked to whatever your `virtualenv` uses.

I guess a change that I could get behind is to change `virtualenv -p python3 venv` in the guide to `virtualenv -p python3.7 venv` to make sure people are not accidentially ending up with a `virtualenv` that may be pointing to Python `3.6`.

Frankly, I'm the first to admit that the Python development environment is very messy compared to other programming stacks. We have to find the right balance to make sure we don't turn the guides into a Python tutorial while at the same time ensure newcomers don't end up frustrated.

For the contributing section, I can also get behind a more prominent warning that notes that we currently only support Python 3.7 and Python 3.8. For the non-developer folks, we really need to step up our game and revive #1205 to get a clean installer for everyone who isn't familiar with Python.

Lilith1410: Ah, ok I understand. Thanks for clarifying the virtualenv-part, that makes so much sense! 

I think the non-developer folks would just follow the quickstart guide and be off with a smoothly running alpha-version. However, a lean installer sounds amazing still! 

As a non-python developer I am still having a hard time setting stuff up x)

PR #1761, opened 2020-05-29
Upgrade py-evm to 0.3.0a16

carver:
### What was wrong?

Get the latest evm changes.

### How was it fixed?

~At least one issue: seems to not persist transactions during beam sync anymore :/. Maybe need to use the new API?~ Nope, this was a local bug with a bad install of py-trie. Looks like an uneventful upgrade.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


cburgdorf:
I merged this (with merge commit! :sweat_smile: ) to be able to rebase on top and gain access to the new APIs.

PR #1763, opened 2020-06-01
Use asyncio.Server as an async ctx manager in our TCP listener

gsalgado:
This was added on py37, and ensures we don't silence crashes in our TCP listener server

Another occurrence of https://github.com/ethereum/snake-charmers-tactical-manual/pull/70

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,6 @@ deps = {
         "websockets>=8.1.0",
         "jsonschema==3.0.1",
         "mypy-extensions>=0.4.3,<0.5.0",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -48,7 +48,6 @@
         "websockets>=8.1.0",
         "jsonschema==3.0.1",
         "mypy-extensions>=0.4.3,<0.5.0",
-        "typing_extensions>=3.7.4,<4.0.0",
```
gsalgado, line 51: We don't need this anymore as we no longer support py36

PR #1765, opened 2020-06-01
Reliably detect trinity process with metrics enabled

njgheorghita:
### What was wrong?
re: https://github.com/ethereum/py-evm/issues/1944

With metrics enabled, the strategy to measure total trinity processes/threads would raise an error on ubuntu. 

There was also a bug for some of the default arguments. For `metrics-influx-port` and `metrics-influx-protocol`, it would use `None` (returned from the `os.environ.get()` call). So I removed those default environment calls, and set the default value directly in the parser - rather than in the `BaseMetricsService` class. 

### How was it fixed?
Changed the strategy to be ubuntu-compatible

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/trinity/components/builtin/metrics/component.py
+++ b/trinity/components/builtin/metrics/component.py
@@ -96,7 +96,7 @@ class MetricsComponent(TrioIsolatedComponent):
         metrics_parser.add_argument(
             '--metrics-influx-port',
             help='Influx DB port. Defaults to ENV var TRINITY_METRICS_INFLUX_DB_PORT or 8086',
-            default=os.environ.get('TRINITY_METRICS_INFLUX_DB_PORT'),
+            default=os.environ.get('TRINITY_METRICS_INFLUX_DB_PORT', 8086),
         )
 
         metrics_parser.add_argument(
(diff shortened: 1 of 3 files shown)


trinity/components/builtin/metrics/component.py
```diff
@@ -95,17 +95,16 @@ def configure_parser(cls,
 
         metrics_parser.add_argument(
             '--metrics-influx-port',
-            help='Influx DB port. Defaults to ENV var TRINITY_METRICS_INFLUX_DB_PORT or 8086',
-            default=os.environ.get('TRINITY_METRICS_INFLUX_DB_PORT'),
```
cburgdorf, line 99: I think you can have the best of both worlds with `os.environ.get('TRINITY_METRICS_INFLUX_DB_PORT', 8086)` So it would still use the env var if present but default to `8086` if that env var doesn't exist.

trinity/components/builtin/metrics/component.py
```diff
@@ -95,17 +95,16 @@ def configure_parser(cls,
...
         )
 
         metrics_parser.add_argument(
             '--metrics-influx-protocol',
             help=(
-                'Influx DB protocol. Defaults to ENV var '
-                'TRINITY_METRICS_INFLUX_DB_PROTOCOL or http'
+                'Influx DB protocol. Defaults to `http`'
             ),
-            default=os.environ.get('TRINITY_METRICS_INFLUX_DB_PROTOCOL'),
```
cburgdorf, line 108: Same here with `os.environ.get('TRINITY_METRICS_INFLUX_DB_PROTOCOL', 'http')`

trinity/components/builtin/metrics/component.py
njgheorghita: Ohh that's beautiful and much better, good to know, thanks!

PR #1766, opened 2020-06-02
LogicAPI.apply() now yields a Future 

gsalgado:
This way callsites can detect misbehaving behaviors/applications that crash or exit before the Connection is cancelled.

### DIFF
--- a/p2p/logic.py
+++ b/p2p/logic.py
@@ -1,9 +1,14 @@
+from __future__ import annotations
 from abc import abstractmethod
+import asyncio
 import contextlib
 from typing import (
     cast,
     AsyncIterator,
     Generic,
+    Iterable,
+    List,
+    Sequence,
     Tuple,
     Type,
 )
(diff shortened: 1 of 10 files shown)


p2p/logic.py
```diff
@@ -74,16 +85,55 @@ class Application(BaseLogic):
...
+        else:
+            await cancel_futures(pending)
+            if len(done) != 1:
+                raise Exception(
+                    "Invariant: asyncio.wait() returned more than one future even "
+                    "though we used return_when=asyncio.FIRST_COMPLETED: %s", done)
+            done_future = list(done)[0]
+            if done_future.exception():
+                raise done_future.exception()
+
```
pipermerriam, line 103: I'm reviewing this at the very end of my day so sorry for not doing a more thorough look over but... Maybe we can do a little better than just yielding an `asyncio.Future` and instead wrap it up in a slightly nicer API.

```python
class FutureWrapper:
    def __await__(self):
         ...  # delegates to awaiting the underlying future

    def wait_first(self, *others):
        ...  # encapsulates the logic of this function to make it easy for **any** call site to do this, rather than only ones that inherit from this class.
```

gsalgado: I think it'd make more sense to have `wait_first()` as a standalone function, then, mainly because it doesn't feel very natural to me to have a `wait_first(others)` method on something that wraps a single future, but also because then we can keep on passing a regular Future instance around. The last commit has that change; how does it look to you?

PR #1768, opened 2020-06-02
Upgrade to py-evm 0.3.0a17

carver:
### What was wrong?

Wanted the newest py-trie APIs supported in the latest py-evm, to help with backfilling cold state.

Related to #854 

### How was it fixed?

Upgrade to latest py-evm, and handle new exception.

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Clean up commit history

[//]: # (For important changes that should go into the release notes please add a newsfragment file as explained here: https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)

[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)
- [ ] Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #1769, opened 2020-06-02
Beam Sync storage & bytecode backfill

### LINKED ISSUE #854: Backfill cold state during Beam sync
This issue is to track some ideas on how to fill in cold state with beam sync.  Mostly a brain dump that hopefully works towards some good ideas.

## Goals

1. Preserve full node caches.
2. Swarming: syncing nodes should be able to serve other syncing nodes with some reasonably good probability.
3. Fill in all cold state within a reasonable amount of time.

In order to have a cache preserving behavior clients need to be pulling *common* state.  i.e. two clients that are beam syncing at the same time should end up pulling roughly the same cold state nodes.  Ideal case is for this property to hold even if one of the clients started their sync at a very different time than the other.

## Questions

- Is it beneficial to prioritize filling in the account state before filling in the contract state?

There might be an argument that account balances and nonces are more valuable information than the actual contract code and state.   Every contract call interacts with the account but doesn't necessarily interact with the state (but it does require the actual code).

### Prioritize cold state reads from alternative sources like Swarm.  

When looking up cold state, we could try to prioritize fetching it from alternative sources such as Swarm.  This could help preserve full node caches while having the beneficial side effect of keeping cold state active in the swarm network.

### Select cold state based on most recent processed block

In order to have multiple syncing clients converge on fetching the same state nodes we could use the most recent block's state reads as the locus for clients to swarm around.  Something like traversing the state trie outwards from the nodes that were read in the most recent block till it finds the closest account that it is missing and fetches it too.  The same procedure could be used for contract state.

This should have the following effects:

- clients asking for data from full nodes will be asking for mostly the same data from the same part of the trie.
- a client that is syncing should always be able to serve a state request to another client provided that the requesting client started syncing *after* them.

carver:
### What was wrong?

Fixes #854 

### How was it fixed?

Combine all storage/bytecode/account missing nodes into a single iterator. Wait to mark account as finished until all storage/bytecode is collected.

### To-Do

- [x] Rebase on #1768 
- [x] py-evm release, and upgrade to published version
- [x] Clean up commit history (_after_ review :+1: )
- [x] Rebase on #1844 
- [x] lint
- [x] Peel off a couple of isolated PRs
- [x] Rebase on merged, isolated PRs
- [x] Refactor duplicate logic for cleaning up paused account search
- [x] Replace all TODOs with issues
- [x] Build special test database with cold data
- [x] Test that accounts are missing
- [x] Backfill accounts
- [x] Test that storage is missing
- [x] Backfill storage
- [x] Test that bytecode is missing (need new fixture!)
- [x] Backfill bytecode
- [x] new issue: Save progress so that exiting and restarting trinity doesn't require rewalking the trie from scratch #1817
- [x] new issue: test that epoch crossover is happening #1815
- [x] new issue: test a state change that triggers a `TraversedPartialPath` to make sure that code path is tested #1816
- [x] ~Probably in new issue:~ find keys based on recent state root

#### Cute Animal Picture

### DIFF
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -6,12 +6,14 @@ import uuid
 
 from async_service import Service, background_asyncio_service
 from eth.consensus import ConsensusContext
+from eth.constants import EMPTY_SHA3
 from eth.db.atomic import AtomicDB
 from eth.db.schema import SchemaV1
 from eth.exceptions import (
     BlockNotFound,
     HeaderNotFound,
 )
+from eth.rlp.accounts import Account
 from eth.vm.forks import (
     MuirGlacierVM,
     PetersburgVM,
@@ -19,8 +21,12 @@ from eth.vm.forks import (
 from eth_utils import decode_hex
 from lahja import ConnectionConfig, AsyncioEndpoint
 import pytest
+import rlp
 from trie import (
     HexaryTrie,
+)
+from trie.constants import (
+    BLANK_NODE_HASH,
 )
 from trie.exceptions import (
     MissingTraversalNode,
@@ -298,7 +304,7 @@ async def _beam_syncing(
             async with background_asyncio_service(import_server):
                 await pausing_endpoint.connect_to_endpoints(gatherer_config)
                 async with background_asyncio_service(client):
-                    yield
+                    yield client
 
 
 # Cases of interest:
@@ -353,6 +359,8 @@ async def test_beam_syncer_loads_recent_state_root(
 
 @pytest.mark.asyncio
 # Cases of interest:
+# -  0: This doesn't test beam sync or backfill, since all state is available immediately.
+#       The test checks things that don't happen if we start at 0, so it's excluded from the range.
 # -  2: Normally we need to look back ~6 headers to look for duplicate uncles. This
 #       tests the lowest block number code path where an alternate code path is triggered,
 #       to look up fewer than usual.
@@ -365,9 +373,10 @@ async def test_beam_syncer_loads_recent_state_root(
 #       duplicate hashes in the same request.
 # - 42: Last block
 # In order to run a deeper test, we can run:
-# @pytest.mark.parametrize('beam_to_block', range(0, 43))
+# @pytest.mark.parametrize('beam_to_block', range(1, 43))
 @pytest.mark.parametrize('beam_to_block', [2, 7, 8, 25, 38, 42])
 async def test_beam_syncer_backfills_all_state(
+        caplog,
         request,
         event_loop,
         event_bus,
@@ -387,8 +396,29 @@ async def test_beam_syncer_backfills_all_state(
         VM_at_0=MuirGlacierVM,
     )
 
-    async with sync_test_service:
+    caplog.set_level(logging.INFO)
+    async with sync_test_service as beam_syncer:
+        # Manually verify that all state is in the state database now
         await wait_for_full_state_db(chaindb_fresh, beam_to_block, timeout=10)
+        beam_syncer.logger.info("State DB complete by manual inspection")
+
+        # Check that backfiller exits on state completion
+        try:
+            await asyncio.wait_for(beam_syncer._backfiller.manager.wait_finished(), timeout=10)
+        except asyncio.TimeoutError:
+            beam_syncer.logger.warning("Backfiller never exited, even though all state is present")
+            if beam_syncer._backfiller._check_complete():
+                beam_syncer.logger.warning("Backfiller thinks it's complete, but never exited")
+            else:
+                beam_syncer.logger.warning(
+                    "Backfiller thinks it's missing %s",
+                    beam_syncer._backfiller._account_tracker,
+                )
+            # Whatever the reason, this TimeoutError means the backfiller service didn't exit
+            raise
+
+    # Double-check that backfiller exited for the right reason.
+    assert "Downloaded all accounts, storage and bytecode state" in caplog.text
 
 
 @pytest.mark.asyncio
@@ -985,7 +1015,6 @@ async def wait_for_block(chain, block, sync_timeout=10):
 
 
 async def wait_for_full_state_db(headerdb, min_block_number, timeout):
-    # Only tracking accounts for now, will add bytecode & storage shortly
     collected_data = collections.defaultdict(int)
 
     def _has_full_state_db(headerdb):
@@ -1001,8 +1030,29 @@ async def wait_for_full_state_db(headerdb, min_block_number, timeout):
         trie = HexaryTrie(raw_db, state_root)
         iterator = NodeIterator(trie)
         try:
-            for _ in iterator.items():
+            for key, value in iterator.items():
                 collected_data["num_accounts"] += 1
+
+                account = rlp.decode(value, sedes=Account)
+
+                if account.code_hash == EMPTY_SHA3:
+                    pass
+                elif account.code_hash not in raw_db:
+                    # missing bytecodes
+                    logging.warning("Missing bytecode for account at 0x%s", key.hex())
+                    return False
+                else:
+                    collected_data["num_bytecodes"] += 1
+
+                if account.storage_root != BLANK_NODE_HASH:
+                    storage_trie = HexaryTrie(raw_db, account.storage_root)
+                    try:
+                        for _key in NodeIterator(storage_trie).keys():
+                            # We don't care what the keys are, just that we can iterate through them
+                            collected_data["num_storage_slots"] += 1
+                    except MissingTraversalNode as exc:
+                        logging.warning("Missing storage for account at 0x%s: %s", key.hex(), exc)
+                        return False
 
         except MissingTraversalNode:
             return False
--- a/trinity/sync/beam/backfill.py
+++ b/trinity/sync/beam/backfill.py
@@ -5,7 +5,7 @@ from functools import partial
 import typing
 from typing import (
     AsyncIterator,
-    AsyncIterable,
+    Dict,
     Iterable,
     NamedTuple,
     Optional,
@@ -19,7 +19,10 @@ from eth.abc import (
     AtomicDatabaseAPI,
     BlockHeaderAPI,
 )
+from eth.constants import EMPTY_SHA3
+from eth.rlp.accounts import Account
 from eth_typing import Hash32
+import rlp
 from trie import (
     HexaryTrie,
     exceptions as trie_exceptions,
@@ -85,6 +88,9 @@ class BeamStateBackfill(Service, QueenTrackerAPI):
         # Track the nodes that we are requesting in the account trie
         self._account_tracker = TrieNodeRequestTracker()
 
+        self._storage_trackers: Dict[Hash32, TrieNodeRequestTracker] = {}
+        self._bytecode_trackers: Dict[Hash32, TrieNodeRequestTracker] = {}
+
         # The most recent root hash to use to navigate the trie
         self._next_trie_root_hash: Optional[Hash32] = None
         self._begin_backfill = asyncio.Event()
@@ -101,6 +107,7 @@ class BeamStateBackfill(Service, QueenTrackerAPI):
         queening_manager = self.manager.run_daemon_child_service(self._queening_queue)
         await queening_manager.wait_started()
         await self._run_backfill()
+        self.manager.cancel()
 
     async def _run_backfill(self) -> None:
         await self._begin_backfill.wait()
@@ -117,8 +124,8 @@ class BeamStateBackfill(Service, QueenTrackerAPI):
 
             if len(required_data) == 0:
                 # Nothing available to request, for one of two reasons:
-                if self._is_complete():
-                    self.logger.info("Downloaded all account state")
+                if self._check_complete():
+                    self.logger.info("Downloaded all accounts, storage and bytecode state")
                     return
                 else:
                     # There are active requests to peers, and we don't have enough information to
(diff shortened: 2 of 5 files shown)


tests/core/integration_test_helpers.py
```diff
@@ -30,6 +30,28 @@
 FUNDED_ACCT = keys.PrivateKey(
     decode_hex("49a7b37aa6f6645917e7b807e9d1c00d4fa71f18343b0d4122a4d2df64dd6fee"))
 
+# How to build your own database:
```
cburgdorf, line 33: :heart:

tests/core/integration_test_helpers.py
```diff
@@ -30,6 +30,28 @@
...
+#       importlib.reload(ifb)
+#       db = AtomicDB(); ifb.build_YOUR_NEW_METHOD(db, num_blocks=SOME_SMALL_TEST_NUMBER)
+#       # inevitable debugging, etc...
+#       # It works!
+#       from eth.db.backends.level import LevelDB
+#       ldb = LevelDB('/tmp/NEW_DB_NAME.ldb')
+#       ifb.build_YOUR_NEW_METHOD(ldb, num_blocks=FULL_TEST_NUMBER)
+#   - Then zip it up:
+#       cd /tmp; zip -r NEW_DB_NAME.ldb.zip NEW_DB_NAME.ldb
+#       cd -
```
cburgdorf, line 51: TIL `cd -` lets you jump back to the previous point. I wonder where else this trick can be applied. I know that `git checkout -` works the same way but it never occurred to me that this may be a broader general concept of "jumping back"

carver: Yeah, I use both of those `-` tricks a lot! It has a more convenient placing on the Dvorak layout, which is nice.

Unfortunately, it doesn't seem have a consistent meaning. I don't even think it's the most common meaning.

I *think* it's most common meaning is to read from stdin instead of a file. Like:
```sh
cat header - footer
```
will output the header file to stdout, then output stdin to stdout, then output the footer file to stdout.

tests/core/p2p-proto/test_sync.py
```diff
@@ -182,7 +187,12 @@ async def test_beam_syncer(
         bob_peer_context=server_context,
         event_bus=event_bus,
     )
-    async with peer_pair as (client_peer, server_peer):
+    backfiller = LatestETHPeerPairFactory(
+        alice_peer_context=client_context,
+        bob_peer_context=server_context,
+        event_bus=event_bus,
+    )
+    async with peer_pair as (client_peer, server_peer), backfiller as (client2_peer, backfill_peer):
```
carver, line 195: Yeah, so this might not help you @cburgdorf with the things you mentioned at standup -- since you really wanted to manage the particular order that peers connect. But I thought it was interesting that it works at all!

trinity/sync/beam/backfill.py
```diff
@@ -78,28 +101,110 @@ async def run(self) -> None:
...
+
+            if len(required_data) == 0:
+                # Nothing left to request
+                if len(self._active_prefixes) == 0:
+                    # TODO save record that full state is available as of X block
+                    self.logger.info("Successfully backfilled all accounts!")
+                    return
+                else:
+                    # Waiting for peers to get back to us, continue loop & wait for new data
+                    self._queening_queue.readd_peasant(peer)
```
pipermerriam, line 124: Is this a typo?  (`readd` instead of `read`)?

carver: lol, nope. I think it could just be `add_peasant`. In my head, I was distinguishing it from the primary peer registration (like the peer is available) vs just putting it back on the queue after using it. Maybe not an important distinction.

pipermerriam: I think it was obvious but I was seeing this as `read` (like as in read a :book: ) with an extra `d`.

- `re_add_peer`
- `add_peer_again`
- `add_peer_back`

dunno, but the namespace collision with `read` is probably good to avoid for future readers.

cburgdorf: TIL: peasant.

Queen, Peasant, I get it! We are building a blockchain client for a decentralized chess game! Wohoo! That's so exciting!

Other candidates that stand more appart from`read`:

- `reinsert_peasant`
- `reattach_peasant`

trinity/sync/beam/backfill.py
```diff
@@ -228,3 +440,62 @@ async def _periodically_report_progress(self) -> None:
                 self._num_requests_by_peer.most_common(show_top_n_peers),
             )
             self._num_requests_by_peer.clear()
+
+
+@dataclass
```
pipermerriam, line 445: I don't think this article should be considered definitive but:

https://medium.com/@jacktator/dataclass-vs-namedtuple-vs-object-for-performance-optimization-in-python-691e234253b9

I was curious whether `NamedTuple` or `dataclass` had better performance characteristics.  It looks like `dataclass` wins out over `NamedTuple` which surprised me.  Just a neat tidbit.

carver: Ah, interesting. Though now that I think about it, it's probably worth the small performance penalty of `NamedTuple` to be immutable.

cburgdorf: Yeah, surprising result. I agree that immutability in itself is such a big win though, that it still may be the better choice here.

carver:
In the home stretch, but this PR is giant, so I'll start peeling pieces off for review, where possible.

tests/core/p2p-proto/test_sync.py
```diff
@@ -255,17 +279,104 @@ async def test_beam_syncer(
...
+        # significantly
+        target_block_number = min(beam_to_block + 10, 129)
+        target_head = chaindb_churner.get_canonical_block_header_by_number(
+            target_block_number,
+        )
+        await wait_for_head(chaindb_fresh, target_head, sync_timeout=5)
+        assert target_head.state_root in chaindb_fresh.db
+
+
+# Interesting cases
```
carver, line 329: Delete comment in favor of issue

tests/core/p2p-proto/test_sync.py
```diff
@@ -255,17 +279,104 @@ async def test_beam_syncer(
...
+        event_bus,
+        chaindb_fresh,
+        chaindb_cold_state,
+        beam_to_block,
+        checkpoint,
+        VM_at_0=MuirGlacierVM,
+    )
+
+    async with sync_test_service:
+        # TODO validate storage slots too!!!
```
carver, line 377: remove out of date comment

trinity/sync/beam/backfill.py
```diff
@@ -76,45 +106,391 @@ async def run(self) -> None:
...
+                request async for request in self._atake(REQUEST_SIZE, self._get_next_node())
+            ])
+
+            if len(required_data) == 0:
+                # Nothing left to request
+                if self._has_any_active_requests():
+                    # Waiting for peers to get back to us, continue loop & wait for new data
+                    self._queening_queue.readd_peasant(peer)
+                    self.logger.debug("Backfill is waiting for more hashes to arrive")
+                    await asyncio.sleep(2)
```
carver, line 145: Reminder from @cburgdorf to move to constant

trinity/sync/beam/backfill.py
```diff
@@ -76,45 +106,391 @@ async def run(self) -> None:
...
+                    return
+
+            self.manager.run_task(self._make_request, peer, required_data)
 
-            self._node_hashes, on_deck = (
-                self._node_hashes[:-1 * REQUEST_SIZE],
-                tuple(self._node_hashes[-1 * REQUEST_SIZE:]),
+    def _has_any_active_requests(self) -> bool:
+        if self._account_tracker.has_active_requests:
+            return True
```
carver, line 156: Rename `is_backfill_complete`?

trinity/sync/beam/backfill.py
```diff
@@ -76,45 +106,391 @@ async def run(self) -> None:
...
+        else:
+            # TODO optimize this check, it could get very long on a long run.
+            #   Maybe delete the storage tracker when it has been fully iterated
+            #   and no active requests remain.
+            return any(
+                storage_tracker.has_active_requests
+                for storage_tracker in self._storage_trackers.values()
```
carver, line 163: Also, check bytecode trackers!

carver:
Okay, chiseled this down to the last piece: bytecode and storage backfill. Ready for review!

trinity/sync/beam/backfill.py
```diff
@@ -153,42 +172,92 @@ async def _missing_trie_hashes(self) -> AsyncIterator[TrackedRequest]:
...
-                #   exit and let the caller deal with it.
-                return
+                if account_components_complete:
+                    # Mark fully downloaded accounts as complete, and do some cleanup
+                    self._mark_account_complete(path_to_leaf, address_hash_nibbles)
+                else:
+                    # Pause accounts that are not fully downloaded, and track the account
+                    #   to resume when the generator exits.
+                    self._account_tracker.pause_review(path_to_leaf)
+                    exhausted_account_leaves.append(path_to_leaf)
```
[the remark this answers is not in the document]
carver: I think there might be some compiling magic happening there:
```
$ ipython
Python 3.7.5 (default, Nov  7 2019, 10:50:52) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.9.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: a = (1, 2)                                                                                                                                                                                                                            

In [2]: b = (3,)                                                                                                                                                                                                                              

In [3]: %timeit a + b                                                                                                                                                                                                                         
42.4 ns ± 0.371 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: c = [1, 2]                                                                                                                                                                                                                            

In [5]: d = 3                                                                                                                                                                                                                                 

In [6]: %timeit c.append(d)                                                                                                                                                                                                                   
49.7 ns ± 1.72 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

So it's not a *big* performance win, but I'm still :+1: on the change to generally avoid the mutable variables.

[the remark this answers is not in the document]
pipermerriam: Performance will likely degrade as the size grows larger because it will have to make new `tuple` objects each time and as those get larger I suspect the `list` based approach would catch up since it should be closer to `O(1)` where-as the `tuple` approach is likely closer to `O(N)`

tests/core/p2p-proto/test_sync.py
```diff
@@ -979,7 +985,6 @@ async def wait_loop():
 
 
 async def wait_for_full_state_db(headerdb, min_block_number, timeout):
-    # Only tracking accounts for now, will add bytecode & storage shortly
     collected_data = collections.defaultdict(int)
 
     def _has_full_state_db(headerdb):
```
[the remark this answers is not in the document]
carver: Added an issue https://github.com/ethereum/trinity/issues/1863

tests/core/p2p-proto/test_sync.py
```diff
@@ -995,9 +1000,30 @@ def _has_full_state_db(headerdb):
...
+
+                if account.storage_root != BLANK_NODE_HASH:
+                    storage_trie = HexaryTrie(raw_db, account.storage_root)
+                    try:
+                        for _key in NodeIterator(storage_trie).keys():
+                            # We don't care what the keys are, just that we can iterate through them
+                            collected_data["num_storage_slots"] += 1
+                    except MissingTraversalNode as exc:
+                        logging.warning("Missing storage for account at 0x%s: %s", key.hex(), exc)
+                        return False
```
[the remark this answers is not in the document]
carver: I like it! We'll have to push py-trie v2 to stable first, I think.

trinity/sync/beam/backfill.py
```diff
@@ -118,7 +124,7 @@ async def _run_backfill(self) -> None:
             if len(required_data) == 0:
                 # Nothing available to request, for one of two reasons:
                 if self._is_complete():
-                    self.logger.info("Downloaded all account state")
+                    self.logger.info("Downloaded all accounts, storage and bytecode state")
```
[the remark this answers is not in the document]
carver: Ooh, good catch, thanks. Yeah, it's problematic that the test didn't catch this scenario. There should definitely be coverage making sure that log prints out, and to see what happens when the backfill service exits (I agree that a crash looks probable, but I still don't feel like I have a grasp on async-service, to know for sure).

Off to tinker...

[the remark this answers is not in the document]
carver: Update: adding the new test caught a bug (the backfiller got stuck with not recognizing a part of the trie as filled, even though it was available).

trinity/sync/beam/backfill.py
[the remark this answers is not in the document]
cburgdorf: Uh, looks like you are right about list outperforming tuple the more items we have to handle:

```
In [15]: a_tuple = tuple(i for i in range(1_000_000))                                                                                                                                                              

In [16]: b_tuple = (1,)                                                                                                                                                                                            

In [17]: %timeit a_tuple + b_tuple                                                                                                                                                                                 
7.96 ms ± 38.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [18]: a_list = list(i for i in range(1_000_000))                                                                                                                                                                

In [19]: b_item = 1                                                                                                                                                                                                

In [20]: %timeit a_list.append(b_item)                                                                                                                                                                             
40.1 ns ± 0.333 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

```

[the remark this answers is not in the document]
carver: Yeah, it's highly unlikely that we would get to 1M, but this would *not* scale well with the number of peers. I'm thinking the worst case scenario (very much not an average scenario) with 50 peers would be:
```
In [6]: max_skipped_leaves = 384 * 50                                                                                

In [7]: at = tuple(i for i in range(max_skipped_leaves))                                                             

In [8]: bt = (1,)                                                                                                    

In [9]: %timeit at + bt                                                                                              
49.7 µs ± 757 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [10]: al = [i for i in range(max_skipped_leaves)]                                                                 

In [11]: bitem = 1                                                                                                   

In [12]: %timeit al.append(bitem)                                                                                    
46.2 ns ± 0.702 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

So yeah, 1000x worse perf. Multiply that across the 50 peers, and the microseconds becomes 2.5ms for every full cycle of asking all peers for data once. So... right on the knife edge. It probably wouldn't make a difference in practice. But it's an easy change back to a list, and it would help make sure that we could support 100s of peers without falling over. Probably worth it.

Issue #1770, opened 2020-06-03
LifecycleError on RegularChainBodySyncer._assign_body_download_to_peers


gsalgado:
That method calls `peer.manager.run_task()` without first checking that the peer is still running, leading to the following LifecycleError:

```
   DEBUG  2020-06-03 11:26:02,837               Manager  <Manager[RegularChainBodySyncer] flags=SRcfe>: task _assign_body_download_to_peers[daemon=True] exited with error: Tasks may not be scheduled if the service is not running
Traceback (most recent call last):
  File "/home/salgado/virtualenvs/trinity37/lib/python3.7/site-packages/async_service/base.py", line 300, in _run_and_manage_task
    await task.run()
  File "/home/salgado/virtualenvs/trinity37/lib/python3.7/site-packages/async_service/asyncio.py", line 35, in run
    await self._async_fn(*self._async_fn_args)
  File "/home/salgado/src/trinity/trinity/sync/full/chain.py", line 277, in _assign_body_download_to_peers
    peer.manager.run_task(self._run_body_download_batch, peer, batch_id, headers)
  File "/home/salgado/virtualenvs/trinity37/lib/python3.7/site-packages/async_service/asyncio.py", line 328, in run_task
    self._common_run_task(task)
  File "/home/salgado/virtualenvs/trinity37/lib/python3.7/site-packages/async_service/base.py", line 275, in _common_run_task
    "Tasks may not be scheduled if the service is not running"
async_service.exceptions.LifecycleError: Tasks may not be scheduled if the service is not running
```

This is fallout from the conversion to new-style services, as before it was possible to schedule tasks on services that were already cancelled

gsalgado:
`FastChainBodySyncer._assign_receipt_download_to_peers` and `HeaderMeatSyncer._match_dl_to_peer` as well

PR #1771, opened 2020-06-04
Ensure peer is still running before we call run_task() on them

gsalgado:
RegularChainBodySyncer._assign_body_download_to_peers and
FastChainBodySyncer._assign_receipt_download_to_peers would
make async calls after looking up a peer and before calling run_task()
on them, causing a LifecycleError sometimes. They now perform the peer
lookup immediately before calling run_task(), ensuring the peer is still
running when we do so.

Closes: #1770

### DIFF
--- a/trinity/sync/common/headers.py
+++ b/trinity/sync/common/headers.py
@@ -641,16 +641,25 @@ class HeaderMeatSyncer(Service, PeerSubscriber, Generic[TChainPeer]):
         def fail_task() -> None:
             self._filler_header_tasks.complete(batch_id, tuple())
 
-        peer = await self._waiting_peers.get_fastest()
-        if not self.sync_progress:
-            await self._init_sync_progress(parent_header, peer)
-
         def complete_task() -> None:
             self._filler_header_tasks.complete(batch_id, (
                 (parent_header, gap, skeleton_peer),
             ))
-        peer.manager.run_task(
-            self._run_fetch_segment, peer, parent_header, gap, complete_task, fail_task)
+
+        download_scheduled = False
+        while not download_scheduled:
+            peer = await self._waiting_peers.get_fastest()
+            if not self.sync_progress:
+                await self._init_sync_progress(parent_header, peer)
+
+            if not peer.manager.is_running:
+                self.logger.debug(
+                    "Tried to fetch headers from %s but it is no longer running", peer)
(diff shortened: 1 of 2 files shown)


trinity/sync/common/headers.py
```diff
@@ -641,16 +641,25 @@ async def _match_dl_to_peer(
...
+            if not self.sync_progress:
+                await self._init_sync_progress(parent_header, peer)
+
+            if not peer.manager.is_running:
+                self.logger.debug(
+                    "Tried to fetch headers from %s but it is no longer running", peer)
+                continue
+
+            peer.manager.run_task(
+                self._run_fetch_segment, peer, parent_header, gap, complete_task, fail_task)
```
carver, line 661: Hm, just noticed a potential issue that's been present since we switched from `BaseService` to `async_service`. When a service exits, does `async-service` now hard-kill any running child tasks, instead of waiting for them to complete?

`fail_task` really needs to run (to reschedule the task for another peer to pick up). So if a peer closes and doesn't finish running `_run_fetch_segment`, then the task would not get re-queued. This whole mechanism was built around the old `BaseService` model of giving the child task a chance to complete when the service shuts down.

This will need a more serious rework, if that's not the case anymore.

gsalgado: > Hm, just noticed a potential issue that's been present since we switched from BaseService to async_service. When a service exits, does async-service now hard-kill any running child tasks, instead of waiting for them to complete?

It will call `task.cancel(); await task` upon termination, but `_run_fetch_segment()` runs `fail_task_fn` in case it catches an `Exception`, so that will do the trick up to py37. On py38 it won't because `asyncio.CancelledError` was changed to inherit from `BaseException`. Anyway, I doubt we actually want to silence an `asyncio.CancelledError` there (which should be expected to happen), so I believe the following is the correct fix:

```
        try:
            completed_headers = await self._fetch_segment(peer, parent_header, length)
        except asyncio.CancelledError:
            fail_task_fn()
            raise
        ...
```

gsalgado: https://github.com/ethereum/trinity/pull/1771/commits/4f8ccc90415129b443326621c2d679b2b855386d

carver:
:+1: :shipit:

PR #1776, opened 2020-06-08
Ensure consistent handling of CancelledErrors between py37/py38

gsalgado:
As of python3.8, asyncio.CancelledError inherits from BaseException
instead of Exception, and we had some callsites that expected to catch
that with a `except Exception`, which wouldn't work on 3.8

Also removes a few left over `except OperationCancelled` blocks, which
are no longer reachable as our old-style services are gone.


trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -197,6 +197,10 @@ async def handle_stream(self,
             try:
                 await event_handler_fn(event)
             except Exception as exc:
+                if isinstance(exc, asyncio.CancelledError):
```
cburgdorf, line 200: *nitpick*: Would this be slightly nicer using multiple except clauses?

```python
           try:
                await event_handler_fn(event)
            except asyncio.CancelledError as exc:
               raise exc
            except Exception as exc:
                self.logger.exception("Suppressed uncaught exception to continue handling events: %s",
                    exc,
                )
```

The downside is that the order of the  `except` blocks becomes important but it may still be more idiomatic?

gsalgado: Yeah, I think it's slightly better; I'll just add a comment as well, so that the `except CancelledError` block is not accidentally removed in the future

PR #1777, opened 2020-06-08
Regenerate 1000pow_headers fixture under latest VM rules

cburgdorf:
### What was wrong?

The current `1000pow_headers` fixture was created under the VM rules of Byzantium (the latest rules at the time of creation). I want to build an uncle chain to this fixture but instead of sticking to byzantium I think it is better to simply update the old fixture.

### How was it fixed?

Updated `1000pow_headers` fixture to use the latest chain rules.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


carver:
Hm, I'm not keen to drop/add ~1MB from the repo every time there's a network change. It will make checking out trinity painfully slow after a while. Just configuring the test to use `Byzantium` sounds better than that to me.

I'm up for considering other options, though.

cburgdorf:
>Hm, I'm not keen to drop/add ~1MB from the repo every time there's a network change

Mmh..ok, I didn't consider that. 

I had actually started by adding an additional parameter to this API to be able to over rule the `latest_mainnet_at(0)`.

https://github.com/ethereum/trinity/blob/fc57875d779d9528df03f629e352c60cc6ecd61b/tests/core/integration_test_helpers.py#L34-L53

But I thought it would be cleaner to just update the fixtures to the latest VM rules instead.

>I'm up for considering other options, though.

Anything else that would come to your mind? Like hosting the zips elsewhere and curling them in?

cburgdorf:
Another idea could be to use a submodule to load in fixtures (similar to `ethereum/tests`). I think that **not** upgrading the fixtures to newer VM rules might get messy over time. E.g. we might need fixtures with newer opcodes to test specific things etc.

carver:
> I had actually started by adding an additional parameter to this API to be able to over rule the `latest_mainnet_at(0)`.

I think this is an okay/good medium term solution.

> I think that **not** upgrading the fixtures to newer VM rules might get messy over time. E.g. we might need fixtures with newer opcodes to test specific things etc.

I don't really expect it to be a *super* common need to test things where POW must be valid, and we need a bunch of headers in a row. But if the cases start piling up, we can think about other options.



> > I'm up for considering other options, though.
> 
> Anything else that would come to your mind? Like hosting the zips elsewhere and curling them in?

I guess? It sounds pretty rough to me. Maybe a submodule? Or AIUI there's a way to trim old objects out of the git history, but obviously we would want to be _very_ careful about doing something like that. That's why I'm still in favor of just passing in the VM to use, for now.

cburgdorf:
Regenerated the uncle fixture on byzantium. Closing

PR #1779, opened 2020-06-11
Add docs on how to start Trinity from a trusted checkpoint

cburgdorf:
### What was wrong?

It wasn't entirely clear how to start Trinity from a checkpoint (question raised in our chat).

### How was it fixed?

Added a cookbook entry.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/cookbook.rst
+++ b/docs/cookbook.rst
@@ -5,6 +5,51 @@ The Cookbook is a collection of simple recipes that demonstrate good practices t
 common tasks. The examples are usually short answers to simple "How do I..." questions that go
 beyond simple API descriptions but also don't need a full guide to become clear.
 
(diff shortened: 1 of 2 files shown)


kikoncuo:
I think the flag is ` --beam-from-checkpoint ` not ` --sync-from-checkpoint ` at least that´s the case for the latest release

cburgdorf:
@kikoncuo it was renamed to `--sync-from-checkpoint` lately because one can also use this flag for `--sync-mode header`. We will have a new release probably between Monday and Wednesday next week.

cburgdorf:
>but I think adding a link to somewhere where this uri scheme is more broadly defined could be useful,

Frankly, we made up that URI scheme out of thin air so the two examples are the only currently supported cases :sweat_smile:

PR #1780, opened 2020-06-12
Mention Görli support in docs

cburgdorf:
### What was wrong?

The docs don't mention support for Görli.

### How was it fixed?

- Expanded cookbook entry
- Mention Görli support on home page

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -26,7 +26,7 @@ The main focus is to enrich the Ethereum ecosystem with a Python implementation 
 * Runs fast and resource friendly
 * Is highly flexible to support:
 
-  * Public chains (including Mainnet, Ropsten and other networks)
+  * Public chains (including Mainnet, Görli, Ropsten and other networks)
   * Private chains
   * Consortium chains
   * Advanced research
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -1,12 +1,11 @@
 Introduction
 ============
 
-Trinity is a program that connects to the Ethereum network to operate as a full or light node.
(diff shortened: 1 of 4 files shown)


docs/introduction.rst
```diff
@@ -1,12 +1,11 @@
...
-Trinity is a program that connects to the Ethereum network to operate as a full or light node.
-It is built on top of Py-EVM which is a new implementation of the Ethereum Virtual Machine (EVM)
-written in Python.
+Trinity is an implementation of an Ethereum network node. It is built on top of Py-EVM which is
+an implementation of the Ethereum Virtual Machine (EVM) written in Python.
 
-Trinity and Py-EVM aim to replace existing Python Ethereum implementations to eventually become the
-defacto standard for the Python ecosystem.
+Trinity and Py-EVM are the successors to the deprecated py-ethereum project and are the current
+the go-to implementations for the Python ecosystem.
```
njgheorghita, line 8: I'd change this to `and are the current go-to implementations`

cburgdorf:
>I've always been confused with the difference between `görli` and `goerli`. It might be confusing to users that the flag is `goerli` and we reference it as `görli` in the docs. But... that's also the grammatically correct way to spell it (i think), so maybe it's the best we can do

See this is just how the non-anglophonic world tries to pick on the anglophonic world because English won the language wars :joy: 

No, but really, when we refer to the name of something or someone like [Chloë Grace Moretz](https://en.wikipedia.org/wiki/Chlo%C3%AB_Grace_Moretz) we should pay attention to the correct writing. For all technical aspects (the CLI flag, code variables etc) we will use the normalized form without accents.

PR #1783, opened 2020-06-12
Update list of bootnodes

cburgdorf:
### What was wrong?

Nothing wrong but I thought it would be a good idea to update the list of bootnodes across the board with the bootnodes from latest geth/master before we cut a release.

### How was it fixed?

Updated list of bootnodes with the latest from geth/master

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


pipermerriam (approved):

p2p/constants.py
```diff
@@ -94,13 +94,16 @@
...
+    # Ethereum Foundation bootnode
+    "enode://a61215641fb8714a373c80edbfa0ea8878243193f57c96eeb44d0bc019ef295abd4e044fd619bfc4c59731a73fb79afe84e9ab6da0c743ceb479cbb6d263fa91@3.11.147.67:30303",  # noqa: E501
 )
 ROPSTEN_BOOTNODES = (
-    'enode://15ac307a470b411745a6f10544ed54c0a14ad640b21f04f523e736e732bf709d4e28c2f06526ecabc03eed226b6d9bee8e433883cd20ab6cbd114bab77a8775d@52.176.7.10:30303',     # noqa: E501
-    'enode://865a63255b3bb68023b6bffd5095118fcc13e79dcf014fe4e47e065c350c7cc72af2e53eff895f11ba1bbb6a2b33271c1116ee870f266618eadfc2e78aa7349c@52.176.100.77:30303',   # noqa: E501
-    'enode://6332792c4a00e3e4ee0926ed89e0d27ef985424d97b6a45bf0f23e51f0dcb5e66b875777506458aea7af6f9e4ffb69f43f3778ee73c81ed9d34c51c4b16b0b0f@52.232.243.152:30303',  # noqa: E501
-    'enode://94c15d1b9e2fe7ce56e458b9a3b672ef11894ddedd0c6f247e0f1d3487f52b66208fb4aeb8179fce6e3a749ea93ed147c37976d67af557508d199d9594c35f09@192.81.208.223:30303',  # noqa: E501
+    "enode://30b7ab30a01c124a6cceca36863ece12c4f5fa68e3ba9b0b51407ccc002eeed3b3102d20a88f1c1d3c3154e2449317b8ef95090e77b312d5cc39354f86d5d606@52.176.7.10:30303",  # noqa: E501
```
gsalgado, line 103: Did you confirm that this is the correct node ID for that host? I've changed it recently because we were not able to bond with it using that node ID (cb5ec2a1a2e76e7eead4eca6590bde1977484c63). It'd be good to confirm that for the other ones you are changing as well -- we shouldn't blindly assume they're right just because they're what's in current geth master

gsalgado (changes requested):

p2p/constants.py
```diff
@@ -53,15 +53,15 @@
...
+    "enode://ca6de62fce278f96aea6ec5a2daadb877e51651247cb96ee310a318def462913b653963c155a0ef6c7d50048bba6e6cea881130857413d9f50a621546b590758@34.255.23.113:30303",   # noqa: E501
+    "enode://279944d8dcd428dffaa7436f25ca0ca43ae19e7bcf94a8fb7d1641651f92d121e972ac2e8f381414b80cc8e5555811c2ec6e1a99bb009b3f53c4c69923e11bd8@35.158.244.151:30303",   # noqa: E501
+    "enode://8499da03c47d637b20eee24eec3c356c9a2e6148d6fe25ca195c7949ab8ec2c03e3556126b0d7ed644675e78c4318b08691b7b57de10e5f0d40d05b09238fa0a@52.187.207.27:30303",   # noqa: E501
+    "enode://103858bdb88756c71f15e9b5e09b56dc1be52f0a5021d46301dbbfb7e130029cc9d0d6f73f693bc29b665770fff7da4d34f3c6379fe12721b5d7a0bcb5ca1fc1@191.234.162.198:30303",   # noqa: E501
+    "enode://715171f50508aba88aecd1250af392a45a330af91d7b90701c436b618c86aaa1589c9184561907bebbb56439b8f8787bc01f49a7c77276c58c1b09822d75e8e8@52.231.165.108:30303",   # noqa: E501
+    "enode://5d6d7cd20d6da4bb83a1d28cadb5d409b64edf314c0335df658c1a54e32c7c4a7ab7823d57c39b6a757556e68ff1df17c748b698544a55cb488b52479a92b60f@104.42.217.25:30303",   # noqa: E501
```
gsalgado, line 64: ISTM most of them were not actually changed at all -- it's just the extra whitespace before the `#` sign. Would you mind removing that so we can easily see what has actually changed?

pipermerriam: I think the diff is showing that then node ids themselves didn't change, only the added space, otherwise the enode URI would be in the darker green?

PR #1786, opened 2020-06-15
Fix beam sync crash with clique consensus

cburgdorf:
### What was wrong?

Beam sync is crashing when being used with clique sync. This is due to the `consensus_context_class` not being properly used.

### How was it fixed?

Set `consensus_context_class` in `BeamChain.configure(..)`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


cburgdorf:
We should have a test for that but I pushed it into an issue for now https://github.com/ethereum/trinity/issues/1787

Issue #1788, opened 2020-06-15
Sync not following new blocks


cburgdorf:
### What is wrong?

Whenever beam sync catches up to the tip of the chain it simply doesn't seem to monitor for new incoming blocks anymore.

```
    INFO  2020-06-15 16:02:37,960            BeamSyncer  Downloading 8 block bodies for uncle validation, before <BlockHeader #2878389 eb36b56a>
    INFO  2020-06-15 16:02:38,003            BeamSyncer  Getting recent block data for uncle validation
    INFO  2020-06-15 16:02:38,155            BeamSyncer  Have all data needed for Beam validation, continuing...
[...]

PR #1790, opened 2020-06-16
Fix BaseHeaderChainSyncer not following freshly created blocks

cburgdorf:
### What was wrong?

As explained in #1788 our `BaseHeaderChainSyncer` currently fails to follow new blocks. It catches up to the tip and then stalls forever.

### How was it fixed?

It looks like this is a case of https://github.com/ethereum/snake-charmers-tactical-manual/pull/70/files (cc @gsalgado ) and I was able to get it to work with this fix.

However, this is breaking one of our tests and I'm beginning to think the proper fix would be to break out of this `yield await` when the `SkeletonSyncer` was canceled.

https://github.com/ethereum/trinity/blob/615d6bcc871ac8c8ae795a39fb55ac6745650cb2/trinity/sync/common/headers.py#L119-L122

Trying to figure out how to do that...

### To-Do

- [x] Clean up commit history

- no release notes entry because the bug did not exist in the past release

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/common/headers.py
+++ b/trinity/sync/common/headers.py
@@ -20,7 +20,8 @@ from typing import (
     NamedTuple,
 )
 
-from async_service import Service, background_asyncio_service
+import async_service
+from async_service import Service, background_asyncio_service, external_asyncio_api
 
 from eth.abc import BlockHeaderAPI
 
@@ -117,8 +118,14 @@ class SkeletonSyncer(Service, Generic[TChainPeer]):
         self._fetched_headers = asyncio.Queue(max_pending_headers)
 
     async def next_skeleton_segment(self) -> AsyncIterator[Tuple[BlockHeader, ...]]:
+        # @external_asyncio_api can not be used on `next_skeleton_segment` directly because
+        # it returns a generator. Maybe that could be fixed though?
+        @external_asyncio_api
+        async def _cancellable_fetch_headers_get(_: Service) -> Tuple[BlockHeader, ...]:
+            return await self._fetched_headers.get()
+
         while self.manager.is_running:
-            yield await self._fetched_headers.get()
+            yield await _cancellable_fetch_headers_get(self)
             self._fetched_headers.task_done()
 
     async def run(self) -> None:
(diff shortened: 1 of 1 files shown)


trinity/sync/common/headers.py
```diff
@@ -923,6 +923,10 @@ async def _build_skeleton(self) -> None:
                 self.logger.debug("At or behind peer %s, skipping skeleton sync", peer)
             else:
                 async with self._get_skeleton_syncer(peer) as syncer:
+                    task = asyncio.create_task(self._full_skeleton_sync(syncer))
+                    await asyncio.wait(
+                        [task, syncer.manager.wait_finished()],
+                        return_when=asyncio.FIRST_COMPLETED)
                     await self._full_skeleton_sync(syncer)
```
gsalgado, line 930: This doesn't look like the anti-pattern I described in https://github.com/ethereum/snake-charmers-tactical-manual/pull/70. I haven't looked into how `_full_skeleton_sync()` is implemented, but I imagine it will crash somehow if `syncer` terminates

cburgdorf: Ah ok. I haven't fully groked yet why the code broke (could be a fall out from the migration to `async-service`).

What happens is that this loop of starting new skeleton syncs doesn't work anymore.

https://github.com/ethereum/trinity/blob/615d6bcc871ac8c8ae795a39fb55ac6745650cb2/trinity/sync/common/headers.py#L919-L926

Whenever the `SkeletonSyncer` cancels itself, we continue to hang on `self._full_skeleton_sync(syncer)` forever. The changes in this PR fixes it but it's probably not the right fix (hence the warnings). But I have an another idea to try out :)

cburgdorf: Ok, warnings are gone when I add `task.cancel()` but now it fails one of our tests :see_no_evil: Investigating...

carver:
Take a look at #1792 which solves the problem AIUI.

trinity/sync/common/headers.py
```diff
@@ -117,9 +118,18 @@ def __init__(self,
         self._fetched_headers = asyncio.Queue(max_pending_headers)
 
     async def next_skeleton_segment(self) -> AsyncIterator[Tuple[BlockHeader, ...]]:
+        # @external_asyncio_api can not be used on `next_skeleton_segment` directly because
+        # it returns a generator. Maybe that could be fixed though?
+        @external_asyncio_api
```
cburgdorf, line 123: If we can not use `@external_asyncio_api` directly on `async def next_skeleton_segment` then this feels like abusing the decorator I guess

carver: Yeah, agreed. It doesn't take too much work in async-service to expose the new API though, AFAICT.

carver: See https://github.com/ethereum/async-service/pull/70

PR #1791, opened 2020-06-16
Also monitor NewBlockHashes in ETHChainTipMonitor

cburgdorf:
### What was wrong?

While working on #1788 I noticed that we can recognize changes at the tip of the chain a little faster if we also listen for `NewBlockHashes` in addition to `NewBlock`.

### How was it fixed?

Add `NewBlockHashes` to the observed events of `ETHChainTipMonitor`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1791.feature.rst
+++ b/newsfragments/1791.feature.rst
@@ -1,0 +1,1 @@
+When monitoring the tip of the chain, also monitor the ``NewBlockHashes`` event.
--- a/newsfragments/1791.feature.rst
+++ b/newsfragments/1791.feature.rst
@@ -1,1 +1,1 @@
-When monitoring the tip of the chain, also monitor the ``NewBlockHashes`` event.
+Keep up with the tip of the chain a bit better. (By also listening to ``NewBlockHashes``)
--- a/trinity/protocol/eth/monitors.py
+++ b/trinity/protocol/eth/monitors.py
@@ -3,4 +3,4 @@ from trinity.protocol.eth import commands
 
 
 class ETHChainTipMonitor(BaseChainTipMonitor):
-    subscription_msg_types = frozenset({commands.NewBlock})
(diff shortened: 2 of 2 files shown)


newsfragments/1791.feature.rst
```diff
@@ -0,0 +1 @@
+When monitoring the tip of the chain, also monitor the ``NewBlockHashes`` event.
```
carver, line 1: This description is helpful for contributors, in the `internal` section. But I do think it belongs here as a feature. From a user's perspective, maybe something like this is clearer:
```suggestion
Keep up with the tip of the chain a bit better. (By also listening to ``NewBlockHashes``)
```

trinity/protocol/eth/monitors.py
```diff
@@ -3,4 +3,4 @@
 
 
 class ETHChainTipMonitor(BaseChainTipMonitor):
-    subscription_msg_types = frozenset({commands.NewBlock})
+    subscription_msg_types = frozenset({commands.NewBlock, commands.NewBlockHashes})
```
carver, line 6: Cool that this is the only change needed. I see that this relies on `peer_pool.highest_td_peer`. Does the peer pool already update the total difficulty for the peer on a `NewBlockHashes` command? (Is it even possible?)

newsfragments/1791.feature.rst
cburgdorf: Yep, that sounds better :+1:

trinity/protocol/eth/monitors.py
cburgdorf: >Does the peer pool already update the total difficulty for the peer on a `NewBlockHashes` command?

It doesn't and I don't think it is possible.

But even if the `NewBlockHashes` do not move the `head_td` forward, monitoring these does still improve our tip tracking as long as the peer is still ahead of us (based on the last `NewBlock`)

PR #1792, opened 2020-06-16
Fix skeleton recycling

### LINKED ISSUE #1788: Sync not following new blocks

carver:
### What was wrong?

Fix #1788 

### How was it fixed?

Parallel work, with a nod to #1790 

When waiting on the next header batch to show up in `next_skeleton_segment()`, exit the iterable if the service exits (rather than hanging, waiting on the next `Queue` element to appear).

Added a test that fails before this bugfix.

As a side-effect, this seemed to uncover some barely-related bug about handling duplicate headers during sync. If the parent of a chain of headers was missing, and the headers were duplicates of in-flight headers being processed, then it would queue them up despite being duplicates. Fixed that bug too.

### To-Do

- [ ] Clean up commit history
- [ ] extract a code block that filters duplicate headers into its own method: 
https://github.com/ethereum/trinity/pull/1792/files#diff-e2a97e2edf2590b9aa6a324ee9648a5cR195
- [ ] internal changelog for trimmed debug2 log, plus an extended hypothesis deadline, for two fewer flaky tests

#### Cute Animal Picture

### DIFF
--- a/p2p/logging.py
+++ b/p2p/logging.py
@@ -1,0 +1,10 @@
+from typing import (
+    Any,
+)
+
+from p2p._utils import trim_middle
+from p2p.constants import LONGEST_ALLOWED_LOG_STRING
+
+
+def loggable(log_object: Any) -> str:
+    return trim_middle(str(log_object), LONGEST_ALLOWED_LOG_STRING)
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -624,4 +624,14 @@ async def wait_for_head(headerdb, header, sync_timeout=10):
             else:
                 break
         assert header_at_block == header
-    await asyncio.wait_for(wait_loop(), sync_timeout)
+    try:
+        await asyncio.wait_for(wait_loop(), sync_timeout)
+    except asyncio.TimeoutError:
+        canonical_head = headerdb.get_canonical_head()
+        logging.error(
+            "Could not finish syncing to %s within %ds, only arrived at %s",
+            header,
+            sync_timeout,
+            canonical_head,
+        )
+        raise
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -5,6 +5,7 @@ import uuid
 from async_service import Service, background_asyncio_service
 from eth.consensus import ConsensusContext
 from eth.db.atomic import AtomicDB
+from eth.db.schema import SchemaV1
 from eth.exceptions import HeaderNotFound
 from eth.vm.forks.petersburg import PetersburgVM
 from eth_utils import decode_hex
@@ -86,7 +87,7 @@ def chaindb_with_gaps(chaindb_fresh, chaindb_1000):
 
 
 @pytest.mark.asyncio
-async def test_skeleton_syncer(request, event_loop, event_bus, chaindb_fresh, chaindb_1000):
+async def test_fast_syncer(request, event_loop, event_bus, chaindb_fresh, chaindb_1000):
 
     client_context = ChainContextFactory(headerdb__db=chaindb_fresh.db)
     server_context = ChainContextFactory(headerdb__db=chaindb_1000.db)
@@ -115,6 +116,7 @@ async def test_skeleton_syncer(request, event_loop, event_bus, chaindb_fresh, ch
 
             head = chaindb_fresh.get_canonical_head()
             assert head == chaindb_1000.get_canonical_head()
+            # TODO assert that the block transactions and uncles are present too.
 
 
 @pytest.mark.asyncio
@@ -413,9 +415,9 @@ async def test_header_syncer(request,
                              event_loop,
                              event_bus,
                              chaindb_fresh,
-                             chaindb_20):
+                             chaindb_1000):
     client_context = ChainContextFactory(headerdb__db=chaindb_fresh.db)
-    server_context = ChainContextFactory(headerdb__db=chaindb_20.db)
+    server_context = ChainContextFactory(headerdb__db=chaindb_1000.db)
     peer_pair = LatestETHPeerPairFactory(
         alice_peer_context=client_context,
         bob_peer_context=server_context,
@@ -433,14 +435,44 @@ async def test_header_syncer(request,
         async with run_peer_pool_event_server(
             event_bus, server_peer_pool, handler_type=ETHPeerPoolEventServer
         ), background_asyncio_service(ETHRequestServer(
-            event_bus, TO_NETWORKING_BROADCAST_CONFIG, AsyncChainDB(chaindb_20.db),
+            event_bus, TO_NETWORKING_BROADCAST_CONFIG, AsyncChainDB(chaindb_1000.db),
         )):
 
-            server_peer.logger.info("%s is serving 20 blocks", server_peer)
-            client_peer.logger.info("%s is syncing up 20", client_peer)
+            server_peer.logger.info("%s is serving 1000 blocks", server_peer)
+            client_peer.logger.info("%s is syncing up 1000", client_peer)
+
+            # Artificially split header sync into two parts, to verify that
+            #   cycling to the next sync works properly. Split by erasing the canonical
+            #   lookups in a middle chunk. We have to erase a range of them because of
+            #   the way that the skeleton sync skips over headers on the first request,
+            #   it expects the peer to have all headers in the 192-header gaps in between.
+            erase_block_numbers = range(500, 700)
+            erased_canonicals = []
+            for blocknum in erase_block_numbers:
+                dbkey = SchemaV1.make_block_number_to_hash_lookup_key(blocknum)
+                canonical_hash = chaindb_1000.db[dbkey]
+                erased_canonicals.append((dbkey, canonical_hash))
+                del chaindb_1000.db[dbkey]
 
             async with background_asyncio_service(client):
-                await wait_for_head(chaindb_fresh, chaindb_20.get_canonical_head())
+                target_head = chaindb_1000.get_canonical_block_header_by_number(
+                    # TODO? Look back from the first erased block number, because the skeleton
+                    #   may not properly fill in right up to the missing tip
+                    erase_block_numbers[0] - 1
+                )
+                await wait_for_head(chaindb_fresh, target_head, sync_timeout=20)
+
+                # TODO validate that the skeleton syncer has cycled??
+
+                # Replace the missing headers so that syncing can resume
+                for dbkey, canonical_hash in erased_canonicals:
+                    chaindb_1000.db[dbkey] = canonical_hash
+
+                # Do we have to do anything here to have the server notify the client
+                #   that it's capable of serving more headers now?
+
+                await wait_for_head(chaindb_fresh, chaindb_1000.get_canonical_head())
+
 
 
 @pytest.mark.asyncio
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -444,8 +444,8 @@ async def test_header_syncer(request,
             # Artificially split header sync into two parts, to verify that
             #   cycling to the next sync works properly. Split by erasing the canonical
             #   lookups in a middle chunk. We have to erase a range of them because of
-            #   the way that the skeleton sync skips over headers on the first request,
-            #   it expects the peer to have all headers in the 192-header gaps in between.
+            #   how the skeleton syncer asks for every ~192 headers. The skeleton request
+            #   would skip right over a single missing header.
             erase_block_numbers = range(500, 700)
             erased_canonicals = []
             for blocknum in erase_block_numbers:
@@ -456,8 +456,6 @@ async def test_header_syncer(request,
 
             async with background_asyncio_service(client):
                 target_head = chaindb_1000.get_canonical_block_header_by_number(
-                    # TODO? Look back from the first erased block number, because the skeleton
-                    #   may not properly fill in right up to the missing tip
                     erase_block_numbers[0] - 1
                 )
                 await wait_for_head(chaindb_fresh, target_head)
@@ -473,7 +471,7 @@ async def test_header_syncer(request,
                     chaindb_1000.db[dbkey] = canonical_hash
 
                 # Do we have to do anything here to have the server notify the client
-                #   that it's capable of serving more headers now?
+                #   that it's capable of serving more headers now? ... Apparently not.
 
                 await wait_for_head(chaindb_fresh, chaindb_1000.get_canonical_head())
 
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -623,4 +623,5 @@ async def wait_for_head(headerdb, header, sync_timeout=10):
                 await asyncio.sleep(0.1)
             else:
                 break
+        assert header_at_block == header
     await asyncio.wait_for(wait_loop(), sync_timeout)
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -460,7 +460,11 @@ async def test_header_syncer(request,
                     #   may not properly fill in right up to the missing tip
                     erase_block_numbers[0] - 1
                 )
-                await wait_for_head(chaindb_fresh, target_head, sync_timeout=20)
+                await wait_for_head(chaindb_fresh, target_head)
+
+                # gut check that we didn't skip past the erased range of blocks
+                head = chaindb_fresh.get_canonical_head()
+                assert head.block_number < erase_block_numbers[0]
 
                 # TODO validate that the skeleton syncer has cycled??
 
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -478,7 +478,6 @@ async def test_header_syncer(request,
                 await wait_for_head(chaindb_fresh, chaindb_1000.get_canonical_head())
 
 
-
 @pytest.mark.asyncio
 async def test_header_gapfill_syncer(request,
                                      event_loop,
--- a/tests/integration/test_trinity_cli.py
+++ b/tests/integration/test_trinity_cli.py
@@ -243,7 +243,7 @@ async def test_does_not_throw_errors_on_short_run(command, unused_tcp_port):
 @pytest.mark.parametrize(
     'command,expected_stderr_logs,unexpected_stderr_logs,expected_file_logs,unexpected_file_logs',
     (
-        (
+        pytest.param(
             # Default run without any flag
             ('trinity',),
             # Expected stderr logs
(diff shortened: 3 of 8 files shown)


trinity/sync/common/headers.py
```diff
@@ -118,8 +118,28 @@ def __init__(self,
 
     async def next_skeleton_segment(self) -> AsyncIterator[Tuple[BlockHeader, ...]]:
         while self.manager.is_running:
-            yield await self._fetched_headers.get()
-            self._fetched_headers.task_done()
+            header_task = asyncio.create_task(self._fetched_headers.get())
```
gsalgado, line 121: Maybe we could wrap `self._fetched_headers.get()` in a method like this:
```
async def _get_fetched_headers(self):
    headers = await self._fetched_headers.get()
    if not headers:
        if self.manager.is_running:
            raise Exception("Got empty headers while service still running, this shouldn't happen")
        break
    yield headers
```

Then in `SkeletonSyncer.run()` we'd add a `finally:` block to clear `self._fetched_headers` and call `self._fetched_headers.put_nowait([])`, causing the generator above to exit. That way we don't have to worry about scheduling/cancelling tasks

carver: Yeah, I see where you're going, and I also don't like all the boilerplate in the current solution. It makes me yearn for the old style from `BaseService`:
```py
try:
  headers = await self.wait(self._fetched_headers.get())
except OperationCancelled:
  break
else:
  yield headers
```

Maybe we can add this to `async-service`? I guess something like:
```py
try:
  header = await self.manager.cancel_on_exit(self._fetched_headers.get())
except CancelledByService:  # or even just an asyncio.CancelledError?
  break
else:
  yield headers
```

trinity/sync/full/chain.py
```diff
@@ -214,39 +216,50 @@ async def _sync_from_headers(
...
+                        "Parent %s of first header %s is already present during sync, ignoring...",
+                        parent_header,
+                        fresh_headers[0],
+                    )
+                    pass
+
                 # Re-register the header tasks, which will now succeed
-                task_integrator.register_tasks(new_headers, ignore_duplicates=True)
+                # Be sure to filter out any duplicates, so we don't re-register them for processing
+                new_headers = task_integrator.register_tasks(fresh_headers, ignore_duplicates=True)
```
carver, line 262: The name changes from `new_headers` to `fresh_headers` above is important because `new_headers` is the value used below for queuing up activities like downloading the black receipts. In the non-exceptional case, `new_headers` is already filtered for duplicates. But in this exceptional case, we were allowing duplicates through by ignoring the returned (filtered) value of `register_tasks()`.

I intend to do the above TODO during this PR, but after a first review:
> extract the following try/except to a function that returns the non-duplicates

I think it's a bit clearer to see the change required before I refactor the code block out into a new method.

trinity/sync/common/headers.py
gsalgado: Yeah, I think that would make sense in async-service, but maybe `run_and_await_task()` would be a better name?

cburgdorf: :+1: for adding such a method to `async-service`. On the naming, I initially found `cancel_on_exit` to be more descriptive but now that I thought about it a big longer, I find that `run_and_await_tasks` actually aligns very nicely with `run_task` just with the difference that this does in fact also awaits the task which is precisely what it does :+1:

gsalgado: Actually, I'm not sure adding this new API to async-service is warranted as it'd only be used here, and ISTM we only need it here because we're passing a running Service across service boundaries. If instead we did something like this:

```
diff --git a/trinity/sync/common/headers.py b/trinity/sync/common/headers.py
index 5acb09f2..e3a62e36 100644
@@ -923,7 +923,8 @@ class BaseHeaderChainSyncer(Service, HeaderSyncerAPI, Generic[TChainPeer]):
                 self.logger.debug("At or behind peer %s, skipping skeleton sync", peer)
             else:
                 async with self._get_skeleton_syncer(peer) as syncer:
-                    await self._full_skeleton_sync(syncer)
+                    await syncer.manager.run_task(self._full_skeleton_sync, syncer)
+                    await syncer.manager.wait_finished()
 
     @contextlib.asynccontextmanager
     async def _get_skeleton_syncer(
```

I believe everything would just work (because when we SkeletonSyncer terminates, either by completing or by losing the peer connection, we'd also cancel the task running `full_skeleton_sync()`. This kind of suggests to me that it'd make more sense to have the `SkeletonSyncer` merged with `BaseHeaderChainSyncer`. Failing that, another alternative would be to change `SkeletonSyncer`'s constructor to take the `_full_skeleton_sync()` function as an argument, so that we don't have to access its internal `.manager` from `BaseHeaderChainSyncer`

tests/core/p2p-proto/test_sync.py
```diff
@@ -433,14 +435,44 @@ async def test_header_syncer(request,
...
+            #   lookups in a middle chunk. We have to erase a range of them because of
+            #   the way that the skeleton sync skips over headers on the first request,
+            #   it expects the peer to have all headers in the 192-header gaps in between.
+            erase_block_numbers = range(500, 700)
+            erased_canonicals = []
+            for blocknum in erase_block_numbers:
+                dbkey = SchemaV1.make_block_number_to_hash_lookup_key(blocknum)
+                canonical_hash = chaindb_1000.db[dbkey]
+                erased_canonicals.append((dbkey, canonical_hash))
+                del chaindb_1000.db[dbkey]
```
cburgdorf, line 455: Clever. I guess we could use the same approach to test the gap syncer "uncle attack" scenario (different test). We wouldn't have a third friendly peer then, just a a serving peer that changed its mind :)

trinity/sync/common/headers.py
cburgdorf: >I'm not sure adding this new API to async-service is warranted as it'd only be used here

Right, but it seems that it would be useful for every service that wants to expose an async API for data retrieval. But guess what, I just went to the `async-service` docs and found this little gem: 

```python
from async_service import Service, background_asyncio_service, external_asyncio_api

class MyService(Service):
    async def run(self):
        ...

    @external_asyncio_api
    async def get_thing(self):
        ...

service = MyService()

# this will fail because the service isn't running yet
await service.get_thing()

async with background_asyncio_service(service) as manager:
    thing = await service.get_thing()

    # now cancel the service
    manager.cancel()

    # this will fail because the service is cancelled.
    thing = await service.get_thing()
``` 

It sounds like all we have to do is to decorate `next_skeleton_segment` with and then it would work as intended raising an `async_service.exceptions.LifecycleError` on cancellation.

>This kind of suggests to me that it'd make more sense to have the `SkeletonSyncer` merged with `BaseHeaderChainSyncer`.

Mmh, I'm not keen on merging 400 lines of mostly unrelated code into `BaseHeaderChainSyncer` because of this. It will generally lower the cohesion of `BaseHeaderChainSyncer` and will make it harder to grasp imho.

>Failing that, another alternative would be to change SkeletonSyncer's constructor to take the `_full_skeleton_sync()` function as an argument, so that we don't have to access its internal `.manager`

I could be ok with that but I think it will reduce the level of abstraction that the `SkeletonSyncer` provides in its current form. From an architectural point of view I like the fact that it simply exposes an `AsyncIterator[Tuple[BlockHeader, ...]]:` for any consumer to use in whichever way they want.

trinity/sync/common/headers.py
```diff
@@ -118,8 +118,24 @@ def __init__(self,
 
     async def next_skeleton_segment(self) -> AsyncIterator[Tuple[BlockHeader, ...]]:
```
cburgdorf, line 119: Unrelated to the issue at hand but I think that the name of this API is a bit misleading. The name suggests: Whenever you need a new segment, call this API. When in fact, this returns a generator that you `async for` over or call `__anext__` on (So, it is the *thing* that is returned that has the *next-ability*, not the method itself)

Naming suggestions:

- `get_segments`
- `observe_segments`
- `yield_segments`

cburgdorf:
image

Sorry, I wanted to convert **my** PR back to draft (to show case another approach) and accidentally ended up on your PR.

trinity/sync/common/headers.py
cburgdorf: I was curious to try out `external_asyncio_api`

https://github.com/ethereum/trinity/pull/1790/commits/0f19bcba68bf642f52ec22ce3ceb758347803600

It's a bit ugly because it can not be applied on `next_skeleton_segment` directly since it returns an `AsyncGenerator` so I had to apply it one level deeper instead. But maybe `external_asyncio_api` can be tweaked to also work in this scenario.

gsalgado: > I could be ok with that but I think it will reduce the level of abstraction that the SkeletonSyncer provides in its current form. From an architectural point of view I like the fact that it simply exposes an AsyncIterator[Tuple[BlockHeader, ...]]: for any consumer to use in whichever way they want.

Fair enough. I'm not really arguing in favor of merging them but note that `SkeletonSyncer` is only used in `BaseHeaderChainSyncer` and it is not even tested in isolation, so it's arguable that this abstraction is hurting us more than it is helping.

carver: > On the naming, I initially found `cancel_on_exit` to be more descriptive but now that I thought about it a big longer, I find that `run_and_await_tasks` actually aligns very nicely with `run_task` just with the difference that this does in fact also awaits the task which is precisely what it does

Maybe just `await_task` to be a bit more concise?

carver:
> That said, I still don't understand where this is coming from and why this wasn't a problem in earlier versions.

Yeah, I also would like to know this. I think this is the commit where I would checkout to see what's going on: https://github.com/ethereum/trinity/commit/59b5ae65d9a7fc73c098a0da949fe96170ad529e (the parent of the commit before the syncer was made into an async-service).

The latest versions push a lot of duplicate headers, so I'd want to add more logs about which headers are coming from the header syncer in which order, and check them out in `test_fast_syncer` (formerly mis-named `test_skeleton_syncer`). The latest version has the header chain catch up to the peer, then exit and re-sync, producing duplicates (I guess because the blocks aren't done persisting yet by the time the second header sync starts). So why is this happening now, and not before?

carver:
Some alternatives, all of which seem to pass tests:
- 07df890 - decorate a method with `@external_asyncio_api` and catch the `LifecycleError`
  - pro: less boilerplate than managing tasks manually
  - con: the decorator name is incorrect (we are not using it as an external API, we are wrapping it in the iterator)
  - con: the decorator name doesn't explain how/why we're using it: to cancel if the service exits
- 1ca427b - move all the service-handling logic into a decorator called `@stop_iteration_on_service_halt`
  - pro: all the service/task handling logic gets shoved into one place
  - con: another awkward use of `@external_asyncio_api` required, but at least outside of the syncer
  - con: not really any shorter than just manually handling the tasks, once you account for the decorator
  - note: CI failures seem unrelated
- cadddd9dd - run `_full_skeleton_sync()` as a task of the skeleton syncer, so it cancels rather than hanging on the dead skeleton syncer service.
  - pro: shortest solution by far
  - con: any other potential consumer of the skeleton segment iterator would have to remember to deal with the same issue
  - con: leaves the general problem unsolved of how to wait on a task only until a service exits
- fc658a4115ae - use a new async-service `manager.await_task` API
  - pro: it works, and is the smallest change to trinity
  - con: depends on a new async-service release, and all the discussion that goes along with a new API
- no code - could write a `@stop_iteration_on_service_halt` in async-service, maybe depending on a new `await_task()` API. But it doesn't save much code compared to fc658a4115ae

Current favorite option: cadddd9dd and open a PR for async-service in the background to eventually migrate to fc658a4115ae.

BTW, I think this is exposing some other issues that we haven't noticed in live integration tests because the syncer was getting stuck. So far it just looks like it's a log that's too big: "asyncio.exceptions.LimitOverrunError: Separator is found, but chunk is longer than limit"

carver:
@cburgdorf @gsalgado Look okay?

tests/core/p2p-proto/test_sync.py
```diff
@@ -623,4 +656,15 @@ async def wait_loop():
...
+        await asyncio.wait_for(wait_loop(), sync_timeout)
+    except asyncio.TimeoutError:
+        canonical_head = headerdb.get_canonical_head()
+        logging.error(
+            "Could not finish syncing to %s within %ds, only arrived at %s",
+            header,
+            sync_timeout,
+            canonical_head,
+        )
+        raise
```
gsalgado, line 670: Maybe it'd be better to raise a new TimeoutError with the msg above?

tests/integration/test_trinity_cli.py
```diff
@@ -254,21 +254,25 @@ async def test_does_not_throw_errors_on_short_run(command, unused_tcp_port):
             {'Started main process', 'Logging initialized'},
             # Unexpected file logs
             {'>>> ping'},
+            id="noflag:starts-both,loginit-file,ping-none",
```
gsalgado, line 257: I didn't know about this. It will certainly help next time one of those fail

p2p/logging.py
```diff
@@ -0,0 +1,10 @@
+from typing import (
+    Any,
+)
+
+from p2p._utils import trim_middle
+from p2p.constants import LONGEST_ALLOWED_LOG_STRING
+
+
+def loggable(log_object: Any) -> str:
+    return trim_middle(str(log_object), LONGEST_ALLOWED_LOG_STRING)
```
gsalgado, line 10: I wonder if this should be enforced by `ExtendedDebugLogger`?  It looks like this is not the first time we're bitten by this: https://github.com/ethereum/trinity/issues/638

trinity/sync/common/headers.py
```diff
@@ -923,7 +924,13 @@ async def _build_skeleton(self) -> None:
                 self.logger.debug("At or behind peer %s, skipping skeleton sync", peer)
             else:
                 async with self._get_skeleton_syncer(peer) as syncer:
-                    await self._full_skeleton_sync(syncer)
+                    # We cannot simply await self._full_skeleton_sync(syncer) here because
+                    #   if the service exits, we get stuck waiting for the next header to appear
+                    #   from skeleton_segments(). By running it async and waiting until the manager
+                    #   exits, we ensure that _full_skeleton_sync gets cancelled when the service
+                    #   exits, even if it's hanging on skeleton_segments().
+                    syncer.manager.run_task(self._full_skeleton_sync, syncer)
```
cburgdorf, line 932: From the comments in `async-service` it sounds like the type hints were supposed to disallow usage of `syncer.manager` so that external manager access would be limited to `syncer.get_manager()` which does not expose the task scheduling functionality. This doesn't seem to be working as intended though (otherwise I would have expected to see a `#type: ignore` here).

Anyway, since we are planning to eventually rewrite this code (see https://github.com/ethereum/async-service/pull/70 ) I'm fine with this solution for now. I will open an issue in `async-service` about the type hinting issue.

cburgdorf: Opened up an issue here: https://github.com/ethereum/async-service/issues/71

cburgdorf:
I'm going ahead and merge this to prepare for a Trinity release today.

carver:
> I'm going ahead and merge this to prepare for a Trinity release today.

Woah, a bit quick on the trigger, buddy. That one had a lot of messy commits in it, which are now on master. Best to avoid merging anything for someone, when it has unchecked to-dos.

In a stable project, I'd say that it should be forbidden. (but since we're just in alpha, it's just mildly irritating...)

cburgdorf:
Ah, I didn't check the commit history but I personally became more relaxed about commit history ever since we started using towncrier to generate nice release notes. 

Sorry, for the trouble, but I didn't want to postpone the release until Monday and rushed for it 😳 :pray:

trinity/sync/common/headers.py
pipermerriam: Not a problem with `async-service`.  I'm pretty sure that if you update the type hint for `_get_skeleton_syncer()` to return a `ServiceAPI` then mypy would start complaining.

cburgdorf: Well if the type hint needs to be `ServiceAPI` for that to work then it is unlikely to prevent access in practice. Because most of the time, as in this case, the type hint will be the concrete service class (e.g. `SkeletonSyncer`)

pipermerriam: Agreed.  Beyond moving `Service.manger` back to the private `Service._manager` I'm not sure how this would be addressed...

carver:
> I personally became more relaxed about commit history ever since we started using towncrier to generate nice release notes.

It's not a huge deal, but having good git commit hygiene is really valuable. When you're trying to track down a bug, tools like `git blame`, `git bisect`, and `git log -S` can speed things up immensely. This merge adds a lot of unnecessary code churn that makes those tools more cumbersome.

cburgdorf:
>It's not a huge deal, but having good git commit hygiene is really valuable

Yeah, I generally agree to that :+1:

PR #1800, opened 2020-06-18
Simplifications to chain, state machine and DB

ralexstokes:
### What was wrong?

Simplifying some parts of the chain, state machine and db abstractions to get ready to implement fork choice.

### How was it fixed?

Mostly removing stuff we don't need and sorting out how to separate our noting of persisting a block from running the fork choice.

- Remove `state_transition` abstraction from the `state_machine` as it was just adding extra overhead
- Making progress towards encapsulating the `chain_db` behind the `chain` abstraction; ultimately the db should be an implementation detail of the chain
- adding `from_genesis` constructors for the `chain` and `chain_db` abstractions; the idea is that to make a "fresh" instance at genesis, you call the `from_genesis` constructor. otherwise the underlying database has already been initialized and you can simply instantiate with `__init__`.
- some other minor API clean ups for things we aren't using or don't need

#### Cute Animal Picture


ralexstokes:
happy to address post-merge feedback but don't want a chain of PRs here (or one massive PR)...

PR #1802, opened 2020-06-19
ComponentManager now stops if a component terminates

gsalgado:
The ComponentManager used to run components as background tasks and rely on the
BaseIsolatedComponent implementation to handle premature termination and send a ShutdownRequest to
the component manager. Now CM keeps track of the tasks running each component and terminates as
soon as any of them terminate.

With this change we may no longer need the whole ShutdownRequest mechanism. If there are no
objections I'll submit a separate PR to remove that


gsalgado:
Yeah, I suppose it'd be nice to have the ability to have short-lived components, but as you pointed out we have never had any, and we don't get new components often, so I'd rather wait until we have an actual use case before we add the extra complexity to the component manager. Also, right now if a component terminates the base `*IsolatedComponent` implementation will send a `ShutdownRequest`, so this is not changing the existing behavior -- it's only making sure we don't fire-and-forget our component tasks, which may lead to unexpected behavior that is really hard to track down as the component errors may not be reported at all

Issue #1795, opened 2020-06-18
Let JSON RPC HTTP listen address be configurable via flag


cburgdorf:
### What is wrong?

Currently, the address that our HTTP server listens on is hardcoded to:

https://github.com/ethereum/trinity/blob/3172e4e92a45aa608fbd761abcf7239eb9b0d1ce/trinity/http/main.py#L26

This is greatly limits the usefulness of Trinity. E.g. when running Trinity on DappNode one can not perform requests against the JSON-RPC API (one  of, if not **the** main use cases for letting Trinity run on a DappNode)

### How can it be fixed

Expose this via a flag. E.g. Geth exposes this as `--rpcaddr 0.0.0.0`. I think that `rpc` is a bit ambigious (after all, one can also use IPC to access the JSON-RPC API). Therfore I think we might want to pick one of these names:

- `http-address`
- `http-listen-address`

carver:
Default should still be 127.0.0.1, but seems reasonable to me :+1:

PR #1803, opened 2020-06-19
Make HTTP Listen Address configurable

cburgdorf:
### What was wrong?

As explained in #1795, Trinity does currently not allow accessing its JSON RPC HTTP API from an external address. This should be configureable.

### How was it fixed?

1. Exposed the listen address as `--http-listen-address` but kept the default at `127.0.0.1`

2. Fixed some casing inconsistencies in the help texts (Server > server)

3. Renamed `--rpcport` to `--http-port` to be more aligned and precise (RPC doesn't necessarily refer to the HTTP API as it is an umbrella term that includes RPC access via HTTP, IPC and potentially WS in the future) 

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/components/builtin/json_rpc/component.py
+++ b/trinity/components/builtin/json_rpc/component.py
@@ -103,12 +103,18 @@ class JsonRpcServerComponent(AsyncioIsolatedComponent):
         arg_parser.add_argument(
             "--disable-rpc",
             action="store_true",
-            help="Disables the JSON-RPC Server",
+            help="Disables the JSON-RPC server",
         )
         arg_parser.add_argument(
             "--enable-http",
             action="store_true",
-            help="Enables the HTTP Server",
+            help="Enables the HTTP server",
+        )
+        arg_parser.add_argument(
+            "--http-listen-address",
+            type=str,
+            help="Address for the HTTP server to listen on",
+            default="127.0.0.1",
         )
         arg_parser.add_argument(
             "--rpcport",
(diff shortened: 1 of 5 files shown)


trinity/components/builtin/json_rpc/component.py
```diff
@@ -103,15 +103,21 @@ def configure_parser(cls, arg_parser: ArgumentParser, subparser: _SubParsersActi
...
         arg_parser.add_argument(
             "--enable-http",
             action="store_true",
-            help="Enables the HTTP Server",
+            help="Enables the HTTP server",
         )
         arg_parser.add_argument(
-            "--rpcport",
+            "--http-listen-address",
+            type=str,
```
cburgdorf, line 115: I initially thought we would want this to be of type `ipaddress` but geth allows this to be bound to a string and `HTTPServer` exposes this as a string, too so I guess binding to some DNS name might be a valid thing (I have no idea, I'm not a networking expert)

cburgdorf:
@hwwhww pinged you for a review since you've last worked on this and I'm also renaming a flag that you might be using.

hwwhww:
@cburgdorf  Hi!
Other files that are using `--rpcport` flags need to be updated too:
1. https://github.com/ethereum/trinity/blob/48188a06d9b9df79eae6dd17b559f77c08e5e600/docs/api/api.cli.rst
2. https://github.com/ethereum/trinity/blob/84381f4fe19b18dddf2d2b210bd0780038728acd/eth2/beacon/scripts/run_beacon_nodes.py#L133

cburgdorf:
@hwwhww Ooops, sloppy me! Thanks for pointing me to these places. Should be fixed!

Issue #1798, opened 2020-06-18
Tracking: spec update v0.11.3 -> v0.12.1


ralexstokes:
edit: moved below

ralexstokes:
reposting some notes i made about the spec PRs and things to consider when we update this:

## things we can do while remaining backwards compatible w/ v0.11.3

- [ ] Rename SigningRoot to SigningData (#1740)
    - just a name change but we elect to stick to the spec when we can
    - assuming we have implemented this type (we may not have!)
        - then just `grep` for it and `sed` away
- [ ]  Remove PERSISTENT_ COMMITTEE_PERIOD in favor of SHARD_COMMITTEE_PERIOD (#1783)
    - Naming on a config property
    - will need to find the `process_voluntary_exits` state transition (sub)function and update accordingly
- [ ]  Disallow empty aggregation bits in attestations (#1780)
    - modifying one of our helpers
    - grepping for `is_valid_indexed_attestation` should reveal location
    - i think we deviate from the spec name here
    - NOTE: this addresses the above comment
- [ ] Fix some ValidatorIndex type casting (#1714)
    - we should do this to adhere to spec
- [ ] Minor typing fixes (#1731)
    - mypy usually catches this stuff for us

## things we can put on a v0.12.x branch to get ready for Altona, etc. (or stuff that is outside the STF

- [ ] Adjust INACTIVITY_PENALTY_QUOTIENT to accurately reflect intention (#1712)
    - just constant update
    - two configs (minimal and mainnet)
    - `eth2.beacon.state_machines.forks.*`
        - two forks `Serenity` and `SkeletonLake` that correspond to
        - the minimal and mainnet config profiles
- [ ] Change MAX_ATTESTER_SLASHINGS from 1 to 2 (#1794)
    - just udpating a constant in config
- [ ] Ensure balances remain unchanged for optimal validators during inactivity leak (#1830)
    - this one is substantive and actually would break for v0.11.3
    - if you want to go ahead and implement that's fine but let's put behind a feature flag (of sorts, a global boolean is fine for now) in the mean time
- [ ] Clarify eth1 timestamp can be less than genesis time (#1756)
    - ensure that we don't error on this in our STF (state transition function)
- [ ] Clarify that eth1 blocks must be deep enough in chain for consideration for genesis (#1834)
    - we should make sure we do this but not urgent for getting a block follower on witti
- [ ] Structure rewards/penalties to allow for more granular testing (#1747, #1826)
    - will likely help w/ testing
[...]

PR #1807, opened 2020-06-22
some spec updates for v0.12.x that can be safely implemented in v0.11.3

ralexstokes:
### What was wrong?

Fixes some things in #1798 that are forwards-compatible with v0.11.3 (the current spec version on `master`)

### How was it fixed?

Address the issues there.

#### Cute Animal Picture


eth2/beacon/attestation_helpers.py
```diff
@@ -38,6 +38,11 @@ def validate_indexed_attestation(
     """
     attesting_indices = indexed_attestation.attesting_indices
 
+    if len(attesting_indices) == 0:
+        raise ValidationError(
+            f"No bits set in attestation's aggregation bits, attestation={indexed_attestation}."
```
g-r-a-n-t, line 43: Maybe the error message should be something like:

`f"The indexed attestation's attesting indices is empty, indexed_attestation={indexed_attestation}."`

Although we're basically testing the same condition, it would still be nice to indicate to the user that we failed to validate an indexed attestation and not a raw attestation.

Issue #1806, opened 2020-06-21
Beam sync stalls


cburgdorf:
### What is wrong?

After a while beam sync stops importing blocks. In the attached log line `80319` is the last time it successfully imported a block. At line `86426` it is the first time that the `BaseBodyChainSyncer` reports it has no write capacity.

```
DEBUG  2020-06-20 09:45:52,898   BaseBodyChainSyncer  (progress, queued, max) of bodies, receipts: [(0, 0, 2048)]. Write capacity? False Importing? True
```
Another interesting thing going on is that there are quite a lot of timed out requests for state data and they start shortly after the last successful block import on line `80663`

```
Timed out requsting state data for block #2906107, retrying...
```

[trinity_dappnode.zip](https://github.com/ethereum/trinity/files/4809079/trinity_dappnode.zip)

Notice this is running:

1. On Görli
2. Against a local Geth Görli node
3. On #1805 which fixes a crash that would otherwise take the whole node down reliably

### How can it be fixed

Need to dig in.

carver:
So the problem seems to start here:
```
   DEBUG  2020-06-20 09:37:39,000        BeamDownloader  ETHPeer (eth, 65) <Session <Node(0x6f621a@91.13.207.112)> cbfe5ab6-fcc7-4b73-9e5c-cc20b4f1f018> returned no urgent nodes from (HexBytes('0x423c7cec6a4b0bec1ea9d78e5f0d4551f9ddb8cc23e87aa8c83a9a4cdf4a38da'),)
```

^ Against a known local node, when beam sync isn't lagging, I wouldn't ever expect to get that message.

It means that beam sync is missing some critical trie node and the peer didn't supply the node body as an answer to a `GetNodeData`.

So that's its own problem that needs to be fixed. But another problem is: it should have been retrying. I don't see any evidence in the logs that `BeamDownloader` made any more state requests. I expect it to keep attempting (because maybe it can get the node from a different peer, or maybe the peer is also beam syncing and will have the node available later). But there were no more `GetNodeData` requests after the "returned no urgent nodes" log line, that I can see.

I scanned the code, and didn't immediately see why that was happening. So I would say this should be split into two issues:
[...]

PR #1811, opened 2020-06-23
Bugfix: beam sync hung when only peer is penalized

carver:
### What was wrong?

A call to `QueeningQueue.get_queen_peer()` is meant to hang when there
is no queen. This fixed bug meant that after a queen was penalized, and
then reinserted at the end of a penalty, the get_queen_peer() call would
hang until a second peer was inserted and retrieved by
pop_fastest_peasant(). That means it hangs forever if you are only
connected to a single peer and it gets penalized.

Related to #1806 

### How was it fixed?

Did a fairly significant refactor of how `QueeningQueue` works: peer inserts now go to a helper method rather than directly into its `WaitingPeers` queue. Every insert does a queen-check. Now, any time a queen is set, an `asyncio.Event()` is fired. This allows us to decouple the `get_queen_peer()` logic from the `WaitingPeers` queue. 

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/queen.py
+++ b/trinity/sync/beam/queen.py
@@ -5,8 +5,6 @@ from typing import Any, FrozenSet, Optional, Type
 
 from async_service import Service
 
-from eth_utils import ValidationError
-
 from p2p.abc import CommandAPI
 from p2p.exceptions import (
     PeerConnectionLost,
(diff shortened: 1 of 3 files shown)


trinity/sync/beam/queen.py
```diff
@@ -134,60 +136,61 @@ async def pop_fastest_peasant(self) -> ETHPeer:
...
-            if force_drop_queen or new_peer_quality < old_queen_quality:
-                old_queen, self._queen_peer = self._queen_peer, peer
-                self._waiting_peers.put_nowait(old_queen)
-                return
+                return True
             else:
-                # nothing to do, peer is slower than the queen
-                return
+                # Quality is designed so that an ascending sort puts the best quality at the front
+                #   of a sequence. So a low value of quality is a better quality.
```
cburgdorf, line 178: *wording nitpick*

>So a low value of quality is a better quality

Maybe something like:

>So, the lower the value, the better the quality.

or 

>So, a lower value means better quality

...would sound a little less confusing.

Issue #1665, opened 2020-04-11
Backfilling block bodies during Beam Sync


carver:
### What is wrong?

Currently, during Beam Sync, Trinity cannot answer any questions about transactions, receipts, or uncles that happen in blocks that were not executed. (Actually, currently, Beam Sync never terminates)

### How can it be fixed

Somewhat similar to #1664 - we could start tracking the missing block bodies in a list of ranges by block number. Then add a service that downloads block bodies (probably only when the Beam Sync lag isn't too high).

This could be very similar (or potentially even reuse) the code from the fast syncing approach, here:
https://github.com/ethereum/trinity/blob/1b32d85b1065d6c97de16887a0433b0a3e675fe7/trinity/sync/full/chain.py#L578

The main difference is that you would replace the header syncer passed in here: https://github.com/ethereum/trinity/blob/1b32d85b1065d6c97de16887a0433b0a3e675fe7/trinity/sync/full/chain.py#L133

with a "header syncer" that just loads the headers that already exist out of the database.

There is some crossover that has to be managed with #1664 -- what if the header is missing because of a checkpoint. The simplest thing to do, I suppose, is to pause this backfiller until the header shows up.

cburgdorf:
Ignoring potential holes due to checkpoint syncing for a moment, do we currently have simple markers in our DB that remembers the highest block number for which have downloaded a consecutive chain of `receipts`, `uncles` and `transactions`?

This came up while working on tracking (https://github.com/ethereum/trinity/pull/1682) because `geth` can tell at any point: What is the latest header, block and receipt that we have synced so far. 

Maybe it would make sense to have markers like you proposed in #1664 for uncles, transactions and receipts individually? And maybe it would also make sense to come up with a special unbounded marker. For instance, when we start beam syncing and we haven't downloaded any uncles, receipts or transactions than e.g. `MISSING_RECEIPTS_NUMBERS` would contain something like `[[0, -1]]` which would basically mean we are missing everything from `0` to infinity.

When we then start backfilling from genesis and e.g. stop after `999 we would update it to `[[1000, -1]]`.

Does that make sense? And related, do you have already started working on this and is this an area where you'd think I could help with?

pipermerriam:
I believe that we don't track this and that it should not be part of py-evm so I believe it's something that trinity would need to keep track of.  It could still store it in the main key/value database.

carver:
> Does that make sense? 

Yeah, most of it!

[...]

PR #1819, opened 2020-06-25
Implement backfill of historical blocks during beam sync

cburgdorf:
### What was wrong?

As explained in #1665 we need to backfill historical blocks during beam sync.

### How was it fixed?

1. Refactored `FastChainBodySyncer` to be a little more flexible to be reused for this.
2. Added a new service `BodyChainGapSyncer` that:

- stay in an infinite loop of checking for block gaps (since they can get re-introduced indefinitely during beam sync)
- allows pause and resume

3. Run the service in `BeamSyncer`
4. Monitor the service to pause and resume based on current block lag.

### Known issues

One thing this doesn't do yet is to ensure the backfill does not use the queen peer. Not sure if this is needed for this PR? There are also questions to resolve like, what if we are only connected to a single peer? Would backfill have to wait for a second peer or would we rather say we use the single peer for backfill despite that not being optimal.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1819.feature.rst
+++ b/newsfragments/1819.feature.rst
@@ -1,0 +1,4 @@
+Implement backfill of historicial blocks during beam sync.
+The backfill automatically pauses if beam sync starts lagging behind and resumes
+as soon as it catches up again. It can also be disabled altogether
+via the ``--disable-backfill`` flag.
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -6,7 +6,10 @@ from async_service import Service, background_asyncio_service
 from eth.consensus import ConsensusContext
 from eth.db.atomic import AtomicDB
 from eth.db.schema import SchemaV1
-from eth.exceptions import HeaderNotFound, BlockNotFound
+from eth.exceptions import (
+    BlockNotFound,
+    HeaderNotFound,
+)
 from eth.vm.forks.petersburg import PetersburgVM
 from eth_utils import decode_hex
 from lahja import ConnectionConfig, AsyncioEndpoint
@@ -34,6 +37,7 @@ from trinity.protocol.les.peer import (
 
 from trinity.sync.beam.chain import (
     BeamSyncer,
+    BodyChainGapSyncer,
 )
 from trinity.sync.beam.queen import QueeningQueue
 from trinity.sync.header.chain import (
@@ -84,6 +88,27 @@ def chaindb_with_gaps(chaindb_fresh, chaindb_1000):
         chaindb_fresh.persist_checkpoint_header(header_at, score_at)
 
     assert chaindb_fresh.get_header_chain_gaps() == (((1, 249), (251, 499)), 501)
+    yield chaindb_fresh
+
+
+@pytest.fixture
+def chaindb_with_block_gaps(chaindb_fresh, chaindb_1000):
+    # Make a chain with gaps. This fixture can not be used in a test alongside `chaindb_fresh`
+    # because it alters the `chaindb_fresh` fixture.
+
+    all_headers = [
+        chaindb_1000.get_canonical_block_header_by_number(block_number)
+        for block_number in range(1, 1001)
+    ]
+    chaindb_fresh.persist_header_chain(all_headers)
+
+    fat_chain = LatestTestChain(chaindb_1000.db)
+    for block_number in (250, 500):
+        block = fat_chain.get_canonical_block_by_number(block_number)
+        receipts = block.get_receipts(chaindb_1000)
+        chaindb_fresh.persist_unexecuted_block(block, receipts)
+
+    assert chaindb_fresh.get_chain_gaps() == (((1, 249), (251, 499)), 501)
     yield chaindb_fresh
 
 
@@ -478,6 +503,72 @@ async def test_header_syncer(request,
 
 
 @pytest.mark.asyncio
+async def test_block_gapfill_syncer(request,
+                                    event_loop,
+                                    event_bus,
+                                    chaindb_with_block_gaps,
+                                    chaindb_1000):
+    client_context = ChainContextFactory(headerdb__db=chaindb_with_block_gaps.db)
+    server_context = ChainContextFactory(headerdb__db=chaindb_1000.db)
+    peer_pair = LatestETHPeerPairFactory(
+        alice_peer_context=client_context,
+        bob_peer_context=server_context,
+        event_bus=event_bus,
+    )
+    async with peer_pair as (client_peer, server_peer):
+
+        syncer = BodyChainGapSyncer(
+            LatestTestChain(chaindb_with_block_gaps.db),
+            chaindb_with_block_gaps,
+            MockPeerPoolWithConnectedPeers([client_peer], event_bus=event_bus)
+        )
+        server_peer_pool = MockPeerPoolWithConnectedPeers([server_peer], event_bus=event_bus)
+
+        async with run_peer_pool_event_server(
+            event_bus, server_peer_pool, handler_type=ETHPeerPoolEventServer
+        ), background_asyncio_service(ETHRequestServer(
+            event_bus, TO_NETWORKING_BROADCAST_CONFIG, AsyncChainDB(chaindb_1000.db),
+        )):
+
+            server_peer.logger.info("%s is serving 1000 blocks", server_peer)
+            client_peer.logger.info("%s is syncing up 1000", client_peer)
+
+            async with background_asyncio_service(syncer):
+                chain_with_gaps = LatestTestChain(chaindb_with_block_gaps.db)
+                fat_chain = LatestTestChain(chaindb_1000.db)
+
+                # Sync the first 100 blocks, then check that pausing/resume works
+                await wait_for_block(
+                    chain_with_gaps, fat_chain.get_canonical_block_by_number(100))
+
+                # Pause the syncer and take note how far we have synced at this point
+                syncer.pause()
+                # We need to give async code a moment to settle before we save the progress to
+                # ensure it has stabilized before we save it.
+                await asyncio.sleep(0.01)
+                paused_chain_gaps = chain_with_gaps.chaindb.get_chain_gaps()
+
+                # Consider it victory if after 0.5s no new blocks were written to the database
+                await asyncio.sleep(0.5)
+                assert paused_chain_gaps == chain_with_gaps.chaindb.get_chain_gaps()
+
+                # Resume syncing
+                syncer.resume()
+
+                await wait_for_block(
+                    chain_with_gaps, fat_chain.get_canonical_block_by_number(1000), sync_timeout=20)
+
+                for block_num in range(1, 1001):
+                    assert chain_with_gaps.get_canonical_block_by_number(
+                        block_num) == fat_chain.get_canonical_block_by_number(block_num)
+
+                # We need to give the async calls a moment to settle before we can read the updated
+                # chain gaps.
+                await asyncio.sleep(0.1)
+                assert chain_with_gaps.chaindb.get_chain_gaps() == ((), 1001)
+
+
+@pytest.mark.asyncio
 async def test_header_gapfill_syncer(request,
                                      event_loop,
                                      event_bus,
@@ -728,5 +819,34 @@ async def wait_for_head(headerdb, header, sync_timeout=10):
             header,
             sync_timeout,
             canonical_head,
+        )
+        raise
+
+
+async def wait_for_block(chain, block, sync_timeout=10):
+    """
+    Await until the block the given ``block`` is found in the ``chain``
+    """
+
+    async def wait_loop():
+        synced_block = None
+        while synced_block != block:
+            try:
+                synced_block = chain.get_canonical_block_by_number(block.number)
+            except BlockNotFound:
+                await asyncio.sleep(0.1)
+            else:
+                break
+        assert synced_block == block
+    try:
+        await asyncio.wait_for(wait_loop(), sync_timeout)
+    except asyncio.TimeoutError:
+        gaps, future_tip = chain.chaindb.get_chain_gaps()
+        logging.error(
+            "Could not finish syncing to %s within %ds, gaps %s, future tip %s",
+            block,
+            sync_timeout,
+            gaps,
+            future_tip,
         )
         raise
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -6,7 +6,7 @@ from async_service import Service, background_asyncio_service
 from eth.consensus import ConsensusContext
 from eth.db.atomic import AtomicDB
 from eth.db.schema import SchemaV1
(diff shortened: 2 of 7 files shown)


trinity/sync/beam/chain.py
```diff
@@ -307,12 +310,29 @@ async def _all_verification_bodies_present(
             headers_with_potential_conflicts: Iterable[BlockHeader]) -> bool:
 
         for header in headers_with_potential_conflicts:
-            if not await self._fast_syncer._should_skip_header(header):
+            if not await body_for_header_exists(self._chain_db, self._chain, header):
                 return False
         return True
 
 
-class RigorousFastChainBodySyncer(FastChainBodySyncer):
```
cburgdorf, line 315: The PR isn't ready for review yet but this commit could be branched out and I wanted to get your opinion on it early. I remember you were a proponent for the composition over inheritance narrative which I also share. So, I went ahead and made this change in a preparation to implement the block back fill.

The `RigorousFastChainBodySyncer` inherited from `FastChainBodySyncer` mostly to replace two methods. At least one of these methods is useful for other places including the the block gap fill syncer. It would be unsustainable to build the new `BodyChainGapSyncer` by inheriting from `RigorousFastChainBodySyncer` and so I took the chance to refactor `FastChainBodySyncer` to take two strategy functions in its constructor. With that change, the `RigorousFastChainBodySyncer` doesn't need to inherit from `FastChainBodySyncer` anymore (theoretically, it could be dropped entirely) and the new `BodyChainGapSyncer` can be build in the same fashion, without inheritance.

What's your opinion on this change?

carver: :clap: Yeah, definitely an improved direction!

carver:
Hm, yeah, I see you having some buffering struggles here. I've been tinkering with the buffering, too, related to #1820 (from full-syncing goerli) -- there was definitely some buggy/unintended stuff there. I got some bugs fixed, but it's still as complex and opaque as it was before, so I'd like to take a minute to try to actually improve the model. But not until I'm done with state backfill.

Here are the changes that made for a good sync experience for me (some of them probably aren't necessary, but I haven't spent the time to isolate which ones): #1825

cburgdorf:
>Hm, yeah, I see you having some buffering struggles here

But those are gone now. They came mostly from the fact that I wasn't aware that `new_sync_headers` of `HeaderSyncerAPI` isn't allowed to finish. When I had `new_sync_headers` to simply yield and finish its finite set of headers I was seeing some funky unreliable behavior but I added an `await asyncio.Future()` at the bottom of the method and this fixed the issues that I was having.

So, that's probably unrelated to #1820.

trinity/sync/full/chain.py
```diff
@@ -591,13 +591,29 @@ class FastChainBodySyncer(BaseBodyChainSyncer):
     highest TD, at which point we must run the StateDownloader to fetch the state for our chain
```
cburgdorf, line 591: I wonder if we should find a better name for `FastChainBodySyncer` as `fast` isn't a syncing strategy for Trinity anymore and it may not the best term to convey the intend.

- `UnexecutingBodyChainSyncer`
- `ExecutionlessBodyChainSyncer`
- `NonExecutingBodyChainSyncer`
- ???

carver:
> One thing this doesn't do yet is to ensure the backfill does not use the queen peer.

From a Beam Sync perspective, this is mostly important for `GetNodeData` calls, instead of block bodies. It's worth thinking through whether we could accidentally starve the tip syncer from getting block bodies when beam sync is lagging, we definitely want to avoid that. But so far that hasn't been a significant bottleneck.

> There are also questions to resolve like, what if we are only connected to a single peer? Would backfill have to wait for a second peer or would we rather say we use the single peer for backfill despite that not being optimal.

One way to approach this is to say that backfill of bodies/headers/state should always be "bonus." If we are having trouble keeping up with the tip for any reason, it seems reasonable to pause *all* of these backfills. I don't think we have any reasonable API for exposing/sharing that lag value, it's just printed to log right now. I agree this can be in a follow-up PR, though.

trinity/sync/full/chain.py
carver: Agree that the name is stale. Maybe `ChainSyncer` is enough to imply the full body, so we can drop the `Body` part, and rely on the other being named a `HeaderSyncer` as a disambiguation. Some more brainstorming:

- `MinerTrustingChainSyncer`
- `SealerTrustingChainSyncer`
- `SkipImportChainSyncer`
- `NonImportChainSyncer`

trinity/sync/full/chain.py
```diff
@@ -625,7 +641,7 @@ async def _sync_from(self) -> BlockHeaderAPI:
         return await self.db.coro_get_canonical_head()
 
     async def run(self) -> None:
-        head = await self._sync_from()
```
carver, line 628: So now it would essentially always be a bug to call `_sync_from()` (or `_should_skip_header()`) directly, right?

That means at the very least we should add a docstring in those functions, to not use them. But then it's a pretty pungent code smell. Probably the best solution is to remove the methods from the class altogether, to avoid accidental calls:
```py
@curry
def is_header_in_db(db, header):
  return db.coro_header_exists(header.hash)
```

Then they can be named for what they do, and the "default" setting up in init is a little clearer:
```diff
-            self._should_skip_header_fn = self._should_skip_header
+            self._should_skip_header_fn = is_header_in_db(db)
```

tests/core/p2p-proto/test_sync.py
```diff
@@ -551,6 +552,11 @@ async def test_sequential_header_gapfill_syncer(request,
                     # We check for 499 because 500 is there from the very beginning (the checkpoint)
                     chaindb_with_gaps, chaindb_1000.get_canonical_block_header_by_number(499)
                 )
+                # This test is supposed to only fill in headers, so the following should fail.
+                # If this ever succeeds it probably means the fixture was re-created with trivial
+                # blocks and the test will fail and remind us what kind of fixture we want here.
+                with pytest.raises(BlockNotFound):
+                    chain_with_gaps.get_canonical_block_by_number(499)
```
carver, line 648: Nice defensive check. Did this happen to you, or you just saw it coming? :)

carver:
> If we are having trouble keeping up with the tip for any reason, it seems reasonable to pause _all_ of these backfills. I don't think we have any reasonable API for exposing/sharing that lag value, it's just printed to log right now. I agree this can be in a follow-up PR, though.

lol, clearly I hadn't read the newsfragment or code yet. Obviously that approach sounds good X) :+1:

tests/core/p2p-proto/test_sync.py
```diff
@@ -87,6 +91,25 @@ def chaindb_with_gaps(chaindb_fresh, chaindb_1000):
...
 
 
+@pytest.fixture
+def chaindb_with_block_gaps(chaindb_fresh, chaindb_1000):
+    # Make a chain with gaps. This fixture can not be used in a test alongside `chaindb_fresh`
+    # because it alters the `chaindb_fresh` fixture.
+
+    for block_number in range(1, 1001):
+        header_at = chaindb_1000.get_canonical_block_header_by_number(block_number)
+        chaindb_fresh.persist_header_chain([header_at])
```
carver, line 101: Silly little thing, but I this should run faster as:
```suggestion
    all_headers = [
        chaindb_1000.get_canonical_block_header_by_number(block_number)
        for block_number in range(1, 1001)
    ]
    chaindb_fresh.persist_header_chain(all_headers)
```
(and I think be a tiny bit faster to scan what it's doing)

trinity/sync/beam/chain.py
```diff
@@ -378,6 +414,117 @@ async def run(self) -> None:
...
+    def pause(self) -> None:
+        """
+        Pause the sync. Pause and resume is a wasteful action and should not happen too frequently.
+        """
+        if self._paused:
+            raise RuntimeError("Invalid action. Can not pause service that is already paused.")
+
+        self._paused = True
+        self._body_syncer.get_manager().cancel()
+        self.logger.debug("BodyChainGapSyncer paused")
```
carver, line 495: Debug2? Looks like we get a more meaningful log up in `_monitor_block_backfill` anyway.

trinity/sync/beam/chain.py
```diff
@@ -378,6 +414,117 @@ async def run(self) -> None:
...
+
+    def resume(self) -> None:
+        """
+        Resume the sync.
+        """
+        if not self._paused:
+            raise RuntimeError("Invalid action. Can not resume service that isn't paused.")
+
+        self._paused = False
+        self.logger.debug("BodyChainGapSyncer resumed")
```
carver, line 505: ditto

trinity/sync/beam/chain.py
```diff
@@ -378,6 +414,117 @@ async def run(self) -> None:
...
+
+    def _get_next_gap(self) -> BlockRange:
+        gaps, future_tip_block = self._db.get_chain_gaps()
+        if len(gaps) == 0:
+            # We do not have gaps in the chain of blocks but we may still have a gap from the last
+            # block up until the highest consecutive written header.
+            header_gaps, future_tip_header = self._db.get_header_chain_gaps()
+            if len(header_gaps) > 0:
+                # The header chain has gaps, find out the lowest missing header
+                highest_consecutive_missing_header, _ = header_gaps[0]
```
carver, line 463: Took me a minute to parse this. Maybe:
```suggestion
                lowest_missing_header, _ = header_gaps[0]
```

trinity/sync/beam/constants.py
```diff
@@ -58,3 +58,13 @@
 # We need MAX_UNCLE_DEPTH + 1 headers to check during uncle validation
 # We need to request one more header, to set the starting tip
 FULL_BLOCKS_NEEDED_TO_START_BEAM = MAX_UNCLE_DEPTH + 2
+
+# The number of blocks that beam sync is allowed to lag behind when resuming backfill
+MAX_LAG_TO_RESUME_BACKFILL = 0
```
carver, line 63: This is going to cause a lot of backfill churn. Reorgs of 1 header happen pretty regularly. Maybe allow lagging 1 or 2?

trinity/sync/beam/constants.py
```diff
@@ -58,3 +58,13 @@
...
+
+# The number of blocks that, when lagged behind, will cause backfill to pause
+MAX_LAG_TO_PAUSE_BACKFILL = 5
```
carver, line 66: :+1:

trinity/sync/common/headers.py
```diff
@@ -547,6 +548,46 @@ def emit(self, headers: Iterable[BlockHeader]) -> None:
...
+        for block_number in range(self._from_header, self._to_header + 1):
+            header = await self._db.coro_get_canonical_block_header_by_number(
+                BlockNumber(block_number)
+            )
+            # Since we are fetching these from the db, it is much more resource friendly to not
+            # batch these to the max_batch_size and instead yield them one by one as we load them.
+            yield (header,)
+
+        # This API is not expected to ever finish
+        await asyncio.Future()
```
carver, line 582: Oof, yeah, I get that you're just writing to make it work the current system. Maybe add an issue to clean up the body syncers, so that sync exits more gracefully when the header syncer stops returning headers?

newsfragments/1819.feature.rst
```diff
@@ -0,0 +1,4 @@
+Implement backfill of historicial blocks during beam sync.
+The backfill automatically pauses if beam sync starts lagging behind and resumes
+as soon as it catches up again. It can also be disabled altogether
+via the ``--disable-backfill`` flag.
```
carver, line 4: I can imagine wanting to disable body backfill but not header backfill. Seems like maybe a GFI. Mind writing that up?

trinity/sync/beam/chain.py
```diff
@@ -110,6 +110,9 @@ def __init__(
             force_beam_block_number: BlockNumber = None,
             enable_backfill: bool = True) -> None:
         self.logger = get_logger('trinity.sync.beam.chain.BeamSyncer')
+        self._chain = chain
+        self._chain_db = chain_db
```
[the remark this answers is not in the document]
cburgdorf: Makes sense :+1:

trinity/sync/beam/chain.py
```diff
@@ -307,12 +310,29 @@ async def _all_verification_bodies_present(
...
 
 
-class RigorousFastChainBodySyncer(FastChainBodySyncer):
+@curry
+async def body_for_header_exists(db: BaseAsyncChainDB,
+                                 chain: AsyncChainAPI,
+                                 header: BlockHeader) -> bool:
+    """
+    Return ``True`` if the body for this header exists, otherwise ``False``.
+    """
```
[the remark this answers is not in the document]
cburgdorf: :+1: Added it to `trinity._utils.header`

trinity/sync/full/chain.py
cburgdorf: That makes a lot of sense :+1:

tests/core/p2p-proto/test_sync.py
cburgdorf: It didn't happen to me but it felt like a good thing to check for :)

tests/core/p2p-proto/test_sync.py
cburgdorf: :+1:

trinity/sync/beam/chain.py
cburgdorf: Ah yes :see_no_evil:

trinity/sync/beam/constants.py
cburgdorf: We do allow lagging one or two headers, in fact we allow lagging up to 5 headers at which point the backfill will pause. What this config says is that we need to catch up completely to a lag of `0` before we turn it on again. So, reducing this number will cause us to resume earlier. In other words, causing us to flip *more often* between `pause` and `resume`.

I'm going to keep it at zero for now but are happy to make a change post-merge if we settle on something else.

trinity/sync/common/headers.py
cburgdorf: Sure https://github.com/ethereum/trinity/issues/1831

newsfragments/1819.feature.rst
cburgdorf: Sounds good https://github.com/ethereum/trinity/issues/1832

trinity/sync/full/chain.py
cburgdorf: >Maybe `ChainSyncer` is enough to imply the full body

Except that we already have a `FastChainSyncer` class on top of the `FastChainBodySyncer` class. I'll leave the name as is for this PR.

In general, I like the last two from your list.

trinity/sync/beam/constants.py
carver: Ah, right. I think the `MAX_` prefix doesn't add much clarity, and might even obscure the meaning. Just `LAG_TO_RESUME_BACKFILL` or even better: `RESUME_BACKFILL_AT_LAG`.

Same for `PAUSE_BACKFILL_AT_LAG`.

Issue #1504, opened 2020-01-23
test_peer_pool_connect (tests/core/p2p-proto/test_server.py) fails randomly


gsalgado:
I've started getting random CI failures on that test recently: https://circleci.com/gh/ethereum/trinity/180545?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Will try to investigate tomorrow but thought I'd open an issue in case somebody is already on it

gsalgado:
I've been debugging some of those random test failures we see on CI, and one thing I've found useful sometimes is to use the `pytest-repeat` plugin to run a given test multiple times, also with `-n 8` (I guess anything higher than number of cores) to overload my machine in the hope of introducing the delays that are triggering those failures on CI. with that I was able to reproduce this locally. Just thought I'd share for when you guys need to reproduce any of those

carver:
Happened again in https://github.com/ethereum/trinity/pull/1821

```
____________________________ test_peer_pool_connect ____________________________
[gw0] linux -- Python 3.7.7 /home/circleci/repo/.tox/py37-eth1-core/bin/python

monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f97f6ccb810>
event_loop = <uvloop.Loop running=False closed=False debug=False>
server = <test_server.ParagonServer object at 0x7f97f6b9ed90>
receiver_remote = <Node(0xca634cae0d49acb401d8a4c6b6fe8c55b70d115bf400769cc1400f3258cd31387574077f301b421bc84df7266c44e9e6d569fc56be00812904767bf5ccd1fc7f@127.0.0.1:59361)>

    @pytest.mark.asyncio
    async def test_peer_pool_connect(monkeypatch, event_loop, server, receiver_remote):
        started_peers = []
    
        async def mock_start_peer(peer):
            nonlocal started_peers
            started_peers.append(peer)
    
        monkeypatch.setattr(server.peer_pool, 'start_peer', mock_start_peer)
    
        initiator_peer_pool = ParagonPeerPool(
            privkey=INITIATOR_PRIVKEY,
            context=ParagonContext(),
        )
        nodes = [receiver_remote]
        async with background_asyncio_service(initiator_peer_pool) as manager:
            await manager.wait_started()
            await initiator_peer_pool.connect_to_nodes(nodes)
            # Give the receiver_server a chance to ack the handshake.
            await asyncio.sleep(0.2)
    
>           assert len(started_peers) == 1
[...]

PR #1823, opened 2020-06-29
Fix intermittent failures in test_peer_pool_connect

gsalgado:
That test used to sleep for 0.2s to wait for the handshake to complete, but
that was not always enough on CI. Now we use an event with a longer
timeout.

Also fixes P2P handshake tests that were causing asyncio warnings
as they were calling negotiate_protocol_handshakes() directly but were
not ensuring the background task started by that was cancelled/awaited at
the end.

Closes: #1504

### DIFF
--- a/tests/core/p2p-proto/test_server.py
+++ b/tests/core/p2p-proto/test_server.py
@@ -90,11 +90,7 @@ async def server(event_bus, receiver_remote):
 
 
 @pytest.mark.asyncio
-async def test_server_incoming_connection(monkeypatch,
-                                          server,
-                                          event_loop,
-                                          receiver_remote,
-                                          ):
+async def test_server_incoming_connection(server, receiver_remote):
     use_eip8 = False
     initiator = HandshakeInitiator(receiver_remote, INITIATOR_PRIVKEY, use_eip8)
     initiator_remote = NodeFactory(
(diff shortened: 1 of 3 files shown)


tests/core/p2p-proto/test_server.py
```diff
@@ -163,14 +159,22 @@ async def test_server_incoming_connection(monkeypatch,
...
-    started_peers = []
+async def test_peer_pool_connect(monkeypatch, server, receiver_remote):
+    peer_started = asyncio.Event()
+
+    start_peer = server.peer_pool.start_peer
 
     async def mock_start_peer(peer):
-        nonlocal started_peers
-        started_peers.append(peer)
+        await start_peer(peer)
```
carver, line 176: Adding this line dosen't have to do with the bug, right?

I guess the rationale to add it is something like: when we mock the method, we want to keep doing whatever the mocked method was doing, to make sure it's not raising any exceptions. (Although we're not actively checking any result that occurs because of `start_peer()` being invoked). Anything I'm missing? If so, maybe add a comment explaining.

gsalgado: It is actually part of the fix, yeah -- it ensures the test does the same thing we do in production, as that way we don't have to manually stop the background task. I'll add a comment

gsalgado: The background streaming task that is started during the handshake is what I mean. It's taken care of by the Connection, but here we were mocking `start_peer()`, which is where the Connection gets created/started

Issue #1822, opened 2020-06-26
Flaky hypothesis test


cburgdorf:
### What is wrong?

https://circleci.com/gh/ethereum/trinity/247074?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

### How can it be fixed

Haven't looked into it but there's a hint that maybe it is just the `deadline` that needs to be turned off.

carver:
Annoying that it seems to be a very fast test usually. Yeah, the deadline is the think I fight with the most on hypothesis. I suppose we aren't really meant to be using hypothesis as a benchmarking tool. I wonder what the logic for converting from warning to error is. The CI boxes are surely shared and are likely to have really high variability.

I guess :+1: to drop the deadline.

PR #1824, opened 2020-06-29
Remove the ShutdownRequest feature from ComponentManager

gsalgado:
The CM used to spawn components and only terminate if a component
explicitly sent it a ShutdownRequest, but now it terminates as soon
as any component terminates, so the ShutdownRequest mechanism became
unnecessary.

### DIFF
--- a/p2p/logic.py
+++ b/p2p/logic.py
@@ -80,6 +80,10 @@ async def wait_first(futures: Sequence[asyncio.Future[None]]) -> None:
 
     If the task running us is cancelled, all futures will be cancelled.
     """
+    for future in futures:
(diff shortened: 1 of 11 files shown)


cburgdorf:
CI fail is the flaky test that was reported in https://github.com/ethereum/trinity/issues/1822

gsalgado:
@cburgdorf There was an actual bug introduced in https://github.com/ethereum/trinity/commit/9d9e5c6233f0e3fdb85aed2f0e5f2decddafda18, but it would only cause a test failure some times. https://github.com/ethereum/trinity/pull/1824/commits/d2e6b70d96d79a68e4b40963e8565999ff380f9b fixes it

PR #1825, opened 2020-06-29
Improve block buffering

### LINKED ISSUE #1820: Header syncer seems to race ahead
### What is wrong?

The debug logs seem to suggest that the header syncer is racing way ahead of the body downloads, at least when doing a full sync. The intended behavior is for the header syncer to only go a few thousand ahead of the tip, during sync.

### How can it be fixed

Just requires some investigation. Probably something to do with the capacity check being in the wrong place, or not getting triggered/cleared correctly.

carver:
### What was wrong?

Fixes #1820 

### How was it fixed?

The key change was being able to use `await` on `OrderedTaskPreparation.register_tasks()` (or really, on a new method called `wait_add_tasks()`). Now, the header and block body sync will pause when there are plenty of blocks in the queue to import, and resume when there are too few.

When `wait_add_tasks()` hangs, it holds onto the header generator which prevents the more headers from being generated (after the header buffer fills up).

### To-Do

- [x] Clean up commit history
- [x] Why does "Ignore delay when inserting a queen-able peasant" commit break beam sync? (move to new issue)

#### Cute Animal Picture

### DIFF
--- a/trinity/_utils/datastructures.py
+++ b/trinity/_utils/datastructures.py
@@ -17,10 +17,10 @@ from itertools import (
 from typing import (
     Any,
     Callable,
+    Collection,
     Dict,
     Generic,
     Iterable,
-    Sequence,
     Set,
     Tuple,
     Type,
@@ -242,7 +242,7 @@ class TaskQueue(Generic[TTask]):
 
         return (next_id, pending_tasks)
 
-    def complete(self, batch_id: int, completed: Sequence[TTask]) -> None:
+    def complete(self, batch_id: int, completed: Collection[TTask]) -> None:
         if batch_id not in self._in_progress:
             raise ValidationError(f"batch id {batch_id} not recognized, with tasks {completed!r}")
 
@@ -339,7 +339,7 @@ class DuplicateTasks(Exception, Generic[TTask]):
     """
     Tried to register a task that was already registered
     """
-    def __init__(self, msg: str, duplicates: Sequence[TTask]) -> None:
+    def __init__(self, msg: str, duplicates: Collection[TTask]) -> None:
         super().__init__(msg)
         self.duplicates = duplicates
 
@@ -363,14 +363,14 @@ class BaseOrderedTaskPreparation(ABC, Generic[TTask, TTaskID]):
     @abstractmethod
     def register_tasks(
             self,
-            tasks: Sequence[TTask],
+            tasks: Collection[TTask],
             ignore_duplicates: bool = False) -> Tuple[TTask, ...]:
         ...
 
     @abstractmethod
     async def wait_add_tasks(
             self,
-            tasks: Sequence[TTask],
+            tasks: Collection[TTask],
             ignore_duplicates: bool = False) -> Tuple[TTask, ...]:
         ...
 
@@ -555,7 +555,7 @@ class OrderedTaskPreparation(
     @to_tuple
     def register_tasks(
             self,
-            tasks: Sequence[TTask],
+            tasks: Collection[TTask],
             ignore_duplicates: bool = False) -> Iterable[TTask]:
         """
         Initiate a task into tracking. By default, each task must be registered
@@ -616,7 +616,7 @@ class OrderedTaskPreparation(
 
     async def wait_add_tasks(
             self,
-            tasks: Sequence[TTask],
+            tasks: Collection[TTask],
             ignore_duplicates: bool = False) -> Tuple[TTask, ...]:
         """Like :meth:`register_tasks` but pauses if there are too many tasks in the pipeline"""
         if self._max_tasks is not None:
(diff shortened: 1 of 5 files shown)


trinity/sync/full/chain.py
```diff
@@ -1151,8 +1160,14 @@ async def _import_ready_blocks(self) -> None:
 
             block = await self._import_queue.get()
             if not self._import_active.locked():
-                self.logger.info(
-                    "Waited %.1fs for %s body",
+                lag = time.time() - block.header.timestamp
+                if lag > 30:  # ~2 blocks
```
carver, line 1164: Only notify us if we are starved for blocks if we are not near the tip. (Of course we will be starved if the blocks haven't been created yet)

p2p/peer_pool.py
```diff
@@ -404,7 +404,7 @@ async def connect_to_nodes(self, nodes: Sequence[NodeAPI]) -> None:
             if not batch:
                 return
 
-            self.logger.info(
+            self.logger.debug(
```
carver, line 407: Now that we pick up peers pretty quickly, we don't have to be so noisy about this.

trinity/_utils/datastructures.py
```diff
@@ -621,6 +626,13 @@ def finish_prereq(self, prereq: TPrerequisite, tasks: Sequence[TTask]) -> None:
             if task_completion.is_complete and self._is_ready(task):
                 self._mark_complete(task_id)
 
+    async def finish_to_capacity(self, prereq: TPrerequisite, tasks: Sequence[TTask]) -> None:
```
carver, line 629: Document and test

trinity/_utils/datastructures.py
```diff
@@ -720,3 +733,13 @@ def _prune(self, prune_task_id: TTaskID) -> None:
         del self._tasks[prune_task_id]
         if prune_task_id in self._declared_finished:
             self._declared_finished.remove(prune_task_id)
+
+    def num_ready(self) -> int:
```
carver, line 737: Document and test

trinity/_utils/datastructures.py
```diff
@@ -720,3 +733,13 @@ def _prune(self, prune_task_id: TTaskID) -> None:
...
+        return self._ready_tasks.qsize()
+
+    def num_unready(self) -> int:
```
carver, line 740: Document and test

trinity/_utils/datastructures.py
```diff
@@ -720,3 +733,13 @@ def _prune(self, prune_task_id: TTaskID) -> None:
...
+        return len(self._unready)
+
+    def num_tasks(self) -> int:
```
carver, line 743: Document and test

trinity/sync/full/chain.py
```diff
@@ -147,7 +147,7 @@ def __init__(self,
         # Track incomplete block body download tasks
         # - arbitrarily allow several requests-worth of headers queued up
         # - try to get bodies from lower block numbers first
-        buffer_size = MAX_BODIES_FETCH * REQUEST_BUFFER_MULTIPLIER
+        buffer_size = MAX_BODIES_FETCH * REQUEST_BUFFER_MULTIPLIER // 10
```
carver, line 150: Just reduce the multiplier I guess?

trinity/sync/full/chain.py
```diff
@@ -792,15 +792,16 @@ async def _assign_receipt_download_to_peers(self) -> None:
...
             peer.manager.run_task(self._run_receipt_download_batch, peer, batch_id, headers)
 
-    def _mark_body_download_complete(
+    async def _mark_body_download_complete(
             self,
             batch_id: int,
             completed_headers: Sequence[BlockHeaderAPI]) -> None:
-        super()._mark_body_download_complete(batch_id, completed_headers)
-        self._block_persist_tracker.finish_prereq(
+        await self._block_persist_tracker.finish_to_capacity(
```
carver, line 799: comment: if OTP `_block_persist_tracker` is "full", then pause here, to avoid racing ahead in downloading block bodies.

carver:
Is the write capacity flag redundant now that each connected pipe enforces its own limit?

carver:
> Is the write capacity flag redundant now that each connected pipe enforces its own limit?

Yes, code looks okay to go without it. I think it's a lot cleaner to have a series of buffered pipes rather than a switch trying to signal fullness back to an earlier part of the pipeline, so it feels like a win.

trinity/_utils/datastructures.py
```diff
@@ -621,6 +626,14 @@ def finish_prereq(self, prereq: TPrerequisite, tasks: Sequence[TTask]) -> None:
             if task_completion.is_complete and self._is_ready(task):
                 self._mark_complete(task_id)
 
+    async def finish_wait(self, prereq: TPrerequisite, tasks: Sequence[TTask]) -> None:
+        """Like :meth:`finish_prereq` but pauses if there are too many ready tasks"""
+        if self._max_ready is not None:
+            while self.num_ready() > 0 and self.num_tasks() > self._max_ready:
```
cburgdorf, line 632: This confuses me. It waits on `self.num_tasks() > self._max_ready` but the docstring says this would wait if there are too many *ready*  tasks. `num_tasks` includes both `ready` and `unready`  tasks though.

carver: Ah, yeah, it's actually the docstring (and name of `_max_ready`) that's wrong. We want to notice that the whole pipeline is full ASAP, without waiting for each of the tasks in it to complete.

But as I'm writing this, I think the await should be on a `register_tasks` level instead of a `finish_wait`. Rewrite coming...

PR #1827, opened 2020-07-01
New integration test syncing from another local trinity instance

gsalgado:
One instance is started with a pre-populated datadir containing some
block headers, and the other connects to it and is expected to
complete a skeleton sync and switch to beam sync.

Closes: #1302


tests/integration/test_trinity_cli.py
```diff
@@ -52,6 +59,41 @@ def amend_command_for_unused_port(command, unused_tcp_port):
     return command
 
 
+@pytest.fixture
+def prepopulated_datadir():
+    fixture_datadir = pathlib.Path(__file__).parent / 'fixtures' / 'trinity_headerchain_datadir'
```
carver, line 64: A zipped datadir would be better, I think, using something similar to this to expose a path to the unzipped data-dir:
https://github.com/ethereum/trinity/blob/8926230973b8ef64b6c78ca9d597aa7fdbe20b0b/tests/core/integration_test_helpers.py#L71-L83

Main benefits I see:
- fewer total bytes added to repo
- follows an existing pattern
- harder/impossible to accidentally use the repo's data-dir directly

tests/integration/test_trinity_cli.py
```diff
@@ -52,6 +59,41 @@ def amend_command_for_unused_port(command, unused_tcp_port):
     return command
 
 
+@pytest.fixture
+def prepopulated_datadir():
```
carver, line 63: What exactly is this pre-populated with? Okay, it's mainnet, but which headers are included? Was there a checkpoint? Were any blocks executed via beam sync?

tests/integration/test_trinity_cli.py
```diff
@@ -52,6 +59,41 @@ def amend_command_for_unused_port(command, unused_tcp_port):
...
+    fixture_datadir = pathlib.Path(__file__).parent / 'fixtures' / 'trinity_headerchain_datadir'
+    with tempfile.TemporaryDirectory() as temp_dir:
+        datadir = pathlib.Path(temp_dir) / 'trinity'
+        shutil.copytree(fixture_datadir, datadir)
+        yield datadir
+
+
+# XXX: Shouldn't we use --trinity-tmp-root-dir on all tests here so that we can run them without
+# interfering with an existing data dir?
```
carver, line 72: Yeah, probably. I can't think of a reason not to, except it's less code, and we can only run one at a time anyway because of VM limitations.

I don't think this is happening, but it wouldn't *totally* shock me if some of the tests only work because an earlier run "warmed up" the data dir. That would obviously be a problem to fix. So :+1: for moving everything to `--trinity-tmp-root-dir` (or maybe all but one? we want to make sure that it works when that flag isn't supplied, too)

tests/integration/test_trinity_cli.py
```diff
@@ -52,6 +59,41 @@ def amend_command_for_unused_port(command, unused_tcp_port):
...
+        command2 = amend_command_for_unused_port(
+            ('trinity', '--trinity-tmp-root-dir', '--disable-discovery',
+             '--preferred-node', serving_enode),
+            port2)
+        async with AsyncProcessRunner.run(command2, timeout_sec=120) as runner2:
+            assert await contains_all(runner2.stderr, {
+                "Adding ETHPeer",
+                "Imported 192 headers",
+                "Caught up to skeleton peer",
+            })
```
carver, line 94: Maybe we can also confirm that runner1 claims to be connected to 1 peer. It's a pretty weak assertion, but probably better than nothing.

gsalgado: Indeed

tests/integration/test_trinity_cli.py
gsalgado: Added a comment:
```
# A mainnet datadir with the first 580 headers, but no block bodies.
```

tests/integration/test_trinity_cli.py
gsalgado: Yeah, I can't think of any reason not to do that

tests/integration/test_trinity_cli.py
cburgdorf: The tests **do** run under temporarily root dirs just not set via CLI flag but read from an environment variable that is set for every test.

https://github.com/ethereum/trinity/blob/8926230973b8ef64b6c78ca9d597aa7fdbe20b0b/tests/conftest.py#L85-L94

Issue #1775, opened 2020-06-08
NetworkDBComponent continues running as long as any of its background services is running


gsalgado:
https://github.com/ethereum/trinity/blob/066e9252d80ee18f38bc22ded990f5d7cd769d7a/trinity/components/builtin/network_db/component.py#L243-L251

As it will certainly not operate normally if one of the services stops, it should probably exit as soon as any of its background services exits

Issue #1773, opened 2020-06-07
MetricsComponent runs background services and never checks on them


gsalgado:
That component is also affected by the [fire-and-forget anti pattern](https://github.com/ethereum/snake-charmers-tactical-manual/pull/70):

https://github.com/ethereum/trinity/blob/066e9252d80ee18f38bc22ded990f5d7cd769d7a/trinity/components/builtin/metrics/component.py#L166-L171

That component runs until the first started service's `wait_finished()` returns (which can be due to a service crashing). Instead it should probably exit as soon as any of the started services` `wait_finished()` returns

PR #1828, opened 2020-07-01
Terminate components when any of their background services do so

gsalgado:
MetricsComponent, JsonRpcServerComponent and NetworkDBComponent would
continue running until all of their background services terminated,
which could lead to malfunction and other crashes. Now they will
exit as soon as any of their services terminate

Closes: #1773
Closes: #1774
Closes: #1775

### DIFF
--- a/trinity/components/builtin/metrics/component.py
+++ b/trinity/components/builtin/metrics/component.py
@@ -4,10 +4,8 @@ from argparse import (
     Namespace,
     _SubParsersAction,
 )
-import contextlib
 from typing import Type
 
-from async_service import background_trio_service
 from eth_utils import ValidationError
 
 from lahja import EndpointAPI
@@ -23,6 +21,7 @@ from trinity.components.builtin.metrics.system_metrics_collector import collect_
 from trinity.extensibility import (
     TrioIsolatedComponent,
 )
+from trinity._utils.services import run_background_trio_services
 
 
 def metrics_service_from_args(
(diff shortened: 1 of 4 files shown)


trinity/components/builtin/metrics/component.py
```diff
@@ -168,4 +171,8 @@ async def do_run(self, event_bus: EndpointAPI) -> None:
                 await stack.enter_async_context(background_trio_service(service))
                 for service in services_to_exit
             ])
-            await managers[0].wait_finished()
+            # If any of our background services terminate, we do so as well.
+            await wait_first([
+                asyncio.create_task(manager.wait_finished())
+                for manager in managers
+            ])
```
cburgdorf, line 178: Given how often we use this pattern, maybe this would justfify new utility functions such as `run_background_trio_services` so that all of this:

```python
services_to_exit = (
   metrics_service,
   system_metrics_collector,
   blockchain_metrics_collector,
)

async with contextlib.AsyncExitStack() as stack:
   managers = tuple([
         await stack.enter_async_context(background_trio_service(service))
         for service in services_to_exit
   ])
   # If any of our background services terminate, we do so as well.
   await wait_first([
         asyncio.create_task(manager.wait_finished())
         for manager in managers
   ])
``` 

could simply become this:

```python
await background_trio_services(
   metrics_service,
   system_metrics_collector,
   blockchain_metrics_collector,
)
```

gsalgado: Well, so far we seem to have only two places using it for asyncio services and one for trio, but I like your suggestion and it does reduce some duplication so here we go: https://github.com/ethereum/trinity/pull/1828/commits/da4799283a18a60116aa683a4ee731758ac2a547

PR #1829, opened 2020-07-01
Add test to prove gap sync recovers from uncle attack

cburgdorf:
### What was wrong?

Our current tests to do not fully assert that we the header gap sync recovers from an uncle attack.

### How was it fixed?

The test already asserted that we *acknowledge* the attack so all this does is to turn our attacker peer into a friendly peer and continue waiting for the sync to succeed.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -576,25 +576,36 @@ async def test_header_gap_fill_detects_invalid_attempt(caplog,
             MockPeerPoolWithConnectedPeers([client_peer], event_bus=event_bus)
         )
         server_peer_pool = MockPeerPoolWithConnectedPeers([server_peer], event_bus=event_bus)
-
+        uncle_chaindb = AsyncChainDB(chaindb_uncle.db)
         async with run_peer_pool_event_server(
             event_bus, server_peer_pool, handler_type=ETHPeerPoolEventServer
         ), background_asyncio_service(ETHRequestServer(
-            event_bus, TO_NETWORKING_BROADCAST_CONFIG, AsyncChainDB(chaindb_uncle.db),
+            event_bus, TO_NETWORKING_BROADCAST_CONFIG, uncle_chaindb,
         )):
 
             server_peer.logger.info("%s is serving 1000 blocks", server_peer)
             client_peer.logger.info("%s is syncing up 1000", client_peer)
 
+            # We check for 499 because 500 exists from the very beginning (the checkpoint)
+            expected_block_number = 499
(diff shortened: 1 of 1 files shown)


tests/core/p2p-proto/test_sync.py
```diff
@@ -576,25 +576,37 @@ async def test_header_gap_fill_detects_invalid_attempt(caplog,
...
                     )
                 except asyncio.TimeoutError:
                     assert "Attempted to fill gap with invalid header" in caplog.text
+                    # Monkey patch the uncle chaindb to effectively make the attacker peer
+                    # switch to the correct chain.
+                    uncle_chaindb.db = chaindb_1000.db
+                    await wait_for_head(
+                        chaindb_with_gaps,
+                        chaindb_1000.get_canonical_block_header_by_number(expected_block_number),
+                        sync_timeout=5,
```
carver, line 606: FWIW, I often reduce the timeouts in local testing significantly to speed up the bug/fix/check cycle. But I have seen issues with CI being occasionally slow, and those flaky failures are really aggravating, so I usually leave it with a nice wide margin. This one especially shouldn't hurt to have a wide margin on the "happy path". The other one is more muddy, because we *have* to wait the full `sync_timeout` in order for the test to succeed.

tests/core/p2p-proto/test_sync.py
```diff
@@ -576,25 +576,37 @@ async def test_header_gap_fill_detects_invalid_attempt(caplog,
...
+            expected_block_number = 499
             async with background_asyncio_service(client):
                 try:
                     await wait_for_head(
-                        # We check for 499 because 500 exists from the
-                        # very beginning (the checkpoint)
-                        chaindb_with_gaps, chaindb_1000.get_canonical_block_header_by_number(499)
+                        chaindb_with_gaps,
+                        chaindb_1000.get_canonical_block_header_by_number(expected_block_number),
+                        sync_timeout=5,
```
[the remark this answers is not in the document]
cburgdorf: Yeah, this is based on the assumption that this is a pure header sync and other tests that are based on 1000 headers make it through with a timeout of `10` seconds (the default). So we might get a way with 5 seconds to sync to the uncle which sits roughly in the middle and then another 5 seconds to sync the second half.

Let's try it :)

tests/core/p2p-proto/test_sync.py
cburgdorf: Ah yes, you're right, this one will shortcut anyway as soon as we reach the target header. I roll that one back but keep the other one for now :)

PR #1830, opened 2020-07-01
Bugfix beam sync first 8 blocks

carver:
### What was wrong?

Beam sync was broken if you try to beam to block number 2 through 7.

### How was it fixed?

Get the genesis header instead of the genesis block, for syncing. Very strange that mypy did not catch this. Seems like just the kind of thing it's supposed to get. :/

Also did a big refactor for an upcoming change in #1769 

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -1,4 +1,5 @@
 import asyncio
+from contextlib import asynccontextmanager
 import logging
 import uuid
 
@@ -161,7 +162,7 @@ async def test_beam_syncer_with_checkpoint_too_close_to_tip(
 
     caplog.set_level(logging.INFO)
     try:
-        await test_beam_syncer(
+        await test_beam_syncer_loads_recent_state_root(
             request,
             event_loop,
             event_bus,
@@ -189,7 +190,7 @@ async def test_beam_syncer_with_checkpoint(
         score=55,
     )
 
-    await test_beam_syncer(
+    await test_beam_syncer_loads_recent_state_root(
         request,
         event_loop,
         event_bus,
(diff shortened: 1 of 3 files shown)


tests/core/p2p-proto/test_sync.py
```diff
@@ -255,17 +260,57 @@ async def test_beam_syncer(
...
+# -  2: Normally we need to look back ~6 headers to look for duplicate uncles. This
+#       tests the lowest block number code path where an alternate code path is triggered,
+#       to look up fewer than usual.
+# -  7: Normally we need to look back ~6 headers to look for duplicate uncles. This
+#       tests the highest block number code path where an alternate code path is triggered,
+#       to look up fewer than usual.
+# This test was reduced to a few cases for speed. To run the full suite, use
+# range(1, 130) for beam_to_block. (and optionally follow the instructions at target_head)
+@pytest.mark.asyncio
+@pytest.mark.parametrize('beam_to_block', [1, 2, 7, 66, 68, 129])
```
carver, line 282: Adding `2, 7` here is essentially the only meaningful change in the whole test file. All the rest is a refactor.

cburgdorf:
>Get the genesis header instead of the genesis block, for syncing. Very strange that mypy did not catch this. Seems like just the kind of thing it's supposed to get. :/

Yes, and it would catch it in `strict` mode. The bitter truth is that a big chunk of Py-EVM / Trinity is still operating in a mode where lots of APIs accept or return `Any`.

That said, even though I knew that, this particular case surprised me as well so I investigated it a little. It turns out that `rlp.Serializable` is typed as `Any` and so everything subclassed from that is too.

In strict mode, mypy would not even allow `BlockHeaderAPI` and the like to exist the way they currently are

>eth/abc.py:85: error: Class cannot subclass 'Serializable' (has type 'Any')

I'm going to investigate a little further if there's a quick win we can get to tighten our typing since this is a very severe weakness.

cburgdorf:
Ok, looks like there's an easy fix for that. We can just expose type hints from pyrlp (PEP561). We don't actually have to add type hints, just expose them so that mypy can infer whatever it wants to infer. I did a quick test and it yields a ton of errors even just in Py-EVM. But a lot of the errors seem like legit bugs! E.g. exceptions that access properties in their message string which do not exist. Some of these in places that are probably rarely hit.

It looks like it's worth doing that and fix all these issues. I'll take care of this.

cburgdorf:
Here's a work in progress PR to address the mypy issue https://github.com/ethereum/py-evm/pull/1948

Issue #1767, opened 2020-06-02
PeerConnectionLost escaping


cburgdorf:
### What is wrong?

This [blasted out of our CI](https://circleci.com/gh/ethereum/trinity/234050) run.

```
DEBUG  06-02 12:30:58  async_process_runner.py  b'\x1b[1m\x1b[31m   ERROR  2020-06-02 12:30:58,526           ETHPeerPool  unexpected error during peer connection\n'
   DEBUG  06-02 12:30:58  async_process_runner.py  b'Traceback (most recent call last):\n'
   DEBUG  06-02 12:30:58  async_process_runner.py  b'  File "/home/circleci/repo/trinity/protocol/common/peer.py", line 239, in maybe_connect_more_peers\n'
   DEBUG  06-02 12:30:58  async_process_runner.py  b'    candidate_counts = await asyncio.gather(*(\n'
   DEBUG  06-02 12:30:58  async_process_runner.py  b'  File "/home/circleci/repo/p2p/peer_pool.py", line 220, in _add_peers_from_backend\n'
   DEBUG  06-02 12:30:58  async_process_runner.py  b'    await self.connect_to_nodes(candidates)\n'
   DEBUG  06-02 12:30:58  async_process_runner.py  b'  File "/home/circleci/repo/p2p/peer_pool.py", line 411, in connect_to_nodes\n'
   DEBUG  06-02 12:30:58  async_process_runner.py  b'    await asyncio.gather(*(self.connect_to_node(node) for node in batch))\n'
   DEBUG  06-02 12:30:58  async_process_runner.py  b'  File "/home/circleci/repo/p2p/peer_pool.py", line 467, in connect_to_node\n'
   DEBUG  06-02 12:30:58  async_process_runner.py  b'    await self.start_peer(peer)\n'
   DEBUG  06-02 12:30:58  async_process_runner.py  b'  File "/home/circleci/repo/p2p/peer_pool.py", line 281, in start_peer\n'
   DEBUG  06-02 12:30:58  async_process_runner.py  b'    self._add_peer(peer, ())\n'
   DEBUG  06-02 12:30:58  async_process_runner.py  b'  File "/home/circleci/repo/p2p/peer_pool.py", line 319, in _add_peer\n'
   DEBUG  06-02 12:30:58  async_process_runner.py  b'    subscriber.register_peer(peer)\n'
[...]

PR #1840, opened 2020-07-06
Several improvements to Peer/Connection APIs 

gsalgado:
These will hopefully make it less likely for us to introduce new code that leaks PeerConnectionLost exceptions

- Peer.p2p_api is now private

- properties that can raise PeerConnectionLost are now methods

- the methods above will now raise a PeerConnectionLost if the peer
  is not running or if the conn is closing (before it'd only happen when
  the conn was closing)

### DIFF
--- a/p2p/connection.py
+++ b/p2p/connection.py
@@ -277,6 +277,8 @@ class Connection(ConnectionAPI, Service):
 
     def has_logic(self, name: str) -> bool:
         if self.is_closing:
+            # This is a safety net, really, as the Peer should never call this if it is no longer
+            # alive.
             raise PeerConnectionLost("Cannot look up subprotocol when connection is closing")
         return name in self._logics
 
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -115,7 +115,7 @@ class BasePeer(Service):
     _event_bus: EndpointAPI = None
 
     base_protocol: BaseP2PProtocol
-    p2p_api: P2PAPI
+    _p2p_api: P2PAPI
 
     def __init__(self,
                  connection: ConnectionAPI,
@@ -252,11 +252,19 @@ class BasePeer(Service):
         pass
 
     async def _handle_disconnect(self, connection: ConnectionAPI, cmd: Disconnect) -> None:
-        self.p2p_api.remote_disconnect_reason = cmd.payload
+        self._p2p_api.remote_disconnect_reason = cmd.payload
         # We run as a daemon child of the connection, so cancel the connection instead of
         # ourselves to ensure asyncio-service doesn't think we're exiting when the connection is
         # still active, as that would cause a DaemonTaskExit.
         self.connection.get_manager().cancel()
+
+    @property
+    def is_alive(self) -> bool:
+        # We need this because when a remote disconnects from us the connection may be closed
+        # before the Disconnect msg is processed and cancels ourselves.
+        if not hasattr(self, 'manager'):
+            return False
+        return self.manager.is_running and not self.connection.is_closing
 
     async def run(self) -> None:
         self._start_time = time.monotonic()
@@ -265,7 +273,7 @@ class BasePeer(Service):
             async with contextlib.AsyncExitStack() as stack:
                 fut = await stack.enter_async_context(P2PAPI().as_behavior().apply(self.connection))
                 futures = [fut]
-                self.p2p_api = self.connection.get_logic('p2p', P2PAPI)
+                self._p2p_api = self.connection.get_logic('p2p', P2PAPI)
 
                 for behavior in self.get_behaviors():
                     if behavior.should_apply_to(self.connection):
@@ -295,8 +303,8 @@ class BasePeer(Service):
         finally:
             for callback in self._finished_callbacks:
                 callback(self)
-            if (self.p2p_api.local_disconnect_reason is None and
-                    self.p2p_api.remote_disconnect_reason is None):
+            if (self.local_disconnect_reason is None and
+                    self.remote_disconnect_reason is None):
                 self._send_disconnect(DisconnectReason.CLIENT_QUITTING)
             # We run as a child service of the connection, but we don't want to leave a connection
             # open if somebody cancels just us, so this ensures the connection gets closed as well.
@@ -331,9 +339,21 @@ class BasePeer(Service):
 
     def _send_disconnect(self, reason: DisconnectReason) -> None:
         try:
-            self.p2p_api.disconnect(reason)
+            self._p2p_api.disconnect(reason)
         except PeerConnectionLost:
             self.logger.debug("Tried to disconnect from %s, but already disconnected", self)
+
+    @property
+    def safe_client_version_string(self) -> str:
+        return self._p2p_api.safe_client_version_string
+
+    @property
+    def local_disconnect_reason(self) -> DisconnectReason:
+        return self._p2p_api.local_disconnect_reason
+
+    @property
+    def remote_disconnect_reason(self) -> DisconnectReason:
+        return self._p2p_api.remote_disconnect_reason
 
 
 class PeerMessage(NamedTuple):
--- a/trinity/protocol/eth/peer.py
+++ b/trinity/protocol/eth/peer.py
@@ -18,6 +18,7 @@ from lahja import (
 )
 
 from p2p.abc import BehaviorAPI, CommandAPI, HandshakerAPI, SessionAPI
+from p2p.exceptions import PeerConnectionLost
 
 from trinity.protocol.common.peer import (
     BaseChainPeer,
@@ -95,8 +96,18 @@ class ETHPeer(BaseChainPeer):
             ETHV65API().as_behavior()
         )
 
+    def get_eth_api(self) -> AnyETHAPI:
+        """
+        Return this peer's ETH API.
+
+        Raises PeerConnectionLost if the peer is no longer alive.
+        """
+        if not self.is_alive:
+            raise PeerConnectionLost(f"{self} is no longer alive")
+        return self._eth_api
+
     @cached_property
-    def eth_api(self) -> AnyETHAPI:
+    def _eth_api(self) -> AnyETHAPI:
         if self.connection.has_protocol(ETHProtocolV63):
             return self.connection.get_logic(ETHV63API.name, ETHV63API)
         if self.connection.has_protocol(ETHProtocolV64):
@@ -107,8 +118,13 @@ class ETHPeer(BaseChainPeer):
             raise Exception("Unreachable code")
 
     def get_extra_stats(self) -> Tuple[str, ...]:
+        """
+        Return extra stats for this peer.
+
+        Raises PeerConnectionLost if the peer is no longer alive.
+        """
         basic_stats = super().get_extra_stats()
-        eth_stats = self.eth_api.get_extra_stats()
+        eth_stats = self.get_eth_api().get_extra_stats()
         return basic_stats + eth_stats
 
 
@@ -231,7 +247,7 @@ class ETHPeerPoolEventServer(PeerPoolEventServer[ETHPeer]):
             self,
             event: GetBlockHeadersRequest) -> Tuple[BlockHeaderAPI, ...]:
         peer = self.get_peer(event.session)
-        return await peer.eth_api.get_block_headers(
+        return await peer.get_eth_api().get_block_headers(
             event.block_number_or_hash,
             event.max_headers,
             skip=event.skip,
@@ -245,7 +261,7 @@ class ETHPeerPoolEventServer(PeerPoolEventServer[ETHPeer]):
         return await self.with_node_and_timeout(
             event.session,
             event.timeout,
-            lambda peer: peer.eth_api.get_receipts(event.headers)
+            lambda peer: peer.get_eth_api().get_receipts(event.headers)
         )
 
     async def handle_get_block_bodies_request(self,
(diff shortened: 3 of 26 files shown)


gsalgado:
@carver I believe this would address the issue you were trying to fix in #1673

p2p/connection.py
```diff
@@ -276,8 +276,6 @@ def remove_logic(self, name: str) -> None:
         self._logics.pop(name)
 
     def has_logic(self, name: str) -> bool:
-        if self.is_closing:
-            raise PeerConnectionLost("Cannot look up subprotocol when connection is closing")
```
gsalgado, line 280: Why I'm removing this: https://github.com/ethereum/trinity/issues/1767#issuecomment-654004827

trinity/sync/beam/queen.py
```diff
@@ -153,20 +146,23 @@ def penalize_queen(self, peer: ETHPeer, delay: float = NON_IDEAL_RESPONSE_PENALT
...
+
         if self._queen_peer is None:
             return True
         elif peer == self._queen_peer:
             return True
         else:
             try:
                 new_peer_quality = _peer_sort_key(peer)
-            except (UnknownAPI, PeerConnectionLost) as exc:
+            except UnknownAPI as exc:
```
carver, line 159: If anything, I'd like to see a move toward dropping `UnknownAPI` from this, instead of `PeerConnectionLost`. It makes sense to me that the peer connection could die at any time, but not that an `ETHPeer` could stop having the `eth_api` attribute.

trinity/sync/beam/queen.py
```diff
@@ -153,20 +153,23 @@ def penalize_queen(self, peer: ETHPeer, delay: float = NON_IDEAL_RESPONSE_PENALT
...
                 new_peer_quality = _peer_sort_key(peer)
-            except (UnknownAPI, PeerConnectionLost) as exc:
+            except PeerConnectionLost as exc:
                 self.logger.debug("Invalid %s, because we can't get speed stats: %r", peer, exc)
                 return False
 
             try:
                 current_queen_quality = _peer_sort_key(self._queen_peer)
-            except (UnknownAPI, PeerConnectionLost) as exc:
+            except PeerConnectionLost as exc:
```
gsalgado, line 172: We no longer need to catch `UnknownAPI` here because `get_eth_api()` will now raise a PeerConnectionLost when the peer is no longer alive, so we'll never end up calling `connection.get_logic()` when the behavior has been removed

trinity/protocol/eth/peer.py
```diff
@@ -95,8 +96,18 @@ def get_behaviors(self) -> Tuple[BehaviorAPI, ...]:
...
         )
 
+    def get_eth_api(self) -> AnyETHAPI:
+        """
+        Return this peer's ETH API.
+
+        Raises PeerConnectionLost if the peer is no longer alive.
+        """
+        if not self.is_alive:
+            raise PeerConnectionLost(f"{self} is no longer alive")
```
gsalgado, line 106: This is a more appropriate place than `Connection.has_logic()` to raise this exception as here we can do it as soon as the peer has been cancelled (which doesn't necessarily mean the connection is closing as we keep it open until the peer has finished). Also by turning it into a method I hope to make it less likely for us to overlook the fact that it can raise a PeerConnectionLost

p2p/peer.py
```diff
@@ -292,11 +300,16 @@ async def run(self) -> None:
                 except BaseException:
                     self.logger.exception("Behavior finished before us, cancelling ourselves")
                     self.manager.cancel()
+                finally:
+                    # We are in an async context so call our finished callbacks before we leave it
+                    # as that may yield control and let other code attempt to use this peer, which
+                    # in turn leads to crashes if they forget to check that the peer is still
+                    # running before attempting to use it.
+                    for callback in self._finished_callbacks:
+                        callback(self)
```
carver, line 309: What about exceptions anywhere in lines 274-294? They won't trigger the finished callbacks anymore.

p2p/peer.py
```diff
@@ -331,10 +344,22 @@ def disconnect_nowait(self, reason: DisconnectReason) -> None:
...
+    def safe_client_version_string(self) -> str:
+        return self._p2p_api.safe_client_version_string
+
+    @property
+    def local_disconnect_reason(self) -> DisconnectReason:
+        return self._p2p_api.local_disconnect_reason
+
+    @property
+    def remote_disconnect_reason(self) -> DisconnectReason:
+        return self._p2p_api.remote_disconnect_reason
```
carver, line 361: Why hide this `p2p_api`?

trinity/protocol/eth/peer.py
carver: What if we don't remove the "logic" from the peer on disconnect?

So `peer.eth_api` would always resolve after the handshake is complete, and it would always return True for `has_logic()` even after disconnect. The `PeerConnectionLost` wouldn't raise until something like `peer.eth_api.get_block_headers()`. Then, failures would happen at the more expected location: when trying to directly interact with the peer.

gsalgado: Some of these logics/behaviors will run `Service`s when applied: https://github.com/ethereum/trinity/blob/f8c0a5bc25a41461f9f1f30e534d87ff3b78b9af/p2p/exchange/logic.py#L13-L36

So we need to stop them when the peer/connection terminates, and can't leave them accessible via `get_logic()` after that as they're no longer running.

p2p/peer.py
gsalgado: Indeed.  I couldn't find a nice way to refactor this code in order to trigger the callbacks in both cases, but it turns out the only place where we use finished callbacks is the `PeerPool`, and my intent here was to try and prevent code from using a `Peer` while it was in the process of being terminated, but I'm pretty sure most of the code that can do that will be keeping references to peers so even if we triggered the callbacks and removed them from the pool, those references would still be there and will still crash if they are used before first checking that the peer is alive. IOW, I think this change doesn't buy us anything so I'll revert it

trinity/protocol/eth/peer.py
carver: > we ... can't leave them accessible via `get_logic()` after that as they're no longer running.

What goes wrong if we leave it accessible after they aren't running?

It seems to me like the right thing we will happen. For example, if we leave `peer.eth_api.get_node_data` accessible, then a PCL gets raised right away if you try to call it:

https://github.com/ethereum/trinity/blob/410cbeb5ccee65129f464a4a134ad20aec985fea/p2p/exchange/manager.py#L51-L54

This is something that all call sites for exchanges need to manage anyway (because the peer could be lost at any point while waiting for a response). So the calling code doesn't really care if the exception happens to be raised immediately.

gsalgado: > What goes wrong if we leave the logic accessible after it isn't running?

Well, if they're `Service`s we could get `LifecycleError`s and other async-service exceptions, but probably others as well since the exposed APIs are likely to expect the service to be running and may crash in unexpected ways if it isn't.

> It seems to me like the right thing we will happen. For example, if we leave peer.eth_api.get_node_data accessible, then a PCL gets raised right away if you try to call it:

In this case, yeah, but I don't think we can assume that for all of them. Note that those abstractions are not very well documented so I'm basically reverse engineering them and trying to figure out why they were designed like this. It seems like a reasonable thing, though, to have the logics only made available while they're running, and stopping them as soon as the connection is cancelled/disconnected

carver: > It seems like a reasonable thing, though, to have the logics only made available while they're running, and stopping them as soon as the connection is cancelled/disconnected

Yeah, definitely makes sense for them to stop. I'm just not sure it makes sense for it to give a PCL on the API.

I guess I don't have a clear picture what else might happen in the sub APIs of the logics either, besides just raising a PCL.  So I guess I'm not opposed to the approach for now.

trinity/sync/beam/state.py
```diff
@@ -489,7 +497,10 @@ async def _make_node_request(
         num_nodes = len(node_hashes)
         self.logger.debug2("Requesting %d nodes from %s", num_nodes, peer)
         try:
-            return await peer.eth_api.get_node_data(node_hashes, timeout=self._reply_timeout)
+            return await peer.get_eth_api().get_node_data(node_hashes, timeout=self._reply_timeout)
+        except PeerConnectionLost:
+            self.logger.debug("Peer %s went away during node request", peer)
+            return tuple()
```
carver, line 503: I think the split of `_make_node_request()` and `_request_nodes()` made more sense when we were more aggressively dropping the peers. Now it seems like the `TimeoutError` can just be one of the handled exceptions above.

If you don't feel like doing that change, then at least this one can be dropped, since it's handled in the layer above. (This change would add some new dead/unreachable code, in the above PCL exception handling).

trinity/sync/beam/queen.py
```diff
@@ -114,17 +114,17 @@ async def pop_fastest_peasant(self) -> ETHPeer:
                 continue
 
             try:
-                peer_is_requesting_nodes = peer.eth_api.get_node_data.is_requesting
+                peer_is_requesting = peer.get_eth_api().get_node_data.is_requesting
```
carver, line 117: Just realized this is a nasty abstraction violation. If QueenTracker was used for any different kind of request, it would still be doing this check on `get_node_data`. (Not related to this PR of course)

trinity/sync/beam/state.py
```diff
@@ -373,6 +379,8 @@ async def _get_nodes_from_peer(
             urgent_node_hashes: Tuple[Hash32, ...],
             predictive_node_hashes: Tuple[Hash32, ...],
             predictive_batch_id: int) -> None:
+        if not peer.is_alive:
+            raise ValueError(f"{peer} is no longer alive")
```
carver, line 383: Why check in this particular method? The very next statement is an await anyway, so we would have to check again in _request_nodes. (if we didn't just handle the PCL)

trinity/sync/common/headers.py
```diff
@@ -1057,15 +1064,18 @@ async def _full_skeleton_sync(self, skeleton_syncer: SkeletonSyncer[TChainPeer])
     async def _validate_peer_is_ahead(self, peer: BaseChainPeer) -> None:
         head = await self._db.coro_get_canonical_head()
         head_td = await self._db.coro_get_score(head.hash)
-        if peer.head_info.head_td <= head_td:
+        # Since we yielded above, we could've lost the peer connection.
+        if not peer.is_alive:
+            raise _PeerBehind(f"{peer} is no longer alive, not a valid target for sync")
+        if peer.get_head_info().head_td <= head_td:
```
carver, line 1070: *nitpick* `elif` just for visual clarity that these are alternate branches (obviously doesn't matter logically).

trinity/sync/common/headers.py
```diff
@@ -832,10 +836,13 @@ async def _request_headers(
             raise
 
     async def _init_sync_progress(self, parent_header: BlockHeaderAPI, peer: TChainPeer) -> None:
+        if not peer.is_alive:
+            raise ValueError(f"{peer} is no longer alive")
```
carver, line 840: Are the checks like these meant to stay in the final PR, or were you just using them to try to catch some cases?

Seems like it will pepper us with uncaught `ValueError`s. That seems strictly worse than the checks not being there in the first place.

Here, for example, it looks like PCL should be just be handled below, with a bare `return`.

gsalgado: My intent was to convey the fact that this method requires an alive peer, forcing its callers to check that before calling it. But maybe that doesn't really buys us anything as a PCL is handled here anyway. I'll revert

trinity/sync/beam/state.py
gsalgado: Here too I was trying to push to the callers the responsibility of checking that the peer was alive. Will revert

trinity/sync/beam/state.py
gsalgado: Merged the two methods as you suggested

trinity/sync/beam/queen.py
gsalgado: I guess the same applies to `_peer_sort_key()`, then? Added FIXME comments to both

carver: Oh, I had an older version of `QueeningQueue` that tried to make the sorting key abstract, but I must have simplified it away at some point. So yeah, at least it's consistent: the class is not *claiming* to be generic. So I think it's not a big concern, until we decide to use it for a different metric.

trinity/sync/beam/state.py
```diff
@@ -470,6 +471,12 @@ async def _request_nodes(
             self.logger.debug("Pending nodes call to %r future cancelled", peer)
             self._queen_tracker.penalize_queen(peer)
             raise
+        except asyncio.TimeoutError:
+            # This kind of exception shouldn't necessarily *drop* the peer,
+            # so capture error, log and swallow
+            self.logger.debug("Timed out requesting %d nodes from %s", num_nodes, peer)
+            self._total_timeouts += 1
+            return tuple()
```
carver, line 479: Ah, I think we should still `self._queen_tracker.penalize_queen(peer)` here.

trinity/sync/common/headers.py
carver: Looks like this one didn't get reverted.

carver:
(Note `trinity/sync/common/headers.py` line 840 that didn't get dropped) -- I don't like how github folds comments high in the thread.

trinity/sync/common/headers.py
gsalgado: Oops, indeed

gsalgado:
FWIW, I ended up reverting this because the connection tracker gets called when a peer is shutting down and it uses the `chain_info` property (https://github.com/ethereum/trinity/blob/01bcf3a9bceb14f0dfd50c9dd610e0c2f35409e1/trinity/components/builtin/network_db/eth1_peer_db/tracker.py#L161), which here became inaccessible during the peer shutdown

In #1884 I use a different approach that doesn't have that limitation

PR #1841, opened 2020-07-06
Fix a number of typing issues and exposed bugs

cburgdorf:
### What was wrong?

A number of wrong type hints + two actual bugs being exposed by upcoming changes in Py-EVM: https://github.com/ethereum/py-evm/pull/1948

### How was it fixed?

- mostly corrected type hints
- two places where different APIs needed to be called.
- one of them turned out to be in a dead `PeerHeaderSyncer` class -> deleted

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/chains/light.py
+++ b/trinity/chains/light.py
@@ -165,7 +165,7 @@ class LightDispatchChain(AsyncChainAPI, Chain):
         Raises HeaderNotFound if it is not found.
         """
         header = self._headerdb.get_canonical_block_header_by_number(block_number)
-        return await self.get_block_by_header(header)
+        return await self.coro_get_block_by_header(header)
 
     def get_canonical_block_hash(self, block_number: BlockNumber) -> Hash32:
         return self._headerdb.get_canonical_block_hash(block_number)
--- a/trinity/chains/light.py
+++ b/trinity/chains/light.py
@@ -33,6 +33,7 @@ from eth.abc import (
     ReceiptAPI,
     SignedTransactionAPI,
     UnsignedTransactionAPI,
+    BlockImportResult,
 )
 from eth.chains.base import (
     AccountState,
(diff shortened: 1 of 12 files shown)


trinity/sync/common/chain.py
```diff
@@ -123,7 +124,7 @@ async def next_header_batch(self) -> AsyncIterator[Tuple[BlockHeaderAPI, ...]]:
                 if last_received_header is None:
                     # Skip over existing headers on the first run-through
                     completed_headers, new_headers = await skip_complete_headers(
-                        all_headers, self.db.coro_header_exists)
+                        all_headers, is_header_in_db(self.db))
                     if len(new_headers) == 0 and len(completed_headers) > 0:
```
cburgdorf, line 128: `coro_header_exists` expects a *hash*  but this is being fed an actual header so I expect the skip here doesn't actually currently work. It's nice that this was now caught by type hints.

trinity/chains/light.py
```diff
@@ -165,7 +165,7 @@ async def coro_get_canonical_block_by_number(self, block_number: BlockNumber) ->
         Raises HeaderNotFound if it is not found.
         """
         header = self._headerdb.get_canonical_block_header_by_number(block_number)
-        return await self.get_block_by_header(header)
```
cburgdorf, line 168: This was now caught as well (*await* called on a block)

cburgdorf:
@carver I also validated that this would have caught the bug which caused me to go down the rabbit whole: https://github.com/ethereum/trinity/commit/1543f9ff92e7a2d69379687ee4f0f37920e2caf3

All these fixes get triggered by these changes in Py-EVM: https://github.com/ethereum/py-evm/pull/1948

So, this is independent of a bigger py-rlp rewrite which I would like to avoid for now.

trinity/sync/common/chain.py
cburgdorf: Oh it gets even better because it turns out that triggered the deletion of a big chunk of dead code. Probably a leftover from the `fast` sync days.

trinity/sync/common/chain.py
```diff
@@ -29,221 +17,12 @@
...
-)
 
 from eth2.beacon.chains.base import (
     BaseBeaconChain
 )
 
-from .types import SyncProgress
-
-
-class PeerHeaderSyncer(Service):
```
carver, line 58: Yeah, this was basically the syncer before the skeleton syncer. I think the original idea was that after a trinity run finishes the skeleton sync, that it could switch over to this simpler approach. But handling that transition has been on the back-burner for a while, and I'm not certain that the wins are big enough. I imagine some of the reason it was still in the repo is that it was painful to delete such hard-earned code (this had to go through a lot of revisions before it was more battle-worthy). But :+1: to remove since I don't see us starting to use it anytime soon.

PR #1848, opened 2020-07-06
Pin `anyio` version to avoid conflict in recent update

ralexstokes:
### What was wrong?

`anyio` updated w/in the last 24 hours to `v1.4.0` which requires the dependency `idna` at `v2.8`.

We have (several?) other dependencies which require `idna@v2.7`.

### How was it fixed?

Pin `anyio` to the previous version to allow the acceptable version of `idna`.

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -118,6 +118,7 @@ deps = {
         PYEVM_DEPENDENCY,
         "ssz==0.2.4",
         "asks>=2.3.6,<3",  # validator client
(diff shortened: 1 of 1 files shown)


ralexstokes:
It seems the way the dockerfile tests are written, this will fail until we merge to `master`.

This seems to resolve the problem locally so I'll go ahead and merge.

Happy to address any post-merge feedback.

PR #1850, opened 2020-07-07
Fix a crasher + flaky test with BodyChainGapSyncer

### LINKED ISSUE #1839: tests/core/p2p-proto/test_sync.py::test_block_gapfill_syncer fails intermittently
https://app.circleci.com/pipelines/github/ethereum/trinity/6649/workflows/800afaa4-3e06-4772-8e73-84aa70d5504d/jobs/249829/steps

```
                    # We need to give the async calls a moment to settle before we can read the updated
                    # chain gaps.
                    await asyncio.sleep(0.1)
>                   assert chain_with_gaps.chaindb.get_chain_gaps() == ((), 1001)
E                   assert ((), 998) == ((), 1001)
E                     At index 1 diff: 998 != 1001
 ```

Probably just need to wait longer before reading the updated chain gaps?

### PR DESCRIPTION
**Pull request by cburgdorf**
### What was wrong?

As reported in #1838 there's a crash when `pause()` is called on a `BodyChainGapSyncer` that is still asynchronously setting itself up.

This also fixes #1839 (flaky test) by increasing two timeouts.

### How was it fixed?

1. Do not assume that the service is already setup in `pause()`
2. Cover the scenario in the existing test
3. Make the idle time of the service configurable and drastically reduce it in the test
4. Increase two timeouts that were too sensitive

### To-Do

- [x] Clean up commit history

- unreleased bug (no changelog entry)

#### Cute Animal Picture

### DIFF
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -649,7 +649,7 @@ async def test_block_gapfill_syncer(request,
                 syncer.pause()
                 # We need to give async code a moment to settle before we save the progress to
                 # ensure it has stabilized before we save it.
-                await asyncio.sleep(0.01)
+                await asyncio.sleep(0.25)
                 paused_chain_gaps = chain_with_gaps.chaindb.get_chain_gaps()
 
                 # Consider it victory if after 0.5s no new blocks were written to the database
@@ -668,7 +668,7 @@ async def test_block_gapfill_syncer(request,
 
                 # We need to give the async calls a moment to settle before we can read the updated
                 # chain gaps.
-                await asyncio.sleep(0.1)
+                await asyncio.sleep(0.25)
                 assert chain_with_gaps.chaindb.get_chain_gaps() == ((), 1001)
 
 
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -618,8 +618,10 @@ async def test_block_gapfill_syncer(request,
         syncer = BodyChainGapSyncer(
             LatestTestChain(chaindb_with_block_gaps.db),
             chaindb_with_block_gaps,
-            MockPeerPoolWithConnectedPeers([client_peer], event_bus=event_bus)
+            MockPeerPoolWithConnectedPeers([client_peer], event_bus=event_bus),
         )
+        # In production, this would be the block time but we want our test to pause/resume swiftly
+        syncer._idle_time = 0.01
         server_peer_pool = MockPeerPoolWithConnectedPeers([server_peer], event_bus=event_bus)
 
         async with run_peer_pool_event_server(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: trinity/sync/beam/chain.py [hunk 1]
```diff
@@ -410,18 +410,26 @@ class BodyChainGapSyncer(Service):
     def __init__(self,
                  chain: AsyncChainAPI,
                  db: BaseAsyncChainDB,
-                 peer_pool: ETHPeerPool) -> None:
+                 peer_pool: ETHPeerPool,
+                 idle_time: int = BLOCK_BACKFILL_IDLE_TIME) -> None:
```
pipermerriam, line 414: What do you think about making this a property of the class? and then accessing it from `self.BLOCK_BACKFILL_IDLE_TIMEOUT` so you can just monkeypatch it in the tests?   Seems *slightly* better than expanding it into the `__init__` signature...

Improvement seems *minor* but maybe cleaner.

trinity/sync/beam/chain.py
```diff
@@ -410,18 +410,26 @@ class BodyChainGapSyncer(Service):
...
         self._chain = chain
         self._db = db
+        self._idle_time = idle_time
         self._peer_pool = peer_pool
         self._paused = False
+        self._body_syncer: FastChainBodySyncer = None
```
pipermerriam, line 420: Shouldn't this be causing a typing error with `mypy`?

cburgdorf: we don't use strict optional in Trinity yet.

trinity/sync/beam/chain.py
cburgdorf: Works for me :)

tests/core/p2p-proto/test_sync.py
```diff
@@ -590,7 +595,7 @@ async def test_block_gapfill_syncer(request,
                 syncer.pause()
                 # We need to give async code a moment to settle before we save the progress to
                 # ensure it has stabilized before we save it.
-                await asyncio.sleep(0.01)
+                await asyncio.sleep(0.5)
```
gsalgado, line 598: That's a rather long unconditional sleep for a test. Isn't there a way we can wait for an event? Otherwise I'd prefer if we called `sleep(0.1)` in a loop, checking if `paused_chain_gaps` is already what we expect on every iteration

tests/core/p2p-proto/test_sync.py
```diff
@@ -609,7 +614,7 @@ async def test_block_gapfill_syncer(request,
 
                 # We need to give the async calls a moment to settle before we can read the updated
                 # chain gaps.
-                await asyncio.sleep(0.1)
+                await asyncio.sleep(0.5)
```
gsalgado, line 617: Same here

tests/core/p2p-proto/test_sync.py
gsalgado: Something like https://github.com/ethereum/trinity/pull/1853/commits/c55e00da63c659470c6746a47ec37119b8149175

gsalgado: Oh, hmm, now that I read the whole test I see this wouldn't work. ISTM this is a very fragile test by its very nature but I can't think of an easy way to improve it :-(

cburgdorf: We probably don't need it to be that long (I haven't actually seen it fail apart from what was reported in #1839). I reduced it to `0.25` for both places now.

PR #1857, opened 2020-07-08
Dropping Noise upgrade in favor of Secio.

g-r-a-n-t:
#### Cute Animal Picture

### DIFF
--- a/trinity/nodes/beacon/host.py
+++ b/trinity/nodes/beacon/host.py
@@ -8,6 +8,7 @@ from libp2p.network.swarm import Swarm
 from libp2p.peer.id import ID as PeerID
 from libp2p.peer.peerinfo import PeerInfo
 from libp2p.peer.peerstore import PeerStore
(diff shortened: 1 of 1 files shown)


trinity/nodes/beacon/host.py
```diff
@@ -56,10 +55,9 @@ def __init__(
         peer_store.add_key_pair(peer_id, key_pair)
 
         muxer_transports_by_protocol = {MPLEX_PROTOCOL_ID: Mplex}
-        noise_key = ed25519.create_new_key_pair()
         security_transports_by_protocol = {
-            TProtocol(noise.PROTOCOL_ID): noise.Transport(
-                key_pair, noise_key.private_key
+            TProtocol(secio.ID): secio.Transport(
```
ralexstokes, line 59: it may be worth just having both here with priority given to `secio`

will likely want to change it either way in the future so not urgent to change this PR further right now

Issue #1856, opened 2020-07-08
Testnet Sync


g-r-a-n-t:
It's currently not possible to import blocks from a multiclient testnet, such as [altona](https://github.com/goerli/altona/tree/master/altona), for a number of reasons. [This branch](https://github.com/ethereum/trinity/compare/master...g-r-a-n-t:testnet-sync) contains the fixes required to connect to and sync altona, but are hacky and need to be cleaned up.

The following things need to be fixed in separate PRs:
- do not ignore multiaddrs passed in under the `bootstrap-nodes` parameter https://github.com/ethereum/trinity/pull/1869
- revert back to secio as the default libp2p transport upgrade https://github.com/ethereum/trinity/pull/1857
- parameterize the chain config (altona, mainnet, minimal) https://github.com/ethereum/trinity/pull/1900
- add a parameter for a genesis state ssz file https://github.com/ethereum/trinity/pull/1864
- handle block_by_range requests that exceed max size https://github.com/ethereum/trinity/pull/1881

ralexstokes:
some thoughts/suggestions:

> * revert back to secio as the default libp2p transport upgrade #1857

we could just have both transports with priority to `secio` (until `noise` is sorted).

> * add a parameter for a genesis state ssz file

i'd suggest rolling this into the `trinity-beacon-trio network-generator` command, with an option to specify the genesis state.

operations will be simpler for us in the future if we can just pass off the one config file for a given testnet...

> * handle block_by_range requests that exceed max size

seems like the lightest weight thing to do right now is just add a method to `trinity.nodes.beacon.full.SyncRequest` that yields a iterable of blocks by range requests (of the appropriate size). down the line, this functionality could be managed by a syncer component that e.g. gets blocks in parallel and more sophisticated things...

PR #1864, opened 2020-07-09
--genesis-state param added to create-network subcommand

g-r-a-n-t:
### What was wrong?

We need to be able to set the initial state of the beacon chain.

### How was it fixed?

Added a parameter to the `create-network` subcommand that allows you to set the genesis state of the chain you're configuring.

#### Cute Animal Picture

### DIFF
--- a/eth2/genesis.py
+++ b/eth2/genesis.py
@@ -1,14 +1,12 @@
 from pathlib import Path
-from typing import Any, Dict
+from typing import Any, Dict, Iterable
 
 from eth.constants import ZERO_HASH32
 from eth_utils import encode_hex
 from ssz.tools.dump import to_formatted_dict
-from ssz.tools.parse import from_formatted_dict
 from typing_extensions import Literal
 
 from eth2.beacon.genesis import initialize_beacon_state_from_eth1
-from eth2.beacon.state_machines.forks.altona.configs import ALTONA_CONFIG
 from eth2.beacon.state_machines.forks.serenity.configs import SERENITY_CONFIG
 from eth2.beacon.state_machines.forks.skeleton_lake.configs import (
     MINIMAL_SERENITY_CONFIG,
@@ -25,26 +23,8 @@ from eth2.beacon.typing import Timestamp
 from eth2.configs import Eth2Config
 
 
-def update_genesis_config_with_time(
-    config: Dict[str, Any], genesis_time: Timestamp
-) -> Dict[str, Any]:
-    config_profile = config["profile"]
-    eth2_config = _get_eth2_config(config_profile)
-    override_lengths(eth2_config)
-
-    existing_state = from_formatted_dict(config["genesis_state"], BeaconState)
-
-    genesis_state = existing_state.set("genesis_time", genesis_time)
-
-    updates = {
-        "genesis_state_root": encode_hex(genesis_state.hash_tree_root),
-        "genesis_state": to_formatted_dict(genesis_state),
-    }
-    return {**config, **updates}
-
-
-def generate_genesis_config(
-    config_profile: Literal["minimal", "mainnet", "altona"], genesis_time: Timestamp
+def genesis_config_with_validators(
+    config_profile: Literal["minimal", "mainnet"], genesis_time: Timestamp = None
 ) -> Dict[str, Any]:
     eth2_config = _get_eth2_config(config_profile)
     override_lengths(eth2_config)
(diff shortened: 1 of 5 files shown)


trinity/components/eth2/network_generator/component.py
```diff
@@ -38,40 +50,35 @@ def configure_parser(
...
+        genesis_group.add_argument(
             "--genesis-delay",
             type=int,
             help="Delay in seconds to use as the genesis time from time of execution",
         )
+        genesis_group.add_argument(
+            "--genesis-state",
+            type=pathlib.Path,
+            help="Path to an SSZ file that contains the initial chain state",
+        )
```
ralexstokes, line 74: there could be cases where we want to get the `genesis_time` from the state while still configuring a custom `genesis_delay`. not that we immediately have a use case but using the `mutex` group option for this keeps us from doing it...

ralexstokes: i'd think we want to start w/ the state, if there is a `genesis_time` then that overwrites the state and the `genesis_delay` works like it did (using `now` as an implicit time) (this is good for one-off genesis events like `trinity-beacon-trio create-network ... && trinity-beacon-trio` at the shell)

eth2/genesis.py
```diff
@@ -78,5 +80,27 @@ def generate_genesis_config(
...
+    config_profile: Literal["minimal", "mainnet", "altona"], genesis_state_path: Path
+) -> Dict[str, Any]:
+    eth2_config = _get_eth2_config(config_profile)
+    override_lengths(eth2_config)
+
+    with open(genesis_state_path, "rb") as genesis_state_file:
+        genesis_state = BeaconState.deserialize(genesis_state_file.read())
+
+    return {
+        "profile": config_profile,
```
ralexstokes, line 93: we may find that this notion of `profile` is not that helpful but i'll just note that defining specific network configs (e.g. `altona`) mixes the spirit of what a `profile` is... this is subtle and may not stand the battle-testing of running testnets!

really we want to take all the data at say: https://github.com/goerli/altona/tree/master/altona and then distill that down into a single file that a trinity node knows how to read. i'm fine w/ defining an `altona` profile here but we should be thinking about how it scales w/ future testnets, etc. i'd prefer to not keep ephemeral ideas hanging around in the code indefinitely...

trinity/components/eth2/network_generator/component.py
```diff
@@ -97,4 +125,4 @@ def _generate_network_as_json(
         )
         output_file_path.parent.mkdir(parents=True, exist_ok=True)
         with open(output_file_path, "w") as output_file:
-            json.dump(genesis_config, output_file)
+            json.dump(genesis_config, output_file, cls=BytesJsonEncoder)
```
ralexstokes, line 128: when are we dumping `bytes` that haven't already been encoded?

trinity/components/eth2/network_generator/component.py
```diff
@@ -81,12 +88,33 @@ def _generate_network_as_json(
...
-                genesis_config = update_genesis_config_with_time(
-                    existing_genesis_config, genesis_time
-                )
-        else:
+                try:
+                    existing_genesis_config = json.load(config_file)
+                except JSONDecodeError as e:
+                    existing_genesis_config = None
+                    cls.logger.warning("failed to load existing config as json: %s", e)
+
```
ralexstokes, line 96: the "check for existing" was to skip an expensive key generation / validator creation if we could....

trinity/components/eth2/network_generator/component.py
```diff
@@ -81,12 +88,33 @@ def _generate_network_as_json(
...
+        if (args.genesis_time or args.genesis_delay) and existing_genesis_config:
+            genesis_config = update_genesis_config_with_time(
+                existing_genesis_config, genesis_time
+            )
+        elif args.genesis_time or args.genesis_delay:
             genesis_config = generate_genesis_config(args.config_profile, genesis_time)
+        else:
+            genesis_config = genesis_config_from_state(
+                args.config_profile, args.genesis_state
+            )
```
ralexstokes, line 115: i think the decisions are:

1. (state, no time): write the state to the config.
2. (state, time): overwrite the time in the state and write this new state to the config.
3. (no state, time): make a new state (from the genesis time) and write that.

trinity/components/eth2/network_generator/component.py
ralexstokes: with existing testnets, we would just skip this step (esp. while our validator is pre-alpha...)

trinity/components/eth2/network_generator/component.py
g-r-a-n-t: So I went with this, but kept the validator key pairs generation via an extra flag, now we have `[--generate-validator-keys | --genesis-state GENESIS_STATE]` and the decisions look roughly like:

1. (state, no time): write the state to the config.
2. (state, time): overwrite the time in the state and write this new state to the config.
3. (no state, time): **use the default** state (from the genesis time) and write that.
4. (generate validators, time) create a genesis state from the self-generated validator keys and update the time
5. (generate validators, no time) same as above, but the default time is used

There's also the case where no values are provided and everything is defaulted. Not sure if we want to go so far as to require a time parameter if a state parameter is not provided.

trinity/components/eth2/network_generator/component.py
g-r-a-n-t: I just removed this, if that ok

trinity/components/eth2/network_generator/component.py
g-r-a-n-t: The `GENESIS_FORK_VERSION` and `BLS_WITHDRAWAL_PREFIX` config values

g-r-a-n-t: right here https://github.com/ethereum/trinity/blob/73968ef497496b8beaba77f5801424cbd9f244f2/eth2/genesis.py#L105 with `GENESIS_FORK_VERSION` and `BLS_WITHDRAWAL_PREFIX`

eth2/genesis.py
g-r-a-n-t: I'll just ignore the altona config stuff in this PR, I have a separte note mentioning this here https://github.com/ethereum/trinity/issues/1856 

But yeah, I agree that altona shouldn't be its own profile.

trinity/components/eth2/network_generator/component.py
ralexstokes: so i'm wondering why this is not working?

https://github.com/ethereum/trinity/blob/73968ef497496b8beaba77f5801424cbd9f244f2/eth2/configs.py#L94-L95

trinity/components/eth2/network_generator/component.py
ralexstokes: we may need it for any local testing we do but perhaps it is fine to drop for now

trinity/components/eth2/network_generator/component.py
g-r-a-n-t: hmm.. it could be that `Version is not bytes`

I may be wrong about `BLS_WITHDRAWAL_PREFIX` not getting encoded, it may just be `GENESIS_FORK_VERSION` that prompted this change

g-r-a-n-t:
```python
>>> from typing import NewType
>>> Version = NewType("Version", bytes)
>>> Version is bytes
False
```

PR #1865, opened 2020-07-10
Use INFO as the default level for async-service's logger

gsalgado:
async-service's DEBUG logs completely drowns everything else as it
accounts for more than 95% of all our DEBUG logs, so by default set it
to INFO.

Also change the number of backup log files we keep from 10 to 20

### DIFF
--- a/trinity/_utils/logging.py
+++ b/trinity/_utils/logging.py
@@ -30,7 +30,7 @@ from trinity._utils.socket import BufferedSocket, IPCSocketServer
 from trinity._utils.ipc import wait_for_ipc
 from trinity.boot_info import BootInfo
 
-LOG_BACKUP_COUNT = 10
+LOG_BACKUP_COUNT = 20
 LOG_MAX_MB = 5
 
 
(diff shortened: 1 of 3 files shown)


trinity/_utils/logging.py
```diff
@@ -154,8 +154,10 @@ def format(self, record: logging.LogRecord) -> str:
 )
 
 
-def set_logger_levels(log_levels: Dict[str, int],
-                      *handlers: logging.Handler) -> None:
+def set_logger_levels(log_levels: Dict[str, int]) -> None:
+    # async-service's DEBUG logs completely drowns our stuff (i.e. more than 95% of all our DEBUG
+    # logs), so by default set it to INFO.
+    logging.getLogger("async_service").setLevel(logging.INFO)
```
pipermerriam, line 160: This doesn't seem like the right location for this.  I'd much rather we hoist this up to a higher level and inject it into the `log_levels` argument that is passed into this function.

gsalgado: Ok, I've tried to do that but ran into https://github.com/ethereum/trinity/issues/1866

I'm not super excited about digging into that, but I've found a way to workaround it and updated this PR with it. Let me know what you think

trinity/_utils/logging.py
```diff
@@ -154,22 +154,22 @@ def format(self, record: logging.LogRecord) -> str:
...
         # The root logger is configured separately
         if name is None:
             continue
 
-        logger = get_logger(name)
+        # Workaround for https://github.com/ethereum/trinity/issues/1866
+        if name == 'async_service':
+            logger = logging.getLogger("async_service")
+        else:
+            logger = get_logger(name)
```
pipermerriam, line 168: So the reason this doesn't work is very likely related to this code from `eth-utils`

https://github.com/ethereum/eth-utils/blob/386863e6b7d95ac5acb11f6ba81619dc88ed6eb9/eth_utils/logging.py#L78-L81

The stdlib `logging` module caches logger instances and that code is ensuring that the logger instance is using the logger class specified...

Maybe a better way to approach this would be something like:

```python
if name.startswith('trinity.') or name.startswith('p2p.'):
    # use get_logger
else:
    # use logging.getLogger
```

PR #1867, opened 2020-07-10
A few tweaks to try and find more reachable/compatible peers

gsalgado:
The main cause I've identified for us not connecting to enough peers was that we were constantly retrying peers that we had already failed to connect to, instead of trying to find and connect to new ones. With the following changes we'll discover more compatible peers and attempt to connect to them before retrying old ones

- Blacklist peers that are unreachable, timeout or have no matching caps (those account for ~60% of handshake errors)

- Blacklist peers that fail the DAO check

- Increase blacklist timeout for all handshake errors. Most of them were set to 10s, effectively causing us to retry these before we had a chance of trying new ones.

- PeerPool now requests a larger batch of peer candidates on every iteration, effectively pushing DiscoveryService # to trigger new peer lookups in order to find enough compatible peers to fulfill our request

### DIFF
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -100,6 +100,9 @@ ALLOWED_PEER_CONNECTION_EXCEPTIONS = cast(Tuple[Type[BaseP2PError], ...], (
 )) + COMMON_PEER_CONNECTION_EXCEPTIONS
 
 
+OVER_PROVISION_MISSING_PEERS = 4
+
+
 class BasePeerPool(Service, AsyncIterable[BasePeer]):
     """
     PeerPool maintains connections to up-to max_peers on a given network.
@@ -199,10 +202,15 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
 
         skip_list = connected_node_ids.union(blacklisted_node_ids)
         should_skip_fn = functools.partial(should_skip_fn, skip_list)
+        # Request a large batch on every iteration as that will effectively push DiscoveryService
+        # to trigger new peer lookups in order to find enough compatible peers to fulfill our
+        # request. There's probably some room for experimentation here in order to find an optimal
+        # value.
+        max_candidates = self.available_slots * OVER_PROVISION_MISSING_PEERS
         try:
             candidates = await asyncio.wait_for(
                 backend.get_peer_candidates(
-                    max_candidates=self.available_slots,
+                    max_candidates=max_candidates,
                     should_skip_fn=should_skip_fn,
                 ),
                 timeout=REQUEST_PEER_CANDIDATE_TIMEOUT,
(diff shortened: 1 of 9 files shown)


tests/p2p/test_connection_tracker_timeout_for_failure.py
```diff
@@ -30,8 +34,12 @@ def _prevent_global_mutation_of_registry():
...
 
 
-def test_get_timeout_for_failure_with_HandshakeFailure():
-    assert get_timeout_for_failure(HandshakeFailure()) == 10
+def test_get_timeout_for_common_handshake_exceptions():
+    common_exc_types = COMMON_PEER_CONNECTION_EXCEPTIONS + (
+        HandshakeFailure, HandshakeFailureTooManyPeers, MalformedMessage,
+        NoMatchingPeerCapabilities)
+    for exc_type in common_exc_types:
+        assert get_timeout_for_failure(exc_type()) > 0
```
carver, line 42: No objection to dropping the tight `==` check on the blackout period. But maybe we can also check an upper bound?

Looks like the current upper bound is 1 hour. It will be painless to increase later if we decide to. At least the test would catch if someone accidentally increases one blackout period too high.

p2p/peer_pool.py
```diff
@@ -435,7 +434,7 @@ async def connect_to_node(self, node: NodeAPI) -> None:
 
         async with self.lock_node_for_handshake(node):
             if self.is_connected_to_node(node):
-                self.logger.debug(
+                self.logger.info(
```
carver, line 437: Seems like a `debug` level kind of thing.

p2p/peer_pool.py
```diff
@@ -199,10 +199,15 @@ async def _add_peers_from_backend(
 
         skip_list = connected_node_ids.union(blacklisted_node_ids)
         should_skip_fn = functools.partial(should_skip_fn, skip_list)
+        # Request a large batch on every iteration as that will effectively push DiscoveryService
+        # to trigger new peer lookups in order to find enough compatible peers to fulfill our
+        # request. There's probably some room for experimentation here in order to find an optimal
+        # value.
+        max_candidates = self.available_slots * 4
```
carver, line 206: Maybe move 4 to a constant. `OVER_PROVISION_MISSING_PEERS` or something.

p2p/peer_pool.py
gsalgado: Indeed, not sure why I've changed it

p2p/peer_pool.py
gsalgado: Sure, but I've added it to this file as it's only used here

pipermerriam:
This log is now a bit too noisy IMHO:

```
    INFO  2020-07-13 17:31:26,020           ETHPeerPool  Initiating 10 peer connection attempts with 17 open peer slots
```

I'd advocate moving it back to `DEBUG` level.

gsalgado:
That change was merged by accident. I reverted it in #1876

PR #1873, opened 2020-07-10
Ensure eth_* JSON-RPC APIs work with all block identifier during beam sync

cburgdorf:
### What was wrong?

During beam sync there's some extra machinery that ensures state dependent `eth_*` JSON-RPC API calls fetch missing state on demand from peers.

This extra machinery did not work when the block was specified via a hex string.

### How was it fixed?

1. Added a call to `to_int_if_hex` in `trinity/rpc/retry.py`

2. Added tests that ensure these APIs work with all kinds of valid block identifier.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/json-rpc/test_rpc_during_beam_sync.py
+++ b/tests/core/json-rpc/test_rpc_during_beam_sync.py
@@ -204,16 +204,19 @@ def fake_beam_syncer(chain, event_bus):
 # Test that eth_getBalance works during beam sync
 
 
+@pytest.mark.parametrize(
+    "at_block", ('latest', '0x0', 0)
+)
 @pytest.mark.asyncio
 async def test_getBalance_during_beam_sync(
-        chain, ipc_request, funded_address, funded_address_initial_balance,
-        fake_beam_syncer):
+        chain, ipc_request, funded_address, funded_address_initial_balance, fake_beam_syncer,
+        at_block):
     """
     Sanity check, if we call eth_getBalance we get back the expected response.
     """
 
     # sanity check, by default it works
-    response = await ipc_request('eth_getBalance', [funded_address.hex(), 'latest'])
+    response = await ipc_request('eth_getBalance', [funded_address.hex(), at_block])
     assert 'error' not in response
     assert response['result'] == hex(funded_address_initial_balance)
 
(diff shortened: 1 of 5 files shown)


tests/core/json-rpc/test_rpc_during_beam_sync.py
```diff
@@ -232,6 +232,38 @@ async def test_getBalance_during_beam_sync(
         assert response['result'] == hex(funded_address_initial_balance)
 
 
+# Test that eth_getTransactionCount works during beam sync
```
carver, line 238: I think the method tame is self-evident. But if you still want to comment something, it seems like a docstring is the right place for it.

tests/core/json-rpc/test_rpc_during_beam_sync.py
```diff
@@ -204,62 +204,66 @@ async def collect_storage(event: CollectMissingStorage):
 # Test that eth_getBalance works during beam sync
 
 
+@pytest.mark.parametrize(
+    "at_block", ('latest', '0x0', 0)
```
carver, line 208: Nice, for adding a check for the hex version of the block number. The order of the decorators looked backwards, so it's nice to confirm they work.

trinity/rpc/retry.py
```diff
@@ -83,7 +84,8 @@ async def check_requested_block_age(chain: Union[AsyncChainAPI, BaseAsyncBeaconC
 
     # Beacon chain doesn't support it
     if not isinstance(chain, BaseAsyncBeaconChainDB):
-        at_block = params.arguments[at_block_name]
+        at_block = to_int_if_hex(params.arguments[at_block_name])
```
carver, line 87: Oh, the decorators didn't work! Since they looked broken to my eyes, it might be better to just revert the decorators, so the `to_hex_if_int` is applied before the retryable check. (I think that should work) It seems nice not to have to convert it from hex to int in two different places.

cburgdorf: Yeah, that's what I was thinking, too but I think I tried applying them in the other order and I still needed to handle them here separately. I will double check that once more.

tests/core/json-rpc/test_rpc_during_beam_sync.py
cburgdorf: I agree, I just wanted to conform to the rest of the test file. I'll clean it up and remove these comments.

trinity/rpc/retry.py
cburgdorf: I double checked, we still need this even after the decorators are flipped. I flipped them nonetheless as it seems to be more accurate.

Issue #1875, opened 2020-07-13
Trinity crash looking up protocol while connection is closing.


pipermerriam:
### What is wrong?

Another probably race condition:

```
Traceback (most recent call last):
  File "/home/piper/python-environments/trinity/lib/python3.8/site-packages/async_service/asyncio.py", line 367, in cleanup_tasks
    yield
  File "/home/piper/python-environments/trinity/lib/python3.8/site-packages/async_service/asyncio.py", line 435, in background_asyncio_service
    yield manager
  File "/home/piper/projects/trinity/trinity/components/builtin/syncer/component.py", line 388, in do_run
    await wait_first([sync_task, asyncio.create_task(node_manager.wait_finished())])
  File "/home/piper/projects/trinity/p2p/logic.py", line 93, in wait_first
    await cancel_futures(pending)
  File "/home/piper/projects/trinity/p2p/logic.py", line 166, in cancel_futures
    raise MultiError(errors)
asyncio.exceptions.TimeoutError: ('Tasks never returned after being cancelled: %s', {<Task finished name='Task-19' coro=<SyncerComponent.launch_sync() done, defined at /home/piper/projects/trinity/trinity/components/builtin/syncer/component.py:390> exception=OSError(9, 'Bad file descriptor')>})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/piper/python-environments/trinity/lib/python3.8/site-packages/asyncio_run_in_process/_child.py", line 205, in run_process
    runner(async_fn, args, to_parent)
  File "/home/piper/python-environments/trinity/lib/python3.8/site-packages/asyncio_run_in_process/_child.py", line 168, in _run_on_asyncio
    result: Any = loop.run_until_complete(_do_async_fn(async_fn, args, to_parent, loop))
  File "uvloop/loop.pyx", line 1456, in uvloop.loop.Loop.run_until_complete
  File "/home/piper/python-environments/trinity/lib/python3.8/site-packages/asyncio_run_in_process/_child.py", line 160, in _do_async_fn
    return await async_fn_task
  File "/home/piper/python-environments/trinity/lib/python3.8/site-packages/asyncio_run_in_process/_child.py", line 85, in _handle_coro
    return await coro_task
  File "/home/piper/projects/trinity/trinity/extensibility/asyncio.py", line 51, in _do_run
    await self.do_run(event_bus)
[...]

Issue #1874, opened 2020-07-10
Trinity crash on AttributeError related to `p2p_api`


pipermerriam:
### What is wrong?

The following error took my node down.

```
   ERROR  2020-07-10 13:21:41,224                  root  Error during trinity run
asyncio_run_in_process._utils.RemoteTraceback:
""" (exception from process: 3583537)
Traceback (most recent call last):
  File "/home/piper/python-environments/trinity/lib/python3.8/site-packages/async_service/asyncio.py", line 367, in cleanup_tasks
    yield
  File "/home/piper/python-environments/trinity/lib/python3.8/site-packages/async_service/asyncio.py", line 435, in background_asyncio_service
    yield manager
  File "/home/piper/projects/trinity/trinity/components/builtin/syncer/component.py", line 388, in do_run
    await wait_first([sync_task, asyncio.create_task(node_manager.wait_finished())])
  File "/home/piper/projects/trinity/p2p/logic.py", line 93, in wait_first
    await cancel_futures(pending)
  File "/home/piper/projects/trinity/p2p/logic.py", line 166, in cancel_futures
    raise MultiError(errors)
asyncio.exceptions.TimeoutError: ('Tasks never returned after being cancelled: %s', {<Task pending name='Task-19' coro=<SyncerComponent.launch_sync() running at /home/piper/projects/trinity/trinity/components/builtin/syncer/component.py:397> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7fd51598ae50>()]>>})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/piper/python-environments/trinity/lib/python3.8/site-packages/asyncio_run_in_process/_child.py", line 168, in _run_on_asyncio
    result: Any = loop.run_until_complete(_do_async_fn(async_fn, args, to_parent, loop))
  File "uvloop/loop.pyx", line 1456, in uvloop.loop.Loop.run_until_complete
  File "/home/piper/python-environments/trinity/lib/python3.8/site-packages/asyncio_run_in_process/_child.py", line 160, in _do_async_fn
    return await async_fn_task
  File "/home/piper/python-environments/trinity/lib/python3.8/site-packages/asyncio_run_in_process/_child.py", line 85, in _handle_coro
    return await coro_task
  File "/home/piper/projects/trinity/trinity/extensibility/asyncio.py", line 51, in _do_run
    await self.do_run(event_bus)
[...]

PR #1876, opened 2020-07-13
Peer.run() no longer crashes if self._p2p_api is not set

gsalgado:
The finally: block in Peer.run() could, in some cases, execute without
self._p2p_api being set, causing it to crash. That is now fixed.

Also, if a Peer behavior crashes unexpectedly, we now propagate that
instead of silently terminating.

Closes: #1875
Closes: #1874
Closes: #1877

### DIFF
--- a/p2p/logic.py
+++ b/p2p/logic.py
@@ -90,7 +90,8 @@ async def wait_first(futures: Sequence[asyncio.Future[None]]) -> None:
         await cancel_futures(futures)
         raise
     else:
-        await cancel_futures(pending)
+        if pending:
+            await cancel_futures(pending)
         if len(done) != 1:
             raise Exception(
                 "Invariant: asyncio.wait() returned more than one future even "
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -293,19 +293,21 @@ class BasePeer(Service):
                 self.connection.start_protocol_streams()
                 self.ready.set()
 
-                try:
-                    await wait_first(futures)
-                except asyncio.CancelledError:
-                    raise
-                except BaseException:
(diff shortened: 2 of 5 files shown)


carver:
Doesn't look like exactly the same thing, but probably related:

```
   ERROR  2020-07-13 15:15:46,739           ETHPeerPool  unexpected error during peer connection
  File "/home/jcarver/code/trinity/trinity/protocol/common/peer.py", line 238, in maybe_connect_more_peers
    for backend in self.peer_backends
  File "/home/jcarver/code/trinity/p2p/peer_pool.py", line 224, in _add_peers_from_backend
    await self.connect_to_nodes(candidates)
  File "/home/jcarver/code/trinity/p2p/peer_pool.py", line 413, in connect_to_nodes
    await asyncio.gather(*(self.connect_to_node(node) for node in batch))
  File "/home/jcarver/code/trinity/p2p/peer_pool.py", line 459, in connect_to_node
    await peer.disconnect(DisconnectReason.TOO_MANY_PEERS)
  File "/home/jcarver/code/trinity/p2p/peer.py", line 319, in disconnect
    self.disconnect_nowait(reason)
  File "/home/jcarver/code/trinity/p2p/peer.py", line 330, in disconnect_nowait
    self._send_disconnect(reason)
  File "/home/jcarver/code/trinity/p2p/peer.py", line 334, in _send_disconnect
    self.p2p_api.disconnect(reason)
AttributeError: 'ETHPeer' object has no attribute 'p2p_api'
```

Should I make a new issue, or do you think this will cover it?

gsalgado:
> Should I make a new issue, or do you think this will cover it?

This will cover it

gsalgado:
Superseded by #1884

Issue #1797, opened 2020-06-18
p2p.exceptions.ConnectionBusy exception taking whole node down


cburgdorf:
### What is wrong?

```
    INFO  2020-06-18 15:27:40,544     BeamStateBackfill  Unexpected err while getting background nodes from ETHPeer (eth, 63) <Session <Node(0xe44bbd@88.99.161.14)> cefd5046-d66e-4a22-97a5-88a0469ab079>: 
    INFO  2020-06-18 15:27:40,548     BeamStateBackfill  Unexpected err while getting background nodes from ETHPeer (eth, 63) <Session <Node(0xb5948a@18.218.250.66)> 1ecca656-ed49-43a8-9656-8f7372921d3e>: 
    INFO  2020-06-18 15:27:44,809           ETHPeerPool  Removing ETHPeer (eth, 63) <Session <Node(0xb5948a@18.218.250.66)> 1ecca656-ed49-43a8-9656-8f7372921d3e> from pool: local_reason=None remote_reason=None
    INFO  2020-06-18 15:27:55,917           ETHPeerPool  Removing ETHPeer (eth, 65) <Session <Node(0x6f621a@91.13.207.112)> daab170d-c0d7-4a9f-a605-768d3f271a51> from pool: local_reason=None remote_reason=None
    INFO  2020-06-18 15:27:58,986           ETHPeerPool  Removing ETHPeer (eth, 63) <Session <Node(0xe44bbd@88.99.161.14)> cefd5046-d66e-4a22-97a5-88a0469ab079> from pool: local_reason=None remote_reason=None
    INFO  2020-06-18 15:27:59,064            FullServer  TCP Listener finished, cancelling Server
    INFO  2020-06-18 15:27:59,090      ComponentManager  Received shutdown request: Node exiting. Triggering shutdown
    INFO  2020-06-18 15:27:59,090      ComponentManager  Shutting down, reason: Node exiting. Triggering shutdown
    INFO  2020-06-18 15:27:59,091      ComponentManager  Stopping components
[1m[31m   ERROR  2020-06-18 15:27:59,090                  Sync  Unexpected error in component <trinity.components.builtin.syncer.component.SyncerComponent object at 0x7ff9fa69c790>, shutting down trinity
Traceback (most recent call last):
  File "/usr/src/app/trinity/p2p/exchange/candidate_stream.py", line 81, in payload_candidates
    timeout=total_timeout * NUM_QUEUED_REQUESTS,
[...]

PR #1879, opened 2020-07-13
Implement pause/resume for header backfill

cburgdorf:
### What was wrong?

As discussed in #1797 Trinity is taken down by a `ConnectionBusy` error if the tip syncer + the historical header backfill syncer both operate on the same peer at the same time.

### How was it fixed?

1. Allow additional `GetBlockHeaders` requests to queue up

2. Implement pause/resume for the historical header backfill and let it be governed by beam syncs lag monitoring. If we lag too far behind the historical backfill will pause.

3. The logic has some overlap with pause/resume of the block backfill so there's a bit of additional refactoring to reduce boilerplate.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1879.bugfix.rst
+++ b/newsfragments/1879.bugfix.rst
@@ -1,0 +1,3 @@
+Fix ``ConnectionBusy`` exception taking whole node down. The fix makes it so
+that additional``GetBlockHeaders`` request (for historical backfill) are allowed
+to be queued up.
--- a/p2p/exchange/constants.py
+++ b/p2p/exchange/constants.py
@@ -10,4 +10,4 @@ ROUND_TRIP_TIMEOUT = 20.0
 # they are sent. This is an estimate of how many get queued locally. The reason we
 # estimate the queue length is to determine how long a timeout to use when
 # waiting for the lock to send the next queued peer request.
-NUM_QUEUED_REQUESTS = 3
+NUM_QUEUED_REQUESTS = 4
--- a/tests/core/p2p-proto/test_sync.py
+++ b/tests/core/p2p-proto/test_sync.py
@@ -809,13 +809,15 @@ async def test_header_gap_fill_detects_invalid_attempt(caplog,
     )
     async with peer_pair as (client_peer, server_peer):
 
+        client_peer_pool = MockPeerPoolWithConnectedPeers([client_peer], event_bus=event_bus)
         client = SequentialHeaderChainGapSyncer(
             LatestTestChain(chaindb_with_gaps.db),
             chaindb_with_gaps,
-            MockPeerPoolWithConnectedPeers([client_peer], event_bus=event_bus)
+            client_peer_pool
         )
         server_peer_pool = MockPeerPoolWithConnectedPeers([server_peer], event_bus=event_bus)
         uncle_chaindb = AsyncChainDB(chaindb_uncle.db)
+
         async with run_peer_pool_event_server(
             event_bus, server_peer_pool, handler_type=ETHPeerPoolEventServer
         ), background_asyncio_service(ETHRequestServer(
@@ -827,11 +829,12 @@ async def test_header_gap_fill_detects_invalid_attempt(caplog,
 
             # We check for 499 because 500 exists from the very beginning (the checkpoint)
             expected_block_number = 499
+            final_header = chaindb_1000.get_canonical_block_header_by_number(expected_block_number)
             async with background_asyncio_service(client):
                 try:
                     await wait_for_head(
                         chaindb_with_gaps,
-                        chaindb_1000.get_canonical_block_header_by_number(expected_block_number),
+                        final_header,
                         sync_timeout=5,
                     )
                 except asyncio.TimeoutError:
@@ -839,9 +842,18 @@ async def test_header_gap_fill_detects_invalid_attempt(caplog,
                     # Monkey patch the uncle chaindb to effectively make the attacker peer
                     # switch to the correct chain.
                     uncle_chaindb.db = chaindb_1000.db
+                    # The hack goes on: Now that our attacker peer turned friendly we may be stuck
+                    # waiting for a new skeleton peer forever. This isn't a real life scenario
+                    # because: a.) an attacker probably won't turn friendly and b.) new blocks and
+                    # peers will constantly yield new skeleton peers.
+                    # This ugly hack will tick the chain tip monitor as we simulate a joining peer.
+                    for subscriber in client_peer_pool._subscribers:
+                        subscriber.register_peer(client_peer)
+
                     await wait_for_head(
                         chaindb_with_gaps,
-                        chaindb_1000.get_canonical_block_header_by_number(expected_block_number)
+                        final_header,
+                        sync_timeout=20,
                     )
                 else:
                     raise AssertionError("Succeeded when it was expected to fail")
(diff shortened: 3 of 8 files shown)


tests/core/p2p-proto/test_sync.py
```diff
@@ -819,21 +840,31 @@ async def test_header_gap_fill_detects_invalid_attempt(caplog,
...
                     # Monkey patch the uncle chaindb to effectively make the attacker peer
                     # switch to the correct chain.
                     uncle_chaindb.db = chaindb_1000.db
+                    # The hack goes on: Now that our attacker peer turned friendly we may be stuck
+                    # waiting for a new skeleton peer forever. This isn't a real life scenario
+                    # because: a.) an attacker probably won't turn friendly and b.) new blocks and
+                    # peers will constantly yield new skeleton peers.
+                    # This ugly hack will tick the chain tip monitor as we simulate a joining peer.
+                    for subscriber in client_peer_pool._subscribers:
+                        subscriber.register_peer(client_peer)
```
cburgdorf, line 862: Let me express my deepest apologies for this hack. I spent two days debugging this test after I noticed how flaky it was. I don't have a better solution to offer in a reasonable time frame :man_shrugging:

p2p/exchange/constants.py
```diff
@@ -10,4 +10,4 @@
 # they are sent. This is an estimate of how many get queued locally. The reason we
 # estimate the queue length is to determine how long a timeout to use when
 # waiting for the lock to send the next queued peer request.
-NUM_QUEUED_REQUESTS = 3
+NUM_QUEUED_REQUESTS = 5
```
carver, line 13: I think 4 is sufficient. It already feels like a lot. If you find that 5 is necessary, I'd be curious to understand why.

tests/core/p2p-proto/test_sync.py
```diff
@@ -773,6 +775,23 @@ async def test_sequential_header_gapfill_syncer(request,
...
+                # may be shorter than the actual gap in the chain.
+                await wait_for_head(
+                    chaindb_with_gaps, chaindb_1000.get_canonical_block_header_by_number(100)
+                )
+
+                # Pause the syncer for a moment and check if it continued syncing (it should not!)
+                client.pause()
+                await asyncio.sleep(1)
+                with pytest.raises(HeaderNotFound):
+                    chaindb_with_gaps.get_canonical_block_header_by_number(249)
```
carver, line 791: I guess we might want to confirm that 249 isn't present before the pause (in case the test runs really fast sometimes). It will help us tell if the test is flaky from a fast run, or if there's actually a bug with pausing.
```suggestion
                await wait_for_head(
                    chaindb_with_gaps, chaindb_1000.get_canonical_block_header_by_number(100)
                )

                # Pause the syncer for a moment and check if it continued syncing (it should not!)
                client.pause()
                # Verify that we stopped the chain fast enough, before the gap was fully filled
                # This is a potential source of test flakiness
                with pytest.raises(HeaderNotFound):
                    chaindb_with_gaps.get_canonical_block_header_by_number(249)
                await asyncio.sleep(1)
                # Make sure that the gap filling doesn't complete for a while
                # We could theoretically get false positives if it's not paused but very slow to fill headers
                with pytest.raises(HeaderNotFound):
                    chaindb_with_gaps.get_canonical_block_header_by_number(249)
```

tests/core/p2p-proto/test_sync.py
carver: I suppose the server peer could send the client peer a `NewBlockHash` message (or `NewBlock`) to trigger sync again? (But if that's not a quick and easy change, then no objection to this hack)

trinity/_utils/pause_resume.py
```diff
@@ -0,0 +1,49 @@
+import asyncio
+
+
+class PauseResume:
+    """
+    A helper class to provide pause / resume functionality to other classes.
+    """
```
carver, line 7: :+1: Nice to have this separated out.

trinity/_utils/pause_resume.py
```diff
@@ -0,0 +1,49 @@
...
+        if self._paused:
+            raise RuntimeError(
+                "Invalid action. Can not pause an operation that is already paused."
+            )
+
+        self._paused = True
+
+    def resume(self) -> None:
+        """
+        Resume the sync.
```
carver, line 33:
```suggestion
        Resume the operation.
```

trinity/sync/beam/chain.py
```diff
@@ -85,7 +85,7 @@
 
 from .backfill import BeamStateBackfill
 from ..header.chain import SequentialHeaderChainGapSyncer
-
+from ..._utils.pause_resume import PauseResume
```
carver, line 88: This seems much quicker to see what's happening with the import:
```py
from trinity.sync.header.chain import SequentialHeaderChainGapSyncer
from trinity._utils.pause_resume import PauseResume
```

Maybe worth adding to the tactical manual. Something like:

Apart from "sibling" imports, we prefer explicit absolute imports. They are easier to read at a glance. They don't require checking the path of the module you are in. Also, relative imports interrupt our alphabetical ordering.

trinity/_utils/pause_resume.py
```diff
@@ -0,0 +1,49 @@
+import asyncio
+
+
+class PauseResume:
```
carver, line 4: Another name if you like it: `Pauser`

newsfragments/1879.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fix ``ConnectionBusy`` exception taking whole node down. The fix makes it so
+that additional``GetBlockHeaders`` request (for historical backfill) are allowed
+to be queued up.
```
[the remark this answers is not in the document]
cburgdorf: It did appear in the last release (which already had the header backfill)

p2p/exchange/constants.py
cburgdorf: Ok, I just read your comment in the other issue

>Three is just a general good-enough number, I don't think we were using the full 3 queued requests for headers before (I think we were only using 2).

I'm good to go with 3 if you think it is sufficient. I still don't fully grasp why as I count 4 potential consumers (skeleton tip, meat tip, skeleton backfill, meat backfill) but I guess it's a little more nuanced and I trust your intuition :)

tests/core/p2p-proto/test_sync.py
cburgdorf: Ah good point!

trinity/sync/beam/chain.py
cburgdorf: Yeah, I agree. The auto import of PyCharm are a curse and blessing :grin:

trinity/_utils/pause_resume.py
cburgdorf: Oh yeah, I like that one.

tests/core/p2p-proto/test_sync.py
cburgdorf: Yeah, that's what I tried first but it didn't work. I'm not entirely sure where this gets stuck but it may be a problem with the peer_pool mock as well. I don't want to go down investigating this now so I'll keep the hack as is for now.

p2p/exchange/constants.py
carver: Wait, I said 4 was sufficient. I think 3 will still get `ConnectionBusy` sometimes. :)

carver: Oh, you picked four, good :)

cburgdorf: Ah yeah, I miss-typed there :sweat_smile:

PR #1883, opened 2020-07-14
Log backfill stats to info

### LINKED ISSUE #1872: Periodic status messages for state backfill
### What is wrong?

There is no indication in the console logs that state backfill is happening.

### How can it be fixed

Print a periodic message with stats on on the progress.  Potential stats are:

- total trie nodes synced
- rate of trie nodes being synced.
- approximate percentage of the trie explored (maybe just treat it as percentage of accounts synced)
- number of contract storages currently syncing
- "width" left and right of the current focal point in the main account trie that has been traversed.

carver:
### What was wrong?

Will fix #1872 

### How was it fixed?

Estimate using the account trie.

Show the completion near the backfill index key as parts per million so that you can see progress every 30s.

Other tiny fixes.

Note that this still doesn't work very well, until #1825 is applied.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/backfill.py
+++ b/trinity/sync/beam/backfill.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 import asyncio
 from collections import Counter
 from functools import partial
+import itertools
 import typing
 from typing import (
     AsyncIterator,
@@ -28,8 +29,14 @@ from trie import (
     exceptions as trie_exceptions,
     fog,
 )
+from trie.constants import (
+    BLANK_NODE_HASH,
+)
 from trie.utils.nibbles import (
     bytes_to_nibbles,
+)
+from trie.utils.nodes import (
+    key_starts_with,
 )
 from trie.typing import (
     HexaryTrieNode,
@@ -46,6 +53,7 @@ from trinity.sync.beam.constants import (
 )
 from trinity._utils.async_iter import async_take
 from trinity._utils.logging import get_logger
+from trinity._utils.timer import Timer
 
 from .queen import (
     QueeningQueue,
@@ -66,9 +74,11 @@ class BeamStateBackfill(Service, QueenTrackerAPI):
     the round-trip-time that peers take to respond to GetNodeData commands.
     """
 
-    _total_processed_nodes = 0
+    _total_added_nodes = 0
     _num_added = 0
     _num_missed = 0
+    _num_accounts_completed = 0
+    _num_storage_completed = 0
     _report_interval = 10
 
     _num_requests_by_peer: typing.Counter[ETHPeer]
@@ -221,7 +231,7 @@ class BeamStateBackfill(Service, QueenTrackerAPI):
                 # Check if account is fully downloaded
                 account_components_complete = self._are_account_components_complete(
                     address_hash_nibbles,
-                    account.code_hash,
+                    account,
                 )
                 if account_components_complete:
                     # Mark fully downloaded accounts as complete, and do some cleanup
@@ -382,6 +392,10 @@ class BeamStateBackfill(Service, QueenTrackerAPI):
         requested, or if there are no more missing nodes.
         """
 
+        if storage_root == BLANK_NODE_HASH:
+            # Nothing to do if the storage has an empty root
+            return
+
         storage_tracker = self._get_storage_tracker(address_hash_nibbles)
         storage_iterator = self._request_tracking_trie_items(
             storage_tracker,
@@ -472,9 +486,12 @@ class BeamStateBackfill(Service, QueenTrackerAPI):
     def _mark_account_complete(self, path_to_leaf: Nibbles, address_hash_nibbles: Nibbles) -> None:
         self._account_tracker.confirm_leaf(path_to_leaf)
 
+        self._num_accounts_completed += 1
+
         # Clear the storage tracker, to reduce memory usage
         #   and the time to check self._check_complete()
         if address_hash_nibbles in self._storage_trackers:
+            self._num_storage_completed += 1
             del self._storage_trackers[address_hash_nibbles]
 
         # Clear the bytecode tracker, for the same reason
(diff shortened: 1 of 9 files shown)


trinity/sync/beam/backfill.py
```diff
@@ -549,22 +554,23 @@ def set_root_hash(self, header: BlockHeaderAPI, root_hash: Hash32) -> None:
...
             await asyncio.sleep(self._report_interval)
 
             if not self._begin_backfill.is_set():
                 self.logger.debug("Beam-Backfill: waiting for new state root")
                 continue
 
-            msg = "all=%d" % self._total_processed_nodes
+            msg = "all=%d" % self._total_added_nodes
             msg += "  new=%d" % self._num_added
             msg += "  missed=%d" % self._num_missed
```
pipermerriam, line 570: maybe shorten `missed -> miss` and consider using formatting that maintains a minimum width of 4 integer characters will make all of these log statements line up well (for both `new` and `miss` metrics).

trinity/sync/beam/backfill.py
```diff
@@ -549,22 +554,23 @@ def set_root_hash(self, header: BlockHeaderAPI, root_hash: Hash32) -> None:
...
+
+            timer = Timer()
             await asyncio.sleep(self._report_interval)
 
             if not self._begin_backfill.is_set():
                 self.logger.debug("Beam-Backfill: waiting for new state root")
                 continue
 
-            msg = "all=%d" % self._total_processed_nodes
+            msg = "all=%d" % self._total_added_nodes
```
pipermerriam, line 568: maybe rename to `total=XXX`

carver:
Oops, I pushed a bunch of stuff from another branch, will get it tomorrow

trinity/sync/beam/backfill.py
carver: FWIW, the all number would practically get up to 9 digits wide if it was alive for the full state download. Not sure I want to add all that space

pipermerriam: For the `total` I think it is fine to let that grow as it needs space since it grows at a slow rate.  For the `new` and `miss` counts, it seems like if we reserve them 4 digits of space we'll typically have uniformly spaced log entries which seems nice.

trinity/sync/beam/backfill.py
```diff
@@ -31,6 +32,9 @@
 from trie.utils.nibbles import (
     bytes_to_nibbles,
 )
+from trie.utils.nodes import (
+    key_starts_with,
```
pipermerriam, line 36: Just noting that these are importing from what should be considered private namespaces.  We should fix this in `py-trie`, both by moving to `trie._utils` and by moving these functions to a public namespace.

trinity/sync/beam/backfill.py
```diff
@@ -549,22 +554,23 @@ def set_root_hash(self, header: BlockHeaderAPI, root_hash: Hash32) -> None:
...
             if not self._begin_backfill.is_set():
                 self.logger.debug("Beam-Backfill: waiting for new state root")
                 continue
 
-            msg = "all=%d" % self._total_processed_nodes
+            msg = "total=%d" % self._total_added_nodes
             msg += "  new=%d" % self._num_added
-            msg += "  missed=%d" % self._num_missed
+            msg += "  miss=%d" % self._num_missed
             msg += "  queen=%s" % self._queening_queue.queen
```
pipermerriam, line 571: I found the presence of the queen in this message odd  I'm not opposed to it, just seemed slightly out of place.

trinity/sync/beam/backfill.py
```diff
@@ -574,6 +580,122 @@ async def _periodically_report_progress(self) -> None:
...
+                    ) / num_storage_trackers
+                else:
+                    active_storage_completion = 1
+
+                self.logger.info(
+                    (
+                        "State backfill: accounts complete ~%.1f%%, active contracts d/l %d,"
+                        " complete storage of active contracts ~%d%%,"
+                        " confirmed width from index: ~%.1fppm, total nodes collected: %d,"
+                        " trie nodes / sec: %.1f"
```
pipermerriam, line 599: Some personal preference for these types of log lines

```
INFO       total=1212123434  accts=83%  storage=82 / 23%  walked=123ppm tnps=300
```

I recognize that you probably went for more verbose descriptive names since these are not metrics we are used to looking at.  My gut says we'll get used to them pretty quickly...

trinity/sync/beam/constants.py
```diff
@@ -42,8 +42,8 @@
 # to establish it (partially because we measure using an exponential average).
 
 # About how long after a block can we request trie data from peers?
-# The value is configurable by client, but tends to be around 120 blocks.
-ESTIMATED_BEAMABLE_BLOCKS = 120
+# The value is configurable by client, but tends to be around 75 blocks.
+ESTIMATED_BEAMABLE_BLOCKS = 75
```
pipermerriam, line 46: My geth node seems to hold state for ~100 blocks.  I'm still :+1: in using this smaller value here.

trinity/sync/beam/backfill.py
```diff
@@ -574,6 +580,122 @@ async def _periodically_report_progress(self) -> None:
...
+            # For now, report every 30s (1/3 as often as the debug report above)
+            if step % 3 == 0:
+                num_storage_trackers = len(self._storage_trackers)
+                if num_storage_trackers:
+                    active_storage_completion = sum(
+                        self._complete_trie_fraction(store_tracker)
+                        for store_tracker in self._storage_trackers.values()
+                    ) / num_storage_trackers
+                else:
+                    active_storage_completion = 1
```
pipermerriam, line 592: This results in showing a `100%` upon first initiating sync.  Maybe this should output as something like `N/A` to indicate no information...

pipermerriam:
Been looking at these logs for a bit this morning and now I feel like I have a better understanding of what information is useful.

1. I think we need to know how many storages have been completed, otherwise the number of active storages is hard to interpret because it is constantly having new ones added and finished ones removed.
2. The % of total accounts completed and the `width` of our current focus on the trie need to adjust their number of decimal places for how they display their number to allow for concise display of very small percentages (maybe down to 6 decimal places) all the way up to displaying the value as a whole number percentage once the number reaches >10%.

None of this is critical stuff.  Just making some notes based on my observations

trinity/sync/beam/backfill.py
carver: Noted in https://github.com/ethereum/py-trie/issues/85

trinity/sync/beam/backfill.py
carver: Yeah, right now, it's giving sort of a simple snapshot of which peer is queen and how stable it is. I don't think it needs to live there for long, so I'll at least add a comment, and maybe remove it. I think looking at `Beam-Sync-Peer-Usage` is probably more informative, anyway.

PR #1884, opened 2020-07-14
A few improvements to how Connection behaviors are handled

gsalgado:
These will hopefully make it less likely for us to introduce new code
that leaks PeerConnectionLost exceptions

- Behaviors are now applied/monitored by Connection, ensuring they're
  applied before a Peer starts and only removed after the Peer terminates

- Do not use cached_properties for Peer.head_info & co as if they are
  called while the Peer is being terminated (e.g. as the connection
  tracker does) they can raise a PeerConnectionLost/UnknownAPI

- Peer will not attempt to send a Disconnect if it's crashed before
  the self._p2_api attribute has been set

Closes: #1874
Closes: #1875

### DIFF
--- a/p2p/abc.py
+++ b/p2p/abc.py
@@ -595,6 +595,13 @@ class BehaviorAPI(ABC):
         ...
 
     @abstractmethod
+    def post_apply(self) -> None:
+        """
+        Called after all behaviors have been applied to the Connection.
+        """
+        ...
+
+    @abstractmethod
     @contextlib.asynccontextmanager
     def apply(self, connection: 'ConnectionAPI') -> AsyncIterator[asyncio.Future[None]]:
         """
@@ -624,6 +631,7 @@ class ConnectionAPI(ServiceAPI):
     #
     # Primary properties of the connection
     #
+    behaviors_applied: asyncio.Event
     is_dial_out: bool
 
     @property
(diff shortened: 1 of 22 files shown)


gsalgado:
@carver With those changes we can keep behaviors available until the Peer has finished, as you suggested in https://github.com/ethereum/trinity/pull/1840#discussion_r451131776

gsalgado:
@carver `test_propagates_behavior_crashes()` was failing intermittently because `run_behaviors()` was cancelling the connection when a behavior raised an exception, when what we really want is to let the exception propagate to async-service, which will then terminate the service and report the exception. The last commit fixes that and makes the changes you suggested

gsalgado:
And the last commit makes one final change so that `ChainInfoAPI` attributes don't raise PeerConnectionLost, as well as making it clear that all `HeadInfoAPI` attributes can do so

carver:
Yup LGTM :+1: 

The `post_apply()` feels like a bit of a hack. My intuition is that it should belong in a context manager somewhere*, but for now I'd rather see the bugs fixed than figure out the ideal pattern. :)

*Something like (pseudocode)
```py
@asynccontextmanager
async def apply_behavior():
  async with super().apply_behavior():
    logic = choose_eth_or_les_api(self.connection)
    self.network_id = logic.network_id
    yield
  ```
_Another reason I'm not pushing for this now is I wonder if there's a compositional approach that's better than this inheritance one._

gsalgado:
> My intuition is that it should belong in a context manager somewhere

That doesn't work because we need it to be called after *all* behaviors are applied. Otherwise the `choose_eth_or_les_api()` call will fail if the `ChainInfo` behavior is applied before the `ETH` behavior

carver:
I included a little piece of this fix in #1883 to pass CI. You will just be able to replace it with your patch during the rebase conflict.

PR #1890, opened 2020-07-14
Remove conflicting cli arg setup

cburgdorf:
### What was wrong?

The `--disable-backfill` flag was setup twice with different info texts but that's not possible and so only the last caption was visible which wasn't accurate.

### How was it fixed?

Removed one of the methods to add the args and adjusted name + text.

### To-Do

- [x] Clean up commit history

- no entry needed: Bug was never released.

#### Cute Animal Picture

### DIFF
--- a/trinity/components/builtin/syncer/component.py
+++ b/trinity/components/builtin/syncer/component.py
@@ -106,22 +106,12 @@ def add_sync_from_checkpoint_arg(arg_group: _ArgumentGroup) -> None:
     )
 
 
-def add_disable_backfill_for_beam_sync_arg(arg_group: _ArgumentGroup) -> None:
+def add_disable_backfill_arg(arg_group: _ArgumentGroup) -> None:
     add_shared_argument(
         arg_group,
         '--disable-backfill',
         action="store_true",
-        help="Disable backfilling entirely (headers, blocks, state)",
-        default=False,
-    )
-
-
-def add_disable_backfill_for_header_sync_arg(arg_group: _ArgumentGroup) -> None:
-    add_shared_argument(
-        arg_group,
-        '--disable-backfill',
-        action="store_true",
-        help="Disable backfilling of headers (introduced through checkpointing)",
+        help="Disable backfilling of historical headers and blocks",
         default=False,
     )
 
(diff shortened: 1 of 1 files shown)


trinity/components/builtin/syncer/component.py
```diff
@@ -106,22 +106,12 @@ def add_sync_from_checkpoint_arg(arg_group: _ArgumentGroup) -> None:
     )
 
 
-def add_disable_backfill_for_beam_sync_arg(arg_group: _ArgumentGroup) -> None:
+def add_disable_backfill_arg(arg_group: _ArgumentGroup) -> None:
     add_shared_argument(
         arg_group,
         '--disable-backfill',
         action="store_true",
-        help="Disable backfilling entirely (headers, blocks, state)",
```
cburgdorf, line 114: I noticed that this doesn't yet govern the backfill of state but I was wondering if it should do that? I think we probably want that and then as a bonus have extra flags to individually disable backfill of state or blocks.

trinity/components/builtin/syncer/component.py
```diff
@@ -106,22 +106,12 @@ def add_sync_from_checkpoint_arg(arg_group: _ArgumentGroup) -> None:
...
-        default=False,
-    )
-
-
-def add_disable_backfill_for_header_sync_arg(arg_group: _ArgumentGroup) -> None:
-    add_shared_argument(
-        arg_group,
-        '--disable-backfill',
-        action="store_true",
-        help="Disable backfilling of headers (introduced through checkpointing)",
```
cburgdorf, line 124: This was the caption that ended up displayed under `trinity --help`

trinity/components/builtin/syncer/component.py
carver: It's a little complicated. We use state backfill as a way to constantly sample the other peers for their speed of service of `GetNodeData`. For now, I think we can't disable it, or we won't correctly switch to the fastest peer for Beam Sync.

cburgdorf: Ah, alright :+1:

PR #1892, opened 2020-07-14
Apply UPNP to TCP as well

pipermerriam:
### What was wrong?

UPNP was only being applied to UDP.

### How was it fixed?

Expanded to apply it to TCP as well.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

tumblr_lkxo8lyf6z1qjehv2o1_400

### DIFF
--- a/newsfragments/1829.bugfix.rst
+++ b/newsfragments/1829.bugfix.rst
@@ -1,0 +1,1 @@
+Apply UPnP to TCP as well.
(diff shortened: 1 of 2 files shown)


pipermerriam:
@carver I'd be curious to have you apply this locally and see if you're getting inbound connections.  It still appears that I'm not getting any inbound connections which suggests something is still wrong.  I can confirm that my router is registering as having these entries so the forwarding rules are getting setup on the router, they just might not be quite working correctly.

carver:
Isn't devp2p all over udp?

pipermerriam:
No, discovery uses UDP, devp2p uses TCP connections (I'm like 99% sure about this).

carver:
Huh I guess not, because I did start getting inbound connections, including a lot of errors!

Issue #1886, opened 2020-07-14
Exception on shutdown


cburgdorf:
### What is wrong?

There's an exception thrown when shutting down after (light) sync. I haven't seen this in a long time.

https://circleci.com/gh/ethereum/trinity/257115

### How can it be fixed

The error suggests that it is related to looking up a subprotocol so there's a high chance this is already being worked on (cc @gsalgado )

gsalgado:
Always good to have the relevant logs/traceback directly on the issue, so here goes:

```
DEBUG  07-14 08:13:26  async_process_runner.py  b'Traceback (most recent call last):\n'

   DEBUG  07-14 08:13:26  async_process_runner.py  b'  File "/home/circleci/repo/.tox/py37-wheel-cli/lib/python3.7/site-packages/async_service/asyncio.py", line 435, in background_asyncio_service\n'

   DEBUG  07-14 08:13:26  async_process_runner.py  b'    yield manager\n'

   DEBUG  07-14 08:13:26  async_process_runner.py  b'  File "/home/circleci/repo/p2p/p2p_api.py", line 101, in apply\n'

   DEBUG  07-14 08:13:26  async_process_runner.py  b'    yield asyncio.create_task(manager.wait_finished())\n'

   DEBUG  07-14 08:13:26  async_process_runner.py  b'  File "/home/circleci/repo/p2p/behaviors.py", line 62, in apply\n'

   DEBUG  07-14 08:13:26  async_process_runner.py  b'    yield task\n'

   DEBUG  07-14 08:13:26  async_process_runner.py  b'  File "/home/circleci/repo/p2p/logic.py", line 143, in apply\n'

   DEBUG  07-14 08:13:26  async_process_runner.py  b'    yield asyncio.create_task(wait_first(futures))\n'

   DEBUG  07-14 08:13:26  async_process_runner.py  b'  File "/home/circleci/repo/p2p/behaviors.py", line 62, in apply\n'

   DEBUG  07-14 08:13:26  async_process_runner.py  b'    yield task\n'

   DEBUG  07-14 08:13:26  async_process_runner.py  b'  File "/home/circleci/repo/p2p/peer.py", line 268, in run\n'

   DEBUG  07-14 08:13:26  async_process_runner.py  b"    self.p2p_api = self.connection.get_logic('p2p', P2PAPI)\n"

[...]

PR #1897, opened 2020-07-15
Swallow PeerConnectionLost in Peer.run()

gsalgado:
That's needed because the Connection may be terminated by the time 
Peer.run() is called

Closes: #1886


carver:
You will need to rebase on #1883 after I merge it -- I included a little piece of this fix in order to pass CI. You will just be able to replace it with your patch during the rebase conflict.

PR #1900, opened 2020-07-15
--config-profile param for beacon node.

g-r-a-n-t:
### What was wrong?
It wasn't possible to select a chain or genesis config when booting up the beacon node, these values were hardcoded.

### How was it fixed?
Added a `--config-profile` param to beacon node that can be used to select the chain on which you would like to run and the genesis config that you would like to use.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth2/configs.py
+++ b/eth2/configs.py
@@ -28,7 +28,7 @@ def _decoder(
         elif field.type is bytes:
             yield field.name, decode_hex(cast(str, data[field.name]))
         elif field.type is Version:
-            yield field.name, decode_hex(cast(str, data[field.name]))
+            yield field.name, Version(decode_hex(cast(str, data[field.name])))
         else:
             yield field.name, int(data[field.name])
 
--- a/tests-trio/conftest.py
+++ b/tests-trio/conftest.py
@@ -3,6 +3,7 @@ import secrets
 import tempfile
 import uuid
 
+from eth2.beacon.state_machines.forks.skeleton_lake import MINIMAL_SERENITY_CONFIG
 from eth_keys.datatypes import PrivateKey
 from lahja import ConnectionConfig
 from lahja.trio.endpoint import TrioEndpoint
@@ -13,11 +14,8 @@ import pytest
 import trio
 
 from eth2._utils.bls import Eth2BLS, bls
-from eth2.beacon.chains.testnet.altona import BeaconChain
+from eth2.beacon.chains.testnet.altona import BeaconChainTest
 from eth2.beacon.constants import FAR_FUTURE_EPOCH
-from eth2.beacon.state_machines.forks.skeleton_lake.configs import (
-    MINIMAL_SERENITY_CONFIG,
-)
 from eth2.beacon.tools.builder.initializer import create_key_pairs_for
 from eth2.beacon.tools.misc.ssz_vector import override_lengths
 from eth2.beacon.types.blocks import BeaconBlockBody, BeaconBlockHeader, BeaconBlock
@@ -114,8 +112,8 @@ def genesis_block(genesis_state):
 
 
 @pytest.fixture
-def chain_config(genesis_state, eth2_config):
-    return BeaconChainConfig(genesis_state, eth2_config, {})
+def chain_config(genesis_state, eth2_config, chain_class):
+    return BeaconChainConfig(genesis_state, eth2_config, {}, beacon_chain_class=chain_class)
 
 
 @pytest.fixture
@@ -125,7 +123,7 @@ def database_dir(tmp_path):
 
 @pytest.fixture
 def chain_class():
-    return BeaconChain
+    return BeaconChainTest
 
 
 @pytest.fixture
--- a/trinity/components/eth2/beacon_trio/component.py
+++ b/trinity/components/eth2/beacon_trio/component.py
@@ -30,6 +30,7 @@ class BeaconNodeComponent(TrioComponent):
         trinity_config = self._boot_info.trinity_config
         beacon_app_config = trinity_config.get_app_config(BeaconAppConfig)
         config = BeaconNodeConfig.from_platform_config(
+            boot_info.args.config_profile,
             trinity_config,
             beacon_app_config,
             boot_info.args.validator_api_port,
@@ -74,6 +75,13 @@ class BeaconNodeComponent(TrioComponent):
             "--orchestration-profile",
             help="[temporary developer option] manage several beacon nodes on one machine",
             default="a",
+        )
+
+        arg_parser.add_argument(
+            "--config-profile",
+            help="the profile used to generate the genesis config",
+            choices=("minimal", "mainnet", "altona"),
+            default="minimal",
         )
 
     @property
--- a/trinity/components/eth2/beacon_trio/component.py
+++ b/trinity/components/eth2/beacon_trio/component.py
@@ -6,7 +6,7 @@ from eth_utils import to_tuple
 from multiaddr import Multiaddr
 
 from trinity.boot_info import BootInfo
-from trinity.config import BeaconAppConfig
+from trinity.config import BeaconTrioAppConfig
 from trinity.constants import BEACON_TESTNET_NETWORK_ID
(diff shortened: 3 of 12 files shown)


eth2/configs.py
```diff
@@ -28,7 +28,7 @@ def _decoder(
         elif field.type is bytes:
             yield field.name, decode_hex(cast(str, data[field.name]))
         elif field.type is Version:
-            yield field.name, decode_hex(cast(str, data[field.name]))
+            yield field.name, Version(decode_hex(cast(str, data[field.name])))
```
ralexstokes, line 31: resolved in #1898.

tests-trio/conftest.py
```diff
@@ -76,7 +74,7 @@ def node_key():
 
 @pytest.fixture
 def eth2_config():
-    config = MINIMAL_SERENITY_CONFIG
+    config = ALTONA_CONFIG
```
ralexstokes, line 77: have you noticed if this extends the time it takes to run these tests by any appreciable amount?

i think we need this as we end up hardcoding the `AltonaStateMachine` at some point.

i'd prefer to keep this as the `minimal` config and patch up the BeaconChain/StateMachine combo in this `conftest` module so that we can keep the lightweight version (unless it doesn't impact test length in which case this is likely simpler!)

trinity/config.py
```diff
@@ -111,6 +108,10 @@
     PRECONFIGURED_NETWORKS
 )
 
+from eth2.beacon.chains.testnet import SkeletonLakeChain
+from eth2.beacon.chains.testnet.altona import BeaconChain as AltonaChain
+from eth2.beacon.chains.base import BeaconChain
```
ralexstokes, line 113: i'd think we want to go ahead and remove the version of `eth2.beacon.chains.base`.

we can `raise NotImplementedError()` for the `mainnet` config if you are worried about not being perfectly accurate here...

trinity/config.py
```diff
@@ -713,20 +716,26 @@ def genesis_time(self) -> Timestamp:
...
         Construct an instance of ``cls`` reading the genesis configuration
         data under the local data directory.
         """
+        if config_profile == "mainnet":
+            beacon_chain_class = BeaconChain
+        elif config_profile == "altona":
+            # TODO: (g-r-a-n-t) We have two different `BaseBeaconChain` classes, which are
+            # used by different chains. This causes some typing issues, which we'll just ignore
+            # until the older one is removed and all chains use the same base class.
+            beacon_chain_class = AltonaChain  # type: ignore
```
ralexstokes, line 733: following the above comment, i'd prefer to drop the older definition and just use the one here called `AltonaChain` you can drop the alias on line 112 if you'd like and then either raise "not implemented" for `mainnet` or simply use the "altona chain" for both

trinity/nodes/beacon/config.py
```diff
@@ -1,10 +1,10 @@
 from pathlib import Path
-from typing import Collection
+from typing import Collection, Type
 
 from eth_keys.datatypes import PrivateKey
 from multiaddr import Multiaddr
 
-from eth2.beacon.chains.testnet.altona import BeaconChain
+from eth2.beacon.chains.base import BaseBeaconChain
```
ralexstokes, line 7: following the convo above, i'd prefer `eth2.beacon.chains.abc import BaseBeaconChain` here

trinity/nodes/beacon/full.py
```diff
@@ -189,7 +189,10 @@ def from_config(
             config.eth2_config,
             config.chain_config,
             config.database_dir,
-            config.chain_class,
+            # TODO: (g-r-a-n-t) We have two different `BaseBeaconChain` classes, which are
+            # used by different chains. This causes some typing issues, which we'll just ignore
+            # until the older one is removed and all chains use the same base class.
+            config.chain_class,  # type: ignore
```
ralexstokes, line 195: can likely drop this change if we do the above suggestions

eth2/configs.py
g-r-a-n-t: we would like to have it set as a `Version` type, as opposed to `bytes`, right?

tests-trio/conftest.py
g-r-a-n-t: it does slow it down quite a bit (~50x)

eth2/configs.py
ralexstokes: ah yes good catch!

tests-trio/conftest.py
```diff
@@ -3,6 +3,7 @@
 import tempfile
 import uuid
 
+from eth2.beacon.state_machines.forks.skeleton_lake import MINIMAL_SERENITY_CONFIG
```
ralexstokes, line 6: where is this used?

trinity/components/builtin/attach/component.py
```diff
@@ -88,8 +88,8 @@ def run_shell(cls, args: Namespace, trinity_config: TrinityConfig) -> None:
             config = trinity_config.get_app_config(Eth1AppConfig)
             with get_eth1_shell_context(config.database_dir, trinity_config) as context:
                 db_shell(is_ipython_available(), context)
-        elif trinity_config.has_app_config(BeaconAppConfig):
-            config = trinity_config.get_app_config(BeaconAppConfig)
+        elif trinity_config.has_app_config(BeaconTrioAppConfig):
+            config = trinity_config.get_app_config(BeaconTrioAppConfig)
```
ralexstokes, line 92: can we leave this file as `BeaconAppConfig`?

i think we should only need to use the `BeaconTrioAppConfig` for the beacon full node

trinity/components/builtin/attach/console.py
```diff
@@ -185,16 +185,14 @@ def get_eth1_shell_context(database_dir: Path,
...
     ipc_path = trinity_config.database_ipc_path
     trinity_already_running = ipc_path.exists()
 
     with _get_base_db(database_dir, ipc_path) as db:
         chain_config = app_config.get_chain_config()
         chaindb = BeaconChainDB(db)
-        chain = chain_config.beacon_chain_class(
-            chaindb,
-        )
+        chain = chain_config.beacon_chain_class()
```
ralexstokes, line 195: same thing here...

trinity/components/eth2/beacon_trio/component.py
```diff
@@ -28,8 +28,9 @@ def __init__(self, boot_info: BootInfo) -> None:
         super().__init__(boot_info)
 
         trinity_config = self._boot_info.trinity_config
-        beacon_app_config = trinity_config.get_app_config(BeaconAppConfig)
-        config = BeaconNodeConfig.from_platform_config(
+        beacon_app_config = trinity_config.get_app_config(BeaconTrioAppConfig)
+        config = BeaconTrioNodeConfig.from_platform_config(
```
ralexstokes, line 32: we shouldn't need to rename this one unless you think it is more clear this way

trinity/initialization.py
```diff
@@ -148,6 +148,15 @@ def initialize_beacon_database(chain_config: BeaconChainConfig,
         chain_config.initialize_chain(base_db)
 
 
+def initialize_beacon_trio_database(chain_config: BeaconTrioChainConfig,
```
ralexstokes, line 151: these are likely used in the `asyncio` app -- and i believe they have been dropped for now in lieu of `from_genesis` class methods that do this logic in the `trio` app

ralexstokes: ah never mind

trinity/nodes/beacon/config.py
ralexstokes: so we should be able to drop `eth2.beacon.chains.base.BaseBeaconChain` now, correct?

tests-trio/conftest.py
g-r-a-n-t: was removed the re-added https://github.com/ethereum/trinity/blob/5acb9d8d985275f448d8829792433f7c7e092aef/tests-trio/conftest.py#L77

PR #1909, opened 2020-07-17
Ensure DappNode dashboard shows Trinity

cburgdorf:
### What was wrong?

When running Trinity via the dappnode package, the current sync status isn't listed on the dashboard"

### How was it fixed?

1. Make configurations in manifest
2. Enable http (the dashboard requests the current head via the JSON-RPC API)

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/dappnode/dappnode_package.json
+++ b/dappnode/dappnode_package.json
@@ -3,7 +3,8 @@
   "version": "0.1.3",
   "description": "Ethereum Python Client (experimental)",
   "avatar": "/ipfs/QmRyJJwuCEYHPf7xkRcoPJqiHQgAN3ay2voSCTrAY1nMet",
-  "type": "service",
+  "type": "library",
+  "chain": "ethereum",
   "image": {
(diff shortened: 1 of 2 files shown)


dappnode/dappnode_package.json
```diff
@@ -3,7 +3,8 @@
   "version": "0.1.3",
   "description": "Ethereum Python Client (experimental)",
   "avatar": "/ipfs/QmRyJJwuCEYHPf7xkRcoPJqiHQgAN3ay2voSCTrAY1nMet",
-  "type": "service",
+  "type": "library",
+  "chain": "ethereum",
```
cburgdorf, line 7: Even if "library" feels odd, these are in fact the intended values:

https://github.com/dappnode/DAppNodePackage-goerli-geth/blob/55f656c45e265f9e40b28b47fae65891e06079ad/dappnode_package.json#L7-L8

Issue #1907, opened 2020-07-17
Geth's checkpoint challenge is ignored by us, causing remote to disconnect


gsalgado:
### What is wrong?

Upon connecting a geth instance to my local trinity, trinity consistently ignores geth's initial checkpoint challenge (https://github.com/ethereum/go-ethereum/blob/6d74d1e5f762e06a6a739a42261886510f842778/eth/handler.go#L344-L361), even though trinity has the header requested. Because of that geth assumes we're useless as we advertise a certain HEAD and do not provide a header prior to that when asked to

It looks like the checkpoint challenge request (the first msg after the handshake) is being dropped because when it arrives the peer has no subscribers, so https://github.com/ethereum/trinity/blob/50a2c376d49e606aaef2a73f0b8a4be5c13d6246/p2p/peer.py#L316-L321 is a no-op

In the original handshake implementation, I remember dealing with something similar, which was addressed by temporarily storing those messages and manually adding them to the subscriber's queues once the peer was added to the pool: https://github.com/ethereum/trinity/blob/50a2c376d49e606aaef2a73f0b8a4be5c13d6246/p2p/peer_pool.py#L307-L328

But we're no longer using the second argument to `_add_peer()`: https://github.com/ethereum/trinity/blob/50a2c376d49e606aaef2a73f0b8a4be5c13d6246/p2p/peer_pool.py#L290

carver:
> But we're no longer using the second argument to `_add_peer()`:
> 
> https://github.com/ethereum/trinity/blob/50a2c376d49e606aaef2a73f0b8a4be5c13d6246/p2p/peer_pool.py#L290

Hah, yeah, that looks obviously broken. :clap: for a clear write-up.

PR #1910, opened 2020-07-19
Only start streaming protocol messages after Peer is fully setup

gsalgado:
We used to start streaming protocol messages in the Peer.run() method,
but that happens before the PeerPool adds its subscribers to the Peer,
so messages could be dropped and that could cause remotes to disconnect
from us. Now it's the PeerPool that starts the streaming after adding
its subscribers, and it is also clearly documented that the streaming
should not start until the peer has subscribers

Closes: #1907

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -26,8 +26,6 @@ from lahja import EndpointAPI
 from cached_property import cached_property
 
 from eth_utils import (
-    to_tuple,
-    ExtendedDebugLogger,
     ValidationError,
 )
 
@@ -220,10 +218,6 @@ class BasePeer(Service):
     def session(self) -> SessionAPI:
         return self.connection.session
 
-    @property
-    def is_closing(self) -> bool:
-        return self.connection.is_closing
-
     def get_extra_stats(self) -> Tuple[str, ...]:
         return tuple()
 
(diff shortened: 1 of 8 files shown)


p2p/peer.py
```diff
@@ -268,6 +259,20 @@ def is_alive(self) -> bool:
...
+            raise LifecycleError("Cannot start protocol streams when peer is not running")
+        elif self.connection.is_streaming_messages:
+            raise LifecycleError("Connection is already streaming messages")
+
+        if len(self._subscribers) == 0:
+            self.logger.warning(
+                "Starting protocol streams while peer has no subscribers. Messages may be dropped")
+        else:
+            self.logger.debug(
+                "Starting protocol streams for %s. Subscirbers: %s", self, self._subscribers)
```
carver, line 273:
```suggestion
                "Starting protocol streams for %s. Subscribers: %s", self, self._subscribers)
```

p2p/peer.py
```diff
@@ -281,17 +286,21 @@ async def run(self) -> None:
...
             self.ready.set()
 
+            if self.connection.is_streaming_messages:
+                raise LifecycleError(
+                    "Connection should not start streaming messages until peer is running "
+                    "and had its subscribers added.")
+
+            self.logger.debug(
+                "Peer %s is running but won't start streaming messages until "
+                "start_protocol_streams() is called")
```
carver, line 303: I get how this log helped when coding the change, but do you think it needs to stay in for merge? Could maybe just be changed to a comment.

p2p/peer.py
```diff
@@ -512,20 +524,6 @@ def subscribe_peer(self, peer: BasePeer) -> Iterator[None]:
             peer.remove_subscriber(self)
 
 
-class MsgBuffer(PeerSubscriber):
```
carver, line 515: Yay for pruning dead code. :+1:

p2p/peer.py
```diff
@@ -443,35 +453,37 @@ def add_msg(self, msg: PeerMessage) -> bool:
         """
         peer, cmd = msg
 
+        if hasattr(self, 'logger'):
+            logger = self.logger  # type: ignore
+        else:
+            logger = logging.getLogger('p2p')
```
pipermerriam, line 459: this won't be guaranteed to be a an `ExtendedDebugLogger` which means the call to `logger.debug2` a few lines down will fail if this code path is taken.

p2p/peer.py
gsalgado: I actually added it to try and remind us of the need to manually trigger the msg streaming in case we introduce a regression. My reasoning being that this would be slightly better than a simple comment as we could notice it either by reading the code or the logs

p2p/peer.py
gsalgado: Right, I've changed it to use `get_logger('p2p.peer.BasePeer')`

Issue #1626, opened 2020-03-23
Make it possible to filter which RPC apis are exposed via HTTP


cburgdorf:
### What is wrong?

Running Trinity with `--enable-http` exposes every JSON-RPC API via HTTP. This includes management APIs such as `admin_addPeer` which can allow a malicious actor to add bad peers to your node.

### How can it be fixed

- Add a new flag `--http-rpc-apis` which takes a comma separated list of modules
  - e.g. `--http-rpc-apis 'net,eth'`
- Disallow running `--enable-http` *without* also providing `--http-rpc-apis`

cburgdorf:
@carver @pipermerriam I've put this on the `beta` milestone. I believe it is pretty essential to be able to expose the `eth` API via http while at the same time disallow things like `admin_addPeer` to be called from anyone who can reach the server via http.

carver:
> * Disallow running `--enable-http` _without_ also providing `--http-rpc-apis`

If we're going to this, we might as well combine into a single command to avoid the extra boilerplate typing, like `--enable-http-apis 'net,eth'`

PR #1911, opened 2020-07-20
Make it configurable which JSON-RPC APIs are exposed via HTTP

cburgdorf:
### What was wrong?

As mentioned in #1626, running Trinity with the JSON-RPC API accessible via HTTP is currently follows an all-or-nothing approach. This is unfortunate and even has security implications (even more so, once we implement `admin_removePeer`) as it may expose sensible administrative APIs to be misused by external attackers. 

### How was it fixed?

1. Remove the `--enable-http` flag
2. Add a `--enable-http-apis` flag that takes either `"*"` as a wildcard (current behavior) or a comma separated list of modules to expose e.g. `"net, eth"`
3. The `RPCServer` has a new `execute_with_access_control` API that allows passing disabled modules. The old `execute` API still exists but delegates to `execute_with_access_control` without any blocked modules. The `IPCServer` continues to use the existing `execute` API as it does not restrict module access.
4. The HTTP server uses the `execute_with_access_control` API and passes in the disabled modules (enabled modules from the CLI are translated into disabled modules in the process). Note: It felt more natural to me to have the RPCServer demand a configuration of disabled modules but have the CLI be focused on which modules are enabled.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/json-rpc/test_rpc_server.py
+++ b/tests/core/json-rpc/test_rpc_server.py
@@ -1,0 +1,54 @@
+import json
+import random
+
+import pytest
+from eth import MainnetChain
+
+from trinity.chains.full import FullChain
+from trinity.config import TrinityConfig
+from trinity.rpc import RPCServer
+from trinity.rpc.modules import initialize_eth1_modules, Admin, Net
+
+
+def build_request(method, params):
+    return {"jsonrpc": "2.0", "method": method, "params": params, "id": random.randrange(1000)}
+
+
+def result_from_response(response_str):
+    response = json.loads(response_str)
+    return (response.get('result', None), response.get('error', None))
+
+
+class MainnetFullChain(FullChain):
+    vm_configuration = MainnetChain.vm_configuration
+    chain_id = MainnetChain.chain_id
+
+
+@pytest.mark.parametrize(
+    "api, param, disabled_modules, expected_result, expected_error",
+    (
+        (
+            ("net_listening", (), (), True, None),
+            ("net_listening", (), (Admin,), True, None),
+            ("net_listening", (), (Net,), None, "Access of net module prohibited"),
+            ("net_listening", (), (Admin, Net,), None, "Access of net module prohibited"),
+        )
+    )
+)
+@pytest.mark.asyncio
+async def test_access_control(event_bus,
+                              api,
+                              param,
+                              disabled_modules,
+                              expected_result,
+                              expected_error):
+
+    chain = MainnetFullChain(None)
+    trinity_config = TrinityConfig(app_identifier="eth1", network_id=1)
+    rpc = RPCServer(initialize_eth1_modules(chain, event_bus, trinity_config), chain, event_bus)
+
+    request = build_request(api, param)
+    response = await rpc.execute_with_access_control(disabled_modules, request)
+    result, error = result_from_response(response)
+    assert result == expected_result
+    assert error == expected_error
--- a/trinity/components/builtin/json_rpc/component.py
+++ b/trinity/components/builtin/json_rpc/component.py
@@ -1,11 +1,14 @@
 from argparse import (
+    Action,
     ArgumentParser,
+    Namespace,
     _SubParsersAction,
 )
 import contextlib
-from typing import Iterator, Tuple, Union
+from typing import Iterator, Tuple, Union, Sequence, Type, Any
 
 from async_service import Service
+from eth_utils import ValidationError, to_tuple
 
 from lahja import EndpointAPI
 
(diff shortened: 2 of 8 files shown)


trinity/rpc/modules/main.py
```diff
@@ -26,12 +26,12 @@ def __init__(self, chain: TChain):
 
 class BaseRPCModule(ABC):
 
-    @property
-    def name(self) -> str:
+    @classmethod
+    def get_name(cls) -> str:
```
cburgdorf, line 30: It's beneficial to be able to read the name from the module class (as exposed to the instance only)

trinity/components/builtin/json_rpc/component.py
```diff
@@ -91,6 +93,32 @@ def chain_for_config(trinity_config: TrinityConfig,
         raise Exception("Unsupported Node Type")
 
 
+@to_tuple
+def get_http_enabled_modules(
+        enabled_modules: str,
+        available_modules: Sequence[BaseRPCModule]) -> Iterator[Type[BaseRPCModule]]:
+    all_module_types = set(type(mod) for mod in available_modules)
+    normalized_str = enabled_modules.lower().strip()
```
pipermerriam, line 101: Small nitpick but I think that you should move this logic out of this function to separate the normalization logic from the business logic.

I'd recommend moving the normalization logic into the CLI parser itself via a custom `Action` class [like this](https://github.com/ethereum/trinity/blob/fc6cea3b866924ef46b325f8874b4a1dd0d25bda/trinity/cli_parser.py#L26-L42)

tests/core/json-rpc/test_rpc_server.py
```diff
@@ -0,0 +1,52 @@
...
+from eth import MainnetChain
+
+from trinity.chains.full import FullChain
+from trinity.config import TrinityConfig
+from trinity.rpc import RPCServer
+from trinity.rpc.modules import initialize_eth1_modules, Admin, Net
+
+
+def build_request(method, params):
+    return {"jsonrpc": "2.0", "method": method, "params": params, "id": 3}
```
gsalgado, line 13: Why is `id` hardcoded to 3? Can we use a random value instead?

trinity/rpc/main.py
```diff
@@ -147,9 +156,25 @@ async def _get_result(self,
         else:
             return result, None
 
-    async def execute(self, request: Dict[str, Any]) -> str:
+    async def execute(self,
```
gsalgado, line 159: Is this still used anywhere? Maybe we could rename it to `_execute()` if it's not used externally?

cburgdorf: Yes, the `IPCServer` still uses `execute` as we don't do access restriction via IPC.

tests/core/json-rpc/test_rpc_server.py
cburgdorf: I just copied that from `test_rpc_fixtures` but we can certainly use a random value :+1:

trinity/components/builtin/json_rpc/component.py
cburgdorf: In the code you linked, the normalization is self contained (turning strings into enodes) but in this case we want to turn a string into a subset of classes that wouldn't be available at normalization time (the `available_modules`).

We *could* do a small normalization step to normalize the string into something like a `Union[str, Sequence[str]` where e.g. `" * "` is turned into `"*"` and `"net,   admin,"` is turned into `("net", "admin")` but I feel that is little gain, if any.

We would still need to take the `Union[str, Sequence[str]` and turn it into a `Sequence[Type[BaseRPCModule]]` at the point where we have access to all the `BaseRPCModule` instances.

But I'm happy to address any points post-merge if maybe I misunderstood or not seeing a good refactoring opportunity!

pipermerriam: > We *could* do a small normalization step to normalize the string into something like a `Union[str, Sequence[str]` where e.g. `" * "` is turned into `"*"` and `"net,   admin,"` is turned into `("net", "admin")` but I feel that is little gain, if any.

This is what I was trying to get at.  It's a small gain.

I think that you can still force it to be `Sequence[str]` and represent `"*"` as `("*",)` which still allows you to do an easy check for the `*` special case.

```python
if tuple(seq) == ("*",):
    ...
```

Acknowleding this is a *nitpick* however it does separate the maybe slightly more error prone string manipulation logic from the less error prone name matching logic.

cburgdorf: Alright, took it :+1:

Issue #1908, opened 2020-07-17
Write integration tests for metrics component


cburgdorf:
### What is wrong?

Starting Trinity with metrics enabled is currently broken (investigation ongoing at this point) and the fact we didn't notice is because we don't have any tests for that area of the code base.

### How can it be fixed

I could imagine two different scenarios for integration tests:

A.) We run a test that spins up Trinity to run against our deployed metrics infrastructure. We would put the credentials in CircleCI environment variables.

PRO: - easy
CON: - could be flaky if the metrics infrastructure or connection to it is unreliable (cough cough, seems offline at the time of writing)

B.) The test spins up its own InfluxDB locally (there may be a ready to use docker image for that to become easier)

PRO: -less flaky
CON: - more effort

cburgdorf:
@njgheorghita Possibly something you could be interested to tackle

njgheorghita:
@cburgdorf Taking the fact that trinity-bench just crashed as a sign, I'd agree that #2 is probably the better solution. In #1903 - do you know approximately how long the node was running before the error occurred? Was it instant? AKA in this test, how long should the node report metrics for to more-or-less conclude that everything works as expected?

cburgdorf:
>AKA in this test, how long should the node report metrics for to more-or-less conclude that everything works as expected?

I would say the test is successful as soon as we were able to read written metrics out of the influxdb. I would say we can start simple with only checking for values that are known to be written immediately such as the system metrics.

PR #1904, opened 2020-07-16
Handle error in metrics reporting

cburgdorf:
### What was wrong?

Currently, Trinity will crash if metrics can't be written to the InfluxDB because of some connection / http errors.

### How was it fixed?

Catch `ConnectionError` and `HttpError` and report them as warnings. The fix also ensures we do at most log one warning every minute.

Fixes #1903

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [x] Clean up commit history

[//]: # (For important changes that should go into the release notes please add a newsfragment file as explained here: https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)

[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)
- [x] Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #1914, opened 2020-07-20
Add integration test for enabled metrics

njgheorghita:
### What was wrong?
Fixes https://github.com/ethereum/trinity/issues/1908

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/.circleci/build_influxdb.sh
+++ b/.circleci/build_influxdb.sh
@@ -4,7 +4,7 @@ sudo apt-get install influxdb;
 sudo apt-get install influxdb-client;
 (sudo influxd -config /etc/influxdb/influxdb.conf) &
 # wait until influxdb service is actually available on port 8086
-(while netstat -lnt | awk '$4 ~ /:8086$/ {exit 1}'; do sleep 5; done;
+(while netstat -lnt | awk '$4 ~ /:8086$/ {exit 1}'; do sleep 1; done;
 sudo influx --execute "CREATE DATABASE trinity";
 sudo influx --execute "USE trinity";
 sudo influx --execute "CREATE USER \"trinity\" WITH PASSWORD 'trinity' WITH ALL PRIVILEGES") &
--- a/.circleci/build_influxdb.sh
+++ b/.circleci/build_influxdb.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+sudo apt-get install influxdb;
+sudo apt-get install influxdb-client;
+(sudo influxd -config /etc/influxdb/influxdb.conf) &
+# wait until influxdb service is actually available on port 8086
+(while netstat -lnt | awk '$4 ~ /:8086$/ {exit 1}'; do sleep 5; done;
+sudo influx --execute "CREATE DATABASE trinity";
+sudo influx --execute "USE trinity";
+sudo influx --execute "CREATE USER \"trinity\" WITH PASSWORD 'trinity' WITH ALL PRIVILEGES") &
+(sleep infinity;)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,6 +71,49 @@ geth_steps: &geth_steps
           - ~/.py-geth
         key: cache-v2-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}-{{ checksum "./.circleci/build_geth.sh" }}
 
+wheel_cli_steps: &wheel_cli_steps
+  working_directory: ~/repo
+  steps:
+    - checkout
+    - run:
+        name: checkout fixtures submodule
+        command: git submodule update --init --recursive
+    - run:
+        name: merge pull request base
+        command: ./.circleci/merge_pr.sh
+    - run:
+        name: merge pull request base (2nd try)
+        command: ./.circleci/merge_pr.sh
+        when: on_fail
+    - run:
+        name: merge pull request base (3nd try)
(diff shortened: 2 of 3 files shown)


njgheorghita:
@cburgdorf This is still more of a draft stage, but I'm ready for input if you have any thoughts. A couple questions on my end...
- Is there any existing mechanism to restrict the test for metrics-enabled to only run in the circle build? It seems unrealistic to require users to have an influxdb running locally.
- The latest `web3` version `5.12` pins `jsonschema>=3.20` but in `trinity`, `jsonschema==3.0.1`. Is that something that should/can be handled in this pr or separately?
- Are the influxdb install/setup commands better suited inside their own script rather than `config.yaml`? like `build_geth.sh`?

.circleci/config.yml
```diff
@@ -27,6 +27,21 @@ common: &common
...
+            sudo apt-get install influxdb-client
+            sudo influxd -config /etc/influxdb/influxdb.conf
+    - run:
+        name: setup trinity on influxdb
+        command: |
+            # wait until influxdb service is actually available on port 8086
+            while netstat -lnt | awk '$4 ~ /:8086$/ {exit 1}'; do sleep 10; done
+            sudo influx --execute "CREATE DATABASE trinity"
+            sudo influx --execute "USE trinity"
+            sudo influx --execute "CREATE USER \"trinity\" WITH PASSWORD 'trinity' WITH ALL PRIVILEGES"
```
cburgdorf, line 44: Yeah, as you mentioned in the issue, I think this should follow the pattern of `/build_geth.sh` and `geth_steps: &geth_steps` to not consume unnecessary time for every job that doesn't need influx to be setup.

tests/integration/test_trinity_cli.py
```diff
@@ -284,6 +285,32 @@ async def test_does_not_throw_errors_on_short_run(command, unused_tcp_port):
...
+    )
+    command = amend_command_for_unused_port(enabled_metrics_command, unused_tcp_port)
+
+    import time
+    async with AsyncProcessRunner.run(command, timeout_sec=120) as runner:
+        assert await contains_all(runner.stderr, {
+            'MetricsService',
+            'Reporting metrics to localhost',
+        })
+        time.sleep(10)
```
cburgdorf, line 305: Instead of the long sleep and if we can't find a good anchor to simply wait on the `contains_all` it may be better to poll repetitively for the expected values (with a small sleep in between) until we hit a timeout. That timeout could be longer like 20 or 30 seconds but in the happy case we would finish the test much sooner in the least amount of time needed.

cburgdorf:
Left comments inline. The `jsonschema` issue is already fixed in latest master.

njgheorghita:
Thanks @cburgdorf ! This is ready for another look-over when you get the chance

.circleci/build_influxdb.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+sudo apt-get install influxdb;
+sudo apt-get install influxdb-client;
+(sudo influxd -config /etc/influxdb/influxdb.conf) &
+# wait until influxdb service is actually available on port 8086
+(while netstat -lnt | awk '$4 ~ /:8086$/ {exit 1}'; do sleep 5; done;
```
cburgdorf, line 7: Wondering if the sleep could be reduced from `5` to `1` to slightly speed things up.

tests/integration/test_trinity_cli.py
```diff
@@ -284,6 +285,37 @@ async def test_does_not_throw_errors_on_short_run(command, unused_tcp_port):
...
+        '--metrics-influx-password=trinity'
+    )
+    command = amend_command_for_unused_port(enabled_metrics_command, unused_tcp_port)
+
+    async with AsyncProcessRunner.run(command, timeout_sec=120) as runner:
+        assert await contains_all(runner.stderr, {
+            'MetricsService',
+            'Reporting metrics to localhost',
+        })
+        t1 = time.time()
```
cburgdorf, line 304: We generally try to avoid super short acronym variable names. Maybe `start_time` would be a little clearer.

tests/integration/test_trinity_cli.py
```diff
@@ -284,6 +285,37 @@ async def test_does_not_throw_errors_on_short_run(command, unused_tcp_port):
...
+                'http://localhost:8086/query?db=trinity&epoch=ns&q=select+%2A'
+                '+from+%22trinity.p2p%2Fpeers.counter%22'
+            )
+            json_response = r.json()
+            if 'series' in json_response['results'][0]:
+                break
+        else:
+            raise Exception('Influxdb request timeout')
+
+        assert len(json_response['results'][0]['series'][0]['values']) > 0
```
cburgdorf, line 316: I'm wondering if this is too broad and doesn't actually test if Trinity has written anything? Because the metrics system is broken in current master (as in, Trinity doesn't even start with metrics enabled) and hence I would assume that this test should currently *not* succeed. The fact that it succeeds, suggests it isn't actually protecting us from the issues we want it to protect us from.

This commit is needed to fix the metrics system https://github.com/ethereum/trinity/pull/1904/commits/3fb6e883a7febf08de4249195cb42620ae7d3cb3

njgheorghita:
@cburgdorf This one slipped through the cracks on my end, but I've updated the test to be time-bound so that we explicitly require that the influxdb is populated within the duration of the test. This test fails if I omit running metrics, so it seems to be testing appropriately now.

cburgdorf (approved):
LGTM

PR #1916, opened 2020-07-20
Longer queen penalty & logging tweaks

carver:
### What was wrong?

If your queen peer is fast at answering data, but giving empty responses, we penalize it. If RTT is ~0.5s, then a 0.5s penalty means that we are still using it as a queen about every other request.

### How was it fixed?

Bump up the penalty to ~2s. A large penalty was more painful when we had a hard time connecting to peers, but now shouldn't have too big an impact.

Also, logging more info about time spent waiting on the missing trie nodes, and intermediate progress during long beam imports.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/importer.py
+++ b/trinity/sync/beam/importer.py
@@ -2,10 +2,12 @@ from abc import abstractmethod
 import asyncio
 from concurrent import futures
 from operator import attrgetter
+import time
 from typing import (
     Any,
     Callable,
     Optional,
+    Sequence,
     Tuple,
     Type,
     TypeVar,
@@ -19,7 +21,9 @@ from eth.abc import (
     BlockAPI,
     BlockHeaderAPI,
     BlockImportResult,
+    ComputationAPI,
     ConsensusContextAPI,
+    ReceiptAPI,
     SignedTransactionAPI,
     StateAPI,
     VirtualMachineAPI,
@@ -52,6 +56,7 @@ from trinity.chains.full import FullChain
 from trinity.exceptions import StateUnretrievable
 from trinity.sync.beam.constants import (
     MAX_SPECULATIVE_EXECUTIONS_PER_PROCESS,
+    MIN_GAS_LOG_WAIT,
     NUM_PREVIEW_SHARDS,
 )
 from trinity.sync.common.events import (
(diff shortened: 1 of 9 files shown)


carver:
@cburgdorf `docker-trinity-beacon-image-build-test` is failing because it clones from master instead of the branch (I think it will be fixed by 9e9fe18) . I am about to wrap up my day, but this red mark is going to block all CI runs until it's fixed. Would you mind taking a look?

> Step 6/10 : RUN git clone https://github.com/ethereum/trinity.git .

trinity/sync/beam/importer.py
```diff
@@ -383,13 +410,77 @@ def make_state_root(self) -> Optional[Any]:
...
 
         def get_beam_stats(self) -> BeamStats:
             return self.state.stats_counter
 
+        def apply_all_transactions(
+            self,
+            transactions: Sequence[SignedTransactionAPI],
+            base_header: BlockHeaderAPI
+        ) -> Tuple[BlockHeaderAPI, Tuple[ReceiptAPI, ...], Tuple[ComputationAPI, ...]]:
+            # TODO find a better hook for this, instead of the copy/pasta from py-evm
```
cburgdorf, line 427: Yeah, the divergence from Py-EVM seems problematic. Either we find good hooks or eventually be able to contribute changes back into Py-EVM to remove it from Trinity back again. Might be a good idea to track this in an issue to not forget about it.

cburgdorf:
I cherry-picked the CI fix into https://github.com/ethereum/trinity/pull/1921 and merged it. CI looks good now :+1:

trinity/sync/beam/importer.py
```diff
@@ -383,13 +410,77 @@ def make_state_root(self) -> Optional[Any]:
...
+
+            receipts = []
+            computations = []
+            previous_header = base_header
+            result_header = base_header
+
+            # Modified from py-evm: count and enumerate transactions, get expected header
+            num_transactions = len(transactions)
+            expected_header = self.get_header()
+            self.last_log_time = time.monotonic()
```
pipermerriam, line 443: Does this really need to be a property on `self`?  It looks like it would be fine as just a transient local variable.

trinity/sync/beam/importer.py
pipermerriam: I assume you can't do the logging inside `apply_transaction` because you are missing the contextual information like the transaction index and the amount of gas used in total?

pipermerriam: I'm not opposed to us doing something simple like:

```python
class VM:
    def after_transaction_apply_hook(self, ...):
        ...
```

Just exposing simple function hooks in places where we find we have need of injecting logic.

trinity/sync/beam/importer.py
carver: With the new hook style, I think it does

carver:
Red CI seems to be #1919

PR #1917, opened 2020-07-20
Eth2 optimizations at the database layer

ralexstokes:
This PR is exploring optimizations at the database layer we can use to speed up sync.

It has yielded other PRs as it explores other optimizations.

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/chains/testnet/altona.py
+++ b/eth2/beacon/chains/testnet/altona.py
@@ -122,7 +122,7 @@ class BeaconChain(BaseBeaconChain):
         assert genesis_state.slot == GENESIS_SLOT
 
         chain_db = BeaconChainDB.from_genesis(
-            base_db, genesis_state, signed_block_class
+            base_db, genesis_state, signed_block_class, config
         )
 
         block_sink = ChainDBBlockSink(chain_db)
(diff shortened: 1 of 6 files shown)


eth2/beacon/committee_helpers.py
```diff
@@ -170,6 +174,8 @@ def _compute_committee(
         yield indices[shuffled_index]
 
 
+# NOTE: cache of 1024 covers "worst case" of every committee getting an attestation on-chain at 4mm ETH at stake.
```
pipermerriam, line 177: These are really useful comments.  Wondering if you can add similar ones to the other functions to make it clearer why each value was chosen.

eth2/beacon/state_machines/forks/serenity/epoch_processing.py
```diff
@@ -404,12 +403,15 @@ def process_rewards_and_penalties(
...
 
-    for index in range(len(state.validators)):
-        index = ValidatorIndex(index)
-        state = increase_balance(state, index, Gwei(rewards_for_attestations[index]))
-        state = decrease_balance(state, index, Gwei(penalties_for_attestations[index]))
+    new_balances = ()
+    for balance, reward, penalty in zip(
+        state.balances, rewards_for_attestations, penalties_for_attestations
+    ):
+        new_balances += (max(balance + reward - penalty, 0),)
```
pipermerriam, line 410: I think you can save yourself a little object allocation via a tuple comprehension:

```python
new_balances = tuple(
    max(balance + reward - penalty, 0)
    for balance, reward, penalty in zip(...)
)
```

eth2/beacon/committee_helpers.py
ralexstokes: agreed -- not sure these are staying in but if so then i'll add more context

PR #1918, opened 2020-07-20
State transition with eth2fastspec

g-r-a-n-t:
### What was wrong?

The state transition is a bit slow and can be optimized using https://github.com/protolambda/eth2fastspec .

### How was it fixed?

Copied in eth2fastspec and made it compatible with our beacon types.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

racoon


eth2/beacon/chains/testnet/altona.py
```diff
@@ -400,4 +400,4 @@ def on_attestation(self, attestation: Attestation) -> None:
 
 
 class BeaconChainTest(BeaconChain):
-    _sm_configuration = ((GENESIS_SLOT, AltonaStateMachineTest),)
+    _sm_configuration = ((GENESIS_SLOT, AltonaStateMachineTest),)  # type: ignore
```
ralexstokes, line 403: why `# type: ignore`?

eth2/beacon/state_machines/forks/altona/eth2fastspec.py
```diff
@@ -0,0 +1,1836 @@
...
+    """
+    return state.randao_mixes[epoch % EPOCHS_PER_HISTORICAL_VECTOR]
+
+
+def int_to_bytes(n: int, length: int) -> bytes:
+    """
+    Return the ``length``-byte serialization of ``n`` in ``ENDIANNESS``-endian.
+    """
+    return n.to_bytes(length=length, byteorder=ENDIANNESS)
+
```
ralexstokes, line 955: look like duplicates

eth2/beacon/state_machines/forks/altona/eth2fastspec.py
```diff
@@ -0,0 +1,1836 @@
...
+        penalty_numerator = effective_balance // increment * slashings_scale
+        penalty = Gwei(penalty_numerator // total_balance * increment)
+        state = decrease_balance(state, index, penalty)
+
+    return state
+
+
+HYSTERESIS_INCREMENT = EFFECTIVE_BALANCE_INCREMENT // HYSTERESIS_QUOTIENT
+DOWNWARD_THRESHOLD = HYSTERESIS_INCREMENT * HYSTERESIS_DOWNWARD_MULTIPLIER
+UPWARD_THRESHOLD = HYSTERESIS_INCREMENT * HYSTERESIS_UPWARD_MULTIPLIER
```
ralexstokes, line 1270: should be derived from config

eth2/beacon/state_machines/forks/altona/eth2fastspec.py
```diff
@@ -0,0 +1,1836 @@
...
+    # Verify that the slots match
+    assert block.slot == state.slot
+    # Verify that the block is newer than latest block header
+    assert block.slot > state.latest_block_header.slot
+    # Verify that proposer index is the correct index
+    proposer_index = epochs_ctx.get_beacon_proposer(state.slot)
+    assert block.proposer_index == proposer_index
+    # Verify that the parent matches
+    assert block.parent_root == state.latest_block_header.hash_tree_root
+    # Cache current block as the new latest block
```
ralexstokes, line 1340: so we generally take a different approach to the spec when handling validations....

e.g. refer all of the `ValidationError` in here: https://github.com/ethereum/trinity/blob/master/eth2/beacon/state_machines/forks/serenity/block_validation.py

the main advantage i see there is being able to detect what exactly has gone wrong...

here we will just get an `AssertionErrror` and then not directly know why...

this could be something to address in a follow-up PR but then we should move this to an issue

eth2/beacon/chains/testnet/altona.py
g-r-a-n-t: The change that caused a type error here was https://github.com/ethereum/trinity/pull/1918/files#diff-d72273019287d30daf35f55acd671fd3L85

It was fine before since `AltonaStateMachineTest` inherits `AltonaStateMachine`, but since changing the Altona chain's state machine class to `AltonaStateMachineFast` that inheritance no longer holds. There may be cleaner ways around this, but figured since the Test class was only used in testing, it was fine.

eth2/beacon/state_machines/forks/altona/eth2fastspec.py
```diff
@@ -0,0 +1,1836 @@
...
+    """
+    return (
+        # Double vote
+        (data_1 != data_2 and data_1.target.epoch == data_2.target.epoch)
+        # Surround vote
+        or (
+            data_1.source.epoch < data_2.source.epoch
+            and data_2.target.epoch < data_1.target.epoch
+        )
+    )
```
ralexstokes, line 1445: these look like re-definitions

eth2/beacon/state_machines/forks/altona/eth2fastspec.py
```diff
@@ -0,0 +1,1836 @@
...
+    """
+    Check if ``leaf`` at ``index`` verifies against the Merkle ``root`` and ``branch``.
+    """
+    value = leaf
+    for i in range(depth):
+        if (index >> i) & 1 == 1:
+            value = hash(branch[i] + value)
+        else:
+            value = hash(value + branch[i])
+    return value == root
```
ralexstokes, line 1678: looks like a redefinition... or close to one

eth2/beacon/state_machines/forks/altona/eth2fastspec.py
```diff
@@ -0,0 +1,1836 @@
...
+    i = 0
+    while True:
+        candidate_index = indices[
+            compute_shuffled_index(ValidatorIndex(i % len(indices)), len(indices), seed)
+        ]
+        random_byte = hash(seed + int_to_bytes(i // 32, length=8))[i % 32]
+        effective_balance = state.validators[candidate_index].effective_balance
+        if effective_balance * MAX_RANDOM_BYTE >= MAX_EFFECTIVE_BALANCE * random_byte:
+            return ValidatorIndex(candidate_index)
+        i += 1
```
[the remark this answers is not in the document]
g-r-a-n-t: it looks like ours is a bit slower, let me look into more

eth2/beacon/state_machines/forks/altona/eth2fastspec.py
```diff
@@ -0,0 +1,1836 @@
...
+
+
+def get_churn_limit(active_validator_count: int) -> int:
+    return max(
+        MIN_PER_EPOCH_CHURN_LIMIT, active_validator_count // CHURN_LIMIT_QUOTIENT
+    )
+
+
+def is_active_validator(v: Validator, epoch: Epoch) -> bool:
+    return v.activation_epoch <= epoch < v.exit_epoch
```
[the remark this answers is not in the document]
g-r-a-n-t: Our `get_churn_limit` is a bit different, as it takes in a state and ends up iterating over each validator in `get_active_validator_indices` to see if they are active. The optimized version takes in an active validator count, which is basically cached in `epochs_ctx`.

Issue #1919, opened 2020-07-21
asyncio Task was destroyed but it is pending!


carver:
### What is wrong?

I've seen this locally and now in CI: https://app.circleci.com/pipelines/github/ethereum/trinity/6944/workflows/7fe39869-557d-4b0b-b7f9-0f2091e3fb73/jobs/263262/steps

```
   DEBUG  07-21 00:26:57  async_process_runner.py  b'\x1b[1m\x1b[31m   ERROR  2020-07-21 00:26:57,347               asyncio  Task was destroyed but it is pending!\n'
   DEBUG  07-21 00:26:57  async_process_runner.py  b'task: <Task pending coro=<wait_first() running at /home/circleci/repo/p2p/logic.py:91> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f9a09bdecd0>()]>>\x1b[0m\n'
   DEBUG  07-21 00:26:57  async_process_runner.py  b'\x1b[1m\x1b[31m   ERROR  2020-07-21 00:26:57,347               asyncio  Task was destroyed but it is pending!\n'
   DEBUG  07-21 00:26:57  async_process_runner.py  b'task: <Task pending coro=<wait_first() running at /home/circleci/repo/p2p/logic.py:91> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f9a09bdee50>()]>>\x1b[0m\n'
   DEBUG  07-21 00:26:57  async_process_runner.py  b'\x1b[1m\x1b[31m   ERROR  2020-07-21 00:26:57,347               asyncio  Task was destroyed but it is pending!\n'
   DEBUG  07-21 00:26:57  async_process_runner.py  b'task: <Task pending coro=<wait_first() running at /home/circleci/repo/p2p/logic.py:91> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f9a09bea950>()]>>\x1b[0m\n'
   DEBUG  07-21 00:26:57  async_process_runner.py  b'\x1b[1m\x1b[31m   ERROR  2020-07-21 00:26:57,347               asyncio  Task was destroyed but it is pending!\n'
[...]

PR #1924, opened 2020-07-22
Ensure Connection.run_behaviors() doesn't leak pending asyncio tasks

gsalgado:
If one of our behaviors raised an exception in their post_apply(),
run_behaviors() would leak pending asyncio tasks, causing asyncio
to complain. This fixes that by ensuring all tasks are done/cancelled

Closes: #1919

### DIFF
--- a/p2p/asyncio_utils.py
+++ b/p2p/asyncio_utils.py
@@ -1,0 +1,74 @@
+from __future__ import annotations
+import asyncio
+import contextlib
+import sys
+from typing import Awaitable, Sequence, Iterable, List, TypeVar
+
+from trio import MultiError
+
+from eth_utils.toolz import first
+
+
+TReturn = TypeVar('TReturn')
+
+
+def create_task(coro: Awaitable[TReturn], name: str) -> asyncio.Task[TReturn]:
+    if sys.version_info >= (3, 8):
+        # Our version of mypy doesn't know about the name argument to create_task()
+        return asyncio.create_task(coro, name=name)  # type: ignore
+    else:
+        return asyncio.create_task(coro)
+
+
+async def wait_first(futures: Sequence[asyncio.Future[None]]) -> None:
+    """
+    Wait for the first of the given futures to complete, then cancels all others.
+
+    If the completed future raised an exception, re-raise it.
+
+    If the task running us is cancelled, all futures will be cancelled.
+    """
+    for future in futures:
+        if not isinstance(future, asyncio.Future):
+            raise ValueError(f"{future} is not an asyncio.Future")
+
+    try:
+        done, pending = await asyncio.wait(futures, return_when=asyncio.FIRST_COMPLETED)
+    except asyncio.CancelledError:
+        await cancel_futures(futures)
+        raise
+    else:
+        if pending:
+            await cancel_futures(pending)
+        if len(done) != 1:
+            raise Exception(
+                "Invariant: asyncio.wait() returned more than one future even "
+                "though we used return_when=asyncio.FIRST_COMPLETED: %s", done)
+        done_future = first(done)
+        if done_future.exception():
+            raise done_future.exception()
+
+
+async def cancel_futures(futures: Iterable[asyncio.Future[None]]) -> None:
+    """
+    Cancel and await for the given futures, ignoring any asyncio.CancelledErrors.
+    """
+    for fut in futures:
+        fut.cancel()
+
+    errors: List[BaseException] = []
+    # Wait for all futures in parallel so if any of them catches CancelledError and performs a
+    # slow cleanup the othders don't have to wait for it. The timeout is long as our component
+    # tasks can do a lot of stuff during their cleanup.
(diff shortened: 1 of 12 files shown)


gsalgado:
So, we need the `type: ignore` comment to please mypy on python3.8, but that upsets it on python3.7. The only way I could find to keep both happy is to set `warn_unused_ignores = False` in mypy.ini

p2p/asyncio_utils.py
```diff
@@ -0,0 +1,67 @@
...
+import sys
+from typing import Any, Awaitable, Sequence, Iterable, List
+
+from trio import MultiError
+
+
+def create_task(coro: Awaitable[Any], name: str) -> asyncio.Task[Any]:
+    if sys.version_info >= (3, 8):
+        # Our version of mypy doesn't know about the name argument to create_task()
+        return asyncio.create_task(coro, name=name)  # type: ignore
```
gsalgado, line 13: So, we need the type: ignore comment to please mypy on python3.8, but that upsets it on python3.7. The only way I could find to keep both happy is to set warn_unused_ignores = False in mypy.ini

carver: Would a more recent version of mypy work better? (obviously don't want to hold this PR up for long, since it's fixing a flaky test, but it seems worth checking if it's a quick fix or doing a quick follow up)

gsalgado: Latest mypy does, but it crashes badly on some of our code:

```
(trinity) salgado@t495~/src/snakecharmers/trinity (issue-1919) $ mypy -p scripts -p trinity -p p2p --config-file /home/salgado/src/snakecharmers/trinity/mypy.ini 
p2p/ecies.py:48: error: "EllipticCurvePrivateKey" has no attribute "private_numbers"
p2p/ecies.py:150: error: "SHA256" has no attribute "block_size"
trinity/components/eth2/eth1_monitor/events.py:25: error: Access to generic instance variables via class is ambiguous
/home/salgado/src/snakecharmers/trinity/trinity/db/orm.py:27: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-a-development-mypy-build
If this issue continues with mypy master, please report a bug at https://github.com/python/mypy/issues
version: 0.782
/home/salgado/src/snakecharmers/trinity/trinity/db/orm.py:27: : note: please use --show-traceback to print a traceback when reporting a bug
```

gsalgado: Turns out it was a bug in the sqlmypy plugin, fixed in the latest version, but the latest mypy complains about a bunch of other things in our code, so I'll submit a separate PR to update mypy which will then unblock this

p2p/asyncio_utils.py
```diff
@@ -0,0 +1,67 @@
...
+        await cancel_futures(futures)
+        raise
+    else:
+        if pending:
+            await cancel_futures(pending)
+        if len(done) != 1:
+            raise Exception(
+                "Invariant: asyncio.wait() returned more than one future even "
+                "though we used return_when=asyncio.FIRST_COMPLETED: %s", done)
+        done_future = list(done)[0]
```
carver, line 42: *nit* `toolz.first()` seems to be a clearer way to grab the first element (I know this was just copy/pasted, but it's an easy change).

p2p/asyncio_utils.py
```diff
@@ -0,0 +1,67 @@
...
+
+def create_task(coro: Awaitable[Any], name: str) -> asyncio.Task[Any]:
+    if sys.version_info >= (3, 8):
+        # Our version of mypy doesn't know about the name argument to create_task()
+        return asyncio.create_task(coro, name=name)  # type: ignore
+    else:
+        return asyncio.create_task(coro)
+
+
+async def wait_first(futures: Sequence[asyncio.Future[None]]) -> None:
```
carver, line 18: I didn't notice any changes in this method, but it's harder to tell because it moved files. Anything worth noting?

p2p/asyncio_utils.py
carver: FWIW, I'm in favor of setting `warn_unused_ignores = False` to get this merged, then doing the mypy PR as an immediate follow-up PR to re-enable the setting.

p2p/asyncio_utils.py
```diff
@@ -0,0 +1,67 @@
+from __future__ import annotations
+import asyncio
+import contextlib
+import sys
+from typing import Any, Awaitable, Sequence, Iterable, List
+
+from trio import MultiError
+
+
+def create_task(coro: Awaitable[Any], name: str) -> asyncio.Task[Any]:
```
pipermerriam, line 10: Should the `Any` here be a `TypeVar('TReturn')` to preserve the return type?

p2p/asyncio_utils.py
gsalgado: No changes indeed, just moved

gsalgado:
Followup PR to re-enable warn_unused_ignores: #1932

PR #1930, opened 2020-07-22
Some baseline eth2 performance optimizations

ralexstokes:
### What was wrong?

We did not cache things that should definitely be cached to avoid recomputing values during an epoch transition.

We also had some obviously inefficient code given the size of the validator set on mainnet-like configurations.

### How was it fixed?

1. Add basic `functools.lru_cache` caching to the functions that do not change over a single epoch.
2. Skip the "immutability penalty" when making a large number of changes to a list of balances.

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/committee_helpers.py
+++ b/eth2/beacon/committee_helpers.py
@@ -1,3 +1,4 @@
+from functools import lru_cache
 import logging
 from typing import Iterable, Sequence, Tuple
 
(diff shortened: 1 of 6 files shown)


pipermerriam:
curious if there is a benchmark suite you guys are using for this?

ralexstokes:
mainly watching sync performance with "slots / second" as a metric.

i'm using the same setup for each measurement so feel confident the numbers are meaningful.

we could easily write a benchmark suite that e.g. loads a bunch of blocks from disk and then processes them but haven't yet.

PR #1932, opened 2020-07-22
Update mypy to 0.782

gsalgado:


gsalgado:
@carver, @pipermerriam, @cburgdorf any idea what we need to please mypy here: https://app.circleci.com/pipelines/github/ethereum/trinity/7023/workflows/d1bb3db1-7373-455c-b4dc-7032a9e1ffa6/jobs/266992

pipermerriam:
@gsalgado yes:

see: https://github.com/ethereum/trinity/blob/691357b2d072017f1a07253c6deb90156e7d8e20/trinity/protocol/les/exchanges.py#L45-L49

We're passing a type as an argument.  Those lines should be something like:

```python
### Wrong
    _normalizer = DefaultNormalizer(
        BlockHeaders,
        Tuple[BlockHeaderAPI, ...],   # note that this is not a type hint...
        normalize_fn=lambda res: res.payload.headers
    )
### Maybe right
    _normalizer = DefaultNormalizer(
        BlockHeaders,
        tuple,   # I have no idea if this is correct but maybe in the right direction...
        normalize_fn=lambda res: res.payload.headers
    )
```

pipermerriam:
@gsalgado 

This comment seems relevant:  https://github.com/ethereum/trinity/blob/691357b2d072017f1a07253c6deb90156e7d8e20/p2p/exchange/normalizers.py#L28-L31

Maybe this used to work in old mypy but new mypy doesn't like it.

gsalgado:
So, needed to add a `type: ignore` in a couple places, but I don't think they're a big deal so we can probably figure them out later

p2p/ecies.py
```diff
@@ -147,7 +150,9 @@ def kdf(key_material: bytes) -> bytes:
     https://github.com/ethereum/go-ethereum/blob/673007d7aed1d2678ea3277eceb7b55dc29cf092/crypto/ecies/ecies.go#L167
     """
     key = b""
-    hash_blocksize = hashes.SHA256().block_size
+    hash_ = hashes.SHA256()
+    # FIXME: Need to find out why mypy thinks SHA256 has no 'block_size' attribute
+    hash_blocksize = hash_.block_size  # type: ignore
```
gsalgado, line 155: For this one, our previous version of mypy would treat the return of `hashes.SHA256()` as `Any`, and that's why it didn't complain

trinity/components/eth2/eth1_monitor/events.py
```diff
@@ -22,7 +22,9 @@ class SSZSerializableEvent(BaseEvent, Generic[TData]):
 
     @classmethod
     def from_data(cls: Type[T], data: TData) -> T:
-        return cls(sedes=cls.sedes, data_bytes=ssz.encode(data))
+        # FIXME: Find out why mypy complains with the following error here:
+        #   Access to generic instance variables via class is ambiguous
```
gsalgado, line 26: This one I couldn't figure out

eth2/_utils/typing.py
```diff
@@ -1,13 +1,13 @@
...
 from typing import Any, TypeVar
 
 from typing_extensions import Protocol
 
 ComparableType = TypeVar("ComparableType", bound="Comparable")
 
 
-@functools.total_ordering
+# Can't use total_ordering here: https://github.com/python/mypy/issues/8539
+# @functools.total_ordering
```
cburgdorf, line 10: I don't see it being moved anywhere...so what happened to the functionality. Do we just not need it?

p2p/asyncio_utils.py
```diff
@@ -14,13 +14,12 @@
...
-        return asyncio.create_task(coro, name=name)  # type: ignore
+        return asyncio.create_task(coro, name=name)
     else:
         return asyncio.create_task(coro)
 
 
-async def wait_first(futures: Sequence[asyncio.Future[None]]) -> None:
+async def wait_first(futures: Sequence[asyncio.Task[Any]]) -> None:
     """
     Wait for the first of the given futures to complete, then cancels all others.
```
cburgdorf, line 24: I'm the first to admit I can't explain the difference between `task` and `future` without looking it up every single time again but maybe we should also update the docstring to mention `task` instead of `future`.

eth2/_utils/typing.py
gsalgado: I guess not. @ralexstokes?

trinity/components/eth2/eth1_monitor/events.py
carver: Looks like the type annotation for `sedes` should be `sedes: ClassVar[Type[TData]]`, to permit a class-level lookup:

https://github.com/python/mypy/issues/3172#issuecomment-294112713

gsalgado: That seems to cause mypy to not accept a sedes argument to the constructor: 
> trinity/components/eth2/eth1_monitor/events.py:27: error: Unexpected keyword argument "sedes" for "SSZSerializableEvent"

I've moved the `cls.sedes` to a line by itself, along with the `type: ignore` to

eth2/_utils/typing.py
ralexstokes: it is being used in what will likely be a deprecated version of the eth2 fork choice...

that is kind of annoying but if the tests pass w/out it then we can remove it for now

PR #1934, opened 2020-07-23
Spread Beam: Ask more peers for the same urgent trie node, to reduce latency

### LINKED ISSUE #1101: Beam Sync: Reduce sensitivity to Queen Peer latency
### What is wrong?

Occasionally, we get stuck in a situation where 2 or more peers have similar performance, and/or highly fluctuating performance. We switch the "queen peer" (best performing) back and forth, which can sometimes cause delays. This seems to cause more of a slowdown than expected (and I *don't* see a lot of "Want to download urgent data, but..." messages).

In general, it's a problem that Beam Sync is so sensitive to quickly and accurately identifying the best peer.

### How can it be fixed

We can reduce fluttering by forcing the new queen to beat the old queen by at least 5% or something.

Worth looking further into why this causes such a performance hit. (Or it might have been a coincidence, that I got a huge block when this fluttering happened)

carver:
### What was wrong?

Fixes #1101 

### How was it fixed?

Added a new protected category of peer besides the queen and peasants, called "knights". These are reserved for use, in this case, to ask for the same urgent node that we ask the queen for.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/state.py
+++ b/trinity/sync/beam/state.py
@@ -19,6 +19,7 @@ from lahja import EndpointAPI
 
 from eth_hash.auto import keccak
 from eth_utils import (
+    clamp,
     to_checksum_address,
     ValidationError,
 )
@@ -30,6 +31,10 @@ from eth_typing import (
 from eth.abc import AtomicDatabaseAPI
 
 from p2p.abc import CommandAPI
+from p2p.asyncio_utils import (
+    cleanup_tasks,
+    create_task,
+)
 from p2p.exceptions import BaseP2PError, PeerConnectionLost
 from p2p.peer import BasePeer, PeerSubscriber
 
@@ -53,7 +58,9 @@ from trinity.sync.beam.queen import (
     QueenTrackerAPI,
 )
 from trinity.sync.beam.constants import (
+    TOO_LONG_PREDICTIVE_PEER_DELAY,
     ESTIMATED_BEAMABLE_SECONDS,
+    MAX_ACCEPTABLE_WAIT_FOR_URGENT_NODE,
     REQUEST_BUFFER_MULTIPLIER,
 )
 from trinity.sync.common.constants import (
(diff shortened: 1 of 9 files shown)


carver:
Pausing work on this until I investigate #1947 further...

trinity/sync/beam/state.py
```diff
@@ -302,32 +341,93 @@ async def _match_urgent_node_requests_to_peers(self) -> None:
...
+        # Process the returned nodes, in the order they complete
+        urgent_timer = Timer()
+        for result_coro in asyncio.as_completed(urgent_requests):
+            nodes_returned, new_nodes, peer = await result_coro
+            time_on_urgent = urgent_timer.elapsed
+
+            # After the first peer returns something, ignore all other pending tasks
+            if len(nodes_returned) > 0:
+                # Stop waiting for other peer responses
+                break
```
carver, line 389: @gsalgado does this look okay to you? The code manually cancels any unfinished tasks below, but I'm not certain that's sufficient.

Basically, we only care about the first one to respond here.

trinity/sync/beam/state.py
```diff
@@ -302,32 +341,93 @@ async def _match_urgent_node_requests_to_peers(self) -> None:
...
+    async def _find_urgent_nodes(
+            self,
+            queen: ETHPeer,
+            urgent_hashes: Tuple[Hash32, ...],
+            batch_id: int) -> None:
+
+        # Generate the request tasks
+        knights = tuple(self._queen_tracker.pop_knights())
+        urgent_requests = [
+            asyncio.create_task(self._store_nodes(peer, urgent_hashes, urgent=True))
```
gsalgado, line 376: Better to use `p2p.async_utils.create_task()`

trinity/sync/beam/state.py
gsalgado: If the task running this is cancelled we wouldn't reach line 419, and so we'd leave pending tasks to be garbage collected, which would cause the infamous asyncio warnings. 

I see you can't use `p2p.asyncio_utils.wait_first()` because you want to stop only when you get a response containing nodes, right? In that case I think you need to wrap the for loop with a `async with async_service.asyncio.cleanup_tasks(*urgent_requests)`. And then you don't need to manually cancel them anymore (lines 419-425)

pipermerriam: Rather than use the one from `async_service` I'd suggest just copy/pasta into local utilities here since that API isn't documented and thus is implicitly not public.

Issue #1913, opened 2020-07-20
Remove stack trace from RPC exception warnings


carver:
### What is wrong?

When finding #1912 , @stobiewan also reported (in gitter) that sync halted:
```
trying to read the chain id was enough to stop trinity syncing...
/trinity/trinity/rpc/main.py", line 105, in _lookup_method
    raise ValueError("Method not implemented: %r" % rpc_method)
ValueError: Method not implemented: 'eth_chainId'
trinity kept running after that but did nothing but NAT port forwarding.
```

That's definitely not supposed to happen. All exceptions are supposed to be caught and returned to the RPC caller as a string, and not be re-raised (or cause any other havoc).

### How can it be fixed

TBD

carver:
This might belong in a first stable milestone instead, but we can demote it from beta when we groom the backlog. (and have created a stable milestone)

cburgdorf:
I can't confirm this from my own test. The sync seems to move on unaffected.

```
    INFO  2020-07-21 11:05:58,913   BaseBodyChainSyncer  blks=1744  txs=507    bps=54   tps=28    elapsed=5.0  head=#54673 a49e..ea4f  age=4y11m1w
 WARNING  2020-07-21 11:05:59,691             RPCServer  RPC method caused exception
Traceback (most recent call last):
  File "/home/cburgdorf/Documents/hacking/ef/trinity/trinity/rpc/main.py", line 103, in _lookup_method
    return getattr(module, method_name)
AttributeError: 'Eth' object has no attribute 'chainid'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/cburgdorf/Documents/hacking/ef/trinity/trinity/rpc/main.py", line 120, in _get_result
    method = self._lookup_method(request['method'])
  File "/home/cburgdorf/Documents/hacking/ef/trinity/trinity/rpc/main.py", line 105, in _lookup_method
    raise ValueError("Method not implemented: %r" % rpc_method)
ValueError: Method not implemented: 'eth_chainid'
    INFO  2020-07-21 11:05:59,805        SkeletonSyncer  Unable to find common ancestor betwen our chain and ETHPeer (eth, 63) <Session <Node(0x08c094@139.5.203.2)> a04eb2bd-653a-4229-a102-a99176c9d124>
[...]

PR #1937, opened 2020-07-23
Do not include stacktrace in RPC warnings

cburgdorf:
### What was wrong?

Our logs include stacktraces that aren't really much beneficial or even feel broken to the user.

#1913 

### How was it fixed?

Removed `exc_info=True`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


newsfragments/1913.feature.rst
```diff
@@ -0,0 +1,2 @@
+Make logs a little cleaner by not including stack traces in warnings
+that are generated by hitting missing RPC endpoints.
```
carver, line 2: maybe this is a doc? or maaybe bugfix?  calling it a feature seems surprising to me

cburgdorf: Yeah, I was wondering...what is this thing..haha. It's a user facing change after all. I'll change it to bugfix.

PR #1940, opened 2020-07-23
Do not start backfill before beam sync started

cburgdorf:
### What was wrong?

Backfill of headers and blocks currently starts before we even managed to kick of the actual start of beam sync. This may cause so much download pressure that we have a hard time to actually get to the point where beam sync kicks off.

### How was it fixed?

Move the backfill to the end of the launch routine to ensure we do not start backfilling before the actually launch routine suceeded.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/chain.py
+++ b/trinity/sync/beam/chain.py
@@ -200,15 +200,6 @@ class BeamSyncer(Service):
             self.manager.cancel()
             return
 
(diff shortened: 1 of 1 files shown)


cburgdorf:
Closing this as @carver has implemented another variant of this in #1941 with some extra goodies.

PR #1941, opened 2020-07-23
Delay backfill start until Beam Sync launch

### LINKED ISSUE #1936: Backfill blocks in reverse order
### What is wrong?

We currently backfill historicial blocks from oldest to newest. It's a fair assumption that recent blocks are generally more "interesting" than older blocks.

In fact, it can be observed that peers may drop us when they ask us about recent block transactions that we don't have.

### How can it be fixed

We can easily swap the direction from which the backfiller syncs historical blocks to start with the most recent blocks and work the way back home in direction towards genesis.

carver:
### What was wrong?

Block backfill can slow down Beam Sync (occasionally, dramatically). It's not urgent, so better to wait until Beam Sync is fully launched.

Also, fixes #1936 

### How was it fixed?

Pause the backfillers at the beginning. Delay the resume/pause monitor until Beam Sync has collected all the data it needs to launch.

I haven't had a chance to manually test it yet.

### To-Do

- [x] Clean up commit history

- [ ] ~Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)~ *not necessary?*

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/chain.py
+++ b/trinity/sync/beam/chain.py
@@ -445,6 +445,7 @@ class BodyChainGapSyncer(Service):
             launch_header_fn=_get_launch_header,
             should_skip_header_fn=body_for_header_exists(self._db, self._chain)
         )
+        self._body_syncer.logger = self.logger
 
     def _get_next_gap(self) -> BlockRange:
         gaps, future_tip_block = self._db.get_chain_gaps()
--- a/trinity/sync/beam/chain.py
+++ b/trinity/sync/beam/chain.py
@@ -45,6 +45,9 @@ from trinity.sync.beam.constants import (
     RESUME_BACKFILL_AT_LAG,
     PREDICTED_BLOCK_TIME,
 )
+from trinity.sync.common.constants import (
+    MAX_BACKFILL_BLOCK_BODIES_AT_ONCE,
+)
 from trinity.sync.common.checkpoint import (
     Checkpoint,
 )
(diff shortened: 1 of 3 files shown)


trinity/sync/beam/chain.py
```diff
@@ -468,7 +478,7 @@ def _get_next_gap(self) -> BlockRange:
             else:
                 raise ValidationError("No gaps in the chain of blocks")
         else:
-            return gaps[0]
+            return gaps[-1]
```
carver, line 481: I haven't thought really hard about this. Maybe this is enough? If there is only one block gap, it still syncs from the beginning of that gap of course. Not sure we can do much about that. But at least at a pivot, it should switch to the blocks that we leapt over.

trinity/sync/beam/chain.py
```diff
@@ -201,13 +201,18 @@ async def run(self) -> None:
             return
 
         if self._enable_backfill:
+            # Pause backfill until we confirm that beam sync is launched
+            self._block_backfill.pause()
+            self._header_backfill.pause()
```
cburgdorf, line 206: Ah, I had come up with something slightly different in #1940 by just moving it to the end of the launch routine. Any thoughts about that?

trinity/sync/beam/chain.py
cburgdorf: Yeah, that looks like a quick win. But I guess we can go even one step further and actually yield the headers in reverse, no?

https://github.com/ethereum/trinity/blob/112f211efa5c6deafccfb6fd67130a9d4708750e/trinity/sync/common/headers.py#L575-L581

trinity/sync/beam/chain.py
carver: Yeah, I'll probably switch to that. :+1:

pipermerriam: This seems fishy to me.  Simply moving it to the end of the launch routine doesn't seem like it will do anything unless there are places that actually block/wait for beam sync to progress to a safe point.

cburgdorf: It should block here imho

https://github.com/ethereum/trinity/blob/81beae51ab34fe2ac57de3cf301730b7522fbcf0/trinity/sync/beam/chain.py#L234

trinity/sync/beam/chain.py
carver: Hm, actually, the block syncers are really not built for handling things in reverse (they prefer to work in order). So I think just picking a maximum range (like we do with the headers) is sufficient and simple as a solution.

PR #1946, opened 2020-07-23
State backfill improvements

carver:
### Changes

Key change: state backfill does a better job of grouping up storage nodes now (it was only ever asking each peer for up to one storage node). Now it asks for many storage nodes at once, from a single peer. This affects test time, and should make `walked=...ppm` increase a bit faster.

- Count how many total requests were made during the logged period for state backfill
- Only display state backfill log if at least one request was made (a precursor for potentially starving backfill of peers during predictive requests, or "spread beam" requests)
- Fixed a log that displayed the wrong peer's information during a test
- Can now turn state backfill on and off, just for the tests (it's a bad idea to turn it off during real operation for now)
- Make a state backfill test a little less likely to be flaky
- Bonus: The beam launchpoint log was really noisy, printing the full repr of a bunch of headers. Converted to string for shorter log

These are fairly independent commits. Reviewing them individually probably makes sense.

### To-Do

- [ ] Clean up commit history

- [x] ~Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)~ fixing unreleased bugs

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/chain.py
+++ b/trinity/sync/beam/chain.py
@@ -45,6 +45,10 @@ from trinity.sync.beam.constants import (
     RESUME_BACKFILL_AT_LAG,
     PREDICTED_BLOCK_TIME,
 )
+from trinity.sync.beam.queen import (
+    QueenTrackerAPI,
+    QueeningQueue,
+)
 from trinity.sync.common.constants import (
     MAX_BACKFILL_BLOCK_BODIES_AT_ONCE,
 )
@@ -120,7 +124,8 @@ class BeamSyncer(Service):
             event_bus: EndpointAPI,
             checkpoint: Checkpoint = None,
             force_beam_block_number: BlockNumber = None,
-            enable_backfill: bool = True) -> None:
+            enable_backfill: bool = True,
+            enable_state_backfill: bool = True) -> None:
         self.logger = get_logger('trinity.sync.beam.chain.BeamSyncer')
 
         self._body_for_header_exists = body_for_header_exists(chain_db, chain)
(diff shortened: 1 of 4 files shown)


trinity/sync/beam/chain.py
```diff
@@ -120,6 +124,7 @@ def __init__(
             event_bus: EndpointAPI,
             checkpoint: Checkpoint = None,
             force_beam_block_number: BlockNumber = None,
+            enable_state_backfill: bool = False,
```
carver, line 127: TODO: swap this one down: the beam sync service is using argument order, and it's better hygiene anyway

Also, default to True

carver:
integration test: "trinity.components.builtin.syncer.etherscan_api.EtherscanAPIError: Invalid status code: 403, Forbidden"
p2p: "p2p.exceptions.HandshakeFailure: Peer failed to send their ENR"

Both of these look unrelated.

trinity/sync/beam/chain.py
```diff
@@ -249,13 +260,16 @@ async def run(self) -> None:
             # Now we can check the lag (presumably ~0) and start backfill
             self.manager.run_daemon_task(self._monitor_historical_backfill)
 
+        if self._enable_state_backfill:
+            # Start state background service
+            self.manager.run_child_service(self._backfiller)
+        else:
+            self.manager.run_child_service(self._queen_queue)
```
cburgdorf, line 267: I assume instead of this if/else we could always run `self.manager.run_child_service(self._queen_queue)` since `self._queen_queue` references `self._backfiller` if state backfill is enabled. But maybe you did the if/else on purpose. Just pointing it out in case it wasn't on purpose.

PR #1947, opened 2020-07-23
Unbundle urgent & predictive node requests

carver:
### What was wrong?

As mentioned in #1934 -- unbundling predictive and urgent node requests might bring a meaningful win. As you can see, the time to receive a response seems to increase with the number of nodes requested. This slows down the current block import.

image

### How was it fixed?

Split the urgent and predictive nodes into two daemon tasks.

Also:
- A rename from `readd_peasant` to `insert_peer`
- Urgent node requests block on the queen peer
- Predictive requests are spammed to as many (non-queen) peers as available. This slows down backfill, which is lower priority.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1947.performance.rst
+++ b/newsfragments/1947.performance.rst
@@ -1,0 +1,2 @@
+Spead up Beam Sync a bit by only asking for critical state data from our fastest peer, instead of
+bundling in the state data needed for pending blocks.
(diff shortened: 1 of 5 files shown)


carver:
Okay, there's a lot more work to do on state sync, but this seems like a good place to pinch it off.

PR #1950, opened 2020-07-24
accept all possible "goodbye reason" values

ralexstokes:
### What was wrong?

We only accepted the few values of the `IntEnum` as "goodbye reason codes" when any 8-byte value is a possibility.

### How was it fixed?

Just subclass `Int` instead.

#### Cute Animal Picture


ralexstokes:
e.g. we got a `250` during a run of the beacon node and it disrupted the sync.

PR #1956, opened 2020-07-27
add chain id and network id for the host eth1 chain

ralexstokes:
### What was wrong?

Fully determine the eth1 system used to bootstrap a particular eth2 chain.

Added in `v0.12.2` in preparation for `medalla` testnet.

### How was it fixed?

Update the config.

#### Cute Animal Picture


ralexstokes:
For reference: https://github.com/ethereum/eth2.0-specs/pull/1988

eth2/beacon/state_machines/forks/skeleton_lake/configs.py
```diff
@@ -58,6 +58,8 @@
     # Fork choice
     SAFE_SLOTS_TO_UPDATE_JUSTIFIED=2,
     # Deposit contract
+    DEPOSIT_CHAIN_ID=999,
+    DEPOSIT_NETWORK_ID=999,
```
g-r-a-n-t, line 62: should these be 5 and 5 to remain consistent with the minimal config in the eth2-specs tests?

https://media.githubusercontent.com/media/ethereum/eth2.0-spec-tests/master/tests/minimal/config/phase0.yaml

ralexstokes: my rationale is that these constants would correspond to chain and network ID for an eth1 chain that hosted this "skeleton lake" eth2 network.

Skeleton Lake at this point is here for historical reasons as it was essentially the "minimal" testnet we used while working w/ cross-client interop some time last year...

I can update them if you want but my thinking is that given the context it is safer if we don't accidentally end up having a SkeletonLake chain hosted off of Goerli

if you disagree i'm happy to update the constants

g-r-a-n-t: That's fine, I was unaware of the background here and just assumed that this was supposed to be the same as the config linked above. Good to know where this Skeleton Lake thing came from.

PR #1958, opened 2020-07-28
Do not start beacon node from scratch everytime

ralexstokes:
### What was wrong?

We had no facility for starting a beacon node from existing chain state, instead re-starting from genesis every time.

### How was it fixed?

NOTE: depends on #1956.

We address this issue by restoring the node's context from any data available in the underlying DB.

This PR is also a good opportunity to upgrade the "start from genesis" semantics to "start from a recent state" semantics to lay the ground work for starting from arbitrary weak subjectivity states (see https://notes.ethereum.org/@adiasg/weak-subjectvity-eth2 for more info).

And along the way I tried to tidy up and/or simplify some of the configuration from the outer layers of the app down to the beacon node (as a `Component` on the `Trinity` platform).

### To-Do

- finish handling restoring fork choice
- test that it all works!
- add issue to properly handle arbitrary "recent state"; right now, need to run genesis before a future state... ideally can use any trusted state but it looks like we would have to track canonical block roots, rather than the full blocks
- add issue about "restarting" from our finalized head every time... can do better by keeping track of the canonical chain segment after the finalized head (but this means the fork choice can store/load this "context" and it looks like our sync may not handle this in the best way)

#### Cute Animal Picture

### DIFF
--- a/eth2/api/http/validator.py
+++ b/eth2/api/http/validator.py
@@ -26,7 +26,7 @@ from eth2.beacon.types.attestations import Attestation, AttestationData
 from eth2.beacon.types.blocks import BeaconBlock, SignedBeaconBlock
 from eth2.beacon.types.checkpoints import Checkpoint
 from eth2.beacon.types.states import BeaconState
-from eth2.beacon.typing import Bitfield, CommitteeIndex, Epoch, Root, Slot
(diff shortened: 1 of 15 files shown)


ralexstokes:
Closing as the important bits are in #1987 and associated PRs.

PR #1961, opened 2020-07-30
Quieter backfill when predictive beam sync active

carver:
### What was wrong?

#1947 depends on peasant peers for predictive nodes, which adds contention with backfill. Predictive nodes should clearly be the higher priority.

### How was it fixed?

Pause backfill if there was any recent predictive node requests.

### To-Do

- [x] Clean up commit history

- [x] Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md) -- hmm, I guess this needs a note? naaah

#### Cute Animal Picture

### DIFF
--- a/trinity/_utils/priority.py
+++ b/trinity/_utils/priority.py
@@ -1,0 +1,70 @@
+import asyncio
+from contextlib import asynccontextmanager
+import time
+from typing import AsyncIterator
+
+
+class SilenceObserver:
+    """
+    This class is helpful for actions that happen in the background, with no urgency.
+    Additionally, we might have urgent tasks that are competing for some shared resource.
+
+    SilenceObserver is used like so:
+    - initialize with a parameter defining how long to wait since some urgent
+          task finished before starting the non-urgent task. This would wait
+          for half a second:
+          ``observer = SilenceObserver(minimum_silence_duration=0.5)``
+    - urgent tasks are called within an ``async with observer.make_noise():`` context
+    - non-urgent tasks block on ``await observer.until_silence()`` before acting
+
+    The until_silence() method won't release until no urgent task has made
+    noise for 0.5s in the example above.
+    """
+    def __init__(self, minimum_silence_duration: float) -> None:
+        self._minimum_silence_duration = minimum_silence_duration
+        self._noise_makers = 0
+        self._last_noise_at = 0.0
+        # Trigger an event when there are no noise makers left
+        self._inactive = asyncio.Event()
+        # We start with no noise makers, so trigger immediately
+        self._inactive.set()
+
+    @asynccontextmanager
+    async def make_noise(self) -> AsyncIterator[None]:
+        """Wrap urgent tasks in this context"""
+        self._noise_makers += 1
+        self._inactive.clear()
+        try:
+            yield
+        finally:
+            self._last_noise_at = time.monotonic()
+            self._noise_makers -= 1
+
+            # Notify anyone waiting if the noise makers have dropped to 0
+            if self._noise_makers == 0:
+                self._inactive.set()
+            elif self._noise_makers < 0:
+                raise ValueError(f"Invalid state: {self._noise_makers} noise makers")
+            else:
+                # No action to take when there are still noise makers left
+                pass
+
+    def _check_silence(self) -> bool:
+        if self._noise_makers > 0:
+            # There are current noise makers, do not trigger silence
+            return False
+        else:
+            time_since_noise = time.monotonic() - self._last_noise_at
+            return time_since_noise > self._minimum_silence_duration
+
+    async def until_silence(self) -> None:
+        """
+        Non-urgent tasks block on this coro. It waits until all the noise
+        makers have gone quiet.
+
+        To avoid churn, wait until there has been no noise for
+        minimum_silence_duration.
+        """
+        while not self._check_silence():
+            await self._inactive.wait()
+            await asyncio.sleep(self._minimum_silence_duration)
--- a/trinity/sync/beam/backfill.py
+++ b/trinity/sync/beam/backfill.py
@@ -45,6 +45,9 @@ from trie.typing import (
 
 from p2p.exceptions import BaseP2PError, PeerConnectionLost
 
+from trinity.protocol.eth.constants import (
+    MAX_STATE_FETCH,
+)
 from trinity.protocol.eth.peer import ETHPeer, ETHPeerPool
 from trinity.sync.beam.constants import (
     EPOCH_BLOCK_LENGTH,
@@ -62,7 +65,8 @@ from .queen import (
     QueenTrackerAPI,
 )
 
-REQUEST_SIZE = 16
+# How many trie nodes to request in each backfill message to peers:
+REQUEST_SIZE = MAX_STATE_FETCH
 
 
 class BeamStateBackfill(Service, QueenTrackerAPI):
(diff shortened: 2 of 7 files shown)


carver:
Weird, the test finishes in 3.5s locally but is timing out after 20s on CI...

trinity/_utils/priority.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @asynccontextmanager
+    async def make_noise(self) -> AsyncIterator[None]:
+        """Wrap urgent tasks in this context"""
+        self._long_enough_silence.clear()
+        self._noise_makers += 1
+        try:
+            yield
+        finally:
+            self._last_noise_end = time.monotonic()
```
pipermerriam, line 37: *name nitpick*  `_last_noise_end_at`?

trinity/_utils/priority.py
```diff
@@ -0,0 +1,68 @@
...
+            return False
+        else:
+            time_since_noise = time.monotonic() - self._last_noise_end
+            return time_since_noise > self._minimum_silence_duration
+
+    def _silence_trigger(self) -> None:
+        if self._check_silence():
+            self._long_enough_silence.set()
+
+    async def silence(self) -> None:
```
pipermerriam, line 54: Alternate names:  `wait_until_quiet`, `wait_quiet`, `wait_silence`  all give a little extra indication that it is *waiting* until no noise is happening.  Current name suggests that maybe it is *silencing* something.

trinity/sync/full/chain.py
```diff
@@ -331,7 +331,7 @@ async def _run_body_download_batch(
             if len(non_trivial_headers) == 0:
                 # peer had nothing to do, so have it get back in line for processing
                 self._body_peers.put_nowait(peer)
-            elif len(completed_headers) > 0:
+            elif len(received_headers) > 0:
```
pipermerriam, line 334: bugfix?

carver: Yeah, it's explained a bit more in the commit c8fba693e6f3438ce73

Basically, we want to demote peers a bit who give us empty responses about block bodies. In the old implementation, we were giving them a pass if we happened to ask them for an empty block. We would have "completed headers", because there was nothing to download, even if they returned no info. So this commit is just a tiny bit stricter in applying the demotion.

trinity/_utils/priority.py
carver: Yeah, something bothers me about `await ..wait`. It sounds natural to say `await ..silence`. But :+1: on avoiding confusion of `silence()` sounding like a verb. Toying with `await ..until_silence()`

pipermerriam: Not arguing, fine with what you proposed.  Just want to point out that `Event.wait()` is a pretty common version of this as well as `asyncio.wait_for`, but I also sort of agree with you...  :woman_shrugging:

carver:
Hm, still don't know why backfill isn't working. A couple requests happen, and then it all kind of stalls out. The amount of backfill data per request just went up a lot. Not sure why that would affect it. But this looks suspicious:

```
   DEBUG  07-30 22:26:49        multiplexer.py  Timeout <Multiplexer[p2p:5,eth:65]>: 
  File "/home/circleci/repo/p2p/multiplexer.py", line 333, in _do_multiplexing
    await self._handle_commands(msg_stream)
  File "/home/circleci/repo/p2p/multiplexer.py", line 345, in _handle_commands
    async for protocol, cmd in msg_stream:
  File "/home/circleci/repo/p2p/multiplexer.py", line 58, in stream_transport_messages
    msg = await transport.recv()
  File "/home/circleci/repo/p2p/tools/memory_transport.py", line 73, in recv
    encoded_sizes = await self.read(8)
  File "/home/circleci/repo/p2p/tools/memory_transport.py", line 65, in read
    return await asyncio.wait_for(self._reader.readexactly(n), timeout=CONN_IDLE_TIMEOUT)
  File "/usr/local/lib/python3.8/asyncio/tasks.py", line 490, in wait_for
    raise exceptions.TimeoutError()
asyncio.exceptions.TimeoutError
```

pipermerriam:
> In a situation with only mocked peers, I can't imagine why we would get a timeout like this.

Maybe something blocking the event loop?

carver:
> > In a situation with only mocked peers, I can't imagine why we would get a timeout like this.
> 
> Maybe something blocking the event loop?

Hah, yeah, just came to the same conclusion. Trying something...

trinity/sync/beam/backfill.py
```diff
@@ -245,6 +246,7 @@ async def _missing_trie_hashes(self) -> AsyncIterator[TrackedRequest]:
                 )
                 async for node_request in subcomponent_hashes_iterator:
                     yield node_request
+                    await asyncio.sleep(0)
```
pipermerriam, line 249: The `async for` should already be releasing the loop on each iteration for this loop.

carver: That's what I thought, too, but [this question writeup](https://stackoverflow.com/questions/55812939/how-to-iterate-over-a-large-list-without-blocking-event-loop) seems to suggest otherwise. I haven't found anything definitive yet either way.

carver:
Okay, does look like I'm on the right track about the total size of the request mattering somehow. Reducing the size back down from `MAX_STATE_FETCH` to 16 seems to make tests pass again. So probably not a problem with `SilenceObserver`.

carver:
Okay, now I'm confused again. Did adding the `await asyncio.sleep(0)` make the difference? Because the commit that did that still failed the test in the same way. But now, reverting everything meaningful since then seems to pass. :angry:

carver:
Sorry to be spamming in here. I can't reproduce locally...

trinity/sync/beam/backfill.py
gsalgado: So, this method (`_missing_trie_hashes`) is cooperative as it calls `await account_iterator.__anext__()`, so we shouldn't need to pepper it with `await asyncio.sleep(0)` calls, and that will cause any of its consumers to periodically yield back to the event loop as well, but that may not happen frequently enough because `subcomponent_hashes_iterator` is not cooperative as it never awaits, so the `async for` above (line 247) will block until all items are consumed. `_missing_storage_hashes` is also not cooperative, and nor is `_missing_bytecode_hashes`

We could either make these cooperative by peppering them with `asyncio.sleep(0)` calls in strategic places, or we can turn them into regular iterators/generators and have their async callsites call `asyncio.sleep(0)` (what you did above). What I think we don't want is to have `asyncio.sleep(0)` calls both in the iterators and in their consumers, as that adds unnecessary context switching. If we turn them into regular iterators, that can't happen, and it also means they could also be used in sync contexts

carver:
> Unless... [4c05aad](https://github.com/ethereum/trinity/commit/4c05aada3fd4b6369cfd25868944498b8557adab) is covering up some bug in `SilenceObserver`...

I couldn't confirm for sure, but the `SilenceObserver` did have a bug if the `call_later` triggers too early, then the coro waiting on it might never wake up. Fixed that in 85ca9bdd0604b206e963877c63a4160aa82f87b5. Looks promising.

trinity/sync/beam/state.py
```diff
@@ -433,16 +438,15 @@ def _is_node_present(self, node_hash: Hash32) -> bool:
         """
         return node_hash in self._db
 
-    async def _node_hashes_present(self, node_hashes: Set[Hash32]) -> None:
+    async def _node_hashes_present(self, node_hashes: Set[Hash32], timeout: float) -> None:
```
pipermerriam, line 441: Just pointing out that `trio` considers this an antipattern.  I don't have a helpful suggestion for how to change this, but it's something maybe worth considering whether you can control this timeout higher up the stack rather than relying on it getting passed down through these call frames.

PR #1964, opened 2020-07-31
yield to the event loop inside of async iterator implementations

pipermerriam:
### What was wrong?

Multiple async iterator implementations were not yielding to the event loop during iteration.

### How was it fixed?

Updated the various implementations of `AsyncIterator` to include `asyncio.sleep(0)` calls.

#### Cute Animal Picture

chicken-red-head

### DIFF
--- a/p2p/multiplexer.py
+++ b/p2p/multiplexer.py
@@ -278,15 +278,8 @@ class Multiplexer(MultiplexerAPI):
             self.raise_if_streaming_error()
             msg_queue = self._protocol_queues[protocol_class]
             while self.is_streaming:
-                try:
-                    # We use an optimistic strategy here of using
-                    # `get_nowait()` to reduce the number of times we yield to
-                    # the event loop.  Since this is an async generator it will
-                    # yield to the loop each time it returns a value so we
-                    # don't have to worry about this blocking other processes.
-                    yield msg_queue.get_nowait()
-                except asyncio.QueueEmpty:
-                    yield await msg_queue.get()
+                await asyncio.sleep(0)
+                yield await msg_queue.get()
 
     #
     # Message reading and streaming API
--- a/p2p/token_bucket.py
+++ b/p2p/token_bucket.py
@@ -2,7 +2,7 @@ import asyncio
 import time
 from typing import (
     Union,
-    AsyncGenerator,
+    AsyncIterator,
 )
 
 
@@ -25,7 +25,7 @@ class TokenBucket:
         self._seconds_per_token = 1 / self._rate
         self._take_lock = asyncio.Lock()
 
-    async def __aiter__(self) -> AsyncGenerator[None, None]:
+    async def __aiter__(self) -> AsyncIterator[None]:
         """
         Can be used as an async iterator to limit the rate at which a loop can
         run.
@@ -79,6 +79,8 @@ class TokenBucket:
         if self._num_tokens < 0:
             sleep_for = abs(self._num_tokens) * self._seconds_per_token
             await asyncio.sleep(sleep_for)
+        else:
+            await asyncio.sleep(0)
 
     def take_nowait(self, num: Union[int, float] = 1) -> None:
         # we calculate this value locally to ensure that in the case of not
--- a/trinity/_utils/async_iter.py
+++ b/trinity/_utils/async_iter.py
@@ -1,3 +1,4 @@
+import asyncio
 from typing import (
     AsyncIterable,
     Set,
@@ -39,3 +40,4 @@ async def async_take(take_count: int, iterator: AsyncIterable[TYield]) -> AsyncI
             yield val
             if taken == take_count:
                 break
+            await asyncio.sleep(0)
--- a/trinity/components/eth2/eth1_monitor/eth1_monitor.py
+++ b/trinity/components/eth2/eth1_monitor/eth1_monitor.py
@@ -303,6 +303,7 @@ class Eth1Monitor(Service):
                             f"Block does not exist for block number={block_number}"
                         )
                     yield block
+                    await trio.hazmat.checkpoint()
             await trio.sleep(self._polling_period)
 
     def _handle_block_data(self, block: Eth1Block) -> None:
--- a/trinity/protocol/common/servers.py
+++ b/trinity/protocol/common/servers.py
@@ -136,6 +136,7 @@ class BasePeerRequestHandler:
         Generates the headers requested, halting on the first header that is not locally available.
         """
         for block_num in block_numbers:
+            await asyncio.sleep(0)
             try:
                 yield await self.db.coro_get_canonical_block_header_by_number(block_num)
             except HeaderNotFound:
--- a/trinity/sync/beam/backfill.py
+++ b/trinity/sync/beam/backfill.py
@@ -201,6 +201,8 @@ class BeamStateBackfill(Service, QueenTrackerAPI):
 
         try:
             while self.manager.is_running:
+                await asyncio.sleep(0)
+
                 # Get the next account
 
                 # We have to rebuild the account iterator every time because...
@@ -295,6 +297,8 @@ class BeamStateBackfill(Service, QueenTrackerAPI):
         starting_index = bytes_to_nibbles(root_hash)
 
         while self.manager.is_running:
+            await asyncio.sleep(0)
+
             try:
                 path_to_node = request_tracker.next_path_to_explore(starting_index)
             except trie_exceptions.PerfectVisibility:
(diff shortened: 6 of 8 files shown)


trinity/sync/full/chain.py
```diff
@@ -184,6 +184,7 @@ async def _sync_from_headers(
         highest_block_num = -1
 
         async for headers in get_headers_coro:
+            await asyncio.sleep(0)
```
gsalgado, line 187: Here it isn't needed because a few lines below we do an `await task_integrator.wait_add_tasks(...)`

trinity/sync/beam/chain.py
```diff
@@ -574,9 +575,11 @@ async def new_sync_headers(
             [str(header) for header in self._launchpoint_headers],
         )
         yield self._launchpoint_headers
+        await asyncio.sleep(0)
```
gsalgado, line 578: Here it isn't needed because a few lines above we do an `await self._at_launchpoint.wait()`

trinity/sync/beam/chain.py
```diff
@@ -574,9 +575,11 @@ async def new_sync_headers(
...
 
         async for headers in self._real_syncer.new_sync_headers(max_batch_size):
             yield headers
+            await asyncio.sleep(0)
```
gsalgado, line 582: As I've mentioned in https://github.com/ethereum/trinity/pull/1961#discussion_r464160990 we probably don't want to do this here (as it just adds extra unnecessary context switches) if `self._real_syncer.new_sync_headers()` is actually cooperative and not just in its signature.

trinity/sync/beam/backfill.py
```diff
@@ -295,6 +297,8 @@ async def _request_tracking_trie_items(
         starting_index = bytes_to_nibbles(root_hash)
 
         while self.manager.is_running:
+            await asyncio.sleep(0)
```
gsalgado, line 300: Same as above

trinity/sync/beam/backfill.py
```diff
@@ -405,6 +409,8 @@ async def _missing_storage_hashes(
 
         storage_tracker = self._get_storage_tracker(address_hash_nibbles)
         while self.manager.is_running:
+            await asyncio.sleep(0)
+
```
gsalgado, line 413: Same as above

trinity/protocol/common/servers.py
```diff
@@ -136,6 +136,7 @@ async def _generate_available_headers(
         Generates the headers requested, halting on the first header that is not locally available.
         """
         for block_num in block_numbers:
+            await asyncio.sleep(0)
```
gsalgado, line 139: This isn't needed as the next line will yield back to the event loop

trinity/sync/beam/backfill.py
```diff
@@ -201,6 +201,8 @@ async def _missing_trie_hashes(self) -> AsyncIterator[TrackedRequest]:
 
         try:
             while self.manager.is_running:
+                await asyncio.sleep(0)
```
gsalgado, line 204: So, in practice we're making this an actual async (i.e. cooperative) iterator, which it claimed to be but wasn't. Another alternative would be to turn it into a regular iterator and let its callsites decide how often they want to yield back to the event loop. That would be useful if, for instance, a callsite wants to get a batch of items and process them all together, as it'd save a lot of context switches since we'd yield back to the event loop once per batch instead of once per item. That feels to me like the correct thing to do here, unless we know that every iteration here is expensive and our callsites process items individually

Indeed its callsite processes items in batches: https://github.com/ethereum/trinity/blob/50477143f9240211d6c45b0eb77572b14ff269c9/trinity/sync/beam/backfill.py#L133

trinity/_utils/async_iter.py
```diff
@@ -39,3 +40,4 @@ async def async_take(take_count: int, iterator: AsyncIterable[TYield]) -> AsyncI
             yield val
             if taken == take_count:
                 break
+            await asyncio.sleep(0)
```
gsalgado, line 43: If `iterator` is actually async (i.e. cooperative and not just a regular iterator with an `async` signature), this is redundant and will actually cause unnecessary context switching

p2p/token_bucket.py
```diff
@@ -79,6 +79,8 @@ async def take(self, num: Union[int, float] = 1) -> None:
         if self._num_tokens < 0:
             sleep_for = abs(self._num_tokens) * self._seconds_per_token
             await asyncio.sleep(sleep_for)
+        else:
+            await asyncio.sleep(0)
```
gsalgado, line 83: Hmm, does it make sense to have an `await sleep(0)` at the end of the method?  Since it returns, wouldn't the control go back to the event loop?

p2p/multiplexer.py
```diff
@@ -278,15 +278,8 @@ async def _stream_protocol_messages(self,
...
-                try:
-                    # We use an optimistic strategy here of using
-                    # `get_nowait()` to reduce the number of times we yield to
-                    # the event loop.  Since this is an async generator it will
-                    # yield to the loop each time it returns a value so we
-                    # don't have to worry about this blocking other processes.
-                    yield msg_queue.get_nowait()
-                except asyncio.QueueEmpty:
-                    yield await msg_queue.get()
+                await asyncio.sleep(0)
```
gsalgado, line 281: AIUI, the sleep(0) would be needed if we were still using `.get_nowait()` as that would never yield back to the event loop, but now this should happen on every `await msg_queue.get()` call

gsalgado (commented):
This was a good find! I left a bunch of comments, but I really wish we could find a way to avoid this from happening again in the future, although I can't think of a way to do so other than documenting it on the tactical manual

p2p/token_bucket.py
carver: Yeah, I'm not as clear that there are situations where `take()` is called a lot in a row, with a huge number of tokens. I would be fine taking this change out, but no problem leaving it in.

trinity/_utils/async_iter.py
carver: If this method was used by a really time-sensitive piece of code, I would be curious to see how much longer it takes to collect 384 nodes when yielding in between each one. But since it's only used for backfill, no objection to making the change and to come back to it later if we start using `async_take()` elsewhere.

trinity/components/eth2/eth1_monitor/eth1_monitor.py
```diff
@@ -303,6 +303,7 @@ async def _new_blocks(self) -> AsyncGenerator[Eth1Block, None]:
                             f"Block does not exist for block number={block_number}"
                         )
                     yield block
+                    await trio.hazmat.checkpoint()
```
carver, line 306: If you feel it makes sense: a note that this changes to `trio.lowlevel.checkpoint()` in more recent versions of trio. (Actually, I only really mention it because it was a little more convoluted to find the docs for the now-deprecated API)

trinity/protocol/common/servers.py
carver: I agree: it seems that `run_in_executor` always yields control (which is how these `coro_` methods are implemented).

trinity/sync/beam/backfill.py
carver: Each iteration can actually become fairly slow. It iterates through the entire existing trie, eventually. The fuller the trie is locally, the more time it looks through the trie before finding something that's missing. If you have to pivot when the database is 99% filled, you will take quite a while before filling up a batch of missing nodes. In fact, it's probably also important to have this deeper in the `_request_tracking_trie_items`, because even finding each single item can take quite a while.

But now that I see it's already in `_request_tracking_trie_items()`, I think this and a couple others are redundant.

trinity/sync/beam/backfill.py
carver: I agree this one can be dropped, because it's already covered by the one inside of `_request_tracking_trie_items()`

trinity/sync/beam/chain.py
```diff
@@ -340,6 +340,7 @@ async def _get_ancestors(self,
             yield parent
             headers_returned += 1
             header = parent
+            await asyncio.sleep(0)
```
carver, line 343: Pretty sure that each `coro_get_block_header_by_hash` yields the event loop.

trinity/sync/beam/chain.py
carver: Yeah, I'm fine with pushing this upstream. Looks like `ManualHeaderSyncer.new_sync_headers` and `BaseHeaderChainSyncer` might need a new sleep.

trinity/sync/full/chain.py
carver: Though that only yields if the task_integrator is overloaded.

In the end, I agree this could be dropped, if we count on `new_sync_headers` to do the heavy lifting of releasing the event loop. If we have a bunch of chained async iterators, the overhead of releasing the event loop at every item *and* at every recursive level, sounds like it might be problematic.

carver (approved):

cburgdorf:
@pipermerriam I noticed this was approved but never merged (and needs to be rebased now). Was there anything holding this back?

PR #1966, opened 2020-08-03
State reading from db.

g-r-a-n-t:
### What was wrong?
We can persist Beacon states, but can't retrieve them.

### How was it fixed?
Implemented Beacon state retrieval.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/db/chain2.py
+++ b/eth2/beacon/db/chain2.py
@@ -74,7 +74,6 @@ class BeaconChainDB(BaseBeaconChainDB):
         chain_db._persist_genesis_data(genesis_state)
         chain_db.persist_state(genesis_state, config)
 
-        chain_db.mark_canonical_block(genesis_block)
         chain_db.mark_justified_head(genesis_block)
         chain_db.mark_finalized_head(genesis_block)
 
@@ -123,16 +122,9 @@ class BeaconChainDB(BaseBeaconChainDB):
         block_root_to_block = SchemaV1.block_root_to_block(block_root)
         self.db[block_root_to_block] = ssz.encode(block)
 
-        block_root_to_block_header = SchemaV1.block_root_to_block_header(block_root)
-        self.db[block_root_to_block_header] = ssz.encode(block.header)
-
         signature = signed_block.signature
         block_root_to_signature = SchemaV1.block_root_to_signature(block_root)
         self.db[block_root_to_signature] = signature
-
-    def mark_canonical_block(self, block: BaseBeaconBlock) -> None:
-        key = SchemaV1.canonical_head_root()
-        self.db[key] = block.hash_tree_root
 
     def mark_canonical_head(self, block: BaseBeaconBlock) -> None:
         canonical_head_root = SchemaV1.canonical_head_root()
@@ -157,8 +149,7 @@ class BeaconChainDB(BaseBeaconChainDB):
         Marks the given ``block`` as finalized and stores each newly finalized state and block at
         their corresponding slot.
         """
-        finalized_head_root = SchemaV1.finalized_head_root()
-        if finalized_head_root in self.db:
+        if block.slot > 0:
             newly_finalized_states = self.get_state_parents(
                 block.state_root, block.slot - self.get_finalized_head(BeaconBlock).slot
             ) + (block.state_root,)
@@ -176,7 +167,7 @@ class BeaconChainDB(BaseBeaconChainDB):
             self.db[slot_to_block_root] = latest_block_header.hash_tree_root
 
         self.db[SchemaV1.slot_to_block_root(block.slot)] = block.hash_tree_root
-        self.db[finalized_head_root] = block.hash_tree_root
+        self.db[SchemaV1.finalized_head_root()] = block.hash_tree_root
 
     def get_finalized_head(self, block_class: Type[BaseBeaconBlock]) -> BaseBeaconBlock:
         finalized_head_root_key = SchemaV1.finalized_head_root()
@@ -208,7 +199,7 @@ class BeaconChainDB(BaseBeaconChainDB):
         Returns a tuple of size ``count`` that contains the state roots from oldest to newest
         that precede the given ``state_root``.
         """
-        for _ in range(0, count):
+        for _ in range(count):
             state_root = self._read_state_parent_state_root(state_root)
             yield state_root
 
@@ -238,8 +229,7 @@ class BeaconChainDB(BaseBeaconChainDB):
         self.db[state_root_to_latest_block_header_root] = block_root
 
         block_root_to_block_header = SchemaV1.block_root_to_block_header(block_root)
-        if block_root_to_block_header not in self.db:
-            self.db[block_root_to_block_header] = ssz.encode(block_header)
+        self.db[block_root_to_block_header] = ssz.encode(block_header)
 
     def _write_state_historical_roots(
         self, state_root: Root, historical_roots: HashableList[Root]
@@ -739,7 +729,7 @@ class BeaconChainDB(BaseBeaconChainDB):
                 state_root
             ),
             finalized_checkpoint=self._read_state_finalized_checkpoint(state_root),
-            genesis_time=self._genesis_time,
+            genesis_time=self.genesis_time,
             genesis_validators_root=self._genesis_validators_root,
             config=config,
         )
@@ -764,7 +754,7 @@ class BeaconChainDB(BaseBeaconChainDB):
         """
         Store data in the database that will never change.
         """
-        self._genesis_time = genesis_state.genesis_time
+        self.genesis_time = genesis_state.genesis_time
         self._genesis_validators_root = genesis_state.genesis_validators_root
         encoded_genesis_time = ssz.encode(genesis_state.genesis_time, ssz.uint64)
         encoded_genesis_validators_root = ssz.encode(
--- a/eth2/beacon/db/chain2.py
+++ b/eth2/beacon/db/chain2.py
@@ -3,7 +3,7 @@ from typing import Iterable, Optional, Sequence, Tuple, Type
 from eth.abc import AtomicDatabaseAPI
 from eth.exceptions import BlockNotFound
 from eth.typing import Hash32
-from eth_utils import to_tuple
+from eth_utils import reversed_return, to_tuple
 from lru import LRU
 import ssz
 from ssz.hashable_list import HashableList
@@ -16,7 +16,7 @@ import eth2.beacon.db.schema2 as SchemaV1
 from eth2.beacon.genesis import get_genesis_block
 from eth2.beacon.helpers import compute_epoch_at_slot
 from eth2.beacon.types.block_headers import BeaconBlockHeader
-from eth2.beacon.types.blocks import BaseBeaconBlock, BaseSignedBeaconBlock
+from eth2.beacon.types.blocks import BaseBeaconBlock, BaseSignedBeaconBlock, BeaconBlock
 from eth2.beacon.types.checkpoints import Checkpoint
 from eth2.beacon.types.eth1_data import Eth1Data
 from eth2.beacon.types.forks import Fork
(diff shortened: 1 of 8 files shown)


eth2/beacon/db/chain2.py
```diff
@@ -487,21 +437,293 @@ def _write_state(self, state: BeaconState, config: Eth2Config) -> None:
...
+            return default_timestamp, default_root, default_root
         genesis_time = ssz.decode(data[:8], ssz.sedes.uint64)
-        genesis_validators_root = ssz.decode(data[8:], ssz.sedes.bytes32)
-        return Timestamp(genesis_time), Root(genesis_validators_root)
+        genesis_validators_root = ssz.decode(data[8:40], ssz.sedes.bytes32)
+        genesis_eth1_block_hash = ssz.decode(data[40:], ssz.sedes.bytes32)
+        return (
+            Timestamp(genesis_time),
+            Root(genesis_validators_root),
+            Root(genesis_eth1_block_hash),
```
ralexstokes, line 725: instead of storing the eth1 hash here, we could consider just writing the entire `randao_mixes` vector in the `from_genesis` constructor.

Then to read, you start w/ that vector and walk backwards from the most recent epoch until you look around `EPOCHS_PER_HISTORICAL_VECTOR` times, overwriting the appropriate position if there is something more recent in the DB.

something to consider -- i dont see any huge problems w/ what you've already done...

eth2/beacon/db/chain2.py
```diff
@@ -320,21 +287,12 @@ def _write_state_balances(
...
+    def _write_state_randao_mix(self, slot: Slot, mix: Root) -> None:
         """
         NOTE: only the root at the current epoch can have changed.
         """
-        # TODO design how to efficiently recreate the list given a state at some slot.
-        key = SchemaV1.state_root_to_randao_mix(state_root)
-        encoding = randao_mixes[current_epoch % EPOCHS_PER_HISTORICAL_VECTOR]
-        self._state_bytes_written += len(encoding)
-        self.db[key] = encoding
+        key = SchemaV1.slot_to_randao_mix(slot)
```
ralexstokes, line 294: i can't guarantee that we wouldn't have two states for the same slot, i.e. during a forking situation on the network.

in this case, we may overwrite our randao mix (and it is conceivable that the mixes hav diverged across chains) so this is safer to store by `state_root`, rather than just `slot`.

eth2/beacon/db/chain2.py
```diff
@@ -288,7 +255,7 @@ def _write_state_validators(
                     encoding = ssz.encode(validator)
                     self._state_bytes_written += len(encoding)
                     self.db[root] = encoding
-                    roots.extend(root)
```
ralexstokes, line 291: good catch -- bc this seems like a bug fix, actually... we could imagine doing something where we only store new validators but simpler to just record the list of roots (and this as i recall was the original intention :) )

eth2/beacon/db/chain2.py
```diff
@@ -52,8 +54,6 @@ def __init__(self, db: AtomicDatabaseAPI) -> None:
         self._state_cache = LRU(STATE_CACHE_SIZE)
         self._state_bytes_written = 0
 
-        self._genesis_time, self._genesis_validators_root = self._get_genesis_data()
```
ralexstokes, line 55: the idea here was to cache these values so that we only have to read them once

eth2/beacon/db/chain2.py
```diff
@@ -141,35 +141,28 @@ def get_finalized_head(self, block_class: Type[BaseBeaconBlock]) -> BaseBeaconBl
         return self.get_block_by_root(finalized_head_root, block_class)
 
     def get_state_by_slot(
-        self, slot: Slot, state_class: Type[BeaconState]
```
ralexstokes, line 144: the idea behind passing the `state_class` here is to eventually enable polymorphism of different state types across forks -- this pattern was inspired by how eth1 forks are handled in `py-evm`.

shall we do `def get_state_by_slot(self, slot: Slot, state_class: ..., config: Eth2Config)` instead?

eth2/beacon/db/chain2.py
```diff
@@ -141,35 +141,28 @@ def get_finalized_head(self, block_class: Type[BaseBeaconBlock]) -> BaseBeaconBl
...
-        # except KeyError:
-        #     raise StateNotFound()
-        # state = ssz.decode(state_data, state_class)
-        # return state
+        return self._read_state(state_root, config)
 
     def _write_state_slot(self, state_root: Root, slot: Slot) -> None:
-        key = SchemaV1.state_root_to_slot(state_root)
+        state_to_slot = SchemaV1.state_root_to_slot(state_root)
+        slot_to_state = SchemaV1.slot_to_state_root(slot)
```
ralexstokes, line 161: there are no docs on this (and it is part of what i intend to add soon) but the current thinking is that anything marked by a `slot` has been finalized, delegating the fork choice (where there could reasonably be forks) to the fork choice module.

this invariant may not be worth its weight in conceptual overhead but it has been my thinking so far...

it seems that adding the `slot -> state` look up here breaks this invariant.

eth2/beacon/db/chain2.py
```diff
@@ -192,40 +185,14 @@ def _write_state_block_header(
...
-        key = SchemaV1.state_root_to_state_root(state_root)
-        encoding = state_roots[slot % SLOTS_PER_HISTORICAL_ROOT]
-        self._state_bytes_written += len(encoding)
-        self.db[key] = encoding
+        state_root_to_block_header_root = SchemaV1.state_root_to_block_header_root(
+            state_root
+        )
+        slot_to_block_header_root = SchemaV1.slot_to_block_header_root(
+            block_header.slot
+        )
```
ralexstokes, line 193: same thing here w/ invariants around what is finalized and what is not...

eth2/beacon/db/chain2.py
```diff
@@ -487,21 +437,293 @@ def _write_state(self, state: BeaconState, config: Eth2Config) -> None:
...
+    @to_tuple
+    def _read_state_randao_mixes(
+        self,
+        state_root: Root,
+        genesis_eth1_block_hash: Root,
+        EPOCHS_PER_HISTORICAL_VECTOR: int,
+        SLOTS_PER_EPOCH: int,
+    ) -> Iterable[Root]:
+        """
+        Reconstructs the ``randa_mixes`` at a given state root.
```
ralexstokes, line 566:
```suggestion
        Reconstructs the ``randao_mixes`` at a given state root.
```

eth2/beacon/db/chain2.py
```diff
@@ -487,21 +437,293 @@ def _write_state(self, state: BeaconState, config: Eth2Config) -> None:
...
+
+        Everytime a block is processed, the ``randao_mixes`` vector is updated with a new
```
ralexstokes, line 568:
```suggestion
        Every time a block is processed, the ``randao_mixes`` vector is updated with a new
```

eth2/beacon/db/chain2.py
```diff
@@ -487,21 +437,293 @@ def _write_state(self, state: BeaconState, config: Eth2Config) -> None:
...
+        ]
+        offset = EPOCHS_PER_HISTORICAL_VECTOR - epochs[0] % EPOCHS_PER_HISTORICAL_VECTOR
+        epochs = epochs[offset:] + epochs[:offset]
+
+        for epoch in epochs:
+            if epoch < 0:
+                yield genesis_eth1_block_hash
+            elif epoch == state_epoch:
+                # yield the randao mix at the particular slot
+                key = SchemaV1.slot_to_randao_mix(state_slot)
```
ralexstokes, line 595: for example, this should be keyed by state_root and slot w/ the current overall strategy...

eth2/beacon/db/chain2.py
g-r-a-n-t: I see, Ill leave this out of the PR for now to keep it from getting more complex. let me know if you think it should go in

eth2/beacon/db/chain2.py
```diff
@@ -121,16 +123,16 @@ def persist_block(self, signed_block: BaseSignedBeaconBlock) -> None:
         block_root_to_block = SchemaV1.block_root_to_block(block_root)
         self.db[block_root_to_block] = ssz.encode(block)
 
+        block_root_to_block_header = SchemaV1.block_root_to_block_header(block_root)
+        self.db[block_root_to_block_header] = ssz.encode(block.header)
+
```
ralexstokes, line 128: just noting that the current code will never touch these block headers as they will have the state root which block headers encountered when writing the state will have the state root blank; see comment below about checking for existence of this data when recreating the state

eth2/beacon/db/chain2.py
```diff
@@ -203,48 +231,15 @@ def _write_state_fork(self, state_root: Root, fork: Fork) -> None:
...
-        NOTE: only the root at ``slot`` (with some offset) will have changed.
-        """
-        # TODO design an efficient way to recreate the list
-        key = SchemaV1.state_root_to_state_root(state_root)
-        encoding = state_roots[slot % SLOTS_PER_HISTORICAL_ROOT]
-        self._state_bytes_written += len(encoding)
-        self.db[key] = encoding
+        block_root_to_block_header = SchemaV1.block_root_to_block_header(block_root)
+        if block_root_to_block_header not in self.db:
+            self.db[block_root_to_block_header] = ssz.encode(block_header)
```
ralexstokes, line 242: this `block_header` will have an empty state root so this conditional will always be true.

refer to the comment above about this.

i think this is fine for now as it really just inflates the DB size on disk and having the headers will likely be useful in the future

eth2/beacon/db/chain2.py
```diff
@@ -121,16 +123,16 @@ def persist_block(self, signed_block: BaseSignedBeaconBlock) -> None:
...
         self.db[block_root_to_signature] = signature
 
     def mark_canonical_block(self, block: BaseBeaconBlock) -> None:
-        slot_to_block_root = SchemaV1.slot_to_block_root(block.slot)
-        self.db[slot_to_block_root] = block.hash_tree_root
-
-        slot_to_state_root = SchemaV1.slot_to_state_root(block.slot)
-        self.db[slot_to_state_root] = block.state_root
+        key = SchemaV1.canonical_head_root()
+        self.db[key] = block.hash_tree_root
```
ralexstokes, line 135: so we should be able to remove this if we move "finality accounting" to the `mark_finalized_head` method...

eth2/beacon/db/chain2.py
```diff
@@ -151,7 +153,29 @@ def get_justified_head(self, block_class: Type[BaseBeaconBlock]) -> BaseBeaconBl
         return self.get_block_by_root(justified_head_root, block_class)
 
     def mark_finalized_head(self, block: BaseBeaconBlock) -> None:
+        """
+        Marks the given ``block`` as finalized and stores each newly finalized state and block at
+        their corresponding slot.
+        """
         finalized_head_root = SchemaV1.finalized_head_root()
+        if finalized_head_root in self.db:
```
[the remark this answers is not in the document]
g-r-a-n-t: only for the genesis block, which could probably be more clear here

eth2/beacon/db/chain2.py
```diff
@@ -526,9 +764,23 @@ def _persist_genesis_data(self, genesis_state: BeaconState) -> None:
         """
         Store data in the database that will never change.
         """
-        genesis_time = ssz.encode(genesis_state.genesis_time, ssz.uint64)
-        genesis_validators_root = ssz.encode(
+        self._genesis_time = genesis_state.genesis_time
+        self._genesis_validators_root = genesis_state.genesis_validators_root
```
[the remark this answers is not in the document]
g-r-a-n-t: these are initialized to their default values in the constructor when loading from genesis, then genesis data is persisted, which is why this is added here. not sure if there is a cleaner way to go about this

PR #1969, opened 2020-08-04
Unnecessary block read removed.

g-r-a-n-t:
### What was wrong?

Fork choice processing would fail with a `BlockNotFound` exception when encountering an attestation with an unknown target root.

### How was it fixed?

We now catch the exception, log a message, and continue without processing the attestation.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth2/beacon/chains/testnet/medalla.py
+++ b/eth2/beacon/chains/testnet/medalla.py
@@ -361,10 +361,7 @@ class BeaconChain(BaseBeaconChain):
     def _get_indices_from_attestation(
         self, attestation: Attestation
     ) -> Collection[ValidatorIndex]:
-        target_block = self._chain_db.get_block_by_root(
-            attestation.data.target.root, BeaconBlock
-        )
-        sm = self.get_state_machine(target_block.slot)
+        sm = self.get_state_machine(attestation.data.slot)
         return get_attesting_indices(
             sm._epochs_ctx, attestation.data, attestation.aggregation_bits
         )


eth2/beacon/chains/testnet/altona.py
```diff
@@ -302,7 +302,8 @@ def _update_fork_choice_with_attestation(
         block_root = attestation.data.beacon_block_root
         target_epoch = attestation.data.target.epoch
         indices = self._get_indices_from_attestation(attestation)
-        fork_choice.on_attestation(block_root, target_epoch, *indices)
+        if len(indices) != 0:
+            fork_choice.on_attestation(block_root, target_epoch, *indices)
```
ralexstokes, line 306: may be a little cleaner to just pass `()` (or its equivalent) to `on_attestation`, which turn around and do nothing (in the event that `len(indices) == 0`.

eth2/beacon/chains/testnet/altona.py
```diff
@@ -361,9 +362,17 @@ def _compute_missing_state(self, target_block: BeaconBlock) -> BeaconState:
...
+            target_block = self._chain_db.get_block_by_root(
+                attestation.data.target.root, BeaconBlock
+            )
+        except BlockNotFound:
+            self.logger.debug(
+                f"failed to load block with root {attestation.data.target.root} "
+                "while processing attestation."
+            )
+            return set()
+
```
ralexstokes, line 375: we should review why this is here... the `Attestation` has the `slot` in it, right?

I believe I wrote this code but unclear why I felt the block read was necessary... upon inspection, it seems like we don't need it...?

g-r-a-n-t: Looks like it should have been changed here https://github.com/ethereum/trinity/pull/1955/files

I don't think this is necessary either.

ralexstokes:
this ready to merge?

Issue #1962, opened 2020-07-30
Blocked event loop canary


pipermerriam:
### What is wrong?

This is sort of a bug report and an architecture idea.

First the bug:   Block imports are reported on a 5 second interval and when I watch the `elapsed` part of the log message it regularly is something like `5.1` seconds and quite often you'll see variance up to even a few extra seconds.  This seems to indicate something is either blocking the event loop or the event loop is simply being overloaded.  I don't know how to tell the difference.

```
    INFO  2020-07-25 07:56:39,208    BodyChainGapSyncer  blks=1     txs=151    bps=0    tps=1     elapsed=5.0  head=#10528187 feee..eab8  age=8h59m31s
    INFO  2020-07-25 08:11:10,446   FastChainBodySyncer  blks=0     txs=0      bps=0    tps=0     elapsed=5.0  head=#10529063 a24a..8af6  age=6h2m3s
    INFO  2020-07-25 08:11:18,144    BodyChainGapSyncer  blks=6     txs=1100   bps=0    tps=8     elapsed=6.3  head=#10529000 2637..a0af  age=6h14m53s
    INFO  2020-07-25 08:11:23,145    BodyChainGapSyncer  blks=0     txs=0      bps=0    tps=8     elapsed=5.0  head=#10529000 2637..a0af  age=6h14m58s
    INFO  2020-07-25 08:11:29,744    BodyChainGapSyncer  blks=0     txs=0      bps=0    tps=7     elapsed=6.6  head=#10529000 2637..a0af  age=6h15m4s
    INFO  2020-07-25 08:11:35,771    BodyChainGapSyncer  blks=11    txs=2430   bps=0    tps=27    elapsed=6.0  head=#10529011 a8be..397e  age=6h12m29s
    INFO  2020-07-25 08:11:41,881    BodyChainGapSyncer  blks=5     txs=912    bps=0    tps=33    elapsed=6.1  head=#10529016 f5b2..7bdd  age=6h11m47s
    INFO  2020-07-25 08:11:46,881    BodyChainGapSyncer  blks=7     txs=1077   bps=0    tps=42    elapsed=5.0  head=#10529023 4e72..9fef  age=6h10m54s
    INFO  2020-07-25 08:11:52,662    BodyChainGapSyncer  blks=5     txs=722    bps=0    tps=46    elapsed=5.8  head=#10529028 3c82..a747  age=6h10m5s
[...]

PR #1970, opened 2020-08-05
Run a task to try and detect when the event loop is overloaded

gsalgado:
Every component now runs a task that wakes up periodically and
emits a warning log with some statistics if the difference between the
time it should have executed and the actual time it was executed is
greater than a certain threshold.

Closes: #1962

### DIFF
--- a/trinity/extensibility/asyncio.py
+++ b/trinity/extensibility/asyncio.py
@@ -1,4 +1,5 @@
 from abc import abstractmethod
+import asyncio
 from typing import (
     Any,
     Callable,
@@ -7,18 +8,20 @@ from typing import (
 from asyncio_run_in_process import run_in_process
 from asyncio_run_in_process.typing import SubprocessKwargs
 from async_service import background_asyncio_service
+from async_service.asyncio import cleanup_tasks
 from lahja import EndpointAPI
 
+from p2p.asyncio_utils import create_task
 
-from trinity._utils.logging import child_process_logging, get_logger
+from trinity._utils.logging import child_process_logging
 from trinity._utils.profiling import profiler
+from trinity._utils.timer import Timer
 
 from .component import BaseIsolatedComponent, TReturn
 from .event_bus import AsyncioEventBusService
 
 
 class AsyncioIsolatedComponent(BaseIsolatedComponent):
-    logger = get_logger('trinity.extensibility.asyncio.AsyncioIsolatedComponent')
 
     async def _run_in_process(
             self,
@@ -28,6 +31,28 @@ class AsyncioIsolatedComponent(BaseIsolatedComponent):
     ) -> TReturn:
         return await run_in_process(async_fn, *args, subprocess_kwargs=subprocess_kwargs)
 
+    # FIXME: Disabled by default
+    async def _loop_monitoring_task(self) -> None:
+        debug_log_since = 0
+        while True:
+            timer = Timer()
+            await asyncio.sleep(self.loop_monitoring_wakeup_interval)
+            delay = timer.elapsed - self.loop_monitoring_wakeup_interval
+            # FIXME: Ideally we'd use a debug2 here, but because of
+            # https://github.com/ethereum/trinity/issues/1971 we need to use debug so only log
+            # every 10th time we're called.
+            debug_log_since += 1
+            if debug_log_since == 10:
+                self.logger.debug("Loop monitoring task called; delay=%.3fs", delay)
+                debug_log_since = 0
+            if delay > self.loop_monitoring_max_delay:
+                pending_tasks = len([task for task in asyncio.all_tasks() if not task.done()])
+                self.logger.warning(
+                    "Event loop blocked or overloaded: delay=%.3fs, tasks=%d",
+                    delay,
+                    pending_tasks,
+                )
+
     async def _do_run(self) -> None:
         with child_process_logging(self._boot_info):
             endpoint_name = self.get_endpoint_name()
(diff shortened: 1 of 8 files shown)


trinity/extensibility/asyncio.py
```diff
@@ -28,35 +31,55 @@ async def _run_in_process(
     ) -> TReturn:
         return await run_in_process(async_fn, *args, subprocess_kwargs=subprocess_kwargs)
 
+    # FIXME: Disabled by default
```
gsalgado, line 34: At first I thought we wouldn't want this to run by default, but since it's not doing any expensive operations, I wonder if for now we should run it by default?

trinity/extensibility/asyncio.py
```diff
@@ -28,35 +31,55 @@ async def _run_in_process(
...
                 endpoint_name,
             )
-            async with background_asyncio_service(event_bus_service):
-                event_bus = await event_bus_service.get_event_bus()
+            loop_monitoring_task = create_task(
+                self._loop_monitoring_task(),
+                f'AsyncioIsolatedComponent/{self.name}/loop_monitoring_task')
+            # FIXME: Must terminate component if loop_monitoring_task terminates.
+            async with cleanup_tasks(loop_monitoring_task):
+                # FIXME: Must terminate component if event_bus_service terminates.
```
gsalgado, line 60: I tried to include a fix for the two FIXMEs above but ran into other issues, so decided to do them in a separate PR

trinity/extensibility/asyncio.py
pipermerriam: Maybe this goes on the list of things we look at disabling once we're stable (like how we write DEBUG logs to disk by default).  I'm :+1: on enabling by default since it gives us valuable feedback on a category of performance that I think would be difficult to get visibility on otherwise.

trinity/extensibility/asyncio.py
```diff
@@ -28,35 +31,55 @@ async def _run_in_process(
...
+    # FIXME: Disabled by default
+    async def _loop_monitoring_task(self) -> None:
+        while True:
+            timer = Timer()
+            await asyncio.sleep(self.loop_monitoring_wakeup_interval)
+            variance = timer.elapsed - self.loop_monitoring_wakeup_interval
+            if variance > self.loop_monitoring_max_variance:
+                pending_tasks = len([task for task in asyncio.all_tasks() if not task.done()])
+                self.logger.warning(
+                    "Event loop blocked or overloaded: variance=%f, tasks=%d",
```
pipermerriam, line 43: After sitting on this a bit, I think we want one more metric here and some extra logic about when to log this warning.

For the logging, I think we want to reduce how often this gets dumped to warning since it would effectively be dumped every 2 seconds if our loop is consistently blocked.   Maybe update with some code to log to `DEBUG` if a message was logged to warning with the last few minutes.

I also think we might like a little extra information here on how *often* this is occuring.  Basically, something like a percentage of the time that the loop ends up blocked.  We could use `p2p.stats.ema.EMA` and record a `0` when the loop doesn't block and a `100` when the loop does block.  This would gives us a percentage to include in this message which gives us information that not only is the loop being blocked, but how often.

gsalgado:
Some notes after running this for a little while

A 0.1s threshold may be too small as we end up hitting that somewhat regularly on the Discovery service. But maybe that won't be an issue if we introduce some logic to control how often we log warnings, as @pipermerriam suggested above. Also, in this limited testing, all warnings from the discovery service had a variance < 0.15s

> Discovery  Event loop blocked or overloaded: variance=0.120986, tasks=52, stats=_EpollStatistics(tasks_waiting_read=15, tasks_waiting_write=0, backend='epoll')

The Sync/PeerPool component also goes above the threshold all the time, and it always has a very high task count, although the variance doesn't seem to be related to that:

>  WARNING  2020-08-06 09:31:17,344       Sync / PeerPool  Event loop blocked or overloaded: variance=26.083164, tasks=1171
>  WARNING  2020-08-06 09:31:29,516       Sync / PeerPool  Event loop blocked or overloaded: variance=0.157939, tasks=1201

There were no other warnings between the two ones above, so the monitoring task ran at least 5 times without any delays. I wonder what most of those tasks are waiting for, because if they were being scheduled regularly I believe I'd have seen more warnings as with more than 1300 tasks we seem to get scheduling delays consistently (https://github.com/ethereum/trinity/issues/1962#issuecomment-669722191)

The number of active tasks from the Sync/PeerPool component is almost always growing; early on we seem to have more tasks terminating than being started (as you can see below), but eventually that changes and every warning reports more active tasks than the previous one. I'd be willing to bet there's a leak leading to that

Also, the first warning was reported with just over 200 tasks, and at one point, with just 383 tasks there was a 6s delay

>  WARNING  2020-08-06 08:20:33,832       Sync / PeerPool  Event loop blocked or overloaded: variance=1.358338, tasks=209
 WARNING  2020-08-06 08:20:35,959       Sync / PeerPool  Event loop blocked or overloaded: variance=0.103121, tasks=340
 WARNING  2020-08-06 08:20:41,252       Sync / PeerPool  Event loop blocked or overloaded: variance=3.284416, tasks=315
 WARNING  2020-08-06 08:20:51,309       Sync / PeerPool  Event loop blocked or overloaded: variance=6.043910, tasks=383
 WARNING  2020-08-06 08:20:58,897       Sync / PeerPool  Event loop blocked or overloaded: variance=1.586895, tasks=279


> WARNING  2020-08-06 10:05:13,708       Sync / PeerPool  Event loop blocked or overloaded: variance=0.167870, tasks=2230
 WARNING  2020-08-06 10:05:22,550       Sync / PeerPool  Event loop blocked or overloaded: variance=0.828003, tasks=2240
 WARNING  2020-08-06 10:05:46,993       Sync / PeerPool  Event loop blocked or overloaded: variance=0.312503, tasks=2244
 WARNING  2020-08-06 10:06:05,113       Sync / PeerPool  Event loop blocked or overloaded: variance=0.103041, tasks=2264
 WARNING  2020-08-06 10:06:15,646       Sync / PeerPool  Event loop blocked or overloaded: variance=0.512331, tasks=2269
 WARNING  2020-08-06 10:06:21,992       Sync / PeerPool  Event loop blocked or overloaded: variance=0.328659, tasks=2269
 WARNING  2020-08-06 10:06:24,785       Sync / PeerPool  Event loop blocked or overloaded: variance=0.773987, tasks=2271
 WARNING  2020-08-06 10:06:51,309       Sync / PeerPool  Event loop blocked or overloaded: variance=0.498698, tasks=2293
 WARNING  2020-08-06 10:07:47,506       Sync / PeerPool  Event loop blocked or overloaded: variance=0.107723, tasks=2315
 WARNING  2020-08-06 10:08:03,866       Sync / PeerPool  Event loop blocked or overloaded: variance=0.335408, tasks=2334

gsalgado:
I thought the number of tasks could be related to the number of connected peers, but that doesn't seem to be the case as early on there were 4 connected peers and only ~400 tasks, but towards the end there was ~2k tasks with just one peer

gsalgado:
Just saw this massive drop on the number of active tasks after beam sync pivoted. Thought it was worth mentioning

> WARNING  2020-08-06 16:37:57,682       Sync / PeerPool  Event loop blocked or overloaded: variance=0.262227, tasks=3178
>  WARNING  2020-08-06 16:38:08,307       BeamSyncService  Beam Sync is lagging by 81 blocks. Pivoting...
>  INFO  2020-08-06 16:38:08,307       BeamSyncService  Pivoting Beam Sync to a newer header...
>  WARNING  2020-08-06 16:38:08,347     HeaderOnlyPersist  Tip <BlockHeader #10605074 b5d3e9f0> is too far behind to Beam Sync, skipping ahead...
    INFO  2020-08-06 16:38:08,985    BodyChainGapSyncer  blks=0     txs=0      bps=0    tps=0     elapsed=5.0  head=#10604766 f679..1412  age=-1y11m4w
    INFO  2020-08-06 16:38:10,741     HeaderOnlyPersist  Caught up to skeleton peer. Switching to beam mode at <BlockHeader #10605158 73c32942>
    INFO  2020-08-06 16:38:11,057            BeamSyncer  Downloading 8 block bodies for uncle validation, before <BlockHeader #10605158 73c32942>
 WARNING  2020-08-06 16:38:20,664       Sync / PeerPool  Event loop blocked or overloaded: variance=0.919026, tasks=322

pipermerriam:
The information coming out of this looks helpful.

trinity/extensibility/asyncio.py
carver: IMO, the log would be a little easier to read with only 3 digits and a unit. Also, "variance" in this context has a name collision with the square of the standard deviation. Maybe "lag" or "delay"? Something like:
```suggestion
                    "Event loop blocked or overloaded: delay=%.3fs, tasks=%d",
```


> Maybe update with some code to log to `DEBUG` if a message was logged to warning with the last few minutes.

Yeah, the `TokenBucket` is used for this in a couple places, and has worked well. We can have a capacity of 5ish and a slow refill rate (~1 per 5 minute). Then you several logs in a row at the beginning to show you that it's really blocked, but then the logs quiet down for long runs. (unless the event loop cleared up and then got blocked again)

That said, there's clearly really useful stuff coming out of these logs already. So I'd definitely prioritize getting this merged over tinkering for too long with the `TokenBucket`.

carver:
> Just saw this massive drop on the number of active tasks after beam sync pivoted. Thought it was worth mentioning
> 
> > WARNING  2020-08-06 16:37:57,682       Sync / PeerPool  Event loop blocked or overloaded: variance=0.262227, tasks=3178
> > ...
> > WARNING  2020-08-06 16:38:20,664       Sync / PeerPool  Event loop blocked or overloaded: variance=0.919026, tasks=322

Funny though that the delay was higher with 1/10th the tasks. The correlation doesn't look super high.

trinity/extensibility/asyncio.py
gsalgado: Also, this is still running as an unsupervised background task and we have no tests for it, so if I introduce bugs it'd stop working and we wouldn't find out easily. I'll merge it with @carver's suggestion, then work on a test and try to address the FIXME so that we terminate the component in case the monitoring task terminates early. Then I can work on @pipermerriam's suggestions knowing that at least regressions that cause it to crash will be caught by tests

gsalgado:
I'm also including a change to emit a debug log every 10th time the task is called, regardless of the delay

Issue #1971, opened 2020-08-07
BaseComponent.logger gets overwritten with a Logger instance


gsalgado:
Inside `BaseComponent.run_process()` self.logger is an `ExtendedDebugLogger`, but then in `BaseComponent._do_run()` it is a `Logger`, so any calls to `debug2()` crash.

Need to figure out what's causing that and fix it

pipermerriam:
this might be relevant: https://github.com/ethereum/eth-utils/blob/master/eth_utils/logging.py#L66-L86

gsalgado:
This is actually caused by the way pickling is implemented for `logging.Logger`: https://github.com/python/cpython/blob/master/Lib/logging/__init__.py#L1740-L1746

Our components are implemented as methods on `*IsolatedComponent`, so when we pass those methods to `run_in_process()`, the `*IsolatedComponent` instance will be pickled (together with the logger) and sent to the subprocess, but in the subprocess the logger will be recreated by calling `getLogger(name)` (the return value of `logging.Logger.__reduce__()`, so it will be a `Logger` instance.

This is not a problem if we pickle/unpickle in the same process as the `ExtendedDebugLogger` instance will be cached in the logging module, so the `getLogger(name)` call will return that

Maybe what we need is a custom `ExtendedDebugLogger.__reduce__()`:

```
    def __reduce__(self):
        return get_extended_debug_logger, (self.name,)
```

PR #1974, opened 2020-08-10
A couple minor improvements to *IsolatedComponent

gsalgado:
- Emit a debug2() log every time the loop monitoring task is called
- Remove unused args from _run_in_process()

Closes: #1971

### DIFF
--- a/trinity/protocol/les/commands.py
+++ b/trinity/protocol/les/commands.py
@@ -12,7 +12,8 @@ from eth_utils import (
     to_tuple,
     ValidationError,
 )
-from eth_utils.curried import (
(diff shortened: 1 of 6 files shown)


trinity/protocol/les/commands.py
```diff
@@ -12,7 +12,8 @@
     to_tuple,
     ValidationError,
 )
-from eth_utils.curried import (
+# For some weird reason mypy thinks eth_utils.curried doesn't have those, but it clearly does.
+from eth_utils.curried import (  # type: ignore
```
pipermerriam, line 16: Can you open an issue to track this so we can figure out why it is happening.

gsalgado: #1982

PR #1976, opened 2020-08-11
Ignore delay when inserting a queen-able peasant (plus bonus code)

carver:
### What was wrong?

If we are blocked waiting for a queen, we want to resolve the situation as quickly as possible. (or if we have a candidate who is a faster queen) Currently, when we re-insert a peasant with a delay, it is always delayed.

### How was it fixed?

So if we are inserting a peasant, but it will be promoted to queen, then insert it immediately, even if the insert was called with a delay. This helps unblock the critical path to collect urgent trie nodes.

Bonus additions:
- Don't block the critical path of collecting urgent trie nodes on a missing queen peer as often
- Better tracking of the number of trie nodes received. (Before, during a timeout, we would consider the nodes as received in the returned count of received nodes) Mostly just relevant to logging.
- Release event loop in state backfill more often.

Changes are nicely isolated by commit.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/backfill.py
+++ b/trinity/sync/beam/backfill.py
@@ -382,6 +382,10 @@ class BeamStateBackfill(Service, QueenTrackerAPI):
                 # If this is just an intermediate node, then we can mark it as confirmed.
                 request_tracker.confirm_prefix(path_to_node, node)
 
+            # We made some DB calls and didn't find anything missing. Yield to the event loop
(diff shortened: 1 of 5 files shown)


carver:
Snuck in one more change. The short timeouts weren't really accomplishing anything, except helping us see how backed up that the event loop & event bus gets sometimes. So increased it quite a bit.

PR #1978, opened 2020-08-13
Add a basic test for the loop monitoring task in *IsolatedComponent

gsalgado:


trinity/extensibility/trio.py
```diff
@@ -43,8 +43,8 @@ async def run_process(self, event_bus: EndpointAPI) -> None:
             # These are expected, when trinity is terminating because of a Ctrl-C
             raise
 
-    # FIXME: Disabled by default
-    async def _loop_monitoring_task(self) -> None:
+    # The EndpointAPI argument here is currently only used by tests.
+    async def _loop_monitoring_task(self, _: EndpointAPI) -> None:
```
cburgdorf, line 47: Did you consider having an `asyncio` / `trio` event on the component and calling `set()` on it inside this method. That way we could get rid of the `EndpointAPI` that is only used for testing I think.

gsalgado: The `*IsolatedComponent` runs in a subprocess as we're using the ComponentManager in the test, so it'd get a different `Event` than whatever we create in the test.

Issue #1613, opened 2020-03-18
Shutdown issues, a braindump


gsalgado:
I've been trying to fix our shutdown so that all tasks started by our asyncio services are terminated and the components exit cleanly without the need for us to kill them with a SIGTERM, but I've been at this for too long now and have identified several issues that I may not be able to fix in one go, so I thought I'd write them down here

~First issue, affecting all our components: https://github.com/ethereum/async-service/issues/62~ (not really an issue)

Another issue is that some of our services use the `run_asyncio_service()` API, which [doesn't wait for the service to finish after `manager.run()` returns](https://github.com/ethereum/async-service/issues/62#issuecomment-600465685). For those services, the `KeyboardInterrupt` comes from the [`Manager.run()` method](https://github.com/ethereum/async-service/blob/master/async_service/asyncio.py#L213)

Also, for all services using the `background_asyncio_service()` API, I get the KeyboardInterrupt in https://github.com/ethereum/async-service/blob/master/async_service/asyncio.py#L414 (instead of [`_run_and_manage_task()`](https://github.com/ethereum/async-service/blob/master/async_service/base.py#L299) as I'd have expected). When that happens, `background_asyncio_service()` moves on to wait for `manager.stop()`, which seemed to hang forever because the only place where we flag a service as finished is in `Manager.run()`, which at this point is no longer running. I thought this one could be easily fixed by changing `_handle_cancelled()` to flag the service as finished once all tasks are finished, but after making that change I realized the `manager.stop()` still didn't seem to return, and after digging further I noticed it was actually raising yet another `KeyboardInterrupt`, and that seems to happen for any async calls inside the `finally` clause of `background_asyncio_service()`. Need to investigate this further unless somebody can shed a light on what's going on here.

~Finally, ISTM that the [AsyncExitStack we use in the ComponentManager](https://github.com/ethereum/trinity/blob/master/trinity/extensibility/component_manager.py#L87-L101) causes us to wait for the components to finish in sequence, and I don't see why we'd want to do that.~ (fixed in #1650)

cburgdorf:
Just pinging @pipermerriam here as he's been the last one who worked on that and I think he formed an opinion about how things *should be* but then got dragged into other things.

pipermerriam:
The sequential unwinding of components by `AsyncExitStack` isn't intentional and seems like it should be reasonably easy to fix.

[...]

PR #1979, opened 2020-08-13
Components now terminate if the eventbus or loop monitoring task exit

gsalgado:


p2p/asyncio_utils.py
```diff
@@ -41,12 +43,19 @@ async def wait_first(tasks: Sequence[asyncio.Task[Any]]) -> None:
...
+    logger = get_logger('p2p.asyncio_utils.wait_first')
     try:
         done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
-    except asyncio.CancelledError:
+    except (KeyboardInterrupt, asyncio.CancelledError) as err:
+        logger.debug("Got %r waiting for %s, cancelling them all", err, tasks)
+        await cancel_tasks(tasks)
+        raise
+    except BaseException:
+        logger.exception("Unexpected error waiting for %s", tasks)
```
gsalgado, line 54: This is to ensure we always cancel all tasks before returning. The `KeyboardInterrupt` happens when trinity is shutting down and `asyncio-rip` injects it in each of our components, which now bloc on a `wait_first()` call to ensure they terminate if either the component itself, the event bus service or the loop monitoring task terminate

trinity/extensibility/component_manager.py
```diff
@@ -82,24 +83,26 @@ async def run(self) -> None:
...
             # logging messages about the component lifecycle.
-            from p2p.asyncio_utils import wait_first
+            from p2p.asyncio_utils import create_task, wait_first
             try:
                 self.logger.info(
                     "Starting components: %s",
                     '/'.join(component.name for component in enabled_components),
                 )
-                context_managers = [component.run() for component in enabled_components]
-                async with AsyncContextGroup(context_managers) as futs:
```
carver, line 92: Is `AsyncContextGroup` unusable or generally a worse solution than this pattern? If so, then maybe it's time to remove it altogether.

p2p/asyncio_utils.py
```diff
@@ -41,12 +43,19 @@ async def wait_first(tasks: Sequence[asyncio.Task[Any]]) -> None:
...
         await cancel_tasks(tasks)
         raise
     else:
+        logger.debug("Task %s finished, cancelling remaining ones: %s", done, pending)
```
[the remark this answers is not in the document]
gsalgado: Right now we will see this when a Connection terminates, so shouldn't be frequent at all. In all other cases where this is used, trinity will be terminated when we get to this point

trinity/extensibility/component_manager.py
gsalgado: Our context managers (the now removed `run()` methods) were simply yielding a task and not performing any cleanups on exit, so there was no point in having them as context managers and using ACG. We could probably remove it but since we should probably always use that instead of `AsyncExitStack` (https://github.com/ethereum/trinity/issues/1613), it might be worth keeping it?

Issue #1980, opened 2020-08-13
Running list of coros that hold the event loop for too long


carver:
These coros have held the event loop for too long, and need to be more cooperative. The longest known blocking time is included:
- Last verified at a2c69cee88626c9
  - [x] `BasePeerPool._periodically_report_stats() running at /home/jcarver/code/trinity/p2p/peer_pool.py:553` 267.3s #1989 
  - [x] `BeamDownloader._store_nodes() done, defined at /home/jcarver/code/trinity/trinity/sync/beam/state.py:531` 6.0s #1984 
  - [x] `BeamBlockImporter._preview_address_load() running at /home/jcarver/code/trinity/trinity/sync/beam/chain.py:811` 5.2s #1984 
  - [x] `BeamStateBackfill.run() running at /home/jcarver/code/trinity/trinity/sync/beam/backfill.py:142` 2.9s #1983 
  - [x] `_run() running at /home/jcarver/code/trinity/trinity/bootstrap.py:275` 0.3s
  - [ ] `coro=<GetNodeDataV65Exchange.__call__() running at /home/jcarver/code/trinity/trinity/protocol/eth/exchanges.py:105>  28.395 seconds`
  - [ ] `coro=<BeamDownloader._get_nodes() done, defined at /home/jcarver/code/trinity/trinity/sync/beam/state.py:571>  26.139 seconds`
  - [ ] `coro=<PingAndDisconnectIfIdle.run.<locals>.set_msg_received() done, defined at /home/jcarver/code/trinity/p2p/p2p_api.py:47>  26.128 seconds`
  - [ ] `<Handle _chain_future.<locals>._set_state created at /home/jcarver/.pyenv/versions/3.8.5/lib/python3.8/asyncio/futures.py:374>  24.48 seconds`
  - [ ] `coro=<BasePeer._handle_subscriber_message() done, defined at /home/jcarver/code/trinity/p2p/peer.py:324>  21.53 seconds`
  - [ ] `coro=<BeamDownloader._node_hashes_present() running at /home/jcarver/code/trinity/trinity/sync/beam/state.py:649>  19.858 seconds`
  - [ ] `coro=<Multiplexer._do_multiplexing() running at /home/jcarver/code/trinity/p2p/multiplexer.py:342>  18.585 seconds` #1991
[...]

PR #1983, opened 2020-08-14
De-async the backfill search for missing nodes

carver:
### What was wrong?

Related to #1980 

`_missing_trie_nodes()` was holding onto the event loop for too long.

### How was it fixed?

There wasn't really a need for it to be async, so synchronize the search for missing nodes, run the method it in a threadpool, and await that.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/backfill.py
+++ b/trinity/sync/beam/backfill.py
@@ -5,9 +5,9 @@ from functools import partial
 import itertools
 import typing
 from typing import (
-    AsyncIterator,
     Dict,
     Iterable,
+    Iterator,
     NamedTuple,
     Optional,
     Set,
@@ -23,6 +23,7 @@ from eth.abc import (
 from eth.constants import EMPTY_SHA3
 from eth.rlp.accounts import Account
 from eth_typing import Hash32
+from eth_utils.toolz import take
 import rlp
 from trie import (
     HexaryTrie,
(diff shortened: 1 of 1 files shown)


trinity/sync/beam/backfill.py
```diff
@@ -174,14 +172,22 @@ async def _run_backfill(self) -> None:
...
+                    self._external_peasant_usage.until_silence(),
+                    self.manager.wait_finished(),
+                ),
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+            if not self.manager.is_running:
+                break
+
+            peer = await self._queening_queue.pop_fastest_peasant()
+
```
carver, line 190: This change looks bigger than it really is, because there was a reordering. Most of the important change is the new `run_in_executor` call.

trinity/sync/beam/backfill.py
```diff
@@ -236,15 +242,15 @@ async def _missing_trie_hashes(self) -> AsyncIterator[TrackedRequest]:
                     starting_root_hash,
                 )
                 try:
-                    next_account_info = await account_iterator.__anext__()
+                    next_account_info = account_iterator.__next__()
```
pipermerriam, line 245: why this and not `next(account_iterator)`?

carver: No good reason. I was just in "delete the async bits" mode. :)  Will update.

PR #1984, opened 2020-08-14
Save trie nodes outside the event loop

carver:
### What was wrong?

Related to #1980 

The event loop is held by `_store_nodes()` for too long.

### How was it fixed?

Save trie nodes outside the event loop. Also, compressed all the `asyncio.Event`s that notify about new data being received down to a single one. I don't see/remember the benefit to having one for each waiting coro.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/state.py
+++ b/trinity/sync/beam/state.py
@@ -117,7 +117,7 @@ class BeamDownloader(Service, PeerSubscriber):
         self._node_tasks = TaskQueue[Hash32](buffer_size, lambda task: 0)
 
         # list of events waiting on new data
-        self._new_data_events: Set[asyncio.Event] = set()
+        self._new_data_event: asyncio.Event = asyncio.Event()
 
         self._peer_pool = peer_pool
 
@@ -533,8 +533,7 @@ class BeamDownloader(Service, PeerSubscriber):
         #   That way, urgent coros don't get stuck hanging until a timeout. This can cause an
         #   especially flaky test_beam_syncer_backfills_all_state[42].
         if urgent or new_nodes:
-            for new_data in self._new_data_events:
-                new_data.set()
+            self._new_data_event.set()
 
         return nodes, new_nodes, peer
 
@@ -552,18 +551,14 @@ class BeamDownloader(Service, PeerSubscriber):
         """
         remaining_hashes = node_hashes.copy()
 
-        # save an event that gets triggered when new data comes in
-        new_data = asyncio.Event()
-        self._new_data_events.add(new_data)
-
         start_time = time.monotonic()
         while remaining_hashes and time.monotonic() - start_time < timeout:
-            await new_data.wait()
+            await self._new_data_event.wait()
 
             found_hashes = set(found for found in remaining_hashes if self._is_node_present(found))
             remaining_hashes -= found_hashes
 
-            new_data.clear()
+            self._new_data_event.clear()
 
         if remaining_hashes:
             self.logger.error(
@@ -573,7 +568,6 @@ class BeamDownloader(Service, PeerSubscriber):
                 time.monotonic() - start_time,
             )
 
-        self._new_data_events.remove(new_data)
         return len(node_hashes) - len(remaining_hashes)
 
     def register_peer(self, peer: BasePeer) -> None:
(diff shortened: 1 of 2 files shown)


carver:
Hm, not having the desired effect. Still holding the event loop for too long. In two different parts of `_get_nodes()`:
- `coro=<BeamDownloader._get_nodes() done, defined at /home/jcarver/code/trinity/trinity/sync/beam/state.py:511>  4.692`
- `coro=<BeamDownloader._get_nodes() running at /home/jcarver/code/trinity/trinity/sync/beam/state.py:516>  4.991`

carver:
This one also prevents `BeamBlockImporter._preview_address_load()` from holding the event loop for too long.

trinity/sync/beam/state.py
```diff
@@ -185,6 +184,16 @@ def _max_spread_beam_factor(self) -> int:
         max_factor = self._num_peers - 1 - self._min_predictive_peers
         return max(0, max_factor)
 
+    def _unique_missing_hashes(self, hashes: Iterable[Hash32]) -> Set[Hash32]:
```
gsalgado, line 187: Maybe `_get_unique_missing_hashes()`, so that we follow the standard practice of using verbs in method names? Otherwise when reading the code it looks like just an attribute

trinity/sync/beam/state.py
```diff
@@ -185,6 +184,16 @@ def _max_spread_beam_factor(self) -> int:
...
+        return set(
+            node_hash for node_hash in hashes if node_hash not in self._db
+        )
+
+    def _unique_present_hashes(self, hashes: Iterable[Hash32]) -> Set[Hash32]:
```
gsalgado, line 192: Same here

trinity/sync/beam/state.py
```diff
@@ -209,13 +218,30 @@ async def _wait_for_nodes(
         else:
             return 0
 
-    def _is_node_missing(self, node_hash: Hash32) -> bool:
-        if len(node_hash) != 32:
-            raise ValidationError(f"Must request node by its 32-byte hash: 0x{node_hash.hex()}")
```
gsalgado, line 214: Are we happy to rely only on mypy to ensure callsites are not passing invalid node hashes? I'd be more comfortable keeping this validation somewhere

trinity/sync/beam/state.py
```diff
@@ -209,13 +218,30 @@ async def _wait_for_nodes(
...
+                except MissingTrieNode as exc:
+                    need_nodes.add(exc.missing_node_hash)
+                else:
+                    completed_account_hashes.add(account_hash)
 
-        self.logger.debug2("checking if node 0x%s is present", node_hash.hex())
+        return need_nodes, completed_account_hashes
 
-        return node_hash not in self._db
+    def _unique_hashes(self, addresses: Collection[Address]) -> Set[Hash32]:
```
gsalgado, line 242: `_get_unique_hashes()`?

trinity/sync/beam/state.py
```diff
@@ -552,19 +609,19 @@ async def _node_hashes_present(self, node_hashes: Set[Hash32], timeout: float) -
...
         start_time = time.monotonic()
         while remaining_hashes and time.monotonic() - start_time < timeout:
-            await new_data.wait()
+            await self._new_data_event.wait()
+            self._new_data_event.clear()
 
-            found_hashes = set(found for found in remaining_hashes if self._is_node_present(found))
+            found_hashes = await loop.run_in_executor(
+                None,
+                self._unique_present_hashes,
```
gsalgado, line 620: It's not clear to me how many of these `run_in_executor()` may run in parallel, but after https://github.com/ethereum/asyncio-run-in-process/pull/26 I'm always weary of using the default executor, both because it may change between python versions but also because we may inadvertently end up having resource contention issues if many parts of the component use the default executor. Don't necessarily need to be dealt with now; I'd be happy with just a TODO/FIXME comment

trinity/sync/beam/state.py
carver: On balance, my thinking is that if we're going to go to all the effort to maintain the typing, then we might as well get a performance win.

Is the concern that mypy will be buggy and not catch an issue?

PR #1989, opened 2020-08-18
Don't block for long on generating a log

carver:
### What was wrong?

Related to #1980 

Slow log emission was causing `BasePeerPool._periodically_report_stats()` to hold the event loop for too long.

### How was it fixed?

Push all logs into a queue, and push them into the socket in a different thread.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/_utils/logging.py
+++ b/trinity/_utils/logging.py
@@ -2,14 +2,17 @@ import contextlib
 import copy
 import logging
 from logging import (
-    StreamHandler
+    StreamHandler,
 )
 from logging.handlers import (
+    QueueHandler,
+    QueueListener,
     RotatingFileHandler,
 )
 import os
 from pathlib import Path
 import pickle
+from queue import Queue
 import socket
 import sys
 from types import TracebackType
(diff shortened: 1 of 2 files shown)


trinity/_utils/logging.py
```diff
@@ -225,7 +228,19 @@ def child_process_logging(boot_info: BootInfo) -> Iterator[None]:
...
+    log_queue: Queue[str] = Queue(-1)
+    queue_handler = QueueHandler(log_queue)
+    queue_listener = QueueListener(
+        log_queue,
+        ipc_handler,
+        respect_handler_level=True,
+    )
+    queue_listener.start()
+
+    logger.addHandler(queue_handler)
```
pipermerriam, line 243: Doesn't need to block this PR, but worth pointing out that in the event of an unfortunate interrupt between here and the `try/except` below we won't call `queue_listener.stop()`.  I don't know what the implications of that are.

carver: :+1: It's an easy enough fix

PR #1990, opened 2020-08-19
Move more Beam state database checks to thread pool

carver:
### What was wrong?

Related to #1980 

`BeamDownloader._download_account` was holding the event loop for too long. I haven't observed the same problem with `_download_storage` yet, but it was bound to happen.

### How was it fixed?

This is all of the database lookups I see in `BeamDownloader`. They are now all moved to a thread pool

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


carver:
Hm docs are failing on master too, but not the latest pull request: https://readthedocs.org/projects/trinity-client/builds/11699537/

PR #1992, opened 2020-08-20
Let go of the event loop more often

carver:
### What was wrong?

Related to #1980 

### How was it fixed?

- Push some more long-running methods out of the event loop
- Release the event loop in the middle of executing a series of methods

Bonuses:
- Fix bug where `set` was modified while `trie_fog` was using it to mark some prefixes as complete
- Improve the mid-import log during Beam Sync

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/multiplexer.py
+++ b/p2p/multiplexer.py
@@ -294,10 +294,12 @@ class Multiplexer(MultiplexerAPI):
                 try:
                     # We use an optimistic strategy here of using
                     # `get_nowait()` to reduce the number of times we yield to
-                    # the event loop.  Since this is an async generator it will
-                    # yield to the loop each time it returns a value so we
-                    # don't have to worry about this blocking other processes.
+                    # the event loop.
                     yield msg_queue.get_nowait()
+
+                    # Manually release the event loop if it won't happen during
+                    #   the queue get().
+                    await asyncio.sleep(0)
                 except asyncio.QueueEmpty:
                     yield await msg_queue.get()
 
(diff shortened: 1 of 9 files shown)


p2p/multiplexer.py
```diff
@@ -290,15 +290,23 @@ async def _stream_protocol_messages(self,
...
-                    # don't have to worry about this blocking other processes.
+                    # the event loop.
                     yield msg_queue.get_nowait()
+
+                    # Manually release the event loop if too many items have
+                    # been sequentially yielded, without releasing the event loop.
+                    current_batch += 1
+                    if current_batch >= max_batch_size:
+                        await asyncio.sleep(0)
+                        current_batch = 0
```
gsalgado, line 307: Why did you chose to release the event loop every 5 items instead of every single item? To me it'd make more sense to have iterators/generators produce as many items as they can if that doesn't require blocking, and then consumers can decide how to batch. However, @pipermerriam found that `trio.ReceiveChannel` releases on every iteration (https://github.com/ethereum/lahja/pull/179#discussion_r465101643) and maybe we should be consistent with that? Either way, I'm not convinced this is the right place to implement the batching

carver: I was trying to honor the original intent there which seemed to be that it would be a bad idea to release it each time. I don't have a better reason than that, and it would certainly be simpler to just release each time. So I'll update it to release each cycle, if no one thinks it's important to be able to pump out batches of these.

PR #1996, opened 2020-08-25
Discovery: msg handlers now await for replies to be sent

gsalgado:
Our msg handlers already run as background tasks, so it made no sense for
them to run the coroutine that sends the reply over the socket in the
background as well.

Also fixes request_enr() and _lookup() so that they only start waiting
for replies after the ENR/FIND_NODE msgs have been sent.

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -443,7 +443,7 @@ class DiscoveryService(Service):
         """
         # No need to use a timeout because bond() takes care of that internally.
         await self.bond(remote.id)
-        token = self.send_enr_request(remote)
+        token = await self.send_enr_request(remote)
         send_chan, recv_chan = trio.open_memory_channel[Tuple[ENR, Hash32]](1)
         try:
             with trio.fail_after(constants.KADEMLIA_REQUEST_TIMEOUT):
@@ -577,7 +577,7 @@ class DiscoveryService(Service):
         nodes_seen: Set[NodeAPI] = set()
 
         async def _find_node(target: bytes, remote: NodeAPI) -> Tuple[NodeAPI, ...]:
-            self.send_find_node_v4(remote, target)
+            await self.send_find_node_v4(remote, target)
             candidates = await self.wait_neighbours(remote)
             if not candidates:
                 self.logger.debug2("got no neighbors from %s, returning", remote)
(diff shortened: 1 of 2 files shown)


p2p/discovery.py
```diff
@@ -719,13 +719,8 @@ async def bootstrap(self) -> None:
...
-            await self.socket.sendto(msg, (ip, port))
-        except OSError:
-            self.logger.exception("Unexpected error sending msg to %s", (ip, port))
-
-    def send(self, node: NodeAPI, msg_type: DiscoveryCommand, payload: Sequence[Any]) -> bytes:
+    async def send(
+            self, node: NodeAPI, msg_type: DiscoveryCommand, payload: Sequence[Any]) -> bytes:
         """
         Pack the given payload using the given msg type and fire a background task to try and
         send it over our socket.
```
carver, line 726: Comment got stale.

p2p/discovery.py
```diff
@@ -577,7 +577,7 @@ async def _lookup(self, target_key: bytes) -> Tuple[NodeAPI, ...]:
         nodes_seen: Set[NodeAPI] = set()
 
         async def _find_node(target: bytes, remote: NodeAPI) -> Tuple[NodeAPI, ...]:
```
carver, line 579: Is this method even used anywhere? I couldn't find it.

p2p/discovery.py
```diff
@@ -1095,6 +1087,22 @@ async def process_pong_v4(self, remote: NodeAPI, token: Hash32, enr_seq: int) ->
             # This means the receiver has already closed, probably because it timed out.
             pass
 
+    @to_tuple
+    def get_neighbours(self, target_id: NodeID) -> Iterator[NodeAPI]:
```
carver, line 1091: Why move this method around? It made it tougher to see what changes there were, if any. (Looks like nothing changed?)

p2p/discovery.py
gsalgado: It's used a few lines down, in

> results = await trio_utils.gather(*next_find_node_queries)

p2p/discovery.py
gsalgado: Nothing changed indeed, I've only moved it so that all msg handlers were next to each other. I usually mention those things, but this time forgot about it, sorry

gsalgado:
> > Our msg handlers already run as background tasks, so it made no sense for them to run the coroutine that sends the reply over the socket in the background as well.
> 
> Hm, but this change affects _every_ send, not just sending replies in response to incoming requests. I tried to follow through some of the `.send()` calls that were not in response to a message, like requesting the ENR. It seems like the couple I looked at were okay, but it would be an easy mistake to make.

That's right, and as I've mentioned in the PR's description, it is intentional as it ensures we only start waiting for ENR/NEIGHBORS messages after the request has actually been sent. There are no other calls to `.send()`, btw
 
> I do not know how the trio `socket.send()` works under the hood, so I'm not sure how big a problem it is. Does it hold the event loop for a long time while sending a message? Is there an equivalent of `PYTHONASYNCIODEBUG=1` that you can run to check if there are any problems with holding the event loop?

`socket.sendto()` is async, it won't hold the loop. This change is about having our coroutines waiting for that to complete before moving on instead of firing it as a background task (via `.run_task()`) and moving on before the message is actually sent.

p2p/discovery.py
carver: OOOh, I was missing that it was an internally defined function instead of an instance method, so the `_find_node` reference looked like it was for some other thing. :+1:

carver:
Cool, sounds good. Still looks good to merge (except for the stale comment). :+1:

p2p/discovery.py
```diff
@@ -719,32 +719,31 @@ async def bootstrap(self) -> None:
...
         """
-        Pack the given payload using the given msg type and fire a background task to try and
-        send it over our socket.
+        Pack the given payload using the given msg type and send it over our socket.
 
         If we get an OSError from our socket when attempting to send it, that will be logged
         and the message will be lost.
         """
         message = _pack_v4(msg_type.id, payload, self.privkey)
-        self.manager.run_task(self._sendto, message, node.address.ip, node.address.udp_port)
```
pipermerriam, line 737: My gut says that this gains us some performance as well, since `async_service` tasks do have some measurable level of overhead in their lifecycle management machinery.

Issue #1999, opened 2020-08-27
Unexpected error handling handshake on inbound peer connections


gsalgado:
Recently started seeing those somewhat frequently

```
   ERROR  2020-08-27 10:51:57,324            FullServer  Unexpected error handling handshake with ('47.244.135.142', 34823)
Traceback (most recent call last):
  File "/home/salgado/src/snakecharmers/trinity/trinity/server.py", line 136, in receive_handshake
    await self._receive_handshake(reader, writer)
  File "/home/salgado/src/snakecharmers/trinity/trinity/server.py", line 189, in _receive_handshake
    await peer.disconnect(DisconnectReason.TOO_MANY_PEERS)
  File "/home/salgado/src/snakecharmers/trinity/p2p/peer.py", line 338, in disconnect
    self.disconnect_nowait(reason)
  File "/home/salgado/src/snakecharmers/trinity/p2p/peer.py", line 349, in disconnect_nowait
    self._send_disconnect(reason)
  File "/home/salgado/src/snakecharmers/trinity/p2p/peer.py", line 353, in _send_disconnect
    self._p2p_api.disconnect(reason)
AttributeError: 'ETHPeer' object has no attribute '_p2p_api'
```

PR #2001, opened 2020-08-27
Fix a crash in Server._receive_handshake()

gsalgado:
That method would attempt to send a Disconnect msg to the remote
if for any reason we decided not to add the peer to the pool after
a successful handshake, but that would crash because to do so we
need a running Peer service and for that we need to add it to the
pool

Closes: #1999

### DIFF
--- a/trinity/server.py
+++ b/trinity/server.py
@@ -19,7 +19,6 @@ from eth.abc import AtomicDatabaseAPI, VirtualMachineAPI
 from pyformance import MetricsRegistry
 
 from p2p.constants import DEFAULT_MAX_PEERS, DEVP2P_V5
-from p2p.disconnect import DisconnectReason
 from p2p.exceptions import (
     HandshakeFailure,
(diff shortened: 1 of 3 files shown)


trinity/server.py
```diff
@@ -185,10 +186,12 @@ async def _receive_handshake(
                 if peer.inbound
             ))
             if total_peers > 1 and inbound_peer_count / total_peers > DIAL_IN_OUT_RATIO:
-                # make sure to have at least 1/4 outbound connections
```
pipermerriam, line 188: was this comment dropped on purpose?

gsalgado: Yep, I don't think it's needed as the code is clear enough, and also it is only accurate as long as DIAL_IN_OUT_RATIO is set 0.25

gsalgado:
@pipermerriam I thought it'd take a major refactoring but indeed it was straightforward. Care to have another look?

PR #2007, opened 2020-09-01
Sync tweaks

carver:
### What was wrong?

- thread unsafe operation crash
- time spent waiting on DB was affecting our perception of the speed of our queen peer
- bad-acting peasants were un-penalized
- when `cancel_pending_tasks` fails due to timeout in another process, it tries to pickle `asyncio.Task` objects and fails
- the block previews run even when the block is being actively (urgently) imported

### How was it fixed?

- copy dict before iterating it
- measure time waiting on queen with *only* the network request
- penalize bad-acting peasants
- include the `repr` of the `Task` instead of the original object, so it's pickle-able
- cancel the block preview when the same block has started importing urgently

### To-Do

- [x] Clean up commit history

- [x] ~Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)~ None of these bugs were released

#### Cute Animal Picture

### DIFF
--- a/p2p/asyncio_utils.py
+++ b/p2p/asyncio_utils.py
@@ -99,8 +99,11 @@ async def cancel_pending_tasks(*tasks: asyncio.Task[Any], timeout: int) -> Async
             # slow cleanup the othders don't have to wait for it.
             done, pending = await asyncio.wait(cancelled, timeout=timeout)
             if pending:
+                # These tasks might need to be pickled to pass across the process boundary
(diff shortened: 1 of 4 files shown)


carver:
Docs build:
> There was a problem with Read the Docs while building your documentation. Please try again later. However, if this problem persists, please report this to us with your build id (11797065).

pipermerriam (approved):

PR #2010, opened 2020-09-02
convert to use eth-enr library

pipermerriam:
### What was wrong?

The code for handling ENR records has been extracted into a standalone library:  https://github.com/ethereum/eth-enr

### How was it fixed?

Updated the local code to use the API exposed by `eth-enr` as well as removing the discovery v5 implementation rather than updating it (since it is slated to be replaced by https://github.com/ethereum/ddht/)

#### Cute Animal Picture

confused-chimp


pipermerriam:
Waiting to merge until I've also completed the migration in https://github.com/ethereum/ddht/ which is looking like it will fuel at least a patch release in the `eth-enr` library.

gsalgado:
This PR removed `tests-trio/p2p-trio/test_discovery.py`, which was for the discv4 implementation

PR #2011, opened 2020-09-02
Beam Sync download speedups

carver:
Generally, we want to make the urgent nodes resolve as quickly as possible, and keep preview nodes out of the way. So this code splits the two scenarios to make sure we run predictive nodes in a separate thread, and urgent nodes in the event loop.

Also, it was expensive to wake up all those coros every time some new data came in. So now when urgent data comes in, it only wakes up the urgent coros. When predictive data comes in, it only wakes up the relevant predictive coros.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

put a cute animal picture link inside the parentheses/front-view-closeup-of-black-and-white-spotted-piglet-on-hay-on-a-sunny-day-881477548-5c8ba7b746e0fb000187a286.jpg)

### DIFF
--- a/trinity/sync/beam/state.py
+++ b/trinity/sync/beam/state.py
@@ -59,6 +59,7 @@ from trinity.sync.beam.queen import (
 )
 from trinity.sync.beam.constants import (
     BLOCK_IMPORT_MISSING_STATE_TIMEOUT,
+    CHECK_PREVIEW_STATE_TIMEOUT,
     ESTIMATED_BEAMABLE_SECONDS,
     MAX_ACCEPTABLE_WAIT_FOR_URGENT_NODE,
     REQUEST_BUFFER_MULTIPLIER,
@@ -74,6 +75,8 @@ class BeamDownloader(Service, PeerSubscriber):
     _total_processed_nodes = 0
     _urgent_processed_nodes = 0
     _predictive_processed_nodes = 0
+    _predictive_found_nodes_woke_up = 0
+    _predictive_found_nodes_during_timeout = 0
     _total_timeouts = 0
     _predictive_requests = 0
     _urgent_requests = 0
@@ -84,6 +87,8 @@ class BeamDownloader(Service, PeerSubscriber):
 
     _num_urgent_requests_by_peer: typing.Counter[ETHPeer]
     _num_predictive_requests_by_peer: typing.Counter[ETHPeer]
+
+    _preview_events: Dict[asyncio.Event, Set[Hash32]]
 
     _num_peers = 0
     # How many extra peers (besides the queen) should we ask for the urgently-needed trie node?
@@ -118,6 +123,7 @@ class BeamDownloader(Service, PeerSubscriber):
 
         # list of events waiting on new data
         self._new_data_event: asyncio.Event = asyncio.Event()
+        self._preview_events = {}
 
         self._peer_pool = peer_pool
 
@@ -152,14 +158,12 @@ class BeamDownloader(Service, PeerSubscriber):
         if urgent:
             num_nodes_found = await self._wait_for_nodes(
                 node_hashes,
-                self._node_tasks,
-                BLOCK_IMPORT_MISSING_STATE_TIMEOUT,
+                urgent,
             )
         else:
             num_nodes_found = await self._wait_for_nodes(
                 node_hashes,
-                self._maybe_useful_nodes,
-                BLOCK_IMPORT_MISSING_STATE_TIMEOUT,
+                urgent,
             )
 
         return num_nodes_found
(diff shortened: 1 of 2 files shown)


trinity/sync/beam/state.py
```diff
@@ -579,25 +608,51 @@ async def _get_nodes(
...
 
+    def _get_preview_waiters(self, new_nodes: NodeDataBundles) -> Tuple[asyncio.Event, ...]:
+        # defensive copy _preview_events, since this method runs in a thread
+        waiters = tuple(self._preview_events.items())
+
+        new_hashes = set(node_hash for node_hash, _ in new_nodes)
+        relevant_waiters = set(
+            waiting_event
+            for waiting_event, node_hashes in waiters
+            if new_hashes & node_hashes
```
pipermerriam, line 652: This method looks like it could be reworked to have an early exit condition:

```
    def _get_preview_waiters(self, new_nodes: NodeDataBundles) -> Tuple[asyncio.Event, ...]:
        new_hashes = set(node_hash for node_hash, _ in new_nodes)
        if not new_hashes:
            return set()
        # defensive copy _preview_events, since this method runs in a thread
        waiters = tuple(self._preview_events.items())

        relevant_waiters = set(
            waiting_event
            for waiting_event, node_hashes in waiters
            if node_hashes
        )
        return tuple(relevant_waiters)
```

pipermerriam: I also don't understand why the set comprehension is needed (instead of just a tuple comprehesion) since `_preview_events` is a mapping with the `waiting_event` objects as keys, which suggests that there shouldn't be any way for duplicates to show up.  Am I missing something?

pipermerriam (approved):
this was a relatively quick pass through the code.  If you feel reasonably good about this already feel free to :shipit: otherwise I can give you a better review first thing tomorrow when my :brain: is less :fried_egg:

carver:
Yup, fine to wait till morning.

trinity/sync/beam/state.py
```diff
@@ -341,12 +364,18 @@ async def download_storage(
         loop = asyncio.get_event_loop()
         # should never take more than 64 attempts to get a full account
         for num_downloads_required in range(64):
-            need_nodes = await loop.run_in_executor(
-                None,
-                self._storage_review,
-                storage_key,
-                storage_root_hash,
-            )
+            if urgent:
```
cburgdorf, line 367: A lot of these methods follow this pattern now:


```python
if urgent:
    self._some_method(params)
else:
     loop.run_in_excecutor(None, self._some_method, params)
```

I guess we could save some lines of code and at the same time make it easier to follow if we introduced one method `_maybe_run_in_executor(urgent: bool, method: Callable[Whatever], *args)` which then decides how to perform the action based on the `urgent` param.

So, then all these individual places that now have the branching logic could get rid of the branching logic and become.

```python
need_nodes = await self._maybe_run_in_executor(urgent, self._some_method, params)
```

trinity/sync/beam/state.py
```diff
@@ -636,31 +691,67 @@ def _store_nodes(
...
+            self._preview_events[wait_event] = node_hashes
         while remaining_hashes and time.monotonic() - start_time < timeout:
-            await self._new_data_event.wait()
-            self._new_data_event.clear()
+            if urgent:
+                await self._new_data_event.wait()
+                self._new_data_event.clear()
+            else:
+                try:
+                    await asyncio.wait_for(wait_event.wait(), timeout=20)
```
cburgdorf, line 714: The undocumented timeout of `20` stands out. Maybe worth putting it into the constants.

cburgdorf (approved):
I left a comment about how I think we could trim down some of the branching logic but overall looks good to me.

trinity/sync/beam/state.py
```diff
@@ -636,31 +691,67 @@ def _store_nodes(
...
         start_time = time.monotonic()
+        if not urgent:
+            wait_event = asyncio.Event()
+            self._preview_events[wait_event] = node_hashes
         while remaining_hashes and time.monotonic() - start_time < timeout:
-            await self._new_data_event.wait()
-            self._new_data_event.clear()
+            if urgent:
+                await self._new_data_event.wait()
+                self._new_data_event.clear()
```
pipermerriam, line 711: Didn't dive very deep, but this might be better suited as an `asyncio.Condition`

pipermerriam (approved):

trinity/sync/beam/state.py
carver: > Am I missing something?

Hm, nope. I think the reason is historical: the method used to return node hashes.

> it could be reworked to have an early exit condition

Probably even better to simply skip the `run_in_executor` call altogether. I'll add that check to the call point.

trinity/sync/beam/state.py
carver: I'm missing what `asyncio.Condition` would add here. I don't see a shared resource that we need to prevent access to (that isn't handled by the event loop).

pipermerriam: It was the pattern of waiting then clearing that caught my eye.  I think as it exists it will work correctly, but there are cases where this pattern can fail and an update and thus end up having to wait longer than it should.

The `asyncio.Condition` API helps prevent this with the lock-then-wait pattern it enforces, which guarantees that you don't accidentally miss an update.

(This is fine to be left as-is if you disagree or simply don't feel like changing it.  I believe as-is it won't exhibit the "bad" behavior)

PR #2013, opened 2020-09-03
A pile of improvements to the DappNode package

cburgdorf:
### What was wrong?

Some issues with the DappNode package:

1. Mapping Trinity's source directory as a volume was preventing us from having clean updates.

2. We need to make the HTTP server accessible from outside ([they all do and it is safe due to the way how DappNode is shielded from the outside world](https://github.com/dappnode/DAppNodePackage-goerli-geth/blob/55f656c45e265f9e40b28b47fae65891e06079ad/build/Dockerfile#L10))

3. Ensure the DappNode package uses the timezone of the host machine

4. Enhance metadata of package (description, github issue tracker etc.)

5. Removed unnecessary file

### How was it fixed?

Mostly just changes to `dappnode_package.json` and the guide.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/dappnode/dappnode_package.json
+++ b/dappnode/dappnode_package.json
@@ -1,7 +1,8 @@
 {
   "name": "trinity.public.dappnode.eth",
   "version": "0.1.3",
-  "description": "Ethereum Python Client (experimental)",
+  "shortDescription": "Ethereum Python Client (experimental)",
+  "description": "Written in Python with great extensibility, the Trinity client is especially interesting for developers. Its novel beam sync strategy makes it possible to have a fully functional client within minutes after launch.",
   "avatar": "/ipfs/QmRyJJwuCEYHPf7xkRcoPJqiHQgAN3ay2voSCTrAY1nMet",
   "type": "library",
   "chain": "ethereum",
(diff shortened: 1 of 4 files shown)


dappnode/trinity.dnp.dappnode.eth.env
```diff
@@ -1 +0,0 @@
-EXTRA_OPTS--trinity-root-dir /trinity
```
cburgdorf, line 1: Note to self: Need to double check by installing a **fresh** version of the package if it is really safe to delete that file. (Can't currently do that as it's in use)

dappnode/dappnode_package.json
```diff
@@ -1,7 +1,8 @@
 {
   "name": "trinity.public.dappnode.eth",
   "version": "0.1.3",
-  "description": "Ethereum Python Client (experimental)",
+  "shortDescription": "Ethereum Python Client (experimental)",
+  "description": "Written in Python and having easy extensibility interfaces, the Trinity client is especially interesting for developers. It's novel beam sync strategy makes it possible to have a fully functional client within minutes after launch.",
```
njgheorghita, line 5: Just a suggestion `Written in Python with easy extensibility ...` - and also `It's` -> `Its`

njgheorghita (approved):
Looks good!

PR #2014, opened 2020-09-03
Beam preview prioritize data by block number

carver:
### What was wrong?

All data is treated equally when downloading preview data.

### How was it fixed?

We track the related block number for the data, then prefer earlier block numbers when querying peers.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


carver:
Oops, realized gsalgado is out. Switching to Piper for review so I can try to get it merged today.

trinity/sync/beam/state.py
```diff
@@ -132,20 +136,24 @@ def __init__(
         self._peer_pool = peer_pool
 
         # Track node data for upcoming blocks
+        self._block_number_lookup = defaultdict(lambda: BlockNumber(0))
         self._maybe_useful_nodes = TaskQueue[Hash32](
             buffer_size,
-            # Everything is the same priority, for now
-            lambda node_hash: 0,
+            # Prefer trie nodes from earliest blocks
+            lambda node_hash: self._block_number_lookup[node_hash],
```
pipermerriam, line 143: Just checking that the priority number is implemented as "small number > high priority" / "large number > low priority".

trinity/sync/beam/state.py
```diff
@@ -165,10 +173,24 @@ async def ensure_nodes_present(
                 urgent,
             )
         else:
+            for node_hash in node_hashes:
+                self._block_number_lookup[node_hash] = block_number
```
pipermerriam, line 177: Should this do something like `self._block_number_lookup[node_hash] = min(block_number, self._block_number_lookup.get(node_hash, block_number))`?  The idea is that a hash might be present in multiple blocks and we'd want to use the *lowest* block number that it was seen in.  Alternatively a simple approach would be to do `self._block_number_lookup.setdefault(node_hash, block_number)` which would at least not overwrite it *later* but wouldn't guarantee that it got the highest priority it deserves if it happens to be seen in a later block before it is seen in an earlier block.

pipermerriam (approved):

trinity/sync/beam/state.py
carver: Yup! Lowest first, just like a [`PriorityQueue`](https://docs.python.org/3/library/asyncio-queue.html#asyncio.PriorityQueue) (which it is, under the hood).

carver:
> Cannot write to _.com: [Errno -2] Name or service not known

No way this is related. Not sure how that ever worked, actually.

cburgdorf:
@carver That test is supposed to make sure Trinity doesn't crash if the metrics endpoint is unreachable (hence the weird endpoint). So, if that did crash the node, that's unexpected...

carver:
> @carver That test is supposed to make sure Trinity doesn't crash if the metrics endpoint is unreachable (hence the weird endpoint). So, if that did crash the node, that's unexpected...

Hm, I don't think trinity crashed though. It looks like the component exited (`trio.Cancelled`) and that wasn't handled smoothly? I'm unfamiliar with the component and trio. Can you take a look at the log? https://app.circleci.com/pipelines/github/ethereum/trinity/7381/workflows/dbd59a4c-8439-4481-b62a-943209c3afe9/jobs/281501

cburgdorf:
I haven't had any luck to reproduce the issue but I looked into it and this is the best explanation and fix that I was able to come up with https://github.com/ethereum/trinity/pull/2028

PR #2015, opened 2020-09-03
use upnp-port-forward library

pipermerriam:
### What was wrong?

Recently created the `upnp-port-forward` library to encapsulate common UPNP logic.  Should be using this rather than the older local version of this code.

### How was it fixed?

Removed local implementation.  Added `upnp-port-forward`

Also sorted dependencies 

#### Cute Animal Picture

animals-stuck-16


setup.py
```diff
@@ -21,43 +21,41 @@
...
-        "ruamel.yaml==0.16.10",
-        "argcomplete>=1.10.0,<2",
+        "lahja>=0.16.0,<0.17",
+        "libp2p==0.1.5",
         "multiaddr>=0.0.8,<0.1.0",
+        "mypy-extensions>=0.4.3,<0.5.0",
+        "plyvel==1.2.0",
         "prometheus-client==0.7.1",
+        "psutil>=5.7.0, <6",
+        "ruamel.yaml==0.16.10",
```
carver, line 49: Looks like a sorting bug. Expected it under `requests`

trinity/components/builtin/upnp/nat.py
```diff
@@ -61,7 +45,7 @@ async def run(self) -> None:
                             self.port,
                             UPNP_PORTMAP_DURATION,
                         )
-                        event = UPnPMapping(external_ip)
+                        event = UPnPMapping(str(external_ip))
```
carver, line 48: Why this change?

carver (approved):

trinity/components/builtin/upnp/nat.py
pipermerriam: The `upnp-port-forward` library returns `Union[ipaddress.IPv4Address, ipaddress.IPv6Address]` but the event uses string representations.  This is just downcasting to the string representation.

PR #2020, opened 2020-09-04
Reduce log level of event loop canary to DEBUG

cburgdorf:
### What was wrong?

This is what watching my Trinity node looks like

image

The event loop being overloaded hints at a real problem that we will have to continue to tackle but I don't think that these logs should be that prominent for the end user.

### How was it fixed?

Reduce log level to DEBUG.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


gsalgado:
Maybe change the current one to DEBUG and have a higher threshold (e.g. 1s) for a warning?

gsalgado (approved):

PR #2022, opened 2020-09-04
Add metrics for blockchain db health

### LINKED ISSUE #1938: Monitor the backfill progress on the grafana dashboard
### What is wrong?

We do currently only track the current head.

image

But this can give a wrong impression because beam sync is supposed to track the head of the chain even long before it has caught up with all historical headers and blocks.

### How can it be fixed

Our `ChainDB` exposes two APIs to find out about the current integrity of the chain, namely:

- `get_header_chain_gaps()`
- `get_chain_gaps()`

Both return `ChainGaps` which consist of sequences of block ranges (`(1, 20), (45, 789), ...` as well as a marker of the first missing header or block at the tip of the chain (e.g. `10_000_457`).

It would be super neat to put this into a nice visualization.

njgheorghita:
### What was wrong?
Fixes #1938 

### How was it fixed?
Calculated missing blocks/headers as a percentage and report these metrics to influx.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/trinity/_utils/connect.py
+++ b/trinity/_utils/connect.py
@@ -2,6 +2,7 @@ from typing import Iterator
 import contextlib
 
 from eth.abc import ChainAPI
+from eth.db.chain import ChainDB
 
 from lahja import EndpointAPI
 from trinity.boot_info import BootInfo
@@ -37,3 +38,11 @@ def get_eth1_chain_with_remote_db(boot_info: BootInfo,
             chain = chain_config.full_chain_class(base_db)
 
         yield chain
+
+
+@contextlib.contextmanager
+def get_eth1_chain_db_with_remote_db(boot_info: BootInfo,
+                                     event_bus: EndpointAPI) -> Iterator[ChainDB]:
+    base_db = DBClient.connect(boot_info.trinity_config.database_ipc_path)
+    with base_db:
+        yield ChainDB(base_db)
--- a/trinity/components/builtin/metrics/blockchain_metrics_collector.py
+++ b/trinity/components/builtin/metrics/blockchain_metrics_collector.py
@@ -6,25 +6,63 @@ from async_service import (
     as_service,
     ManagerAPI,
 )
+from eth.typing import ChainGaps
+from eth_typing import BlockNumber
 from p2p import trio_utils
 
 from lahja import EndpointAPI
 from trinity.boot_info import BootInfo
 from trinity.components.builtin.metrics.registry import HostMetricsRegistry
-from trinity._utils.connect import get_eth1_chain_with_remote_db
+from trinity._utils.connect import (
+    get_eth1_chain_with_remote_db,
+    get_eth1_chain_db_with_remote_db,
+)
 
 
 class BlockchainStats(NamedTuple):
     latest_block: int
+    header_db_completeness: float
+    block_db_completeness: float
 
 
 def read_blockchain_stats(boot_info: BootInfo, event_bus: EndpointAPI) -> BlockchainStats:
-    with get_eth1_chain_with_remote_db(boot_info, event_bus) as chain:
-        latest_block_number = chain.get_canonical_head().block_number
-
+    latest_block_number = read_latest_block_number(boot_info, event_bus)
+    header_db_gaps = read_header_db_gaps(boot_info, event_bus)
+    block_db_gaps = read_block_db_gaps(boot_info, event_bus)
     return BlockchainStats(
         latest_block=latest_block_number,
+        header_db_completeness=calculate_completeness(header_db_gaps, latest_block_number),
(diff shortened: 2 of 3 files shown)


njgheorghita:
Screen Shot 2020-09-04 at 12 42 38 PM
Screen Shot 2020-09-04 at 12 28 50 PM
@cburgdorf Curious what your thoughts are on this approach for visualizing db health. From the available options in grafana, it doesn't seem like there are any better options for visualizing db gaps except for 1. building a custom plugin (probably not worth the time effort) or 2. displaying the gaps via a table. I don't have a good sense for how important it is to be able to identify where the gaps actually are rather than a general overview of the db health.

cburgdorf:
I think tracking gaps as a percentage is fine if there are no better built-in visualizations. I don't think it is that important to see where exactly the gaps are.

But judging from the screenshots, the block visualization looks odd. I guess it is saying that we don't have any *gaps* in the chain of blocks because if we have e.g. blocks 1 - 3 (so we are not missing block 2 which would create a gap) but I think in terms of completeness we should calculate the percentage relative to the tip of the chain. In other words, we if the tip is `10795698` and we have blocks 1 - 3 then we should't say we have 100 % of the blocks when it is actually close to 0 %.

njgheorghita:
@cburgdorf Thanks, sounds good! I've updated the metrics to compare db health relative to the overall chain tip rather than the db tip.
Screen Shot 2020-09-08 at 9 25 56 AM

trinity/_utils/connect.py
```diff
@@ -37,3 +39,19 @@ def get_eth1_chain_with_remote_db(boot_info: BootInfo,
...
+    with base_db:
+        yield ChainDB(base_db)
+
+
+@contextlib.contextmanager
+def get_eth1_header_db_with_remote_db(boot_info: BootInfo,
+                                      event_bus: EndpointAPI) -> Iterator[HeaderDB]:
+    base_db = DBClient.connect(boot_info.trinity_config.database_ipc_path)
+    with base_db:
+        yield HeaderDB(base_db)
```
cburgdorf, line 57: I'm not opposed on having two distinct APIs but let me point out that `ChainDB` inherits from `HeaderDB` so everything that needs a `HeaderDB` can also be done on a `ChainDB`.

trinity/components/builtin/metrics/blockchain_metrics_collector.py
```diff
@@ -6,35 +6,78 @@
...
+
 @as_service
 async def collect_blockchain_metrics(manager: ManagerAPI,
                                      boot_info: BootInfo,
                                      event_bus: EndpointAPI,
                                      registry: HostMetricsRegistry,
                                      frequency_seconds: int) -> None:
-    blockchain_block_gauge = registry.gauge('trinity.blockchain/head/block.gauge')
+    blockchain_head_gauge = registry.gauge('trinity.blockchain/head/block.gauge')
+    header_db_health_gauge = registry.gauge('trinity.blockchain/health/header_db.gauge')
```
cburgdorf, line 76: Wondering if `health` is the best to convey what this represents. Since we are calculating the percentage of blocks and headers that we have synced so far, maybe something like `completeness` would be more clear?

trinity/components/builtin/metrics/blockchain_metrics_collector.py
```diff
@@ -6,35 +6,78 @@
...
+    with get_eth1_header_db_with_remote_db(boot_info, event_bus) as headerdb:
+        return headerdb.get_header_chain_gaps()
+
+
+def read_block_db_gaps(boot_info: BootInfo, event_bus: EndpointAPI) -> ChainGaps:
+    with get_eth1_chain_db_with_remote_db(boot_info, event_bus) as chaindb:
+        return chaindb.get_chain_gaps()
+
+
+def calculate_gap_health(gaps: ChainGaps, tip: BlockNumber) -> float:
```
cburgdorf, line 55: Same here, maybe one of these would be a bit clearer than the generic "health":

- `calculate_completeness` 
- `calculate_percent_completeness`
- `calculate_completeness_percent`
- `calculate_completeness_in_percent`
- `calculate_integrity` 
- `calculate_percent_integrity`
- `calculate_integrity_percent`
- `calculate_integrity_in_percent`

cburgdorf (approved):
Looks good! Just some naming recommendations.

cburgdorf:
I assume the new widgets aren't yet deployed on the dashboard, right? (I'm not seeing them)

njgheorghita:
Correct, I was thinking of getting a couple more metrics completed before pushing a new docker image to k8s but i'm happy to do that today if you're curious? Lmk if you'd like to see them in action, it shouldn't take too long

cburgdorf:
No rush, just wanted to reassure myself I'm not missing anything.

PR #2025, opened 2020-09-07
Remove p2p.BaseService & co

gsalgado:
Also remove cancel_token from our dependencies


gsalgado:
@carver I couldn't change https://github.com/ethereum/trinity/blob/d95a7775c98f048a6770ddbbaa9ca4358e3abcc5/tests/core/task-queue-utils/test_task_queue.py#L69 to not use a `CancelToken`, and since it's not a trivial test I thought I'd check with you first before I rewrite it and we accidentally end up with a test that doesn't test what it should. Do you think we can easily change it so that we can drop `cancel_token` from our deps or should I just leave it for now?

pipermerriam (approved):
:heart:

carver:
> @carver I couldn't change
> 
> https://github.com/ethereum/trinity/blob/d95a7775c98f048a6770ddbbaa9ca4358e3abcc5/tests/core/task-queue-utils/test_task_queue.py#L69
> 
> to not use a `CancelToken`, and since it's not a trivial test I thought I'd check with you first before I rewrite it and we accidentally end up with a test that doesn't test what it should. Do you think we can easily change it so that we can drop `cancel_token` from our deps or should I just leave it for now?

Yeah, it may take a bit of work. Chaining isn't used here, so it wouldn't be that hard to add a small internal version of the cancel token in utils that has a limited, similar API. No objection to removing.

gsalgado:
I don't think it's worth investing the time to implement a simpler `CancelToken` in trinity just for this test, so I'd rather keep the dependency and leave the test unchanged for now

Issue #2018, opened 2020-09-04
Peer Pool crashes, dies but does not crash the whole node


cburgdorf:
### What is wrong?

It appears that my peer pool crashed around 4:19 and from there did not recover again (no further logs from `ETHPeerPool` after the crash).

```
[1m[33m WARNING  2020-09-04 04:19:18,179       Sync / PeerPool  Event loop blocked or overloaded: delay=2.005s, tasks=74[0m
[1m[31m   ERROR  2020-09-04 04:19:18,181               asyncio  Exception in callback <TaskWakeupMethWrapper object at 0x7fa47b456e90>
handle: <Handle <TaskWakeupMethWrapper object at 0x7fa47b456e90>>
Traceback (most recent call last):
  File "uvloop/cbhandles.pyx", line 70, in uvloop.loop.Handle._run
  File "/usr/local/lib/python3.7/site-packages/async_service/base.py", line 300, in _run_and_manage_task
    await task.run()
  File "/usr/local/lib/python3.7/site-packages/async_service/base.py", line 165, in run
    await self.child_manager.run()
  File "/usr/local/lib/python3.7/site-packages/async_service/asyncio.py", line 246, in run
    for _, exc_value, exc_tb in self._errors
trio.MultiError: TimeoutError('Tasks never returned after being cancelled: %s', {<Task cancelled coro=<wait_first() done, defined at /usr/src/app/trinity/p2p/asyncio_utils.py:31>>, <Task cancelled coro=<wait_first() done, defined at /usr/src/app/trinity/p2p/asyncio_utils.py:31>>, <Task cancelled coro=<wait_first() done, defined at /usr/src/app/trinity/p2p/asyncio_utils.py:31>>, <Task cancelled coro=<wait_first() done, defined at /usr/src/app/trinity/p2p/asyncio_utils.py:31>>}), TimeoutError('Tasks never returned after being cancelled: %s', {<Task cancelled coro=<wait_first() done, defined at /usr/src/app/trinity/p2p/asyncio_utils.py:31>>, <Task cancelled coro=<wait_first() done, defined at /usr/src/app/trinity/p2p/asyncio_utils.py:31>>, <Task cancelled coro=<wait_first() done, defined at /usr/src/app/trinity/p2p/asyncio_utils.py:31>>, <Task cancelled coro=<wait_first() done, defined at /usr/src/app/trinity/p2p/asyncio_utils.py:31>>})

Details of embedded exception 1:

  Traceback (most recent call last):
[...]

PR #2026, opened 2020-09-08
Do not let timeout errors escape from wait_first

cburgdorf:
### What was wrong?

None of the places that used `wait_first` caught the possible `asyncio.TimeoutError` that is being raised when the cancelled task isn't stopping in the appropriate timeframe.

### How was it fixed?

Citing @gsalgado from #2018 

>We should probably catch TimeoutError and log it as a warning/error wherever we call wait_first(). These could be caused by a bug when terminating any long-running-task/service but may also happen when the event loop is overloaded, so we probably don't want to let them crash a component.

Caught `asyncio.TimeoutError` and logged it as a warning.

### To-Do

- [x] Clean up commit history

- [ ] ~~Add entry to the [release notes](https://github.com/ethereum/trinity/blob/master/newsfragments/README.md)~~ (I think the bug is only in master)

#### Cute Animal Picture

### DIFF
--- a/p2p/connection.py
+++ b/p2p/connection.py
@@ -131,10 +131,17 @@ class Connection(ConnectionAPI, Service):
                     for behavior in behaviors:
                         behavior.post_apply()
                     await wait_first(futures, max_wait_after_cancellation=2)
+                except asyncio.TimeoutError:
+                    self.logger.warning(
+                        "Timed out waiting for tasks to terminate after cancellation: %s",
+                        futures
+                    )
                 except PeerConnectionLost:
                     # Any of our behaviors may propagate a PeerConnectionLost, which is to be
                     # expected as many Connection APIs used by them can raise that. To avoid a
                     # DaemonTaskExit since we're returning silently, ensure we're cancelled.
+                    pass
+                finally:
                     self.manager.cancel()
 
     async def run_peer(self, peer: 'BasePeer') -> None:
--- a/trinity/components/builtin/syncer/component.py
+++ b/trinity/components/builtin/syncer/component.py
@@ -375,7 +375,14 @@ class SyncerComponent(AsyncioIsolatedComponent):
             # returns.
             node_manager_task = create_task(
                 node_manager.wait_finished(), f'{NodeClass.__name__} wait_finished() task')
-            await wait_first([sync_task, node_manager_task], max_wait_after_cancellation=2)
+            tasks = [sync_task, node_manager_task]
+            try:
+                await wait_first(tasks, max_wait_after_cancellation=2)
+            except asyncio.TimeoutError:
+                self.logger.warning(
+                    "Timed out waiting for tasks to terminate after cancellation: %s",
+                    tasks
+                )
 
     async def launch_sync(self,
                           node: Node[BasePeer],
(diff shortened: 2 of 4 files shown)


p2p/connection.py
```diff
@@ -131,10 +131,14 @@ async def run_behaviors(self, behaviors: Tuple[BehaviorAPI, ...]) -> None:
...
                         behavior.post_apply()
                     await wait_first(futures, max_wait_after_cancellation=2)
+                except asyncio.TimeoutError:
+                    self.logger.warning("Timed out during async task cancellation")
                 except PeerConnectionLost:
                     # Any of our behaviors may propagate a PeerConnectionLost, which is to be
                     # expected as many Connection APIs used by them can raise that. To avoid a
                     # DaemonTaskExit since we're returning silently, ensure we're cancelled.
+                    pass
+                finally:
```
cburgdorf, line 141: Seemed appropriate to move the cancellation into a `finally`

p2p/connection.py
```diff
@@ -131,10 +131,14 @@ async def run_behaviors(self, behaviors: Tuple[BehaviorAPI, ...]) -> None:
                     for behavior in behaviors:
                         behavior.post_apply()
                     await wait_first(futures, max_wait_after_cancellation=2)
+                except asyncio.TimeoutError:
+                    self.logger.warning("Timed out during async task cancellation")
```
gsalgado, line 135: We need a more descriptive warning so that we can track down the rogue tasks. Something like `.warning("Timed out waiting for tasks to terminate after cancellation: %s", futures)`

trinity/components/builtin/syncer/component.py
```diff
@@ -375,7 +375,10 @@ async def do_run(self, event_bus: EndpointAPI) -> None:
             # returns.
             node_manager_task = create_task(
                 node_manager.wait_finished(), f'{NodeClass.__name__} wait_finished() task')
-            await wait_first([sync_task, node_manager_task], max_wait_after_cancellation=2)
+            try:
+                await wait_first([sync_task, node_manager_task], max_wait_after_cancellation=2)
+            except asyncio.TimeoutError:
+                self.logger.warning("Timed out during async task cancellation")
```
gsalgado, line 381: Same here

trinity/extensibility/asyncio.py
```diff
@@ -58,20 +58,26 @@ async def _do_run(self) -> None:
...
-                            [do_run_task, eventbus_task, loop_monitoring_task],
-                            max_wait_after_cancellation,
-                        )
+                        try:
+                            await wait_first(
+                                [do_run_task, eventbus_task, loop_monitoring_task],
+                                max_wait_after_cancellation,
+                            )
+                        except asyncio.TimeoutError:
+                            self.logger.warning("Timed out during async task cancellation")
```
gsalgado, line 80: The two above as well

trinity/extensibility/component.py
```diff
@@ -262,7 +262,10 @@ async def run() -> None:
         async with _run_eventbus_for_component(component, connect_to_endpoints) as event_bus:
             async with _run_asyncio_component_in_proc(component, event_bus) as component_task:
                 sigint_task = asyncio.create_task(got_sigint.wait())
-                await wait_first([component_task, sigint_task], max_wait_after_cancellation=2)
+                try:
+                    await wait_first([component_task, sigint_task], max_wait_after_cancellation=2)
+                except asyncio.TimeoutError:
+                    logger.warning("Timed out during async task cancellation")
```
gsalgado, line 268: Here too

gsalgado (approved):
Thanks for taking care of this; I was going to do it myself but got sidetracked on #2025

PR #2031, opened 2020-09-09
Discovery V5

pipermerriam:
### What was wrong?

We want to start moving towards supporting and using the discovery v5 network.

### How was it fixed?

Minimal implementation of the ability to run v5 discovery alongside v4.  (not v5.1)

#### Cute Animal Picture

animals-stuck-leop_3458605k


pipermerriam:
Got this working in a very minimal way locally.  Some upstream changes to DDHT need to be merged prior to this being useful.

We also very likely want a better ENRDB that lets us filter by capabilities.

We also need to integrate the extra `eth` field into our discv5 ENR record.

Issue #2008, opened 2020-09-02
ConnectionBusy exception thrown from BeamDownloader


cburgdorf:
### What is wrong?

I've noticed the following exception.

```
f2b7-40b4-a77a-6ba6d13da819>: Timed out waiting for NodeDataV65 request lock or connection: Connection-<Session <Node(0xd4062d@91.13.194.238)> 1c231c61-f2b7-40b4-a77a-6ba6d13da819>[0m
   DEBUG  2020-09-02 12:59:58,686        BeamDownloader  Problem downloading nodes from peer, dropping...
Traceback (most recent call last):
  File "/usr/src/app/trinity/p2p/exchange/candidate_stream.py", line 81, in payload_candidates
    timeout=total_timeout * NUM_QUEUED_REQUESTS,
  File "/usr/local/lib/python3.7/asyncio/tasks.py", line 449, in wait_for
    raise futures.TimeoutError()
concurrent.futures._base.TimeoutError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/src/app/trinity/trinity/sync/beam/state.py", line 681, in _request_nodes
    node_hashes, timeout=self._reply_timeout)
  File "/usr/src/app/trinity/trinity/protocol/eth/exchanges.py", line 110, in __call__
    timeout,
  File "/usr/src/app/trinity/p2p/exchange/exchange.py", line 73, in get_result
    timeout,
  File "/usr/src/app/trinity/p2p/exchange/manager.py", line 59, in get_result
    async for payload in stream.payload_candidates(request, tracker, timeout=timeout):
  File "/usr/src/app/trinity/p2p/exchange/candidate_stream.py", line 85, in payload_candidates
    f"Timed out waiting for {self.response_cmd_name} request lock "
p2p.exceptions.ConnectionBusy: Timed out waiting for NodeDataV65 request lock or connection: Connection-<Session <Node(0xd4062d@91.13.194.238)> 1c231c61-f2b7-40b4-a77a-6ba6d13da819>
```

It seems that after that error the client was left in a broken state and the logs where dominated by this warning.

```
[...]

PR #2039, opened 2020-09-10
Send fewer non-urgent GetNodeData queries to peers

carver:
### What was wrong?

#2008 

### How was it fixed?

In general, it's not common for a query to already be active, but just in case, we'd rather not queue up more requests for preview or backfill data. (Urgent data, on the other hand, should always be retrieved as quickly as possible, so we allow urgent requests to stack up)

I'm not really sure that this will fix 2008, but it won't hurt, and it might help.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/backfill.py
+++ b/trinity/sync/beam/backfill.py
@@ -188,6 +188,17 @@ class BeamStateBackfill(Service, QueenTrackerAPI):
 
             peer = await self._queening_queue.pop_fastest_peasant()
 
+            # skip over peer if it has an active data request
+            while peer.eth_api.get_node_data.is_requesting:
+                self.logger.debug(
+                    "Want backfill nodes from %s, but it has an active request, skipping...",
+                    peer,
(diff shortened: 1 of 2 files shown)


cburgdorf:
:+1:  I'm trying latest master + this PR on my DappNode overnight.

carver:
@cburgdorf I fixed the typo. Looks good to run now.

cburgdorf:
The good news is that I haven't gotten any `ConnectionBusy` error with this patch. The bad news is that my node eventually starved and I found it idling without peers this morning. But looking through the logs I can't make a connection between these issues. I will create a separate issue for that to examine.

cburgdorf:
Created https://github.com/ethereum/trinity/issues/2045 for the failed run.

cburgdorf (approved):
I can confirm that this fixes the `ConnectionBusy` issue :+1:

PR #2040, opened 2020-09-11
Report beam sync pivot rate and events via metrics

njgheorghita:
### What was wrong?
Track the number of pivots/hour

### How was it fixed?
Create an event that gets triggered every time a pivot happens. Subscribe to this event and track the rate of pivots with the builtin pyformance weighted moving average, and also send an "annotation" to influxdb to display to event as a red line on the graphs.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/trinity/components/builtin/metrics/abc.py
+++ b/trinity/components/builtin/metrics/abc.py
@@ -22,3 +22,7 @@ class MetricsServiceAPI(ServiceAPI):
     @abstractmethod
     def registry(self) -> MetricsRegistry:
         ...
+
+    @abstractmethod
+    async def send_annotation(self, annotation_data: str) -> None:
+        ...
--- a/trinity/components/builtin/metrics/service/base.py
+++ b/trinity/components/builtin/metrics/service/base.py
@@ -1,13 +1,40 @@
 import time
+import aiohttp
+from typing import Dict
+import base64
 from abc import abstractmethod
+from urllib import parse
 from http.client import HTTPException
 
 from async_service import Service
-from pyformance.reporters import InfluxReporter
+from pyformance.reporters.influx import InfluxReporter
 
 from trinity.components.builtin.metrics.abc import MetricsServiceAPI
 from trinity.components.builtin.metrics.registry import HostMetricsRegistry
 from trinity._utils.logging import get_logger
+
+
+class ExtendedInfluxReporter(InfluxReporter):
+    """
+    ``InfluxReporter`` extended to enable sending annotations to InfluxDB
+    """
+    async def send_annotation(self, annotation_data: str) -> None:
+        url = parse.urlunparse((
+            self.protocol,  # scheme
+            f"{self.server}:{self.port}",  # netloc
+            "/write",  # path
+            '',
+            f"db={self.database}&precision=s",  # query
+            ''
+        ))
+        auth_header = self._generate_auth_header()
+        async with aiohttp.ClientSession() as session:
+            await session.post(url, data=annotation_data, headers=auth_header)
+
+    def _generate_auth_header(self) -> Dict[str, str]:
+        auth_string = ("%s:%s" % (self.username, self.password)).encode()
+        auth = base64.b64encode(auth_string)
+        return {"Authorization": "Basic %s" % auth.decode("utf-8")}
 
 
 class BaseMetricsService(Service, MetricsServiceAPI):
@@ -32,14 +59,14 @@ class BaseMetricsService(Service, MetricsServiceAPI):
         self._influx_server = influx_server
         self._reporting_frequency = reporting_frequency
         self._registry = HostMetricsRegistry(host)
-        self._reporter = InfluxReporter(
+        self._reporter = ExtendedInfluxReporter(
             registry=self._registry,
             database=influx_database,
             username=influx_user,
             password=influx_password,
             protocol=protocol,
             port=port,
-            server=influx_server
+            server=influx_server,
         )
 
     logger = get_logger('trinity.components.builtin.metrics.MetricsService')
@@ -51,6 +78,9 @@ class BaseMetricsService(Service, MetricsServiceAPI):
         metrics instruments can be registered and retrieved.
         """
         return self._registry
+
+    async def send_annotation(self, annotation_data: str) -> None:
+        await self._reporter.send_annotation(annotation_data)
 
     async def run(self) -> None:
         self.logger.info("Reporting metrics to %s", self._influx_server)
--- a/trinity/components/builtin/metrics/service/noop.py
+++ b/trinity/components/builtin/metrics/service/noop.py
@@ -39,5 +39,8 @@ class NoopMetricsService(Service, MetricsServiceAPI):
     async def continuously_report(self) -> None:
         pass
 
+    async def send_annotation(self, annotation_data: str) -> None:
+        pass
+
 
 NOOP_METRICS_SERVICE = NoopMetricsService()
--- a/trinity/components/builtin/metrics/sync_metrics_registry.py
+++ b/trinity/components/builtin/metrics/sync_metrics_registry.py
@@ -1,0 +1,24 @@
+import time
+
+from eth_typing import BlockNumber
+
+from trinity.components.builtin.metrics.abc import MetricsServiceAPI
+
+
+class SyncMetricsRegistry:
+    """
+    Registry to track weighted moving average of pivot events, and report to InfluxDB.
+    """
+    def __init__(self, metrics_service: MetricsServiceAPI) -> None:
+        self.metrics_service = metrics_service
+        self.pivot_meter = metrics_service.registry.meter('trinity.p2p/sync/pivot_rate.meter')
+
+    async def record_pivot(self, block_number: BlockNumber) -> None:
+        # record pivot and send event annotation to influxdb
+        self.pivot_meter.mark()
+        pivot_time = int(time.time())
+        post_data = (
+            f'events title="beam pivot @ block {block_number}",'
+            f'text="pivot event",tags="{self.metrics_service.registry.host}" {pivot_time}'
+        )
+        await self.metrics_service.send_annotation(post_data)
--- a/trinity/components/builtin/syncer/component.py
+++ b/trinity/components/builtin/syncer/component.py
@@ -34,8 +34,10 @@ from trinity.boot_info import BootInfo
 from trinity.config import (
     Eth1AppConfig,
 )
+from trinity.components.builtin.metrics.abc import MetricsServiceAPI
 from trinity.components.builtin.metrics.component import metrics_service_from_args
 from trinity.components.builtin.metrics.service.asyncio import AsyncioMetricsService
+from trinity.components.builtin.metrics.sync_metrics_registry import SyncMetricsRegistry
 from trinity.components.builtin.metrics.service.noop import NOOP_METRICS_SERVICE
 from trinity.constants import (
     NETWORKING_EVENTBUS_ENDPOINT,
@@ -136,7 +138,8 @@ class BaseSyncStrategy(ABC):
                    chain: AsyncChainAPI,
                    base_db: AtomicDatabaseAPI,
                    peer_pool: BasePeerPool,
-                   event_bus: EndpointAPI) -> None:
+                   event_bus: EndpointAPI,
+                   metrics_service: MetricsServiceAPI = None) -> None:
         ...
 
 
@@ -152,7 +155,8 @@ class NoopSyncStrategy(BaseSyncStrategy):
                    chain: AsyncChainAPI,
                    base_db: AtomicDatabaseAPI,
                    peer_pool: BasePeerPool,
-                   event_bus: EndpointAPI) -> None:
+                   event_bus: EndpointAPI,
+                   metrics_service: MetricsServiceAPI = None) -> None:
 
         logger.info("Node running without sync (--sync-mode=%s)", self.get_sync_mode())
         await asyncio.Future()
@@ -170,7 +174,8 @@ class FullSyncStrategy(BaseSyncStrategy):
                    chain: AsyncChainAPI,
                    base_db: AtomicDatabaseAPI,
                    peer_pool: BasePeerPool,
-                   event_bus: EndpointAPI) -> None:
+                   event_bus: EndpointAPI,
+                   metrics_service: MetricsServiceAPI = None) -> None:
 
         syncer = FullChainSyncer(
             chain,
@@ -205,7 +210,11 @@ class BeamSyncStrategy(BaseSyncStrategy):
                    chain: AsyncChainAPI,
                    base_db: AtomicDatabaseAPI,
                    peer_pool: BasePeerPool,
-                   event_bus: EndpointAPI) -> None:
+                   event_bus: EndpointAPI,
+                   metrics_service: MetricsServiceAPI = None) -> None:
+
+        # create registry for tracking pivot metrics
+        sync_metrics_registry = SyncMetricsRegistry(metrics_service)
 
         syncer = BeamSyncService(
             chain,
@@ -215,7 +224,8 @@ class BeamSyncStrategy(BaseSyncStrategy):
             event_bus,
             args.sync_from_checkpoint,
             args.force_beam_block_number,
-            not args.disable_backfill
+            not args.disable_backfill,
+            sync_metrics_registry,
         )
 
         async with background_asyncio_service(syncer) as manager:
@@ -239,7 +249,8 @@ class HeaderSyncStrategy(BaseSyncStrategy):
                    chain: AsyncChainAPI,
                    base_db: AtomicDatabaseAPI,
                    peer_pool: BasePeerPool,
-                   event_bus: EndpointAPI) -> None:
+                   event_bus: EndpointAPI,
+                   metrics_service: MetricsServiceAPI = None) -> None:
 
         syncer = HeaderChainSyncer(
             chain,
@@ -265,7 +276,8 @@ class LightSyncStrategy(BaseSyncStrategy):
                    chain: AsyncChainAPI,
                    base_db: AtomicDatabaseAPI,
                    peer_pool: BasePeerPool,
-                   event_bus: EndpointAPI) -> None:
+                   event_bus: EndpointAPI,
+                   metrics_service: MetricsServiceAPI = None) -> None:
 
         syncer = LightChainSyncer(
             chain,
@@ -397,6 +409,7 @@ class SyncerComponent(AsyncioIsolatedComponent):
             node.base_db,
             node.get_peer_pool(),
             event_bus,
+            node.metrics_service,
         )
 
 
--- a/trinity/sync/beam/service.py
+++ b/trinity/sync/beam/service.py
@@ -8,6 +8,7 @@ from eth_typing import BlockNumber
 from eth.abc import AtomicDatabaseAPI
 
 from trinity.chains.base import AsyncChainAPI
+from trinity.components.builtin.metrics.sync_metrics_registry import SyncMetricsRegistry
 from trinity.db.eth1.chain import BaseAsyncChainDB
 from trinity.protocol.eth.peer import ETHPeerPool
 from trinity.sync.beam.constants import (
@@ -31,7 +32,8 @@ class BeamSyncService(Service):
             event_bus: EndpointAPI,
             checkpoint: Checkpoint = None,
             force_beam_block_number: BlockNumber = None,
-            enable_header_backfill: bool = False) -> None:
+            enable_header_backfill: bool = False,
+            sync_metrics_registry: SyncMetricsRegistry = None) -> None:
         self.logger = get_logger('trinity.sync.beam.service.BeamSyncService')
         self.chain = chain
         self.chaindb = chaindb
@@ -41,6 +43,7 @@ class BeamSyncService(Service):
         self.checkpoint = checkpoint
         self.force_beam_block_number = force_beam_block_number
         self.enable_header_backfill = enable_header_backfill
+        self.sync_metrics_registry = sync_metrics_registry
 
     async def run(self) -> None:
         head = await self.chaindb.coro_get_canonical_head()
@@ -72,6 +75,9 @@ class BeamSyncService(Service):
             do_pivot = await self._monitor_for_pivot(beam_syncer)
             if do_pivot:
                 self.logger.info("Pivoting Beam Sync to a newer header...")
+                if self.sync_metrics_registry:
+                    latest_block = beam_syncer._body_syncer._latest_block_number
+                    await self.sync_metrics_registry.record_pivot(latest_block)
             else:
                 self.logger.info("No pivot requested. Leaving Beam Syncer closed...")
                 break
--- a/newsfragments/2040.feature.rst
+++ b/newsfragments/2040.feature.rst
@@ -1,0 +1,1 @@
+Report beam sync pivot rate to metrics for visualization on grafana dashboard.
--- a/trinity/sync/full/chain.py
+++ b/trinity/sync/full/chain.py
@@ -160,10 +160,10 @@ class BaseBodyChainSyncer(Service, PeerSubscriber):
         # Keep track of some statistics, which is useful for deciding if syncing has stalled
 
         # What is the largest block number reported by a header? (before importing the block)
-        self._highest_header_number = 0
+        self._highest_header_number = BlockNumber(0)
 
         # What is the most recently imported block number? (after importing the block)
-        self._latest_block_number = 0
+        self._latest_block_number = BlockNumber(0)
 
     async def run(self) -> None:
         with self.subscribe(self._peer_pool):


njgheorghita:
@cburgdorf I still have some hunting down to do to get the tests passing, but I'll pick it up tomorrow. Until then, if you don't mind giving this a quick :eyes: and jot down any initial thoughts, that would be great!

Screen Shot 2020-09-10 at 2 46 33 PM

trinity/sync/beam/service.py
```diff
@@ -72,6 +73,8 @@ async def _pivot_loop(self) -> None:
             do_pivot = await self._monitor_for_pivot(beam_syncer)
             if do_pivot:
                 self.logger.info("Pivoting Beam Sync to a newer header...")
+                latest_block_number = beam_syncer._body_syncer._latest_block_number
+                await self.event_bus.broadcast(PivotEvent(latest_block_number), BroadcastConfig(internal=True))
```
njgheorghita, line 77: Is `latest_block_number` here useful info? Or is it more interesting to display the block number that we are pivoting from?

p2p/peer_pool.py
```diff
@@ -138,6 +140,11 @@ def __init__(self,
             metrics_registry = MetricsRegistry()
         self._active_peer_counter = metrics_registry.counter('trinity.p2p/peers.counter')
         self._peer_reporter_registry = self.get_peer_reporter_registry(metrics_registry)
+        self._sync_metrics_registry = SyncMetricsRegistry(
+            metrics_registry,
+            metrics_reporter,
+            self._event_bus
+        )
```
cburgdorf, line 147: I don't think the `PeerPool` is the right place to drive this. You can consider the peer pool and the sync as two fairly decoupled things. In fact, in some idealistic future both of them would run in independent processes but we aren't there yet. 

I actually think that we shouldn't need any changes to the `p2p/*` namespace for this new metric.

I think there are two way possible ways forward:

1. Pass the `metrics_service.registry` into the sync code and then let the sync code create whatever gauges, counters, meters etc it needs and directly set their values. We can also make this a bit more cleaner by wrapping the bare registry and creating a more type safe version only for the sync. Something like:

```python

class SyncMetricsRegistry:

    def __init__(self, registry: MetricsRegistry):
        self._registry = registry

   @property
   def pivot_gauge(self) -> SimpleGauge:
       return self.registry.gauge('trinity.p2p/sync/pivot_rate.gauge')
```
  

2. Alternatively, the sync code could stay unaware of the registry and just raise the (internal) event (as you do now) and then there is code listening for the event that writes it to the gauge. But that code shouldn't live inside the peer pool which is unaware of the sync.

p2p/metrics.py
```diff
@@ -60,3 +66,65 @@ def reset_peer_meters(self, peer_id: int) -> None:
...
+        pivots_over_last_hour = self.get_pivots_over_last_hour()
+        self.pivot_gauge.set_value(pivots_over_last_hour)
+
+    def record_pivot(self, block_number: int) -> None:
+        # record pivot and send event annotation to influxdb
+        current_time = int(time.time())
+        self.pivot_timestamps.append(current_time)
+        self._post_pivot_annotations(current_time, block_number)
+        pivots_over_last_hour = self.get_pivots_over_last_hour(current_time)
+        self.pivot_gauge.set_value(pivots_over_last_hour)
```
cburgdorf, line 110: I'm wondering if the `gauge` is the correct instrument here. Thinking about the different instruments always makes my head spin but a `Meter` already tracks the time and calculates the rate.

https://pyformance.readthedocs.io/en/latest/
>Measures the rate of events over time. Useful to track how often a certain portion of your application gets requests so you can set resources accordingly. Tracks the mean rate (the overall rate since the meter was reset) and the rate statistically significant regarding only events that have happened in the last 1, 5, and 15 minutes (Exponentally weighted moving average).

p2p/metrics.py
```diff
@@ -60,3 +66,65 @@ def reset_peer_meters(self, peer_id: int) -> None:
...
+        self.username = metrics_reporter.username
+        self.password = metrics_reporter.password
+        self.host = metrics_registry.host
+        self.event_bus = event_bus
+        self.event_bus.subscribe(PivotEvent, lambda event: self.record_pivot(event.payload))
+        self.pivot_gauge = metrics_registry.gauge('trinity.p2p/sync/pivot_rate.gauge')
+        self.pivot_gauge.set_value(0)
+        self.pivot_timestamps = []
+
+    def trigger_sync_reports(self) -> None:
```
cburgdorf, line 99: I would think that we do not even need to have our own continuous trigger to calculate the rate of pivots if we used a `meter`. Pivots would simply occur as they occur and the `meter` would calculate the rate for us I think. And the MetricsService has its own reporting interval that continuously pushes the values upstream.

cburgdorf (commented):
Left comments inline, I think some code ended up being at the wrong place. I also think the authentication code shouldn't be there and probably the pivots per hour calculation is off (and maybe doesn't even need to be there if a `Meter` would do the job for us)

Looking at this graphic...

Screen Shot 2020-09-10 at 2 46 33 PM

The pivots per hour seem to only be going up but that doesn't match with the red dottet lines that I'm counting. At 14:30 we have 5 pivots that happend roughly in a timespan of 5 minutes so that means we should have a rate of 60 pivots per hour (5 per 5 minutes = 5*12 = 60 per 1 hour). But then the timespan between pivots **increases** which means the rate of pivots per hour should be going **down**. At 14:35 we only have one pivot more than before so where we previously had 5 pivots per five minutes we now only have 6 pivots per 10 minutes which means our rate should have **decreased** to 36 pivots per hours (6 pivots per 10 minutes = 6 * 6 = 36 per 1 hour).

njgheorghita:
@cburgdorf Thanks for the feedback! So I guess `pivots / hour` is not the most accurate name, a more appropriate name would be `number of pivots over the last hour`. My sense here is that the rate might be a bit misleading, and the raw number of pivots over the past hour is more useful metrics. 

> The pivots per hour seem to only be going up but that doesn't match with the red dottet lines that I'm counting. At 14:30 we have 5 pivots that happend roughly in a timespan of 5 minutes so that means we should have a rate of 60 pivots per hour (5 per 5 minutes = 5*12 = 60 per 1 hour). But then the timespan between pivots increases which means the rate of pivots per hour should be going down. At 14:35 we only have one pivot more than before so where we previously had 5 pivots per five minutes we now only have 6 pivots per 10 minutes which means our rate should have decreased to 36 pivots per hours (6 pivots per 10 minutes = 6 * 6 = 36 per 1 hour).

I'm not sure I agree with how you calculate `pivots/hour` here. In one example you extrapolate a time range of 5 minutes to calculate the hourly rate - and in another example you use a time range of 10 minutes. The time ranges choose seem arbitrary which will lead to an inaccurate hourly rate. Shouldn't the hourly rate be simply the number of pivots over the past 60 minutes? Though, I could be misunderstanding your example. 

Though as you mentioned, I agree that a meter is a better option than a gauge. Though, I was confused by the [pyformance API docs](https://pyformance.readthedocs.io/en/latest/reference.html) - which from what I can see have just forgotten to document the actual [`Meter class`](https://github.com/omergertel/pyformance/blob/master/pyformance/meters/meter.py).

p2p/metrics.py
njgheorghita: Well, the hourly rate (according to how I calculated it as the # of pivots over the past hour) changes continuously (aka every second). This trigger mechanism is important to fire continuously b/c if we only report the rate when a pivot occurs, then let's say no more pivots occur, the hourly rate will be stuck on the last reported value rather than decreasing gradually to reflect the decrease in pivots.

p2p/metrics.py
```diff
@@ -60,3 +66,65 @@ def reset_peer_meters(self, peer_id: int) -> None:
...
+                 metrics_registry: MetricsRegistry,
+                 metrics_reporter: InfluxReporter,
+                 event_bus: EndpointAPI
+                 ) -> None:
+        self.protocol = metrics_reporter.protocol
+        self.server = metrics_reporter.server
+        self.port = metrics_reporter.port
+        self.database = metrics_reporter.database
+        self.username = metrics_reporter.username
+        self.password = metrics_reporter.password
```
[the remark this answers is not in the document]
njgheorghita: The pyformance `MetricsReporter` (located on `MetricsService`) cannot handle `annotations` for influx (aka events displayed as a red line on the graph), so I had to implement my own post request to influxdb to save these events, which requires authentication.

p2p/metrics.py
njgheorghita: Ahhh, but I misunderstood your point here that `IF` we use a meter this trigger is no longer necessary. 👍

p2p/metrics.py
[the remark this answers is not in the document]
cburgdorf: In that case I think we should derive from `InfluxReporter` and use the new derived reporter here.

https://github.com/ethereum/trinity/blob/2cfe64d2a87f2de316858d5d5825f50b61b143b8/trinity/components/builtin/metrics/service/base.py#L35-L43

But I think the this kind of downstream code should not be concerned with authentication again.

cburgdorf:
>My sense here is that the rate might be a bit misleading, and the raw number of pivots over the past hour is more useful metrics.

I think that the raw number of pivots that happened is fine to just roughly read from the overlayed dotted lines. But what I'm really interested in is to see if the rate at which pivots happen slows down over time.

>In one example you extrapolate a time range of 5 minutes to calculate the hourly rate - and in another example you use a time range of 10 minutes.

I think I just used [cross multiplication](https://en.wikipedia.org/wiki/Cross-multiplication) to arrive at the rate. We do agree that the rate was slowing down there, no? In the first five minutes pivots happened roughly every minute (so 60 per hour) and then after the there's a five minute window of not having pivots so clearly pivots become *less frequent* there.

I basically would like to see pivots per hour (or per last 1000 blocks could also work if we want to decouple from the notion of time) in the same way as we display packets per second (except that pivots happen at a much slower frequency)

image

p2p/metrics.py
```diff
@@ -60,3 +66,65 @@ def reset_peer_meters(self, peer_id: int) -> None:
...
+        self.host = metrics_registry.host
+        self.event_bus = event_bus
+        self.event_bus.subscribe(PivotEvent, lambda event: self.record_pivot(event.payload))
```
pipermerriam, line 94: This stood out to me as a potential antipattern.  `__init__` methods that *do stuff* have the potential to be problematic.  Is there any way that we can actually manage the lifecycle of this subscription so that it gets installed and removed?  Less idea but maybe better than this might be an extra method that activates the subscription (and maybe some safety instrumentation on the various reporting methods to disallow reporting if the subscription hasn't been activated).

p2p/metrics.py
```diff
@@ -60,3 +66,65 @@ def reset_peer_meters(self, peer_id: int) -> None:
...
+        self.pivot_gauge = metrics_registry.gauge('trinity.p2p/sync/pivot_rate.gauge')
+        self.pivot_gauge.set_value(0)
+        self.pivot_timestamps = []
```
pipermerriam, line 97: Just pointing out that the current implementation will cause this data structure to grow unbounded.  It'll take a lot of pivots before it's large so practically not a concern, but still something I think we should avoid.  Consider using a `collections.dequeu(maxlen=XXX)` which will automatically push items off the other end once the structure is full.

njgheorghita:
@cburgdorf I've moved the code to track pivots from the `p2p` code to the `Syncer` component. I'm not convinced this is the optimal place for the code to live, so let me know if there's a better location. Also, we're now tracking the weighted moving average of pivot events (pyformance's builtin tracking equation), it seems like this metric is now closer to how you envisioned it?
Screen Shot 2020-09-11 at 10 25 05 AM
.

trinity/components/builtin/metrics/service/base.py
```diff
@@ -1,15 +1,36 @@
...
 from trinity._utils.logging import get_logger
 
 
+class ExtendedInfluxReporter(InfluxReporter):
+    """
+    `InfluxReporter` extended to enable sending annotations to InfluxDB
+    """
+    def send_annotation(self, annotation_data: str) -> None:
+        path = f"/write?db={self.database}&precision=s"
+        url = f"{self.protocol}://{self.server}:{self.port}{path}"
```
pipermerriam, line 24: Just noting that the python standard library comes with a tool that does this:

https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlunparse

pipermerriam (approved):

trinity/components/builtin/metrics/service/base.py
```diff
@@ -1,15 +1,36 @@
...
 
 from trinity.components.builtin.metrics.abc import MetricsServiceAPI
 from trinity.components.builtin.metrics.registry import HostMetricsRegistry
+from trinity.components.builtin.metrics.sync_metrics_registry import SyncMetricsRegistry
 from trinity._utils.logging import get_logger
 
 
+class ExtendedInfluxReporter(InfluxReporter):
+    """
+    `InfluxReporter` extended to enable sending annotations to InfluxDB
```
cburgdorf, line 20: Better to use double backticks because in rst syntax single backticks only give you cursive text as in *InfluxReporter* whereas double backticks give you the code formatting as in `InfluxReporter`

trinity/components/builtin/metrics/service/base.py
```diff
@@ -52,6 +74,21 @@ def registry(self) -> HostMetricsRegistry:
         """
         return self._registry
 
+    @property
+    def reporter(self) -> ExtendedInfluxReporter:
+        """
+        Return the :class:`trinity.components.builtin.metrics.service.base.ExtendedInfluxReporter`
+        with influxdb connection details and annotations supported.
+        """
+        return self._reporter
```
cburgdorf, line 83: I think I would prefer to not expose the `ExtendedInfluxReporter` here and instead enrich the `BaseMetricsService`with a `def send_annotation(self, annotation_data: str)` method (just a passthrough to the underlying reporter). That would be more in line with the general [tell, don't ask](https://martinfowler.com/bliki/TellDontAsk.html) and prevent the reporter from becoming a leaky abstraction.

trinity/components/builtin/metrics/service/base.py
```diff
@@ -52,6 +74,21 @@ def registry(self) -> HostMetricsRegistry:
...
+
+    def subscribe_to_pivot_events(self, event_bus: EndpointAPI) -> None:
```
cburgdorf, line 85: I would also argue that we don't want that method on the metrics service. I would consider the metrics service to be agnostic from any concrete metric that we want to track. It's just there to provide the generic foundation to send metrics but shouldn't cater to a specific one. I also don't think there is a strong argument why that code would *need* to sit inside this class.

trinity/components/builtin/metrics/service/noop.py
```diff
@@ -23,6 +24,7 @@ def __init__(self,
                  reporting_frequency: int = 10):
 
         self._registry = NoopMetricsRegistry()
+        self._reporter = ExtendedInfluxReporter()
```
cburgdorf, line 27: We can also remove this if we follow my advice from above to not expose the reporter at all.

trinity/components/builtin/metrics/service/noop.py
```diff
@@ -32,6 +34,14 @@ def registry(self) -> NoopMetricsRegistry:
         """
         return self._registry
 
+    @property
+    def reporter(self) -> ExtendedInfluxReporter:
+        """
+        Return the :class:`trinity.components.builtin.metrics.service.base.ExtendedInfluxReporter`
+        with dummy influxdb connection details.
+        """
+        return self._reporter
```
cburgdorf, line 43: And this.

trinity/components/builtin/metrics/sync_metrics_registry.py
```diff
@@ -0,0 +1,39 @@
...
+                 ) -> None:
+        self.host = metrics_registry.host
+        self.pivot_meter = metrics_registry.meter('trinity.p2p/sync/pivot_rate.meter')
+        self.reporter = metrics_reporter
+        self.event_bus = event_bus
+
+    def subscribe_to_pivot_events(self) -> None:
+        self.event_bus.subscribe(PivotEvent, lambda event: self.record_pivot(event.payload))
+
+    def record_pivot(self, block_number: int) -> None:
```
cburgdorf, line 31: `block_number` should be of type `BlockNumber`

trinity/components/builtin/metrics/service/base.py
```diff
@@ -1,15 +1,36 @@
...
 
+class ExtendedInfluxReporter(InfluxReporter):
+    """
+    `InfluxReporter` extended to enable sending annotations to InfluxDB
+    """
+    def send_annotation(self, annotation_data: str) -> None:
+        path = f"/write?db={self.database}&precision=s"
+        url = f"{self.protocol}://{self.server}:{self.port}{path}"
+        auth_header = self._generate_auth_header()
+        requests.post(url, data=annotation_data, headers=auth_header)
```
cburgdorf, line 26: It just occurred to me that this means one synchronous request per send annotation which isn't optimal especially in the same process that runs the peerpool / sync. It will slow down the event loop. I realize that the internals of `InfluxReporter` create synchronous requests as well but it's not making the situation any better :-/

It would be great to spend some thought to figure out how to make that request async (I believe `requests` is strictly snychronous though)

trinity/components/builtin/metrics/sync_metrics_registry.py
```diff
@@ -0,0 +1,39 @@
...
+                 metrics_reporter: InfluxReporter,
+                 event_bus: EndpointAPI
+                 ) -> None:
+        self.host = metrics_registry.host
+        self.pivot_meter = metrics_registry.meter('trinity.p2p/sync/pivot_rate.meter')
+        self.reporter = metrics_reporter
+        self.event_bus = event_bus
+
+    def subscribe_to_pivot_events(self) -> None:
+        self.event_bus.subscribe(PivotEvent, lambda event: self.record_pivot(event.payload))
```
cburgdorf, line 29: This doesn't take care of unsubscribing which is a problem. But I'll leave another comment on a different line number that will argue for getting rid of the event bus here anyway.

trinity/components/builtin/syncer/component.py
```diff
@@ -215,7 +215,8 @@ async def sync(self,
             event_bus,
             args.sync_from_checkpoint,
             args.force_beam_block_number,
-            not args.disable_backfill
+            not args.disable_backfill,
+            args.enable_metrics,
```
cburgdorf, line 219: I'm beginning to lean towards passing in `SyncMetricsRegistry` here and exposing a property `enabled`. That way we can get rid of using the event bus as the used mechanism. While I do think event bus events, especially internal ones, aren't much performance overhead, the overhead may start to accumulate as we begin to track more and more things. Especially more frequent ones.

If you pass in the `SyncMetricsRegistry`, you can simply call `metrics_registry.record_pivot(pivot)` (or ideally `await metrics_registry.record_pivot(pivot)` if you figure out how to make the `POST` request `async`) without having to go through the event bus at all.

trinity/components/builtin/metrics/abc.py
```diff
@@ -22,3 +24,11 @@ def __init__(self,
...
     def registry(self) -> MetricsRegistry:
         ...
+
+    @property
+    @abstractmethod
+    def reporter(self) -> InfluxReporter:
+        ...
+
+    def subscribe_to_pivot_events(self, event_bus: EndpointAPI) -> None:
+        ...
```
cburgdorf, line 34: I left more comments further down, just wanted to come back here and comment that these changes could be rolled back then.

trinity/components/builtin/syncer/component.py
```diff
@@ -358,6 +359,7 @@ async def do_run(self, event_bus: EndpointAPI) -> None:
 
         if boot_info.args.enable_metrics:
             metrics_service = metrics_service_from_args(boot_info.args, AsyncioMetricsService)
+            metrics_service.subscribe_to_pivot_events(event_bus)
```
cburgdorf, line 362: If we follow my advice of getting rid of the event bus you won't up needing this and don't need to care about how to deal with the (currently unhandled) subscription.

trinity/sync/beam/service.py
```diff
@@ -72,6 +75,12 @@ async def _pivot_loop(self) -> None:
             do_pivot = await self._monitor_for_pivot(beam_syncer)
             if do_pivot:
                 self.logger.info("Pivoting Beam Sync to a newer header...")
+                if self.metrics_enabled:
+                    latest_block_number = beam_syncer._body_syncer._latest_block_number
+                    await self.event_bus.broadcast(
```
cburgdorf, line 80: This place would end up as `metrics_registry.record_pivot(latest_block_number)`

cburgdorf (commented):
Alright left my comments inline. I think we need to iterate one more time but I think we are getting close!

trinity/sync/beam/service.py
```diff
@@ -75,12 +74,11 @@ async def _pivot_loop(self) -> None:
             do_pivot = await self._monitor_for_pivot(beam_syncer)
             if do_pivot:
                 self.logger.info("Pivoting Beam Sync to a newer header...")
-                if self.metrics_enabled:
+                print("XXXXXXXXXXXXXXXXXXXXXXXXXX")
+                if self.sync_metrics_registry.enabled: # what if metrics_registry is None?
```
cburgdorf, line 78: If we make the `sync_metrics_registry` optional, then dropping the `enabled` and just going for `if sync_metrics_registry` would be fine by me.

trinity/components/builtin/metrics/service/base.py
```diff
@@ -17,13 +17,14 @@
...
+    ``InfluxReporter`` extended to enable sending annotations to InfluxDB
     """
-    def send_annotation(self, annotation_data: str) -> None:
+    async def send_annotation(self, annotation_data: str) -> None:
         path = f"/write?db={self.database}&precision=s"
         url = f"{self.protocol}://{self.server}:{self.port}{path}"
         auth_header = self._generate_auth_header()
-        requests.post(url, data=annotation_data, headers=auth_header)
+        async with aiohttp.ClientSession() as session:
+            await session.post(url, data=annotation_data, headers=auth_header)
```
cburgdorf, line 27: Nice, I skimmed over `aiohttp` but it wasn't immediately obvious to me whether they provide that functionality or not. Thanks for following up on that!

trinity/components/builtin/syncer/component.py
```diff
@@ -215,11 +226,12 @@ async def sync(self,
...
             args.force_beam_block_number,
-            not args.disable_backfill
+            not args.disable_backfill,
+            sync_metrics_registry,
         )
 
-        async with background_asyncio_service(syncer) as manager:
-            await manager.wait_finished()
+        services = (syncer, metrics_service,) if metrics_service else (syncer,)
+        await run_background_asyncio_services(services)
```
njgheorghita, line 234: @cburgdorf Ok, I think I finally figured out where we were going wrong... but if you can validate my theory that would be super! Basically we were never running this `metrics_service` within the syncer component. I've changed this, and everything (both annotations and the 15min pivot rate) are reporting correctly to grafana. However, if there's something I'm missing with this fix, let me know!

cburgdorf: Ah, sorry, I totally missed that this is creating its own instance of the `MetricsService` (line 215). That should not be. There is already one instance created a bit further down.

https://github.com/ethereum/trinity/blob/e5beac6ff65dedb1b8c2692012bca5dab7b69eef/trinity/components/builtin/syncer/component.py#L359-L364

That also explains why we were seeing the log messages to appear twice. We only want one single instance of the `MetricsService` per process (if a process even needs its own which is the case here). The instance that already exists is currently used by the peer pool but since both sync and peer pool run in the same process they can both share that instance.

trinity/sync/beam/service.py
```diff
@@ -72,6 +75,9 @@ async def _pivot_loop(self) -> None:
             do_pivot = await self._monitor_for_pivot(beam_syncer)
             if do_pivot:
                 self.logger.info("Pivoting Beam Sync to a newer header...")
+                if self.sync_metrics_registry:
+                    latest_block = BlockNumber(beam_syncer._body_syncer._latest_block_number)
```
cburgdorf, line 79: I think that `beam_syncer._body_syncer._latest_block_number` is already of type `BlockNumber` so that you don' need to wrap it again. If it is of type `int` than a type annotation inside body syncer needs to be done to fix it on that end because it really should be `BlockNumber`. It's not a big issue but it just crossed my eyes so I wanted to bring it up.

trinity/components/builtin/syncer/component.py
njgheorghita: 👍 I've removed the redundant `metrics_service` and everything is working as expected locally, so I think we're finally ready for a proper review on this pr 😅

trinity/components/builtin/metrics/service/base.py
cburgdorf: I'm still not a fan of exposing the reporter here (not sure why this was marked as resolved) and would instead prefer `def send_annotation(self, annotation_data: str)` to the `BaseMetricsService` instead. I think exposing the reporter and using it down the road is a bit of a code smell but I don't feel too strong about it to hold back the PR. It could also be fixed with a follow up PR.

trinity/components/builtin/metrics/registry.py
```diff
@@ -46,7 +46,7 @@ class NoopMetricsRegistry(HostMetricsRegistry):
     """
 
     def __init__(self) -> None:
-        pass
+        self.host = None
```
cburgdorf, line 49: I would prefer to pass the `host` into the `SyncMetricsRegistry` if it needs it instead of making it mandatory here.

trinity/components/builtin/syncer/component.py
```diff
@@ -205,7 +210,14 @@ async def sync(self,
...
                    base_db: AtomicDatabaseAPI,
                    peer_pool: BasePeerPool,
-                   event_bus: EndpointAPI) -> None:
+                   event_bus: EndpointAPI,
+                   metrics_service: MetricsServiceAPI = None) -> None:
+
+        # create registry for tracking pivot metrics
+        sync_metrics_registry = SyncMetricsRegistry(
+            metrics_service.registry,
+            metrics_service.reporter,
```
cburgdorf, line 219: This could take in the whole `metrics_service` so that `SyncMetricsRegistry` can call `metrics_service.send_annotation(...)` internally and hence getting rid of the leaky `reporter` abnstraction.

cburgdorf (approved):
Left a few more architectural cleanup  suggestions but only minor stuff and overall this looks good to me. The change became much smaller which is usually a good sign :)

Looking forward to see the metric going live soon!

njgheorghita:
> I'm still not a fan of exposing the reporter here (not sure why this was marked as resolved) and would instead prefer def send_annotation(self, annotation_data: str) to the BaseMetricsService instead. I think exposing the reporter and using it down the road is a bit of a code smell but I don't feel too strong about it to hold back the PR. It could also be fixed with a follow up PR.

@cburgdorf Thanks for the help in getting this through! I must have resolved it on accident at some point in the review process, but I've gone ahead and made the changes here rather than in a later pr. If you can give it a quick scan to double-check that everything looks good, I'll merge and then get this up on the dashboard!

cburgdorf:
@njgheorghita Almost forgot to answer. Thanks for making the changes. That looks great! :+1:

PR #2041, opened 2020-09-11
wait_first() no longer raises TimeoutError when there are rogue tasks

gsalgado:
If any tasks don't return after being cancelled, we simply log a warning
now. That's consistent with asyncio's behavior of logging a warning in
case a task terminates but its result/exception are ignored.


gsalgado:
@cburgdorf guess you forgot about this one? :-)

cburgdorf (approved):
I'm fine with that :+1:

cburgdorf:
@gsalgado Sorry, that slipped through the cracks...

Issue #2035, opened 2020-09-10
Trinity may skip a preferred-peer when running with --enable-experimental-eth1-peer-tracking


cburgdorf:
### What is wrong?

Here is Trinity running with a preferred node **and** the `--enable-experimental-eth1-peer-tracking` flag

The log [clearly shows](https://gist.github.com/cburgdorf/2d6dca510a8ec78fbce209ff9722a3c5#file-gistfile1-txt-L544) Trinity is started with a preferred peer but even after several minutes (and after establishing various connections to different peers) it has not once tried to reach out to our specified peers.

Here's the same without using `--enable-experimental-eth1-peer-tracking`

https://gist.github.com/cburgdorf/a36a95b321294fe3002d577c2c2e3669

We can see that it reaches out to `172.33.0.2` pretty much immediately. 

I *think* what happened is that my local good peer has ended up on the block list and got filtered out. So I believe there may be two issues:

1. Why did it end up on that list (if that is true)

2. How can we make it so that a preferred node can bypass this filter

### How can it be fixed

If it really is the `skip_filter_fn` that causes the `preferred_node` to be filtered out then I think that could be fixed here.

https://github.com/ethereum/trinity/blob/c3f7c02a74a9b7a0f61f90a56e2f841bfb7b22d7/p2p/discovery.py#L244-L250

It looks like the derived `PreferredNodeDiscoveryService` should have the power to overrule the skip list if it is trying to skip one of the preferred peers.

cburgdorf:
Fixed by #2042

PR #2042, opened 2020-09-11
Never skip preferred peers from list of connection candidates

gsalgado:


cburgdorf (approved):
Nice :+1: I think we should create `2042.bugfix.rst` fragment for that to make sure it gets mentioned in the next release notes. Also a test would be a nice bonus if you feel like going that extra mile ;)

cburgdorf:
Fixes #2035

gsalgado:
Added a newsfragment file, but when I tried to add a test I noticed we no longer have any discovery tests left. Turns out they were removed in #2010

pipermerriam:
This has resulted in an unfortunate side effect.

```
   DEBUG  2020-09-15 21:14:46,459     ConnectionTracker  <Node(0xc74e5d@45.79.78.217)> will not be retried for 33m8s because HandshakeFailure
   DEBUG  2020-09-15 21:14:46,489     ConnectionTracker  <Node(0x9440be@45.79.76.193)> will not be retried for 36m2s because HandshakeFailure
   DEBUG  2020-09-15 21:14:48,936     ConnectionTracker  <Node(0x9440be@45.79.76.193)> will not be retried for 37m23s because HandshakeFailure
   DEBUG  2020-09-15 21:14:54,572     ConnectionTracker  <Node(0x9440be@45.79.76.193)> will not be retried for 38m42s because HandshakeFailure
   DEBUG  2020-09-15 21:14:57,000     ConnectionTracker  <Node(0x9440be@45.79.76.193)> will not be retried for 39m59s because HandshakeFailure
   DEBUG  2020-09-15 21:15:02,170     ConnectionTracker  <Node(0x9440be@45.79.76.193)> will not be retried for 41m12s because HandshakeFailureTooManyPeers
   DEBUG  2020-09-15 21:15:03,024     ConnectionTracker  <Node(0x9440be@45.79.76.193)> will not be retried for 42m24s because HandshakeFailure
   DEBUG  2020-09-15 21:15:08,847     ConnectionTracker  <Node(0x9440be@45.79.76.193)> will not be retried for 43m34s because HandshakeFailure
   DEBUG  2020-09-15 21:15:12,537     ConnectionTracker  <Node(0x9440be@45.79.76.193)> will not be retried for 44m42s because HandshakeFailure
   DEBUG  2020-09-15 21:15:14,514     ConnectionTracker  <Node(0x9440be@45.79.76.193)> will not be retried for 45m48s because HandshakeFailure
   DEBUG  2020-09-15 21:15:20,188     ConnectionTracker  <Node(0x9440be@45.79.76.193)> will not be retried for 46m53s because HandshakeFailure
```

If you have a preferred peer which you can't connect to for some reason, the API will now continually hammer attempting to connect to that peer, resulting in them accumulating large retry delays.  We still try to connect to them so this isn't critical, but it is resulting in an unfortunate situation where our network DB will "dislike" these peers when they are not specified as preferred.

gsalgado:
We could not blacklist preferred peers, or revert this change. Any preference, @pipermerriam and @cburgdorf?

cburgdorf:
>the API will now continually hammer attempting to connect to that peer, resulting in them accumulating large retry delays

The API not giving up on the preferred peer (with increasing delays) strikes me as the right thing to do.

>but it is resulting in an unfortunate situation where our network DB will "dislike" these peers when they are not specified as preferred.

Yeah, so to me, it is that part that needs to be fixed. I think the user consciously picking a preferred peer should have preference over the nodes heuristics to categorize a peer as good or bad.

pipermerriam:
This isn't immediately urgent since as long as the peer is provided via `--preferred-peer` they will still be given as a candidate and connected to if possible(even though they are actively blacklisted).

I believe this won't be trivial to re-work since our blacklisting logic is relatively heavy handed so I think we can ignore it for a while.  I'll open an issue.

PR #2044, opened 2020-09-11
Ressurrect and fix tests-trio/p2p-trio/test_discovery.py

gsalgado:
It was removed by mistake in #2010


pipermerriam (approved):
Thanks for catching this.

PR #2046, opened 2020-09-15
py-evm and trinity are no longer part of the same repo

indi-ca:
Simple text change


cburgdorf:
Thank you for the contribution :pray:

Issue #1895, opened 2020-07-14
Error on inbound connection


carver:
### What is wrong?

```wait_for=<_GatheringFuture pending cb=[<TaskWakeupMethWrapper object at 0x7f8028142c10>()]>>
   ERROR  2020-07-14 14:11:42,771            FullServer  Unexpected error handling handshake with ('195.14.184.226', 58872)
Traceback (most recent call last):
  File "/home/jcarver/code/trinity/trinity/server.py", line 135, in receive_handshake
    await self._receive_handshake(reader, writer)
  File "/home/jcarver/code/trinity/trinity/server.py", line 160, in _receive_handshake
    protocol_handshakers=handshakers,
RuntimeError: coroutine ignored GeneratorExit
   ERROR  2020-07-14 14:11:42,774               asyncio  Task exception was never retrieved
future: <Task finished coro=<<async_generator_athrow without __name__>()> exception=RuntimeError('cannot reuse already awaited coroutine')>
Traceback (most recent call last):
  File "/home/jcarver/code/trinity/p2p/multiplexer.py", line 289, in _stream_protocol_messages
    yield await msg_queue.get()
RuntimeError: cannot reuse already awaited coroutine
```

### How can it be fixed

:man_shrugging:

gsalgado:
This is weird; my node can receive incoming connections (at least from another locally running trinity instance). Can you try that to see if it works? Is this on master or on #1892?

carver:
> Is this on master or on #1892?

It first showed up when I started with #1892 -- but then because UPnP was configured, even when I ran it without the patch I was getting the error. It seems to go away when I do this:
```patch
diff --git a/trinity/server.py b/trinity/server.py
index 9402862b..e006f03f 100644
--- a/trinity/server.py
+++ b/trinity/server.py
@@ -172,7 +172,10 @@ class BaseServer(Service, Generic[TPeerPool]):
             peer = factory.create_peer(connection)
 
             if self.peer_pool.is_full:
-                await peer.disconnect(DisconnectReason.TOO_MANY_PEERS)
+                try:
+                    await peer.disconnect(DisconnectReason.TOO_MANY_PEERS)
+                except AttributeError:
+                    pass
                 return
             elif not self.peer_pool.is_valid_connection_candidate(peer.remote):
                 await peer.disconnect(DisconnectReason.USELESS_PEER)
```
(weirdly the exception that inspired the patch was logged separately, later)

So I suspect it will actually get fixed by #1840

[...]

PR #2049, opened 2020-09-18
Ensure stream_protocol_messages() returns when multiplexer stops streaming

gsalgado:
That method could end up waiting forever on a queue.get() call if the
multiplexer stops streaming (i.e. adding items to the queue), and then
asyncio.BaseServer would eventually throw a GeneratorExit into it
(https://bugs.python.org/issue30083) causing a lot of cryptic errors

Closes: #1895

### DIFF
--- a/p2p/handshake.py
+++ b/p2p/handshake.py
@@ -367,22 +367,14 @@ async def receive_dial_in(reader: asyncio.StreamReader,
         writer=writer,
         private_key=private_key,
     )
-    try:
-        multiplexer, devp2p_receipt, protocol_receipts = await negotiate_protocol_handshakes(
-            transport=transport,
-            p2p_handshake_params=p2p_handshake_params,
-            protocol_handshakers=protocol_handshakers,
-        )
-    except BaseException:
-        # Note: This is one of two places where we manually handle closing the
-        # reader/writer connection pair in the event of an error during the
-        # peer connection and handshake process.
-        # See `p2p.auth.handshake` for the other.
-        try:
-            await transport.close()
-        except ConnectionResetError:
-            transport.logger.debug("Could not wait for transport to close")
-        raise
+
+    # In case of any errors, negotiate_protocol_handshakes() will close the multiplexer,
+    # which in turn closes the transport, so we don't need to worry about that here.
+    multiplexer, devp2p_receipt, protocol_receipts = await negotiate_protocol_handshakes(
+        transport=transport,
+        p2p_handshake_params=p2p_handshake_params,
+        protocol_handshakers=protocol_handshakers,
+    )
 
     connection = Connection(
         multiplexer=multiplexer,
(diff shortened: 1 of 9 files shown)


gsalgado:
I was hopeful that fixing #1895 would magically make #2045 go away as well, but that didn't happen :-(

p2p/multiplexer.py
```diff
@@ -271,37 +276,37 @@ def stream_protocol_messages(self,
...
-                except asyncio.QueueEmpty:
-                    yield await msg_queue.get()
+                else:
+                    # Need to run this with a timeout so that we return in case the multiplexer
+                    # stops streaming.
+                    # Using a short timeout here (0.1s) caused msg queues to get full and messages
+                    # to be discarded. Using 0.5s should not be a problem as it only means this
+                    # coroutine may take up to 0.5s to return after the multiplexer stops.
+                    try:
+                        yield await asyncio.wait_for(msg_queue.get(), timeout=0.5)
```
carver, line 307: Another option to consider is something like adding an event that triggers when streaming ends, and then this does a `wait_first([msg_queue.get(), stream_end.wait()])`. That way we don't have to cycle this loop `2 * num_peers` times per second, even when peers are idle.

p2p/handshake.py
```diff
@@ -367,22 +367,14 @@ async def receive_dial_in(reader: asyncio.StreamReader,
...
-            p2p_handshake_params=p2p_handshake_params,
-            protocol_handshakers=protocol_handshakers,
-        )
-    except BaseException:
-        # Note: This is one of two places where we manually handle closing the
-        # reader/writer connection pair in the event of an error during the
-        # peer connection and handshake process.
-        # See `p2p.auth.handshake` for the other.
-        try:
-            await transport.close()
```
carver, line 382: Why don't we need to close the transport anymore if there's a problem during the handshake?

carver (approved):

p2p/multiplexer.py
gsalgado: I did consider that, but we need the return value of the `msg_queue.get()` call, so `wait_first()` wouldn't work. It could be done with `asyncio.wait()` and then checking whether the finished task was one or the other, but the code would become less straightforward for what seemed to be a not very significant improvement.  Happy to reconsider that if you think it's worth it though

p2p/handshake.py
gsalgado: `negotiate_protocol_handshakes()` takes care of that -- I've added a comment explaining that (line 371 in the diff)

p2p/multiplexer.py
carver: Not a huge deal, but also not sure why we need to check quite as often, since it is checking for a corner case and has some performance hit. Would it be just as fine to timeout every 5?

gsalgado: Yeah, it should be ok to use a timeout of 5s

PR #2050, opened 2020-09-18
Un-clog the predictive task queue

### LINKED ISSUE #2047: _preview_address_load() never completing
### What is wrong?

There's a leak in the number of active coroutines. After applying this patch to `async-service`: https://github.com/ethereum/async-service/pull/77 it raised:

```
  File "/home/jcarver/code/trinity/trinity/sync/full/chain.py", line 1109, in _preview_ready_blocks
    await self._block_importer.preview_transactions(
  File "/home/jcarver/code/trinity/trinity/sync/beam/chain.py", line 802, in preview_transactions
    self.manager.run_task(self._preview_address_load, header, parent_state_root, transactions)
  File "/home/jcarver/code/async-service/async_service/asyncio.py", line 337, in run_task
    self._common_run_task(task)
  File "/home/jcarver/code/async-service/async_service/base.py", line 289, in _common_run_task
    self._add_child_task(task.parent, task)
  File "/home/jcarver/code/async-service/async_service/base.py", line 304, in _add_child_task
    raise TooManyChildrenException(
async_service.exceptions.TooManyChildrenException: Tried to add more than 1000 child tasks. Most common tasks: [('_preview_address_load[daemon=False]', 1000), ('run[daemon=False]', 1)]
```
### How can it be fixed

Some options:
- Patch the problem, by putting a timeout on `_preview_address_load()`
- Figure out why it's not finishing. Maybe... Wrong state root? Predictive hang that is never woken up by urgent data? Something else?

carver:
### What was wrong?

Fix #2047 

The predictive node task queue backs up easily, until it can't accept any more tasks. I haven't figured out why yet. But it turns out that the queue was staying clogged even when the trie nodes were available, because there was no check for node presence when pulling the task off the queue.

The issue presented as _preview_address_load() never completing, and ending up with way too many live coroutines.

### How was it fixed?

Check whether any of the predictive node download tasks are stale (because the node was downloaded, or generated locally), before asking any peers for the data. This seems to resolve the coroutine leak.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/beam/state.py
+++ b/trinity/sync/beam/state.py
@@ -555,6 +555,21 @@ class BeamDownloader(Service, PeerSubscriber):
                 # Re-attempt
                 continue
 
+            # Find any hashes that were discovered through other means, like urgent requests:
+            existing_hashes = await asyncio.get_event_loop().run_in_executor(
+                None,
+                self._get_unique_present_hashes,
+                hashes,
+            )
+            # If any hashes are already found, clear them out and retry
+            if existing_hashes:
+                # Wake up any paused preview threads
+                await self._wakeup_preview_waiters(existing_hashes)
+                # Clear out any tasks that are no longer necessary
+                await self._maybe_useful_nodes.complete(batch_id, tuple(existing_hashes))
+                # Restart from the top
+                continue
+
             try:
                 peer = await asyncio.wait_for(
                     self._queen_tracker.pop_fastest_peasant(),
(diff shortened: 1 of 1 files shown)


trinity/sync/beam/state.py
```diff
@@ -645,21 +660,26 @@ async def _get_nodes(
...
+        # Wake up any coroutines waiting for the particular data that was returned.
+        #   (If no data returned, then no coros should wake up, and we can skip the block)
+        preview_waiters = await asyncio.get_event_loop().run_in_executor(
+            None,
+            self._get_preview_waiters,
+            node_hashes,
+        )
+        for waiter in preview_waiters:
+            waiter.set()
+            await asyncio.sleep(0)
```
gsalgado, line 678: Do we still need to sleep here?

gsalgado (approved):
LGTM, but I have to say I don't fully understand all the implications of this change

trinity/sync/beam/state.py
carver: Ah, maybe not. I think `set()`ing an event can actually be quite slow when `PYTHONASYNCIODEBUG=1`, which is why I added in the first place. Maybe I'll annotate with a timer to see if it's plenty fast with the debug off. (I hope it would be)

carver: Nothing meaningful showed up with a Timer, so I removed the sleep for now.

PR #2051, opened 2020-09-18
Write component to support preferred nodes with discovery disabled

### LINKED ISSUE #440: Decouple preferred nodes plugin from discovery plugin
1. The concept of preferred nodes should be completely removed from discovery.
2. A preferred node plugin should be written up that activates if the `--preferred-node` flag is provided.  That plugin monitors the peer pool, and sends `ConnectToNodeCommand` events when it finds that the peer pool is not empty and isn't connected to one of the preferred nodes.
    - A cursory glance suggests that this will require a new event based API for querying what nodes the peer pool is connected to.

This would completely decouple discovery and the concept of preferred nodes.

_Originally posted by @pipermerriam in https://github.com/ethereum/trinity/pull/436#issuecomment-474992361_

*Note from @carver: should we remove the constraint that the peer pool must be empty? I don't understand the motivation.*

njgheorghita:
### What was wrong?
Plugin written that continuously pings the peer pool for its connected peers. If there's room in the pool, connect to an eligible, preferred peer. Isolates the concept of "preferred nodes" from discovery, so that a user can supply a `--preferred-node` with `--discovery-disabled`.

Fixes #440 

### How was it fixed?
- Removed the `StaticDiscoveryService` that was a temporary stopgap from #436
- Wrote a `PreferredNodeComponent` that will send a `ConnectToNodeCommand` for a preferred node to the peer pool if the pool is not full.
- It's not obvious to me that this requires its own test since the use case seems to be covered in [this test](https://github.com/ethereum/trinity/blob/a525735d342366a6d987a66a7f90b47ae4832d51/tests/integration/test_trinity_cli.py#L72).

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/trinity/components/builtin/preferred_node/component.py
+++ b/trinity/components/builtin/preferred_node/component.py
@@ -1,0 +1,122 @@
+import random
+import time
+from typing import Dict, Iterator, List
+
+from async_service import Service, background_trio_service
+from eth_utils import to_tuple
+from lahja import EndpointAPI
+
+from p2p import trio_utils
+from p2p.abc import NodeAPI
+from p2p.constants import DEFAULT_MAX_PEERS
+from trinity._utils.logging import get_logger
+from trinity.constants import TO_NETWORKING_BROADCAST_CONFIG
+from trinity.extensibility import TrioIsolatedComponent
+from trinity.protocol.common.events import (
+    ConnectToNodeCommand,
+    GetConnectedPeersRequest,
+)
+
+
+class PreferredNodeComponent(TrioIsolatedComponent):
+    """
+    Monitor peer pool and attempt to connect to a preferred node
+    when the pool is not at full capacity.
+    """
+    name = "PreferredNode"
+
+    @property
+    def is_enabled(self) -> bool:
+        return bool(self.preferred_nodes)
+
+    @property
+    def max_peers(self) -> int:
+        if self._boot_info.args.max_peers is not None:
+            return self._boot_info.args.max_peers
+        return DEFAULT_MAX_PEERS
+
+    @property
+    def preferred_nodes(self) -> List[NodeAPI]:
+        return self._boot_info.args.preferred_nodes
+
+    async def do_run(self, event_bus: EndpointAPI) -> None:
+        service = PreferredNodeService(event_bus, self.preferred_nodes, self.max_peers)
+        async with background_trio_service(service) as manager:
+            await manager.wait_finished()
+
+
+class PreferredNodeService(Service):
+    preferred_node_recycle_time: int = 300
+    pool_availability_check_period: int = 5
+    _preferred_node_tracker: Dict[NodeAPI, float] = None
+
+    def __init__(
+        self, event_bus: EndpointAPI, preferred_nodes: List[NodeAPI], max_peers: int
+    ) -> None:
+        self.event_bus = event_bus
+        self.max_peers = max_peers
+        self.preferred_nodes = preferred_nodes
+        self._preferred_node_tracker = {}
+        self.logger = get_logger('trinity.components.preferred_node.PreferredNodeService')
+
+    async def run(self) -> None:
+        while self.manager.is_running:
+            await self.periodically_try_to_fill_pool_with_preferred_nodes()
+
+    async def periodically_try_to_fill_pool_with_preferred_nodes(self) -> None:
+        """
+        Check pool for available peer slots, and brodacast command to connect
+        to preferred nodes if there's room in the pool.
+        """
+        # seed _preferred_node_tracker so all preferred nodes are considered eligible initially
+        for node in self.preferred_nodes:
+            self._preferred_node_tracker[node] = time.monotonic() - self.preferred_node_recycle_time
+
+        await self.event_bus.wait_until_any_endpoint_subscribed_to(GetConnectedPeersRequest)
+
+        async for _ in trio_utils.every(self.pool_availability_check_period):
+            response = await self.event_bus.request(GetConnectedPeersRequest())
+            available_slots = self.max_peers - len(response.peers)
+            if available_slots > 0:
+                connected_peers = [peer.session.remote for peer in response.peers]
+                eligible_nodes = self._sample_eligible_preferred_nodes(
+                    connected_peers,
+                    available_slots
+                )
+                for node in eligible_nodes:
+                    await self._connect_to_preferred_node(node)
+
+    async def _connect_to_preferred_node(self, node: NodeAPI) -> None:
+        """
+        Broadcast a command for peer pool to connect to a preferred node,
+        and track the time of attempted connection.
+        """
+        self._preferred_node_tracker[node] = time.monotonic()
+        await self.event_bus.broadcast(
+            ConnectToNodeCommand(node),
+            TO_NETWORKING_BROADCAST_CONFIG,
+        )
+
+    def _sample_eligible_preferred_nodes(
+            self,
+            connected_peers: List[NodeAPI],
+            available_slots: int) -> List[NodeAPI]:
+        """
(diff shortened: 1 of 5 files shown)


trinity/components/builtin/preferred_node/component.py
```diff
@@ -0,0 +1,117 @@
...
+        connection attempt has not been made within preferred_node_recycle_time.
+        """
+        try:
+            random_preferred_node = self._get_random_preferred_node()
+        except NoEligibleNodes:
+            pass
+        else:
+            if random_preferred_node not in connected_peers:
+                await self.event_bus.broadcast(
+                    ConnectToNodeCommand(random_preferred_node),
```
carver, line 93: If I set 5 preferred nodes, I expect to try to connect to all 5 right away. An approach that fills as fast as possible might look something like:
```py
empty_slots = self.max_peers - len(response.peers)
preferred = self._get_eligible_preferred_nodes()
missing_preferred = set(preferred) - set(response.peers)
attempt_peers = random.sample(missing_preferred, empty_slots)
for peer in attempt_peers:
  self._preferred_node_tracker[peer] = time.monotonic()
  await self.event_bus.broadcast(
    ...

File: trinity/components/builtin/preferred_node/component.py [hunk 2]
```diff
@@ -0,0 +1,117 @@
...
+
+    @to_tuple
+    def _get_eligible_preferred_nodes(self) -> Iterator[NodeAPI]:
+        """
+        Return nodes from the preferred_nodes which have not been used within
+        the last preferred_node_recycle_time
+        """
+        for node in self.preferred_nodes:
+            last_used = self._preferred_node_tracker[node]
+            if time.time() - last_used > self.preferred_node_recycle_time:
```
carver, line 105:
```suggestion
            if time.monotonic() - last_used > self.preferred_node_recycle_time:
```

Monotonic is better here to handle daylight savings and other clock shifts.

trinity/components/builtin/preferred_node/component.py
njgheorghita: Cool, yeah that makes sense. I took this motivation one step further, so that with every periodic call - we try to fill ALL available slots in the peer pool with a preferred peer (prior, it was always one by one). This doesn't seem problematic to me, since if we prefer certain nodes there's no problem with always trying to shove as many preferred nodes as possible into our pool.

njgheorghita (commented):
Thanks @carver ! I updated to the suggested changes, let me know if there's anything else that should be updated.

trinity/components/builtin/preferred_node/component.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        Return preferred_nodes that aren't currently connected and don't have an attempted
+        connection within the last preferred_node_recycle_time.
+        """
+        missing_preferred = [node for node in self.preferred_nodes if node not in connected_peers]
+        sample_size = min(available_slots, len(missing_preferred))
+        attempt_nodes = random.sample(missing_preferred, sample_size)
+        for node in attempt_nodes:
+            last_used = self._preferred_node_tracker[node]
+            if time.monotonic() - last_used > self.preferred_node_recycle_time:
```
carver, line 117: I think it's better to filter out these recently-attempted nodes before sampling, so that you can try to connect to as many nodes as possible.

carver (approved):
Yup, think this should be good to go, after this one last change!

trinity/components/builtin/preferred_node/component.py
```diff
@@ -0,0 +1,118 @@
...
+            return self._boot_info.args.max_peers
+        return DEFAULT_MAX_PEERS
+
+    @property
+    def preferred_nodes(self) -> List[NodeAPI]:
+        return self._boot_info.args.preferred_nodes
+
+    async def do_run(self, event_bus: EndpointAPI) -> None:
+        # Don't start service even if users provide `--max-peers 0` alongside preferred nodes
+        if self.max_peers > 0:
```
pipermerriam, line 45: This is odd to me.  This service doesn't "inject" peers into the peer pool.  It only responds to a request for peer candidates.  It seems to me that we should start it as long as any preferred peers were provided.  If `max_peers == 0` then the peer pool should never end up asking for connection candidates.

Otherwise, in some future iteration of the client that allows us to change `max-peers` at runtime, this service will not start, and then at some later time, we bump `max-peers` up to something non-zero and this service will never provide the preferred peers.

pipermerriam: Lol, I just realized that this service does *indeed* inject peer connections into the pool.

I think my logic still holds though.  It should just do nothing if the peer pool has no capacity which will always be the case when `max_peers == 0`

trinity/components/builtin/preferred_node/component.py
```diff
@@ -0,0 +1,118 @@
...
+            await self.periodically_try_to_fill_pool_with_preferred_nodes()
+
+    async def periodically_try_to_fill_pool_with_preferred_nodes(self) -> None:
+        """
+        Check pool for available peer slots, and brodacast command to connect
+        to preferred nodes if there's room in the pool.
+        """
+        # seed _preferred_node_tracker so all preferred nodes are eligible
+        for node in self.preferred_nodes:
+            self._preferred_node_tracker[node] = time.monotonic() - self.preferred_node_recycle_time
```
pipermerriam, line 76: These two lines shouldn't be necessary since you used a `collections.defaultdict`.  Since the dict defaults to `0` for all values, then all of the peers should always return a last-used-time that is eligible for inclusion.

trinity/components/builtin/preferred_node/component.py
carver: > in some future iteration of the client that allows us to change `max-peers` at runtime, this service will not start, and then at some later time, we bump `max-peers` up to something non-zero and this service will never provide the preferred peers.

I agree that we might as well support this.

trinity/components/builtin/preferred_node/component.py
carver: I believe the reason for this choice is that the absolute value of `monotonic` is undefined. I suppose the defaultdict could initialize with `time.monotonic() - self.preferred_node_recycle_time`

trinity/components/builtin/preferred_node/component.py
njgheorghita: Works for me, just to clarify my understanding - this requires a new event to communicate with the peer pool and request its `max_peers`? Then, this service will constantly poll for the `max_peers` value and only attempt to inject preferred nodes when there are available slots in the pool.

trinity/components/builtin/preferred_node/component.py
njgheorghita: It's maybe not the most elegant solution here, but since we only update the tracker when a connection attempt is made, it's necessary to 1. seed the tracker for all peers and 2. seed the tracker with a value that's guaranteed to be greater than the recycle time. This also has the effect that every peer is considered eligible on first pass, which is the behavior we want. If this still smells funny though, I can do some re-working here

trinity/components/builtin/preferred_node/component.py
pipermerriam: No, I don't think so.  I think we should just currently bank on the service being a noop (do nothing forever) in the event that `max_peers == 0`.  If at a later date we make `max_peers` adjustable during runtime, we'll do the infrastructure work of exposing that data over the event bus, but right now it would be premature.

trinity/components/builtin/preferred_node/component.py
pipermerriam: I think what is throwing me off is that you're using a `collections.defaultdict` but then your seeding it with data.  I don't believe the current implementation needs anything more than a regular `dict`.  Probably sufficient just to change to that.

Issue #2052, opened 2020-09-21
MalformedMessage during handshake on seemingly valid Hello msg


gsalgado:
I'm seeing lots of unexpected errors during handshake like the following:

```
 WARNING  2020-09-21 08:39:19,080             Transport  Deserializing list length (1) does not match sedes (2) decoding header (b'\x00\x00\xa2\xc1\x80\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'), (body=b'\x80\xf8\x9f\x05\xb8>Nethermind/v1.8.103-0-6bc67f44b-20200917/X64-Windows/Core3.1.5\xd8\xc5\x83eth>\xc5\x83eth?\xc5\x83eth@\xc5\x83ethA\x82v_\xb8@7\x1a\xe0\xd0\x073\x88\xeb\xd2\xb1\xa5e\xc2]\x145\x1e}\xdd\xbb\x8ee[\x8a*\x12\xbe\xda\xc2^\x1e\x1c\x94\x1f\xf3}\x89\xb2\xfd\xa1+J\xc7\x8b\x18c\x7f\xff\xf8Ub\xe3/\xf0\xc5\x1a\xd8\xbd\xc0t\xa0\x97\xae\xb1'), command_id=0)
   ERROR  2020-09-21 08:39:19,080            FullServer  Unexpected error handling handshake with ('82.26.85.206', 65335)
Traceback (most recent call last):
  File "/home/salgado/src/snakecharmers/trinity/p2p/transport.py", line 264, in recv
    header_data = _decode_header_data(padded_header[3:])
  File "/home/salgado/src/snakecharmers/trinity/p2p/transport.py", line 59, in _decode_header_data
    header_data = rlp.decode(data, sedes=HEADER_DATA_SEDES, strict=False)
  File "/home/salgado/virtualenvs/trinity/lib/python3.8/site-packages/rlp/codec.py", line 228, in decode
    obj = sedes.deserialize(item, **kwargs)
  File "/home/salgado/virtualenvs/trinity/lib/python3.8/site-packages/eth_utils/functional.py", line 45, in inner
    return callback(fn(*args, **kwargs))
  File "/home/salgado/virtualenvs/trinity/lib/python3.8/site-packages/rlp/sedes/lists.py", line 86, in deserialize
    raise ListDeserializationError(
rlp.exceptions.ListDeserializationError: Deserializing list length (1) does not match sedes (2)

The above exception was the direct cause of the following exception:

[...]

PR #2054, opened 2020-09-21
Support older versions of the RLPx protocol

gsalgado:
Closes: #2052


carver (commented):
Hm, been having more of those mysterious timeout failures lately. Aggravating. I restarted CI.

Assuming it passes, good to :shipit: !

carver (approved):

PR #2057, opened 2020-09-23
Fix flaky P2P server tests

gsalgado:
A couple tests could sometimes attempt to connect to a server
    before it was actually listening, and since the PeerPool
    suppresses the UnreachableError (as it is expected), the tests would
    then fail with a TimeoutError, Now the test fixture will only yield
    once the server is listening and the PeerPool used in tests does
    not suppress any errors when connecting.
    
    Also better logs on PeerPool to help debugging and better asyncio
    task names for connection behaviors.

Closes: #1504

### DIFF
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -110,6 +110,10 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
     _report_metrics_interval = 15  # for influxdb/grafana reporting
     _peer_boot_timeout = DEFAULT_PEER_BOOT_TIMEOUT
     _event_bus: EndpointAPI = None
+    # If set to True, any exceptions in ALLOWED_PEER_CONNECTION_EXCEPTIONS will be logged and
+    # suppressed. Should only be set to False in tests we we want to ensure a connection is
+    # successful.
+    _suppress_allowed_connection_exceptions: bool = True
 
     _handshake_locks: ResourceLock[NodeAPI]
     peer_reporter_registry_class: Type[PeerReporterRegistry[Any]] = PeerReporterRegistry[BasePeer]
@@ -488,7 +492,7 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
 
         if self._handshake_locks.is_locked(node):
             self.logger.info(
-                "Asked to connect to node when handshake lock is already locked, will wait")
+                "Asked to connect to %s when handshake lock is already locked, will wait", node)
 
         async with self.lock_node_for_handshake(node):
             if self.is_connected_to_node(node):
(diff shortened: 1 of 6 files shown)


trinity/server.py
```diff
@@ -119,6 +120,7 @@ async def run(self) -> None:
         try:
             async with tcp_listener:
                 self.manager.run_daemon_child_service(self.peer_pool)
+                self.ready.set()
```
carver, line 123: Should we wait for the peer pool to be confirmed launched here first before we mark the server as ready?

p2p/peer_pool.py
```diff
@@ -502,7 +506,13 @@ async def connect_to_node(self, node: NodeAPI) -> None:
                 async with self._connection_attempt_lock:
                     peer = await self.connect(node)
             except ALLOWED_PEER_CONNECTION_EXCEPTIONS:
-                return
+                if self._suppress_allowed_connection_exceptions:
+                    # These are all logged in self.connect(), so we simply return.
+                    return
+                raise
```
carver, line 512: *nitpick* style guide suggests doing the explicit branch here, like:
```suggestion
                else:
                    raise
```

trinity/protocol/common/peer.py
```diff
@@ -222,20 +222,11 @@ async def maybe_connect_more_peers(self) -> None:
...
-                ))
-                last_candidates_count = sum(candidate_counts)
-            except asyncio.CancelledError:
-                # no need to log this exception, this is expected
-                raise
-            except Exception:
-                self.logger.exception("unexpected error during peer connection")
-                # Continue trying to connect to peers, even if there was a
-                # surprising failure during one of the attempts.
-                continue
```
carver, line 238: Hm, so we seem to be in a challenging position here, where we want this to fail/raise in the test, but not in the actual execution. I suspect we'll want to be more forgiving during regular execution, eventually, but if you think this is the right thing to do now, no objection. :ok_hand:

carver (approved):
LGTM

trinity/server.py
gsalgado: That's not necessary as the only thing the PeerPool does when running is try to connect to peers

p2p/peer_pool.py
gsalgado: :+1:

trinity/protocol/common/peer.py
gsalgado: My intent was just to move the suppression/logging to `connect_to_node()`, where we can also log the remote to which we were trying to connect, but when merging the actual test fix with my git-stashed logging improvements I accidentally made `connect_to_node()` re-raise. I think for now we want it to always re-raise in tests but during actual execution just log an exception, like we do currently

p2p/peer_pool.py
```diff
@@ -110,10 +110,10 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
     _report_metrics_interval = 15  # for influxdb/grafana reporting
     _peer_boot_timeout = DEFAULT_PEER_BOOT_TIMEOUT
     _event_bus: EndpointAPI = None
-    # If set to True, any exceptions in ALLOWED_PEER_CONNECTION_EXCEPTIONS will be logged and
-    # suppressed. Should only be set to False in tests we we want to ensure a connection is
-    # successful.
-    _suppress_allowed_connection_exceptions: bool = True
+    # If set to True, exceptions raised when connecting to a remote peer will be logged (DEBUG if
+    # it's in ALLOWED_PEER_CONNECTION_EXCEPTIONS or ERROR if not) and suppressed. Should only be
+    # set to False in tests we we want to ensure a connection is successful.
```
carver, line 115:
```suggestion
    # set to False in tests if we want to ensure a connection is successful.
```
?

PR #2058, opened 2020-09-23
Add new block component to propagate new block messages

### LINKED ISSUE #1133: Relay `NewBlock` messages
### What is wrong?

We don't currently relay `NewBlock` messages.  Part of being a good network citizen is realaying these messages.

### How can it be fixed

1. figure out what the right rules are for re-broadcasting these.
2. setup a long running service that listens for this command and does the gossip.

njgheorghita:
### What was wrong?
Fixes #1133 

> Setup a long-running service that listens for this (`NewBlock`) command and does the gossip.

> Figure out what the right rules are for re-broadcasting these.

Rules sourced from [devp2p spec](https://github.com/ethereum/devp2p/blob/master/caps/eth.md)

1. When a `NewBlock` announcement message is received from a peer.
    - [x] Verify the basic validity of the block and check that the POW value is valid
    - [x] Send the block to a small fraction of connected peers (sqroot of total # of peers) using the `NewBlock` message    
        
2. Import block into local chain ...
    - [x] Once block is fully processed - send `NewBlockHashes` to all peers it didn't notify earlier
    - [x] A node should never send a block announcement back to a peer which previously announced the same block. This is usually achieved by remembering a large set of block hashes recently relayed to or from each peer.
    - [ ] The reception of a block announcement may also trigger chain sync if the block
               is not the immediate successor of the client's current latest block
              (it doesn't seem to me that this step is within the scope of this pr)
   

### How was it fixed?
Wrote a `NewBlockComponent` that handles the above tasks. Also introduced a `BroadcastImportedBlock` event to trigger block broadcasts after they've been imported.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/trinity/components/builtin/new_block/component.py
+++ b/trinity/components/builtin/new_block/component.py
@@ -1,0 +1,152 @@
+import math
+import random
+from typing import (
+    Dict,
+    Iterable,
+    List,
+    Tuple,
+)
+
+from async_service import (
+    Service,
+    background_trio_service,
+)
+from eth.abc import BlockHeaderAPI, BlockAPI
+from eth.consensus.pow import check_pow
+from eth_utils import (
+    ValidationError,
+    to_tuple,
+)
+from lahja import EndpointAPI
+import trio
+
+from p2p.abc import SessionAPI
+from trinity.constants import TO_NETWORKING_BROADCAST_CONFIG
+from trinity.extensibility import TrioIsolatedComponent
+from trinity.protocol.eth.events import NewBlockEvent
+from trinity.protocol.eth.payloads import (
+    NewBlockHash,
+    NewBlockPayload,
+)
+from trinity.protocol.eth.peer import (
+    ETHProxyPeerPool,
+    ETHProxyPeer,
+)
+from trinity.sync.common.events import NewBlockImported
+from trinity._utils.logging import get_logger
+
+
+class NewBlockComponent(TrioIsolatedComponent):
+    """
+    Propogate newly received and imported blocks to peers, according to devp2p rules.
+    https://github.com/ethereum/devp2p/blob/master/caps/eth.md#block-propagation
+    """
+    name = "NewBlockComponent"
+
+    @property
+    def is_enabled(self) -> bool:
+        return True
+
+    async def do_run(self, event_bus: EndpointAPI) -> None:
+        proxy_peer_pool = ETHProxyPeerPool(event_bus, TO_NETWORKING_BROADCAST_CONFIG)
+        async with background_trio_service(proxy_peer_pool):
+            service = NewBlockService(event_bus, proxy_peer_pool)
+            async with background_trio_service(service) as manager:
+                await manager.wait_finished()
+
+
+class NewBlockService(Service):
+
+    logger = get_logger('trinity.components.new_block.NewBlockService')
+
+    def __init__(self, event_bus: EndpointAPI, peer_pool: ETHProxyPeerPool) -> None:
+        self._event_bus = event_bus
+        self._peer_pool = peer_pool
+        # tracks which peers have seen a block
+        # todo: old blocks need to be pruned to avoid unbounded growth of tracker
+        self._peer_block_tracker: Dict[bytes, List[str]] = {}
+
+    async def run(self) -> None:
+        self.manager.run_daemon_task(self._handle_imported_blocks)
+
+        async for event in self._event_bus.stream(NewBlockEvent):
+            self.manager.run_task(self._handle_new_block, event.session, event.command.payload)
+
+    async def _handle_imported_blocks(self) -> None:
+        async for event in self._event_bus.stream(NewBlockImported):
+            await self._broadcast_imported_block(event.block)
+
+    async def _handle_new_block(self, sender: SessionAPI, block: NewBlockPayload) -> None:
+        header = block.block.header
+        sender_peer_str = str(ETHProxyPeer.from_session(
+            sender,
+            self._event_bus,
+            TO_NETWORKING_BROADCAST_CONFIG
+        ))
+
+        # Add peer to tracker if we've seen this block before
+        if header.hash in self._peer_block_tracker:
+            if sender_peer_str not in self._peer_block_tracker[header.hash]:
+                self._peer_block_tracker[header.hash].append(sender_peer_str)
+        else:
+            # Verify the validity of block, add to tracker and broadcast to eligible peers
+            try:
+                check_pow(
+                    header.block_number,
+                    header.mining_hash,
+                    header.mix_hash,
+                    header.nonce,
+                    header.difficulty
+                )
+            except ValidationError:
+                self.logger.info("Received invalid block from peer: %s", sender_peer_str)
+            else:
+                self._peer_block_tracker[header.hash] = [sender_peer_str]
+                await self._broadcast_newly_seen_block(header)
+
+    async def _broadcast_imported_block(self, block: BlockAPI) -> None:
+        """
+        Broadcast `NewBlockHashes` for newly imported block to eligible peers,
+        aka those that haven't seen the block.
+        """
+        all_peers = await self._peer_pool.get_peers()
+        if block.hash in self._peer_block_tracker:
+            eligible_peers = self._filter_eligible_peers(all_peers, block.hash)
+        else:
+            self._peer_block_tracker[block.hash] = []
+            eligible_peers = all_peers
+
+        new_block_hash = NewBlockHash(hash=block.hash, number=block.number)
+        for peer in eligible_peers:
+            target_peer = await self._peer_pool.ensure_proxy_peer(peer.session)
+            target_peer.eth_api.send_new_block_hashes((new_block_hash,))
+            self._peer_block_tracker[block.hash].append(str(target_peer))
+            # add checkpoint here to guarantee the event loop is released per iteration
+            await trio.sleep(0)
+
+    async def _broadcast_newly_seen_block(self, header: BlockHeaderAPI) -> None:
+        """
+        Broadcast `NewBlock` for newly received block to square root of
+        total # of connected peers, aka those that haven't seen the block.
+        """
+        all_peers = await self._peer_pool.get_peers()
+        eligible_peers = self._filter_eligible_peers(all_peers, header.hash)
+        number_of_broadcasts = int(math.sqrt(len(all_peers)))
+        sample_size = min(len(eligible_peers), number_of_broadcasts)
+        broadcast_peers = random.sample(eligible_peers, sample_size)
+
+        for peer in broadcast_peers:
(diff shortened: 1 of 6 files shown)


njgheorghita (commented):
Here's an initial pass at part 1. of this issue, any feedback on the general implementation would be great!

> Once block is fully processed - send NewBlockHashes to all peers it didn't notify earlier

In terms of part 2, to assert when a block is considered "fully processed", is this as simple as listening for a `StatelessBlockImportDone` event?

trinity/components/builtin/new_block/component.py
```diff
@@ -0,0 +1,96 @@
...
+        sender_peer_str = str(ETHProxyPeer.from_session(
+            sender,
+            self._event_bus,
+            TO_NETWORKING_BROADCAST_CONFIG
+        ))
+        
+        # Add peer to tracker if we've seen this block before
+        if header.mining_hash in self._peer_block_tracker:
+            # what if the same peer broadcasts duplicate new blocks?
+            if sender_peer_str not in self._peer_block_tracker[header.mining_hash]:
```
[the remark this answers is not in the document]
carver: One lens to use is: could a malicious peer send us the same block over and over again to DoS us? It's not obvious that it would be terrible for this list to get 1000s of elements long with duplicates, so maybe it's not a problem. The presence check is also pretty cheap, though, since the total number of connected peers is also small. So I guess I would keep it.

trinity/components/builtin/new_block/component.py
```diff
@@ -0,0 +1,99 @@
...
+            try:
+                check_pow(
+                    header.block_number,
+                    header.mining_hash,
+                    header.mix_hash,
+                    header.nonce,
+                    header.difficulty
+                )
+            except ValidationError:
+                pass
```
cburgdorf, line 80: It's probably worth to log a message that `sender` has send us an invalid block.

trinity/sync/beam/importer.py
```diff
@@ -491,7 +491,7 @@ def _broadcast_import_complete(
             future.result() if completed else None,
             future.exception() if completed else None,
         ),
-        broadcast_config,
+        # broadcast_config,
```
njgheorghita, line 494: @cburgdorf Could you can help clear up my misunderstanding of what's going wrong here. In order for the `NewBlockComponent` to "hear" a `StatelessBlockImportDone` event, I removed this `broadcast_config` which afaik means that rather than just broadcasting to the networking process, it broadcasts to the entire event bus. This seems not ideal. So I tried a couple different ways to subscribe to the `StatelessBlockImportDone` event, but was unable to make that connection. Could you point me in the right direction / help me understand how to implement the proper communication between the importer and the `NewBlockComponent`? I can also jump on a call if it's easier to describe in person.

cburgdorf: >which afaik means that rather than just broadcasting to the networking process, it broadcasts to the entire event bus.

That is correct but sending it to the networking process only wouldn't help you here anyway because you want to listen to it in your new component process.

But there's a second important aspect to it if we look at the `BroadcastConfig` in detail

https://github.com/ethereum/trinity/blob/a487b20e0917840360e9e582f4460fcee2f62d35/trinity/sync/beam/importer.py#L592

It is generated from an incoming event. That in turn means that it will only be send as a reply to the specific code that originated the event. 

https://github.com/ethereum/trinity/blob/a487b20e0917840360e9e582f4460fcee2f62d35/trinity/sync/beam/chain.py#L766

So even if multiple places within the networking process would listen for it, only the one that send the `DoStatelessBlockImport` would receive this event as a reply.

I think removing the `broadcast_config` entirely is actually the right thing to do here and its not as bad as you may think. The event bus has a mechanism to pass subscriptions across different endpoints which in practice means it won't actually send the event to an endpoint that does not have a single subscription to the event.

Now, there is one important thing that I'm not 100 % sure of and I you should double check that. If you do remove the `broadcast_config`, will it still be a valid answer ending up as a reply at

https://github.com/ethereum/trinity/blob/a487b20e0917840360e9e582f4460fcee2f62d35/trinity/sync/beam/chain.py#L766

I *think* it is but I don't remember exactly and it seems the functionality isn't covered by Lahja tests. If it turns out that this would break the request / response, then I think just broadcasting your own event would be a reasonable workaround.

njgheorghita: > I think it is but I don't remember exactly and it seems the functionality isn't covered by Lahja tests. If it turns out that this would break the request / response, then I think just broadcasting your own event would be a reasonable workaround.

Thanks! This made a lot of sense, it did seem to break the request / response. When using `StatelessBlockImportDone` to trigger the broadcast, beam sync would pause (aka not sync the next block) until the next pivot occurred, indicating to me that `import_done` was never resolved. I went ahead and introduced a new event, `BroadcastImportedBlock` instead.

trinity/components/builtin/new_block/component.py
```diff
@@ -0,0 +1,148 @@
...
+
+class NewBlockService(Service):
+
+    logger = get_logger('trinity.components.new_block.NewBlockService')
+
+    def __init__(self, event_bus: EndpointAPI, peer_pool: ETHProxyPeerPool) -> None:
+        self._event_bus = event_bus
+        self._peer_pool = peer_pool
+        # tracks which peers have seen a block
+        self._peer_block_tracker: Dict[bytes, List[str]] = {}
```
carver, line 65: Looks like this is going to grow forever, so it at least needs a comment & issue noting that this should be resolved. At some point, old enough headers would have to be removed from the object, to avoid unbounded memory growth.

trinity/components/builtin/new_block/component.py
```diff
@@ -0,0 +1,148 @@
...
+            eligible_peers = self._filter_eligible_peers(all_peers, block.hash)
+        else:
+            self._peer_block_tracker[block.hash] = []
+            eligible_peers = all_peers
+
+        new_block_hash = NewBlockHash(hash=block.hash, number=block.number)
+        for peer in eligible_peers:
+            target_peer = await self._peer_pool.ensure_proxy_peer(peer.session)
+            target_peer.eth_api.send_new_block_hashes((new_block_hash,))
+            self._peer_block_tracker[block.hash].append(str(target_peer))
```
carver, line 121: I'm not sure how being in Trio changes things, but I would worry about something like this holding onto the event loop for too long. It depends on the implementation of `ensure_proxy_peer`. If that's not guaranteed to release the loop, then you might want to add trio's version of `await asyncio.sleep(0)`.

carver (approved):
Seems pretty close, GTG after a couple notes.

trinity/sync/common/events.py
```diff
@@ -140,6 +140,17 @@ class StatelessBlockImportDone(BaseEvent):
     exception: BaseException
 
 
+@dataclass
+class BroadcastImportedBlock(BaseEvent):
```
cburgdorf, line 144: Maybe `NewBlockImported` would be a more idiomatic name. The current name is tailored around the consuming component but there's nothing stopping anyone to build other components that use this event for other things.

trinity/sync/common/events.py
```diff
@@ -140,6 +140,17 @@ class StatelessBlockImportDone(BaseEvent):
...
+    """
+    Event that is only emitted after a new block has been successfully imported.
+    This event communicates to the `NewBlockComponent` that a block is ready to
```
cburgdorf, line 147: I would drop the second sentence as this is just one current use case but the use cases for components that listen for freshly imported blocks could be endless.

cburgdorf (approved):

trinity/components/builtin/new_block/component.py
njgheorghita: > If that's not guaranteed to release the loop...

It's my (limited) understanding that this is not much of a concern in trio, since every `await` is a checkpoint (which checks to see if it's time to switch to another task).... but maybe @cburgdorf can help confirm/deny this?

cburgdorf: I honestly can't tell. @gsalgado is the one who understands all the nitty gritty details of `asyncio` and `trio`.

gsalgado: According to https://trio.readthedocs.io/en/stable/reference-core.html#checkpoints, trio currently always switches to a new task at every checkpoint (e.g. an `await something(...)` call), so right now there's no risk of us running all iterations of this for-loop without releasing the event loop, regardless of how `ensure_proxy_peer()` is implemented. If `ensure_proxy_peer()` were actually a sync method (albeit having an async signature) and trio's scheduler behavior changed, then we could end up running all iterations of this for-loop without releasing the event loop, which I believe is what @carver was worried about? I'd say we just need to make sure `ensure_proxy_peer()` is not a sync method (i.e. it has one or more trio checkpoints, which is something any method with an async signature should have anyway)

carver: Yup, I see from the linked docs that trio's goal is to *always* release the event loop in every trio-level async function. This is different from asyncio. For example, getting an item from a queue will not release the loop if there is an item ready. So it requires less investigation to make sure `ensure_proxy_peer` is okay here. As long as every code path inside the method eventually calls into some base-level trio async function, then we should be good.

njgheorghita: Thanks for the explanation @gsalgado! Just to help clear my understanding, while `ensure_proxy_peer` does have the `async` signature and contains a trio checkpoint, that checkpoint is only touched if the inclusion check fails...

https://github.com/ethereum/trinity/blob/65bf6acaf510cc6cf1600a793c2be9c3cacfdae2/trinity/protocol/common/peer_pool_event_bus.py#L347-L359

> As long as every code path inside the method eventually calls into some base-level trio async function..

So, we could have the case where iterate many times without releasing the event loop, but it seems like in this case it's not a concern since the function will return the `ProxyPeer` immediately if it exists inside `connected_peers`?

gsalgado: > So, we could have the case where we iterate many times without releasing the event loop, but it seems like in this case it's not a concern since the function will return the ProxyPeer immediately if it exists inside connected_peers?

I think @carver's concern was not with the performance of `ensure_proxy_peer()` itself, but with that of the other calls performed on every iteration of the loop. Currently, the fact that `ensure_proxy_peer()` may return immediately (without any trio checkpoints) is not a problem given how the trio scheduler works, but as their docs said that may change. I was going to suggest adding an `else:` block to `ensure_proxy_peer()` with a `trio.checkpoint()` call, but that could cause the return statement to crash if the peer was removed from `self.connected_peers()` by another coroutine, so maybe the best option is to add the checkpoint at the beginning of the method, with a comment explaining why we do so

gsalgado: Or maybe add a checkpoint at the end of every iteration of the for loop where this is called

PR #2060, opened 2020-09-23
Asyncify metrics reporting to influx

### LINKED ISSUE #2048: Report metrics to Influx asynchronously
* Trinity version: x.x.x
* OS: osx/linux/win
* Python version (output of `python --version`):
* Environment (output of `pip freeze`):

### What is wrong?
Currently the `pyformance.reporters.InfluxReporter` uses the `requests` module, which is synchronous, to push metrics data to InfluxDB. We want to replace this with an asynchronous request using `aiohttp`. 

### How can it be fixed
Extend `pyformance.reporters.InfluxReporter` to use an async call with `aiohttp`.

njgheorghita:
### What was wrong?
Fixes #2048

Posting metrics to influxdb via the `pyformance.InfluxReporter` was done synchronously, and would be better if it was async.

### How was it fixed?
Due to compatibility issues between `trio` and `aiohttp`, I had to introduce the [`asks`](https://asks.readthedocs.io/en/latest/) library in order to execute calls to influxdb asynchronously.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/trinity/components/builtin/metrics/service/asyncio.py
+++ b/trinity/components/builtin/metrics/service/asyncio.py
@@ -7,5 +7,5 @@ class AsyncioMetricsService(BaseMetricsService):
 
     async def continuously_report(self) -> None:
         while self.manager.is_running:
-            super().report_now()
+            await self.report_now()
             await asyncio.sleep(self._reporting_frequency)
--- a/trinity/components/builtin/metrics/service/base.py
+++ b/trinity/components/builtin/metrics/service/base.py
@@ -1,11 +1,16 @@
-import time
-import aiohttp
-from typing import Dict
-import base64
 from abc import abstractmethod
-from urllib import parse
+import base64
+import functools
 from http.client import HTTPException
+import time
+from typing import (
+    Callable,
+    Dict,
+    TypeVar,
+)
+from urllib import parse
 
+from asks import Session
 from async_service import Service
 from pyformance.reporters.influx import InfluxReporter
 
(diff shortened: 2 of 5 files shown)


trinity/components/builtin/metrics/service/asyncio.py
```diff
@@ -7,5 +7,5 @@ class AsyncioMetricsService(BaseMetricsService):
 
     async def continuously_report(self) -> None:
         while self.manager.is_running:
-            super().report_now()
+            await super().report_now()
```
pipermerriam, line 10: why is the `super()` call needed here?

trinity/components/builtin/metrics/service/base.py
```diff
@@ -19,22 +19,51 @@ class ExtendedInfluxReporter(InfluxReporter):
     ``InfluxReporter`` extended to enable sending annotations to InfluxDB
     """
     async def send_annotation(self, annotation_data: str) -> None:
+        url = self._get_post_url()
+        await self._post(url, annotation_data)
+
+    def _generate_auth_header(self) -> Dict[str, str]:
+        auth_string = ("%s:%s" % (self.username, self.password)).encode()
+        auth = base64.b64encode(auth_string)
+        return {"Authorization": "Basic %s" % auth.decode("utf-8")}
```
pipermerriam, line 28: f-string?

trinity/components/builtin/metrics/service/base.py
```diff
@@ -19,22 +19,51 @@ class ExtendedInfluxReporter(InfluxReporter):
...
-        auth_string = ("%s:%s" % (self.username, self.password)).encode()
-        auth = base64.b64encode(auth_string)
-        return {"Authorization": "Basic %s" % auth.decode("utf-8")}
+        return url
+
+    # async implementation of InfluxReporter.report_now
+    async def report_async(self,
+                           registry: HostMetricsRegistry = None,
+                           timestamp: int = None) -> None:
+        timestamp = timestamp or int(round(self.clock.time()))
```
pipermerriam, line 49: Just noting that if `timestamp == 0` this will produce a new timestamp which is probably not the desired behavior.

trinity/components/builtin/metrics/service/base.py
```diff
@@ -19,22 +19,51 @@ class ExtendedInfluxReporter(InfluxReporter):
...
+        metrics = (registry or self.registry).dump_metrics()
+        post_data = []
+        for key, metric_values in metrics.items():
+            if not self.prefix:
+                table = key
+            else:
+                table = f"{self.prefix}.{key}" % (self.prefix, key)
```
pipermerriam, line 56: this is a confusing combination of `%` formatting and f-string.  What is going on here?

trinity/components/builtin/metrics/service/base.py
```diff
@@ -19,22 +19,51 @@ class ExtendedInfluxReporter(InfluxReporter):
...
+    def _get_post_url(self) -> str:
         url = parse.urlunparse((
             self.protocol,  # scheme
             f"{self.server}:{self.port}",  # netloc
             "/write",  # path
-            '',
+            '',  # params
             f"db={self.database}&precision=s",  # query
-            ''
+            '',  # fragment
```
pipermerriam, line 41: You could use [`urllib.parse.ParseResult`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.ParseResult) which is a built-in named-tuple type which would allow you to document this using keyword arguments rather than comments.

trinity/components/builtin/metrics/service/base.py
njgheorghita: Yeah this whole method is gross, I pulled it straight from the [`pyformance.InfluxReporter` implementation](https://github.com/omergertel/pyformance/blob/master/pyformance/reporters/influx.py#L77) but I'll clean this method up.

njgheorghita: Ahh, but this particular instance was my mistake of combining the two.

trinity/components/builtin/metrics/service/base.py
```diff
@@ -19,22 +20,51 @@ class ExtendedInfluxReporter(InfluxReporter):
...
     def _generate_auth_header(self) -> Dict[str, str]:
         auth_string = ("%s:%s" % (self.username, self.password)).encode()
         auth = base64.b64encode(auth_string)
-        return {"Authorization": "Basic %s" % auth.decode("utf-8")}
+        return {"Authorization": f"Basic {auth.decode('utf-8')}"}
+
+    async def _post(self, data: str) -> None:
+        url = self._get_post_url()
+        auth_header = self._generate_auth_header()
+        s = Session(url, headers=auth_header)
```
cburgdorf, line 33: We try to avoid single letter variables. This could be `session`

trinity/components/builtin/metrics/service/base.py
```diff
@@ -87,10 +117,10 @@ async def run(self) -> None:
         self.manager.run_daemon_task(self.continuously_report)
         await self.manager.wait_finished()
 
-    def report_now(self) -> None:
+    async def report_now(self) -> None:
         try:
-            self._reporter.report_now()
-        except (HTTPException, ConnectionError) as exc:
+            await self._reporter.report_async()
+        except (HTTPException, ConnectionError, socket.gaierror) as exc:
```
cburgdorf, line 123: https://docs.python.org/3/library/socket.html#socket.gaierror suggests that this would not be some kind of permanent error. Do I speculate correctly that this is coming from this test:

https://github.com/ethereum/trinity/blob/0c7089e71d6a33fca3b1ea5aabe8d221f6a3e2ce/tests/integration/test_trinity_cli.py#L279

If that's the case then I think we need to change the URL in that test to something that is actually legal but will still fail / timeout. Because all the test tries to prove is that a temporarily unavailable metrics endpoint won't take Trinity down.

I think if someone configures an entirely illegal domain for the metrics then we *do want*  to reject that on startup. And frankly, when I wrote that test it didn't occur to me that ` _.com` may in fact be an illegal URL.

cburgdorf (approved):
Looks good :+1: Left two comments inline but just nitpicks.

trinity/components/builtin/metrics/service/base.py
pipermerriam, line 33: I'd recommend caching and re-using this session object.

pipermerriam (approved):

trinity/components/builtin/metrics/service/base.py
njgheorghita: ahhh, this isn't the first time you've had to remind me 🤦 sooner or later i'll quit it with these single letter variables

trinity/components/builtin/metrics/service/base.py
njgheorghita: eagle eye! that's exactly it

Issue #2053, opened 2020-09-21
Core testsuite failing on CI with "Too long with no output (exceeded 10m0s)" error


gsalgado:
Been seeing those quite frequently recently: https://app.circleci.com/pipelines/github/ethereum/trinity/7501/workflows/b572d9b4-ba28-4717-acad-44d35e7402ea/jobs/285262 (both on py37-core and py38-core)

gsalgado:
`tests/core/p2p-proto/test_sync.py::test_beam_syncer_loads_recent_state_root` is the one that is sometimes hanging: https://app.circleci.com/pipelines/github/ethereum/trinity/7535/workflows/2e7d23be-ca81-4b2a-bc9d-175e63ac6e2f/jobs/286232. 

From the above failure I thought it was hanging when running with `True-2` as parameters, but locally I can reproduce it consistently (running in a loop) with `True-7`

gsalgado:
And it seems to hang because `async-service` fails to cancel all `BeamDownloader` tasks: https://gist.github.com/gsalgado/f54d9833f84aef66576920c65690211a#file-gistfile1-txt-L2339

gsalgado:
https://github.com/ethereum/async-service/pull/79 has a fix that will stop that test from hanging

gsalgado:
Seeing this again: https://app.circleci.com/pipelines/github/ethereum/trinity/7726/workflows/e64bdc48-dc22-496c-88e7-0d0e8795be31/jobs/291982

gsalgado:
And again... https://app.circleci.com/pipelines/github/ethereum/trinity/7763/workflows/b902e1f2-2f77-4b12-9f0b-4f372025ceae/jobs/293064

Now I can't seem to reproduce it locally, though :-(

PR #2065, opened 2020-09-28
Update async-service

gsalgado:
Closes: #2053


carver:
I updated async-service so it uses towncrier like the other projects now. Just a reminder for next time to include a news fragment and run `make notes` before the release.

carver (approved):

PR #2066, opened 2020-09-28
Fix QueeningQueue to not manually raise CancelledError

gsalgado:
QueeningQueue.pop_fastest_peasant() would raise a CancelledError when the
service finished, but that method runs in a task created by QueeningQueue's
parent service (e.g. BeamDownloader), so that CancelledError would cause
the parent service to terminate as well

This was triggering a bug in async-service that was ultimately causing our
test suite to hang: https://github.com/ethereum/trinity/issues/2053

### DIFF
--- a/trinity/sync/beam/state.py
+++ b/trinity/sync/beam/state.py
@@ -538,7 +538,9 @@ class BeamDownloader(Service, PeerSubscriber):
         for them. Without waiting for a response from the peer, continue and check if more
         predictive trie nodes are requested. Repeat indefinitely.
         """
-        while self.manager.is_running:
+        # If self._queen_tracker terminates we need to exit as well, so check that on every
+        # iteration.
+        while self.manager.is_running and self._queen_tracker.get_manager().is_running:
             try:
                 batch_id, hashes = await asyncio.wait_for(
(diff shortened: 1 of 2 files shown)


trinity/sync/beam/state.py
```diff
@@ -593,11 +594,6 @@ async def _match_predictive_node_requests_to_peers(self) -> None:
                 cancel_attempt = True
             else:
                 if peer.eth_api.get_node_data.is_requesting:
-                    self.logger.debug(
-                        "Want predictive nodes from %s, but it has an active request, skipping...",
-                        peer,
-                    )
-                    self._queen_tracker.insert_peer(peer, NON_IDEAL_RESPONSE_PENALTY)
```
gsalgado, line 600: I've removed this because that's already done in `pop_fastest_peasant()`

carver: I think this should stay in. When  `pop_fastest_peasant()` reinserts the peer, it doesn't return the peer. Since the peer has been returned in this code path, we need to reinsert it before cancelling the request attempt.

carver (approved):

Issue #2068, opened 2020-09-30
Crash in metrics reporting


pipermerriam:
### What is wrong?

My node crashed today with this.

```
Sep 30 21:30:43 localhost launch-trinity.sh[17611]: <Manager[BeamSyncService] flags=SRcfe>: task run[daemon=False] exited with error: [Errno -2] Name or service not known
Sep 30 21:30:43 localhost launch-trinity.sh[17611]: Traceback (most recent call last):
Sep 30 21:30:43 localhost launch-trinity.sh[17611]:   File "/root/env/lib/python3.8/site-packages/async_service/base.py", line 324, in _run_and_manage_task
Sep 30 21:30:43 localhost launch-trinity.sh[17611]:     await task.run()
Sep 30 21:30:43 localhost launch-trinity.sh[17611]:   File "/root/env/lib/python3.8/site-packages/async_service/asyncio.py", line 35, in run
Sep 30 21:30:43 localhost launch-trinity.sh[17611]:     await self._async_fn(*self._async_fn_args)
Sep 30 21:30:43 localhost launch-trinity.sh[17611]:   File "/root/env/lib/python3.8/site-packages/trinity/sync/beam/service.py", line 60, in run
Sep 30 21:30:43 localhost launch-trinity.sh[17611]:     await self._pivot_loop()
Sep 30 21:30:43 localhost launch-trinity.sh[17611]:   File "/root/env/lib/python3.8/site-packages/trinity/sync/beam/service.py", line 80, in _pivot_loop
Sep 30 21:30:43 localhost launch-trinity.sh[17611]:     await self.sync_metrics_registry.record_pivot(latest_block)
Sep 30 21:30:43 localhost launch-trinity.sh[17611]:   File "/root/env/lib/python3.8/site-packages/trinity/components/builtin/metrics/sync_metrics_registry.py", line 24, in record_pivot
Sep 30 21:30:43 localhost launch-trinity.sh[17611]:     await self.metrics_service.send_annotation(post_data)
Sep 30 21:30:43 localhost launch-trinity.sh[17611]:   File "/root/env/lib/python3.8/site-packages/trinity/components/builtin/metrics/service/base.py", line 130, in send_annotation
Sep 30 21:30:43 localhost launch-trinity.sh[17611]:     await self._reporter.send_annotation(annotation_data)
[...]

PR #2069, opened 2020-10-01
Use aiohttp for reporting metrics annotations

njgheorghita:
### What was wrong?
Fixes https://github.com/ethereum/trinity/issues/2068

### How was it fixed?
The `asks` library is incompatible within an asyncio service, so it was updated to use `aiohttp`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/trinity/components/builtin/metrics/service/asyncio.py
+++ b/trinity/components/builtin/metrics/service/asyncio.py
@@ -1,9 +1,16 @@
+import aiohttp
 import asyncio
 
 from trinity.components.builtin.metrics.service.base import BaseMetricsService
 
 
 class AsyncioMetricsService(BaseMetricsService):
+    async def async_post(self, data: str) -> None:
+        # use asyncio-compatible aiohttp for async http calls
+        url = self.reporter._get_post_url()
+        auth_header = self.reporter._generate_auth_header()
+        async with aiohttp.ClientSession() as session:
+            await session.post(url, data=data, headers=auth_header)
 
     async def continuously_report(self) -> None:
         while self.manager.is_running:
--- a/trinity/components/builtin/metrics/service/base.py
+++ b/trinity/components/builtin/metrics/service/base.py
@@ -1,16 +1,14 @@
 from abc import abstractmethod
 import base64
-import functools
 from http.client import HTTPException
 import time
 from typing import (
+    Awaitable,
     Callable,
     Dict,
-    TypeVar,
 )
 from urllib import parse
 
-from asks import Session
 from async_service import Service
 from pyformance.reporters.influx import InfluxReporter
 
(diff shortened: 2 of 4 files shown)


trinity/components/builtin/metrics/service/base.py
```diff
@@ -127,7 +133,10 @@ def registry(self) -> HostMetricsRegistry:
         return self._registry
 
     async def send_annotation(self, annotation_data: str) -> None:
-        await self._reporter.send_annotation(annotation_data)
+        try:
+            await self._reporter.send_annotation(annotation_data)
+        except (HTTPException, ConnectionError) as exc:
```
njgheorghita, line 138: The bug in #2068 is fixed with the change in `ExtendedInfluxReporter` above, but this doesn't protect against other bugs that might cause crashes later on when sending annotations. Since annotations are only reported occasionally (every time we pivot) it seems to me acceptable that we just log the exception, as opposed to the time based caching strategy used in `report_now`.

trinity/components/builtin/metrics/service/base.py
```diff
@@ -81,7 +84,8 @@ async def report_async(self,
             line = f"{table} {values} {timestamp}"
             post_data.append(line)
         data = "\n".join(post_data)
-        await self._post(data)
+        # metrics component is trio-based, which requires the asks library
+        await self.asks_session.post(data=data)
```
cburgdorf, line 88: The metrics service comes in two flavors to be able to run in trio based components (such as the metrics component itself) and asyncio based components such as the peer pool / sync. 

As the code stands, we have one method `report_async` which assumes the class is running in a `trio` environment and then we have `send_annotation` in the same class which assumes it is called in an `asyncio` environment.

I think what we need to do is to ensure that we have two versions of the reporter for both flavors. I mean, two different classes just like we have the `TrioMetricsService` and `AsyncioMetricsService`.Then `report_async` can collect all the data and then call something like `post_async` which ends up calling the trio/asyncio based method depending on the environment it is running in.

I'm :-1: on mixing having two methods on the same class where one expects to be invoked in a trio environment and the other expects to be invoked in an asyncio environment.

njgheorghita: Cool, that makes sense to me!

njgheorghita:
ping @cburgdorf for another :eyes: when you get the chance

trinity/components/builtin/metrics/service/base.py
```diff
@@ -83,6 +68,49 @@ async def report_async(self,
...
+    @property  # type: ignore
+    @cache
+    def session(self) -> Session:
+        url = self._get_post_url()
+        auth_header = self._generate_auth_header()
+        return Session(url, headers=auth_header)
+
+    async def _post(self, data: str) -> None:
+        await self.session.post(data=data)
+
```
cburgdorf, line 113: I think this is a case where "composition over inheritance" would make the code a little stronger.
Basically we don't need `AsyncioInfluxReporter` and `TrioInfluxReporter` if we change `ExtendedInfluxReporter` to accept one more argument of type `Callable[[str], Awaitable[None]]` which will be the `async` `_post` function `trio` or `asyncio` based depending on the circumstances.

Inside the `BaseMetricsService` this could lookup the method like this:

```python
self._reporter = ExtendedInfluxReporter(
            registry=self._registry,
            database=influx_database,
            username=influx_user,
            password=influx_password,
            protocol=protocol,
            port=port,
            server=influx_server,
            # This would mean we define it on the existing TrioMetricsService / AsyncioMetricsService
            # And put an abstract version on the BaseMetricsService
            async_post=self.async_post
        )
```

trinity/components/builtin/metrics/service/asyncio.py
```diff
@@ -1,9 +1,32 @@
...
+            influx_server,
+            influx_user,
+            influx_password,
+            influx_database,
+            host,
+            port,
+            protocol,
+            reporting_frequency,
+            AsyncioInfluxReporter,
+        )
```
cburgdorf, line 29: If we follow https://github.com/ethereum/trinity/pull/2069/files#r505418099 this change could be reverted.

trinity/components/builtin/metrics/service/trio.py
```diff
@@ -1,9 +1,32 @@
...
+            influx_server,
+            influx_user,
+            influx_password,
+            influx_database,
+            host,
+            port,
+            protocol,
+            reporting_frequency,
+            TrioInfluxReporter,
+        )
```
cburgdorf, line 29: If we follow https://github.com/ethereum/trinity/pull/2069/files#r505418099 this change can be reverted

cburgdorf (approved):
All in all looks good. I think it can be rewritten a little leaner with the "composition over inheritance" mindset. I know I actually recommended creating two classes `AsyncioInfluxReporter` and `TrioInfluxReporter` but now that I see it and that it really only boils down to one `async_post` function in essence I think this could be refactored further.

njgheorghita:
@cburgdorf thanks for the suggestion! I've made the update to a composition approach. If you can give it just one more :eyes: then I think we'll be good to go here. It also seems like we need a `py-evm` release (with the new `rlp` version) to happen for the tests to pass?

trinity/sync/beam/constants.py
```diff
@@ -71,7 +71,7 @@
 
 # About how long after a block can we request trie data from peers?
 # The value is configurable by client, but tends to be around 120 blocks.
-ESTIMATED_BEAMABLE_BLOCKS = 120
+ESTIMATED_BEAMABLE_BLOCKS = 1
```
cburgdorf, line 75: @njgheorghita guess this slipped through from testing

cburgdorf (approved):
LGTM except for the change of `ESTIMATED_BEAMABLE_BLOCKS` which looks like it slipped through from testing.

Also, the RLP mess needs to be settled first. I haven't looked into it yet but it wasn't on my radar this would break Trinity CI :see_no_evil: 

I'm inclined to cut a Py-EVM release today even if its Friday (policy forbids Friday releases) because I'll be out next week. Unless someone else volunteers to cut the release on Monday. I'll try to resolve that later in chat when 'murica wakes up.

trinity/sync/beam/constants.py
njgheorghita: ahhh soo!

PR #2073, opened 2020-10-08
Install RLP rust backend explicitly

cburgdorf:
### What was wrong?

`rlp 2.0.0a1` came with a fix dependency against a binary library written in Rust. Unfortunately, the supported binaries leave some platforms unsupported (all 32 bit, ARM etc). A new version `rlp 2.0.0a2` addresses that by making the rust backend entirely optional. That in turn means it has to be enabled explicitly on an application level (e.g. Trinity).

### How was it fixed?

Added dependency against `"rlp[rust-backend]>=2.0.0-alpha.2, <3"`

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,8 @@ deps = {
         "pyformance==0.4",
         # requests 2.21 is required to support idna 2.8 which is required elsewhere
         "requests>=2.21,<3",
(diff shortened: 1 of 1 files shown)


cburgdorf:
This is still depending on a cascade of other releases (see https://github.com/ethereum/py-trie/pull/119)

Issue #1912, opened 2020-07-20
Implement eth_chainId


carver:
### What is wrong?

The rpc endpoint `eth_chainId` is missing.

### How can it be fixed

Add the `chainId` endpoint in `trinity/rpc/modules/eth.py`.

PR #2076, opened 2020-10-12
Add eth_chainId rpc endpoint

njgheorghita:
### What was wrong?
Adds support for the `eth_chainId` rpc endpoint.
Fixes https://github.com/ethereum/trinity/issues/1912

### How was it fixed?
It now reports the chain id.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/trinity/rpc/modules/eth.py
+++ b/trinity/rpc/modules/eth.py
@@ -30,6 +30,7 @@ from eth_utils import (
     encode_hex,
     int_to_big_endian,
     is_integer,
+    to_hex,
     to_wei,
     ValidationError,
 )
@@ -172,6 +173,10 @@ class Eth(Eth1ChainRPCModule):
     async def blockNumber(self) -> str:
         num = self.chain.get_canonical_head().block_number
         return hex(num)
+
+    async def chainId(self) -> str:
+        chain_id = self.chain.chain_id
+        return to_hex(chain_id)
 
     @format_params(identity, to_int_if_hex)
     @retryable(which_block_arg_name='at_block')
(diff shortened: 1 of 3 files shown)


trinity/rpc/modules/eth.py
```diff
@@ -173,6 +175,13 @@ async def blockNumber(self) -> str:
         num = self.chain.get_canonical_head().block_number
         return hex(num)
 
+    async def chainId(self) -> str:
+        response = await self.event_bus.request(
+            NetworkIdRequest(),
+            TO_NETWORKING_BROADCAST_CONFIG
+        )
+        return to_hex(text=str(response.network_id))
+
```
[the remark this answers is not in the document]
njgheorghita: Ahh, good point. I was operating under the assumption that since `mainnet` / `goerli` / `ropsten` are the only supported chains (and it's unlikely trinity will support etc) that it won't make a difference, but I'll make the update to a more robust chain id endpoint.

trinity/rpc/modules/eth.py
```diff
@@ -173,6 +174,10 @@ async def blockNumber(self) -> str:
         num = self.chain.get_canonical_head().block_number
         return hex(num)
 
+    async def chainId(self) -> str:
+        chain_id = self.chain.chain_id
+        return to_hex(text=str(chain_id))
```
carver, line 179: If `chain_id` is an int here, then I think we can skip some conversion steps here with:
```suggestion
        return to_hex(chain_id)
```

njgheorghita: Ahh, yup I completely forgot that `int`s were also accepted as a default arg in these conversion functions.

carver (approved):

Issue #1905, opened 2020-07-16
Add admin_removePeer JSON-RPC API


cburgdorf:
### What is wrong?

It's valuable to be able to remove peers via the JSON-RPC admin API.

### How can it be fixed

Lazy hint: Very similar to how `admin_addPeer` is implemented

PR #2077, opened 2020-10-12
Support admin_removePeer RPC endpoint

njgheorghita:
### What was wrong?
JSONRPC support for `admin_removePeer`

Fixes https://github.com/ethereum/trinity/issues/1905

[besu implementation](https://besu.hyperledger.org/en/1.2.1/Reference/Pantheon-API-Methods/#admin_removepeer)
[geth implementation](https://github.com/ethereum/go-ethereum/blob/c0c01612e95dd20f125154646b38283ab780f357/node/api.go#L81)

### How was it fixed?
Added `admin_removePeer` RPC endpoint.
Returns `True` if peer was found in peer pool and an attempt to disconnect was made, `False` otherwise.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/p2p/peer_pool.py
+++ b/p2p/peer_pool.py
@@ -284,6 +284,16 @@ class BasePeerPool(Service, AsyncIterable[BasePeer]):
             peer.connection.get_manager().wait_started(), timeout=PEER_READY_TIMEOUT)
         await peer.connection.run_peer(peer)
 
+    async def disconnect_from_peer(
+        self, peer_session: SessionAPI, reason: DisconnectReason
+    ) -> None:
+        if peer_session in self.connected_nodes:
+            peer = self.connected_nodes[peer_session]
+            self.logger.info("Disconnecting from %s.", peer)
+            await peer.disconnect(reason)
+        else:
+            self.logger.info("Unable to disconnect from %s, since peer is not connected to.", peer)
+
     async def add_inbound_peer(self, peer: BasePeer) -> None:
         try:
             await self._start_peer(peer)
--- a/tests/core/json-rpc/test_ipc.py
+++ b/tests/core/json-rpc/test_ipc.py
@@ -25,6 +25,7 @@ from trinity.nodes.events import (
 )
 from trinity.protocol.common.events import (
     ConnectToNodeCommand,
+    DisconnectFromPeerCommand,
     PeerCountRequest,
     PeerCountResponse,
 )
@@ -91,9 +92,6 @@ async def get_ipc_response(
         request_msg,
         event_loop,
         event_bus):
-
-    # Give event subsriptions a moment to propagate.
-    await asyncio.sleep(0.01)
 
     assert wait_for(jsonrpc_ipc_pipe_path), "IPC server did not successfully start with IPC file"
 
@@ -654,6 +652,14 @@ GOOD_KEY = (
             },
         ),
         (
+            build_request('admin_removePeer', ['enode://none@[::]:30303']),
+            {
+                'jsonrpc': '2.0',
+                'id': 3, 'error':
+                'public key must be a 128-character hex string'
+            },
+        ),
+        (
             build_request('admin_addPeer', [f'enode://{GOOD_KEY}@[::]:30303']),
             {
                 'jsonrpc': '2.0',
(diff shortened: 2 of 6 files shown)


tests/core/json-rpc/test_ipc.py
```diff
@@ -707,6 +729,59 @@ async def test_admin_addPeer_fires_message(
...
+
+            result = await get_ipc_response(
+                jsonrpc_ipc_pipe_path,
+                request,
+                event_loop,
+                event_bus
+            )
+            assert result == {'id': 3, 'jsonrpc': '2.0', 'result': True}
+
+            event = await asyncio.wait_for(future, timeout=0.1, loop=event_loop)
```
gsalgado, line 757: Why do you need to use `ensure_future()` and then wait for it here? Can't you instead call `event = await asyncio.wait_for(bus.wait_for(...), timeout=...)` here directly?

trinity/protocol/common/peer_pool_event_bus.py
```diff
@@ -163,6 +165,14 @@ async def handle_connect_to_node_requests(self) -> None:
             self.logger.debug('Received request to connect to %s', command.remote)
             self.manager.run_task(self.peer_pool.connect_to_node, command.remote)
 
+    async def handle_disconnect_from_peer_requests(self) -> None:
+        async for command in self.event_bus.stream(DisconnectFromPeerCommand):
+            self.logger.warning(
+                'Received request to disconnect to %s. Reason: %s',
```
gsalgado, line 171: s/disconnect to/disconnect from?

trinity/rpc/modules/admin.py
```diff
@@ -56,6 +58,23 @@ async def addPeer(self, uri: str) -> None:
             TO_NETWORKING_BROADCAST_CONFIG
         )
 
+    async def removePeer(self, uri: str) -> bool:
+        validate_enode_uri(uri, require_ip=True)
+        peer_to_remove = Node.from_uri(uri)
+
+        response = await self.event_bus.request(GetConnectedPeersRequest())
+        if response.peers:
```
gsalgado, line 66: What will `response.peers` be if there are no connected peers? If it's None, it'd be better to check for that explicitly, otherwise this check is redundant as the iteration below will be a no-op in case it's an empty sequence

trinity/protocol/common/peer_pool_event_bus.py
gsalgado: Also, do we really want this to be a warning?

p2p/peer_pool.py
```diff
@@ -284,6 +284,14 @@ async def _start_peer(self, peer: BasePeer) -> None:
             peer.connection.get_manager().wait_started(), timeout=PEER_READY_TIMEOUT)
         await peer.connection.run_peer(peer)
 
+    async def disconnect_from_peer(
+        self, peer_session: SessionAPI, reason: DisconnectReason
+    ) -> None:
```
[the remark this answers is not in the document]
njgheorghita: Line length comes in at `101` characters, but if you have a shorter function name in mind, I can update it.

trinity/rpc/modules/admin.py
njgheorghita: Good catch - it returns an empty tuple.

tests/core/json-rpc/test_ipc.py
njgheorghita: It seems like this is not possible, and results in a timeout even when a large timeout is set. The design for this test was borrowed heavily from a similar test for [admin_addPeer](https://github.com/ethereum/trinity/blob/master/tests/core/json-rpc/test_ipc.py#L685)

p2p/peer_pool.py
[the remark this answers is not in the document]
gsalgado: No need to, it's fine like this

tests/core/json-rpc/test_ipc.py
gsalgado: I'm weary of those things we have to do for no obvious reason, so I'd like to understand why it's needed so that we can either write it in a better way or at least add a comment explaining why it's needed. I believe this is needed because we need some coroutine to be waiting for the event when `get_ipc_response()` is called (as it will cause the event to be sent, and if nobody is listening it will be lost?), but using `ensure_future()` doesn't guarantee the coroutine will be scheduled before the event is sent so in theory the test could fail. 

That seems to be what's going on, as I just noticed this hack to ensure the event subscribers get a chance to run before `get_ipc_response(): https://github.com/ethereum/trinity/blob/3f2a091c526912bbf7f24f8e69370c197332e777/tests/core/json-rpc/test_ipc.py#L96

We should be able to get rid of that hack if we add a `await event_bus.wait_until_any_endpoint_subscribed_to(<event>)` immediately after the `ensure_future()` here and in the addPeer test. Can you give that a try?
I can't think of a way to write that

njgheorghita: sure! thanks for the feedback, I'll give it a shot tomorrow

njgheorghita: I've updated the tests - and everything is as expected locally (circleci tests are failing right now b/c of a version conflict with `rlp` library). Are the tests now in line with what you had in mind?

gsalgado (approved):

Issue #1520, opened 2020-01-31
`cached_property` results in `Any` types


pipermerriam:
### What is wrong?

The `cached_property` decorator results in return values having `Any` 

### How can it be fixed

Fix is present in https://github.com/ethereum/trinity/pull/1519 and needs to be extracted into a stand-alone PR

Elnaril:
Hello @pipermerriam 
I am happy to give a try to this issue if it's fine for you ?

As per my current understanding, this would mean:

1. Apply commit a177bef89b7c89853d3cb1f86175c6865523a9ef (so your modified mypy.ini file from the PR #1519)
2. For each of the  "Returning Any from function declared to return _Some_Type_" error ( I found 57 of them), apply one of the 3 fixes you mentioned:
    - wrap return values in an extra level of type casting
    - use cast to force the types...
    - do a type: ignore

Is it correct ?

pipermerriam:
@Elnaril yes, that seems like the correct approach.  Feel free to open up a PR and we'll take a look.

Elnaril:
@pipermerriam Thank you for your feedback. I'll create a PR asap with the new mypy.ini and start from there.

PR #2078, opened 2020-10-12
Issue #1520 - `cached_property` results in `Any` types

Elnaril:
### What was wrong?
The cached_property decorator results in return values having Any type.
(See issue #1520)

### How was it fixed?
Applying commit a177bef89b7c89853d3cb1f86175c6865523a9ef which is an update of the mypy.ini file making mypy to warn when the type Any is returned whereas the function declares returning another type.
Then fix any error found by mypy.

### To-Do

- [x] Apply commit a177bef89b7c89853d3cb1f86175c6865523a9ef updating mypy.ini
- [x] Fix mypy errors

#### Cute Animal Picture


Elnaril:
Initial number of errors found by mypy: 57

image
image

Elnaril:
For a start, I fixed 2 errors with 'type: ignore'.
I chose this method as it seems to me to be the less invasive, code-wise and performance-wise (though typing.cast() is not that far behind I suppose), since we're heading anyway with a dirty solution.
@pipermerriam , Is it fine for you ?

Elnaril:
Thanks ! I'll carry on with the remaining errors then.

Elnaril:
Trinity package is done.
I'm looking now into the p2p package

Elnaril:
I guess this failure is not directly related to my release note.
Is it possible to re-trigger py37-rpc-state-istanbul ?

Elnaril:
@pipermerriam : All checks have passed and I added a release note.

p2p/commands.py
```diff
@@ -72,10 +76,12 @@ def decode(self, data: bytes) -> TCommandPayload:
...
 class SnappyCodec(CompressionCodecAPI):
     def compress(self, data: bytes) -> bytes:
-        return snappy.compress(data)
+        # type ignored to fix https://github.com/ethereum/trinity/issues/1520
+        return snappy.compress(data)  # type: ignore
 
     def decompress(self, data: bytes) -> bytes:
-        return snappy.decompress(data)
+        # type ignored to fix https://github.com/ethereum/trinity/issues/1520
+        return snappy.decompress(data)  # type: ignore
```
pipermerriam, line 84: I *think* this one could be fixed by adding a `./stubs/snappy.pyi` file with the following content:

```python
def compress(data: bytes) -> bytes:
    ...
def decompress(data: bytes) -> bytes:
    ...
```

Some care would need to be taken to duplicate any other interfaces that we import from the `snappy` namespace.

p2p/discovery.py
```diff
@@ -220,7 +220,8 @@ def is_bond_valid_with(self, node_id: NodeID) -> bool:
             pong_time = self._last_pong_at[node_id]
         except KeyError:
             return False
-        return pong_time > (time.monotonic() - constants.KADEMLIA_BOND_EXPIRATION)
+        # type ignored to fix https://github.com/ethereum/trinity/issues/1520
+        return pong_time > (time.monotonic() - constants.KADEMLIA_BOND_EXPIRATION)  # type: ignore
```
pipermerriam, line 224: This one is a little confusing for me.  Why can't `mypy` understand this statement as evaluating to a `bool`?  What does it think it is?

p2p/discovery.py
```diff
@@ -1273,7 +1274,8 @@ def _pack_v4(cmd_id: int, payload: Sequence[Any], privkey: datatypes.PrivateKey)
     encoded_data = cmd_id_bytes + rlp.encode(payload)
     signature = privkey.sign_msg(encoded_data)
     message_hash = keccak(signature.to_bytes() + encoded_data)
-    return message_hash + signature.to_bytes() + encoded_data
+    # type ignored to fix https://github.com/ethereum/trinity/issues/1520
+    return message_hash + signature.to_bytes() + encoded_data  # type: ignore
```
pipermerriam, line 1278: my guess is that this is because of line 1274: `encoded_data = cmd_id_bytes + rlp.encode(payload)` where the call to `rlp.encode` causes `encoded_data` to be any `Any` type.  This *might* be a case where using a `cast` could save us from having to `type: ignore` this and that it would be cleaner.

p2p/discovery.py
```diff
@@ -1365,7 +1368,8 @@ async def receive(
 
 
 def node_id_from_pubkey(pubkey: eth_keys.keys.PublicKey) -> NodeID:
-    return keccak(pubkey.to_bytes())
+    # type ignored to fix https://github.com/ethereum/trinity/issues/1520
+    return keccak(pubkey.to_bytes())  # type: ignore
```
pipermerriam, line 1372: In this case, I think our preferred approach would be `return NodeID(keccak(...))`

p2p/kademlia.py
```diff
@@ -66,19 +66,23 @@ def __init__(self, ip: str, udp_port: int, tcp_port: int) -> None:
 
     @property
     def is_loopback(self) -> bool:
-        return self._ip.is_loopback
+        # type ignored to fix https://github.com/ethereum/trinity/issues/1520
```
pipermerriam, line 69: Can you check whether this is remedied by adding an explicit type hint for `self._ip`

```python
class Address(AddressAPI):
    _ip = Union[ipaddress.IPv4Address, ipaddress.IPv6Address]
    # or maybe just the v4 version
    _ip: ipaddress.IPv4Address
```

It seems like there could be an easy route to these not needing to be `type: ignore`

trinity/components/builtin/beam_preview/component.py
```diff
@@ -33,7 +33,8 @@ class BeamChainPreviewComponent(AsyncioIsolatedComponent):
 
     @property
     def is_enabled(self) -> bool:
-        return self._boot_info.args.sync_mode.upper() == SYNC_BEAM.upper()
+        # type ignored to fix https://github.com/ethereum/trinity/issues/1520
+        return self._boot_info.args.sync_mode.upper() == SYNC_BEAM.upper()  # type: ignore
```
pipermerriam, line 37: I don't think it is appropriate to do it in this PR but if you'd like to try something in a follow up, I believe that `self._boot_info.args` is of type `argparse.Namespace` which is effectively like an `Any` type.

I've had the idea of creating our own subclass of that:

```python
class TrinityArgs(argparse.Namespace):
    sync_mode: str
    port: int
    ...
```

This would allow us to have a strongly typed class to use for the result of CLI argument parsing and thus have slightly more robust type checking when interacting with it.

trinity/protocol/eth/peer.py
```diff
@@ -272,8 +274,8 @@ async def handle_get_node_data_request(self,
     async def handle_get_pooled_transactions_request(
             self,
             event: GetPooledTransactionsRequest) -> NodeDataBundles:
-
-        return await self.with_node_and_timeout(
+        # type ignored to fix https://github.com/ethereum/trinity/issues/1520
+        return await self.with_node_and_timeout(  # type: ignore
```
pipermerriam, line 278: These would be nice to know more about why they need the `type: ignore` as it isn't clear from just reading the code.

trinity/sync/beam/backfill.py
```diff
@@ -868,7 +869,8 @@ def delete_cache(self, prefix: Nibbles) -> None:
 
     @property
     def is_complete(self) -> bool:
-        return self._trie_fog.is_complete
+        # type ignored to fix https://github.com/ethereum/trinity/issues/1520
+        return self._trie_fog.is_complete  # type: ignore
```
pipermerriam, line 873: Another place where it isn't entirely clear why the ignore is needed.  I think it's because our `py-trie` library isn't exposing type hints so the `self._trie_fog` is `Any` typed.  Might be worth explicitly commenting on that here since the linked issue only really talks about rlp and cached property.

pipermerriam (approved):
General :+1: I commented on some of the `type: ignore` comments could use a better comment describing why since many of them aren't actually explained well by the linked issue.  An astute reader will *know* that `rlp.decode(...)` produces `Any` types, but not necessarily why `return receipts.payload.receipts[0]` is resulting in an `Any` type.

Elnaril:
@pipermerriam : Thank you very much for your feedbacks.
I'll look into them and update accordingly.

p2p/commands.py
Elnaril: - snappy.pyi created and 'type: ignore' removed for snappy compress and decompress functions.
- commit https://github.com/ethereum/trinity/pull/2078/commits/1e4c8d2d04028f0d5496314e77ca93d1fe68e77d
- Not sure why the tests failed after this commit as I don't reproduce the errors.

cburgdorf: >Not sure why the tests failed after this commit as I don't reproduce the errors.

This is on me ([see my comment](https://github.com/ethereum/trinity/pull/2069#pullrequestreview-510344768)) :see_no_evil: I'm waiting for other team mates to wake up to decide about cutting a Py-EVM release today to fix these unrelated issues. We normally don't cut releases on Fridays so I don't want to jump ahead just yet.

p2p/discovery.py
Elnaril: Yes, it puzzled me also each time I encountered the mypy error for a statement with a comparison operator ...

p2p/discovery.py
Elnaril: Actually, mypy infers the type `Any` for the 3 parts of the return statement, even for `signature.to_bytes()`
image
I've let for the moment the `type: ignore`, but let me know if you would prefer to cast the whole return statement instead.

p2p/discovery.py
Elnaril: Indeed, that would be a clean way to solve the issue.
Commit https://github.com/ethereum/trinity/pull/2078/commits/d04a4da49ef232a1a962cac8ba640900b1cfc595

pipermerriam:
Thanks, will re-trigger tests once the Py-EVM issue is resolved.  Thanks for the follow up.

p2p/kademlia.py
Elnaril: - You were right, it works pretty well with a typing hint on `_ip`.
- I took the opportunity to re-order the imports from the `typing` module.
- The typing issue in the `ip_packed` method was then converted to: _'Mypy: Incompatible return value type (got "**bytes**", expected "**str**")'_.
To fix that I tried first to cast the output bytes into a string: `return str(self._ip.packed)`. But it made _py38-eth1-core_ fail in tests/conftest.py while loading the constants `DEFAULT_PREFERRED_NODES` raising a `ValueError` from ipaddress.py/`ip_address(address)`.
So, I just used cast() instead ...
- commit: https://github.com/ethereum/trinity/pull/2078/commits/2d6a5a99acea2cb8379f2b63dc319aaa2e51f75a

p2p/commands.py
Elnaril: @cburgdorf : Thank you for letting me know :)

Elnaril:
Ok, thank you. In the meantime, I'll carry on with the remaining points.

trinity/components/builtin/beam_preview/component.py
Elnaril: Sounds great to me !!
Should I create an issue for this ? Or you or a team member would do it ?

pipermerriam: https://github.com/ethereum/trinity/issues/2084

trinity/protocol/eth/peer.py
Elnaril: I have improved the comments for the 4 concerned methods.
Commit: https://github.com/ethereum/trinity/pull/2078/commits/2d4a8aa3aa97e05aad2e004f14ecf1573439f617

trinity/sync/beam/backfill.py
Elnaril: I have improved the comment.
Commit: https://github.com/ethereum/trinity/pull/2078/commits/7fc11f8ac1543fb04cb8a9248083e94e7d0a7bad

Elnaril:
@pipermerriam : I addressed and commented all the points you mentioned.
Could you please re-trigger the CI tests once the underlying issue is fixed ?

Elnaril:
@pipermerriam
Hello,
I've "re-triggered" the tests and now all of them passed except `py37-long_run_integration` for which I don't reproduce the error: 

`____ ERROR at teardown of test_does_not_throw_errors_on_long_run[command0] _____`
`...`
`E                               OSError: [Errno 39] Directory not empty: 'ropsten'`

Is there still an unrelated issue ? Or would it be possible to re-trigger this test ?

Elnaril:
Hello @pipermerriam ,
I'm not sure I notified this PR was ready for review the right way ... Is directly mentioning the reviewer in a comment the preferred method ?

Elnaril:
Hello @pipermerriam 
Shall we finalize this PR as well ?

PR #2072, opened 2020-10-07
#1785 - On Clique, checkpoints need to be aligned with Clique checkpoints

Elnaril:
### What was wrong?
Using --sync-from-checkpoint eth://block/byetherscan/latest on Clique networks, like Görli, crashed the application with unexpected exceptions like SnapshotNotFound.

### How was it fixed?
 - The consensus algorithm name was added for each network supported by Trinity, ethash for Mainnet and Ropsten, clique for Görli as constants in trinity/constants.py, along with the epoch_length.
- When retrieving the latest block number for a clique network, Trinity will now search for the latest block number satisfying blocknumber % epoch_length == 0

### To-Do

- First things first : overall, is it a correct way to solve the issue ?
- I've set the new constants, along with a new typing hint class in trinity/constants.py, with the network ids. Is it the right place ?
- I thought of adding a unit test in trinity/tests/integration/test_trinity_cli.py, similar to test_does_not_throw_errors_on_short_run, but that would mean to set the timeout at roughly 120s to be sure the SnapshotNotFound exception is not raised. But it feels a bit too long to me for a unit test ...
Maybe I just should go for a test about the new method get_latest_clique_checkpoint_block_number  in /trinity/tests/integration/test_etherscan_checkpoint_resolver.py ?
- Last point : I tried to read all coding guidelines, but ... maybe I miss some of them ?

#### Cute Animal Picture
Love this section !! :-)
put a cute animal picture link inside the parentheses

PR #2079, opened 2020-10-13
Fix flaky tests

njgheorghita:
### What was wrong?
The `wheel-cli` tests were acting flaky since the introduction of the `NewBlock` component in #2058 

### How was it fixed?
Added an explicit return to an async function in the `NewBlock` component.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/trinity/components/builtin/new_block/component.py
+++ b/trinity/components/builtin/new_block/component.py
@@ -123,6 +123,7 @@ class NewBlockService(Service):
             self._peer_block_tracker[block.hash].append(str(target_peer))
             # add checkpoint here to guarantee the event loop is released per iteration
             await trio.sleep(0)
+        return
 
     async def _broadcast_newly_seen_block(self, header: BlockHeaderAPI) -> None:
         """


njgheorghita:
@gsalgado It seems as though I may have caused some [flaky tests](https://github.com/ethereum/trinity/pull/2072) with adding the `NewBlockComponent` in #2058. Adding an explicit return to the daemon task seems to have quieted down the failing tests in this pr and [here](https://github.com/ethereum/trinity/pull/2082). Does this fix make sense to you? I'm not convinced the explicit return actually changes the behavior in any meaningful way. If you agree, is there any other spot that sticks out to you in the `NewBlockComponent` as potentially problematic?

gsalgado:
That explicit return should have no effect whatsoever. Are you sure these failures were introduced in #2058?

njgheorghita:
@gsalgado Thanks, and no I'm not sure at all. However, I have noticed the `NewBlockComponent` pop up in a few of the various stack traces, which makes me suspicious. The `NewBlockComponent` is a trio-based service - could there be compatibility issues with `pytest.asyncio` tests?

gsalgado:
ISTM the failure is caused by influxdb not accepting connections, no? In https://circleci.com/api/v1.1/project/github/ethereum/trinity/289639/output/107/0?file=true&allocation-id=5f858223c9af110362af44d2-0-build%2FCB5BA71 I see the following:

```
>           raise ConnectionError(e, request=request)
E           requests.exceptions.ConnectionError: HTTPConnectionPool(host='localhost', port=8086): Max retries exceeded with url: /query?db=trinity&epoch=ns&q=select+%2A+from+%22trinity.p2p%2Fpeers.counter%22+WHERE+time+%3E%3D+%272020-10-13T10%3A36%3A42Z%27 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fc0599d6390>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

Maybe influxdb failed to start or crashed?  I see the influxdb build script wait until it's listening on port 8086, but maybe it should actually connect via HTTP to ensure it's actually working?

gsalgado:
Indeed, https://app.circleci.com/pipelines/github/ethereum/trinity/7652/workflows/f68af374-60c3-432e-90fe-766998d5775e/jobs/289738 didn't build influxdb even though the build_influx script was there. After I rebased it built influxdb and the test passed. I'm not sure what's the intermittent test failure you're seeing, then?

njgheorghita:
Ahh sorry, the test failure I was referencing was from [this pr](https://github.com/ethereum/trinity/pull/2072) - which also disappeared after retriggering the tests.

PR #2081, opened 2020-10-13
WIP: Witness fetching upon receipt of NewBlock

gsalgado:


njgheorghita:
I think it'll help the failing tests here if you rebase against master. In the `wheel-cli` steps on circle there should be a `build influxdb` step

trinity/components/builtin/new_block/component.py
```diff
@@ -156,3 +169,45 @@ def _filter_eligible_peers(self,
...
+    except Exception as err:
+        logger.warning(
+            "Error fetching witnesses for block %s from %s: %s", header.hash, peer, err)
+        return tuple()
+    else:
+        if witness_hashes:
+            logger.debug(
+                "Got witness hashes for block %s, asking BeamSyncer to fetch them", header.hash)
+            await event_bus.broadcast(
+                CollectMissingTrieNodes(
```
gsalgado, line 204: @carver how do you feel about triggering the download of the trie nodes immediately when we receive a `NewBlock`, as opposed to waiting until we're about to import the block?

carver: No problem, especially if we're prioritizing the download order by the older blocks, so we don't clog the request queue with recent state downloads.

gsalgado:
Split into separate PRs

PR #2089, opened 2020-10-22
Support various mining consensus when validating newly seen blocks

### LINKED ISSUE #2087: NewBlockService attempts a POW-check regardless of the chain's mining method
### What is wrong?

It calls `check_pow()` on every received block, regardless of the chain's mining method. 

### How can it be fixed

Use a seal-validation method appropriate for the configured chain

njgheorghita:
### What was wrong?
Fixes #2087 

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/trinity/components/builtin/new_block/component.py
+++ b/trinity/components/builtin/new_block/component.py
@@ -11,8 +11,10 @@ from async_service import (
     Service,
     background_trio_service,
 )
-from eth.abc import BlockHeaderAPI, BlockAPI
-from eth.consensus.pow import check_pow
+from eth.abc import (
+    BlockAPI,
+    BlockHeaderAPI,
+)
 from eth_utils import (
     ValidationError,
     to_tuple,
@@ -21,6 +23,7 @@ from lahja import EndpointAPI
 import trio
 
 from p2p.abc import SessionAPI
+from trinity.boot_info import BootInfo
 from trinity.constants import TO_NETWORKING_BROADCAST_CONFIG
 from trinity.extensibility import TrioIsolatedComponent
 from trinity.protocol.eth.events import NewBlockEvent
@@ -33,6 +36,7 @@ from trinity.protocol.eth.peer import (
     ETHProxyPeer,
 )
 from trinity.sync.common.events import NewBlockImported
+from trinity._utils.connect import get_eth1_chain_with_remote_db
 from trinity._utils.logging import get_logger
 
 
@@ -50,7 +54,7 @@ class NewBlockComponent(TrioIsolatedComponent):
     async def do_run(self, event_bus: EndpointAPI) -> None:
         proxy_peer_pool = ETHProxyPeerPool(event_bus, TO_NETWORKING_BROADCAST_CONFIG)
         async with background_trio_service(proxy_peer_pool):
-            service = NewBlockService(event_bus, proxy_peer_pool)
+            service = NewBlockService(event_bus, proxy_peer_pool, self._boot_info)
             async with background_trio_service(service) as manager:
                 await manager.wait_finished()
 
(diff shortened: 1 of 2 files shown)


trinity/components/builtin/new_block/component.py
```diff
@@ -91,15 +95,14 @@ async def _handle_new_block(self, sender: SessionAPI, block: NewBlockPayload) ->
...
-                    header.mix_hash,
-                    header.nonce,
-                    header.difficulty
+                with get_eth1_chain_with_remote_db(self._boot_info, self._event_bus) as chain:
+                    chain.validate_seal(header)
+            except ValidationError as exc:
+                self.logger.info(
+                    "Received invalid block from peer: %s. Reason: %s",
+                    sender_peer_str,
+                    exc,
```
njgheorghita, line 104: @gsalgado Could I get your thoughts on the solution here for using seal validation rather than `check_pow` for every chain. This solution validates the seal correctly for pow chains - but when it's connected to clique/goerli - it will not validate a single block because the timestamp on the newly seen block header is consistently 6-8 seconds ahead of `time.time()`.
https://github.com/ethereum/py-evm/blob/5b177029d4d08454fa097bcd451c89b1d4a92e89/eth/consensus/clique/_utils.py#L128-L131

gsalgado: That's weird. I'm not familiar with clique/goerli at all, but I wonder how you're getting blocks with a future timestamp

njgheorghita: Ok thanks, yeah it seems weird to me also - but I'll dig a little deeper and see if I can uncover what's going on

njgheorghita: @cburgdorf Any thoughts on this as the clique expert? All newly received headers validate properly after adding a buffer of `15` seconds to the header validation below...

https://github.com/ethereum/py-evm/blob/5b177029d4d08454fa097bcd451c89b1d4a92e89/eth/consensus/clique/_utils.py#L128-L131

It seems like [other clients](https://ethereum.stackexchange.com/questions/27485/how-do-ethereum-clients-cope-with-future-blocks) allow for a 30-second buffer.

cburgdorf: Calling `chain.validate_seal` is the right approach. Notice that this will only check the basic integrity of the header for Görli and one would also need to call `validate_seal_extension` to check the actual consensus rules for clique (but unlike in PoW, this can not be done out of context in Clique).

Regarding the timestamp: [Geth also rejects](https://github.com/ethereum/go-ethereum/blob/7b748e550a3234c5240427d0164a058829dc83dc/consensus/clique/clique.go#L251-L254) headers with timestamps from the future. Are you saying that all new blocks constantly arrive with time stamps from the future? Did you check if the system time of your machines might be off by a few seconds? Is this happening on the Görli network or some other network?

carver:
@njgheorghita Yeah, let's treat the timestamp issue as separate. Apart from linting, I think this is basically good to go. Are you able to wrap it up soon?

njgheorghita:
@carver Yup! I created an [issue](https://github.com/ethereum/trinity/issues/2092) to track the timestamp question. I can make the merge with a 👍

trinity/components/builtin/new_block/component.py
```diff
@@ -91,15 +99,13 @@ async def _handle_new_block(self, sender: SessionAPI, block: NewBlockPayload) ->
...
-                    header.mining_hash,
-                    header.mix_hash,
-                    header.nonce,
-                    header.difficulty
+                with get_eth1_chain_with_remote_db(self._boot_info, self._event_bus) as chain:
+                    chain.validate_seal(header)
+            except ValidationError as exc:
+                self.logger.info(
+                    "Received invalid block from peer: %s. %s",
+                    sender_peer_str, exc,
```
carver, line 107: Generally, keeping the try/except as tight as possible is a good thing. Like if `get_eth1_chain_with_remote_db` could raise a validation exception, we wouldn't want it mixed in here. It's not a practical concern at the moment, AFAIK, but there's no harm in tightening up the try block. So, structurally:

```py
with ...
  try:
    chain.validate
  except:
```

carver (approved):

PR #2090, opened 2020-10-22
Disable resolver in dockerfile for consistent builds

njgheorghita:
### What was wrong?
The `docker-trinity-image-build-test` CI run has been consistently flaky for me recently.

It seems to me that the problem is similar to the one described [in this issue](https://github.com/pypa/pip/issues/8713) - where essentially the `--use-feature=2020-resolver` pytest flag is very slow and inefficient. I've seen multiple instances of the resolver installing many different versions of the same dependency while trying to resolve the dependency tree. This poor performance is acknowledged and is under active improvements [by the pip team](https://github.com/pypa/pip/issues/8664). The `docker-trinity-image-build-test` consistently times out during the install step with the `2020-resolver` enabled.

The pip team has already released some improvements to the resolver in `pip==20.2.4`, but  it seems like whatever improvements were included in `20.2.4` are not sufficient to prevent timeouts in the docker install step. They intend to release further improvements in `20.3`.

### How was it fixed?

The solution here is to just remove the `--use-feature=2020-resolver` flag for the time being. This results in consistently quick & passing CI tests. 

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -9,8 +9,8 @@ COPY . /usr/src/app
 RUN apt-get update
 RUN apt-get -y install libsnappy-dev
 
-RUN pip install -e .[dev]  --no-cache-dir --use-feature=2020-resolver
-RUN pip install -U trinity --no-cache-dir --use-feature=2020-resolver
+RUN pip install -e .[dev]  --no-cache-dir
+RUN pip install -U trinity --no-cache-dir
 
 RUN echo "Type \`trinity\` to boot or \`trinity --help\` for an overview of commands"
 


njgheorghita:
@carver I noticed you included the `2020-resolver` flag [here](https://github.com/ethereum/trinity/pull/1992/files). I'm not sure how important you consider this flag to be? It seems to me like if we don't mind the `docker-trinity-image-build-test` behaving mildly flaky, then it's not necessary to remove the flag.

carver:
Flaky is definitely bad, so I agree that doing nothing is not the best resolution.

I have had experiences where the *only* way to install it correctly was to use the new resolver. As long as it's green, I'm fine with it for now. The behavior will be default soon enough.

njgheorghita:
> I have had experiences where the only way to install it correctly was to use the new resolver. 

Ok, cool - yeah it seems like that's no longer the case. Do you think it's worth leaving a note? When you first introduced the flag - how obvious of a fix was it to introduce the flag?

carver:
It wasn't totally obvious, but since it will be the default soon, we don't have to spend too much time on it.

njgheorghita:
Cool, i'll merge this in if you can give the official 👍

carver (approved):

PR #2099, opened 2020-11-10
Implement AsyncWitnessDB, to store recent block witness hashes

gsalgado:


tests/core/p2p-proto/test_wit_db.py
```diff
@@ -0,0 +1,27 @@
...
+from trinity.protocol.wit.db import AsyncWitnessDB
+from trinity.tools.factories import Hash32Factory
+
+
+@pytest.mark.asyncio
+async def test_db():
+    wit_db = AsyncWitnessDB(AtomicDB())
+
+    hash1 = Hash32Factory()
+    with pytest.raises(KeyError):
```
carver, line 14: I like the pattern we've used with creating a special exception (like `WitnessHashesUnavailable`) so as not to confuse a missing witness with some unrelated `KeyError` bug. :+1: if you're up for adding that here.

trinity/protocol/wit/db.py
```diff
@@ -0,0 +1,61 @@
...
+        return tuple(rlp.decode(self.db[self._make_block_witness_hashes_lookup_key(block_hash)]))
+
+    def persist_witness_hashes(
+            self, block_hash: Hash32, witness_hashes: Tuple[Hash32, ...]) -> None:
+        try:
+            recent_blocks_with_witnesses = list(self._get_recent_blocks_with_witnesses())
+        except KeyError:
+            recent_blocks_with_witnesses = []
+
+        if len(recent_blocks_with_witnesses) == self._max_witness_history:
```
carver, line 51: If there was a bug that caused the witness list to get too long, this check would stop removing old witnesses.

One option is to do this:
```suggestion
        while len(recent_blocks_with_witnesses) >= self._max_witness_history:
```

trinity/protocol/wit/db.py
```diff
@@ -0,0 +1,61 @@
...
+        witness_hashes: Tuple[Hash32, ...],
+    ) -> None:
+        ...
+
+
+class AsyncWitnessDB(AsyncWitnessDataBaseAPI):
+    _recent_blocks_with_witnesses_lookup_key = b'recent-blocks-with-witness-hashes'
+    _max_witness_history = 256
+
+    def __init__(self, db: AtomicDatabaseAPI) -> None:
```
carver, line 32: What do you think about requiring an Async DB input here? Maybe you already considered that approach, I'm just curious what issues you run into.

carver (approved):
Yay for small PRs. Looks good!

trinity/protocol/wit/db.py
gsalgado: You mean an `*DatabaseAPI` that provides async methods? Do we have something like that already? Not sure what that'd give us, though

PR #2100, opened 2020-11-10
Implement the wit p2p sub-protocol

gsalgado:
Includes #2099; only last commit should be reviewed

### DIFF
--- a/trinity/components/builtin/request_server/component.py
+++ b/trinity/components/builtin/request_server/component.py
@@ -3,7 +3,7 @@ from argparse import (
     _SubParsersAction,
 )
 
-from async_service import Service, background_asyncio_service
+from async_service import Service
 from lahja import EndpointAPI
 
 from eth.db.backends.base import BaseAtomicDB
@@ -23,6 +23,8 @@ from trinity.extensibility import (
 )
 from trinity.protocol.eth.servers import ETHRequestServer
 from trinity.protocol.les.servers import LightRequestServer
+from trinity.protocol.wit.servers import WitRequestServer
+from trinity._utils.services import run_background_asyncio_services
 
 
 class RequestServerComponent(AsyncioIsolatedComponent):
@@ -46,7 +48,7 @@ class RequestServerComponent(AsyncioIsolatedComponent):
         base_db = DBClient.connect(trinity_config.database_ipc_path)
         with base_db:
             if trinity_config.has_app_config(Eth1AppConfig):
-                server = self.make_eth1_request_server(
+                eth_server = self.make_eth1_request_server(
                     trinity_config.get_app_config(Eth1AppConfig),
                     base_db,
                     event_bus,
@@ -54,8 +56,10 @@ class RequestServerComponent(AsyncioIsolatedComponent):
             else:
                 raise Exception("Trinity config must have eth1 config")
 
-            async with background_asyncio_service(server) as manager:
-                await manager.wait_finished()
+            wit_server = self.make_wit_request_server(
+                trinity_config.get_app_config(Eth1AppConfig), base_db, event_bus)
+
+            await run_background_asyncio_services([eth_server, wit_server])
 
     @classmethod
     def make_eth1_request_server(cls,
@@ -83,3 +87,10 @@ class RequestServerComponent(AsyncioIsolatedComponent):
             raise Exception(f"Unsupported Database Mode: {app_config.database_mode}")
 
         return server
+
+    @classmethod
+    def make_wit_request_server(cls,
+                                app_config: Eth1AppConfig,
+                                base_db: BaseAtomicDB,
+                                event_bus: EndpointAPI) -> Service:
+        return WitRequestServer(event_bus, TO_NETWORKING_BROADCAST_CONFIG, base_db)
--- a/trinity/protocol/common/validators.py
+++ b/trinity/protocol/common/validators.py
@@ -1,6 +1,7 @@
 from abc import abstractmethod
 import collections
 from typing import (
+    Any,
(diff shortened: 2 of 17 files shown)


trinity/protocol/eth/peer.py
```diff
@@ -150,7 +189,7 @@ class ETHPeerFactory(BaseChainPeerFactory):
     async def get_handshakers(self) -> Tuple[HandshakerAPI[Any], ...]:
         headerdb = self.context.headerdb
         head = await headerdb.coro_get_canonical_head()
-        total_difficulty = await headerdb.coro_get_score(head.hash)
+        total_difficulty = await self.context.headerdb.coro_get_score(head.hash)
```
carver, line 192: What's this change about? It looks unnecessary.

trinity/protocol/eth/peer.py
```diff
@@ -183,11 +222,16 @@ async def get_handshakers(self) -> Tuple[HandshakerAPI[Any], ...]:
...
-        return (
+        eth_handshakers = (
             ETHV63Handshaker(handshake_v63_params),
             ETHHandshaker(handshake_params, head.block_number, fork_blocks, highest_eth_protocol)
         )
 
+        return eth_handshakers + self._get_wit_handshakers()
+
+    def _get_wit_handshakers(self) -> Tuple[HandshakerAPI[Any], ...]:
+        return (WitnessHandshaker(),)
```
carver, line 233: What's the reason for this indirection? I was expecting something like this above:
...
```py
            ETHHandshaker(handshake_params, head.block_number, fork_blocks, highest_eth_protocol),
            WitnessHandshaker(),
        )

File: trinity/protocol/eth/peer.py [hunk 3]
```diff
@@ -229,9 +278,31 @@ async def run(self) -> None:
...
+        self.run_daemon_request(
+            GetBlockWitnessHashesRequest, self.handle_get_block_witness_hashes_request)
 
         await super().run()
 
+    @async_fire_and_forget
+    async def handle_send_block_witness_hashes_command(
+            self, event: SendBlockWitnessHashesEvent) -> None:
+        peer = self.get_peer(event.session)
+        if not hasattr(peer, 'wit_api'):
```
carver, line 290: Since `wit_api` is always present (according to the XXX comment), this doesn't seem to do much now. But I guess it's good to have in place for later when that changes.

carver (approved):
Looks good! I ignored the commit that's part of #2099 in this diff.

trinity/components/builtin/request_server/component.py
```diff
@@ -46,18 +48,19 @@ async def do_run(self, event_bus: EndpointAPI) -> None:
...
                 raise Exception("Trinity config must have eth1 config")
 
-            async with background_asyncio_service(server) as manager:
-                await manager.wait_finished()
+            wit_server = self.make_wit_request_server(
+                trinity_config.get_app_config(Eth1AppConfig), base_db, event_bus)
+
+            await run_background_asyncio_services([eth_server, wit_server])
 
-    @classmethod
```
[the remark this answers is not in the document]
gsalgado: Yeah, I think this was from an attempted refactoring. Reverted

trinity/components/builtin/request_server/component.py
```diff
@@ -83,3 +86,9 @@ def make_eth1_request_server(cls,
             raise Exception(f"Unsupported Database Mode: {app_config.database_mode}")
 
         return server
+
+    def make_wit_request_server(cls,
```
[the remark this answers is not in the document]
gsalgado: Yep, made it a classmethod for consistency

trinity/protocol/common/validators.py
```diff
@@ -155,3 +156,8 @@ def _validate_sequence(self, block_numbers: Sequence[BlockNumber]) -> None:
                 'Duplicate headers returned.\n'
                 f'- duplicates: {humanize_integer_sequence(sorted(duplicates))}\n'
             )
+
+
+def match_payload_request_id(request: Any, response: Any) -> None:
+    if request.request_id != response.payload.request_id:
+        raise ValidationError("Request `id` does not match")
```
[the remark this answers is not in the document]
gsalgado: I've included the whole response.payload, in fact

trinity/protocol/eth/handshaker.py
```diff
@@ -144,6 +144,7 @@ async def do_handshake(self,
         Raise HandshakeFailure if the handshake is not successful.
         """
 
+        self.logger.debug("Performing %s handshake with %s", protocol, multiplexer.remote)
```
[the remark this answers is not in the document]
gsalgado: I think it's worth leaving, yeah. That way we can see all handshakes we attempt with a given peer, including the ones that are a no-op like wit/0

trinity/protocol/eth/peer.py
gsalgado: Yep, more leftover from attempted refactorings, I believe. Reverted

trinity/protocol/eth/peer.py
gsalgado: We need to override that method in a test; will add a comment

trinity/protocol/eth/peer.py
gsalgado: Actually, the comment above is about `ETHProxyPeer.wit_api`, which will always be set, whereas this is about `ETHPeer.wit_api`, which won't be set if the remote does not support wit/0, so if these error logs show up somewhere it means we have a bug on our side or a remote peer is sending wit/0 requests without performing the wit handshake

Issue #2096, opened 2020-11-03
NewBlock component propagates received blocks incorrectly


njgheorghita:
### What is wrong?
According to the [devp2p](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#block-propagation) spec, a `NewBlock` message should be propagated after receiving a `NewBlock` message. Currently, the `NewBlockService` propagates a `BlockHeadersV65` command after validating an inbound block. 

### How can it be fixed
Update `NewBlockService. _broadcast_imported_block()` to propagate the right command.

gsalgado:
Also worth noting that for blocks mined locally, we currently send just a `NewBlockHashes` msg to each of our peers, whereas geth first sends a `NewBlock` to a subset of the peers and then a `NewBlockHashes` to each of them: https://github.com/ethereum/go-ethereum/blob/43c278cdf93d5469702fd1c2f570dbf3c1718ff0/eth/handler.go

gsalgado:
Oh, and `NewBlockComponent` doesn't actually send any `NewBlockHashes` msg over the wire as it emits a `NewBlockHashesEvent`, which is the event use to notify of a newly received msg and not the one to cause the `PeerPoolEventServer` to send a msg

gsalgado:
I need those fixes to test the witness propagation, so will take care of them

njgheorghita:
@gsalgado did you still need these fixes? I'm back from time off, and can pick them up but don't want to duplicate work if you've already made progress on this

gsalgado:
I have a fix for this in #2081 already, will move that into a separate PR soon

PR #2101, opened 2020-11-10
Fix a couple bugs in NewBlockComponent

gsalgado:
- Upon receiving a NewBlock msg, broadcast a NewBlock msg instead of
  BlockHeadersV65

- Actually send NewBlockHashes msgs. Before it was simply firing a
  NewBlockHashesEvent, which is meant to be used when we receive a
  msg of that type

Closes: #2096

#### Cute Animal Picture

### DIFF
--- a/trinity/components/builtin/new_block/component.py
+++ b/trinity/components/builtin/new_block/component.py
@@ -1,7 +1,8 @@
+import collections
 import math
 import random
 from typing import (
-    Dict,
+    DefaultDict,
     Iterable,
     List,
     Tuple,
@@ -13,7 +14,6 @@ from async_service import (
 )
 from eth.abc import (
     BlockAPI,
-    BlockHeaderAPI,
 )
 from eth_utils import (
     ValidationError,
@@ -28,6 +28,7 @@ from trinity.constants import TO_NETWORKING_BROADCAST_CONFIG
 from trinity.extensibility import TrioIsolatedComponent
 from trinity.protocol.eth.events import NewBlockEvent
 from trinity.protocol.eth.payloads import (
+    BlockFields,
     NewBlockHash,
     NewBlockPayload,
 )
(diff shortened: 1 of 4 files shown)


trinity/components/builtin/new_block/component.py
```diff
@@ -82,10 +82,12 @@ async def run(self) -> None:
 
     async def _handle_imported_blocks(self) -> None:
         async for event in self._event_bus.stream(NewBlockImported):
-            await self._broadcast_imported_block(event.block)
+            block = event.block
+            self.logger.debug("NewBlockImported: %s", block)
+            await self._broadcast_new_block_hashes(event.block)
```
njgheorghita, line 87: minor nitpick: is there a reason for using `event.block` here instead of `block`?

trinity/components/builtin/new_block/component.py
```diff
@@ -82,10 +82,12 @@ async def run(self) -> None:
...
 
-    async def _handle_new_block(self, sender: SessionAPI, block: NewBlockPayload) -> None:
-        header = block.block.header
+    async def _handle_new_block(self, sender: SessionAPI, payload: NewBlockPayload) -> None:
+        header = payload.block.header
```
njgheorghita, line 90: ahh, yeah that's a significantly better variable name

trinity/protocol/eth/events.py
```diff
@@ -155,6 +155,25 @@ class SendBlockBodiesEvent(PeerPoolMessageEvent):
...
+class SendNewBlockHashesEvent(PeerPoolMessageEvent):
+    """
+    Event to proxy a ``ETHPeer.sub_proto.send_new_block_hashes`` call from a proxy peer to the
+    actual peer that sits in the peer pool.
+    """
+    session: SessionAPI
+    command: NewBlockHashes
+
+
+class SendNewBlockEvent(PeerPoolMessageEvent):
```
njgheorghita, line 168: should this event also be marked as a `dataclass`?

njgheorghita (approved):
Looks good! Thanks for cleaning this up!

trinity/components/builtin/new_block/component.py
gsalgado: Nope, just an oversight

trinity/protocol/eth/events.py
gsalgado: `PeerPoolMessageEvent` already is, so I'm not sure it's necessary to have that again here, but for consistency with the others I'll add

PR #2103, opened 2020-11-12
Download block witnesses upon receiving a NewBlock msg

gsalgado:
First download the witness hashes, then emit an event to
cause the BeamSyncer to download the actual trie nodes

### DIFF
--- a/trinity/components/builtin/new_block/component.py
+++ b/trinity/components/builtin/new_block/component.py
@@ -1,5 +1,7 @@
+import asyncio
 import collections
 import math
+import pathlib
 import random
 from typing import (
     DefaultDict,
@@ -14,8 +16,11 @@ from async_service import (
 )
 from eth.abc import (
     BlockAPI,
+    BlockHeaderAPI,
 )
+from eth_typing import Hash32
 from eth_utils import (
+    ExtendedDebugLogger,
     ValidationError,
     to_tuple,
 )
@@ -25,6 +30,7 @@ import trio
 from p2p.abc import SessionAPI
 from trinity.boot_info import BootInfo
 from trinity.constants import TO_NETWORKING_BROADCAST_CONFIG
+from trinity.db.manager import DBClient
 from trinity.extensibility import TrioIsolatedComponent
 from trinity.protocol.eth.events import NewBlockEvent
 from trinity.protocol.eth.payloads import (
@@ -36,7 +42,8 @@ from trinity.protocol.eth.peer import (
     ETHProxyPeerPool,
     ETHProxyPeer,
 )
-from trinity.sync.common.events import NewBlockImported
+from trinity.protocol.wit.db import AsyncWitnessDB
+from trinity.sync.common.events import CollectMissingTrieNodes, NewBlockImported
 from trinity._utils.connect import get_eth1_chain_with_remote_db
 from trinity._utils.logging import get_logger
 
@@ -88,11 +95,12 @@ class NewBlockService(Service):
 
     async def _handle_new_block(self, sender: SessionAPI, payload: NewBlockPayload) -> None:
         header = payload.block.header
-        sender_peer_str = str(ETHProxyPeer.from_session(
+        sender_peer = ETHProxyPeer.from_session(
             sender,
             self._event_bus,
             TO_NETWORKING_BROADCAST_CONFIG
-        ))
+        )
+        sender_peer_str = str(sender_peer)
 
         # Add peer to tracker if we've seen this block before
         if header.hash in self._peer_block_tracker:
@@ -109,6 +117,7 @@ class NewBlockService(Service):
                         sender_peer_str, exc,
                     )
                 else:
+                    self.manager.run_task(self._fetch_witnesses, sender_peer, header)
                     self._peer_block_tracker[header.hash] = [sender_peer_str]
                     # Here we only broadcast a NewBlock msg to a subset of our peers, and once the
                     # block is imported into our chain a NewBlockImported event will be generated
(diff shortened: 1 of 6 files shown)


trinity/components/builtin/new_block/component.py
```diff
@@ -156,3 +165,50 @@ def _filter_eligible_peers(self,
...
+        logger.warning(
+            "Error fetching witnesses for block %s from %s: %s", header.hash, peer, err)
+        return tuple()
+    else:
+        if witness_hashes:
+            logger.debug(
+                "Got witness hashes for block %s, asking BeamSyncer to fetch them", header.hash)
+            await event_bus.broadcast(
+                CollectMissingTrieNodes(
+                    witness_hashes, urgent=True, block_number=header.block_number))
```
gsalgado, line 206: Should this be an urgent request? Not sure

trinity/sync/beam/chain.py
```diff
@@ -945,6 +947,17 @@ async def _launch_server(self) -> None:
         self.manager.run_daemon_task(self._provide_missing_account_tries)
         self.manager.run_daemon_task(self._provide_missing_bytecode)
         self.manager.run_daemon_task(self._provide_missing_storage)
+        self.manager.run_daemon_task(self._fetch_missing_trie_nodes)
+
+    async def _fetch_missing_trie_nodes(self) -> None:
+        async for event in self._event_bus.stream(CollectMissingTrieNodes):
+            # XXX: Or should we use _wait_for_nodes() instead?
```
gsalgado, line 954: For urgent requests it will simply delegate to `_wait_for_nodes()`, but for non-urgent it does some extra stuff I don't quite understand

trinity/sync/common/events.py
```diff
@@ -126,6 +126,22 @@ def expected_response_type() -> Type[MissingStorageResult]:
         return MissingStorageResult
 
 
+@dataclass
+class MissingTrieNodesResult(BaseEvent):
+    num_nodes_collected: int = 0
+
+
+@dataclass
+class CollectMissingTrieNodes(BaseRequestResponseEvent[MissingTrieNodesResult]):
```
carver, line 135: Maybe add a quick note to use the more-specific events whenever possible. In the case of witness metadata, we can't distinguish between missing accounts, code, storage. Generally we want to, because some day we'll be able to ask peers for proofs of a given account or storage.

trinity/sync/beam/chain.py
carver: Context: the beam downloader prioritizes downloading by block number, so that the earliest predictive nodes are requested first. I think that's still valuable for when the chain falls behind. Maybe it's not critical for a prototype, but leaving the one layer of indirection through `ensure_nodes_present` doesn't seem to cost us too much here.

Since we have the block number available, we might as well check if the block number is too old here (same as below, where the block number check is in the analogous `_provide_missing_account_tries`, etc).

trinity/protocol/eth/peer.py
```diff
@@ -304,7 +304,9 @@ async def handle_get_block_witness_hashes_request(
             event: GetBlockWitnessHashesRequest) -> Tuple[Hash32, ...]:
         peer = self.get_peer(event.session)
         if not hasattr(peer, 'wit_api'):
-            self.logger.error(
+            # Currently this is expected to happen as NewBlockComponent doesn't know which peers
+            # support the witness protocol so it always tries to fetch witnesses.
```
carver, line 308: It seems so strange to me that we don't have an easier way to get access to this information. But I'm fine with kicking that can down the road a bit. Maybe deserves a quick issue, though?

trinity/components/builtin/new_block/component.py
carver: Usually, urgent means that the block importer is actively importing the block. Maybe it would be pretty quick/easy to check our own tip and check if the new block is only a block or two ahead of our tip?

If not, I'm :+1: with leaving this as hard-coded to `True` for this prototyping. As a long-term strategy it's not great. For times when Beam Sync starts to fall behind, catching up would be more difficult. So a comment or issue would be great.

trinity/components/builtin/new_block/component.py
```diff
@@ -156,3 +165,50 @@ def _filter_eligible_peers(self,
...
                 yield peer
+
+    async def _fetch_witnesses(self, peer: ETHProxyPeer, header: BlockHeaderAPI) -> None:
+        await fetch_witnesses(
+            peer, header, self._event_bus, self._boot_info.trinity_config.database_ipc_path,
+            self.logger)
+
+
+# XXX: Should probably be moved to trinity/sync/beam/ as we'll want to use this in the block
+# previewer.
```
carver, line 176: How are you imagining using it there? Maybe if the new block witness isn't at our local tip block number +1, then the block importer would re-request the hashes as urgent when it catches up to this new block. But as I said in the earlier comment, I think it's :ok_hand:  to skip that logic if you try it and it looks hairy.

trinity/components/builtin/new_block/component.py
```diff
@@ -156,3 +165,50 @@ def _filter_eligible_peers(self,
...
+        if witness_hashes:
+            logger.debug(
+                "Got witness hashes for block %s, asking BeamSyncer to fetch them", header.hash)
+            await event_bus.broadcast(
+                CollectMissingTrieNodes(
+                    witness_hashes, urgent=True, block_number=header.block_number))
+            base_db = DBClient.connect(database_ipc_path)
+            with base_db:
+                wit_db = AsyncWitnessDB(base_db)
+                wit_db.persist_witness_hashes(header.hash, witness_hashes)
```
carver, line 210: Hm, just realized/noticed that we do not neatly handle what happens if we get different answers from different peers. We just overwrite the last witness-hashes at the same block. This would be especially bad if the second peer to send us the witness hashes gives us a list with one hash (to grief us). We also append the block to the witness history, so our history gets shorter than intended...

This might be a problem soonish, but I think it can be all handled in the witness db (storing and reading a union of all the hashes instead of overwriting the previous), so I don't think it needs to be fixed in this PR.

carver (approved):

trinity/components/builtin/new_block/component.py
gsalgado: This comment was from before the call we had yesterday where we agreed to not add witness downloading capabilities to the importer/previewer for the time being. I'll just remove the comment for now

trinity/components/builtin/new_block/component.py
gsalgado: Ok, I'll add an XXX comment here.

trinity/protocol/eth/peer.py
gsalgado: It's because an `ETHProxyPeer` knows only the session ID of the connection and exposes only an `eth_api` property which is implied to work on any connection. I believe it was implemented like that because it has to be pickle-able. I'll file an issue

trinity/components/builtin/new_block/component.py
gsalgado: Since we fire a `CollectMissingTrieNodes` event with the actual node hashes every time we get a `NewBlock` msg, getting different witness hashes from different peers will just cause us to shorten our witness history in the DB, and potentially send an incomplete list of hashes to any peers that ask for it. I agree it's probably better to handle this in the witness db as you suggest

trinity/sync/beam/chain.py
gsalgado: I wonder if it's worth doing that check here as well, given that currently we only run this when receiving a `NewBlock` msg, which is unlikely to be about an old block, no? I'll add a comment for now

carver:
Are you going to switch it over to `NewBlockHashes` in this PR, or a follow-up one? (It seemed like that's the direction you were headed in chat)

gsalgado:
I'm going to do it in a separate PR as I ended up making that change in another branch which included metrics for witnesses, and it'd be annoying to cherry pick that from there.

Issue #2105, opened 2020-11-13
Do not overwrite existing witness hashes for a given block in the DB


gsalgado:
Hm, just realized/noticed that we do not neatly handle what happens if we get different answers from different peers. We just overwrite the last witness-hashes at the same block. This would be especially bad if the second peer to send us the witness hashes gives us a list with one hash (to grief us). We also append the block to the witness history, so our history gets shorter than intended...

This might be a problem soonish, but I think it can be all handled in the witness db (storing and reading a union of all the hashes instead of overwriting the previous), so I don't think it needs to be fixed in this PR.

_Originally posted by @carver in https://github.com/ethereum/trinity/pull/2103#discussion_r522516017_

carver:
I can take a hack at this

carver:
So we need to be sure to handle any race conditions, like two hash sets being saved at roughly the same time, and they both read out an empty set of hashes for a given blockhash. So might need a lock.

Current plan is to return the union of all hashes reported by all peers. It's probably worse to collect not enough data than to collect too much. Someday we might have to deal with griefing by sending way too many hashes, but for now I think empty witness griefing is a bigger concern.

Someday we might want something like a `Counter()` of the hashes returned by all peers, but for now I think a simple union of the set is the best next step.

PR #2106, opened 2020-11-17
Record metrics for block witnesses and missing state

gsalgado:


trinity/components/builtin/new_block/component.py
```diff
@@ -166,50 +178,70 @@ def _filter_eligible_peers(self,
             if str(peer) not in self._peer_block_tracker[block_hash]:
                 yield peer
 
-    async def _fetch_witnesses(self, peer: ETHProxyPeer, header: BlockHeaderAPI) -> None:
+    async def _fetch_witnesses(
+            self, peer: ETHProxyPeer, block_hash: Hash32, block_number: BlockNumber) -> None:
+        # FIXME: Return immediately if we already have witness hashes for the given block hash in
+        # our wit db.
```
carver, line 184: Do you agree that this will be moot when https://github.com/ethereum/trinity/issues/2105 is closed? If so, then the comment can be removed.

trinity/components/builtin/new_block/component.py
```diff
@@ -166,50 +178,70 @@ def _filter_eligible_peers(self,
...
             # of our tip, as otherwise when beam sync start to falls behind it may be more
             # difficult to catch up.
             urgent = True
-            await event_bus.broadcast(
-                CollectMissingTrieNodes(witness_hashes, urgent, header.block_number))
+            try:
+                # Sometimes we get a NewBlock/NewBlockHashes msg before the BeamSyncer
+                # service has started, and there will be no subscribers to CollectMissingTrieNodes
+                # in that case. This ensures we wait for it to start before attempting to fire
+                # CollectMissingTrieNodes events.
```
carver, line 230: Nit - this comment is about the `await event_bus.wait_until_any`... line, so makes more sense to me one line down.

trinity/sync/beam/chain.py
```diff
@@ -755,6 +766,25 @@ async def import_block(
...
+                    "No witness hashes for block %s. Import will be slow",
+                    encode_hex(block.hash),
+                )
+            else:
+                missing = self._state_downloader._get_unique_missing_hashes(wit_hashes)
+                self.logger.debug(
+                    "Missing %d witnesses out of %d for block %s",
+                    len(missing), len(wit_hashes), encode_hex(block.hash))
+                if missing:
+                    self.metrics_registry.counter('trinity.sync/block_witness_missing').inc()
```
carver, line 784: This name confused me at first, thinking that it meant we didn't have access to the witness. I understand this is more like `block_witness_uncollected`, now, if you like that name.

trinity/sync/beam/chain.py
```diff
@@ -755,6 +766,25 @@ async def import_block(
...
+                wit_hashes = wit_db.get_witness_hashes(block.hash)
+            except WitnessHashesUnavailable:
+                self.logger.debug(
+                    "No witness hashes for block %s. Import will be slow",
+                    encode_hex(block.hash),
+                )
+            else:
+                missing = self._state_downloader._get_unique_missing_hashes(wit_hashes)
+                self.logger.debug(
+                    "Missing %d witnesses out of %d for block %s",
```
carver, line 781: The combined thing is the witness. Maybe something like:
```suggestion
                    "Missing %d nodes from witness of %d for block %s",
```

carver (approved):
A lot of the new logs refer to a block by its complete block hash. I find that I usually want to know the block number too, and rarely want the whole hash, so just the `str(block)` is what looks best in the logs to me.

My guess is that you are logging the full hash because that's easiest for you to manually inspect the database. If that's right, then by all means leave it as is. But if the full hash isn't critical to you in the logs, then I'd recommend just logging the `"%s", block`.

trinity/sync/beam/chain.py
gsalgado: Went with `Missing %d nodes out of %d from witness of block %s`

trinity/sync/beam/chain.py
gsalgado: Yeah, that's better!

gsalgado: For the metrics I went with `block_witness_incomplete` and `block_witness_complete`

gsalgado:
> My guess is that you are logging the full hash because that's easiest for you to manually inspect the database

I actually wanted the encoded hash because when reading the logs on vim I could easily jump between all references to that block, but indeed I don't need the whole thing and having the block number is definitely a plus, so I've changed as you suggested.

PR #2107, opened 2020-11-17
Fetch block witness upon receipt of NewBlockHashes msgs

gsalgado:
Before we were fetching upon receipt of NewBlock msgs, but those are only
forwarded to a subset of the peers (per the spec), so if we didn't get them
from a peer that supports the wit protocol we'd end up missing witnesses
for certain blocks

Also ensures we do not fetch witness hashes again if they are already in our DB

Includes #2106

### DIFF
--- a/trinity/components/builtin/new_block/component.py
+++ b/trinity/components/builtin/new_block/component.py
@@ -36,7 +36,7 @@ from trinity.constants import TO_NETWORKING_BROADCAST_CONFIG
 from trinity.db.manager import DBClient
 from trinity.exceptions import WitnessHashesUnavailable
 from trinity.extensibility import TrioIsolatedComponent
-from trinity.protocol.eth.events import NewBlockEvent
+from trinity.protocol.eth.events import NewBlockEvent, NewBlockHashesEvent
 from trinity.protocol.eth.payloads import (
     BlockFields,
     NewBlockHash,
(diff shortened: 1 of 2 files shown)


trinity/components/builtin/new_block/component.py
```diff
@@ -180,8 +191,18 @@ def _filter_eligible_peers(self,
 
     async def _fetch_witnesses(
             self, peer: ETHProxyPeer, block_hash: Hash32, block_number: BlockNumber) -> None:
-        # FIXME: Return immediately if we already have witness hashes for the given block hash in
-        # our wit db.
```
carver: Ah, I see, so I guess you can ignore my comment about this in #2106

trinity/components/builtin/new_block/component.py
```diff
@@ -128,8 +141,6 @@ async def _handle_new_block(self, sender: SessionAPI, payload: NewBlockPayload)
                         sender_peer_str, exc,
                     )
                 else:
-                    self.manager.run_task(
-                        self._fetch_witnesses, sender_peer, header.hash, header.block_number)
```
carver: Why stop processing it from `NewBlock` messages? It doesn't seem like it would hurt anything, and might help us if a client does something funky like skip sending `NewBlockHashes` if they send a `NewBlock`.

carver (approved):
Ignored #2106 in this review. LGTM!

trinity/components/builtin/new_block/component.py
gsalgado: Actually it'd still be possible that we get multiple `NewBlockHashes` msgs with the same block before we have a chance to fetch/store the witness hashes, so could end up making multiple requests and they'd overwrite each other's hashes. Ideally we should still fix #2105

trinity/components/builtin/new_block/component.py
gsalgado: I thought it'd be unnecessary but it turns out a client who forwards us a `NewBlock` (done before importing the block) mustn't send us a `NewBlockHashes` after importing, so it's indeed important to do it in both cases. I'll revert this change

PR #2109, opened 2020-11-19
Store union of hashes for block witness

### LINKED ISSUE #2105: Do not overwrite existing witness hashes for a given block in the DB

carver:
### What was wrong?

Fixes #2105 

### How was it fixed?

Store union of the witness hashes, when witness metadata comes in for the same block.

### To-Do

- [x] ~Test where two persists interleave their database accesses, to force the race condition that one witness overwrites the other~
- [x] ~Add a lock to pass ^ test~
- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/p2p-proto/test_wit_db.py
+++ b/tests/core/p2p-proto/test_wit_db.py
@@ -48,6 +48,50 @@ async def test_witness_for_recent_blocks():
 
 
 @pytest.mark.asyncio
+async def test_witness_history_on_repeat_blocks():
+    """
+    Repeated blocks should not consume more slots in the limited history of block witnesses
+    """
+    wit_db = AsyncWitnessDB(MemoryDB())
+    hash1 = Hash32Factory()
+    hash1_witnesses = tuple(Hash32Factory.create_batch(5))
+    await wit_db.coro_persist_witness_hashes(hash1, hash1_witnesses)
+
+    hash2 = Hash32Factory()
+    await wit_db.coro_persist_witness_hashes(hash2, tuple(Hash32Factory.create_batch(5)))
+
+    # *almost* push the first witness out of history
+    for _ in range(wit_db._max_witness_history - 2):
+        await wit_db.coro_persist_witness_hashes(Hash32Factory(), Hash32Factory.create_batch(2))
+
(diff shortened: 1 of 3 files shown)


gsalgado:
Sorry, I keep forgetting to review this. Will do so tomorrow, but just thought I'd mention I found a bug that causes `persist_witness_hashes()` to crash with a KeyError when we call it multiple times with the same hash. This test exhibits it: https://github.com/ethereum/trinity/commit/d0b33b51f19a3d84d32bf47440dcbd4b759cdfb2

Is that bug present in this branch as well?

carver:
Yup, looks like that bug is fixed in this branch. I added the test. It's similar to another test I added, but seems good to have another explicit one. :+1:

tests/core/p2p-proto/test_wit_db.py
```diff
@@ -42,3 +45,66 @@ async def test_witness_for_recent_blocks():
...
 
     assert len(wit_db._get_recent_blocks_with_witnesses()) == wit_db._max_witness_history
+
+
+@pytest.mark.asyncio
+async def test_witness_history_on_repeat_blocks():
+    """
+    Repeated blocks should not consume more slots in the limited history of block witnesses
+    """
+    wit_db = AsyncWitnessDB(AtomicDB())
```
gsalgado, line 55: AsyncWitnessDB was changed recently to wrap a regular `DatabaseAPI` instead of `AtomicDatabaseAPI`, so we could use a `MemoryDB()` here instead

gsalgado (approved):
LGTM

PR #2110, opened 2020-11-27
More peer/sync metrics

gsalgado:
- Beam sync lag (in number of blocks)
    - Block import time
    - Peer count with/without witness support

Includes #2112

### DIFF
--- a/trinity/sync/beam/chain.py
+++ b/trinity/sync/beam/chain.py
@@ -808,7 +808,8 @@ class BeamBlockImporter(BaseBlockImporter, Service):
             try:
                 wit_hashes = wit_db.get_witness_hashes(block.hash)
             except WitnessHashesUnavailable:
-                self.logger.debug("No witness hashes for block %s. Import will be slow", block)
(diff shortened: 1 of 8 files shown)


trinity/sync/beam/chain.py
```diff
@@ -771,7 +771,8 @@ async def import_block(
             try:
                 wit_hashes = wit_db.get_witness_hashes(block.hash)
             except WitnessHashesUnavailable:
-                self.logger.debug("No witness hashes for block %s. Import will be slow", block)
+                self.logger.info("No witness hashes for block %s. Import will be slow", block)
+                self.metrics_registry.counter('trinity.sync/block_witness_hashes_missing').inc()
```
carver, line 775: Worth also adding the witness present counter in the else block?

carver (approved):

trinity/sync/beam/chain.py
gsalgado: In the else block we already have two counters: `block_witness_incomplete` and `block_witness_complete`

PR #2113, opened 2020-11-30
Move witness fetcher to BeamSyncService

gsalgado:


trinity/components/builtin/new_block/component.py
```diff
@@ -102,14 +95,30 @@ async def run(self) -> None:
...
+                except trio.TooSlowError:
+                    self.logger.warning(
+                        "No subscribers for FetchBlockWitness, couldn't fetch witness for %s",
+                        humanize_hash(new_block_hash.hash),
+                    )
+                    continue
                 self.manager.run_task(
-                    self._fetch_witnesses, sender_peer, new_block_hash.hash, new_block_hash.number)
+                    self._event_bus.request,
+                    FetchBlockWitness(event.session, new_block_hash.hash, new_block_hash.number)
```
carver, line 120: The other `FetchBlockWitness` has a first parameter of `sender`. Is that what's supposed to be happening here, too?

trinity/sync/beam/service.py
```diff
@@ -128,3 +150,122 @@ async def _monitor_for_pivot(self, beam_syncer: BeamSyncer) -> bool:
...
+        if event.preferred_peer is not None:
+            try:
+                preferred_peer = cast(ETHPeer, self.peer_pool.connected_nodes[event.preferred_peer])
+            except KeyError:
+                # This means the peer has disconnected since we fired the FetchBlockWitness event.
+                pass
+            else:
+                if hasattr(preferred_peer, 'wit_api'):
+                    queried_peers.append(preferred_peer)
+                    witness_hashes = await _fetch_witness(
```
carver, line 188: I think it would be clearer to have the instance method and the unbound method use different names.

trinity/sync/beam/service.py
```diff
@@ -128,3 +150,122 @@ async def _monitor_for_pivot(self, beam_syncer: BeamSyncer) -> bool:
...
+                peer for peer in self.peer_pool.connected_nodes.values()
+                if hasattr(peer, 'wit_api') and peer not in queried_peers
+            ]
+            if not pending_peers:
+                break
+            peer = cast(ETHPeer, pending_peers[0])
+            queried_peers.append(peer)
+            witness_hashes = await _fetch_witness(
+                peer, block_hash, block_number, self.event_bus, self.base_db,
+                self.metrics_registry, self.logger)
```
carver, line 203: Don't all peers don't send a `NewBlockHash` message for every block they get? (or a `NewBlock`)

If that's right, I get the sense that we can only ask the peer who sent the `NewBlockHash` message for the witness. (Using the `NewBlockHash` as a proxy for "I have the witness available"). Then the whole method gets simpler, and I think this code block gets deleted.

trinity/sync/beam/service.py
```diff
@@ -128,3 +150,122 @@ async def _monitor_for_pivot(self, beam_syncer: BeamSyncer) -> bool:
...
+            except asyncio.TimeoutError:
+                logger.warning(
+                    "No subscribers for CollectMissingTrieNodes, cannot fetch witness for %s",
+                    block_str,
+                )
+                return witness_hashes
+            wit_db = AsyncWitnessDB(db)
+            wit_db.persist_witness_hashes(block_hash, witness_hashes)
+            result = await event_bus.request(
+                CollectMissingTrieNodes(witness_hashes, urgent, block_number))
```
carver, line 263: It's a little hard to spot because the whole method changed files, but I noticed these lines switched order. (Which seems like the right choice, now). Anything else worth pointing out that changed during the move?

carver (approved):
I got caught in a similar naming trap a little while ago as I see in this PR: the list of hashes isn't the witness. It's a meta-witness. I think it's worth trying to be ultra-clear here. For example, seeing a `BlockWitnessResult` makes me think that "we have collected all the state necessary to import the block." But I think the event actually means something like "The hashes needed to collect the state for a block are available." I suppose that could be named something like `BlockMetaWitnessCollected`.

There are a number of other places where "witness" could be switched to "meta-witness." (or "witness hashes" if you like that better)

trinity/sync/beam/service.py
carver: Ah, I see in #2114 that you want to be able to trigger a fetch without any known peer. Since that gets called rarely we hope, maybe this block is explicitly something that only happens when the sending peer is `None`?

trinity/sync/beam/service.py
gsalgado: No, it's just that before we were using a fire-and-forget event (`event_bus.broadcast(...)`) to fetch the trie nodes, whereas now we wait for them (`event_bus.request(...)`), so by persisting the hashes ASAP we make sure other `FetchBlockWitness` requests for the same block will be a no-op as we are already have one in progress

trinity/sync/beam/service.py
gsalgado: That's right, but there's also the case where the peer who sent us the `NewBlockHash` does not support the wit protocol, and in that case we'll try to fetch the meta-witness from another peer. I'll add a comment to make that clear

gsalgado:
> For example, seeing a BlockWitnessResult makes me think that "we have collected all the state necessary to import the block." But I think the event actually means something like "The hashes needed to collect the state for a block are available."

The `BlockWitnessResult` event actually means we have collected all the state necessary for the block, although only the meta-witness is included in the event. This is something that changed in this PR as before we were letting the trie nodes be fetched in the background, after returning the witness hashes, whereas now the `_fetch_witness()` function waits for the trie nodes so when it returns we have actually fetched the block witness.

> There are a number of other places where "witness" could be switched to "meta-witness." (or "witness hashes" if you like that better)

I remember you suggested that in a previous PR and I tried to keep it consistent on this one. I think the confusion here was because it was not obvious that the witness-fetching APIs now wait for the trie nodes as well

trinity/components/builtin/new_block/component.py
gsalgado: Yep, in the other one the event's session is passed in as the method's `sender` argument : https://github.com/ethereum/trinity/blob/519e4cbc67697ca410c7bc82d5d508b79669ba87/trinity/components/builtin/new_block/component.py#L94

trinity/sync/beam/service.py
gsalgado: I've renamed the method to `_ensure_witness()` as it will keep trying until it has managed to fetch the witness hashes from any of our connected peers and then wait for the actual trie nodes to arrive.

trinity/components/builtin/new_block/component.py
carver: Ooh, I didn't follow that `session` here was a handle to the peer.

trinity/sync/beam/service.py
carver: But why not wait for that other peer that supports `wit` to send us the `NewBlockHash`? I thought we were using that as the signal that the meta-witness was ready.

gsalgado: I thought it'd be possible for us to get a NewBlockHash from a non-wit peer, then forward that to our wit-supporting peers and thus cause the wit-supporting peer to not send us a NewBlockHash msg, but now that I've tried to confirm, it seems that can only happen for NewBlock so I guess in case `preferred_peer` is not None we could query just that peer. I'll make that change

gsalgado:
@carver does https://github.com/ethereum/trinity/pull/2113/commits/4ed31be86b2a8c84e92d725635e30dbdb067790a look good?

carver:
Yup, looks good! (Apart from aggravating CI failures)

PR #2114, opened 2020-11-30
BeamSyncer.import_block(): fetch witness if missing

gsalgado:
Includes #2113; only last commit needs to be reviewed


carver (approved):
Ah, I had the same issue when reading the PR title as in #2113 -- "witness" means the full state data to me. I would have understood more clearly/quickly if it said "fetch meta-witness if missing".

PR #2115, opened 2020-11-30
Fix handling of witness hashes request when they don't exist in the DB

gsalgado:


carver:
Yup, last commit looks good.

Issue #1959, opened 2020-07-28
Easier UPnP support for more routers


pipermerriam:
### What is wrong?

Currently, we have a hard coded list of service names in our UPnP code which is used to setup port forwarding.

https://github.com/ethereum/trinity/blob/0b05a02f06fabe13ce999f8ef7b4bf09ee204885/trinity/components/builtin/upnp/nat.py#L141-L145

This list was manually populated by inspecting values found on routers that I had access to.  In order to support more routers we need to gather other values used by other routers.  This isn't a trivial thing for users to do.

### How can it be fixed

We should add a subcommand to the UPnP component:

```
$ trinity export-upnp-info
```

This command would walk through all of the devices found on the network, and all of the services those devices expose and dump that data to the terminal.  This data could then be provided by a user to facilitate adding support to their router.

In addition, we can add an extra `--upnp-service-name` flag which allows the user to provide a service name which will be added to the list of known service names.  This would allow a workflow of:

- dumping the values with `$ trinity export-upnp-info`
- figuring out which one is the right one
- trying it to confirm that it works via `--upnp-service-name WANSomethingOrOther`
- reporting the new service name to be added to the main list.

And last, we can go ahead and maybe log all the service names to the debug logs unless any sees an obvious reason not to do this.

PR #2116, opened 2020-12-02
Issue #1959 easier upnp support

Elnaril:
## What was wrong?
Only a limited number of upnp services are listed to setup portforwarding.
The list is hardcoded, it is not straight forward for users to identify the one used by their routers, and it is currently not possible to provide a specific one to Trinity.
See Trinity Issue https://github.com/ethereum/trinity/issues/1959 for more details

## How was it fixed?
The proposed enhancement is divided in 2 parts:

 1/ In ethereum/upnp-port-forward/client.py:
 - A new method `fetch_add_portmapping_services` retrieves and returns all available services which can map ports.
 - The method `setup_port_map` can now receive a service name as a new argument and uses it first to when it tries to map a port.
This part has been submitted with the PR https://github.com/ethereum/upnp-port-forward/pull/5

 2/ In ethereum/trinity:
 - A new component `UpnpServiceNamesComponent` exposes the new command line positional argument `export-upnp-info`, calls `fetch_add_portmapping_services` and print the results, if any, on the terminal.
 - The component `UpnpComponent` exposes now a new command line optional argument `--upnp-service-name` which allows the user to provide a upnp service name to Trinity. This service name is provided to the function `setup_port_map`.

### To-Do

**This PR depends on https://github.com/ethereum/upnp-port-forward/pull/5 to pass the CI tests.**

- [ ] Clean up commit history

[//]: # (For important changes that should go into the release notes please add a newsfragment file as explained here: https://github.com/ethereum/upnp-port-forward/blob/master/newsfragments/README.md)

[//]: # (See: https://upnp-port-forward.readthedocs.io/en/latest/contributing.html#pull-requests)
- [ ] Add entry to the [release notes](https://github.com/ethereum/upnp-port-forward/blob/master/newsfragments/README.md)

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,6 @@ deps = {
         "SQLAlchemy>=1.3.3,<2",
         'trio>=0.16.0,<0.17',
         'trio-typing>=0.5.0,<0.6',
-        "upnpclient>=0.0.8,<1",
     ],
     'trinity': [
         "aiohttp==3.6.0",
--- a/trinity/components/builtin/upnp/component.py
+++ b/trinity/components/builtin/upnp/component.py
@@ -7,10 +7,8 @@ import logging
 
 from async_service import background_trio_service
 from lahja import EndpointAPI
-from upnp_port_forward import (
-    fetch_add_portmapping_services,
-    NoPortMapServiceFound,
-)
+from upnp_port_forward import NoPortMapServiceFound
+from upnp_port_forward.tools.export import fetch_add_portmapping_services
 
 from trinity.config import TrinityConfig
 from trinity.extensibility import (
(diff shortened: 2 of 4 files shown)


pipermerriam:
@Elnaril thank you for doing this work.  I'm having a small change of heart and thinking that I'd rather drop this into the actual upnp library itself.  You could put it in `upnp_port_forward/tools/export.py` and make it so that you can do either `python upnp_port_forward/tools/export.py` or `python -m upnp_port_forward.tools.export` and have it dump to stderr.

Elnaril:
@pipermerriam thanks for the review.
I understand that the _export_ part can be done by the `upnp-port-forward` library, but what about the `--upnp-service-name` part ?
If we don't want to be able to provide a specific service name to Trinity, do we still need to implement the corresponding feature in `upnp-port-forward` ?

pipermerriam:
Ah yes, we would still want the ability to specify the device.

Elnaril:
@pipermerriam , indeed moving the export part in `upnp-port-forward` makes great sense to me, but would it really help the users of Trinity if it's only accessible from there ?
If someone is interested in using Trinity, it may not be straightforward to think of using one of its libraries to discover the available upnp services, especially if using Docker. But it may be more common for a Trinity user to consult Trinity help and find out there is an option to do it from Trinity, even if Trinity is just a pass-through.
I understand though that this option would be a bit remote to Trinity's actual purpose ... it's just a thought to help the spreading of Trinity ...

pipermerriam:
:+1: I'm ok mirroring the export functionality here.

PR #2122, opened 2021-02-19
[POC] Implement 'proxy' mode for bridge infrastructure

pipermerriam:
A minimal POC that demonstrates using Trinity to gain access to the ETH protocol by:

1. Using infura to serve block headers
2. Disconnecting from peers that request block bodies/receipts/nodedata
3. Ignoring gossip (this should be fixed to actually implement basic gossip as to not be a leach on the network)

This is an experiment in how we might gain access to some of the gossip messages from the ETH protocol so that they can be bridged with alternate implementations like: https://notes.ethereum.org/YtI7e-R5RSK13cklq8j3Ug?view


pipermerriam:
Running this I was able to maintain a healthy peer pool of 25 peers.

Implementing minimal relaying of block gossip and transactions would make this less of a leech on the network

